PR #1, opened 2019-06-14
rehearse: Get changed periodics

droslean:

### DIFF
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -229,3 +229,29 @@ func GetPresubmitsForClusterProfiles(prowConfig *prowconfig.Config, profiles []c
 	}
 	return ret
 }
+
+// GetChangedPeriodics compares the periodic jobs from two prow configs and returns a list the changed periodics.
+func GetChangedPeriodics(prowMasterConfig, prowPRConfig *prowconfig.Config, logger *logrus.Entry) []prowconfig.Periodic {
+	var changedPeriodics []prowconfig.Periodic
+	masterPeriodicsPerName := getPeriodicsPerName(prowMasterConfig.JobConfig.AllPeriodics())
+
+	for name, job := range getPeriodicsPerName(prowPRConfig.JobConfig.AllPeriodics()) {
+		if job.Agent == string(pjapi.KubernetesAgent) {
+			masterPeriodics := masterPeriodicsPerName[name]
+			if !equality.Semantic.DeepEqual(masterPeriodics.Spec, job.Spec) {
+				logger.WithFields(logrus.Fields{logJobName: job.Name, logDiffs: convertToReadableDiff(masterPeriodics.Spec, job.Spec, objectSpec)}).Info(chosenJob)
+				changedPeriodics = append(changedPeriodics, job)
+			}
+		}
+	}
+
+	return changedPeriodics
+}
+
+func getPeriodicsPerName(periodics []prowconfig.Periodic) map[string]prowconfig.Periodic {
+	ret := make(map[string]prowconfig.Periodic, len(periodics))
+	for _, periodic := range periodics {
+		ret[periodic.Name] = periodic
+	}
+	return ret
+}
--- a/pkg/diffs/diffs_test.go
+++ b/pkg/diffs/diffs_test.go
@@ -527,3 +527,121 @@ func TestGetPresubmitsForClusterProfiles(t *testing.T) {
 		})
 	}
 }
+
+func TestGetChangedPeriodics(t *testing.T) {
+	basePeriodic := []prowconfig.Periodic{
+		{
+			JobBase: prowconfig.JobBase{
+				Agent: "kubernetes",
+				Name:  "test-base-periodic",
+				Spec: &v1.PodSpec{
+					Containers: []v1.Container{{
+						Command: []string{"ci-operator"},
+						Args:    []string{"--artifact-dir=$(ARTIFACTS)", "--target=images"},
+					}},
+				},
+			},
+		},
+	}
+
+	testCases := []struct {
+		name            string
+		configGenerator func() (before, after *prowconfig.Config)
+		expected        []prowconfig.Periodic
+	}{
+		{
+			name: "no differences mean nothing is identified as a diff",
+			configGenerator: func() (*prowconfig.Config, *prowconfig.Config) {
+				return makeConfigWithPeriodics(basePeriodic), makeConfigWithPeriodics(basePeriodic)
+			},
+			expected: nil,
+		},
+		{
+			name: "new job added",
+			configGenerator: func() (*prowconfig.Config, *prowconfig.Config) {
+				var p []prowconfig.Periodic
+				var pNew prowconfig.Periodic
+				deepcopy.Copy(&p, basePeriodic)
+
+				pNew = p[0]
+				pNew.Name = "test-base-periodic-new"
+				p = append(p, pNew)
+
+				return makeConfigWithPeriodics(basePeriodic), makeConfigWithPeriodics(p)
+
+			},
+			expected: func() []prowconfig.Periodic {
+				var p []prowconfig.Periodic
+				var pNew prowconfig.Periodic
+				deepcopy.Copy(&p, basePeriodic)
+				pNew = p[0]
+				pNew.Name = "test-base-periodic-new"
+
+				return append([]prowconfig.Periodic{}, pNew)
+			}(),
+		},
+		{
+			name: "different spec is identified as a diff - single change",
+			configGenerator: func() (*prowconfig.Config, *prowconfig.Config) {
+				var p []prowconfig.Periodic
+				deepcopy.Copy(&p, basePeriodic)
+				p[0].Spec.Containers[0].Command = []string{"test-command"}
+				return makeConfigWithPeriodics(basePeriodic), makeConfigWithPeriodics(p)
+
+			},
+			expected: func() []prowconfig.Periodic {
+				var p []prowconfig.Periodic
+				deepcopy.Copy(&p, basePeriodic)
+				p[0].Spec.Containers[0].Command = []string{"test-command"}
+				return p
+			}(),
+		},
+		{
+			name: "different spec is identified as a diff - massive changes",
+			configGenerator: func() (*prowconfig.Config, *prowconfig.Config) {
+				var p []prowconfig.Periodic
+				deepcopy.Copy(&p, basePeriodic)
+				p[0].Spec.Containers[0].Command = []string{"test-command"}
+				p[0].Spec.Containers[0].Args = []string{"testarg", "testarg", "testarg"}
+				p[0].Spec.Volumes = []v1.Volume{{
+					Name: "test-volume",
+					VolumeSource: v1.VolumeSource{
+						EmptyDir: &v1.EmptyDirVolumeSource{},
+					}},
+				}
+				return makeConfigWithPeriodics(basePeriodic), makeConfigWithPeriodics(p)
+			},
+			expected: func() []prowconfig.Periodic {
+				var p []prowconfig.Periodic
+				deepcopy.Copy(&p, basePeriodic)
+				p[0].Spec.Containers[0].Command = []string{"test-command"}
+				p[0].Spec.Containers[0].Args = []string{"testarg", "testarg", "testarg"}
+				p[0].Spec.Volumes = []v1.Volume{{
+					Name: "test-volume",
+					VolumeSource: v1.VolumeSource{
+						EmptyDir: &v1.EmptyDirVolumeSource{},
+					}},
+				}
+				return p
+			}(),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			before, after := testCase.configGenerator()
+			p := GetChangedPeriodics(before, after, logrus.NewEntry(logrus.New()))
+			if !reflect.DeepEqual(testCase.expected, p) {
+				t.Fatalf("Name:%s\nExpected %#v\nFound:%#v\n", testCase.name, testCase.expected, p)
+			}
+		})
+	}
+}
+
+func makeConfigWithPeriodics(p []prowconfig.Periodic) *prowconfig.Config {
+	return &prowconfig.Config{
+		JobConfig: prowconfig.JobConfig{
+			Periodics: p,
+		},
+	}
+}


cmd/pj-rehearse/main.go
```diff
@@ -265,37 +267,45 @@ func rehearseMain() int {
 	}
 	loggers := rehearse.Loggers{Job: logger, Debug: debugLogger.WithField(prowgithub.PrLogField, prNumber)}
 
-	toRehearse := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
-	metrics.RecordChangedPresubmits(toRehearse)
-	metrics.RecordOpportunity(toRehearse, "direct-change")
+	periodicsToRehearse := diffs.GetChangedPeriodics(masterConfig.Prow, prConfig.Prow, logger)
+	// TODO: metrics
```
stevekuznetsov: Why not here?

pkg/rehearse/jobs.go
```diff
@@ -110,24 +110,60 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
 	return &rehearsal, nil
 }
 
-func filterJobs(changedPresubmits map[string][]prowconfig.Presubmit, allowVolumes bool, logger logrus.FieldLogger) config.Presubmits {
-	ret := config.Presubmits{}
+func makeRehearsalPeriodic(source *prowconfig.Periodic, prNumber int) (prowconfig.Periodic, error) {
```
stevekuznetsov: My comment on refs seems to have been lost here

cmd/pj-rehearse/main.go
droslean: Sure, I will include the metrics in this PR.

pkg/rehearse/jobs.go
droslean: oh. I will address this.

pkg/rehearse/jobs.go
```diff
@@ -158,85 +187,141 @@ func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
...
+	loggers      Loggers
+	allowVolumes bool
+	templateMap  map[string]string
+}
+
+// NewJobConfigurer ...
+func NewJobConfigurer(presubmits config.Presubmits, periodics []prowconfig.Periodic, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool, templates []config.ConfigMapSource, profiles []config.ConfigMapSource) *JobConfigurer {
+	return &JobConfigurer{
+		presubmits:   filterPresubmits(presubmits, allowVolumes, loggers.Job),
+		periodics:    filterPeriodics(periodics, allowVolumes, loggers.Job),
```
bbguimaraes: These are only necessary in the corresponding `Configure*Rehearsal` method, they can be parameters instead of members.

pkg/rehearse/jobs.go
```diff
@@ -158,85 +187,141 @@ func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
...
+		ciopConfigs:  ciopConfigs,
+		templates:    templates,
```
bbguimaraes: If `templateMap` is filled here:
- `fillTemplateMap` won't be potentially called twice
- `templates` can be removed

pkg/rehearse/jobs.go
```diff
@@ -353,28 +438,32 @@ func replaceClusterProfiles(volumes []v1.Volume, profiles []config.ConfigMapSour
...
-	rehearsals []*prowconfig.Presubmit
-	prNumber   int
-	prRepo     string
-	refs       *pjapi.Refs
-	loggers    Loggers
-	pjclient   pj.ProwJobInterface
+	dryRun         bool
+	allowPeriodics bool
+	presubmits     []*prowconfig.Presubmit
+	periodics      []prowconfig.Periodic
```
bbguimaraes: Why `[]*` vs. `[]`?

pkg/rehearse/jobs.go
```diff
@@ -353,28 +438,32 @@ func replaceClusterProfiles(volumes []v1.Volume, profiles []config.ConfigMapSour
...
 
-	dryRun     bool
-	rehearsals []*prowconfig.Presubmit
-	prNumber   int
-	prRepo     string
-	refs       *pjapi.Refs
-	loggers    Loggers
-	pjclient   pj.ProwJobInterface
+	dryRun         bool
+	allowPeriodics bool
```
bbguimaraes: In what case is this necessary?

pkg/rehearse/jobs.go
```diff
@@ -482,10 +571,28 @@ func (e *Executor) submitRehearsals() ([]*pjapi.ProwJob, error) {
 		e.loggers.Job.WithFields(pjutil.ProwJobFields(created)).Info("Submitted rehearsal prowjob")
 		pjs = append(pjs, created)
 	}
+
+	for _, job := range e.periodics {
+
+		if !e.allowPeriodics {
+			e.loggers.Job.Warn("Periodics are not allowed to be executed.")
+			break
+		}
```
bbguimaraes: (Assuming `allowPeriodics` is necessary, see above) this can be tested once.

pkg/rehearse/jobs.go
```diff
@@ -496,3 +603,15 @@ func (e *Executor) submitRehearsal(job *prowconfig.Presubmit) (*pjapi.ProwJob, e
...
+	labels := make(map[string]string)
+	for k, v := range job.Labels {
+		labels[k] = v
+	}
+
+	prowJob := pjutil.NewProwJob(pjutil.PeriodicSpec(job), labels)
+	e.loggers.Job.WithFields(pjutil.ProwJobFields(&prowJob)).Info("Submitting a new prowjob.")
+
+	return e.pjclient.Create(&prowJob)
+}
```
bbguimaraes: Only `job` differs between these two functions, can they be DRYed?

pkg/rehearse/jobs.go
bbguimaraes: [Nevermind](https://github.com/openshift/ci-operator-prowgen/pull/175#pullrequestreview-249356080).

I would rather roll this out and have the results (and not have an extra flag), since the rehearse job is optional, but I'm fine with it.

pkg/rehearse/jobs.go
droslean: The incorrect one is the `[]*prowconfig.Presubmit` which we return [here](https://github.com/openshift/ci-tools/blob/master/pkg/rehearse/jobs.go#L90). Unfortunately,  we did more changes based on that(metrics, executor etc...) and we need another PR to clear this up.

pkg/rehearse/jobs.go
droslean: nope. the job is different type. I could use some reflection but I don't think its a good idea.

pkg/rehearse/jobs.go
droslean: it is being testing only once.. I guess its better to put it out from executor and don't include periodics in the `JobConfigurer`

pkg/rehearse/jobs.go
droslean: Let's keep it for now.

pkg/rehearse/jobs.go

pkg/rehearse/jobs.go
droslean: You are right, but I would like to keep them here for now to make the API more extensible and avoid more refactoring in the future.

pkg/diffs/diffs.go
```diff
@@ -229,3 +229,29 @@ func GetPresubmitsForClusterProfiles(prowConfig *prowconfig.Config, profiles []c
...
+	prPeriodics := getPeriodicsPerName(prowPRConfig.JobConfig.AllPeriodics())
+
+	var changedPeriodics []prowconfig.Periodic
+	for name, job := range prPeriodics {
+		if job.Agent == string(pjapi.KubernetesAgent) {
+			if !equality.Semantic.DeepEqual(masterPeriodics[name].Spec, job.Spec) {
+				logger.WithFields(logrus.Fields{logJobName: job.Name, logDiffs: convertToReadableDiff(masterPeriodics[name].Spec, job.Spec, objectSpec)}).Info(chosenJob)
+				changedPeriodics = append(changedPeriodics, job)
+			}
+		}
```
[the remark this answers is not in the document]
droslean: its only 1 line and also the logger fields are different. Its not worth it to dry it.

pkg/diffs/diffs.go
```diff
@@ -229,3 +229,29 @@ func GetPresubmitsForClusterProfiles(prowConfig *prowconfig.Config, profiles []c
...
+// GetChangedPeriodics returns the changed presubmits mapped by their name.
+func GetChangedPeriodics(prowMasterConfig, prowPRConfig *prowconfig.Config, logger *logrus.Entry) []prowconfig.Periodic {
+	masterPeriodics := getPeriodicsPerName(prowMasterConfig.JobConfig.AllPeriodics())
+	prPeriodics := getPeriodicsPerName(prowPRConfig.JobConfig.AllPeriodics())
+
+	var changedPeriodics []prowconfig.Periodic
+	for name, job := range prPeriodics {
+		if job.Agent == string(pjapi.KubernetesAgent) {
+			if !equality.Semantic.DeepEqual(masterPeriodics[name].Spec, job.Spec) {
+				logger.WithFields(logrus.Fields{logJobName: job.Name, logDiffs: convertToReadableDiff(masterPeriodics[name].Spec, job.Spec, objectSpec)}).Info(chosenJob)
```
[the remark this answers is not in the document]
droslean: addressed

droslean:
This should be ready. Only one thing that @stevekuznetsov mentioned, regarding how we want to treat the `git-ref` in the periodic jobs. I found out that most of the periodics are already using the `git-ref` and also some of them the have also the `extra_refs`. In my point of view we should leave them untouched, and leave the configuration to the user. 
Also, another problem is how we will give information to the user for the rehearsed periodic job. @petr-muller  suggested to have a presubmits that watches a periodic job, which is good idea. Other simplier way is to output in the deck's url.

pkg/diffs/diffs.go
```diff
@@ -229,3 +229,29 @@ func GetPresubmitsForClusterProfiles(prowConfig *prowconfig.Config, profiles []c
 	}
 	return ret
 }
+
+// GetChangedPeriodics returns the changed presubmits mapped by their name.
+func GetChangedPeriodics(prowMasterConfig, prowPRConfig *prowconfig.Config, logger *logrus.Entry) []prowconfig.Periodic {
+	masterPeriodics := getPeriodicsPerName(prowMasterConfig.JobConfig.AllPeriodics())
+	prPeriodics := getPeriodicsPerName(prowPRConfig.JobConfig.AllPeriodics())
```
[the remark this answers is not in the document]
bbguimaraes: It is not necessary to convert `prowPRConfig` to a map, it can be iterated over directly.

pkg/diffs/diffs.go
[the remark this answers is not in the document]
bbguimaraes: The entire inner loop is the same (or at least it should be, I don't know why they are different) and could be a separate function called by both functions.

pkg/rehearse/jobs.go
bbguimaraes: `templates` can be removed.

pkg/rehearse/jobs.go
bbguimaraes: The functions are now so small that the duplication is fine.

pkg/rehearse/jobs.go
droslean: oh boy, I forgot it. addressed now.

stevekuznetsov:
We should update `crier` upstream to have them report the same way postsubmits and presubmits do if they have `extra_refs` set. Until then, we should report it in the `rehearse` main job.

pkg/rehearse/jobs.go
```diff
@@ -110,45 +110,76 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
...
 		return fmt.Errorf("cannot rehearse jobs that have Command different from simple 'ci-operator'")
 	}
 
-	for _, arg := range container.Args {
-		if strings.HasPrefix(arg, "--git-ref") || strings.HasPrefix(arg, "-git-ref") {
-			return fmt.Errorf("cannot rehearse jobs that call ci-operator with '--git-ref' arg")
+	if !allowGitRef {
+		for _, arg := range container.Args {
+			if strings.HasPrefix(arg, "--git-ref") || strings.HasPrefix(arg, "-git-ref") {
+				return fmt.Errorf("cannot rehearse jobs that call ci-operator with '--git-ref' arg")
```
stevekuznetsov: Why do we allow this for periodics?

pkg/rehearse/jobs.go
```diff
@@ -158,85 +189,133 @@ func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
...
+
+			env.Value = string(ciOpConfigContent)
+			env.ValueFrom = nil
 		}
 	}
+	return nil
+}
 
-	return &rehearsal, nil
+// JobConfigurer ...
```
stevekuznetsov: TODO?

pkg/rehearse/jobs.go
```diff
@@ -158,85 +189,133 @@ func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
...
-// ConfigureRehearsalJobs filters the jobs that should be rehearsed, then return a list of them re-configured with the
-// ci-operator's configuration inlined.
-func ConfigureRehearsalJobs(toBeRehearsed config.Presubmits, ciopConfigs config.CompoundCiopConfig, prNumber int, loggers Loggers, allowVolumes bool, templates []config.ConfigMapSource, profiles []config.ConfigMapSource) []*prowconfig.Presubmit {
-	var templateMap map[string]string
-	if allowVolumes {
-		templateMap = make(map[string]string, len(templates))
-		for _, t := range templates {
-			templateMap[filepath.Base(t.Filename)] = t.TempCMName("template")
-		}
+// NewJobConfigurer ...
```
stevekuznetsov: TODO?

pkg/rehearse/jobs.go
```diff
@@ -158,85 +189,133 @@ func filterJob(source *prowconfig.Presubmit, allowVolumes bool) error {
...
-			}
-
 			jobLogger.WithField(logRehearsalJob, rehearsal.Name).Info("Created a rehearsal job to be submitted")
 			rehearsals = append(rehearsals, rehearsal)
 		}
 	}
+	return rehearsals
+}
+
+// ConfigurePeriodicRehearsals ...
```
stevekuznetsov: TODO?

pkg/rehearse/jobs.go
droslean: Most of our periodics, already have `git-ref`. I guess this is how our stakeholders wanted to configure. Also, the `extra_refs` used to be ignore by ci-operator but this have been fixed by now. Shall we force them do otherwise?

pkg/rehearse/jobs.go
droslean: jesus.. let me address them

pkg/rehearse/jobs.go
droslean: meaning that we need to change all of our current periodics in release repo.

stevekuznetsov: Let's make them opt-in. This is why. Opt in and follow the rules if you want to be rehearsed.

stevekuznetsov:
@petr-muller @bbguimaraes would love your thoughts to add LGTM here

petr-muller:
> @petr-muller would love your thoughts to add LGTM here

Ack, I plan to have a look today

bbguimaraes:
LGTM, but I still think we are adding too much duplication, and the kind that can generate bugs and/or make maintenance unnecessarily harder.

I would much rather see functions like these instead of duplicate code (not a mandate, take them as suggestions):

```go
// pkg/diff/diff.go
func jobChanged(name string, job0, job1 *prowconfig.JobBase, logger *logrus.Entry, logFields *logrus.Fields) bool {
       if job1.Agent != string(pjapi.KubernetesAgent) {
               return false
       }
       if job0.Agent != string(pjapi.KubernetesAgent) {
               logger.WithFields(*logFields).WithFields(logrus.Fields{logJobName: name, logDiffs: convertToReadableDiff(job0.Agent, job1.Agent, objectAgent)}).Info(chosenJob)
               return true
       }
       if !equality.Semantic.DeepEqual(job0.Spec, job1.Spec) {
               logger.WithFields(*logFields).WithFields(logrus.Fields{logJobName: name, logDiffs: convertToReadableDiff(job0.Spec, job1.Spec, objectSpec)}).Info(chosenJob)
               return true
       }
       return false
}

// pkg/rehearse/jobs.go
func makeRehearsalJob(job *prowconfig.JobBase, prNumber int, name, repo, branch string) {
       job.Name = fmt.Sprintf("rehearse-%d-%s", prNumber, name)
       gitrefArg := fmt.Sprintf("--git-ref=%s@%s", repo, branch)
       job.Spec.Containers[0].Args = append(job.Spec.Containers[0].Args, gitrefArg)
       if job.Labels == nil {
               job.Labels = make(map[string]string, 1)
       }
       job.Labels[rehearseLabel] = strconv.Itoa(prNumber)
}
```

droslean:
@bbguimaraes Will do a follow up for this.

droslean:
After discussing this with @petr-muller we decide to break this PR in smaller ones to make the `periodics` introduction to the rehearse tool more stable and easy to follow up.

PR #6, opened 2019-06-21
prowgen: add IaaS-agnostic installer test type

bbguimaraes:


cmd/ci-operator-prowgen/main.go
```diff
@@ -27,6 +27,29 @@ const (
 	sentryDsnSecretName = "sentry-dsn"
 	sentryDsnMountPath  = "/etc/sentry-dsn"
 	sentryDsnSecretPath = "/etc/sentry-dsn/ci-operator"
+
+	openshiftInstallerRandomCmd = `set -eux
+target=$(awk < /usr/local/e2e-targets \
```
petr-muller: Ugh, awk. [If only we had Python available](https://docs.python.org/3/library/random.html#random.choices) :(

I don't like the fact that we have duplicate lists of platforms (here and in the `/usr/local/e2e-targets` weights file). I'm thinking we could make `/usr/local/e2e-targets` a sourceable bash file, covering L32-L41 of this file. Then we would just source it here and it would set `template ` and `CLUSTER_TYPE` accordingly.

cmd/ci-operator-prowgen/main.go
```diff
@@ -84,7 +107,7 @@ func (o *options) process() error {
 // Generate a PodSpec that runs `ci-operator`, to be used in Presubmit/Postsubmit
 // Various pieces are derived from `org`, `repo`, `branch` and `target`.
 // `additionalArgs` are passed as additional arguments to `ci-operator`
-func generatePodSpec(info *config.Info, target string, additionalArgs ...string) *kubeapi.PodSpec {
+func generatePodSpec(info *config.Info, target string, cmd, args []string, additionalArgs ...string) *kubeapi.PodSpec {
```
petr-muller: I don't like how this this makes all default case callsites cluttered with `nil`, `nil`. I like the advice in https://dave.cheney.net/practical-go/presentations/qcon-china.html#_design_apis_for_their_default_use_case. I think the only caller that cares about `cmd` and `args` (`generatePodSpecRandom`) can overwrite the `Containers[0].Command` and `Containers[0].Args` fields of the returned podspec, eliminating the impact on other callsites.

cmd/ci-operator-prowgen/main.go
```diff
@@ -277,6 +305,78 @@ func generatePodSpecTemplate(info *config.Info, release string, test *cioperator
 	return podSpec
 }
 
+func generatePodSpecRandom(info *config.Info, test *cioperatorapi.TestStepConfiguration) *kubeapi.PodSpec {
+	cmd := []string{"-c", fmt.Sprintf(openshiftInstallerRandomCmd, test.As)}
+	podSpec := generatePodSpec(info, test.As, []string{"bash"}, cmd)
+	profiles := []string{"aws", "azure4", "vsphere"}
```
petr-muller: I think this should be extracted somewhere more prominent to be more easily found when platforms are added / removed.

cmd/ci-operator-prowgen/main.go
bbguimaraes: A couple of points (in decreasing order of importance):
- What's wrong with awk? =)
- This is a quick-and-dirty (actually, the quickest-and-dirtiest) implementation: the goal of the task is to get the jobs in place as soon as possible, then refine. The job is generated by `prowgen`, so changing the implementation is trivial.
- I treated the format of `e2e-targets` as definitive. If/when this becomes Go code, maybe YAML would make more sense. We can bikeshed the exact format, but I went for "as-simple-as-possible plain text". I think we specially don't want to tie it to a specific implementation, as that is very likely to change.
- Isn't it available? I assumed it were, but even then I still think `awk` is unbeatable for ad-hoc text processing. The `python` version (with `random.choices`, an equivalent implementation would be even longer):

```python
import random, sys
ws, ns = zip(*((float(w), n) for (w, n) in map(str.split, sys.stdin)))
print(random.choices(ns, weights=ws)[0])
```

*sorry for the little essay, I wanted to answer other questions that were likely to come up*

cmd/ci-operator-prowgen/main.go
bbguimaraes: Heh, I spent ~5ms thinking about the interface. Agreed, I will change it to something more decent.

cmd/ci-operator-prowgen/main.go
bbguimaraes: Hmm, I certainly should have used the constants we already have, now that I think about it. I'll move this to a constant.

bbguimaraes:
Comments addressed.

I am keeping `awk` for now to get this merged ASAP. I developed a [Go prototype](https://github.com/bbguimaraes/ci-tools/commit/1cc0975a0cea40a7e7351eed44f5513ba375ded0) that we can transition to once the jobs are in place.

PR #9, opened 2019-06-27
ci-operator: check extra refs in gitSourceStep

bbguimaraes:
Treat extra refs the same way as other steps (e.g. source).  Found when
testing a periodic job with a `project_image` `build_root`.


bbguimaraes:
Added as an ad-hoc check to unblock people testing their jobs, but we should find a way to treat this properly to avoid adding this code to every step that touches refs and extra refs.

PR #12, opened 2019-06-28
prowgen: DRY out JobBase generation

petr-muller:
Generate the JobBase parts of both Presubmits and Postsubmit in a single
method.

Additionally, remove some of the regex-handling code for Postsubmits. We
never passed anything else than `true` to the `treatBranchesAsExplicit`
param, so I've removed the parameter and simplified the API. Also, we
took some measures to handle regex-y branches, but the branches come
from **ci-operator** config (not Prowjob config), so they are never
regexes, so we do not need to handle that.

/cc @stevekuznetsov @droslean @bbguimaraes @hongkailiu

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -28,6 +28,9 @@ const (
 	sentryDsnMountPath  = "/etc/sentry-dsn"
 	sentryDsnSecretPath = "/etc/sentry-dsn/ci-operator"
 
+	presubmitPrefix  = "pull"
+	postsubmitPrefix = "branch"
+
 	openshiftInstallerRandomCmd = `set -eux
 target=$(awk < /usr/local/e2e-targets \
     --assign "r=$RANDOM" \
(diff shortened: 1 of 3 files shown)


cmd/ci-operator-prowgen/main.go
```diff
@@ -399,18 +401,25 @@ func generatePresubmitForTest(name string, info *config.Info, podSpec *kubeapi.P
...
+		UtilityConfig: prowconfig.UtilityConfig{
+			DecorationConfig: &v1.DecorationConfig{SkipCloning: &newTrue},
+			Decorate:         true,
+		},
+	}
+}
 
+func generatePresubmitForTest(name string, info *config.Info, podSpec *kubeapi.PodSpec) *prowconfig.Presubmit {
+	if len(info.Variant) > 0 {
+		name = fmt.Sprintf("%s-%s", info.Variant, name)
```
stevekuznetsov: Why is this more appropriate outside of `generateJobBase`?

petr-muller: We need to have this string on this level to make the Reporter/RerunCommand/Trigger properties of presubmits (L425-L429).

PR #13, opened 2019-06-28
Prowgen: Rewrite pruning stale jobs 

petr-muller:
First commit is https://github.com/openshift/ci-tools/pull/12

We tried to be smart before, but pruning jobs when we are visiting a
file has problems. First, we miss pruning opportunities, second, we
badly treat cases when multiple sources generate jobs into the same
file (like variant jobs).

Add a (default on, but possible to suppress) separate pruning pass after
we generate the jobs. The jobs are labelled with a different label when
we generate them (meaning "we generated this now"). After the generation
pass, we walk everything again and remove the jobs that were not
generated in this run. This also imprves the pruning so that it removes
files with empty jobconfig.

I intend to add few more integration tests for pruning for the usual cases
(removal of promotion, removal of ci-op config, variants), but at last the code
is here for review

/cc @stevekuznetsov @bbguimaraes @droslean @hongkailiu


pkg/jobconfig/files.go
```diff
@@ -499,3 +474,81 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
+}
+
+func isGenerated(job prowconfig.JobBase) bool {
+	_, generated := job.Labels[ProwJobLabelGenerated]
+	return generated
+}
+
+func PruneStaleJobs(jobDir string, oldLabel, newLabel ProwgenLabel) error {
+	if err := OperateOnJobConfigDir(jobDir, func(jobConfig *prowconfig.JobConfig, info *Info) error {
+		if info.Type == "periodics" {
```
stevekuznetsov: Technically this isn't necessary right? We do nothing without the labels

pkg/jobconfig/files.go
```diff
@@ -499,3 +474,81 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
+	genLabel, generated := job.Labels[ProwJobLabelGenerated]
+	return generated && genLabel != string(newLabel)
+}
+
+func isGenerated(job prowconfig.JobBase) bool {
+	_, generated := job.Labels[ProwJobLabelGenerated]
+	return generated
+}
+
+func PruneStaleJobs(jobDir string, oldLabel, newLabel ProwgenLabel) error {
```
stevekuznetsov: Why are the labels parameterized here?

pkg/jobconfig/files.go
```diff
@@ -499,3 +474,81 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
+				}
+				if isGenerated(job.JobBase) {
+					job.Labels[ProwJobLabelGenerated] = string(oldLabel)
+				}
+
+				pruned.Postsubmits[repo] = append(pruned.Postsubmits[repo], job)
+			}
+		}
+
+		for repo, jobs := range pruned.Presubmits {
```
stevekuznetsov: You only add to `pruned.Presubmits` when you have a reason to so I don't think you'll ever hit a case where you have an empty slice?

pkg/jobconfig/files.go
petr-muller: It's not necessary, but it causes a read-and-write on periodics, which makes noise in the repo (periodics are handcrafted, so it fixes ordering, removes omitempty fields...).

pkg/jobconfig/files.go
petr-muller: Right. A remnant from previous solution :)

pkg/jobconfig/files.go
petr-muller: I did not want to make pkg/jobconfig/files.go` coupled with prowgen `main.go` (more than it already is). The code in `main.go` determines how exactly are the jobs labelled, so I wanted to avoid the pkg to rely on that.

pkg/jobconfig/files.go
stevekuznetsov: ah

pkg/jobconfig/files.go
stevekuznetsov: But we won't ever prune with a different set of labels so the urge not to couple seems misplaced?

petr-muller: I wouldn't say misplaced, but I hear what you say. How about I move this method to prowgen `main` and simplify it?

petr-muller:
The PR actually pruning some stale jobs: https://github.com/openshift/release/pull/4254. It needs to be applied when this one merges to avoid breaking openshift/release for everyone.

petr-muller:
Addressed feedback, also refactored so the pruning is testable and wrote some tests.

bbguimaraes:
While this does make things better, I am not sure it addresses the problem effectively. But a few questions first:

Given

```
$ find ci-operator/config/ -name '*__*'
$ git log -n 1 -- '*__*'
commit 2d19d03cd3f852c701f846a5d25b4c2b3d914a2f
Author: Clayton Coleman <ccoleman@redhat.com>
Date:   Sat Feb 16 13:45:18 2019 -0500

    Remove the rhel variant jobs - origin will build from ubi instead

    We'll build and publish a consistent set of output.
```

Do we still care about variants?  If those are removed, do we go back to a one-to-one `ci-operator/config/$x` → `ci-operator/jobs/$x` mapping?

Do we care that this is also incompatible with the `--from-file` flag?

```
$ ci-operator-prowgen --from-file ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml --to-dir ci-operator/jobs/ 2> /dev/null
$ git diff --shortstat
 1079 files changed, 196 insertions(+), 128310 deletions(-)
```

petr-muller:
> While this does make things better, I am not sure it addresses the problem effectively.

Yeah, it does two passes. I know. I believe it's still good enough, it still runs in few seconds. You could do it in one pass if you loaded everything into memory, generated new stuff, pruned it and then wrote everything, but then it would basically be a full rewrite, which I did not want to do.

> Do we still care about variants? If those are removed, do we go back to a one-to-one `ci-operator/config/$x` → `ci-operator/jobs/$x` mapping?

Variants are still occasionally useful to someone. In fact, @stevekuznetsov recently mentioned he suggested them to someone, it did not work because of pruning and that triggered me enough to fix the pruning.

> Do we care that this is also incompatible with the `--from-file` flag?

I cared enough to add a switch to suppress pruning. Maybe `--from-file` should imply `--prune=false`?

stevekuznetsov:
> I cared enough to add a switch to suppress pruning. Maybe --from-file should imply --prune=false?

If they are incompatible, yes it should. This otherwise LGTM

bbguimaraes:
Sorry, I didn't have time to respond earlier.

> > While this does make things better, I am not sure it addresses the problem effectively.
>
> Yeah, it does two passes. I know. I believe it's still good enough, it still runs in few seconds.

I didn't mean "efficiently", "A few seconds" is fine with me, even if we are doubling it.

> > Do we care that this is also incompatible with the `--from-file` flag?
>
> I cared enough to add a switch to suppress pruning. Maybe `--from-file` should imply `--prune=false`?

Then it will not remove old jobs — obviously, but that's why it's incompatible: there's no good answer.

> You could do it in one pass if you loaded everything into memory, generated new stuff, pruned it and then wrote everything, but then it would basically be a full rewrite, which I did not want to do.

I asked if variants were still worth maintaining because they were what really broke the algorithm we have, based on `OperateOnCIOperatorConfigDir`.

I am fine with adding this as a compensation for that problem. I am just arguing we won't have an appropriate solution until we reconsider that part of the code.

/lgtm

petr-muller:
> Then it will not remove old jobs — obviously, but that's why it's incompatible: there's no good answer.

Maybe we could just kill `--from-file`; I wonder how much it is actually used.

> > You could do it in one pass if you loaded everything into memory, generated new stuff, pruned it and then wrote everything, but then it would basically be a full rewrite, which I did not want to do.
> 
> I asked if variants were still worth maintaining because they were what really broke the algorithm we have, based on `OperateOnCIOperatorConfigDir`.
> 
> I am fine with adding this as a compensation for that problem. I am just arguing we won't have an appropriate solution until we reconsider that part of the code.

I'm pragmatic here. Prowgen works well enough to not cause trouble and investing effort into internal changes does not bring much value IMHO. I was thinking about the "proper" approach but it's not straightforward too (dealing with handcrafted jobs is harder in that model). For me it's simply not worth the effort.

PR #14, opened 2019-07-03
prowgen: make --from-file imply --prune=false

petr-muller:
It's better to not prune stale jobs than remove the existing ones. We could also consider removing `--from-file` entirely, WDYT?

Also clean up some imports.

/cc @bbguimaraes @stevekuznetsov

PR #16, opened 2019-07-04
prowgen: operate on sub-directories

bbguimaraes:
I decided to give this a time limit of 30m and… it was much easier to implement than I had initially thought. If people are happy with the overall design and implementation, I will work on making it prettier (documentation, comments, etc.).

As discussed in https://github.com/openshift/ci-tools/pull/13 and https://github.com/openshift/ci-tools/pull/14, `--from-file` is incompatible with our new pruning method. It does have advantages, though, that I think we should maintain. Namely:

- Speed: it is ~70x faster on my machine, although the difference is seconds vs. centiseconds.
- Scope: it allows someone working on jobs in a single repository (what I believe is our most common non-expert usage) to focus on their jobs. See https://github.com/openshift/ci-tools/pull/14#issuecomment-508120974.  Correctness is still enforced by the pre-submit check in `openshift/release`, which remains unchanged.

The problem with `--from-file`, AFAICT, is that variants broke the direct mapping we had from `ci-operator` configuration files to job files. We can no longer safely consider each input file in isolation, which is the basis of our current iteration algorithm.

We can, however, consider a *sub-directory* in isolation: all variants come from the same leaf directory. This change makes the iteration on both generation and pruning look only at the directories passed as non-flag arguments (previously ignored), similar to `git` commands, e.g. `git diff -- dir0 dir1`.

Backward compatibility is maintained by iterating over the entire tree when no argument is given. All tests from https://github.com/openshift/ci-tools/pull/15 pass. I've based this PR on them, the original changes are:

` 3 files changed, 26 insertions(+), 18 deletions(-)`


petr-muller:
> We can, however, consider a _sub-directory_ in isolation: all variants come from the same leaf directory.

Very nice! I like the approach.

bbguimaraes:
So… I decided to read our documentation. What a concept, I know. Actually, I opened the file to remove the references to `--from-file`. But it turns out we've always supported passing a sub-directory to `--from-dir`: https://github.com/openshift/ci-tools/blob/6718aa2f9936aa2a55ef2d555c23ba602fbbfef3/CI_OPERATOR_PROWGEN.md#generate-prow-jobs-for-new-ci-operator-config.

I tested it and it works the same as my implementation (unsurprisingly, in retrospect), except for the pruning stage. I will, in that order, fix pruning, remove the extra code for handling non-flag arguments, and never again tell anyone to RTFM.

cmd/ci-operator-prowgen/main.go
```diff
@@ -506,8 +501,9 @@ func generateJobs(
 
 // generateJobsToDir returns a callback that knows how to generate prow job configuration
 // into the dir provided by consuming ci-operator configuration
-func generateJobsToDir(dir string, label jc.ProwgenLabel) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
+func generateJobsToDir(dir string, label jc.ProwgenLabel, visited sets.String) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 	return func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
+		visited.Insert(filepath.Join(info.Org, info.Repo))
```
petr-muller: This unfortunately does not cover (uncommon but real) corner case when some changes make us to stop visiting a directory entirely, which happens when we remove all config for a given component. Sometimes this happens when repositories are renamed or archived (I remember both happening).

bbguimaraes: Heh, I realized that five seconds after pushing, but didn't think people would be monitoring the PR =)

I'll add an integration test for that. I had a hard time yesterday trying to reconcile `--from-dir` with `--prune`, in the end I think I'll have to revert to my previous implementation with separate arguments.

I'll see if I can find some time in the next few days, I'm trying not to spend too much time on this.

bbguimaraes:
- rebased
- added a configuration removal test
- reverted to using separate arguments
- added an integration test
- removed `--from-file` because it cannot work with pruning
- removed `--prune` becase pruning should now be safe for all operations
- removed references to `--from-dir path/to/subdir` from the documentation

The last two items can be reverted if there are objections.

The reason for reverting to separate arguments is that I could not find a way, with just `--from-dir` and `--to-dir`, to determine how to limit pruning without using heuristics while scanning the input directory. That becomes straightforward when the arguments are names of directories under both the input and output directories.

bbguimaraes:
Commit order because Github has one job and fails at it:

`$ git log --oneline --reverse master..`
b4ee9363 prowgen: test complete removal
a5f3f02c prowgen: remove --from-file
99d497cf prowgen: remove --prune
605ba962 prowgen: operate on sub-directories

PR #24, opened 2019-07-18
introduce generic test image test type

bparees:


pkg/api/types.go
```diff
@@ -440,6 +441,16 @@ type OpenshiftInstallerUPIClusterTestConfiguration struct {
 // chosen randomly and runs conformance tests.
 type OpenshiftInstallerRandomClusterTestConfiguration struct{}
 
+// OpenshiftInstallerCustomTestImageClusterTestConfiguration describes a
+// test that provisions a cluster using openshift-installer and
+// executes a command in the image specified by the job configuration.
+type OpenshiftInstallerCustomTestImageClusterTestConfiguration struct {
```
stevekuznetsov: Do we need a new type? Why not just add the field to `OpenshiftInstallerSrcClusterTestConfiguration` and default to `LOCAL_IMAGE_SRC`?

bparees: discussed in slack.  "because LOCAL_IMAGE_SRC is a special case hack" is the short summary.

PR #26, opened 2019-07-22
Build ipi_deprovision image

hongkailiu:
https://jira.coreos.com/browse/DPTP-471

### DIFF
--- a/images/ipi-deprovision/Dockerfile
+++ b/images/ipi-deprovision/Dockerfile
@@ -1,0 +1,13 @@
+FROM registry.svc.ci.openshift.org/openshift/hive-v4.0:hive
+LABEL maintainer="hongkliu@redhat.com"
+
+COPY cmd/ipi-deprovision/ipi-deprovision.sh  /usr/bin/ipi-deprovision.sh
+
+# Install awscli
+RUN chmod +x /usr/bin/ipi-deprovision.sh && \
+      curl -O https://bootstrap.pypa.io/get-pip.py && \
+      python get-pip.py && \
+      rm -fv get-pip.py && \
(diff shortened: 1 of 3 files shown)


images/ipi-deprovision/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM registry.svc.ci.openshift.org/openshift/hive-v4.0:hive
+LABEL maintainer="hongkliu@redhat.com"
+
+COPY ipi-deprovision/ipi-deprovision.sh  /usr/bin/ipi-deprovision.sh
```
petr-muller: This assumes the context directory is `cmd`, I think this should be `cp cmd/ipi-deprovision/ipi-deprovision.sh destination`

images/ipi-deprovision/Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+
+# Install awscli
+RUN chmod +x /usr/bin/ipi-deprovision.sh && \
+      curl -O https://bootstrap.pypa.io/get-pip.py && \
+      python get-pip.py && \
+      rm -fv get-pip.py && \
+      pip install awscli
```
petr-muller: I'm wondering whether we should peg this to a specific version to have control over what we are building upon.

images/ipi-deprovision/Dockerfile
hongkailiu: Done.

images/ipi-deprovision/Dockerfile
hongkailiu: Done.

PR #29, opened 2019-07-23
Install less in ipi-deprivision image

hongkailiu:
less is required by aws cli
https://unix.stackexchange.com/questions/295340/how-to-install-less-on-oracle-linux-server

### DIFF
--- a/images/ipi-deprovision/Dockerfile
+++ b/images/ipi-deprovision/Dockerfile
@@ -1,6 +1,8 @@
 FROM registry.svc.ci.openshift.org/openshift/hive-v4.0:hive
 LABEL maintainer="hongkliu@redhat.com"
 
+ARG JQ_VERSION=1.6
+
 COPY cmd/ipi-deprovision/ipi-deprovision.sh  /usr/bin/ipi-deprovision.sh
 
 # Install awscli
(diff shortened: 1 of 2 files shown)


images/ipi-deprovision/Dockerfile
```diff
@@ -1,13 +1,21 @@
...
 COPY cmd/ipi-deprovision/ipi-deprovision.sh  /usr/bin/ipi-deprovision.sh
 
 # Install awscli
 RUN chmod +x /usr/bin/ipi-deprovision.sh && \
       curl -O https://bootstrap.pypa.io/get-pip.py && \
       python get-pip.py && \
       rm -fv get-pip.py && \
-      pip install awscli==1.16.202
+      pip install awscli==1.16.202 && \
+      curl -L -o jq "https://github.com/stedolan/jq/releases/download/jq-${JQ_VERSION}/jq-linux64" && \
```
stevekuznetsov: On Fedora at least, `jq` is an RPM -- is it not on CentOS?

stevekuznetsov: Or from `pip`?

hongkailiu: only my 2 cents ... our base image is our beloved rhel image.
`jq` in fedora/centos can come with yum/dnf. In the case of centos, it comes from epel-release which i do not want to install :) https://serverfault.com/questions/768026/how-to-install-jq-on-rhel6-5

Did not go for with fedora although the jq in my pc is from dnf.

hongkailiu: Forgot to mention, i cannot test this locally with docker image directly because the yum repo url points to a svc in cluster. Silly me, I must have known this already.

stevekuznetsov: OK, that makes sense

PR #30, opened 2019-07-23
Check if AWS_SHARED_CREDENTIALS_FILE is set

hongkailiu:
follow-up of https://github.com/openshift/release/pull/4489

### DIFF
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -1,15 +1,23 @@
 #!/bin/sh
 
+set -o errexit
+set -o nounset
+set -o pipefail
+
 if [ -z "${AWS_SHARED_CREDENTIALS_FILE}" ]; then echo "ENV VAR AWS_SHARED_CREDENTIALS_FILE is not set"; exit 1; fi
 
 ### expirationDate is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
 MY_DATE=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
+### TODO: ask @abhinav.dahiya the right list of regions we should focus on
+### here is only for test
 declare -a regions=("us-east-1" "us-east-2" "us-west-1")
 
 handle_cluster () {
   local cluster_name;
   cluster_name=$1
   echo "handling cluster: ${cluster_name} ..."
+  ### TODO: will test hiveutil later
+  #./bin/hiveutil aws-tag-deprovision "${cluster_name}=owned"
 }
 
 for r in "${regions[@]}"
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+if [ -z "${AWS_SHARED_CREDENTIALS_FILE}" ]; then echo "ENV VAR AWS_SHARED_CREDENTIALS_FILE is not set"; exit 1; fi
+
 ### expirationDate is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
 MY_DATE=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
 declare -a regions=("us-east-1" "us-east-2" "us-west-1")
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -1,27 +1,42 @@
-#!/bin/sh
+#!/bin/bash
 
 set -o errexit
 set -o nounset
 set -o pipefail
 
-if [ -z "${AWS_SHARED_CREDENTIALS_FILE}" ]; then echo "ENV VAR AWS_SHARED_CREDENTIALS_FILE is not set"; exit 1; fi
+### cluster_age_cutoff is going to be used as argument in the aws command below
+### to filter out the expired clusters with the tags of the key expirationDate
+### which is always with timezone +00:00, eg, "2019-07-23T22:35+0000".
+### CLUSTER_TTL controlls the relative time to the running time of this script.
+### eg, export CLUSTER_TTL="22 hours ago"
+echo "Searching for clusters with a TTL of ${CLUSTER_TTL}"
+cluster_age_cutoff="$(TZ=":Africa/Abidjan" date --date="${CLUSTER_TTL}" '+%Y-%m-%dT%H:%M+0000')"
 
-### expirationDate is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
-MY_DATE=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
-### TODO: ask @abhinav.dahiya the right list of regions we should focus on
-### here is only for test
-declare -a regions=("us-east-1" "us-east-2" "us-west-1")
+echo "cluster_age_cutoff: ${cluster_age_cutoff}"
+
+### eg, export AWS_REGIONS="us-east-1;us-west-1"
+echo "take regions from env. var.: AWS_REGIONS: ${AWS_REGIONS}"
+IFS=';' read -r -a regions <<< "$AWS_REGIONS"
 
 handle_cluster () {
-  local cluster_name;
+  local cluster_name
   cluster_name=$1
-  echo "handling cluster: ${cluster_name} ..."
+  local expirationDateValue
+  expirationDateValue=$2
+  echo "handling cluster: ${cluster_name} with expirationDate: ${expirationDateValue}"
   ### TODO: will test hiveutil later
   #./bin/hiveutil aws-tag-deprovision "${cluster_name}=owned"
 }
 
+
+
 for r in "${regions[@]}"
 do
   echo "doing region ${r} ..."
-  aws ec2 describe-vpcs --output json --region "${r}" | jq --arg date "${MY_DATE}" -r -S '.Vpcs[] | select (.Tags[]? | (.Key == "expirationDate" and .Value < $date)) | .Tags[] | select (.Value == "owned") | .Key' | while read line; do handle_cluster ${line}; done
+  json_output=$(aws ec2 describe-vpcs --output json --region "${r}")
+  for cluster in $(echo ${json_output} | jq --arg date "${cluster_age_cutoff}" -r -S '.Vpcs[] | select (.Tags[]? | (.Key == "expirationDate" and .Value < $date)) | .Tags[] | select (.Value == "owned") | .Key')
+  do
+    expirationDateValue=$(echo ${json_output} | jq --arg cl "${cluster}" -r -S '.Vpcs[] | select (.Tags[]? | (.Key == $cl and .Value == "owned")) | .Tags[] | select (.Key == "expirationDate") | .Value')
+    handle_cluster "${cluster}" "${expirationDateValue}"
+  done
 done


cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,7 +1,16 @@
 #!/bin/sh
 
+set -o errexit
+set -o nounset
+set -o pipefail
+
+export AWS_SHARED_CREDENTIALS_FILE=/usr/local/e2e-aws-cluster-profile/.awscred
```
stevekuznetsov: Since this is dependent on the place where the secret is mounted in the job, could we set this as an env var in the job? Or mount the awscred to the "normal" place that the AWS libraries look?

hongkailiu: Done.

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+set -o pipefail
+
+if [ -z "${AWS_SHARED_CREDENTIALS_FILE}" ]; then echo "ENV VAR AWS_SHARED_CREDENTIALS_FILE is not set"; exit 1; fi
+
 ### expirationDate is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
-MY_DATE=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
-declare -a regions=("us-east-1" "us-east-2" "us-west-1")
+my_date=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
+
+if [[ -n "${IPI_DEPROVISION_DATE-}" ]]; then
```
stevekuznetsov: Hmm sorry -- we will want this to be relative to the time that is run, right? So the config option would be how old --

```
$ date --date='1 hour ago'
Tue Jul 23 16:54:10 PDT 2019
```

hongkailiu: Done.

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
 #!/bin/sh
```
stevekuznetsov: So I don't think we want to run POSIX shell. If we have `bash`, let's use it.

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
 
+set -o errexit
+set -o nounset
+set -o pipefail
```
stevekuznetsov: For instance, POSIX shell does not support this.

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+
+if [ -z "${AWS_SHARED_CREDENTIALS_FILE}" ]; then echo "ENV VAR AWS_SHARED_CREDENTIALS_FILE is not set"; exit 1; fi
```
stevekuznetsov: nit: prefer

```sh
if [[ -z "${AWS_SHARED_CREDENTIALS_FILE:-}" ]]; then
    echo "Required variable \$AWS_SHARED_CREDENTIALS_FILE unset"
    exit 1
fi
```

stevekuznetsov: Also, can't we just omit this? The AWS library will fail when no credentials are found, and this script could be used locally with the credentials in `$HOME`

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+
 ### expirationDate is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
```
stevekuznetsov: This comment is out of place

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
-MY_DATE=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
-declare -a regions=("us-east-1" "us-east-2" "us-west-1")
+my_date=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
+
+if [[ -n "${IPI_DEPROVISION_DATE_ARG-}" ]]; then
```
stevekuznetsov: We should validate that this is set, as without it, the script is not useful.

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+if [ -z "${AWS_SHARED_CREDENTIALS_FILE}" ]; then echo "ENV VAR AWS_SHARED_CREDENTIALS_FILE is not set"; exit 1; fi
+
 ### expirationDate is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
-MY_DATE=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
-declare -a regions=("us-east-1" "us-east-2" "us-west-1")
+my_date=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
```
stevekuznetsov: nit: we could name this `cluster_age_cutoff` or something similar

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+
+if [[ -n "${IPI_DEPROVISION_DATE_ARG-}" ]]; then
+  ### eg, export IPI_DEPROVISION_DATE_ARG="22 hours ago"
+  echo "take env. var.: IPI_DEPROVISION_DATE_ARG: ${IPI_DEPROVISION_DATE_ARG}"
```
stevekuznetsov: nit: maybe format like

```sh
echo "Searching for clusters with a TTL of ${IPI_DEPROVISION_DATE_ARG}"
```

or

```sh
echo "Searching for clusters older than ${my_date}
```

cmd/ipi-deprovision/ipi-deprovision.sh
stevekuznetsov: nit: that's a lot of names in the env var! we could just use something like `${CLUSTER_TTL}`

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+  my_date="$(TZ=":Africa/Abidjan" date --date="${IPI_DEPROVISION_DATE_ARG}" '+%Y-%m-%dT%H:%M+0000')"
+fi
+
+echo "my_date: ${my_date}"
+
+### James: CI only runs in us-east-1.
```
stevekuznetsov: we don't need to keep this comment around, I don't think

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+declare -a regions=("us-east-1")
+### for more than one region:
+# declare -a regions=("us-east-1" "us-east-2" "us-west-1")
+if [[ -n "${IPI_DEPROVISION_REGIONS-}" ]]; then
```
stevekuznetsov: Do we need this? We can add it later if we do

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+  ### eg, export IPI_DEPROVISION_REGIONS="us-east-1;us-west-1"
+  echo "take regions from env. var.: IPI_DEPROVISION_REGIONS: ${IPI_DEPROVISION_REGIONS}"
+  IFS=';' read -r -a regions <<< "$IPI_DEPROVISION_REGIONS"
```
stevekuznetsov: FYI `read -a` and `<<<` here-string are not defined in POSIX shell :)

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+fi
+
+echo "regions: ${regions}"
```
stevekuznetsov: likely do not need this debugger

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
 
 handle_cluster () {
   local cluster_name;
   cluster_name=$1
   echo "handling cluster: ${cluster_name} ..."
```
stevekuznetsov: might be nice to see the date it expired at, as well

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,44 @@
...
+  ### TODO: will test hiveutil later
+  #./bin/hiveutil aws-tag-deprovision "${cluster_name}=owned"
 }
 
 for r in "${regions[@]}"
 do
   echo "doing region ${r} ..."
-  aws ec2 describe-vpcs --output json --region "${r}" | jq --arg date "${MY_DATE}" -r -S '.Vpcs[] | select (.Tags[]? | (.Key == "expirationDate" and .Value < $date)) | .Tags[] | select (.Value == "owned") | .Key' | while read line; do handle_cluster ${line}; done
+  aws ec2 describe-vpcs --output json --region "${r}" | jq --arg date "${my_date}" -r -S '.Vpcs[] | select (.Tags[]? | (.Key == "expirationDate" and .Value < $date)) | .Tags[] | select (.Value == "owned") | .Key' | while read line; do handle_cluster ${line}; done
```
stevekuznetsov: nit: piping into `while read line` is a bit esoteric, we could:


```sh
for cluster in $( aws ec2 describe... ); do
    handle_cluster "${cluster}"
done
```

(and with that formatting we hardly need the `handle_cluster` function since inlining it is simple

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Done. CP/Paste error.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Good to know.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Deleted.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Improved a bit.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: rename: Done.
validation: not sure how to validate it. Could do it as a follow-up i think.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Done.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Deleted.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Deleted.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Good to know.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Deleted.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: /hold ... I want to give this a try.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Done.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: /hold cancel
There should be a way to do it with one search and pass 2 params into the function.
Let us iterate.

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,38 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+### cluster_age_cutoff is going to be used as argument in the aws command below
+### to filter out the expired clusters with the tags of the key expirationDate
+### which is always with timezone +00:00, eg, "2019-07-23T22:35+0000"
+cluster_age_cutoff=$(TZ=":Africa/Abidjan" date '+%Y-%m-%d')
+
+if [[ -n "${CLUSTER_TTL-}" ]]; then
```
stevekuznetsov: So if this is not set, we will remove all clusters with an expiration date older than now? This is a dangerous default -- we should enforce that this parameter is passed and not use a default

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -1,17 +1,38 @@
...
+
+### eg, export AWS_REGIONS="us-east-1;us-west-1"
+echo "take regions from env. var.: AWS_REGIONS: ${AWS_REGIONS}"
+IFS=';' read -r -a regions <<< "$AWS_REGIONS"
 
 handle_cluster () {
   local cluster_name;
   cluster_name=$1
-  echo "handling cluster: ${cluster_name} ..."
+  echo "handling cluster: ${cluster_name} with expirationDate: $(aws ec2 describe-vpcs --output json --region "us-east-1" --filters "Name=tag:${cluster_name},Values=owned" | jq -r '.Vpcs[].Tags[] | select (.Key=="expirationDate") | .Value')"
```
stevekuznetsov: nit: this will make runtime much slower as we now make O(N) aws calls (rather than one list) and be also more costly on API use -- we can save the output of `aws describe-vpcs --output json` into a file and post-process that or use `jq` in the first call to select the cluster name and expiration date, either way would work.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Done.

cmd/ipi-deprovision/ipi-deprovision.sh
hongkailiu: Done.
I like the 2nd way more but i can implement the 1st one for the moment.

PR #37, opened 2019-07-25
Fix vendors

droslean:


stevekuznetsov:
What did you end up doing for this to work?

droslean:
I tracked all the revisions that each dependency needs.

`k8s.io/client-go` and `k8s.io/apimachinery` to `kubernetes-1.14.0` tag
`github.com/openshift/client-go` needed to be at `master`. That series started working after I found out that `github.com/openshift/api` needed to point to `master` as well. Those had to be updated manually, because `go mod` was pointing to absolute wrong revisions.

PR #41, opened 2019-07-29
Remove git-ref arg, use extra_refs with workdir.

droslean:


pkg/rehearse/jobs.go
```diff
@@ -111,8 +111,22 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
...
+	// This check is to satisfy the tests and make sure that we won't hit any panic in case of any
+	// bugs while loading the prow config.
+	// It is go idiom that the string{"org","repo"} has length equals to '2'.
+	if len(orgRepo) > 1 {
+		rehearsal.ExtraRefs = []pjapi.Refs{
+			{
+				Org:     orgRepo[0],
+				Repo:    orgRepo[1],
+				BaseRef: branch,
+				WorkDir: true,
```
petr-muller: What happens if we attempt to rehearse something that already has extra refs? Do we just overwrite them?

droslean: I didn't think of that. Nice catch.. I will address this. Thanks.

pkg/rehearse/jobs.go
```diff
@@ -111,8 +111,20 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
 	rehearsal.Context = fmt.Sprintf("ci/rehearse/%s/%s/%s", repo, branch, shortName)
 	rehearsal.RerunCommand = defaultRehearsalRerunCommand
 
-	gitrefArg := fmt.Sprintf("--git-ref=%s@%s", repo, branch)
-	rehearsal.Spec.Containers[0].Args = append(source.Spec.Containers[0].Args, gitrefArg)
+	orgRepo := strings.Split(repo, "/")
+
+	// We will never hit different length because it will fail while we are loading the prowconfig.
```
stevekuznetsov: Alternatively, we could define:

```go
type orgRepo struct {
    org, repo string
}
```

Use this when we load and downstream from the loading code, we will have the type system help us not make any mistakes or need to make any comments like this

pkg/rehearse/jobs_test.go
```diff
@@ -361,8 +386,8 @@ func makeTestData() (int, string, string, *pjapi.Refs) {
 	testPrNumber := 123
 	testNamespace := "test-namespace"
 	testRefs := &pjapi.Refs{
-		Org:     "testRepo",
```
stevekuznetsov: Yikes -- what happened here?

pkg/rehearse/jobs_test.go
```diff
@@ -529,72 +558,54 @@ func TestExecuteJobsPositive(t *testing.T) {
...
-				[]string{"arg1", fmt.Sprintf("--git-ref=%s@not-master", targetRepo)},
-			).Spec,
+				testRefs, targetOrg, targetRepo, "not-master").Spec,
 		}},
 		{
 			description: "two jobs in a separate repos",
 			jobs: map[string][]prowconfig.Presubmit{
-				targetRepo:        {*makeTestingPresubmit("job1", "ci/prow/job1", []string{"arg1"}, "master")},
-				anotherTargetRepo: {*makeTestingPresubmit("job2", "ci/prow/job2", []string{"arg1"}, "master")},
+				targetRepo:        {*makeTestingPresubmit("job1", "ci/prow/job1", targetOrg, targetRepo, "master")},
```
stevekuznetsov: Shouldnt keys for presubmits be `org/repo` format?

pkg/rehearse/jobs_test.go
droslean: That was an old typo

pkg/rehearse/jobs_test.go
droslean: those are being used only to fill the `ExtraRefs`

pkg/rehearse/jobs_test.go
stevekuznetsov: Did you change both ways? Why wasn't this test broken before or after?

pkg/rehearse/jobs_test.go
stevekuznetsov: Right but you're using it as the key, isn't that weird? Why?

pkg/rehearse/jobs_test.go
droslean: I guess we didn't even use that.... I will check why it wasn't failing...

pkg/rehearse/jobs_test.go
droslean: ah.. sry I misunderstood you above. Yea it seems that we don't care for the key much, but you are right we need to be consistent here. I will change this.

droslean: addressed

pkg/rehearse/jobs_test.go
droslean: Well, we used those refs to pass them in the `Executor`, but we didn't evaluate the actual refs but only the execution. We could always extend the test but I think that they are ok as they are now.

pkg/rehearse/jobs.go
droslean: we get the `org/repo` from the key of the `map[string][]Presubmit` type... I wish we had a better way to get the values but for now this is the best that we can do.. either way we have to `.Split` it in some point of the code.

stevekuznetsov: Right, I was just saying `Split()` it quickly and send the `struct` afterwords. No biggie.

pkg/rehearse/jobs.go
```diff
@@ -111,8 +111,14 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
...
 	rehearsal.RerunCommand = defaultRehearsalRerunCommand
 
-	gitrefArg := fmt.Sprintf("--git-ref=%s@%s", repo, branch)
-	rehearsal.Spec.Containers[0].Args = append(source.Spec.Containers[0].Args, gitrefArg)
+	orgRepo := strings.Split(repo, "/")
+	rehearsal.ExtraRefs = append(rehearsal.ExtraRefs, pjapi.Refs{
+		Org:     orgRepo[0],
+		Repo:    orgRepo[1],
+		BaseRef: branch,
+		WorkDir: true,
```
stevekuznetsov: We might want to make sure that the other ones don't set this so we have only one workdir setting (does prow validate this?)

droslean: `Prow` will use the first `workDir` that will find in the slice. `ci-operator` too... We don't validate multiple `workDir`s .. I guess this is good, if our stakeholders for some reason would want to override it

stevekuznetsov: Seems reasonable

PR #48, opened 2019-08-01
ci-operator: inject path alias to refs

droslean:
Recent prow changes that make use of the `workdir: true` in the refs, change the idea of how we treat the refs in `ci-operator`. 
Now we are just assigning the `PathAlias` to the `refs` but that breaks a different logic that we are now able to use in the `pj-rehearse` tool.

Rehearsals have a ref that point to the `release` repo PR and `extra_refs` with `workdir: true` that will be the main repo to be cloned and run the tests against that. 
The problem was that ci-operator is assigning the `PathAlias` to the `refs` which is, in this case, the `release` repo instead. 

To solve this issue, this PR is changing the logic with injecting the `PathAlias` to the ref that has a `workdir: true` instead. If there isn't any, then assign it to the `refs` as usual.

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -116,22 +116,12 @@ func (s *sourceStep) Run(ctx context.Context, dry bool) error {
 		s.resources,
 	)
 
-	var refs []interface{}
-	// periodic jobs may have no refs to clone
-	if s.jobSpec.Refs != nil {
-		ref := s.jobSpec.Refs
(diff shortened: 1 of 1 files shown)


smarterclayton:
Looks reasonable to me, have you been able to test this before we merge?

droslean:
I tested it, both as a normal job and as a rehearsal. Both passed.

PR #51, opened 2019-08-05
add rehearse label to generated presubmits

droslean:
We are about to add an opt-in mechanism for the rehearsals that will be allowed to be executed only if they hold the `pj-rehearse.openshift.io/can-be-rehearsed: "true"` label. 
With this PR, we include this label to all generated ci-operator presubmits.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -398,8 +398,12 @@ func generateConfigMapVolume(name string, templates []string) kubeapi.Volume {
 	return ret
 }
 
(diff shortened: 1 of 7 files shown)


droslean:
/hold

PR for the rehearse tool is on the way.

petr-muller:
Hmm. This needs a companion PR to openshift/release, which won't be allowed to merge because of the prowgen check. I think `breaking-changes` test was somehow made required during the transition from ci-operator-prowgen to ci-tools repo? (@bbguimaraes ?) I think it was not required in the past, allowing changes like this to merge but reminding us to prepare the opeshift/release PR in advance...

PR #54, opened 2019-08-05
Populate env. var. in test pods

hongkailiu:
https://jira.coreos.com/browse/DPTP-473

The list of env. var. names are defined:
https://github.com/openshift/test-infra/blob/master/prow/jobs.md#job-environment-variables

### DIFF
--- a/pkg/api/job_spec.go
+++ b/pkg/api/job_spec.go
@@ -8,9 +8,7 @@ import (
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 )
 
-// JobSpec is a superset of the upstream spec, but
-// we do not import it as importing test-infra is a
-// massive hassle.
+// JobSpec is a superset of the upstream spec.
 type JobSpec struct {
 	downwardapi.JobSpec `json:",inline"`
 
--- a/pkg/steps/pod.go
+++ b/pkg/steps/pod.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/test-infra/prow/pod-utils/decorate"
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
@@ -192,19 +193,11 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
 	}
 }
 
-func getEnv(s *podStep) []coreapi.EnvVar {
-	var env []coreapi.EnvVar
+func (s *podStep) generatePodForStep(image string, containerResources coreapi.ResourceRequirements) (*coreapi.Pod, error) {
 	envMap, err := downwardapi.EnvForSpec(s.jobSpec.JobSpec)
 	if err != nil {
-		return env
+		return nil, err
 	}
-	for k, v := range envMap {
-		env = append(env, coreapi.EnvVar{Name: k, Value: v})
-	}
-	return env
-}
-
-func (s *podStep) generatePodForStep(image string, containerResources coreapi.ResourceRequirements) (*coreapi.Pod, error) {
 	pod := &coreapi.Pod{
 		ObjectMeta: meta.ObjectMeta{
 			Name: s.config.As,
@@ -226,7 +219,7 @@ func (s *podStep) generatePodForStep(image string, containerResources coreapi.Re
 			Containers: []coreapi.Container{
 				{
 					Image:                    image,
-					Env:                      getEnv(s),
+					Env:                      decorate.KubeEnv(envMap),
 					Name:                     s.name,
 					Command:                  []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + s.config.Commands},
 					Resources:                containerResources,
--- a/pkg/steps/pod.go
+++ b/pkg/steps/pod.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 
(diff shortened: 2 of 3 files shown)


pkg/steps/pod.go
```diff
@@ -191,6 +192,47 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
 	}
 }
 
+func (s *podStep) getEnvMap() map[string]string {
```
stevekuznetsov: We do vendor `test-infra` in `ci-tools`, can we import and use the function from upstream?

hongkailiu: Added a `spec` type conversion function.
As a followup, I think we could `import` the type from upstream and reuse it in ours.
Maybe I did not get the point of redefining the spec in our code.

droslean:
Why don't we just inject them into the `src` image?

pkg/steps/pod.go
petr-muller: I think originally ci-operator did not vendor `test-infra` b/c it's a hell of a vendor. Not sure if that's the reason, but could be.

pkg/api/job_spec.go
```diff
@@ -111,3 +112,11 @@ func ResolveSpecFromEnv() (*JobSpec, error) {
 
 	return spec, nil
 }
+
+func (s *JobSpec) GetJobSpecUpstream() (*downwardapi.JobSpec, error) {
+	spec := &downwardapi.JobSpec{}
+	if err := json.Unmarshal([]byte(s.RawSpec()), spec); err != nil {
+		return nil, err
+	}
+	return spec, nil
```
petr-muller: I think the check is redundant, just return `spec, err`?

hongkailiu: Removed.

hongkailiu:
/hold
blocked by https://github.com/openshift/ci-tools/pull/56

pkg/api/job_spec.go
```diff
@@ -8,6 +8,7 @@ import (
 	"strings"
 
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 )
 
 // JobSpec is a superset of the upstream spec, but
```
stevekuznetsov: Lets update this comment

hongkailiu: Yes. Will do during the rebasing.

pkg/steps/pod.go
```diff
@@ -191,6 +192,18 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
 	}
 }
 
+func getEnv(s *podStep) []coreapi.EnvVar {
```
stevekuznetsov: can you make a PR upstream to expose the `kubeEnv` function? We will want to re-use it, and need the sorting behavior for stable test output

hongkailiu: Sure ... let me do the upstream PR first ...
Testing is still unstable ... :(

hongkailiu:
/hold
blocked by https://github.com/openshift/ci-tools/pull/59

hongkailiu:
/retest unit
for stability: 1

hongkailiu:
/retest unit
for stability: 2

hongkailiu:
Did several times ... for `/test unit`

https://prow.svc.ci.openshift.org/?repo=openshift%2Fci-tools&type=presubmit&pull=54

Yeah, stability seems back. `sorting` does help as @stevekuznetsov pointed out.

Let us wait for `dep` upgrade and I will rebase accordingly.

pkg/steps/pod_test.go
```diff
@@ -14,6 +15,13 @@ import (
 	"github.com/openshift/ci-tools/pkg/api"
 )
 
+const (
+	buildID      = "test-build-id"
+	jobName      = "very-cool-prow-job"
+	pjID         = "prow-job-id"
+	jobSpecValue = `{"type":"periodic","job":"periodic-ci-azure-e2e-applysecurityupdates","buildid":"21","prowjobid":"ec28bec2-b7a4-11e9-af8e-0a58ac108dbc","extra_refs":[{"org":"openshift","repo":"openshift-azure","base_ref":"master"}]}`
```
stevekuznetsov: Might as well use a presubmit or batch config as well to test all the env vars we don't set with periodics

hongkailiu: Will do today.
I was keeping things as simple as possible yesterday, because, you know, the sorting thing was eating me alive.

hongkailiu: Done.

pkg/steps/pod.go
```diff
@@ -191,6 +193,15 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
```
stevekuznetsov: We already have the spec -- call the `decorate.KubeEnv(downwardapi.EnvForSpec())` on `podStep.jobSpec` and we don't need this method

pkg/steps/pod.go
```diff
@@ -191,6 +193,15 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
+	envMap, err := downwardapi.EnvForSpec(s.jobSpec.JobSpec)
+	if err != nil {
```
stevekuznetsov: This should get bubbled up

pkg/steps/pod_test.go
```diff
@@ -71,7 +93,13 @@ func preparePodStep(t *testing.T, namespace string) (*podStep, stepExpectation,
 	return ps.(*podStep), specification, client
 }
 
-func makeExpectedPod(step *podStep, phaseAfterRun v1.PodPhase) *v1.Pod {
+func makeExpectedPod(step *podStep, phaseAfterRun v1.PodPhase, t *testing.T) *v1.Pod {
```
stevekuznetsov: So in general there is a good reason not to have helpers like this. What are we testing? I know you're new to this code but in general we should prefer hard-coded literals for test structures. When we confound the testing with large helpers, we lose the simplicity of input/output checking. Furthermore, you are using `json.Marshal` as well as `decorate.KubeEnv` in the creation of the environment -- the test should not care. These are implementation details today of how these values are created. Your test should validate that the output `PodSpec` is correct.

pkg/steps/pod_test.go
```diff
@@ -351,3 +394,36 @@ func expectedPodStepTemplate() *podStep {
 		},
 	}
 }
+
+func TestGetEnv(t *testing.T) {
```
stevekuznetsov: We don't need this func so we can drop the test

hongkailiu:
/hold
checking ...

pkg/steps/pod.go
hongkailiu: Done.

pkg/steps/pod.go
hongkailiu: Done.

pkg/steps/pod_test.go
hongkailiu: Done.

PR #56, opened 2019-08-06
Embed jobSpec from downwardapi

hongkailiu:


pkg/api/job_spec.go
```diff
@@ -2,25 +2,17 @@ package api
...
 // massive hassle.
 type JobSpec struct {
-	Type      ProwJobType `json:"type,omitempty"`
-	Job       string      `json:"job,omitempty"`
-	BuildId   string      `json:"buildid,omitempty"`
-	ProwJobID string      `json:"prowjobid,omitempty"`
-
-	Refs      *Refs  `json:"refs,omitempty"`
-	ExtraRefs []Refs `json:"extra_refs,omitempty"`
+	downwardapi.JobSpec
```
stevekuznetsov: I think this needs to be JSON inline? Does this work locally for you?

hongkailiu: Added just in case.

Probably inline is not necessary for json.
https://play.golang.org/p/az46KDeLW1I

But i saw many inline at embedding, eg,

https://github.com/kubernetes/test-infra/blob/5e888585cfcca5b1593d3e8d143e442b348530b6/prow/config/config.go#L243

stevekuznetsov:
@hongkailiu update the title not to be WIP if this can merge

PR #58, opened 2019-08-06
make rehearse tool to check for requred label to rehearse a job

droslean:


pkg/rehearse/jobs.go
```diff
@@ -42,7 +42,8 @@ const (
 	logCiopConfigFile            = "ciop-config-file"
 	logCiopConfigRepo            = "ciop-config-repo"
 
-	clusterTypeEnvName = "CLUSTER_TYPE"
+	clusterTypeEnvName  = "CLUSTER_TYPE"
+	canBeRehearsedLabel = "pj-rehearse.openshift.io/can-be-rehearsed"
```
stevekuznetsov: Can this be shared and imported?

pkg/rehearse/jobs.go
```diff
@@ -168,22 +180,28 @@ func filterPeriodics(changedPeriodics []prowconfig.Periodic, allowVolumes bool,
 	return periodics
 }
 
+func filterLabels(labels map[string]string) error {
```
stevekuznetsov: `filter` makes me think of taking a collection and choosing a subset of it using some rule ... this is not really doing that.

pkg/rehearse/jobs.go
droslean: yea, it turns out that we need that for the tests as well.

pkg/rehearse/jobs.go
droslean: you are right. `validateLabels` is the best name here.

pkg/rehearse/jobs.go
stevekuznetsov: So actually in the test file I would not use the import -- we want someone who changes the value of a constant like this to have tests break so they know they're changing functionality. But sharing this between rehearse and gen is good

pkg/rehearse/jobs.go
```diff
@@ -159,6 +166,11 @@ func filterPeriodics(changedPeriodics []prowconfig.Periodic, allowVolumes bool,
 	for _, periodic := range changedPeriodics {
 		jobLogger := logger.WithField("job", periodic.Name)
 
+		if err := validateLabels(periodic.Labels); err != nil {
```
stevekuznetsov: It's OK if someone does not opt into rehearsal. These are not really errors -- we should just 


```go
	if value, ok := labels[CanBeRehearsedLabel]; !ok || value != "true" {
		continue
	}
```

test/pj-rehearse-integration/candidate/ci-operator/jobs/super/duper/super-duper-master-presubmits.yaml
```diff
@@ -157,50 +169,6 @@ presubmits:
             cpu: 10m
       serviceAccountName: ci-operator
     trigger: ((?m)^/test( all| integration| changed),?(\s+|$))
-  - agent: kubernetes
```
stevekuznetsov: Why did this change?

droslean: it changed order. No worries, I added only the corresponding changes. I will update the integration tests in another PR.

pkg/rehearse/jobs.go
droslean: This is what am doing here. Check `validateLabels()`

pkg/rehearse/jobs.go
```diff
@@ -168,22 +180,28 @@ func filterPeriodics(changedPeriodics []prowconfig.Periodic, allowVolumes bool,
...
-		if arg == "--promote" {
-			return fmt.Errorf("cannot rehearse jobs that call ci-operator with '--promote' arg")
+	if len(container.Command) > 0 && container.Command[0] == "ci-operator" {
+		for _, arg := range container.Args {
+			if strings.HasPrefix(arg, "--git-ref") || strings.HasPrefix(arg, "-git-ref") {
+				return fmt.Errorf("cannot rehearse jobs that call ci-operator with '--git-ref' arg")
+			}
+			if arg == "--promote" {
+				return fmt.Errorf("cannot rehearse jobs that call ci-operator with '--promote' arg")
+			}
```
petr-muller: Personally I would drop these checks too and let the label have full control. Are we afraid of malice or mistakes? If malice, then it's still going to be possible to craft a PR that will run ci-op with promote. If mistakes, then I think it would be a unique mistake - generated jobs won't have the label so copy-paste should not be an issue, and nobody should ever write a `ci-operator --promote` job by hand...

droslean: addresed

droslean: I removed also the `git-ref` check, after the new changes in `ci-operator`

pkg/rehearse/jobs.go
stevekuznetsov: I mean continue, not raise an error

pkg/rehearse/jobs_test.go
```diff
@@ -132,8 +132,9 @@ func TestReplaceClusterProfiles(t *testing.T) {
 func makeTestingPresubmitForEnv(env []v1.EnvVar) *prowconfig.Presubmit {
 	return &prowconfig.Presubmit{
 		JobBase: prowconfig.JobBase{
-			Agent: "kubernetes",
-			Name:  "test-job-name",
+			Agent:  "kubernetes",
+			Name:   "test-job-name",
+			Labels: map[string]string{CanBeRehearsedLabel: "true"},
```
stevekuznetsov: Especially if more parts of our system start to rely on a specific label value, we want someone who changes the constant to break the tests. Using the constant _in_ the test is counterproductive there

droslean: that makes perfect sense. I will address it.

droslean: addressed

PR #67, opened 2019-08-08
Don't change refs at runtime unless running locally 

stevekuznetsov:
Use `canonical_go_repository` to generate `path_alias`

These settings are identical, so we should use the CI Operator
configuration to generate an appropriate set of jobs.

---

Don't change refs at runtime unless running locally

When running with a Prow-provided set of Refs, we should honor those and
not attempt to change them in CI Operator. If a user is passing
`--git-ref` manually, though, we have to construct a set of refs
dynamically at runtime.

---

depends on #66 
/assign @petr-muller @droslean

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -11,7 +11,7 @@ import (
 
 	"github.com/openshift/ci-tools/pkg/promotion"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


stevekuznetsov:
(this is known to be a breaking change)

petr-muller:
/lgtm

The code changes look fine, but I won't pretend I have a full idea of how this pathAlias stuff works on all layers...

PR #68, opened 2019-08-08
determine ref with workdir when cloning source from git

droslean:


pkg/steps/git_source.go
```diff
@@ -27,23 +27,19 @@ func (s *gitSourceStep) Inputs(ctx context.Context, dry bool) (api.InputDefiniti
...
-	return handleBuild(s.buildClient, buildFromSource(s.jobSpec, "", api.PipelineImageStreamTagReferenceRoot, buildapi.BuildSource{
-		Type:       buildapi.BuildSourceGit,
-		ContextDir: s.config.ContextDir,
-		Git: &buildapi.GitBuildSource{
-			URI: fmt.Sprintf("https://github.com/%s/%s.git", refs.Org, refs.Repo),
-			Ref: refs.BaseRef,
-		},
-	}, s.config.DockerfilePath, s.resources), dry, s.artifactDir)
+
+	log.Printf("Nothing to build source image from, no refs")
```
stevekuznetsov: this is an error and should fail the job

pkg/steps/git_source_test.go
```diff
@@ -0,0 +1,107 @@
...
+	}{
+		{
+			testName:  "no workdir, nil Refs/ExtraRefs, expect nil",
+			refs:      nil,
+			extraRefs: nil,
+			expected:  nil,
+		},
+		{
+			testName:  "no workdir, expect Refs",
+			refs:      makeTestRefs(),
```
stevekuznetsov: We should prefer to have raw literal inputs and outputs to test cases unless they severely impede readability. Factoring out commonality and DRY in production code is useful as it ensures that there is one authoritative means by which some concept is represented. That does not apply to these tests -- there is nothing shared between the `refs` you choose to have in this test case versus the one after it. For all we care, they should be totally different. 

I would prefer to have test cases that are entirely self-contained within one `testCase`. As soon as I am needing to read helpers, it is harder to grok what a test case is doing and how and _why_ it shares logic with other tests.

droslean: What you said make sense. For my defense, I didn't focus on the actual values of the `refs`, but I wanted to make sure which of them will be chosen. 
Either way, I will remove the helpers and have raw inputs and outputs since those cases are closer on what we should expect from the production.

pkg/steps/git_source.go
droslean: addressed

pkg/steps/git_source_test.go
droslean: addressed

PR #71, opened 2019-08-08
Fix timeout command in ipi-deprovision script

hongkailiu:


cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -26,7 +26,7 @@ handle_cluster () {
   local region
   region=$3
   echo "handling cluster: ${cluster_name} in region ${region} with expirationDate: ${expirationDateValue}"
-  timeout -v 10m ./bin/hiveutil aws-tag-deprovision --loglevel debug "${cluster_name}=owned" --region "${region}"
+  timeout 10m ./bin/hiveutil aws-tag-deprovision --loglevel debug "${cluster_name}=owned" --region "${region}" || (echo "timeout and died with $?" && false)
```
wking: `timeout and died with $?` isn't quite right, since we could have `hiveutil` exit before the timeout.  For example:

```console
$ timeout 4 false || echo "exited $?"
exited 1
```

Maybe `${cluster_name} deletion exited $?`?

hongkailiu: right.

stevekuznetsov:
Technically if your `echo` fails, you won't get an `errexit`. Why don't we just not do the `|| echo` stuff?

wking:
> Technically if your `echo` fails, you won't get an `errexit`. Why don't we just not do the `||` echo stuff?

Fine with me too.

/lgtm

PR #75, opened 2019-08-09
Small bug fixes

bbguimaraes:


cmd/ci-operator/main.go
```diff
@@ -714,6 +714,9 @@ func (o *options) initializeNamespace() error {
 			if err != nil {
 				return fmt.Errorf("could not retrieve secret %s: %v", secret.Name, err)
 			}
+			if l := len(secret.Data); l != 0 && existing.Data == nil {
```
droslean: nit: you can simplify this as
```go
if len(secret.Data) > 0 && existing.Data == nil 
```

pkg/defaults/defaults.go
```diff
@@ -154,17 +154,19 @@ func FromConfig(
...
 			buildSteps = append(buildSteps, initialReleaseStep)
 
-		} else if rawStep.TestStepConfiguration != nil && rawStep.TestStepConfiguration.OpenshiftInstallerClusterTestConfiguration != nil && rawStep.TestStepConfiguration.OpenshiftInstallerClusterTestConfiguration.Upgrade {
-			var err error
-			step, err = clusterinstall.E2ETestStep(*rawStep.TestStepConfiguration.OpenshiftInstallerClusterTestConfiguration, *rawStep.TestStepConfiguration, params, podClient, templateClient, secretGetter, artifactDir, jobSpec)
-			if err != nil {
-				return nil, nil, fmt.Errorf("unable to create end to end test step: %v", err)
+		} else if testStep := rawStep.TestStepConfiguration; testStep != nil {
+			if testStep.OpenshiftInstallerClusterTestConfiguration != nil {
+				if testStep.OpenshiftInstallerClusterTestConfiguration.Upgrade {
```
droslean: nit: those are a lot of `if`s. Can you put them in a func to be easier to read. Something like 
```go
func isStepWithClusterUpgrade(rawStep api.StepConfiguration) bool {
	testStep: = rawStep.TestStepConfiguration
	if testStep != nil && testStep.OpenshiftInstallerClusterTestConfiguration != nil {
		if testStep.OpenshiftInstallerClusterTestConfiguration.Upgrade {
			return true
		}
	}
	return false
}
```

cmd/ci-operator/main.go
droslean: meaning that it's ok to use `len(secret.Data)` below.

PR #76, opened 2019-08-09
 applyconfig: handle templates

petr-muller:
This is one of the things needed to move all Prow config under core-services to be covered under the dry run job.

Basically run `oc process path | oc apply -f -` for templates. Detecting
a template means we need to read and parse it unfortunately, but we
vendor the necessary pieces anyway so why not.

Some abstractions were necessary, so I wrote a simple mock executor for testing it all.

@hongkailiu we could use the same principle to handle the special monitoring snowflake CRs (Prometheuses?) that need to be deleted and created.

/cc @stevekuznetsov @bbguimaraes @droslean @hongkailiu


cmd/applyconfig/applyconfig_test.go
```diff
@@ -64,47 +69,181 @@ func TestIsStandardConfig(t *testing.T) {
...
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			args := makeOcArgs(tc.path, tc.user, tc.dry)
-			if !reflect.DeepEqual(args, tc.expected) {
-				t.Errorf("Expected '%v', got '%v'", tc.expected, args)
+			cmd := makeOcApply(tc.path, tc.user, tc.dry)
+			command := strings.Join(cmd.Args, " ")
```
stevekuznetsov: This fails to test an errant `Command` with one arg like `--foo bar` (where we would want two args, `--foo` `bar`)

cmd/applyconfig/applyconfig.go
```diff
@@ -84,24 +89,86 @@ func makeOcArgs(path, user string, dry bool) []string {
...
+
+func (c *configApplier) asGenericManifest() error {
+	cmd := makeOcApply(c.path, c.user, c.dry)
+	_, err := c.runAndCheck(cmd, "apply")
+	return err
+}
+
+func (c configApplier) asTemplate() error {
+	ocProcess := []string{"process", "-f", c.path}
+	if c.user != "" {
```
stevekuznetsov: Is it possible to factor this such that we only have one place that appends `--dry-run` and `--as` ?

petr-muller: The difference here is that `oc process` does not have `--dry-run`, so the only common part is `--as`, but I can extract that, sure.

cmd/applyconfig/applyconfig_test.go
petr-muller: Addressed everywhere

cmd/applyconfig/applyconfig.go
petr-muller: Done

petr-muller:
@stevekuznetsov PTAL at the code now, I'll need to figure out why `ci/prow/validate-vendor` fails on me :/

petr-muller:
https://github.com/openshift/ci-tools/pull/76/commits/7c4a78e5ebfd68b42571bab33a76463f4848e318
/shrug

petr-muller:
@stevekuznetsov addressed

code changes: ~10m
test: ~15m
figuring out why the testing template is not correctly parsed: ~30m (template from [documentation](https://docs.openshift.com/container-platform/4.1/openshift_images/using-templates.html#templates-writing_using-templates) has `apiVersion: v1`, but needs `template.openshift.io/v1`)
https://github.com/openshift/ci-tools/commit/7c4a78e5ebfd68b42571bab33a76463f4848e318: ~15m

PR #77, opened 2019-08-10
add org/repo label in template pods

droslean:
Since `ci-operator` can get this kind of information from `jobSpec`, this PR adds a label to template pods with the org and repo details.

Example:
 `"ci.openshift.io/refs.org": "openshift"`
`"ci.openshift.io/refs.repo": "origin"`

Why?
We can use this label in Prometheus queries. We can track the template jobs by org/repo and we will be able to gather metrics, etc...

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -31,6 +31,11 @@ import (
 	"github.com/openshift/ci-tools/pkg/junit"
 )
 
+const (
+	RefsOrgLabel  = "ci.openshift.io/refs.org"
+	RefsRepoLabel = "ci.openshift.io/refs.repo"
+)
+
 type templateExecutionStep struct {
 	template       *templateapi.Template
 	params         api.Parameters
@@ -82,6 +87,14 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
 				s.template.Parameters[i].Value = strings.Replace(format, api.ComponentFormatReplacement, component, -1)
 			}
 		}
+	}
+
+	if refs := s.jobSpec.JobSpec.Refs; refs != nil {
+		if s.template.ObjectLabels == nil {
+			s.template.ObjectLabels = make(map[string]string)
+		}
+		s.template.ObjectLabels[RefsOrgLabel] = refs.Org
+		s.template.ObjectLabels[RefsRepoLabel] = refs.Repo
 	}
 
 	if len(s.artifactDir) > 0 {


pkg/steps/template.go
```diff
@@ -46,6 +55,16 @@ func (s *templateExecutionStep) Inputs(ctx context.Context, dry bool) (api.Input
 	return nil, nil
 }
 
+func getOrgRepoFromVars(jobspec downwardapi.JobSpec) string {
```
stevekuznetsov: You have the job spec in the template step already, just use the values from that?

pkg/steps/template.go
```diff
@@ -24,13 +24,22 @@ import (
...
 	templateclientset "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/junit"
 )
 
+const (
+	RepoOwnerVar       = "REPO_OWNER"
+	RepoNameVar        = "REPO_NAME"
+	RepoOwnerNameLabel = "template.openshift.io/template-repo-owner-name"
```
stevekuznetsov: Could we reuse the `prow.k8s.io/refs.{org,repo} ` label names? Or `ci.openshift.io`

pkg/steps/template.go
droslean: I am using the jobspec from the template step already, [here](https://github.com/openshift/ci-tools/pull/77/files#diff-37ed7e4b3e5f7bb5d23118bff1f75bd9R106)

pkg/steps/template.go
droslean: `ci.openshift.io/refs-repo-owner-name`. I can't use `.{org,repo}` in the label name because we need to get the org/repo from the value. This will helps as to use the actual values as variables in grafana.

droslean: Do you mean having 2 labels? like `ci.openshift.io/refs.org` and `ci.openshift.io/refs.repo`?

pkg/steps/template.go
```diff
@@ -46,6 +55,16 @@ func (s *templateExecutionStep) Inputs(ctx context.Context, dry bool) (api.Input
 	return nil, nil
 }
 
+func getOrgRepoFromVars(jobspec downwardapi.JobSpec) string {
+	vars := make(map[string]string)
+	envMap, _ := downwardapi.EnvForSpec(jobspec)
```
stevekuznetsov: I don't understand. Just do spec.refs.org. No need to create variables from it...

pkg/steps/template.go
stevekuznetsov: Yes, two labels

pkg/steps/template.go
droslean: let me address that :/

pkg/steps/template.go
droslean: addressed

pkg/steps/template.go
droslean: addressed

PR #81, opened 2019-08-13
add branch label in template pods

droslean:
/cc @stevekuznetsov

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -32,8 +32,9 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


droslean:
to be used in [release/pull/4728](https://github.com/openshift/release/pull/4728)

stevekuznetsov:
/hold

Why doesn't this break any tests? :)

droslean:
> Why doesn't this break any tests? :)

Because we don't have any tests!!!!!

PR #82, opened 2019-08-13
Drop CI template bootstrap_ip code.

dprince:
As the openshift-installer now has 'gather bootstrap' support
we should drop this example from the CI templates.

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -441,43 +441,6 @@ objects:
           echo "Gathering artifacts ..."
           mkdir -p /tmp/artifacts/pods /tmp/artifacts/nodes /tmp/artifacts/metrics /tmp/artifacts/bootstrap /tmp/artifacts/network
 
-
-          if [ -f /tmp/artifacts/installer/terraform.tfstate ]
-          then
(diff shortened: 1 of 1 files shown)


abhinavdahiya:
The template is used for both 4.1 and 4.2 clusters. And the 'gather bootstrap' is a 4.2 feature. Therefore we cannot drop this.

dprince:
> The template is used for both 4.1 and 4.2 clusters. And the 'gather bootstrap' is a 4.2 feature. Therefore we cannot drop this.

Would it be reasonable to wrap this code in a block "if openshift-install version | grep v4.1; then"

Goal would be to make this also work for OpenStack per your suggestions here: https://github.com/openshift/release/pull/4716

dprince:
> > The template is used for both 4.1 and 4.2 clusters. And the 'gather bootstrap' is a 4.2 feature. Therefore we cannot drop this.
> 
> Would it be reasonable to wrap this code in a block "if openshift-install version | grep v4.1; then"
> 
> Goal would be to make this also work for OpenStack per your suggestions here: [openshift/release#4716](https://github.com/openshift/release/pull/4716)

Pushed this instead: https://github.com/openshift/ci-tools/pull/83

PR #83, opened 2019-08-13
Rely on 'gather bootstrap' for CI templates > 4.1

dprince:
Update the clusterinstall CI template so that it skips the
inline bootstrap_ip code if the installer version matches 4.1.
All other CI versions supported (4.2, and greater) would have
'gather bootstrap' to replace this functionality.

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -442,7 +442,8 @@ objects:
           mkdir -p /tmp/artifacts/pods /tmp/artifacts/nodes /tmp/artifacts/metrics /tmp/artifacts/bootstrap /tmp/artifacts/network
 
 
-          if [ -f /tmp/artifacts/installer/terraform.tfstate ]
+          # NOTE: OpenShift 4.2 and greater have native 'gather bootstrap' support
+          if [ -f /tmp/artifacts/installer/terraform.tfstate ] && openshift-install version | grep v4.1 &>/dev/null
           then
               # we don't have jq, so the python equivalent of
               # jq '.modules[].resources."aws_instance.bootstrap".primary.attributes."public_ip" | select(.)'


pkg/steps/clusterinstall/template.go
```diff
@@ -442,7 +442,8 @@ objects:
           mkdir -p /tmp/artifacts/pods /tmp/artifacts/nodes /tmp/artifacts/metrics /tmp/artifacts/bootstrap /tmp/artifacts/network
 
 
-          if [ -f /tmp/artifacts/installer/terraform.tfstate ]
+          # NOTE: OpenShift 4.2 and greater have native 'gather bootstrap' support
+          if [ -f /tmp/artifacts/installer/terraform.tfstate ] && openshift-install version | grep v4.1 &>/dev/null
```
abhinavdahiya: The CI runs have 0.0.1- style versions. i'm not sure if CI jobs running on release-4.1 branches will have the 4.1XX version set??

dprince: I got it from one of these:

https://storage.googleapis.com/origin-ci-test/logs/release-openshift-ocp-installer-e2e-aws-4.1/347/artifacts/e2e-aws/installer/.openshift_install.log

Which has: "OpenShift Installer v4.1.12-201908130938-dirty"

Are there other branches to account for?

stevekuznetsov:
@abhinavdahiya @dprince is this still active?

dprince:
> @abhinavdahiya @dprince is this still active?

By my assessment this patch was still okay. Last I looked the 4.1 tests had version strings set like "OpenShift Installer v4.1.12-201908130938-dirty". I dunno, maybe I'm missing something on another set of tests where the version string isn't set correctly.

I would say this is nice to have but not required. Just trying to clean up this template a bit as the code here confused me a bit.

abhinavdahiya:
> /hold

I don't think we should be doing this version magic... the current setup works already.

dprince:
> > /hold
> 
> I don't think we should be doing this version magic... the current setup works already.

It works, but duplicates the new bootstrap gather feature in the installer codebase now. Going forward this code should probably get removed. The version code I was adding here was just meant to clarify when that removal could happen. Not a big deal either way though. I guess when 4.1 is dropped someone can come back and just remove this manually.

PR #84, opened 2019-08-13
Inject resources in template pods, refactor label injection

droslean:
This PR adds the ability to define resources for the tests using templates.

- [X] Inject resources in template pods
- [X] Refactor label injection logic
- [X] Unittests for both


pkg/steps/template.go
```diff
@@ -59,6 +65,30 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
...
+		containerResources, err := resourcesFor(s.resources.RequirementsForStep(s.template.Name))
+		if err != nil {
+			return fmt.Errorf("unable to calculate resources for %s: %s", s.template.Name, err)
+		}
+
+		for _, object := range s.template.Objects {
+			var pod *coreapi.Pod
+			yaml.Unmarshal(object.Raw, &pod)
+
+			if pod != nil {
```
smarterclayton: You can only do this for things that are pods. This doesn't check that things are pods.

stevekuznetsov: What's the type to deserialize into that has GVR ?

droslean: My intention here was to check if the `Unmarshal` didn't work because the object can be different than `pod` and the pod var will continue to be `nil`. In this case, we don't want to continue.

droslean: Also, the Unmarshal won't raise an error if the object is different than `pod`.

stevekuznetsov: Easy to do something like this, though: https://github.com/openshift/release/blob/297849cc49ed9c4d1ba9ae94c5104f36653f1ec8/tools/release-ci/pkg/buildrunner/runbuild.go#L65-L72

pkg/steps/template_test.go
```diff
@@ -0,0 +1,219 @@
...
+			templateFakeCs.Fake.PrependReactor("create", "templateinstances", func(action coretesting.Action) (bool, runtime.Object, error) {
+				createAction := action.(coretesting.CreateAction)
+				instance := createAction.GetObject().(*templateapi.TemplateInstance)
+				expectedTemplate := tc.expected()
+
+				if !equality.Semantic.DeepEqual(expectedTemplate, instance.Spec.Template) {
+					t.Fatal(diff.ObjectDiff(expectedTemplate, instance.Spec.Template))
+				}
+
+				// No errors found, skip the test.
```
stevekuznetsov: This smells -- why do we need it?

pkg/steps/template_test.go
```diff
@@ -0,0 +1,219 @@
...
+				t.Skip()
+
+				return true, nil, nil
+			})
+
+			template := tc.crippleFunc()
+			templateExecutionStep := TemplateExecutionStep(&template, api.NewDeferredParameters(), podClient, tClient, "", makeTestJobSpec(), makeTestResourceConfiguration())
+			templateExecutionStep.Run(context.Background(), false)
```
stevekuznetsov: Doesn't this return an error?

pkg/steps/template_test.go
```diff
@@ -0,0 +1,219 @@
...
+		Requests: coreapi.ResourceList{
+			"cpu":    *resource.NewQuantity(3, resource.DecimalSI),
+			"memory": *resource.NewQuantity(8*1024*1024*1024, resource.BinarySI)},
+		Limits: coreapi.ResourceList{"memory": *resource.NewQuantity(10*1024*1024*1024, resource.BinarySI)},
+	}
+}
+
+func makeTestResourceConfiguration() api.ResourceConfiguration {
+	return map[string]api.ResourceRequirements{
+		"*": {
```
stevekuznetsov: Is this ever used?

stevekuznetsov:
I think we need to focus on a maintainable approach to testing. In general, I think there are some strategies that we can employ here that will make our lives easier and our job simpler:

1. We should factor logic into small, simple functions (not methods). Even if there is only one call-site for a function and it is not helping to promote re-use across a code-base, it is useful to factor self-contained logic into functions. Functions, not methods, allow us to have clearly defined inputs and outputs for stateless effort. Tests for functions are very simple -- list the inputs, list the outputs, call the function and assert that you get what you expect. For instance, in this file we could have the following functions:
  a. partition a template's objects into Pods and non-Pods
  b. apply a set of resources to a Pod
  c. inject an artifacts container to a Pod (you'll notice there's quite a lot of duplicate work in what you've written and what function already implements this step)
  d. append a set of labels to a slice of generic objects
  e. combine a set of Pods and non-Pods back into a template
  f. resolve a set of parameters using a parameter generator
  g. create an `TemplateInstance` from a `Template`
2. Use test-specific helpers when they improve the readability of a test _for a reader looking at  a single test case they did not write_. A reader will look at test code when a test case breaks. Very, very often they look at one case only, as they broke one case only, or they broke everything by violating something really basic but only need to read one case to determine how. The `crippleFunc` pattern (generators and mutators) was originally implemented for the rehearsal jobs test. You will notice there that the mutator function takes the base object as input and creates a mutated copy. When a very complex object is being operated on _and the intent of the test is to assert a small difference_, this approach is useful as it allows the assertion in the test to only present the part of the object that should have changed. Even then, though, the base object is present in the test case definition and all the changes are presented in the mutator. The helpers in this pull request hide test functionality, and make reading more difficult, not easier. It is reasonable for a test to be verbose if it is not simple to present the inputs and outputs concisely. There is no intrinsic value in shortening a test file. Remember, the less a reader has to stray from the definition inside of one test case, the easier their life will be.
3. Shorten test files by reducing the number of tests written at a high level of abstraction. Building on the first comment in this list, if there exists some logic (for example: if a memory request is set for a key that matches the template name, the test Pod inherits that request) it should be tested fully at the smallest unit. By testing the combinations of options at the lowest  level, each test case is reduced to the smallest functional unit and is thereby shorter. If each of the functions above have full unit test coverage, it is then reasonable to have a limited number of (long, complex) integration tests that verify client interaction at a high level (the provided `TemplateInstance` is created, errors are bubbled up if they occur, etc). The function in 1.b. could have a number of cases, and tests at higher levels of abstraction would not need to cover the more than one case:
  a. no resource requests are defined, a Pod named `"test"` exists but is not changed
  b. resource requests are defined, the Pod is not named `"test"` and is not changed
  c. resource requests are defined, the Pod has no pre-existing resource requests and inherits resources from the config
  d. resource requests are defined, the Pod has pre-existing resource requests and has those overridden by the config

I don't necessarily expect you to implement all of this in the PR here. Please note, however, that the tests as written do not even cover the changes that I asked to be tested in the past PR (references in the job spec are added as labels to all objects in the template). If you factor the work you are adding here into a stateless function (not a method) you can write tests for that function without involving any clients. I would also suggest DRYing out your code with the code that adds artifacts containers to Pods.

pkg/steps/template_test.go
droslean: We need to do this or change the template's condition, Otherwise, the test will be stuck waiting.

pkg/steps/template_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	podClient := NewPodClient(fakecs.Core(), nil, nil)
+	templateFakeCs := templatefake.NewSimpleClientset()
+	templateV1 := templateFakeCs.TemplateV1()
+	tClient := NewTemplateClient(templateV1, templateV1.RESTClient())
+
+	for _, tc := range testCases {
+		t.Run(tc.testID, func(t *testing.T) {
+			templateFakeCs.Fake.PrependReactor("create", "templateinstances", func(action coretesting.Action) (bool, runtime.Object, error) {
+				createAction := action.(coretesting.CreateAction)
```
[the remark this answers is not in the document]
droslean: since the `PrependReactor` will fire only in `create` actions, theoretically it will never panic. Either way, I agree that we should check it first.

pkg/steps/template_test.go
droslean: Yep.. I will address this.

pkg/steps/template_test.go
droslean: I wanted to add a case that there are multiple resources in the config. It's not used, and it's not supposed to.

pkg/steps/template.go
```diff
@@ -48,6 +57,52 @@ type templateExecutionStep struct {
...
+					}
+				}
+
+				corev1Scheme := runtime.NewScheme()
+				utilruntime.Must(coreapi.AddToScheme(corev1Scheme))
+				corev1Codec := serializer.NewCodecFactory(corev1Scheme).LegacyCodec(coreapi.SchemeGroupVersion)
+
+				template.Objects[i] = runtime.RawExtension{
+					Raw:    []byte(runtime.EncodeOrDie(corev1Codec, pod)),
+					Object: pod.DeepCopyObject(),
```
stevekuznetsov: Are we doing a deep copy to not change the original? Wouldn't we want to do the deep copy before we change resources?

pkg/steps/template.go
```diff
@@ -48,6 +57,52 @@ type templateExecutionStep struct {
...
+				for index, container := range pod.Spec.Containers {
+					if container.Name == TestContainerName {
+						pod.Spec.Containers[index].Resources = containerResources
+						break
+					}
+				}
+
+				corev1Scheme := runtime.NewScheme()
+				utilruntime.Must(coreapi.AddToScheme(corev1Scheme))
+				corev1Codec := serializer.NewCodecFactory(corev1Scheme).LegacyCodec(coreapi.SchemeGroupVersion)
```
stevekuznetsov: Can we create the codec once instead of for every pod we need to serialize?

pkg/steps/template.go
```diff
@@ -48,6 +57,52 @@ type templateExecutionStep struct {
...
+					Raw:    []byte(runtime.EncodeOrDie(corev1Codec, pod)),
+					Object: pod.DeepCopyObject(),
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func injectLabelsToTemplatePod(jobSpec *api.JobSpec, template *templateapi.Template) {
```
stevekuznetsov: nit: you are injecting to the template labels, not to a pod (directly) so the function name is misleading

pkg/steps/template.go
```diff
@@ -90,13 +149,8 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
...
 	}
 
-	if refs := s.jobSpec.JobSpec.Refs; refs != nil {
-		if s.template.ObjectLabels == nil {
-			s.template.ObjectLabels = make(map[string]string)
-		}
-		s.template.ObjectLabels[RefsOrgLabel] = refs.Org
-		s.template.ObjectLabels[RefsRepoLabel] = refs.Repo
-		s.template.ObjectLabels[RefsBranchLabel] = refs.BaseRef
+	if s.jobSpec != nil {
```
stevekuznetsov: can `jobSpec` ever be `nil`?

pkg/steps/template_test.go
```diff
@@ -0,0 +1,367 @@
...
+							Object: pod.DeepCopyObject(),
+						}
+					}(),
+				},
+			},
+		},
+		{
+			testID: "resource requests are defined, pod has no container named 'test' and is not changed",
+			resources: api.ResourceConfiguration{
+				"test-template": {
```
stevekuznetsov: Do we need a case for when nothing is defined for the specific template name?

pkg/steps/template.go
droslean: theoretically no... I will remove that.

pkg/steps/template_test.go
droslean: I don't think that we need to check that in these tests. This case should belong to [resourcesFor](https://github.com/openshift/ci-tools/blob/master/pkg/steps/source.go#L469) unit tests.

pkg/steps/template.go
droslean: well, this `DeepCopyObject` is just to get the pod as an object for the `RawExtension`.  This pod is the result of the `UnMarsall`  so technically we don't need to `DeepCopy` before changing the resources. So, in the end, we do the `DeepCopy` to actually change the original.

pkg/steps/template.go
droslean: addressed

pkg/steps/template.go
droslean: addressed

pkg/steps/template.go
droslean: addressed

pkg/steps/clusterinstall/clusterinstall.go
```diff
@@ -94,7 +94,7 @@ func E2ETestStep(
 		params = api.NewOverrideParameters(params, overrides)
 	}
 
-	step := steps.TemplateExecutionStep(template, params, podClient, templateClient, artifactDir, jobSpec)
+	step := steps.TemplateExecutionStep(template, params, podClient, templateClient, artifactDir, jobSpec, nil)
```
stevekuznetsov: Don't we have `config.Resources` in this case?

droslean: It seems that we don't. But that smells, we should have. I will check what it takes to implement pass the resources there.

droslean: addressed

PR #85, opened 2019-08-13
[WIP] new test type design

bbguimaraes:
First prototype of the new test type. This version is capable of reproducing the most common template test (AWS E2E). See the corresponding WIP branch [here](https://github.com/bbguimaraes/release/tree/steps).

Some parts of the code are still specific to E2E tests, but I've isolated them in the last commit and am open to suggestions on how to make them more generic.

Other than that, next steps are (sup)porting more test types, adding more tests (which we can finally have), and making sure we have 100% feature parity with templates.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -155,7 +155,7 @@ func FromConfig(
 
 		} else if testStep := rawStep.TestStepConfiguration; testStep != nil {
 			if testStep.MultiStageTestConfiguration != nil {
-				step = steps.MultiStageTestStep(*testStep, config, podClient, secretGetter, artifactDir, jobSpec)
+				step = steps.MultiStageTestStep(*testStep, config, params, podClient, secretGetter, artifactDir, jobSpec)
 			} else if testStep.OpenshiftInstallerClusterTestConfiguration != nil {
 				if testStep.OpenshiftInstallerClusterTestConfiguration.Upgrade {
 					var err error
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -155,7 +155,7 @@ func FromConfig(
 
 		} else if testStep := rawStep.TestStepConfiguration; testStep != nil {
 			if testStep.MultiStageTestConfiguration != nil {
-				step = steps.MultiStageTestStep(*testStep, config, podClient, artifactDir, jobSpec)
+				step = steps.MultiStageTestStep(*testStep, config, podClient, secretGetter, artifactDir, jobSpec)
 			} else if testStep.OpenshiftInstallerClusterTestConfiguration != nil {
 				if testStep.OpenshiftInstallerClusterTestConfiguration.Upgrade {
 					var err error
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -21,13 +21,15 @@ import (
 )
 
 const (
-	multiStageTestLabel = "ci.openshift.io/multi-stage-test"
-	secretMountPath     = "/var/run/secrets/ci.openshift.io/multi-stage"
+	multiStageTestLabel     = "ci.openshift.io/multi-stage-test"
+	clusterProfileMountPath = "/var/run/secrets/ci.openshift.io/cluster-profile"
+	secretMountPath         = "/var/run/secrets/ci.openshift.io/multi-stage"
 )
 
 type multiStageTestStep struct {
 	dry             bool
 	name            string
+	profile         api.ClusterProfile
 	config          *api.ReleaseBuildConfiguration
 	podClient       PodClient
 	secretClient    coreclientset.SecretsGetter
@@ -61,6 +63,7 @@ func newMultiStageTestStep(
 	}
 	return &multiStageTestStep{
 		name:         testConfig.As,
+		profile:      testConfig.MultiStageTestConfiguration.ClusterProfile,
 		config:       config,
 		podClient:    podClient,
 		secretClient: secretClient,
@@ -78,6 +81,14 @@ func (s *multiStageTestStep) Inputs(ctx context.Context, dry bool) (api.InputDef
 
 func (s *multiStageTestStep) Run(ctx context.Context, dry bool) error {
 	s.dry = dry
+	if s.profile != "" {
+		if !dry {
+			profileSecret := fmt.Sprintf("%s-cluster-profile", s.name)
+			if _, err := s.secretClient.Secrets(s.jobSpec.Namespace).Get(profileSecret, meta.GetOptions{}); err != nil {
+				return fmt.Errorf("could not find secret %q: %v", profileSecret, err)
+			}
+		}
+	}
 	if err := createSecret(s.secretClient.Secrets(s.jobSpec.Namespace), s.name, s.dry); err != nil {
 		return fmt.Errorf("failed to create secret: %v", err)
 	}
@@ -162,8 +173,15 @@ func (s *multiStageTestStep) generatePods(steps []api.TestStep) ([]coreapi.Pod, 
 		if owner := s.jobSpec.Owner(); owner != nil {
 			pod.OwnerReferences = append(pod.OwnerReferences, *owner)
 		}
+		if s.profile != "" {
+			addProfile(s.name, s.profile, pod)
+			container.Env = append(container.Env, coreapi.EnvVar{
+				Name:  "KUBECONFIG",
+				Value: filepath.Join(secretMountPath, "kubeconfig"),
+			})
+		}
 		if s.artifactDir != "" && step.ArtifactDir != "" {
-			pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, coreapi.VolumeMount{
+			container.VolumeMounts = append(container.VolumeMounts, coreapi.VolumeMount{
 				Name:      "artifacts",
 				MountPath: step.ArtifactDir,
 			})
@@ -185,6 +203,27 @@ func addSecret(secret string, pod *coreapi.Pod) {
 	pod.Spec.Containers[0].VolumeMounts = append(pod.Spec.Containers[0].VolumeMounts, coreapi.VolumeMount{
 		Name:      secret,
 		MountPath: secretMountPath,
+	})
+}
+
+func addProfile(name string, profile api.ClusterProfile, pod *coreapi.Pod) {
+	volumeName := "cluster-profile"
+	pod.Spec.Volumes = append(pod.Spec.Volumes, coreapi.Volume{
+		Name: volumeName,
+		VolumeSource: coreapi.VolumeSource{
+			Secret: &coreapi.SecretVolumeSource{
+				SecretName: name + "-cluster-profile",
+			},
+		},
+	})
+	container := &pod.Spec.Containers[0]
+	container.VolumeMounts = append(container.VolumeMounts, coreapi.VolumeMount{
+		Name:      volumeName,
+		MountPath: clusterProfileMountPath,
+	})
+	container.Env = append(container.Env, coreapi.EnvVar{
+		Name:  "CLUSTER_TYPE",
+		Value: strings.Split(string(profile), "-")[0],
 	})
 }
 
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -29,8 +29,11 @@ const (
 type multiStageTestStep struct {
 	dry             bool
 	name            string
+	releaseInitial  string
+	releaseLatest   string
 	profile         api.ClusterProfile
 	config          *api.ReleaseBuildConfiguration
+	params          api.Parameters
 	podClient       PodClient
 	secretClient    coreclientset.SecretsGetter
 	artifactDir     string
@@ -42,17 +45,19 @@ type multiStageTestStep struct {
 func MultiStageTestStep(
 	testConfig api.TestStepConfiguration,
 	config *api.ReleaseBuildConfiguration,
+	params api.Parameters,
 	podClient PodClient,
 	secretClient coreclientset.SecretsGetter,
 	artifactDir string,
 	jobSpec *api.JobSpec,
 ) api.Step {
-	return newMultiStageTestStep(testConfig, config, podClient, secretClient, artifactDir, jobSpec)
+	return newMultiStageTestStep(testConfig, config, params, podClient, secretClient, artifactDir, jobSpec)
 }
 
 func newMultiStageTestStep(
 	testConfig api.TestStepConfiguration,
 	config *api.ReleaseBuildConfiguration,
+	params api.Parameters,
 	podClient PodClient,
 	secretClient coreclientset.SecretsGetter,
 	artifactDir string,
@@ -65,6 +70,7 @@ func newMultiStageTestStep(
 		name:         testConfig.As,
 		profile:      testConfig.MultiStageTestConfiguration.ClusterProfile,
 		config:       config,
+		params:       params,
 		podClient:    podClient,
 		secretClient: secretClient,
 		artifactDir:  artifactDir,
@@ -87,6 +93,13 @@ func (s *multiStageTestStep) Run(ctx context.Context, dry bool) error {
 			if _, err := s.secretClient.Secrets(s.jobSpec.Namespace).Get(profileSecret, meta.GetOptions{}); err != nil {
 				return fmt.Errorf("could not find secret %q: %v", profileSecret, err)
 			}
+		}
+		var err error
+		if s.releaseInitial, err = s.params.Get("RELEASE_IMAGE_INITIAL"); err != nil {
+			return err
+		}
+		if s.releaseLatest, err = s.params.Get("RELEASE_IMAGE_LATEST"); err != nil {
+			return err
 		}
 	}
 	if err := createSecret(s.secretClient.Secrets(s.jobSpec.Namespace), s.name, s.dry); err != nil {
@@ -126,6 +139,10 @@ func (s *multiStageTestStep) Requires() (ret []api.StepLink) {
 	}
 	if needsRelease {
 		ret = append(ret, api.ReleaseImagesLink())
+	}
+	if s.profile != "" {
+		ret = append(ret, s.params.Links("RELEASE_IMAGE_INITIAL")...)
+		ret = append(ret, s.params.Links("RELEASE_IMAGE_LATEST")...)
 	}
 	return
 }
@@ -175,10 +192,11 @@ func (s *multiStageTestStep) generatePods(steps []api.TestStep) ([]coreapi.Pod, 
 		}
 		if s.profile != "" {
 			addProfile(s.name, s.profile, pod)
-			container.Env = append(container.Env, coreapi.EnvVar{
-				Name:  "KUBECONFIG",
-				Value: filepath.Join(secretMountPath, "kubeconfig"),
-			})
+			container.Env = append(container.Env, []coreapi.EnvVar{
+				{Name: "KUBECONFIG", Value: filepath.Join(secretMountPath, "kubeconfig")},
+				{Name: "RELEASE_IMAGE_INITIAL", Value: s.releaseInitial},
+				{Name: "RELEASE_IMAGE_LATEST", Value: s.releaseLatest},
+			}...)
 		}
 		if s.artifactDir != "" && step.ArtifactDir != "" {
 			container.VolumeMounts = append(container.VolumeMounts, coreapi.VolumeMount{
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	multiStageTestLabel = "ci.openshift.io/multi-stage-test"
+	secretMountPath     = "/var/run/secrets/ci.openshift.io/multi-stage"
 )
 
 type multiStageTestStep struct {
@@ -29,6 +30,7 @@ type multiStageTestStep struct {
 	name            string
 	config          *api.ReleaseBuildConfiguration
 	podClient       PodClient
+	secretClient    coreclientset.SecretsGetter
 	artifactDir     string
 	jobSpec         *api.JobSpec
 	pre, test, post []api.TestStep
@@ -39,16 +41,18 @@ func MultiStageTestStep(
 	testConfig api.TestStepConfiguration,
 	config *api.ReleaseBuildConfiguration,
 	podClient PodClient,
+	secretClient coreclientset.SecretsGetter,
 	artifactDir string,
 	jobSpec *api.JobSpec,
 ) api.Step {
-	return newMultiStageTestStep(testConfig, config, podClient, artifactDir, jobSpec)
+	return newMultiStageTestStep(testConfig, config, podClient, secretClient, artifactDir, jobSpec)
 }
 
 func newMultiStageTestStep(
 	testConfig api.TestStepConfiguration,
 	config *api.ReleaseBuildConfiguration,
 	podClient PodClient,
+	secretClient coreclientset.SecretsGetter,
 	artifactDir string,
 	jobSpec *api.JobSpec,
 ) *multiStageTestStep {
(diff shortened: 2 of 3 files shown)


pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+		}
+		name := s.name + "-" + step.Name
+		pod, err := generateBasePod(s.jobSpec, resources)
+		pod.ObjectMeta.Name = name
+		pod.ObjectMeta.Annotations[annotationContainersForSubTestResults] = step.Name
+		container := &pod.Spec.Containers[0]
+		container.Name = step.Name
+		container.Image = step.Image
+		container.Command = []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + step.Commands}
+		container.Env = append(container.Env, []coreapi.EnvVar{
```
stevekuznetsov: I think this has changed (perhaps since you last looked) -- look for Hongkai's commit to see how to add all env vars

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+	if errs != nil {
+		return nil, utilerrors.NewAggregate(errs)
+	}
+	return ret, nil
+}
+
+func (s newE2ETestStep) runPods(ctx context.Context, pods []coreapi.Pod) error {
+	go func() {
+		<-ctx.Done()
+		for _, pod := range pods {
```
stevekuznetsov: Could do this with a label selector too, to make for one call

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+			}
+		}
+	}()
+	for _, pod := range pods {
+		log.Printf("Executing %q", pod.Name)
+		if _, err := createOrRestartPod(s.podClient.Pods(s.jobSpec.Namespace), &pod); err != nil {
+			return fmt.Errorf("failed to create or restart %q pod: %v", pod.Name, err)
+		}
+		if err := waitForPodCompletion(s.podClient.Pods(s.jobSpec.Namespace), pod.Name, nil, false); err != nil {
+			return fmt.Errorf("%q pod %q failed: %v", s.name, pod.Name, err)
```
stevekuznetsov: Do we want, for instance, on cleanup tasks after the test has run, for all steps to try to run regardless?

pkg/steps/pod.go
```diff
@@ -193,42 +193,48 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
-	}
+	}, nil
+}
 
+func (s *podStep) generatePodForStep(image string, containerResources coreapi.ResourceRequirements) (*coreapi.Pod, error) {
+	pod, err := generateBasePod(s.jobSpec, containerResources)
+	if err != nil {
+		return nil, err
+	}
+	pod.ObjectMeta.Name = s.config.As
```
stevekuznetsov: All of these look like the same post-generation edits as in the other file. Why not pass them in or otherwise DRY them out?

pkg/steps/new_e2e_test.go
```diff
@@ -13,10 +13,27 @@ import (
...
+func (p testParameters) Get(name string) (string, error)  { return p.m[name], nil }
+func (p testParameters) Links(name string) []api.StepLink { return nil }
+
 func TestGenPods(t *testing.T) {
 	config := api.TestStepConfiguration{
 		As: "test",
 	}
+	params := testParameters{
+		m: map[string]string{
+			"IMAGE_IMAGE0": "registry.example.com/test/image0",
```
stevekuznetsov: Why didn't we see this affect the expected test output?

pkg/api/types.go
```diff
@@ -307,6 +310,14 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerCustomTestImageClusterTestConfiguration *OpenshiftInstallerCustomTestImageClusterTestConfiguration `json:"openshift_installer_custom_test_image,omitempty"`
 }
 
+type TestStep struct {
+	Name        string               `json:"name,omitempty"`
+	Image       string               `json:"image,omitempty"`
+	Commands    string               `json:"commands,omitempty"`
+	ArtifactDir string               `json:"artifact_dir,omitempty"`
+	Resources   ResourceRequirements `json:"resources,omitempty"`
```
stevekuznetsov: Why not use the resource mapping we already have?

cmd/ci-operator/main.go
```diff
@@ -268,7 +271,20 @@ func (o *options) Complete() error {
...
+	if o.stepLibraryPath != "" {
+		data, err := ioutil.ReadFile(o.stepLibraryPath)
+		if err != nil {
+			return fmt.Errorf("failed to read step library (%v)", err)
+		}
+		var l []api.TestStep
+		if err := yaml.Unmarshal(data, &l); err != nil {
+			return fmt.Errorf("failed to load step library (%v)", err)
+		}
+		o.stepLibrary = make(map[string]api.TestStep, len(l))
```
stevekuznetsov: If you want this data structure, lets just serialize it instead of having to do post-processing. What happens if two names collide?

pkg/steps/new_e2e.go
```diff
@@ -139,12 +145,26 @@ func (s newE2ETestStep) genPods(steps []api.TestStep) ([]coreapi.Pod, error) {
 		if owner := s.jobSpec.Owner(); owner != nil {
 			pod.OwnerReferences = append(pod.OwnerReferences, *owner)
 		}
+		if step.ArtifactDir != "" {
+			if artifacts == nil {
```
stevekuznetsov: Having a possibly-nil worker is a bit awkward. Why not:

```go
type collectionRecord struct{ podName, containerName string }
var toCollect []collectionRecord

// ....

for _, step := range steps {
    if step.ArtifactDir != "" {
        // add the volume mount and container
        toCollect = append(toCollect, collectionRecord{podName: pod.Name, containerName: container.Name})
    }
}

if len(toCollect) > 0 {
    // create worker
    // return notifier
}
```

pkg/steps/new_e2e.go
```diff
@@ -139,12 +145,26 @@ func (s newE2ETestStep) genPods(steps []api.TestStep) ([]coreapi.Pod, error) {
 		if owner := s.jobSpec.Owner(); owner != nil {
 			pod.OwnerReferences = append(pod.OwnerReferences, *owner)
 		}
+		if step.ArtifactDir != "" {
```
stevekuznetsov: Do we ever allow for an unset artifacts dir?

pkg/steps/new_e2e_test.go
```diff
@@ -94,8 +94,8 @@ func TestGenPods(t *testing.T) {
 			}},
 		},
 	}}
-	step := newNewE2ETestStep(config, params, nil, &jobSpec, nil)
-	ret, err := step.genPods(steps)
+	step := newNewE2ETestStep(config, params, nil, "", &jobSpec, nil)
+	ret, _, err := step.genPods(steps)
```
stevekuznetsov: Shouldn't we be testing the output of this?

pkg/steps/new_e2e_test.go
```diff
@@ -94,8 +94,8 @@ func TestGenPods(t *testing.T) {
 			}},
 		},
 	}}
-	step := newNewE2ETestStep(config, params, nil, &jobSpec, nil)
-	ret, err := step.genPods(steps)
+	step := newNewE2ETestStep(config, params, nil, "", &jobSpec, nil)
```
stevekuznetsov: How come adding artifacts didn't change any test output?

pkg/steps/new_e2e.go
```diff
@@ -181,6 +199,50 @@ func imageForStep(image string, params api.Parameters) (string, error) {
 	return image, nil
 }
 
+func addSecretMount(pod *coreapi.Pod) {
+	path := "/tmp/installer"
```
stevekuznetsov: this seems awfully specific

pkg/steps/new_e2e.go
```diff
@@ -66,14 +75,20 @@ func (s *newE2ETestStep) Run(ctx context.Context, dry bool) error {
...
+	if secret, err = createSecret(s.secretClient, s.jobSpec.Namespace, s.artifactDir); err != nil {
+		errs = append(errs, fmt.Errorf("failed to create secret: %v", err))
+	}
+	if errs == nil {
+		if err = s.runSteps(ctx, s.tests, secret); err != nil {
 			errs = append(errs, fmt.Errorf("%q test steps failed: %v", s.name, err))
 		}
 	}
-	if err := s.runSteps(ctx, s.post); err != nil {
+	if err = s.runSteps(ctx, s.post, secret); err != nil {
```
stevekuznetsov: This should remain `err :=` ?

pkg/steps/new_e2e.go
```diff
@@ -66,14 +75,20 @@ func (s *newE2ETestStep) Run(ctx context.Context, dry bool) error {
...
 		errs = append(errs, fmt.Errorf("%q pre steps failed: %v", s.name, err))
-	} else {
-		if err := s.runSteps(ctx, s.tests); err != nil {
+	}
+	var secret string
+	if secret, err = createSecret(s.secretClient, s.jobSpec.Namespace, s.artifactDir); err != nil {
+		errs = append(errs, fmt.Errorf("failed to create secret: %v", err))
+	}
+	if errs == nil {
+		if err = s.runSteps(ctx, s.tests, secret); err != nil {
```
stevekuznetsov: This should remain `err :=` ?

pkg/steps/new_e2e.go
```diff
@@ -66,14 +75,20 @@ func (s *newE2ETestStep) Run(ctx context.Context, dry bool) error {
...
+	if err = s.runSteps(ctx, s.pre, ""); err != nil {
 		errs = append(errs, fmt.Errorf("%q pre steps failed: %v", s.name, err))
-	} else {
-		if err := s.runSteps(ctx, s.tests); err != nil {
+	}
+	var secret string
+	if secret, err = createSecret(s.secretClient, s.jobSpec.Namespace, s.artifactDir); err != nil {
+		errs = append(errs, fmt.Errorf("failed to create secret: %v", err))
+	}
+	if errs == nil {
```
stevekuznetsov: The logic here is getting a bit nuts. Why not just return when we can't make the secret?

pkg/steps/new_e2e.go
```diff
@@ -66,14 +75,20 @@ func (s *newE2ETestStep) Run(ctx context.Context, dry bool) error {
 		return nil
 	}
 	var errs []error
-	if err := s.runSteps(ctx, s.pre); err != nil {
+	var err error
+	if err = s.runSteps(ctx, s.pre, ""); err != nil {
```
stevekuznetsov: Don't we need the secret in the pre-steps so the installer can commit the KUBECONFIG to it in the first place?

pkg/steps/new_e2e.go
```diff
@@ -181,6 +199,50 @@ func imageForStep(image string, params api.Parameters) (string, error) {
...
+	if os.IsNotExist(err) {
+		return "", nil
+	} else if err != nil {
+		return "", err
+	}
+	secret := coreapi.Secret{
+		ObjectMeta: meta.ObjectMeta{Name: "installer"},
+		Data:       make(map[string][]byte, len(files)),
+	}
+	for _, f := range files {
```
stevekuznetsov: This seems fragile -- if the secret is write-able (or if we use a configmap) the pre-steps can just self-manage this

pkg/steps/new_e2e.go
```diff
@@ -265,3 +271,73 @@ func (s newE2ETestStep) runPods(ctx context.Context, pods []coreapi.Pod, notifie
 	}
 	return nil
 }
+
+type openshiftInstallerTestStep struct {
```
stevekuznetsov: I don't understand what is going on in this commit

pkg/api/config.go
```diff
@@ -110,8 +110,9 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: '%s' is not valid value, should be [a-zA-Z0-9_.-]", fieldRoot, num, test.As))
 		}
 
-		if len(test.Commands) == 0 {
-			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
+		hasCommands, hasSteps := len(test.Commands) != 0, len(test.Steps) != 0
+		if !(hasCommands || hasSteps) || (hasCommands && hasSteps) {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%d]: either `commands` or `steps` (but not both) should be set", fieldRoot, num))
```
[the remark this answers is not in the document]
bbguimaraes: Separate error messages are fine with me.

As for validation in general, as you said below, it is very incomplete. This was mostly a reminder to do it, I didn't want to spend too much time writing validation code  because I expect the configuration format to change.

pkg/api/config.go
```diff
@@ -110,8 +110,9 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: '%s' is not valid value, should be [a-zA-Z0-9_.-]", fieldRoot, num, test.As))
 		}
 
-		if len(test.Commands) == 0 {
-			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
+		hasCommands, hasSteps := len(test.Commands) != 0, len(test.Steps) != 0
+		if !(hasCommands || hasSteps) || (hasCommands && hasSteps) {
```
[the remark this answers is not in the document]
bbguimaraes: Do you mean the actual `^` operator? That is how I would do it in a language where booleans are `int`s, but I could not find find a way to do it in go.

pkg/api/config.go
```diff
@@ -186,7 +187,7 @@ func validateReleaseTagConfiguration(fieldRoot string, input ReleaseTagConfigura
 
 func validateClusterProfile(fieldRoot string, p ClusterProfile) []error {
 	switch p {
-	case ClusterProfileAWS, ClusterProfileAWSAtomic, ClusterProfileAWSCentos, ClusterProfileAWSCentos40, ClusterProfileAWSGluster, ClusterProfileAzure4, ClusterProfileGCP, ClusterProfileGCP40, ClusterProfileGCPHA, ClusterProfileGCPCRIO, ClusterProfileGCPLogging, ClusterProfileGCPLoggingJournald, ClusterProfileGCPLoggingJSONFile, ClusterProfileGCPLoggingCRIO, ClusterProfileOpenStack, ClusterProfileVSphere:
+	case "", ClusterProfileAWS, ClusterProfileAWSAtomic, ClusterProfileAWSCentos, ClusterProfileAWSCentos40, ClusterProfileAWSGluster, ClusterProfileAzure4, ClusterProfileGCP, ClusterProfileGCP40, ClusterProfileGCPHA, ClusterProfileGCPCRIO, ClusterProfileGCPLogging, ClusterProfileGCPLoggingJournald, ClusterProfileGCPLoggingJSONFile, ClusterProfileGCPLoggingCRIO, ClusterProfileOpenStack, ClusterProfileVSphere:
```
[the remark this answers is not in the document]
bbguimaraes: Mostly a remnant from the idea of having an empty value be "choose one randomly", which is what I used initially to trigger the new code path.

I'll remove it, It has no place in this PR.

pkg/defaults/defaults.go
```diff
@@ -415,7 +420,9 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
 		switch {
```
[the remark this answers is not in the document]
bbguimaraes: In my defense, I only extended the nonsense =)

The change to the last case is also not necessary anymore.

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+		name:      config.As,
+		podClient: podClient,
+		jobSpec:   jobSpec,
+		pre:       config.PreSteps,
+		tests:     config.Steps,
+		post:      config.PostSteps,
+	}
+}
+
+func (s *newE2ETestStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
```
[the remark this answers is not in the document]
bbguimaraes: That is added in a subsequent commit, as you probably saw after commenting, but even then it is not correct. Input resolution happens before we know which images come from `release_tag`, so it won't work.

Note that the same mechanism is used in `imageForStep`, but by then we have already imported the images. I also got strange results when I tested this where the variables were not added to `params` because the IS seemed to be empty. I'll have to investigate this.

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+			errs = append(errs, fmt.Errorf("%q test steps failed: %v", s.name, err))
+		}
+	}
+	if err := s.runSteps(ctx, s.post); err != nil {
+		errs = append(errs, fmt.Errorf("%q post steps failed: %v", s.name, err))
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func (s *newE2ETestStep) Done() (bool, error) { return false, nil }
```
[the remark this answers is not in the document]
bbguimaraes: TODO. Tests pass without it =)

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+func (s *newE2ETestStep) Name() string        { return s.name }
+func (s *newE2ETestStep) Description() string {
+	return fmt.Sprintf("Run cluster install %s", s.name)
```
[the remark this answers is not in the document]
bbguimaraes: Yeah, so is naming of types and files and all the rest. I'll polish when the implementation is solid.

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+		ret = append(ret, api.InternalImageLink(api.PipelineImageStreamTagReference(step.Image)))
+	}
+	return
+}
+
+func (s *newE2ETestStep) Creates() []api.StepLink { return nil }
+func (s *newE2ETestStep) Provides() (api.ParameterMap, api.StepLink) {
+	return nil, nil
+}
+func (s *newE2ETestStep) SubTests() []*junit.TestCase { return nil }
```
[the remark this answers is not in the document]
bbguimaraes: Yes, everything related to junit is TODO.

pkg/steps/new_e2e.go
```diff
@@ -0,0 +1,147 @@
...
+			continue
+		}
+		name := s.name + "-" + step.Name
+		pod, err := generateBasePod(s.jobSpec, resources)
+		pod.ObjectMeta.Name = name
+		pod.ObjectMeta.Annotations[annotationContainersForSubTestResults] = step.Name
+		container := &pod.Spec.Containers[0]
+		container.Name = step.Name
+		container.Image = step.Image
+		container.Command = []string{"/bin/sh", "-c", "#!/bin/sh\nset -eu\n" + step.Commands}
```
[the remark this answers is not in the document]
bbguimaraes: I copied this from `pod.go`, which uses `sh`. Later on I found that the template I used as reference requires `bash`, which is why it is changed in the last commit.

We should change `pod.go` to also use `bash`, should be backwards compatible.

pkg/steps/new_e2e.go
bbguimaraes: Yes, the branch I linked above is based on https://github.com/openshift/release/commit/dc0e476cbb24354a30a58694687657b143aad698, which is a few weeks old now, but I looked briefly at the diff and didn't see anything major.

I really don't want to chase the templates, so I'll keep monitoring them for big changes but will only pull the changes when we are ready.

pkg/steps/new_e2e.go
bbguimaraes: I debated whether `post` should short circuit or if the steps should `exit 0`. I went for the latter only because it made the implementation cleaner.

I believe the templates don't fail if `teardown` fails.

pkg/steps/new_e2e_test.go
bbguimaraes: Because `steps[0].Image` should have been just `image0`. Bad test.

pkg/api/types.go
bbguimaraes: Because the granularity there is per-test, this allows for different resource requirements for each step. I don't really understand why it is a separate top-level field instead of being a field in `TestStepConfiguration`, but we can extend it if there is a good reason.

pkg/steps/new_e2e.go
bbguimaraes: We don't even need the `struct` because we know what `podName` will be based on `containerName`. I also want to check if we can use a single worker for all three phases.

pkg/steps/new_e2e.go
bbguimaraes: "Unset" here means the flag wasn't present in the command line, which is a normal mode of operation. A more general question is: should we extract the information we need via artifacts, which I'll address below…

pkg/steps/new_e2e_test.go
bbguimaraes: Yes, tests are by no means comprehensive yet.

pkg/steps/new_e2e_test.go
bbguimaraes: Because the step doesn't have an `ArtifactDir`.

pkg/steps/new_e2e.go
bbguimaraes: I am having a really hard time with naming, suggestions highly welcome.

pkg/steps/new_e2e.go
bbguimaraes: Artifact from a previous implementation, all `err`s can be scoped now.

pkg/steps/new_e2e.go
bbguimaraes: Depends if we want `post` if `pre` failed.

pkg/steps/new_e2e.go
bbguimaraes: The main reason for repurposing artifacts is because the `installer` image doesn't have `oc` in it. If we can enforce that `pre` steps must be able to modify a secret, I'll gladly remove this.

pkg/steps/new_e2e.go
bbguimaraes: First of all, in the best scenario it is intended to be removed. Up to this point, the implementation is completely generic and has no E2E- (or even Openshift-) specific bits.

The code in `openshiftInstallerTestStep` is everything I still don't know how to make generic that is required for the E2E AWS test.. It is also a possible implementation where different test types would share the step back-end but have test-specific extensions, but I would prefer if we could make all that part of configuration / step implementation.

pkg/api/config.go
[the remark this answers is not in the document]
stevekuznetsov: I was being silly :)

pkg/steps/new_e2e.go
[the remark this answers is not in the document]
stevekuznetsov: I don't think the job will finish without it, though

pkg/steps/new_e2e.go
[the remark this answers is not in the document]
stevekuznetsov: I'm amazed `pod.go` also uses `sh` and this hasn't broken anyone yet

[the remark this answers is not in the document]
stevekuznetsov: Let's update that in a separate PR

pkg/steps/new_e2e.go
stevekuznetsov: It's more of a question of what API we want to expose to users here

pkg/steps/new_e2e.go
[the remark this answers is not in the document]
bbguimaraes: what_if_I_told_you.png

pkg/steps/new_e2e.go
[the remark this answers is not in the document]
bbguimaraes: https://github.com/openshift/ci-tools/pull/87

cmd/ci-operator/main.go
bbguimaraes: I agreed initially, but then realized we would need code here anyway to validate that the key and the `name` field match (or similar treatment) and that duplicates would be silently ignored. Loading a list and validating that names are not duplicated seems better.

pkg/steps/new_e2e.go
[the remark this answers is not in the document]
stevekuznetsov: oh no

PR #86, opened 2019-08-14
Adding checks for additional possible infra failures

ewolinetz:
cc @stevekuznetsov @wking

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -323,7 +323,9 @@ func isInfraReason(reason buildapi.StatusReason) bool {
 }
 
 func hintsAtInfraReason(logSnippet string) bool {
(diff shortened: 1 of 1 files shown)


pkg/api/config_test.go
```diff
@@ -75,7 +75,7 @@ func TestValidateTests(t *testing.T) {
 			id: "test without `commands`",
 			tests: []TestStepConfiguration{
 				{
-					As:                         "test",
+					As: "test",
```
wking: Looks like you washed this through an older `gofmt`?

wking:
what do we use this for?  Is this cache-busting?

ewolinetz:
@wking basically... We had noticed on one of my jobs that the image builds had failed due to `yum` and retesting would immediately fail because of the pre-existing namespace

PR #87, opened 2019-08-15
ci-operator: DRY pod generation, use bash

bbguimaraes:


pkg/steps/pod.go
```diff
@@ -193,44 +193,45 @@ func PodStep(name string, config PodStepConfiguration, resources api.ResourceCon
...
-	}
+	}, nil
+}
 
+func (s *podStep) generatePodForStep(image string, containerResources coreapi.ResourceRequirements) (*coreapi.Pod, error) {
+	pod, err := generateBasePod(s.jobSpec, s.config.As, s.name, image, containerResources)
+	if err != nil {
+		return nil, err
+	}
+	pod.Spec.ServiceAccountName = s.config.ServiceAccountName
```
stevekuznetsov: Why not pass in the SA name and commands to `generateBasePod`?

bbguimaraes: I added `command`, but the SA is usually not set. That is the case for the new test type, and would be the case here if the step wasn't also used by `create_release.go`.

PR #88, opened 2019-08-20
fix unmarhal error log

droslean:
It was wrong to return with an error here and assume that the rest of the objects can't be unmarshaled. Also, logging is not needed here.

/cc @stevekuznetsov


pkg/steps/template.go
```diff
@@ -71,7 +71,7 @@ func injectResourcesToTemplatePods(template *templateapi.Template, resources api
 		for i, object := range template.Objects {
 			var pod *coreapi.Pod
 			if err := yaml.Unmarshal(object.Raw, &pod); err != nil {
-				return fmt.Errorf("couldn't unmashal pod %s: %v", pod.Name, err)
+				continue
```
stevekuznetsov: Can we add a test that has a template with non-Pod objects in it?

PR #89, opened 2019-08-20
WIP Add a tests case in templates, dry out logic of checking object if they are pods

droslean:
follow-up of #88


pkg/steps/artifacts.go
```diff
@@ -731,3 +733,10 @@ func getContainerStatuses(pod *coreapi.Pod) []coreapi.ContainerStatus {
 	statuses = append(statuses, pod.Status.ContainerStatuses...)
 	return statuses
 }
+
+func isObjectPod(obj map[string]interface{}) bool {
```
stevekuznetsov: This is a good factor but it still leaves the two call-sites doing the iteration and marshalling twice. Could we do something like:

```go
func operateOnTemplatePods(template **templateapi.Template, work func(pod *v1.Pod)) error {}
```

PR #90, opened 2019-08-20
Add a new tool: autobumper

hongkailiu:
https://jira.coreos.com/browse/DPTP-518


droslean:
Why do we need to copy `autobumber` from `test-infra`?

hongkailiu:
when this pr gets merged, I will add the relevant piece into our job config.
So the whole build procedure is in place.
Then we can focus on our customization.

hongkailiu:
Why bump the dep. version of test-infra (I think we should bump its version for each PR in ci-tools so that we can find problem as early as possible):
Without bumping, the latest autobumper code does not work with our version of test-infra in vendor.

cmd/autobumper/main.go
```diff
@@ -0,0 +1,334 @@
...
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/robots/pr-creator/updater"
+)
+
+const (
+	prowPrefix      = "gcr.io/k8s-prow/"
+	testImagePrefix = "gcr.io/k8s-testimages/"
+	prowRepo        = "https://github.com/kubernetes/test-infra"
+	testImageRepo   = prowRepo
+	oncallAddress   = "https://storage.googleapis.com/kubernetes-jenkins/oncall.json"
```
droslean: well, we definitely don't need that.

hongkailiu: Just a starting point ... we will do our customization later.
But since you get confused too, I will start with an empty main.

PR #94, opened 2019-08-20
Inject resources in template pod's container named "test"

droslean:
With this patch, we allow resources injection to the template pod's test container. 

The logic is the following:
* If the test container has already resources defined, use them.
* else If the ci-operator config has resources defined with the template's name, inject the resources to the test container.
* else If there are only the default resources defined in the ci-operator's config, use them instead.

Note: The resources will be injected only if the template has a pod and the pod has a container named "test"

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -36,10 +36,13 @@ const (
 	RefsOrgLabel    = "ci.openshift.io/refs.org"
 	RefsRepoLabel   = "ci.openshift.io/refs.repo"
 	RefsBranchLabel = "ci.openshift.io/refs.branch"
+
+	TestContainerName = "test"
 )
 
 type templateExecutionStep struct {
 	template       *templateapi.Template
+	resources      api.ResourceConfiguration
 	params         api.Parameters
 	templateClient TemplateClient
 	podClient      PodClient
@@ -92,7 +95,7 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
 		}
 	}
 
-	operateOnTemplatePods(s.template, s.artifactDir)
+	operateOnTemplatePods(s.template, s.artifactDir, s.resources)
 	injectLabelsToTemplate(s.jobSpec, s.template)
 
 	if dry {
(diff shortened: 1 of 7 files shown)


pkg/api/types.go
```diff
@@ -86,6 +86,18 @@ func (c ResourceConfiguration) RequirementsForStep(name string) ResourceRequirem
 	return req
 }
 
+func (c ResourceConfiguration) RequirementsForTemplateStep(name string) ResourceRequirements {
```
stevekuznetsov: I don't think this is correct -- instead of never using the default, we should only use it when the template itself does not set resource requirements

pkg/steps/template.go
```diff
@@ -48,6 +57,61 @@ type templateExecutionStep struct {
 	subTests []*junit.TestCase
 }
 
+func injectResourcesToTemplatePods(template *templateapi.Template, resources api.ResourceConfiguration) error {
+	if resources != nil {
```
stevekuznetsov: nit: passing an argument into a function and then using it to determine if work should be done is a smell -- don't call the function at all if you don't have resources to set

pkg/steps/template.go
```diff
@@ -48,6 +57,61 @@ type templateExecutionStep struct {
...
+		containerResources, err := resourcesFor(resources.RequirementsForTemplateStep(template.Name))
+		if err != nil {
+			return fmt.Errorf("unable to calculate resources for %s: %s", template.Name, err)
+		}
+
+		corev1Scheme := runtime.NewScheme()
+		utilruntime.Must(coreapi.AddToScheme(corev1Scheme))
+		corev1Codec := serializer.NewCodecFactory(corev1Scheme).LegacyCodec(coreapi.SchemeGroupVersion)
+
+		for i, object := range template.Objects {
```
stevekuznetsov: This is still really duplicated with the other call-site. Why not  do this:

```go
func operateOnTemplatePods(template *templateapi.Template, func(pod *coreapi.Pod)) error {}
```

The fact that the code in the other file does crazy JSON marshalling is not important -- we should refactor it to mutate the Pod in Go code and only marshal it at the end to make the raw JSON, just like you are doing here for this. It is not important that we iterate over the list twice. O(2N) === O(N).

stevekuznetsov: @smarterclayton could you please give us the reasoning behind the JSON-based workflow you implemented to inject the artifacts container?

droslean: this is why I left the PR as WIP.  I will address this. Any more insights are welcome.

droslean:
/hold
Depends on #96

droslean:
WIP -> upgrade the integration tests.

pkg/steps/template.go
```diff
@@ -50,6 +53,22 @@ type templateExecutionStep struct {
 	subTests []*junit.TestCase
 }
 
+func injectResourcesToPod(pod *coreapi.Pod, templateName string, resources api.ResourceConfiguration) error {
+	containerResources, err := resourcesFor(resources.RequirementsForStep(templateName))
```
droslean: That will return the resources for the specific template, but if there is nothing defined it will return the '*' resources.

pkg/steps/template_test.go
```diff
@@ -230,11 +233,81 @@ func TestOperateOnTemplatePods(t *testing.T) {
 				},
 			},
 		},
+		{
+			testID: "resources defined in the config but the test container has existing resources, no changes expected",
```
petr-muller: I think we also want to test the other scenario, where the resources are propagated from the config to the container.

droslean: addressed

PR #95, opened 2019-08-20
ci-operator: add TestStep type

bbguimaraes:
Slightly modified subset of https://github.com/openshift/ci-tools/pull/85.

Open questions:

- Is a step with just a `name` the best way to reference a pre-defined step?  See the original PR for details and https://github.com/bbguimaraes/release/tree/steps for an example.
- Is `resources` required? That is the case for other ci-operator tests, but not for templates.
- `TestStep` is a terrible name.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -307,6 +307,14 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerCustomTestImageClusterTestConfiguration *OpenshiftInstallerCustomTestImageClusterTestConfiguration `json:"openshift_installer_custom_test_image,omitempty"`
 }
 
+type TestStep struct {
+	Name        string               `json:"name,omitempty"`
+	Image       string               `json:"image,omitempty"`
+	Commands    string               `json:"commands,omitempty"`
+	ArtifactDir string               `json:"artifact_dir,omitempty"`
+	Resources   ResourceRequirements `json:"resources,omitempty"`
+}
+
 // Secret describes a secret to be mounted inside a test
 // container.
 type Secret struct {
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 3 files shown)


stevekuznetsov:
As a first pass, we should be concerned with identifying the set of information necessary to work with downstream/at runtime. Let's validate that every field exists, resources are set, etc. No concept of name resolution yet

bbguimaraes:
Another one:

- We need a way to reference an image that is not part of the pipeline,

Container tests don't have that problem because they simply don't allow it, templates use the `IMAGE_`/`LOCAL_IMAGE_` prefixes.

pkg/api/types.go
```diff
@@ -307,6 +311,14 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerCustomTestImageClusterTestConfiguration *OpenshiftInstallerCustomTestImageClusterTestConfiguration `json:"openshift_installer_custom_test_image,omitempty"`
 }
 
+type TestStep struct {
+	Name        string               `json:"name,omitempty"`
+	Image       string               `json:"image,omitempty"`
```
stevekuznetsov: As this is a `string` and not a `PipelineImageReference` don't we already allow any pull spec here?

bbguimaraes: There are two points in the execution where this is used:

- When configuration is loaded, to determine dependencies between (ci-operator) steps.
- When the pods are created.

The biggest problem is #1. We know at that point the images that are part of the pipeline, but not those that come from `tag_specification`.

For #2, we can always use `pipeline:src`, `stable:tests`, or even a full spec, which is what I've been using, but since #1 requires us to have that information, that should not be required.

stevekuznetsov: So we generally operate on the concept that if an image is named `foo` and there exists a `foo` in the images brought in by the `tag_specification`, the built image overrides that which is pulled in. So, we can accept a simple string here and:

 - if it matches something that is built by the pipeline or the `images`, create that link in the steps
 - if it does not, expect it to be provided by the import from `tag-specification`

bbguimaraes: That's what I wanted to hear/read.

PR #97, opened 2019-08-22
rehearse jobs when optional and always_run changes.

droslean:
It makes sense to rehearse the job when it changes to non-optional and when `always_run` changes to `true`.

That will help our stakeholders to make sure that they won't block anything.


bbguimaraes:
LGTM, can we add tests?

PR #98, opened 2019-08-22
Implement multi-stage test registry interface

AlexNPavel:
This implements a few new structs to allow us to read in user test configs and a function to expand the user's config (which may contain references, chains, etc) into a full internal test representation. This contains some duplicate code from `TestStepConfiguration` as this is essentially the evolution of that struct, but I did not want to trample that code while this new config format is still in a WIP state.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -271,6 +271,12 @@ func validateTestConfigurationType(fieldRoot string, test TestStepConfiguration,
 		typeCount++
 		validationErrors = append(validationErrors, validateClusterProfile(fmt.Sprintf("%s", fieldRoot), testConfig.ClusterProfile)...)
 	}
+	if testConfig := test.MultiStageTestConfiguration; testConfig != nil {
+		typeCount++
+		if testConfig.ClusterProfile != "" {
+			validationErrors = append(validationErrors, validateClusterProfile(fmt.Sprintf("%s", fieldRoot), testConfig.ClusterProfile)...)
+		}
+	}
 	if test.OpenshiftInstallerRandomClusterTestConfiguration != nil {
 		typeCount++
 	}
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -294,6 +294,7 @@ type TestStepConfiguration struct {
 
 	// Only one of the following can be not-null.
 	ContainerTestConfiguration                                *ContainerTestConfiguration                                `json:"container,omitempty"`
+	MultiStageTestConfiguration                               *MultiStageTestConfiguration                               `json:"steps,omitempty"`
 	OpenshiftAnsibleClusterTestConfiguration                  *OpenshiftAnsibleClusterTestConfiguration                  `json:"openshift_ansible,omitempty"`
 	OpenshiftAnsibleSrcClusterTestConfiguration               *OpenshiftAnsibleSrcClusterTestConfiguration               `json:"openshift_ansible_src,omitempty"`
 	OpenshiftAnsibleCustomClusterTestConfiguration            *OpenshiftAnsibleCustomClusterTestConfiguration            `json:"openshift_ansible_custom,omitempty"`
@@ -307,12 +308,25 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerCustomTestImageClusterTestConfiguration *OpenshiftInstallerCustomTestImageClusterTestConfiguration `json:"openshift_installer_custom_test_image,omitempty"`
 }
 
+// TestStep is the external representation of a test step that allows users to
+// write less verbose test configs. It gets converted to an internal TestStep
+// struct that represents the full configuration that ci-operator can use.
 type TestStep struct {
-	Name        string               `json:"name,omitempty"`
-	Image       string               `json:"image,omitempty"`
-	Commands    string               `json:"commands,omitempty"`
-	ArtifactDir string               `json:"artifact_dir,omitempty"`
-	Resources   ResourceRequirements `json:"resources,omitempty"`
+	Name          string               `json:"name,omitempty"`
+	Documentation string               `json:"documentation,omitempty"`
+	Image         string               `json:"image,omitempty"`
+	Commands      string               `json:"commands,omitempty"`
+	ArtifactDir   string               `json:"artifact_dir,omitempty"`
+	Resources     ResourceRequirements `json:"resources,omitempty"`
+}
+
+// MultiStageTestConfiguration is a flexible configuration mode that allows tighter control over
+// the multiple stages of end to end tests
+type MultiStageTestConfiguration struct {
+	ClusterProfile ClusterProfile `json:"cluster_profile"`
+	Pre            []TestStep     `json:"pre,omitempty"`
+	Test           []TestStep     `json:"test,omitempty"`
+	Post           []TestStep     `json:"post,omitempty"`
 }
 
 // Secret describes a secret to be mounted inside a test
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -1,0 +1,55 @@
+package registry
+
+import (
+	api "github.com/openshift/ci-tools/pkg/api"
+	types "github.com/openshift/ci-tools/pkg/steps/types"
+	"k8s.io/apimachinery/pkg/util/errors"
+)
+
+type Resolver interface {
+	Resolve(config api.MultiStageTestConfiguration) (types.TestFlow, error)
+}
+
+// Registry will hold all the registry information needed to convert between the
+// user provided configs referencing the registry and the internal, complete
+// representation
+type registry struct{}
+
+func NewResolver() Resolver {
+	return &registry{}
+}
+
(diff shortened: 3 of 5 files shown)


pkg/api/types.go
```diff
@@ -315,6 +316,34 @@ type TestStep struct {
 	Resources   ResourceRequirements `json:"resources,omitempty"`
 }
 
+// TestStepCompact is the external representation of a test step that allows users to
+// write less verbose test configs
+type TestStepCompact struct {
```
stevekuznetsov: Usually we have the same name under two different packages. Can we move the internal one to `pkg/steps` or similar and then we won't need the `-Compact` suffix?

pkg/api/test_flow.go
```diff
@@ -0,0 +1,33 @@
+package api
+
+func ExpandFlow(flow TestFlowCompact) (TestFlow, []error) {
```
stevekuznetsov: Let's create a new package for this `pkg/registry` and a formal interface for it, maybe:

```go
type Resolver interface {
    Resolve (input api.TestFlow) (output steps.TestFlow, err error)
}
```

pkg/api/test_flow.go
```diff
@@ -0,0 +1,33 @@
...
+	expandedFlow := TestFlow{}
+	for _, external := range flow.Pre {
+		newStep := TestStep{}
```
stevekuznetsov: Maybe we can DRY this out with a function like

```go
func toInternal(input api.TestStep) (output steps.TestStep) {}
```

pkg/api/types.go
```diff
@@ -307,6 +307,7 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerCustomTestImageClusterTestConfiguration *OpenshiftInstallerCustomTestImageClusterTestConfiguration `json:"openshift_installer_custom_test_image,omitempty"`
 }
 
+// TestStep is an internal representation of a full test step that ci-operator can run
 type TestStep struct {
 	Name        string               `json:"name,omitempty"`
```
stevekuznetsov: After moving this, the internal types also won't need `json` annotations as they are not read from config

pkg/api/config.go
```diff
@@ -211,6 +211,24 @@ func searchForTestDuplicates(tests []TestStepConfiguration) []error {
 	return nil
 }
 
+func searchForTestConfigDuplicates(tests []TestConfig) []error {
```
stevekuznetsov: The new test type should be an optional field under the `TestStepConfiguration`, so this function will be redundant with `searchForTestDuplicates`

pkg/api/config.go
```diff
@@ -211,6 +211,24 @@ func searchForTestDuplicates(tests []TestStepConfiguration) []error {
...
+		}
+	}
+
+	if testNames != nil {
+		return []error{fmt.Errorf("tests: found duplicated test: (%s)", strings.Join(testNames, ","))}
+	}
+	return nil
+}
+
 func validateTestConfigurationType(fieldRoot string, test TestStepConfiguration, release *ReleaseTagConfiguration) []error {
```
stevekuznetsov: Once we add the `TestFlow` as an option for a `TestStepConfiguration`, we should update the `validateTestConfigurationType` as well

pkg/registry/resolver.go
```diff
@@ -0,0 +1,54 @@
...
+
+import (
+	api "github.com/openshift/ci-tools/pkg/api"
+	types "github.com/openshift/ci-tools/pkg/steps/types"
+)
+
+// Registry will hold all the registry information needed to convert between the
+// user provided configs referencing the registry and the internal, complete
+// represenation
+type Registry struct{}
```
stevekuznetsov: Using an interface here will allow us to mock out this functionality in tests that do need to validate the internals.

If we have:

```go
type Resolver interface {
    Resolve(config api.MultiStageTestConfiguration) (types.TestFlow, []error)
}

type registry struct{}

func (r *registry) Resolve(config api.MultiStageTestConfiguration) (types.TestFlow, []error) {}

func NewResolver() Resolver {
  return &registry{}
}
```

Then, a downstream user can require something like this:

```go
type SomeCIOperatorThing struct {
    resolver registry.Resolver
}
```

This lets us pass in `NewResolver()` in the production code, then mock out the functionality otherwise in tests. For instance, we might write a test that CI Operator actually contacts the resolver, so we could write a test mock like:

```go
type recordingResolver struct {
    requests []api.MultiStageTestConfiguration
    mockResponse types.TestFlow
    mockError []error
}

func (r *recordingResolver) Resolve(config api.MultiStageTestConfiguration) (types.TestFlow, []error) {
    r.requests = append(r.requests, config)
    return r.mockResponse. r.mockError
}
```

This allows us, in a test, to use dependency injection to:

 - record the calls to the resolver and assert that we see what we want
 - output a specific response and assert that the downstream logic uses it correctly
 - output an error and assert that it is handled correctly

pkg/registry/resolver.go
```diff
@@ -0,0 +1,54 @@
...
+
+func (r Registry) Resolve(config api.MultiStageTestConfiguration) (types.TestFlow, []error) {
```
stevekuznetsov: Let's return one `error` -- we can use `errors.NewAggregate(many, separate, errors)` to coalesce many errors into one. If the downstream consumer is not interested in ever breaking this array of errors apart, we do not need to return more than one `error`.

pkg/registry/resolver.go
```diff
@@ -0,0 +1,54 @@
...
+		}
+		expandedFlow.Test = append(expandedFlow.Test, newStep)
+	}
+	if resolveErrors != nil {
+		return types.TestFlow{}, resolveErrors
+	}
+	return expandedFlow, nil
+}
+
+func toInternal(input api.TestStep) (types.TestStep, []error) {
```
stevekuznetsov: nit: since this cannot error we can omit that from the signature

It is still OK to have the error in the signature for `Resolve()` since we expect it to be possible to return an error in the future.

pkg/registry/resolver.go
```diff
@@ -0,0 +1,54 @@
...
+	newStep := types.TestStep{}
```
stevekuznetsov: nit: this could be

```go
return types.TestStep{
    Name: input.Name,
    Image: input.Image,
    Commands: input.Commands,
    ArtifactDir: input.ArtifactDir,
    Resources: input.Resources,
}
```

pkg/steps/types/types.go
```diff
@@ -0,0 +1,19 @@
+package steptypes
+
+import (
+	api "github.com/openshift/ci-tools/pkg/api"
+)
+
+type TestStep struct {
```
stevekuznetsov: These could use GoDoc

pkg/steps/types/types.go
```diff
@@ -0,0 +1,19 @@
...
+	Name        string
+	Image       string
+	Commands    string
+	ArtifactDir string
+	Resources   api.ResourceRequirements
+}
+
+type TestFlow struct {
```
stevekuznetsov: You mentioned this on Slack but just so we don't lose it, we should pass through the cluster profile here

pkg/api/types.go
```diff
@@ -307,6 +308,8 @@ type TestStepConfiguration struct {
 	OpenshiftInstallerCustomTestImageClusterTestConfiguration *OpenshiftInstallerCustomTestImageClusterTestConfiguration `json:"openshift_installer_custom_test_image,omitempty"`
 }
 
+// TestStep is the external representation of a test step that allows users to
+// write less verbose test configs
 type TestStep struct {
```
stevekuznetsov: Let's also add

```go
Documentation string `json:documentation`
```

Since we will want that for the registry.

pkg/api/config.go
```diff
@@ -271,6 +271,10 @@ func validateTestConfigurationType(fieldRoot string, test TestStepConfiguration,
 		typeCount++
 		validationErrors = append(validationErrors, validateClusterProfile(fmt.Sprintf("%s", fieldRoot), testConfig.ClusterProfile)...)
 	}
+	if testConfig := test.MultiStageTestConfiguration; testConfig != nil {
+		typeCount++
+		validationErrors = append(validationErrors, validateClusterProfile(fmt.Sprintf("%s", fieldRoot), testConfig.ClusterProfile)...)
```
bbguimaraes: `validateClusterProfile` requires the field to have a value. I expect there will be tests that don't involve a cluster, so it should not be required.

pkg/registry/resolver.go
```diff
@@ -0,0 +1,55 @@
...
+	"k8s.io/apimachinery/pkg/util/errors"
+)
+
+type Resolver interface {
+	Resolve(config api.MultiStageTestConfiguration) (types.TestFlow, error)
+}
+
+// Registry will hold all the registry information needed to convert between the
+// user provided configs referencing the registry and the internal, complete
+// represenation
```
petr-muller: `represenTation`

PR #99, opened 2019-08-26
add job pathalias to the extraref with workdif

droslean:
Adds the job's config `path_alias` to the `extraRef`. With this, we assume that the `path_alias` in the extrarefs will override the `path_alias` in the Refs while cloning. 

A proper fix should be done upstream.


droslean:
/hold 

Testing...

PR #92, opened 2019-08-20
Inject org/repo/branch labels to pods created by ci-operator

droslean:
/cc @stevekuznetsov

PR #100, opened 2019-08-26
ci-operator multi-stage: MultiStageTestStep

bbguimaraes:
Basic implementation of the multi-stage test type. Subset of
https://github.com/openshift/ci-tools/pull/85, based on
https://github.com/openshift/ci-tools/pull/98. Design document:
https://docs.google.com/document/d/1md-1BMf4_7mtKgGVoeZ3jOh4zSIBSjwl6vTTAYESwIM.

Fully-functional but lacking management of artifacts, the test secret, cluster
profile, and release image(s).

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -103,33 +103,39 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 	validationErrors = append(validationErrors, searchForTestDuplicates(input)...)
 
 	for num, test := range input {
+		fieldRootN := fmt.Sprintf("%s[%d]", fieldRoot, num)
 		if len(test.As) == 0 {
-			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: is required", fieldRoot, num))
+			validationErrors = append(validationErrors, fmt.Errorf("%s.as: is required", fieldRootN))
 		} else if test.As == "images" {
-			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: should not be called 'images' because it gets confused with '[images]' target", fieldRoot, num))
+			validationErrors = append(validationErrors, fmt.Errorf("%s.as: should not be called 'images' because it gets confused with '[images]' target", fieldRootN))
 		} else if ok := regexp.MustCompile("^[a-zA-Z0-9_.-]*$").MatchString(test.As); !ok {
-			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].as: '%s' is not valid value, should be [a-zA-Z0-9_.-]", fieldRoot, num, test.As))
+			validationErrors = append(validationErrors, fmt.Errorf("%s.as: '%s' is not valid value, should be [a-zA-Z0-9_.-]", fieldRootN, test.As))
 		}
-
-		if len(test.Commands) == 0 {
-			validationErrors = append(validationErrors, fmt.Errorf("%s[%d].commands: is required", fieldRoot, num))
+		if hasCommands, hasSteps := len(test.Commands) != 0, test.MultiStageTestConfiguration != nil; !hasCommands && !hasSteps {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: either `commands` or `steps` should be set", fieldRootN))
+		} else if hasCommands && hasSteps {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `commands` and `steps` are mutually exclusive", fieldRootN))
+		} else if hasSteps {
+			validationErrors = append(validationErrors, validateTestSteps(fieldRootN+".steps.pre", test.MultiStageTestConfiguration.Pre)...)
+			validationErrors = append(validationErrors, validateTestSteps(fieldRootN+".steps.test", test.MultiStageTestConfiguration.Test)...)
+			validationErrors = append(validationErrors, validateTestSteps(fieldRootN+".steps.post", test.MultiStageTestConfiguration.Post)...)
 		}
 
 		if test.Secret != nil {
 			// TODO: Move to upstream validation when vendoring is fixed
 			// currently checking against DNS RFC 1123 regexp
 			if ok := regexp.MustCompile("^[a-z0-9]([-a-z0-9]*[a-z0-9])?$").MatchString(test.Secret.Name); !ok {
-				validationErrors = append(validationErrors, fmt.Errorf("%s[%d].name: '%s' secret name is not valid value, should be [a-z0-9]([-a-z0-9]*[a-z0-9]", fieldRoot, num, test.Secret.Name))
+				validationErrors = append(validationErrors, fmt.Errorf("%s.name: '%s' secret name is not valid value, should be [a-z0-9]([-a-z0-9]*[a-z0-9]", fieldRootN, test.Secret.Name))
 			}
 			// validate path only if name is passed
 			if test.Secret.MountPath != "" {
 				if ok := filepath.IsAbs(test.Secret.MountPath); !ok {
-					validationErrors = append(validationErrors, fmt.Errorf("%s[%d].path: '%s' secret mount path is not valid value, should be ^((\\/*)\\w+)+", fieldRoot, num, test.Secret.MountPath))
+					validationErrors = append(validationErrors, fmt.Errorf("%s.path: '%s' secret mount path is not valid value, should be ^((\\/*)\\w+)+", fieldRootN, test.Secret.MountPath))
 				}
 			}
 		}
 
-		validationErrors = append(validationErrors, validateTestConfigurationType(fmt.Sprintf("%s[%d]", fieldRoot, num), test, release)...)
+		validationErrors = append(validationErrors, validateTestConfigurationType(fieldRootN, test, release)...)
 	}
 	return validationErrors
 }
--- a/pkg/api/types_test.go
+++ b/pkg/api/types_test.go
@@ -134,3 +134,65 @@ func TestOverlay(t *testing.T) {
 		})
 	}
 }
+
+func TestBuildsImage(t *testing.T) {
+	conf := ReleaseBuildConfiguration{
+		Images: []ProjectDirectoryImageBuildStepConfiguration{
+			{To: "this-image-is-in-the-images-field"},
+		},
+	}
+	for _, tc := range []struct {
+		name  string
+		image string
+		want  bool
+	}{{
+		name:  "not in `images`",
+		image: "this-image-is-not-in-the-images-field",
+	}, {
+		name:  "in `images`",
+		image: "this-image-is-in-the-images-field",
+		want:  true,
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			if ret := conf.BuildsImage(tc.image); ret != tc.want {
+				t.Errorf("got %v, want %v", ret, tc.want)
+			}
+		})
+	}
+}
+
+func TestIsPipelineImage(t *testing.T) {
+	conf := ReleaseBuildConfiguration{
+		InputConfiguration: InputConfiguration{
+			BaseImages: map[string]ImageStreamTagReference{
+				"base-img": {},
+			},
+			BaseRPMImages: map[string]ImageStreamTagReference{
+				"base-rpm-img": {},
+			},
+		},
+		BinaryBuildCommands:     "make",
+		TestBinaryBuildCommands: "make test-bin",
+		RpmBuildCommands:        "make rpms",
+		Images:                  []ProjectDirectoryImageBuildStepConfiguration{{To: "img"}},
+	}
+	for _, tc := range []struct {
+		name string
+		want bool
+	}{
+		{name: "base-img", want: true},
+		{name: "base-rpm-img", want: true},
+		{name: "root", want: true},
+		{name: "bin", want: true},
+		{name: "test-bin", want: true},
+		{name: "rpms", want: true},
+		{name: "img"},
+		{name: "404"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			if ret := conf.IsPipelineImage(tc.name); ret != tc.want {
+				t.Errorf("want %v, got %v", tc.want, ret)
+			}
+		})
+	}
+}
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -1,7 +1,6 @@
 package defaults
 
 import (
-	"crypto/sha256"
 	"fmt"
 	"log"
 	"net/url"
@@ -99,7 +98,7 @@ func FromConfig(
 
 	params := api.NewDeferredParameters()
 	params.Add("JOB_NAME", nil, func() (string, error) { return jobSpec.Job, nil })
-	params.Add("JOB_NAME_HASH", nil, func() (string, error) { return fmt.Sprintf("%x", sha256.Sum256([]byte(jobSpec.Job)))[:5], nil })
+	params.Add("JOB_NAME_HASH", nil, func() (string, error) { return jobSpec.JobNameHash(), nil })
 	params.Add("JOB_NAME_SAFE", nil, func() (string, error) { return strings.Replace(jobSpec.Job, "_", "-", -1), nil })
 	params.Add("NAMESPACE", nil, func() (string, error) { return jobSpec.Namespace, nil })
 
@@ -155,7 +154,9 @@ func FromConfig(
 			buildSteps = append(buildSteps, initialReleaseStep)
 
 		} else if testStep := rawStep.TestStepConfiguration; testStep != nil {
-			if testStep.OpenshiftInstallerClusterTestConfiguration != nil {
+			if testStep.MultiStageTestConfiguration != nil {
+				step = steps.MultiStageTestStep(*testStep, config, podClient, jobSpec)
+			} else if testStep.OpenshiftInstallerClusterTestConfiguration != nil {
 				if testStep.OpenshiftInstallerClusterTestConfiguration.Upgrade {
 					var err error
 					step, err = clusterinstall.E2ETestStep(*testStep.OpenshiftInstallerClusterTestConfiguration, *testStep, params, podClient, templateClient, secretGetter, artifactDir, jobSpec)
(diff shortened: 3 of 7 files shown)


pkg/api/types_test.go
```diff
@@ -134,3 +134,61 @@ func TestOverlay(t *testing.T) {
...
+			BaseRPMImages: map[string]ImageStreamTagReference{
+				"base-rpm-img": {},
+			},
+		},
+		BinaryBuildCommands:     "make",
+		TestBinaryBuildCommands: "make test-bin",
+		RpmBuildCommands:        "make rpms",
+		Images:                  []ProjectDirectoryImageBuildStepConfiguration{{To: "img"}},
+	}
+	for _, i := range []string{
```
petr-muller: Personally I would prefer this to be `t.Run()` based so it's executed in parallel and does not stop on first failure.

pkg/defaults/defaults.go
```diff
@@ -155,7 +154,12 @@ func FromConfig(
 			buildSteps = append(buildSteps, initialReleaseStep)
 
 		} else if testStep := rawStep.TestStepConfiguration; testStep != nil {
-			if testStep.OpenshiftInstallerClusterTestConfiguration != nil {
+			if testStep.MultiStageTestConfiguration != nil {
+				var err error
+				if step, err = steps.MultiStageTestStep(*testStep, config, podClient, jobSpec); err != nil {
+					return nil, nil, fmt.Errorf("unable to create end to end test step: %v", err)
```
petr-muller: Are we actually calling this `end to end step` in the user-facing documentation? The design doc calls it multi-stage tests, I think we should call it that way even here.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+}
+
+func (s *multiStageTestStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
+	return nil, nil
+}
+
+func (s *multiStageTestStep) Run(ctx context.Context, dry bool) error {
+	if dry {
+		return nil
```
petr-muller: The other steps output a JSON of what would they submit while in dry mode, I think we should do the same.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+	if err := s.runSteps(ctx, s.post, false); err != nil {
+		errs = append(errs, fmt.Errorf("%q post steps failed: %v", s.name, err))
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func (s *multiStageTestStep) Done() (bool, error) { return false, nil }
+func (s *multiStageTestStep) Name() string        { return s.name }
+func (s *multiStageTestStep) Description() string {
+	return fmt.Sprintf("Run test %s", s.name)
```
petr-muller: "Run multi-stage test..."?

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+	if needsImages {
+		ret = append(ret, api.ImagesReadyLink())
+	}
+	if needsRelease {
+		ret = append(ret, api.ReleaseImagesLink())
+	}
+	return
+}
+
+func (s *multiStageTestStep) Creates() []api.StepLink { return nil }
```
petr-muller: Hmm... `templateExecutionStep` and `podStep` return an empty slice, not nil. Does this matter?

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+func (s *multiStageTestStep) Provides() (api.ParameterMap, api.StepLink) {
+	return nil, nil
+}
+func (s *multiStageTestStep) SubTests() []*junit.TestCase { return s.subTests }
+
+func (s multiStageTestStep) runSteps(ctx context.Context, steps []api.TestStep, shortCircuit bool) error {
```
petr-muller: Hmm... is the non-pointer receiver here intentional?

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+		}...)
+		if owner := s.jobSpec.Owner(); owner != nil {
+			pod.OwnerReferences = append(pod.OwnerReferences, *owner)
+		}
+		ret = append(ret, *pod)
+	}
+	if errs != nil {
+		return nil, utilerrors.NewAggregate(errs)
+	}
+	return ret, nil
```
petr-muller: This could be simplified to just `return ret, utilerrors.NewAggregate(errs)` as `NewAggregate` is documented to return nil for an empty slice.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+	return s.runPods(ctx, pods, shortCircuit)
+}
+
+func (s multiStageTestStep) genPods(steps []api.TestStep) ([]coreapi.Pod, error) {
+	var ret []coreapi.Pod
+	var errs []error
+	for _, step := range steps {
+		image := step.Image
+		if s.config.IsPipelineImage(step.Image) {
+			image = api.PipelineImageStream + ":" + step.Image
```
petr-muller: would prefer `Sprintf` over concatenation

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+		}
+		resources, err := resourcesFor(step.Resources)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		name := s.name + "-" + step.Name
```
petr-muller: here too

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+		<-ctx.Done()
+		log.Printf("cleanup: Deleting pods with label %s=%s", multiStageTestLabel, s.name)
+		if err := deletePods(s.podClient.Pods(s.jobSpec.Namespace), s.name); err != nil {
+			log.Printf("failed to delete pods with label %s=%s: %v", multiStageTestLabel, s.name, err)
+		}
+	}()
+	var errs []error
+	for _, pod := range pods {
+		log.Printf("Executing %q", pod.Name)
+		if _, err := createOrRestartPod(s.podClient.Pods(s.jobSpec.Namespace), &pod); err != nil {
```
petr-muller: I think there is a very small opportunity for false pod name sharing. If I'm not mistaken, pod name will be `$flowname-$stepname` (https://github.com/openshift/ci-tools/pull/100/files#diff-1c5508a2984039a0cdbdce9c3650e0a7R132), so if someone is unlucky enough to have two flows like `flow=flow step=test-step` and `flow=flow-test step=step`, we could falsely attempt to share that Pod.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+
+func (s multiStageTestStep) runSteps(ctx context.Context, steps []api.TestStep, shortCircuit bool) error {
+	pods, err := s.genPods(steps)
+	if err != nil {
+		return err
+	}
+	return s.runPods(ctx, pods, shortCircuit)
+}
+
+func (s multiStageTestStep) genPods(steps []api.TestStep) ([]coreapi.Pod, error) {
```
droslean: No need to cut the name here. let it be `generatePods`

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+	var errs []error
+	for _, pod := range pods {
+		log.Printf("Executing %q", pod.Name)
+		if _, err := createOrRestartPod(s.podClient.Pods(s.jobSpec.Namespace), &pod); err != nil {
+			errs = append(errs, fmt.Errorf("failed to create or restart %q pod: %v", pod.Name, err))
+			if shortCircuit {
+				break
+			}
+		}
+		if err := waitForPodCompletion(s.podClient.Pods(s.jobSpec.Namespace), pod.Name, nil, false); err != nil {
```
droslean: Do we need to pass a `notifier` here?

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+			errs = append(errs, fmt.Errorf("%q pod %q failed: %v", s.name, pod.Name, err))
+			if shortCircuit {
+				break
+			}
+		}
+	}
+	if errs == nil {
+		return nil
```
droslean: no need for that

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+	subTests        []*junit.TestCase
+}
+
+func MultiStageTestStep(
+	testConfig api.TestStepConfiguration,
+	config *api.ReleaseBuildConfiguration,
+	podClient PodClient,
+	jobSpec *api.JobSpec,
+) (api.Step, error) {
+	return newMultiStageTestStep(testConfig, config, podClient, jobSpec), nil
```
droslean: why we need an extra unexported function and pass again the same args there, to return a new struct?

Also, this doesn't return any error.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+type multiStageTestStep struct {
+	name            string
+	config          *api.ReleaseBuildConfiguration
+	podClient       PodClient
+	jobSpec         *api.JobSpec
+	pre, test, post []api.TestStep
+	subTests        []*junit.TestCase
+}
+
+func MultiStageTestStep(
```
droslean: I would prefer to name this `NewMultiStageTestStep`

pkg/steps/multi_stage.go
droslean: Yep, we will need to parse all three stages.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,202 @@
...
+			errs = append(errs, err)
+			continue
+		}
+		name := s.name + "-" + step.Name
+		pod, err := generateBasePod(s.jobSpec, name, step.Name, []string{"/bin/bash", "-c", "#!/bin/bash\nset -eu\n" + step.Commands}, image, resources)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		pod.Labels[multiStageTestLabel] = s.name
```
droslean: since you are doing this, can you add also the [org/repo/branch](https://github.com/openshift/ci-tools/blob/master/pkg/steps/template.go#L35-L37) labels?

pkg/defaults/defaults.go
bbguimaraes: No longer relevant because of https://github.com/openshift/ci-tools/pull/100#discussion_r318066351.

pkg/steps/multi_stage.go
bbguimaraes: No.

pkg/steps/multi_stage.go
bbguimaraes: Yes. Is there a problem?

pkg/steps/multi_stage.go
bbguimaraes: Didn't know that.

pkg/steps/multi_stage.go
bbguimaraes: Similar situations can occur today with container tests — and probably other areas where pods and other resources are created.

pkg/steps/multi_stage.go
bbguimaraes: Yes.

pkg/steps/multi_stage.go
bbguimaraes: Error removed, the extra function is used by the tests.

pkg/steps/multi_stage.go
bbguimaraes: It follows the convention used by all other steps.

pkg/steps/multi_stage.go
bbguimaraes: That should happen automatically after https://github.com/openshift/ci-tools/pull/92 merges.

pkg/api/types_test.go
```diff
@@ -134,3 +134,61 @@ func TestOverlay(t *testing.T) {
...
+		name: "in `images`",
+		conf: ReleaseBuildConfiguration{
+			Images: []ProjectDirectoryImageBuildStepConfiguration{
+				{To: "image"},
+			},
+		},
+		want: true,
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			if ret := tc.conf.BuildsImage("image"); ret != tc.want {
```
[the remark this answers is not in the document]
bbguimaraes: Not sure what you mean, do you want `image: "image"` repeated in all cases?

pkg/steps/multi_stage.go
bbguimaraes: It's more complex here, but I think that's doable.

pkg/steps/multi_stage.go
bbguimaraes: Well, that's what I asked.

Using a value receiver was based on the (reasonable but not optimal) assumption that it is what methods that don't mutate the object should use. After reading some golang [nonsense](https://github.com/golang/go/wiki/CodeReviewComments#receiver-type), I suppose the size of the object is the contention here? Changed.

pkg/api/types_test.go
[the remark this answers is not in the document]
bbguimaraes: I think it's easier to read the test code here (a one-line conditional), but… changed.

pkg/api/config.go
```diff
@@ -103,33 +103,39 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
 	validationErrors = append(validationErrors, searchForTestDuplicates(input)...)
 
 	for num, test := range input {
```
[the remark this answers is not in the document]
bbguimaraes: We don't, I'll create a separate PR then update this one.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,199 @@
...
+			}
+		}
+		if err := waitForPodCompletion(s.podClient.Pods(s.jobSpec.Namespace), pod.Name, nil, false); err != nil {
+			errs = append(errs, fmt.Errorf("%q pod %q failed: %v", s.name, pod.Name, err))
+			if shortCircuit {
+				break
+			}
+		}
+	}
+	if errs == nil {
```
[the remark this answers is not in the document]
bbguimaraes: Nice.

pkg/steps/multi_stage.go
```diff
@@ -0,0 +1,199 @@
...
+func deletePods(client coreclientset.PodInterface, test string) error {
+	err := client.DeleteCollection(
+		&meta.DeleteOptions{},
+		meta.ListOptions{
+			LabelSelector: fields.Set{
+				multiStageTestLabel: test,
+			}.AsSelector().String(),
+		},
+	)
+	if errors.IsNotFound(err) {
```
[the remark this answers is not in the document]
bbguimaraes: There are three cases:

- no error → `nil (== err)`
- not found → `nil`
- other errors → `err`

Are you proposing:
```go
if err != nil && errors.IsNotFound(err) {
    return err
}
return nil
```
?

pkg/steps/multi_stage_test.go
```diff
@@ -0,0 +1,235 @@
...
+		config: api.ReleaseBuildConfiguration{
+			Images: []api.ProjectDirectoryImageBuildStepConfiguration{
+				{To: "from-images"},
+			},
+		},
+		steps: []api.TestStep{{From: "from-images"}},
+		req:   []api.StepLink{api.ImagesReadyLink()},
+	}, {
+		name:  "needs pipeline image",
+		steps: []api.TestStep{{From: "src"}},
```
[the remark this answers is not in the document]
bbguimaraes: We have all cases covered in `IsPipelineImage`'s test and one of them here, do you still think that's necessary?

pkg/steps/multi_stage_test.go
```diff
@@ -0,0 +1,235 @@
...
+		req: []api.StepLink{
+			api.InternalImageLink(
+				api.PipelineImageStreamTagReference(
+					api.PipelineImageStreamTagReferenceSource)),
+		},
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			step := multiStageTestStep{config: &tc.config, test: tc.steps}
+			ret := step.Requires()
+			if len(ret) == len(tc.req) {
```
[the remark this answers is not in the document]
bbguimaraes: The comparison between steps must be `Step.Matches`, not simple equality.

pkg/steps/multi_stage_test.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	step := newMultiStageTestStep(config.Tests[0], &config, nil, &jobSpec)
+	ret, err := step.generatePods(config.Tests[0].MultiStageTestConfiguration.Test)
+	if err != nil {
+		t.Fatal(err)
+	}
+	expected := []coreapi.Pod{{
+		ObjectMeta: meta.ObjectMeta{
+			Name: "test-step0",
+			Labels: map[string]string{
```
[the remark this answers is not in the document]
bbguimaraes: This was based on master before #92 merged.

pkg/steps/multi_stage_test.go
```diff
@@ -0,0 +1,235 @@
...
+			"test-test0", "test-test1",
+			"test-post0", "test-post1",
+		},
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			fakecs := fake.NewSimpleClientset()
+			var pods []*coreapi.Pod
+			fakecs.PrependReactor("create", "pods", func(action clientgo_testing.Action) (bool, runtime.Object, error) {
+				pod := action.(clientgo_testing.CreateAction).GetObject().(*coreapi.Pod)
+				for _, f := range tc.failures {
```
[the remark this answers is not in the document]
bbguimaraes: You'll hear from my orthopedist.

bbguimaraes:
Rebased to include #92, but only 185ce214^..1e7f9a45 are new commits.

pkg/steps/multi_stage.go
droslean: If we don't mutate the object, we don't need to pass the pointer.

pkg/steps/multi_stage_test.go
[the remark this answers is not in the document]
bbguimaraes: I should have been more explicit in my comment, I corrected that when I rebased.

bbguimaraes:
I think all reviews have been addressed. Rebased, squashed.

bbguimaraes:
Follow-up work identified in the review (other than those listed in the original description):
- add tests for the validation code changed in this PR
- log pods as JSON in dry-run mode

bbguimaraes:
Ah, it took me a while to understand what the notifier question was about. The final version of the code, with artifact handling, does pass an artifact worker as a notifier — that's what I was thinking about when I read and responded to the question.

It is not necessary here because the lifetime of the pod is exactly how much we want to wait. There is even a higher-level [function](https://github.com/openshift/ci-tools/blob/af8a90a2faf965eeda949dc1c607c48d3ffcda3e/pkg/steps/pod.go#L294-L304) we could use that does exactly that, but it would add noise to the subsequent PRs.

stevekuznetsov:
ACK

/hold cancel

PR #102, opened 2019-08-28
pkg/registry: add support for test references

AlexNPavel:
Adds support for test references via a step map that is passed to the registry in the `NewResolver` function.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -282,6 +282,9 @@ func validateTestConfigurationType(fieldRoot string, test TestStepConfiguration,
 		if testConfig.ClusterProfile != "" {
 			validationErrors = append(validationErrors, validateClusterProfile(fmt.Sprintf("%s", fieldRoot), testConfig.ClusterProfile)...)
 		}
+		validationErrors = append(validationErrors, validateTestSteps(fmt.Sprintf("%s.Pre", fieldRoot), testConfig.Pre)...)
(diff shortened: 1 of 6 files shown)


pkg/api/config.go
```diff
@@ -293,24 +296,44 @@ func validateTestConfigurationType(fieldRoot string, test TestStepConfiguration,
...
+			continue
 		}
-		if len(s.From) == 0 {
-			ret = append(ret, fmt.Errorf("%s: `from` is required", fieldRootI))
+		if s.TestStep == nil && s.Reference == nil {
+			ret = append(ret, fmt.Errorf("%s: field cannot be empty", fieldRootI))
+			continue
+		}
+		if s.Reference != nil {
+			// can this be true if s.Reference != nil?
```
bbguimaraes: Yes: https://play.golang.org/p/KRv7KyyRBsZ.

pkg/registry/resolver.go
```diff
@@ -44,12 +62,23 @@ func (r *registry) Resolve(config api.MultiStageTestConfiguration) (types.TestFl
...
+
+func toInternal(input api.TestStep) (types.TestStep, error) {
 	return types.TestStep{
 		As:          input.As,
 		From:        input.From,
 		Commands:    input.Commands,
 		ArtifactDir: input.ArtifactDir,
 		Resources:   input.Resources,
-	}
+	}, nil
```
bbguimaraes: Is the `error` necessary?

PR #103, opened 2019-08-30
Add ci-operator integration tests

droslean:
Adds integration tests for `ci-operator`.


test/ci-operator-integration/expected_files/expected_with_template.json
```diff
@@ -0,0 +1,1533 @@
+{
```
stevekuznetsov: I'm surprised we are not seeing api group version and kind in these, so that makes me think we're using the wrong serializer

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,40 @@
...
+trap 'rm -rf "${WORKDIR}"' EXIT
+
+readonly TEST_ROOT="$( dirname "${BASH_SOURCE[0]}")"
+readonly EXPECTED="${TEST_ROOT}/expected_files/expected.json"
+readonly EXPECTED_WITH_TEMPLATE="${TEST_ROOT}/expected_files/expected_with_template.json"
+readonly DRY_RUN_JSON="${WORKDIR}/ci-op-dry.json"
+readonly DRY_RUN_WITH_TEMPLATE_JSON="${WORKDIR}/ci-op-template-dry.json"
+
+echo "[INFO] Running ci-operator in dry-mode..."
+if ! ci-operator --dry-run --namespace testns --config ${TEST_ROOT}/config/test-config.yaml  --git-ref=openshift/origin@master 2>&1 |  grep -v "^[0-9]" > ${DRY_RUN_JSON}; then
```
stevekuznetsov: I think I'd prefer us to use a `$JOB_SPEC` -- I don't really care how well we support a `--git-ref`

stevekuznetsov: We should not `grep -v` here but instead use `diff -I`

test/ci-operator-integration/compare_json.py
```diff
@@ -0,0 +1,44 @@
...
+import re
+from pprint import pprint
+
+def parse_args():
+	parser = argparse.ArgumentParser(description='Blah blah')
+	parser.add_argument('--files-to-compare',nargs='+', required=True, help='Files to compare')
+	args = parser.parse_args()
+	return args
+
+# Replace all sha256 hashes
```
stevekuznetsov: Also should be `diff -I` on specific fields we don't need to match

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,40 @@
...
+readonly DRY_RUN_WITH_TEMPLATE_JSON="${WORKDIR}/ci-op-template-dry.json"
+
+echo "[INFO] Running ci-operator in dry-mode..."
+if ! ci-operator --dry-run --namespace testns --config ${TEST_ROOT}/config/test-config.yaml  --git-ref=openshift/origin@master > ${DRY_RUN_JSON} 2>&1; then
+  echo "ERROR: ci-operator failed:"
+  exit 1
+fi
+
+
+if ! diff <(sort ${EXPECTED} | grep -vE "^[0-9]{4}/[0-9]{2}/[0-9]{2}|sha256:.*") <(sort ${DRY_RUN_JSON} | grep -vE "^[0-9]{4}/[0-9]{2}/[0-9]{2}|sha256:"); then
```
stevekuznetsov: So let's use `diff -I` instead of `grep -v` -- we can specify regexes to ignore in `diff` with `-I` and it would be good to do that

stevekuznetsov: What is `sort` doing here? sorting the entire JSON file?

droslean: It seems so but the diff output is helpful enough to find the difference

droslean:
Should rebase after #104 lands.

test/ci-operator-integration/run.sh
droslean: addressed

droslean:
/hold

blocked by #109

stevekuznetsov:
We need to use the external codec to serialize with group/version/kind

droslean:
> We need to use the external codec to serialize with group/version/kind

Now we are not using any codec to serialize any of those objects. The group/version/kind in some object is missing because of the way they are created. I tried also to decode the objects with the serializer but the g,b,k are still missing. 
I will try to find out a good reason why this is happening.

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+function sort_nested_JSON_to_file {
+   jq --argfile a $1 -n 'def post_recurse(f): def r: (f | select(. != null) | r), .; r; def post_recurse: post_recurse(.[]?); ($a | (post_recurse | arrays) |= sort) as $a | $a' > $2
```
bbguimaraes: Other than the complexity of this, we don't want to reorder all fields. Some of them (e.g. the `command` field in a container) need to be in a particular order.

We should identify which specific objects needs to be ordered.

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,61 @@
...
+
+echo "[INFO] Running ci-operator in dry-mode..."
+if ! ci-operator --dry-run --namespace ${TEST_NAMESPACE} --config ${TEST_CONFIG} > ${DRY_RUN_JSON} 2>/dev/null; then
+  echo "ERROR: ci-operator failed:"
+  exit 1
+fi
+
+
+sort_nested_JSON_to_file ${DRY_RUN_JSON} ${DRY_RUN_JSON_SORTED}
+if ! diff -b -I "[0-9a-f]\{40\}" -I "[0-9a-f]\{64\}" ${EXPECTED} ${DRY_RUN_JSON_SORTED}; then
```
bbguimaraes: Why do we need `--ignore-space-change`?

test/ci-operator-integration/run.sh
bbguimaraes: Ordering can also be implemented in the logger, in go, if it gets anywhere near this level of complexity.

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,61 @@
...
+	echo "ERROR: differences has been found"
```
bbguimaraes: s/has/have/

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,61 @@
...
+  exit 1
+fi
+
+sort_nested_JSON_to_file ${DRY_RUN_WITH_TEMPLATE_JSON} ${DRY_RUN_WITH_TEMPLATE_JSON_SORTED}
+if ! diff -b -I "[0-9a-f]\{40\}" -I "[0-9a-f]\{64\}" ${EXPECTED_WITH_TEMPLATE} ${DRY_RUN_WITH_TEMPLATE_JSON_SORTED}; then
+	echo "ERROR: differences has been found"
+	exit 1
+fi
+
+echo "[INFO] Success"
```
bbguimaraes: Lines are indented in a few different ways in this file.

test/ci-operator-integration/config/test-config.yaml
```diff
@@ -0,0 +1,246 @@
...
+  verify:
+    limits:
+      memory: 12Gi
+    requests:
+      cpu: "3"
+      memory: 8Gi
+rpm_build_commands: make build-rpms
+tag_specification:
+  name: "ci-operator-test"
+  namespace: ci
```
bbguimaraes: Is this created by our automation tools?

test/ci-operator-integration/run.sh
bbguimaraes: For the `{40}` expression, don't we control the git hashes?

bbguimaraes: Can we make the second expression `@sha256:[0-9a-f]{64}` to be more explicit and reduce the chance of false negatives?

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
petr-muller: Please quote everything to avoid ugly errors and also so we can use shellcheck to validate these files:

```
$ shellcheck test/ci-operator-integration/run.sh

In test/ci-operator-integration/run.sh line 4:
   jq --argfile a $1 -n 'def post_recurse(f): def r: (f | select(. != null) | r), .; r; def post_recurse: post_recurse(.[]?); ($a | (post_recurse | arrays) |= sort) as $a | $a' > $2
                  ^-- SC2086: Double quote to prevent globbing and word splitting.
                                                                                                                                                                                   ^-- SC2086: Double quote to prevent globbing and word splitting.


In test/ci-operator-integration/run.sh line 31:
if ! ci-operator --dry-run --namespace ${TEST_NAMESPACE} --config ${TEST_CONFIG} > ${DRY_RUN_JSON} 2>/dev/null; then
                                                                  ^------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                                   ^-------------^ SC2086: Double quote to prevent globbing and word splitting.


In test/ci-operator-integration/run.sh line 37:
sort_nested_JSON_to_file ${DRY_RUN_JSON} ${DRY_RUN_JSON_SORTED}
                         ^-------------^ SC2086: Double quote to prevent globbing and word splitting.
                                         ^--------------------^ SC2086: Double quote to prevent globbing and word splitting.


In test/ci-operator-integration/run.sh line 38:
if ! diff -b -I "[0-9a-f]\{40\}" -I "[0-9a-f]\{64\}" ${EXPECTED} ${DRY_RUN_JSON_SORTED}; then
                                                     ^---------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                 ^--------------------^ SC2086: Double quote to prevent globbing and word splitting.


In test/ci-operator-integration/run.sh line 50:
if ! ci-operator --dry-run --namespace ${TEST_NAMESPACE} --config ${TEST_CONFIG} --template ${TEST_TEMPLATE} --target test-template --artifact-dir ${ARTIFACT_DIR} > ${DRY_RUN_WITH_TEMPLATE_JSON} 2>/dev/null; then
                                                                  ^------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                                            ^--------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                                                                                                   ^-------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                                                                                                                     ^---------------------------^ SC2086: Double quote to prevent globbing and word splitting.


In test/ci-operator-integration/run.sh line 55:
sort_nested_JSON_to_file ${DRY_RUN_WITH_TEMPLATE_JSON} ${DRY_RUN_WITH_TEMPLATE_JSON_SORTED}
                         ^---------------------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                       ^-- SC2086: Double quote to prevent globbing and word splitting.


In test/ci-operator-integration/run.sh line 56:
if ! diff -b -I "[0-9a-f]\{40\}" -I "[0-9a-f]\{64\}" ${EXPECTED_WITH_TEMPLATE} ${DRY_RUN_WITH_TEMPLATE_JSON_SORTED}; then
                                                     ^-----------------------^ SC2086: Double quote to prevent globbing and word splitting.
                                                                               ^-- SC2086: Double quote to prevent globbing and word splitting.

For more information:
```

bbguimaraes: +1

pkg/steps/testing.go
```diff
@@ -37,10 +71,28 @@ func (dl *DryLogger) AddObject(o runtime.Object) {
 	dl.objects = append(dl.objects, o)
 }
 
-// GetObjects returns the list of objects.
+// GetObjects returns the list of decoded objects with kind and version included.
 func (dl *DryLogger) GetObjects() []runtime.Object {
 	dl.RLock()
 	defer dl.RUnlock()
+
+	for _, object := range dl.objects {
```
stevekuznetsov: Why is this necessary?

test/ci-operator-integration/config/test-config.yaml
droslean: Nope. I added this imagestream because it won't change or so. Any other suggestions are welcome.

test/ci-operator-integration/run.sh
stevekuznetsov: We should order explicitly in the Go code where it's needed

test/ci-operator-integration/run.sh
stevekuznetsov: Yeah right now you are ignoring any 40 and 60-char alpha strings?

pkg/steps/testing.go
droslean: We need to encode/decode each object to include the missing `Kind` `APIVersion` etc fields.

test/ci-operator-integration/config/test-config.yaml
bbguimaraes: Yeah, the test being dependent on external state is not great, but no one has come up with a better idea. I was just commenting that we want that ImageStream tracked in `openshift/release`.

test/ci-operator-integration/run.sh
droslean: The drylogger holds a `[]runtime.Object` that needs to be sorted in some way. I will try to check it.

test/ci-operator-integration/run.sh
droslean: addressed to ignore only `@sha256:[0-9a-f]{64}`

pkg/steps/testing.go
```diff
@@ -37,10 +71,28 @@ func (dl *DryLogger) AddObject(o runtime.Object) {
 	dl.objects = append(dl.objects, o)
 }
 
-// GetObjects returns the list of objects.
+// GetObjects returns the list of decoded objects with kind and version included.
```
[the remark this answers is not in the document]
droslean: I'd prefer to have this operation here or maybe it would be better to unexport this and do this operation in another method.

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,61 @@
...
+readonly DRY_RUN_WITH_TEMPLATE_JSON_SORTED="${WORKDIR}/ci-op-template-dry_sorted.json"
+
+readonly ARTIFACT_DIR="${WORKDIR/artifacts}"
+
+export JOB_SPEC='{"type":"presubmit","job":"pull-ci-openshift-release-master-ci-operator-integration","buildid":"0","prowjobid":"uuid","refs":{"org":"openshift","repo":"ci-tools","base_ref":"master","base_sha":"af8a90a2faf965eeda949dc1c607c48d3ffcda3e","pulls":[{"number":1234,"author":"droslean","sha":"538680dfd2f6cff3b3506c80ca182dcb0dd22a58"}]}}'
+unset BUILD_ID
+
+echo "[INFO] Running ci-operator in dry-mode..."
+if ! ci-operator --dry-run --namespace ${TEST_NAMESPACE} --config ${TEST_CONFIG} > ${DRY_RUN_JSON} 2>/dev/null; then
+  echo "ERROR: ci-operator failed:"
```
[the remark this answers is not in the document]
droslean: I didn't show any message when the ci-operator fails.

test/ci-operator-integration/run.sh
droslean: addressed

test/ci-operator-integration/config/test-config.yaml
droslean: That is a good follow-up. I need also to investigate the case that the imagestream will be empty and so on.

pkg/steps/testing.go
stevekuznetsov: What uses `GetObjects`? I would just expect the logger to encode once on output

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,49 @@
...
+readonly DRY_RUN_WITH_TEMPLATE_JSON="${WORKDIR}/ci-op-template-dry.json"
+
+readonly ARTIFACT_DIR="${WORKDIR/artifacts}"
+
+export JOB_SPEC='{"type":"presubmit","job":"pull-ci-openshift-release-master-ci-operator-integration","buildid":"0","prowjobid":"uuid","refs":{"org":"openshift","repo":"ci-tools","base_ref":"master","base_sha":"af8a90a2faf965eeda949dc1c607c48d3ffcda3e","pulls":[{"number":1234,"author":"droslean","sha":"538680dfd2f6cff3b3506c80ca182dcb0dd22a58"}]}}'
+unset BUILD_ID
+
+echo "[INFO] Running ci-operator in dry-mode..."
+if ! ci-operator --dry-run --namespace "${TEST_NAMESPACE}" --config "${TEST_CONFIG}" > "${DRY_RUN_JSON}" 2>/dev/null; then
+    echo "ERROR: ci-operator failed."
```
stevekuznetsov: You redirect stderr to /dev/null so nobody will know why :)

test/ci-operator-integration/run.sh
```diff
@@ -0,0 +1,49 @@
...
+    exit 1
+fi
+
+if ! diff -b -I "[0-9a-f]\{40\}" -I "[0-9a-f]\{64\}" "${EXPECTED}" "${DRY_RUN_JSON}"; then
```
stevekuznetsov: What is the output of this `diff` without the `-I`?

test/ci-operator-integration/run.sh
droslean: addressed.

pkg/steps/testing.go
droslean: I did the encoding/decoding here to iterate in the objects just once.

droslean: therefore, in the `main()` we can just call the `.GetObjects()` and don't care about the encode/decode and the order of the objects.

test/ci-operator-integration/run.sh
stevekuznetsov: I would rather us use `-I` to skip specific YAML keys instead of values.

test/ci-operator-integration/config/test-config.yaml
droslean: So we have to maintain this imagetream. However, even if we add this `is` in release repo, it will never be going to be automated. Not to mention the image that we have to push there...

bbguimaraes: Do we require *valid* values for a dry-run? Can't we just use

```
$ oc create imagestream --dry-run -o yaml test
$ oc create imagestreamtag --dry-run --from test -o yaml test:test
```

?

droslean: We do actually required valid values. `ci-operator` even in a `dry-run` mode will actually get the images from the imagestream.

droslean:
@stevekuznetsov @petr-muller @bbguimaraes 
There are some design decisions that we can make here. 

- We can avoid the encode/decode/sort logic by only encode and print the objects and let the bash do the sorting and let `diff` to compare lines with json objects. 
- We can somehow maintain imagestreams to be used in `base_images` `tag_specification`, but that will require to maintain a whole new project in the `release` repo, plus scripts etc...
- Be more selective in ignoring the `sha256` hash in specific fields, by using a crazy `jq` command.

Do you think that we can address any of the above statements in the current PR or in an follow-up?

test/ci-operator-integration/config/test-config.yaml
bbguimaraes: But that's just text read from the kube object, isn't it? There is no requirement that the pull spec actually refer to real images.

pkg/steps/testing.go
bbguimaraes: I agree with :steve: here. As I said elsewhere about `GetObjects`, the API is a bit strange for a type called `DryLogger`.

I would expect something to the effect of

```go
func (dl *DryLogger) Log(w io.Writer) error {
 	dl.RLock()
    objs := dl.GetObjects()
    dl.RUnlock() 
    // encode/decode/order
    // log
}
```

bbguimaraes:
A list of the fields that require special treatment would be helpful to guide that decision. Filtering them explicitly (either in `ci-operator` or in the test script) documents that in code, which is why I proposed it.

pkg/steps/testing.go
droslean: addressed

cmd/ci-operator/main.go
```diff
@@ -488,8 +488,9 @@ func (o *options) Run() error {
 		}
 
 		if o.dry {
-			objects, _ := json.MarshalIndent(dryLogger.GetObjects(), "", "  ")
-			fmt.Printf("%s\n", objects)
+			if err := dryLogger.Log(); err != nil {
+				fmt.Printf("%v", err)
```
stevekuznetsov: nit: this will lose a newline, use

```go
fmt.Println(err)
```

pkg/steps/testing.go
```diff
@@ -37,13 +76,97 @@ func (dl *DryLogger) AddObject(o runtime.Object) {
 	dl.objects = append(dl.objects, o)
 }
 
-// GetObjects returns the list of objects.
+func (dl *DryLogger) AddBuild(build *buildapi.Build) {
```
stevekuznetsov: nit: you used to sort the deep copy, now you sort the object someone hands you -- do the deep copy first

pkg/steps/testing.go
```diff
@@ -37,13 +76,97 @@ func (dl *DryLogger) AddObject(o runtime.Object) {
...
+		if tag.From != nil {
+			ist.Tag.From.Name = trimSHA256(ist.Tag.From.Name)
+		}
+	}
+	dl.AddObject(ist.DeepCopyObject())
+}
+
+func trimSHA256(value string) string {
+	reg := regexp.MustCompile("@sha256:[0-9a-f]{64}")
+	return reg.ReplaceAllString(value, "")
```
stevekuznetsov: nit: to make it clearer to someone reading, we could replace with `@sha256:SHA` or something

pkg/steps/testing.go
```diff
@@ -37,13 +76,97 @@ func (dl *DryLogger) AddObject(o runtime.Object) {
...
 func (dl *DryLogger) GetObjects() []runtime.Object {
 	dl.RLock()
 	defer dl.RUnlock()
 	return dl.objects
 }
 
+// Log encode/decode the objects and print them as a valid JSON array.
+func (dl *DryLogger) Log() error {
+	var errors []error
+	objects := dl.GetObjects()
```
stevekuznetsov: do we need `GetObjects()`? why not just access the field

pkg/steps/testing.go
```diff
@@ -37,13 +76,97 @@ func (dl *DryLogger) AddObject(o runtime.Object) {
...
+		buffer := bytes.NewBuffer(b)
+
+		err := encoder.Encode(object, buffer)
+		if err != nil {
+			fmt.Printf("Unexpected encode error '%v'\n", err)
+			errors = append(errors, err)
+			continue
+		}
+
+		_, _, err = decoder.Decode(buffer.Bytes(), nil, object)
```
stevekuznetsov: What does the `meta.Accessor()` return without encode/decode?

droslean: If we don't encode/decode the `Kind` and `APIVersion` will be missing.

pkg/steps/testing.go
droslean: `DeepCopyObject()` is converting the build to a `runtime.Object`

pkg/steps/testing.go
droslean: addressed

pkg/steps/testing.go
droslean: addressed

cmd/ci-operator/main.go
droslean: addressed

pkg/steps/testing.go
bbguimaraes: We can still cast it back safely, it's very surprising for a logger to mutate the objects.

pkg/steps/testing.go
```diff
@@ -44,6 +83,90 @@ func (dl *DryLogger) GetObjects() []runtime.Object {
...
+		}
+		jAccessor, err := meta.Accessor(objects[j])
+		if err != nil {
+			errors = append(errors, fmt.Errorf("couldn't create accessor: %v", err))
+		}
+
+		iKind := objects[i].GetObjectKind().GroupVersionKind().Kind
+		jKind := objects[j].GetObjectKind().GroupVersionKind().Kind
+
+		return iKind+iAccessor.GetName() < jKind+jAccessor.GetName()
```
bbguimaraes: This looks strange, why not do two comparisons?

pkg/steps/testing.go
bbguimaraes: Yeah, sorry, that's what I meant to write in my comment.

pkg/steps/testing.go
droslean: There are objects with the same name. By adding `kind` we make sure that the ordering will be more stable.

bbguimaraes: By "two comparisons", I meant first compare the kinds, and only if they are the same compare the names.

bbguimaraes:
Judging by the silence, I assume people don't care about the untracked test objects. Two minor comments are still open, but I'm fine with merging as is.

/lgtm
/hold

Feel free to merge when desired.

PR #105, opened 2019-08-30
ci-operator multistage: implement chains

AlexNPavel:
Add chains to the multistage registry. The registry first unrolls chains to form an unrolled `[]api.TestStep` and then continues to process as before.

### DIFF
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -16,12 +16,14 @@ type Resolver interface {
 // user provided configs referencing the registry and the internal, complete
 // representation
 type registry struct {
-	stepsByName map[string]api.LiteralTestStep
+	stepsByName  map[string]api.LiteralTestStep
+	chainsByName map[string][]api.TestStep
 }
 
-func NewResolver(stepsByName map[string]api.LiteralTestStep) Resolver {
+func NewResolver(stepsByName map[string]api.LiteralTestStep, chainsByName map[string][]api.TestStep) Resolver {
 	return &registry{
-		stepsByName: stepsByName,
+		stepsByName:  stepsByName,
+		chainsByName: chainsByName,
 	}
 }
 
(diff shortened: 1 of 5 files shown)


pkg/registry/resolver.go
```diff
@@ -49,7 +51,22 @@ func (r *registry) Resolve(config api.MultiStageTestConfiguration) (types.TestFl
 }
 
 func (r *registry) process(steps []api.TestStep) (internalSteps []types.TestStep, errs []error) {
+	// unroll chains
+	var unrolledSteps []api.TestStep
 	for _, external := range steps {
+		if external.Chain != nil {
+			var err error
+			steps, err := r.unrollChain(external)
```
stevekuznetsov: nit: this `steps` shadows the `steps` from the function parameter, let's use a different name for clarity

pkg/registry/resolver_test.go
```diff
@@ -137,9 +140,122 @@ func TestResolve(t *testing.T) {
...
+					Resources: api.ResourceRequirements{
+						Requests: api.ResourceList{"cpu": "1000m"},
+						Limits:   api.ResourceList{"memory": "2Gi"},
+					}},
+			}},
+			Post: []api.TestStep{{
+				Reference: &teardownRef,
+			}},
+		},
+		chainMap: map[string][]api.TestStep{
```
stevekuznetsov: We need another test case for nested chains

pkg/registry/resolver.go
```diff
@@ -66,6 +75,29 @@ func (r *registry) process(steps []api.TestStep) (internalSteps []types.TestStep
...
+	}
+	return
+}
+
+func (r *registry) unrollChains(input []api.TestStep) (unrolledSteps []api.TestStep, errs []error) {
+	for _, step := range input {
+		if step.Chain != nil {
+			chain, ok := r.chainsByName[*step.Chain]
+			if !ok {
+				return []api.TestStep{}, []error{fmt.Errorf("invalid step chain: %s", *step.Chain)}
```
stevekuznetsov: Maybe "unknown" instead of "invalid"?

PR #109, opened 2019-09-02
Introduce a dry run logger to ci-operator

droslean:
DryLogger holds the generated object from each step. This can be used to JSON Marshal only once the whole array and generate a valid JSON output to be used for tests etc...

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -27,6 +27,7 @@ import (
 	rbacapi "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
 	"k8s.io/client-go/rest"
@@ -392,8 +393,9 @@ func (o *options) Run() error {
 		log.Printf("Ran for %s", time.Now().Sub(start).Truncate(time.Second))
 	}()
 
+	dryLogger := &steps.DryLogger{}
 	// load the graph from the configuration
-	buildSteps, postSteps, err := defaults.FromConfig(o.configSpec, o.jobSpec, o.templates, o.writeParams, o.artifactDir, o.promote, o.clusterConfig, o.targets.values)
+	buildSteps, postSteps, err := defaults.FromConfig(o.configSpec, o.jobSpec, o.templates, o.writeParams, o.artifactDir, o.promote, o.clusterConfig, o.targets.values, dryLogger)
 	if err != nil {
 		return fmt.Errorf("failed to generate steps from config: %v", err)
 	}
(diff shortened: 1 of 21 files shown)


pkg/steps/testing.go
```diff
@@ -7,19 +7,37 @@ import (
...
 )
 
+// DryLogger holds the information of all objects that have been created from a dry run.
+type DryLogger struct {
+	objects []runtime.Object
+}
+
+// AddObject is adding an object to the list.
+func (dl *DryLogger) AddObject(o runtime.Object) {
+	dl.objects = append(dl.objects, o)
```
bbguimaraes: Each step is executed in a separate goroutine, will this be safe?

droslean: oh boy.. of course. I will add a lock here.

droslean:
> The structure looks good, one concurrency concern. Can we add a test?

The only test that we can add here is to check if the drylogger `AddObject` and `GetObjects`. I don't think that we need them.

pkg/steps/testing.go
bbguimaraes: Shouldn't it be locked for writing?

https://golang.org/pkg/sync/#RWMutex
> The lock can be held by an arbitrary number of readers or a single writer.

bbguimaraes: Also, for completeness even though that doesn't matter here, we should lock in `GetObjects` too (for reading).

droslean: addressed

bbguimaraes:
/lgtm

/hold
if people want to take a final look.

pkg/steps/testing.go
```diff
@@ -5,21 +5,45 @@ package steps
...
+func (dl *DryLogger) AddObject(o runtime.Object) {
+	dl.Lock()
+	defer dl.Unlock()
+	dl.objects = append(dl.objects, o)
+}
+
+// GetObjects returns the list of objects.
+func (dl *DryLogger) GetObjects() []runtime.Object {
+	dl.RLock()
+	defer dl.RUnlock()
```
petr-muller: Is this actually safe though? We return just the header for writing but the underlying memory is the same...

droslean: If we assign the objects in `x` with `GetObjects()` and it, the `DryLogger` won't change.
example: https://play.golang.org/p/S-heq951JOy

bbguimaraes: I think Petr's question was about the opposite case, `AddObjects` after a slice has been obtained with `GetObjects`.

As I understand it, slices are a `(pointer, length)` pair, so even if `objects` were modified (and even if it had to be reallocated to grow), the slice would still be (outdated, but) valid. Garbage collection should ensure the memory remains valid as long as the slice also is.

This does not guarantee that the objects in the slice cannot be modified, but I
don't think that is a problem here.

cmd/ci-operator/main.go
```diff
@@ -485,6 +487,11 @@ func (o *options) Run() error {
 			time.Sleep(time.Second)
 		}
 
+		if o.dry {
```
stevekuznetsov: Why don't we store the value of `dry` in the logger and use it once, when we output?

cmd/ci-operator/main.go
```diff
@@ -392,8 +393,9 @@ func (o *options) Run() error {
 		log.Printf("Ran for %s", time.Now().Sub(start).Truncate(time.Second))
 	}()
 
+	dryLogger := &steps.DryLogger{}
```
stevekuznetsov: Couldn't we make this a singleton?

cmd/ci-operator/main.go
droslean: good idea. I will do it in a followup

cmd/ci-operator/main.go
droslean: I will check this. We won't have to pass the drylogger to all the steps too...Thanks.

PR #110, opened 2019-09-03
ci-operator multi-stage: gather artifacts

bbguimaraes:
When given an output artifact directory, for each step that declares an
artifact directory, copy files from the container to
`$artifact_dir/$test_name/$step_name/`.

---

Tests involving `pkg/steps/artifacts.go` are impractical and it would be a lot
of work to invent a separate test infrastructure, so tests are limited to pod
generation. I am open for ideas, though, and to working on that in a separate
PR.

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -163,26 +163,27 @@ func (s *multiStageTestStep) runPods(ctx context.Context, pods []coreapi.Pod, sh
 	var errs []error
 	for _, pod := range pods {
 		log.Printf("Executing %q", pod.Name)
-		if s.dry {
-			if err := dumpObject(&pod); err != nil {
-				errs = append(errs, err)
-			}
-			continue
-		}
-		if _, err := createOrRestartPod(s.podClient.Pods(s.jobSpec.Namespace), &pod); err != nil {
-			errs = append(errs, fmt.Errorf("failed to create or restart %q pod: %v", pod.Name, err))
-			if shortCircuit {
-				break
-			}
-		}
-		if err := waitForPodCompletion(s.podClient.Pods(s.jobSpec.Namespace), pod.Name, nil, false); err != nil {
-			errs = append(errs, fmt.Errorf("%q pod %q failed: %v", s.name, pod.Name, err))
+		if err := s.runPod(ctx, &pod); err != nil {
+			errs = append(errs, err)
 			if shortCircuit {
 				break
 			}
 		}
 	}
 	return utilerrors.NewAggregate(errs)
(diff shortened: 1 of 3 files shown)


pkg/steps/multi_stage.go
```diff
@@ -145,6 +154,13 @@ func (s *multiStageTestStep) generatePods(steps []api.TestStep) ([]coreapi.Pod,
 		if owner := s.jobSpec.Owner(); owner != nil {
 			pod.OwnerReferences = append(pod.OwnerReferences, *owner)
 		}
+		if s.artifactDir != "" && step.ArtifactDir != "" {
```
stevekuznetsov: Why do we need to check both?

pkg/steps/multi_stage_test.go
```diff
@@ -118,7 +124,7 @@ func TestGeneratePods(t *testing.T) {
 		},
 		Namespace: "namespace",
 	}
-	step := newMultiStageTestStep(config.Tests[0], &config, nil, &jobSpec)
+	step := newMultiStageTestStep(config.Tests[0], &config, nil, "artifact_dir", &jobSpec)
```
stevekuznetsov: Where is this value used?

pkg/steps/multi_stage.go
bbguimaraes: `s.artifactDir`: the user has requested artifact gathering via `--artifact-dir`, which is where this value ultimately comes from. There is nowhere to output to if that's not the case.

`step.ArtifactDir`: the step has declared an artifact directory. There is no reason to add the mount if it doesn't.

pkg/steps/multi_stage_test.go
bbguimaraes: Nowhere, but it triggers the conditional in your previous comment.

bbguimaraes:
I'm not sure what you mean by "concatenation behavior", but I've explained why there are not tests for the actual artifact gathering in the description.

pkg/steps/multi_stage.go
bbguimaraes: The pod step (unbeknownst to me when I wrote the multi-stage code) uses the exact same logic: https://github.com/openshift/ci-tools/blob/36ac9f630ea7308baacfba2779564d6f5cf544eb/pkg/steps/pod.go#L131.

bbguimaraes: Although maybe we should add the volume even if `--artifact-dir` is not specified? If a test has `artifact_dir`, I would think it requires that directory to exist.

But… it has worked so far, so /shrug?

PR #115, opened 2019-09-04
Autoowners

hongkailiu:
https://jira.coreos.com/browse/DPTP-519

/cc @stevekuznetsov


hongkailiu:
/hold

dep. on https://github.com/kubernetes/test-infra/pull/14191

cmd/autoowners/main.go
```diff
@@ -0,0 +1,441 @@
...
+const (
+	doNotEdit     = "# DO NOT EDIT; this file is auto-generated using tools/populate-owners.\n"
+	ownersComment = "# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md\n"
+	//ownersAliasesComment = "# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#owners_aliases\n"
+
+	githubOrg  = "openshift"
+	githubRepo = "release"
+)
+
+// owners is copied from k8s.io/test-infra/prow/repoowners's Config
```
stevekuznetsov: Since we have this in vendor can we just use it on this one?

hongkailiu: It was actually added the `yaml` tags.

hongkailiu:
/hold

Need do more manually tests since the http client and yaml write function have been changed.

hongkailiu:
/hold cancel

@stevekuznetsov I will test manually with container.
Ready for review and merge.

cmd/autoowners/main.go
```diff
@@ -3,31 +3,34 @@ package main
...
+	"k8s.io/test-infra/prow/repoowners"
 )
 
 const (
-	doNotEdit            = "# DO NOT EDIT; this file is auto-generated using tools/populate-owners.\n"
-	ownersComment        = "# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md\n"
-	ownersAliasesComment = "# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#owners_aliases\n"
+	doNotEdit     = "# DO NOT EDIT; this file is auto-generated using tools/populate-owners.\n"
+	ownersComment = "# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md\n"
+	//ownersAliasesComment = "# See the OWNERS_ALIASES docs: https://git.k8s.io/community/contributors/guide/owners.md#owners_aliases\n"
```
wking: Why did `OWNERS_ALIASES` handling get removed here?  Does that mean that [openshift/release's `OWNERS_ALIASES`][1] is no longer maintained?  I thought that was useful for maintaining `OWNERS` files in non-job directories (e.g. openshift/release#1791), although I see openshift/release#4279 decided to hard-code lists of people in `ci-operator/config/openshift/installer/OWNERS` (which are now manually maintained by... somebody?).

[1]: https://github.com/openshift/release/blob/18abe54df94f3e6d6155c08af2068a29ff526a9f/OWNERS_ALIASES#L1

hongkailiu: The current autoownners tool does the sync of OWNERS and OWNERS_ALIAS from the upstream repos ONLY to the jobs/config/template folders via a periodic job, e.g., https://prow.svc.ci.openshift.org/prowjob?prowjob=98e2303f-0f6e-11ea-8989-0a58ac10564c

You can see that we have `ignore-repo` args there if some repo does not want us to do this sync. _If the upstream does not have an OWNERS file, the utility will ignore syncing it for those paths._

With that said, top level OWNERS related files are not synced by the tool. So if we want to defined alias, we need to maintain manually. It is one file used by release repo, maybe that is OK?

4729 was probably before this tool was taken to production. And the changed owner file was controlled by the tool now.
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/installer/OWNERS#L1-L16

wking: > So if we want to defined alias, we need to maintain manually.

I expect we're going to have stale entries in these files then, because folks are probably going to forget to bump them (or an `OWNERS_ALIASES` entry) when they adjust the installer approver set, and I still think we want to be using the installer approver set for that template directory.

PR #118, opened 2019-09-05
ci-operator multistage: add loading code for registry

AlexNPavel:
This adds new structs for the registry configs and code to load the configs into the structs that the registry Resolver can use.


pkg/api/types.go
```diff
@@ -341,16 +341,50 @@ type TestStepConfiguration struct {
...
 
+// RegistryReferenceConfig is the struct that step references are unmarshalled into
+type RegistryReferenceConfig struct {
+	Reference RegistryReference `json:"step,omitempty"`
+}
+
+// RegistryReference contains the LiteralTestStep of a reference as well as the documentation for the step
+type RegistryReference struct {
+	LiteralTestStep `json:",inline"`
+	Documentation   string `json:"documentation,omitempty"`
```
stevekuznetsov: In general these are the actual types that we will expect people to be filling in when they make PRs to add new files to the registry, so we will link them the GoDoc page to read when they have questions on how to configure something, so we always want doc on what exported fields are for, too

stevekuznetsov: (and let's make sure our comments here are written for that audience)

pkg/load/load.go
```diff
@@ -32,3 +34,119 @@ func Config(path string) (*api.ReleaseBuildConfiguration, error) {
...
+	if err != nil {
+		return "", api.LiteralTestStep{}, err
+	}
+	baseDir := filepath.Dir(file.Name())
+	command, err := ioutil.ReadFile(filepath.Join(baseDir, step.Reference.Commands))
+	if err != nil {
+		return "", api.LiteralTestStep{}, err
+	}
+	step.Reference.Commands = string(command)
+	fmt.Printf("Ref is %+v\n", step)
```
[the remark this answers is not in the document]
AlexNPavel: Yeah, these were debugging messages that I forgot to remove. I don't think we need this logged in production

cmd/testReg/main.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"flag"
+	"fmt"
+	"log"
+
+	"github.com/openshift/ci-tools/pkg/load"
+	"github.com/openshift/ci-tools/pkg/registry"
+)
+
+func main() {
```
stevekuznetsov: If we want to land this in the current PR we should use the normal workflow (`type options struct` etc) and be a little more rigorous, like adding help text to flags, etc

PR #124, opened 2019-09-06
Refactor artifacts container injection to Template pods.

droslean:
Trying to land this kind of changes, we met a lot of complications. 
Another PR that removes the unnecessary variables in the template has to be merged in order for this PR to work.

The previous attempt was crashing to Decode the pod from the template because of [this](https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-src.yaml#L151) variable.

ref: [release/pull/4950](https://github.com/openshift/release/pull/4950)

### DIFF
--- a/test/ci-operator-integration/expected_files/expected_with_template.json
+++ b/test/ci-operator-integration/expected_files/expected_with_template.json
@@ -1164,6 +1164,7 @@
             "ci-operator.openshift.io/save-container-logs": "true",
             "ci-operator.openshift.io/wait-for-container-artifacts": "teardown"
           },
+          "creationTimestamp": null,
           "name": "${JOB_NAME_SAFE}",
           "namespace": "${NAMESPACE}"
         },
@@ -1247,6 +1248,15 @@
               ],
               "image": "${LOCAL_IMAGE_SRC}",
               "name": "test",
+              "resources": {
+                "limits": {
+                  "memory": "200Mi"
+                },
+                "requests": {
+                  "cpu": "10m",
+                  "memory": "10Mi"
+                }
+              },
               "terminationMessagePolicy": "FallbackToLogsOnError",
               "volumeMounts": [
                 {
@@ -1363,6 +1373,15 @@
               ],
               "image": "${IMAGE_INSTALLER}",
               "name": "setup",
+              "resources": {
+                "limits": {
+                  "memory": "200Mi"
+                },
+                "requests": {
+                  "cpu": "10m",
+                  "memory": "10Mi"
+                }
+              },
               "terminationMessagePolicy": "FallbackToLogsOnError",
               "volumeMounts": [
                 {
@@ -1419,6 +1438,15 @@
               ],
               "image": "${IMAGE_TESTS}",
               "name": "teardown",
+              "resources": {
+                "limits": {
+                  "memory": "200Mi"
+                },
+                "requests": {
+                  "cpu": "10m",
+                  "memory": "10Mi"
+                }
+              },
               "terminationMessagePolicy": "FallbackToLogsOnError",
               "volumeMounts": [
                 {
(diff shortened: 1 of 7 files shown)


test/ci-operator-integration/expected_files/expected_with_template.json
```diff
@@ -1151,21 +1151,57 @@
         ]
       },
       {
-        "apiVersion": "v1",
         "kind": "Pod",
+        "apiVersion": "v1",
```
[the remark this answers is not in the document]
droslean: because before we were just injecting the json stuff there, but with the refactoring, we do the serializing.

droslean:
> Can we run things through jq or something? I want to see the actual diff here, minus reordering

The difference is the ordering plus some extra `resources: {}` and `status: {}` fields

test/ci-operator-integration/expected_files/expected_with_template.json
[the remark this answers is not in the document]
droslean: Also, note that the `json.Marshal` does some kind of keys ordering, but the serializer does not.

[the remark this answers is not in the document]
stevekuznetsov: I'm confused -- aren't we running the Pod through the serializer regardless of what happens to it first?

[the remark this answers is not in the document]
droslean: we do that in this refactoring. Before we were getting the bytes with `json.Unmarshal` [here](https://github.com/openshift/ci-tools/blob/master/pkg/steps/artifacts.go#L571) and the crazy json keys injection we just update the object.Raw with those `[]bytes` [here](https://github.com/openshift/ci-tools/blob/master/pkg/steps/artifacts.go#L600). The integration tests have been created based on that output. It makes sense for the new refactoring to create a new output

test/ci-operator-integration/expected_files/expected_with_template.json
```diff
@@ -1151,21 +1151,57 @@
         ]
       },
       {
-        "apiVersion": "v1",
```
stevekuznetsov: Let's commit a PR first that adds a `jq --sort-keys` step to the test so that we don't have a diff here on ordering.

droslean: I will prepare a PR for it. In the meanwhile this is the diff from the ordered files:
```
1167d1166
<           "creationTimestamp": null,
1251d1249
<               "resources": {},
1368d1365
<               "resources": {},
1425d1421
<               "resources": {},
1468d1463
<               "resources": {},
1495,1496c1490
<         },
<         "status": {}
---
>         }

```

droslean: PR: #153

droslean:
/hold 

Integration tests should have fail. It seems that the jq version 1.5 doesn't have the `--sort-keys`

droslean:
/hold cancel

Integration tests fixed. Also, it is supposed to fail here.

droslean:
@stevekuznetsov Now [commit](https://github.com/openshift/ci-tools/pull/124/commits/e745a445d315c3df837591a18611ea5bd75321f1) makes clear what are the actual differences in the integration tests.

test/ci-operator-integration/expected_files/expected_with_template.json
```diff
@@ -1247,6 +1248,7 @@
               ],
               "image": "${LOCAL_IMAGE_SRC}",
               "name": "test",
+              "resources": {},
```
stevekuznetsov: Can we add config to the test so that we inject some resources? :)

droslean: on it!

droslean: addressed

PR #128, opened 2019-09-10
Address static code linting warnings

stevekuznetsov:
We've got a couple hundred of these things hanging around ... 

@openshift/openshift-team-developer-productivity-test-platform

### DIFF
--- a/pkg/interrupt/interrupt.go
+++ b/pkg/interrupt/interrupt.go
@@ -38,17 +38,6 @@ type Handler struct {
 	once   sync.Once
 }
 
-// Chain creates a new handler that invokes all notify functions when the critical section exits
-// and then invokes the optional handler's notifications. This allows critical sections to be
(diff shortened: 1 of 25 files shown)


pkg/interrupt/interrupt.go
```diff
@@ -42,12 +42,6 @@ type Handler struct {
 // and then invokes the optional handler's notifications. This allows critical sections to be
 // nested without losing exactly once invocations. Notify functions can invoke any cleanup needed
 // but should not exit (which is the responsibility of the parent handler).
-func Chain(handler *Handler, notify ...func()) *Handler {
```
petr-muller: Remove the comment too

stevekuznetsov: Booo, intellij!

stevekuznetsov:
We could add a whitelist to a lint job, I guess. Mitt be hard to track white listed failures that move inside a file

PR #130, opened 2019-09-11
Add the `fail` field in clonerefs options.

droslean:
after [test-infra/pull/14280](https://github.com/kubernetes/test-infra/pull/14280) landed, we can make use of this option.

If we fail to clone any of the repositories, ci-operator shouldn't proceed with the tests.


droslean:
WIP -> fixing the vendors.. Also, it seems that more changes are needed to sync with upstream latest changes.

droslean:
relies on #129 
/hold

droslean:
rebased after #129 

/hold cancel

PR #131, opened 2019-09-11
[WIP] ci-operator-configresolver: initial commit

AlexNPavel:
This PR adds a new command: `ci-operator-configresolver`. It is intended to be run in the cluster and be used by the ci-operator to return a ci-operator config with the `MultiStageTestConfiguration` steps fully expanded via an HTTP GET request.

The command expects a config directory where every branch is a subdirectory and each subdirectory has config files in the form of `$org-$repo-$branch.yaml`. This will allow us to mount the configmaps containing the configs easily when running in a cluster.

It also has a registry directory that it watches for changes. Any time the registry directory (or any of its subdirectories) have a change (detected via fsnotify), the registry is automatically reloaded.


cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/fsnotify/fsnotify"
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/load"
+	"github.com/openshift/ci-tools/pkg/registry"
+	types "github.com/openshift/ci-tools/pkg/steps/types"
+	log "github.com/sirupsen/logrus"
+)
+
+var (
+	resolver   registry.Resolver
```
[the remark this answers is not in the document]
stevekuznetsov: We could use a closure, something like this:

```go
func resolverFunc(resolver registry.Resolver) http.HandlerFunc {
    return func(w http.ResponseWriter, r *http.Request) {
        // do something here with resolver
    }
}

func main() {
    resolver := registry.NewResolver()
    http.HandleFunc("/config", resolverFunc(resolver))
}
```

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,177 @@
...
+			Test:           internalStepsToAPISteps(resolvedFlow.Test),
+			Post:           internalStepsToAPISteps(resolvedFlow.Post),
+		}
+		newStep.MultiStageTestConfiguration = &multiStageConfig
+		result = append(result, newStep)
+	}
+	return result, nil
+}
+
+func internalStepsToAPISteps(steps []types.TestStep) []api.TestStep {
```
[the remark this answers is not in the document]
stevekuznetsov: Yeah, crap. Before, we only converted one sub-structure so we only needed an internal variant of that one thing. Now, we're operating on the whole config so we need an internal structure that holds the resolved config. @bbguimaraes thoughts?

go.mod
```diff
@@ -14,6 +14,7 @@ require (
 	github.com/docker/spdystream v0.0.0-20181023171402-6480d4af844c // indirect
 	github.com/elazarl/goproxy v0.0.0-20190711103511-473e67f1d7d2 // indirect
 	github.com/elazarl/goproxy/ext v0.0.0-20190711103511-473e67f1d7d2 // indirect
+	github.com/fsnotify/fsnotify v1.4.7
```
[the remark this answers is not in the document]
stevekuznetsov: Ah, makes sense.

cmd/ci-operator-configresolver/main.go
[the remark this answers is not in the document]
bbguimaraes: TBH, I never understood the purpose of `pkg/steps/types.go` (and actually have been maintaining a local patch that aliases those back to the `api` types). The [only reference I've found](https://github.com/openshift/release/pull/5019) is:

> This contains some duplicate code from TestStepConfiguration as this is essentially the evolution of that struct, but I did not want to trample that code while this new config format is still in a WIP state.

so it seems we should just merge them back?

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,189 @@
...
+			return
+		}
+		variant := r.URL.Query().Get("variant")
+		info := config.Info{
+			Org:     org,
+			Repo:    repo,
+			Branch:  branch,
+			Variant: variant,
+		}
+		info.Filename = filepath.Join(o.configPath, branch, info.Basename())
```
bbguimaraes: Is this directory structure possible with the ConfigMap structure we have?

https://github.com/openshift/release/blob/725610d085583d0af88eafbe292cbb56897376b9/core-services/prow/02_config/_plugins.yaml#L1198-L1259

bbguimaraes:
How are we going to mount the registry files into resolver pods? ConfigMap mounts are [amusing](https://github.com/openshift/ci-secret-mirroring-controller/pull/9#issuecomment-518222625), so we should make sure whatever directory structure and tracking mechanism we use interact correctly with them.

cmd/ci-operator-configresolver/main.go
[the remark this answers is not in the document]
stevekuznetsov: We have two separate types and two separate things:

 1. a type that users use to configure their tests
 2. a type that ci-operator uses to run the tests

It would not be good to use a type that has possibly-nil fields, aliases, etc post-resolution, since those are syntactic sugar for the user in config.

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,189 @@
...
+
+		config, err := load.Config(info.Filename)
```
stevekuznetsov: I think I was expecting to see a server that has an up-to-date mapping `config.ByFilename`, validate on load (which happens on file changes) and not load or validate on user request

cmd/ci-operator-configresolver/main.go
AlexNPavel: For our configmaps that contain the ci-operator configs, it should work fine. The names of the keys are filenames and when mounted everything _should_ work as expected. I can try this out on a test cluster to make sure though.

cmd/ci-operator-configresolver/main.go
AlexNPavel: Ok, I can do that

cmd/ci-operator-configresolver/main.go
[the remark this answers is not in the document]
bbguimaraes: From a purely abstract point of view, I understand, which is why I hadn't mentioned it before.

The differences, however, are minimal: post-resolution, the `workflow` in each test and `chain` and `reference` in each step must be `nil`. Everything else is exactly the same. I think those are simple and reasonable invariants and that it's not worth it to complicate the implementation with separate types.

cmd/ci-operator-configresolver/main.go
bbguimaraes: I don't think we can create separate directories for each branch. Some of the patterns in the configuration have wildcards for the branch part, or for the entire file name.

AlexNPavel:
@bbguimaraes I tested this in a cluster and reloading seems to work correctly since this code keeps track of all subdirectories including the data subdirectories used in the configmap mounts. One minor thing we might need to look into is the fact that this does trigger a lot of reloads. Using a flattened version of the registry that I have in this PR mounted from a configmap, changing 1 file caused the registry to reload 17 times, which isn't ideal behavior. It may also make sense to keep track of removals to prevent too many watches from being open on files/directories that don't exist anymore.

stevekuznetsov:
> changing 1 file caused the registry to reload 17 times, 

Why? What were the 17 events seen by `fs-notify`?

AlexNPavel:
So, I've been looking at what fsnotify events we're receiving when we do various operations. Here are some logs: 
```
ON CREATE
time="2019-09-16T22:16:37Z" level=info msg="Received CREATE event for /registry/..2019_09_16_22_16_37.172096673"
time="2019-09-16T22:16:37Z" level=info msg="Received CHMOD event for /registry/..2019_09_16_22_16_37.172096673"
time="2019-09-16T22:16:37Z" level=info msg="Received CREATE event for /registry/ipi-deprovision-chain.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received RENAME event for /registry/..data_tmp"
time="2019-09-16T22:16:37Z" level=info msg="Received CREATE event for /registry/..data"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-install-install-commands.sh"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-install-install-ref.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-install-rbac-commands.sh"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-deprovision-deprovision-ref.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-workflow.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-install-chain.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-deprovision-must-gather-commands.sh"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-install-rbac-ref.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-deprovision-deprovision-commands.sh"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800/ipi-deprovision-must-gather-ref.yaml"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800"
time="2019-09-16T22:16:37Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_15_17.471299800"

ON MODIFY
time="2019-09-16T22:18:55Z" level=info msg="Received CREATE event for /registry/..2019_09_16_22_18_55.401007839"
time="2019-09-16T22:18:55Z" level=info msg="Received CHMOD event for /registry/..2019_09_16_22_18_55.401007839"
time="2019-09-16T22:18:55Z" level=info msg="Received RENAME event for /registry/..data_tmp"
time="2019-09-16T22:18:55Z" level=info msg="Received CREATE event for /registry/..data"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-install-install-commands.sh"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-install-install-ref.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-install-rbac-commands.sh"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-deprovision-chain.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-deprovision-deprovision-ref.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-workflow.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-install-chain.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-deprovision-must-gather-commands.sh"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-install-rbac-ref.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-deprovision-deprovision-commands.sh"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318/ipi-deprovision-must-gather-ref.yaml"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318"
time="2019-09-16T22:18:55Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_17_44.538588318"

ON DELETE
time="2019-09-16T22:25:33Z" level=info msg="Received CREATE event for /registry/..2019_09_16_22_25_33.412163723"
time="2019-09-16T22:25:33Z" level=info msg="Received CHMOD event for /registry/..2019_09_16_22_25_33.412163723"
time="2019-09-16T22:25:33Z" level=info msg="Received RENAME event for /registry/..data_tmp"
time="2019-09-16T22:25:33Z" level=info msg="Received CREATE event for /registry/..data"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/ipi-deprovision-chain.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-install-install-commands.sh"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-install-install-ref.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-install-rbac-commands.sh"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-deprovision-chain.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-deprovision-deprovision-ref.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-workflow.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-install-chain.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-deprovision-must-gather-commands.sh"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-install-rbac-ref.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-deprovision-deprovision-commands.sh"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431/ipi-deprovision-must-gather-ref.yaml"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for /registry/..2019_09_16_22_24_14.614883431"
time="2019-09-16T22:25:33Z" level=info msg="Received REMOVE event for "
```

Would it be reasonable to assume we don't need to reload the registry on REMOVE events (other than removing old watches)? It seems the main issue with this would be that if we delete a ref/chain/workflow from the registry, it would still remain in our registry after the configmap is updated, since that REMOVE happens after all the other non-REMOVE events.

stevekuznetsov:
Oh, because the timestamps are changing -- we should look only at the actual data, not the k8s symlinks and machinery

AlexNPavel:
I don't think we can do that with fsnotify

AlexNPavel:
I have updated the logic for registry reloading with fsnotify (I will use the same logic for config directory reloading when I add that). Now when an event occurs, a new go routine is created that waits 1 second before reloading the registry. All other events that occur do not trigger a registry reload and just update the watches until the registry reloading go routine finishes waiting 1 second (the go routine resets the boolean used to skip a reload before running the reload function in case another event occurs while the registry is in the process of being reloaded)

cmd/ci-operator-configresolver/main.go
```diff
@@ -155,30 +156,49 @@ func main() {
 	opts := options{}
 	flag.StringVar(&opts.regPath, "registry", "", "Path to step registry")
 	flag.StringVar(&opts.configPath, "config", "", "Path to config dirs")
+	verbose := flag.Bool("verbose", false, "Enable debug logging messages")
```
stevekuznetsov: Add the `logrus` level as a flag itself:

```go

	level, err := logrus.ParseLevel(o.logLevel)
	if err != nil {
		return fmt.Errorf("invalid --log-level: %v", err)
	}
	logrus.SetLevel(level)
```

cmd/ci-operator-configresolver/main.go
```diff
@@ -155,30 +156,49 @@ func main() {
...
+					log.Debugf("Removing %s from watches", event.Name)
+					watcher.Remove(event.Name)
+				}
+				if !reloadSet {
+					reloadSet = true
+					// use separate go routine to allow other events to run while the reloader waits
+					go func() {
+						time.Sleep(time.Second)
+						// unset reloadSet before running reload function in case a change occurs during the reload
+						reloadSet = false
```
stevekuznetsov: You're writing and reading a value from separate routines, which will be racy. Let's use the `sync` package for concurrency: (this is rough, will need refinement and testing)

```go
type coalescer struct {
    lock sync.Mutex
    loading bool
    loader sync.Cond
    latest *Registry
}

func (c *coalescer) reload() {
    c.Lock()
    if c.loading {
        // someone else is loading it, so we wait
        c.loader.L.Lock()
        c.Unlock()
        c.loader.Wait()
        c.loader.L.Unlock()
        return c.latest
    }
    // nobody else is loading
    c.loading = true
    c.Unlock()

    // reload the registry, wait an extra second after if you want
    registry := registry.Load()
    c.Lock()
    c.latest = registry
    c.Unlock()

    // inform the waiters that we are done
    c.loader.L.Lock()
    c.loader.Broadcast()
    c.loader.L.Unlock()
}
```

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,251 @@
...
+	}
+	resolver := registry.NewResolver(references, chain, workflows)
+	c.Lock()
+	c.latest = &resolver
+	c.Unlock()
+	log.Info("Registry reloaded")
+
+	// inform the waiters that we are done
+	c.loader.L.Lock()
+	c.loading = false
```
stevekuznetsov: You're setting this without holding `c.Lock()`

cmd/ci-operator-configresolver/main.go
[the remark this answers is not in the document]
stevekuznetsov: I disagree. Implicit rules about which parts of a structure are or are not nil are hard to remember, challenging to learn and grow over time. k8s in general has very explicitly not chosen this path. Prow organically grew to a total chaos of fields that may or may not be set. I don't want to see us go in that direction here.

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,251 @@
...
+}
+
+type coalescer struct {
+	sync.Mutex
+	loading bool
+	loader  *sync.Cond
+	latest  *registry.Resolver
+}
+
+func (c *coalescer) reload(o options) {
```
stevekuznetsov: The options are invariant, let's choose which of them we need (the paths?) and make them part of the struct, not passed in every call

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,251 @@
...
+	c.Lock()
+	if c.loading {
+		// someone else is loading it, so we wait
+		log.Debug("Registry already being reloaded; waiting")
+		c.loader.L.Lock()
+		c.Unlock()
+		c.loader.Wait()
+		c.loader.L.Unlock()
+		log.Debug("Registry reloaded by separate goroutine, returning")
+		return
```
stevekuznetsov: How does this work? I don't see you return anything and I don't see anything use the `latest` value. Whatever is going to be accessing `latest` will also be racy unless it locks, so IMO you should return a reference here since all the locking is already handled.

test/ci-op-configresolver-integration/run.sh
```diff
@@ -0,0 +1,56 @@
...
+# Purpose:  prepends a command to a trap
+#
+# - 1st arg:  code to add
+# - remaining args:  names of traps to modify
+#
+# Example:  trap_add 'echo "in trap DEBUG"' DEBUG
+#
+# See: http://stackoverflow.com/questions/3338030/multiple-bash-traps-for-the-same-signal
+#===================================================================
+function trap_add() {
```
stevekuznetsov: Every time I see Bash like this I wince -- let's not use it. We can implement this more concisely

test/ci-op-configresolver-integration/run.sh
```diff
@@ -0,0 +1,56 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOTDIR=$(pwd)
+WORKDIR="$( mktemp -d )"
+trap_add 'rm -rf "${WORKDIR}"' EXIT
+
+pushd $WORKDIR
+go build -o resolver $ROOTDIR/cmd/ci-operator-configresolver
```
stevekuznetsov: As with the other tests, expect up-to-date binaries in $PATH

test/ci-op-configresolver-integration/run.sh
```diff
@@ -0,0 +1,56 @@
...
+go build -o tester $ROOTDIR/test/ci-op-configresolver-integration/main.go
+# copy registry to tmpdir to allow tester to modify registry
+cp -a $ROOTDIR/test/ci-op-configresolver-integration/ tests
+./resolver -config tests/configs -registry tests/registry &
+PID=$!
+trap_add 'kill "${PID}"' EXIT
```
stevekuznetsov: You can just do this at the end:


```sh
for job in $( jobs -p ); do
    wait "${job}"
done
```

test/ci-op-configresolver-integration/run.sh
```diff
@@ -0,0 +1,56 @@
...
+# wait for registry to be resolved
```
stevekuznetsov: We will want a health and readiness check anyway for the Pod, so we should poll that.

stevekuznetsov: See https://github.com/kubernetes/test-infra/blob/master/prow/pjutil/health.go

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,251 @@
...
+				}
+			case err := <-watcher.Errors:
+				log.Errorf("error: %v", err)
+			}
+		}
+	}()
+	populateWatcher(watcher, opts.regPath)
+
+	http.HandleFunc("/config", resolveConfig(&opts))
+	http.ListenAndServe(":8080", nil)
```
stevekuznetsov: Since we already have it imported, we might as well use the interrupts library for this https://github.com/kubernetes/test-infra/blob/master/prow/interrupts/interrupts.go#L152

test/ci-op-configresolver-integration/main.go
```diff
@@ -0,0 +1,143 @@
...
+								Requests: api.ResourceList{"cpu": "1000m", "mem": "2Gi"},
+							}},
+					},
+				},
+			},
+		},
+	},
+}
+
+func main() {
```
stevekuznetsov: This seems complex. At a black-box integration level, we will have:

 - registry data
 - config data
 - HTTP requests
 - expected HTTP responses

I almost wonder if we can just write a script that runs the server on the test data and loops through the requests and responses to validate them.

cmd/ci-operator-configresolver/main.go
stevekuznetsov: (unit tests help)

stevekuznetsov: Make sure to run `go test -race` to catch data races

test/ci-op-configresolver-integration/run.sh
AlexNPavel: If `./tester` panics then we won't get to the end of the function so we need a trap for that situation

cmd/ci-operator-configresolver/main.go
```diff
@@ -70,7 +70,9 @@ func (c *coalescer) coalesce(o *options) {
 
 	// inform the waiters that we are done
 	c.loader.L.Lock()
+	c.Lock()
```
stevekuznetsov: Now, this can deadlock. I don't think that's what you want :)

test/ci-op-configresolver-integration/run.sh
stevekuznetsov: Why will it panic? We shouldn't code around panics, they are execptional

stevekuznetsov:
This has been implemented in a series of pulls
/close

PR #143, opened 2019-09-16
Don't add extraRefs if the job belongs to the same org/repo with refs.

droslean:
If the rehearse job belongs to the same org/repo with the refs, the `extrarefs` injection is not needed. Otherwise, `clonerefs` will fail.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -11,7 +11,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
@@ -88,7 +88,7 @@ func NewCMClient(clusterConfig *rest.Config, namespace string, dry bool) (corecl
 	return cmClient.ConfigMaps(namespace), nil
 }
 
-func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber int) (*prowconfig.Presubmit, error) {
+func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber int, refs *pjapi.Refs) (*prowconfig.Presubmit, error) {
 	var rehearsal prowconfig.Presubmit
 	deepcopy.Copy(&rehearsal, source)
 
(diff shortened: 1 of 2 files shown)


test/pj-rehearse-integration/expected.yaml
```diff
@@ -512,11 +512,6 @@
         entrypoint: gcr.io/k8s-prow/entrypoint:v20190129-0a3c54c
         initupload: gcr.io/k8s-prow/initupload:v20190129-0a3c54c
         sidecar: gcr.io/k8s-prow/sidecar:v20190129-0a3c54c
-    extra_refs:
-    - base_ref: cluster-profile
-      org: openshift
-      repo: origin
-      workdir: true
```
petr-muller: This looks suspicious... This does not seem like the case where `extra_refs` should be removed (refs are not nil and are openshift/release, extra_refs are openshift/origin)

:confused:

pkg/rehearse/jobs.go
```diff
@@ -101,12 +101,19 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber
...
-			rehearsal.ExtraRefs = append(rehearsal.ExtraRefs, pjapi.Refs{
-				Org:     orgRepo[0],
-				Repo:    orgRepo[1],
-				BaseRef: branch,
-				WorkDir: true,
-			})
+			jobOrg := orgRepo[0]
+			jobRepo := orgRepo[1]
+
+			if refs != nil {
```
petr-muller: I think we need to add `ExtraRefs` when `refs==nil` though, for periodics?

droslean: there is no org/repo to add. In the periodics, the user has to add the `extra_refs` by himself.

test/pj-rehearse-integration/expected.yaml
droslean: No, quite the opposite. This presubmit is under the `openshift/origin` repo in the [job config](https://github.com/openshift/ci-tools/blob/master/test/pj-rehearse-integration/master/ci-operator/jobs/super/duper/super-duper-cluster-profile-presubmits.yaml#L2).

petr-muller: Well, yeah - that's why I would expect it to have openshift/release refs (to report stuff to github) and openshift/origin extra_refs with workdir: true to be able to work with its code.

droslean: OH! yea

pkg/rehearse/jobs.go
petr-muller: Ah, we're relying on `repo==""` for periodics...

petr-muller:
@droslean Would you mind adding one more testcase for https://github.com/openshift/ci-tools/pull/143#discussion_r325172349, e.g. having same org and different repo? It would be useful, we managed to hit that pitfall :)

droslean:
Sure thing.

droslean:
@petr-muller test case added.

PR #147, opened 2019-09-16
ci-operator: make the artifact worker testable

bbguimaraes:
Not for the faint-hearted.  As promised (and since regretted) in
https://github.com/openshift/ci-tools/pull/110, test artifact gathering in
multi-stage tests, making internal APIs more generic where required.

`waitForPodCompletion` and friends are very complex and messy functions, but
changes are as small as possible and tests are added as soon as it's practical.
This also makes it possible to add test to those functions.

Additionally, I've verified multiple executions of container, template, and
multi-stage tests using the new binary.

Each commit is independent.

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -422,7 +422,7 @@ func (w *ArtifactWorker) downloadArtifacts(podName string, hasArtifacts bool) er
 	return nil
 }
 
-func (w *ArtifactWorker) CollectFromPod(podName string, hasArtifactsContainer bool, hasArtifacts []string, waitForContainers []string) {
+func (w *ArtifactWorker) CollectFromPod(podName string, hasArtifacts []string, waitForContainers []string) {
 	w.lock.Lock()
 	defer w.lock.Unlock()
 
@@ -541,11 +541,7 @@ func (w *ArtifactWorker) Done(podName string) bool {
 
 func addArtifactContainersFromPod(pod *coreapi.Pod, worker *ArtifactWorker) {
 	var containers []string
-	var hasArtifactsContainer bool
 	for _, container := range append(append([]coreapi.Container{}, pod.Spec.InitContainers...), pod.Spec.Containers...) {
-		if container.Name == "artifacts" {
-			hasArtifactsContainer = true
-		}
 		if !containerHasVolumeName(container, "artifacts") {
 			continue
 		}
@@ -555,7 +551,7 @@ func addArtifactContainersFromPod(pod *coreapi.Pod, worker *ArtifactWorker) {
 	if names := pod.Annotations[annotationWaitForContainerArtifacts]; len(names) > 0 {
 		waitForContainers = strings.Split(names, ",")
 	}
-	worker.CollectFromPod(pod.Name, hasArtifactsContainer, containers, waitForContainers)
+	worker.CollectFromPod(pod.Name, containers, waitForContainers)
 }
 
 func containerHasVolumeName(container coreapi.Container, name string) bool {
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -52,8 +52,8 @@ type ContainerNotifier interface {
 	// Complete indicates the specified pod has completed execution, been deleted, or that no further
 	// Notify() calls can be made.
 	Complete(podName string)
-	// Done returns true if the specified pod name has already completed any work it had pending.
-	Done(podName string) bool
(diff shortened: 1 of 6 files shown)


pkg/steps/artifacts.go
```diff
@@ -155,6 +155,10 @@ func (n *TestCaseNotifier) SubTests(prefix string) []*junit.TestCase {
 	return tests
 }
 
+type podExecInterface interface {
```
stevekuznetsov: super nit: Go idiom is to call this `podExecutor` after what it knows how to do

pkg/steps/artifacts.go
```diff
@@ -181,16 +190,14 @@ func copyArtifacts(podClient PodClient, into, ns, name, containerName string, pa
 		args = append(args, "-C", s, ".")
 	}
 
-	u := podClient.RESTClient().Post().Resource("pods").Namespace(ns).Name(name).SubResource("exec").VersionedParams(&coreapi.PodExecOptions{
+	e, err := podClient.Exec(ns, name, &coreapi.PodExecOptions{
```
stevekuznetsov: nit: the number of single letter vars here makes this hard to read

pkg/steps/artifacts_test.go
```diff
@@ -447,3 +454,70 @@ func TestTestCaseNotifier_SubTests(t *testing.T) {
...
+	}
+	return fmt.Errorf("unexpected command: %v", e.command)
+}
+
+func TestArtifactWorker(t *testing.T) {
+	tmp, err := ioutil.TempDir("", "")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(tmp)
```
stevekuznetsov: handle the error:

```go
defer func() {
    if err := os.RemoveAll(tmp); err != nil {
        t.Errorf("couldn't clean up tmpdir: %v", err)
    }
}()
```

pkg/steps/artifacts_test.go
```diff
@@ -447,3 +454,70 @@ func TestTestCaseNotifier_SubTests(t *testing.T) {
...
+	pod := "pod"
+	podClient := testPodClient{
+		PodsGetter: fake.NewSimpleClientset().CoreV1(),
+		namespace:  "namespace",
+		name:       pod,
+	}
+	w := NewArtifactWorker(podClient, tmp, podClient.namespace)
+	w.CollectFromPod(pod, true, []string{"container"}, nil)
+	w.Complete(pod)
+	for !w.Done(pod) {
```
stevekuznetsov: can we add a timeout here so pathological behavior can't happen

pkg/steps/artifacts_test.go
```diff
@@ -507,7 +507,10 @@ func TestArtifactWorker(t *testing.T) {
 	w := NewArtifactWorker(podClient, tmp, podClient.namespace)
 	w.CollectFromPod(pod, []string{"container"}, nil)
 	w.Complete(pod)
-	for !w.Done(pod) {
+	select {
+	case <-w.Done(pod):
+	case <-time.After(time.Second):
+		t.Fatal("timout waiting for artifact worker to finish")
```
stevekuznetsov: `s/timout/timeout/`

pkg/steps/artifacts_test.go
stevekuznetsov: NVM, you did it in a future commit

pkg/steps/artifacts.go
bbguimaraes: I thought about using that exact name, but it could be confused for something that executes pods, instead of commands in pods, so I named it after kube's PodInterface (which itself is not a great name). Changed to `podCmdExecutor`, hopefully somewhere in the middle.

pkg/steps/artifacts.go
bbguimaraes: I tried to touch the existing code as little as possible, mostly moved things around.

pkg/steps/artifacts_test.go
bbguimaraes: Copy-pasted, `grep` says we have a bunch of other tests to fix =)

pkg/steps/artifacts_test.go
bbguimaraes: Yes, I didn't want to spend too much time on it. The final solution is not great, but the wait time is fractions of a millisecond on my computer, so hopefully the one second timeout will be enough in our overloaded nodes.

PR #154, opened 2019-09-18
fix jq 1.5 flag in integration tests script

droslean:


droslean:
/hold

Let me add a check no make sure that diff won't exit 0 again.

PR #155, opened 2019-09-18
integration: Add sorting command and update expected files

droslean:
When we run `diff` with $(commands), it will still exit 0 even if the commands will fail. 

To prevent that, this PR add some extra checks to the jq commands.


bbguimaraes:
There's no need to sort the expected files.

We can also add `jq --sort-keys` (use the long version of the flags in scripts) to the end of the `ci-operator` command instead of using more files.

PR #162, opened 2019-09-18
Remove the commit releated header in OWNERS file

hongkailiu:
Using master/head as commit generates unnecessary modification for OWNERS file.
Ideally, we should use the last commit on the file and modify it only when the last commit is changed (as today, this need to implement list-commits function at upstream).

With limit time on this task, we remove the headers.

/cc @stevekuznetsov

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -22,9 +22,8 @@ import (
 )
 
 const (
-	doNotEdit     = "# DO NOT EDIT; this file is auto-generated using tools/populate-owners."
+	doNotEdit     = "# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools."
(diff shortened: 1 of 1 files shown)


petr-muller:
Maybe we can leave the header but without mentioning specific commit, to at least hint that the file should not me modified?

cmd/autoowners/main.go
```diff
@@ -22,10 +22,6 @@ import (
 )
 
 const (
-	doNotEdit     = "# DO NOT EDIT; this file is auto-generated using tools/populate-owners."
```
stevekuznetsov: I mean we can keep this -- (and update the tool name), just not make it commit-specific, so it does not change.

hongkailiu: Done.

PR #163, opened 2019-09-18
Add ci-operator-configresolver

AlexNPavel:
This adds a new ci-operator-configresolver command.

This PR is a breakout of PR #131 to make it easier to review and move forward.

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -1,0 +1,159 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/ghodss/yaml"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+)
+
+const (
+	orgQuery     = "org"
+	repoQuery    = "repo"
+	branchQuery  = "branch"
+	variantQuery = "variant"
+)
+
+type filenameToConfig map[string]api.ReleaseBuildConfiguration
+
+type options struct {
+	configs     filenameToConfig
+	configPath  string
+	logLevel    string
+	address     string
+	gracePeriod time.Duration
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.configPath, "config", "", "Path to config dirs")
+	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
+	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
+	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
+	fs.Parse(os.Args[1:])
+	return o
+}
+
+func validateOptions(o options) error {
+	_, err := log.ParseLevel(o.logLevel)
+	if err != nil {
+		return fmt.Errorf("invalid --log-level: %v", err)
+	}
+	return nil
+}
+
+func missingQuery(w http.ResponseWriter, field string) {
+	w.WriteHeader(http.StatusBadRequest)
+	fmt.Fprintf(w, "%s query missing or incorrect", field)
+	return
+}
+
+func resolveConfig(o *options) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			w.WriteHeader(http.StatusNotImplemented)
+			w.Write([]byte(http.StatusText(http.StatusNotImplemented)))
+			return
+		}
+		org := r.URL.Query().Get(orgQuery)
+		if org == "" {
+			missingQuery(w, orgQuery)
+			return
+		}
+		repo := r.URL.Query().Get(repoQuery)
+		if repo == "" {
+			missingQuery(w, repoQuery)
+			return
+		}
+		branch := r.URL.Query().Get(branchQuery)
+		if branch == "" {
+			missingQuery(w, branchQuery)
+			return
+		}
+		variant := r.URL.Query().Get(variantQuery)
+		info := config.Info{
+			Org:     org,
+			Repo:    repo,
+			Branch:  branch,
+			Variant: variant,
+		}
+
+		config, ok := o.configs[info.Basename()]
+		if !ok {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("config does not exist"))
+			return
+		}
+		jsonConfig, err := json.MarshalIndent(config, "", "  ")
+		if err != nil {
+			w.WriteHeader(http.StatusInternalServerError)
+			fmt.Fprintf(w, "failed to marshal config to JSON: %v", err)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
+		w.Write(jsonConfig)
+	}
+}
+
+// loadFilenameToConfig generates a new filenameToConfig map.
+func loadFilenameToConfig(configPath string) (filenameToConfig, error) {
+	log.Debug("Reloading configs")
+	configs := filenameToConfig{}
+	err := filepath.Walk(configPath, func(path string, info os.FileInfo, err error) error {
+		ext := filepath.Ext(path)
+		if info != nil && !info.IsDir() && (ext == ".yml" || ext == ".yaml") {
+			data, err := ioutil.ReadFile(path)
+			if err != nil {
+				return fmt.Errorf("failed to read ci-operator config (%v)", err)
+			}
+
+			var configSpec *api.ReleaseBuildConfiguration
+			if err := yaml.Unmarshal(data, &configSpec); err != nil {
+				return fmt.Errorf("failed to load ci-operator config (%v)", err)
+			}
+
+			if err := configSpec.ValidateAtRuntime(); err != nil {
+				return fmt.Errorf("invalid ci-operator config: %v", err)
+			}
+			log.Debugf("Adding %s to filenameToConfig", filepath.Base(path))
+			configs[filepath.Base(path)] = *configSpec
+		}
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	log.Info("Configs reloaded")
+	return configs, nil
+}
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(o)
+	if err != nil {
+		log.Fatalf("invalid options: %v", err)
+	}
+	level, _ := log.ParseLevel(o.logLevel)
+	log.SetLevel(level)
+
+	o.configs, err = loadFilenameToConfig(o.configPath)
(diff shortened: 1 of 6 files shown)


cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,135 @@
...
+	})
+	if err != nil {
+		return nil, err
+	}
+	log.Info("Configs reloaded")
+	return configs, nil
+}
+
+func main() {
+	opts := options{}
```
droslean: check [how](https://github.com/openshift/ci-tools/blob/master/cmd/pj-rehearse/main.go#L44-L79) we are used to parsing the flags.

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,135 @@
...
+		}
+		info.Filename = filepath.Join(o.configPath, branch, info.Basename())
+
+		config, ok := o.configs[info.Basename()]
+		if !ok {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("config does not exist"))
+			return
+		}
+		jsonConfig, err := json.Marshal(config)
```
droslean: use `MarshalIndent` instead.

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,135 @@
...
+type filenameToConfig map[string]api.ReleaseBuildConfiguration
+
+func resolveConfig(o *options) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
+			w.WriteHeader(http.StatusNotImplemented)
+			w.Write([]byte(http.StatusText(http.StatusNotImplemented)))
+			return
+		}
+		org := r.URL.Query().Get("org")
```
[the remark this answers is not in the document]
AlexNPavel: But this is parsing the result from an http.Request, and thus cannot be constant right?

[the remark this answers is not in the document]
droslean: I mean, `r.URL.Query().Get(constant)`

test/ci-operator-configresolver-integration/main.go
```diff
@@ -0,0 +1,88 @@
+package main
```
droslean: There is no need to create a new binary to run and compare the results. You can be advised from the `pj-rehearse` and `ci-operator` integration tests. Since you are hardcoding the configs, you can have a JSON expected file and `diff` it, while you will be able to `GET` via curl in the bash script.

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+# This test runs the ci-operator-configresolver and verifies that it returns
+# properly resolved configs
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOTDIR=$(pwd)
```
droslean: Use `"$( dirname "${BASH_SOURCE[0]}")"` instead

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -0,0 +1,28 @@
...
+WORKDIR="$( mktemp -d )"
+trap "rm -rf ${WORKDIR}" EXIT
+
+pushd $WORKDIR
```
droslean: no need for pushd and popd

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -0,0 +1,28 @@
...
+go build -o tester $ROOTDIR/test/ci-operator-configresolver-integration/main.go
+# copy registry to tmpdir to allow tester to modify registry
+cp -a $ROOTDIR/test/ci-operator-configresolver-integration/ tests
```
droslean: no need to copy anything.

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -0,0 +1,28 @@
...
+ci-operator-configresolver -config tests/configs -log-level debug &
+PID=$!
+# wait for registry to be resolved
+sleep 1
```
droslean: nit: I am not feeling comfortable to rely on a sleep command to wait when the service is ready to accept connections. Maybe it is better to expose another route just for it.

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -0,0 +1,28 @@
...
+if ! output=$( ./tester -serverAddress "http://127.0.0.1:8080" ); then
```
droslean: Use the `ci-operator-configresolver` command directly. This is going to be resolved in the `src` image when the integration tests will run.

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -0,0 +1,28 @@
...
+set -o errexit
+set -o nounset
+set -o pipefail
+
+ROOTDIR=$(pwd)
+WORKDIR="$( mktemp -d )"
+trap "rm -rf ${WORKDIR}" EXIT
+
+pushd $WORKDIR
+go build -o tester $ROOTDIR/test/ci-operator-configresolver-integration/main.go
```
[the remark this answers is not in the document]
droslean: Note: the tests will run in an image that will contain the `ci-operator-configresolver` binary of the corresponding PR.

test/ci-operator-configresolver-integration/main.go
AlexNPavel: I was intending on using this to more easily modify files and verify that the resolver picks up the changes automatically in future tests (this functionality will be added in a future PR). Thinking about it though, we could do this in bash a bit more easily, we would just need more input files we can copy over/modify and expected result files to compare against. I'll change it to simple bash diffs for now.

test/ci-operator-configresolver-integration/run.sh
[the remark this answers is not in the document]
AlexNPavel: This was building the separate test binary that wouldn't be included in the install. I will be removing that though, so this line will be removed.

test/ci-operator-configresolver-integration/run.sh
AlexNPavel: This was for the tester binary, not the configresolver. I'm removing the tester binary though, so this won't be necessary.

test/ci-operator-configresolver-integration/run.sh
AlexNPavel: The main issue is that, while the loading of the config files and exposing the service is really quick, it takes slightly longer than making the actual GET request. I think maybe having a loop with a timeout might work. We could have it attempt to get the proper config, and then try every second for 5 seconds, failing if the output is still not correct.

We'll probably need to do something like that when we add automatic reloading as well, as we might get a stale result from just immediately checking the output after updating a registry/config file.

test/ci-operator-configresolver-integration/run.sh
AlexNPavel: The reason that I'm copying them in is because we will be changing files in the registry/config dirs and verifying that the resolver picks up the changes. It's best to make all those changes from the temp workdir instead of the rootdir.

test/ci-operator-configresolver-integration/run.sh
AlexNPavel: `${BASH_SOURCE[0]}` only gets the relative path, but since we will be changing directory into a tmpdir we need an absolute path.

test/ci-operator-configresolver-integration/run.sh
stevekuznetsov: Using the upstream libraries to expose a health and readiness endpoint and using that to gauge when the service is ready to accept the request is preferred, but can be a follow-up.

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+const (
+	orgQuery     = "org"
+	repoQuery    = "repo"
+	branchQuery  = "branch"
+	variantQuery = "variant"
+)
+
+type filenameToConfig map[string]api.ReleaseBuildConfiguration
```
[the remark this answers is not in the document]
stevekuznetsov: (we don't have the `org/repo/` subdir structure and thereby cannot use the normal loaders, or generate normal `Info` objects. That's a shame.)

stevekuznetsov:
> IMHO you don't have to write integration tests for this tool. Just mock an http server using net/http/httptest and add unit tests.

@droslean using a mock HTTP server would be useful for people _relying_ on this, as you could inject fake handlers and not need to contact a server anywhere. For the actual server, we do want to test it end-to-end.

cmd/ci-operator-configresolver/main.go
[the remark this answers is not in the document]
AlexNPavel: We can't build a correct `config.Info` struct since we only have the filename and not the full directory structure, so `config.ByFilename` doesn't work for this (unless we want to intentionally omit the `Info` struct; we would still need to have a separate function to load the configs because the `LoadConfigByFilename` function sets the filename using the directory structure (because it sets the basename based on the `Info` struct, which is generated by `InfoFromPath`)).

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,160 @@
...
+			return
+		}
+		variant := r.URL.Query().Get(variantQuery)
+		info := config.Info{
+			Org:     org,
+			Repo:    repo,
+			Branch:  branch,
+			Variant: variant,
+		}
+		info.Filename = filepath.Join(o.configPath, branch, info.Basename())
```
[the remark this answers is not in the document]
AlexNPavel: Since we have separate configmaps for each branch, they get mounted to their own subdirs. That being said, there is no reason to set the `info.Filename` field anymore. That was used for the earlier PR for this which loaded the config file dynamically as requested from clients. I'll just remove this as it's unnecessary. The way that we load configs now is directly from the `filenameToConfig` map.

test/ci-operator-configresolver-integration/run.sh
AlexNPavel: While using exposing health and readiness endpoints can be good to indicate whether or not the server is ready to serve configs, it doesn't help much for the tests checking if the server picked up changes made on disk. The server will still be ready; it will just serve outdated information until it detects the changes and updates the internal mappings.

cmd/ci-operator-configresolver/main.go
```diff
@@ -0,0 +1,160 @@
...
+	"github.com/ghodss/yaml"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+)
+
+const (
+	orgQuery     = "org"
```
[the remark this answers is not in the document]
AlexNPavel: The main thing we gain is that now we are using a named constant so we are less prone to typos if we want to use any of these values elsewhere in this file. Typos should be caught by tests, but it is nice to have the IDE and `go vet` immediately identify that something is wrong. We also only have to change the value in one location instead of 2 if we decide the change a value.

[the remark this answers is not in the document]
stevekuznetsov: >  if we want to use any of these values elsewhere in this file

Do we?

stevekuznetsov:
@AlexNPavel needs a reabse

PR #165, opened 2019-09-19
ci-operator multi-stage: add integration test

bbguimaraes:


bbguimaraes:
Note to reviewers: I refuse to make any change that alters the number of lines added.

bbguimaraes:
As discussed previously, the order in which the pods are executed is unfortunately lost by `--determine-output`, but is verified by a [unit test](https://github.com/openshift/ci-tools/blob/bd95c0686be14dbdfd3757216feec81a1ad91026/pkg/steps/multi_stage_test.go#L214).

PR #166, opened 2019-09-19
Refactor labels injection and add tests.

droslean:
This patch only moves the label injection code to its own function and adds unit tests for it.

### DIFF
--- a/pkg/steps/template_test.go
+++ b/pkg/steps/template_test.go
@@ -6,12 +6,16 @@ import (
 	coreapi "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 
+	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/pod-utils/downwardapi"
+
 	"k8s.io/apimachinery/pkg/api/equality"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/diff"
 
 	templateapi "github.com/openshift/api/template/v1"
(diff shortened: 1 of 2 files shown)


pkg/steps/template_test.go
```diff
@@ -238,3 +242,114 @@ func TestOperateOnTemplatePods(t *testing.T) {
...
+			testID: "nil refs in jobspec, no injection expected",
+			jobSpec: &api.JobSpec{
+				JobSpec: downwardapi.JobSpec{
+					Refs: nil,
+				},
+			},
+			template: &templateapi.Template{
+				TypeMeta:   meta.TypeMeta{Kind: "Template", APIVersion: "template.openshift.io/v1"},
+				ObjectMeta: meta.ObjectMeta{Name: "test-template"},
+				Objects: []runtime.RawExtension{
```
stevekuznetsov: This field is not germane to this test. we should remove it

droslean: addressed

pkg/steps/template_test.go
```diff
@@ -238,3 +242,90 @@ func TestOperateOnTemplatePods(t *testing.T) {
...
+							Raw:    []byte(runtime.EncodeOrDie(corev1Codec, pod)),
+							Object: pod.DeepCopyObject(),
+						}
+					}(),
+				},
+			},
+			expected: &templateapi.Template{
+				TypeMeta:   meta.TypeMeta{Kind: "Template", APIVersion: "template.openshift.io/v1"},
+				ObjectMeta: meta.ObjectMeta{Name: "test-template"},
+				Objects: []runtime.RawExtension{
```
stevekuznetsov: I meant _any_ Object field . We just need to check that the template has ObjectLabels set, so all this is extra in every test case

droslean: oops, yea. I removed the others as well.

PR #167, opened 2019-09-19
Add tool auto-config-brancher

hongkailiu:
Automate @petr-muller routine on Monday. For example, https://github.com/openshift/release/pull/5029

```
/user/bin/determinize-ci-operator --config-dir ./ci-operator/config --current-release ${CURRENT_RELEASE} --future-release ${FUTURE_RELEASE} --confirm
/user/bin/config-brancher --config-dir ./ci-operator/config --current-release ${CURRENT_RELEASE} --future-release ${FUTURE_RELEASE} --confirm
/user/bin/ci-operator-prowgen --from-dir ./ci-operator/config --to-dir ./ci-operator/jobs
```

### DIFF
--- a/cmd/autoconfigbrancher/main.go
+++ b/cmd/autoconfigbrancher/main.go
@@ -1,0 +1,188 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/experiment/autobumper/bumper"
+	"k8s.io/test-infra/prow/config/secret"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/github"
+)
+
+const (
+	githubOrg   = "openshift"
+	githubRepo  = "release"
+	githubLogin = "openshift-bot"
+	githubTeam  = "openshift/openshift-team-developer-productivity-test-platform"
+)
+
+var (
+	count = 0
+)
+
+type options struct {
+	githubLogin    string
+	githubToken    string
+	gitName        string
+	gitEmail       string
+	targetDir      string
+	assign         string
+	currentRelease string
+	futureReleases flagutil.Strings
+	debugMode      bool
+}
+
+func parseOptions() options {
+	var o options
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	flag.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
+	flag.StringVar(&o.githubToken, "github-token", "", "The path to the GitHub token file.")
+	flag.StringVar(&o.gitName, "git-name", "", "The name to use on the git commit. Requires --git-email. If not specified, uses the system default.")
+	flag.StringVar(&o.gitEmail, "git-email", "", "The email to use on the git commit. Requires --git-name. If not specified, uses the system default.")
+	flag.StringVar(&o.targetDir, "target-dir", "", "The directory containing the target repo.")
+	flag.StringVar(&o.assign, "assign", githubTeam, "The github username or group name to assign the created pull request to.")
+	flag.StringVar(&o.currentRelease, "current-release", "", "Configurations targeting this release will get branched.")
+	flag.Var(&o.futureReleases, "future-release", "Configurations will get branched to target this release, provide one or more times.")
+	flag.BoolVar(&o.debugMode, "debug-mode", false, "Enable the DEBUG level of logs if true.")
+	flag.Parse()
+	return o
+}
+
+func validateOptions(o options) error {
+	if o.githubLogin == "" {
+		return fmt.Errorf("--github-login cannot be empty string")
+	}
+	if o.githubToken == "" {
+		return fmt.Errorf("--github-token is mandatory")
+	}
+	if (o.gitEmail == "") != (o.gitName == "") {
+		return fmt.Errorf("--git-name and --git-email must be specified together")
+	}
+	if o.targetDir == "" {
+		return fmt.Errorf("--target-dir is mandatory")
+	}
+	if o.assign == "" {
+		return fmt.Errorf("--assign is mandatory")
+	}
+	if o.currentRelease == "" {
+		return fmt.Errorf("--current-release is mandatory")
+	}
+	if len(o.futureReleases.Strings()) == 0 {
+		return fmt.Errorf("--future-release is mandatory")
+	}
+	return nil
+}
+
+func hasChanges() (bool, error) {
+	cmd := "git"
+	args := []string{"status", "--porcelain"}
+	logrus.WithField("cmd", cmd).WithField("args", args).Info("running command ...")
+	combinedOutput, err := exec.Command(cmd, args...).CombinedOutput()
+	if err != nil {
+		logrus.WithField("cmd", cmd).Debugf("output is '%s'", string(combinedOutput))
+		return false, err
+	}
+	return len(strings.TrimSuffix(string(combinedOutput), "\n")) > 0, nil
+}
+
+func commitIfNeeded(msg, author string) {
+	changed, err := hasChanges()
+	if err != nil {
+		logrus.WithError(err).Fatal("error occurred when checking changes")
+	}
+	if changed {
+		count++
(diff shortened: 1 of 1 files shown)


cmd/autoconfigbrancher/helper/helper.go
```diff
@@ -0,0 +1,24 @@
+package helper
```
stevekuznetsov: Why does this need to be a separate package?

cmd/autoconfigbrancher/helper/helper.go
```diff
@@ -0,0 +1,24 @@
...
+	stdout := &OutputWriter{}
+	stderr := &OutputWriter{}
+	c := exec.Command(cmd, args...)
+	c.Stdout = stdout
+	c.Stderr = stderr
+	e := c.Run()
+	return stdout.output, stderr.output, e
+}
+
+type OutputWriter struct {
```
stevekuznetsov: This looks a lot like `bytes.Buffer()`

cmd/autoconfigbrancher/main.go
```diff
@@ -0,0 +1,170 @@
...
+	i, err := strconv.Atoi(strings.TrimSuffix(string(stdout), "\n"))
+	if err != nil {
+		return false, err
+	}
+	return i > 0, nil
+}
+
+func run(cmd string, args ...string) {
+	logrus.WithField("cmd", cmd).WithField("args", args).Info("running command ...")
+	if _, stderr, err := helper.Call(cmd, args...); err != nil {
```
stevekuznetsov: Why supress `stdout`?

cmd/autoconfigbrancher/main.go
```diff
@@ -0,0 +1,170 @@
...
+		logrus.WithError(err).WithField("string(stderr)", string(stderr)).Fatal("Failed to run command")
```
stevekuznetsov: Don't add the stderr as a field, just put `string(stderr)` into `Fatal()`

cmd/autoconfigbrancher/main.go
```diff
@@ -0,0 +1,170 @@
...
+
+	cmd = "/usr/bin/config-brancher"
+	run(cmd, args...)
+
+	cmd = "/usr/bin/ci-operator-prowgen"
+	args = []string{"--from-dir", "./ci-operator/config", "--to-dir", "./ci-operator/jobs"}
+	run(cmd, args...)
+
+	// Here we use `git-rev-list`, together with the tag `before` to check
+	// if any new commits has been generated so far by the release-tools above.
```
stevekuznetsov: What is actually making commits? I don't see it anywhere

cmd/autoconfigbrancher/helper/helper.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+func Call(cmd string, args ...string) ([]byte, []byte, error) {
+	stdout := &OutputWriter{}
+	stderr := &OutputWriter{}
+	c := exec.Command(cmd, args...)
+	c.Stdout = stdout
+	c.Stderr = stderr
+	e := c.Run()
+	return stdout.output, stderr.output, e
```
stevekuznetsov: I think you can just use `os.Cmd.Output()` or `CombinedOutput()`

cmd/autoconfigbrancher/helper/helper.go
hongkailiu: I want to keep this open to other tools for running `bash` command.
I feel it is cleaner this way than using `bumper` from upstream.

cmd/autoconfigbrancher/main.go
hongkailiu: In this tool, we did not use `stdout` at all, except `git-rev-list`.
This function is an shortcut for log, call, fital in the `main` function.

cmd/autoconfigbrancher/helper/helper.go
stevekuznetsov: I think this package has very little utility as noted in other comments, we should consider removing it

cmd/autoconfigbrancher/main.go
hongkailiu: I thought the 3-tool calls generate commits.
I cannot test it on master of release repo today because it does not generate any change.

cmd/autoconfigbrancher/main.go
stevekuznetsov: Right, but we likely want logs to actually show what the tools are doing in case we need to debug them at all.

cmd/autoconfigbrancher/helper/helper.go
hongkailiu: Removing ...

cmd/autoconfigbrancher/main.go
stevekuznetsov: No, the tools just change files, they do not do any changes to `git`

cmd/autoconfigbrancher/helper/helper.go
hongkailiu: Done.

cmd/autoconfigbrancher/helper/helper.go
hongkailiu: Using `Cmd.Output()` now as you pointed out.
This struct is removed.

cmd/autoconfigbrancher/main.go
hongkailiu: Not sure if I did what you want. So keep this comment open.

cmd/autoconfigbrancher/main.go
hongkailiu: Done.

cmd/autoconfigbrancher/main.go
hongkailiu: Then I misunderstood your comment from JIRA card.
Working on this now.

cmd/autoconfigbrancher/helper/helper.go
hongkailiu: Done.

cmd/autoconfigbrancher/main.go
```diff
@@ -0,0 +1,177 @@
...
+	if err != nil {
+		return false, err
+	}
+	return i > 0, nil
+}
+
+func run(cmd string, args ...string) {
+	logrus.WithField("cmd", cmd).WithField("args", args).Info("running command ...")
+	if out, err := exec.Command(cmd, args...).Output(); err != nil {
+		logrus.WithField("cmd", cmd).Debugf("output is '%s'", string(out))
```
stevekuznetsov: You likely want to either use `CombinedOutput()` or continue using `Output()` but cast the error and extract stder from it, so we show stderr on failure

cmd/autoconfigbrancher/main.go
hongkailiu: Replace with the new command.
Resolving this one.

hongkailiu:
Found another problem ...
https://github.com/openshift/release/pull/5112/files

Now the tool generate PR ... after git pull on release repo.

However, it updates PR which is closed already. :(

hongkailiu:
Changed a new matchTitle ... it looks OK now.

https://github.com/openshift/release/pull/5113

hongkailiu:
lots of checking is not happy with the change :(

cmd/autoconfigbrancher/main.go
hongkailiu: Done.

cmd/autoconfigbrancher/main.go
hongkailiu: Now I see what you meant ... chose `CombinedOutput` as it is simpler.

cmd/autoconfigbrancher/main.go
```diff
@@ -0,0 +1,187 @@
...
+	return nil
+}
+
+func hasChanges() (bool, error) {
+	cmd := "git"
+	args := []string{"status", "--porcelain"}
+	logrus.WithField("cmd", cmd).WithField("args", args).Info("running command ...")
+	combinedOutput, err := exec.Command(cmd, args...).CombinedOutput()
+	if err != nil {
+		return false, err
```
petr-muller: nit: if the command ever fails, I want to see its error output - please at least log it as debug like you do in `run()`

hongkailiu: I missed that one.

PR #161, opened 2019-09-18
Add a tool to trigger images postsubmits from changes

stevekuznetsov:
When CI Operator configuration is changed in a merge, it may impact the
output images that will be promoted. These changes do not take effect
until the next merge happens to the branch in question, which is often
not close in time to when the configuration changes. Therefore, we add a
component that will run to analyze merged configuration changes and
determine if an image re-build is necessary and trigger it if so.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/cc @openshift/openshift-team-developer-productivity-test-platform

PR #169, opened 2019-09-20
Ignore the changes on modules.txt

hongkailiu:
./vendor/modules.txt can be modified (reordering the lines in it)
even if the dependencies are not changed. This forces us to run
`make validate-vendor` for each commit.

We ingore the changes on that file in the check.

https://github.com/openshift/ci-tools/pull/161#issuecomment-533296228

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ validate-vendor:
 	GO111MODULE=on GOPROXY=https://proxy.golang.org go mod tidy
 	GO111MODULE=on GOPROXY=https://proxy.golang.org go mod vendor
 	git status -s ./vendor/ go.mod go.sum
-	test -z "$$(git status -s ./vendor/ go.mod go.sum)"
+	test -z "$$(git status -s ./vendor/ go.mod go.sum | grep -v vendor/modules.txt)"
 .PHONY: validate-vendor
 
 lint:


hongkailiu:
I did not find any official doc about `modules.txt`. Most importantly whether or not it is used in `go build/install` in vendor mode. It should not in my opinion.

Makefile
```diff
@@ -20,7 +20,7 @@ validate-vendor:
 	GO111MODULE=on GOPROXY=https://proxy.golang.org go mod tidy
 	GO111MODULE=on GOPROXY=https://proxy.golang.org go mod vendor
 	git status -s ./vendor/ go.mod go.sum
-	test -z "$$(git status -s ./vendor/ go.mod go.sum)"
+	test -z "$$(git status -s ./vendor/ go.mod go.sum | grep -v vendor/modules.txt)"
```
droslean: maybe `git status` is not a good idea to compare the differences. We can grab the old modules file and compare with the new version after the commands but both sorted.

hongkailiu: my 2 cents:
Without understanding the meaning of the file ... How to compare? How can we be sure sorting wont impact the build?

Before the check is the strictest, they have to be literally identical.
Now we ignore that file (maybe we shouldn't), which is the most lenient.

The official doc/src for the file would be great to support checking between those two.

droslean: The sorting is changing because of the concurrent import scanning (that's is also why it is quick). However, let's ignore it in the `git status` command, but let's also add an extra step to compare it. Lets create a copy `modules.txt.orig` before the `go mod` commands, and then 
`diff <(sort modules.txt.orig) <(sort modules.txt)`

hongkailiu: _It appears that the BuildList itself is sorted. If there is any nondeterminism, it likely comes from concurrent import scanning, but I haven't been able to reproduce it in a unit-test so far._

It wont bite us if we add the `diff` in our check. I understand that. I will add to see if it only change is the ordering.

Just wanna point out, we have a different scenario than the issue there.
* In the issue, they generate vendor on the fly in CI and module.txt is changed with the same code base. We never had this. For us, whenever we do the make-validate-vendor locally, the check will pass in ci-pod.
* In our case, we change the code with our commits (but not adding/removing the dependencies), then module.txt is sometime (maybe always?) changed.

The unit test the GoLang person did must test a different thing because we have seen it so often and he cannot reproduce.

My question is still valid i think: Is there any GoLang doc saying, _we can get the same build if ordering of that file is changed_.

droslean: > My question is still valid i think: Is there any GoLang doc saying, we can get the same build if ordering of that file is changed.

I don't have any clue. It is worth investigating.

PR #170, opened 2019-09-20
determinize all ci-operator configs

petr-muller:
I don't think there's a good reason for determinizing only the configs that would be branched. @stevekuznetsov WDYT?


stevekuznetsov:
I originally did this just because people were using comments, special sorting, formatting etc in YAML which would get needlessly clobbered -- the impact of determinization is a loss in UX

petr-muller:
> I originally did this just because people were using comments, special sorting, formatting etc in YAML which would get needlessly clobbered -- the impact of determinization is a loss in UX

Ah, so only clobber those where we expect machine-write to? That makes sense.

/shrug
/close

PR #171, opened 2019-09-20
ci-operator multi-stage: create secret

bbguimaraes:
Each test is assigned a secret with a pre-defined name (already
accessible via the `$JOB_NAME_SAFE` environment variable) that will be
mounted at a specific path in all pods (from `pre` to `post`) and can be
used to pass a small amount of data to subsequent steps.

### DIFF
--- a/cmd/secret-wrapper/main.go
+++ b/cmd/secret-wrapper/main.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"os/exec"
+	"os/signal"
+	"syscall"
+
+	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const (
+	DIR = "/tmp/secret"
+)
+
+func main() {
+	if err := run(os.Args); err != nil {
+		fmt.Fprintln(os.Stderr, err)
+		os.Exit(1)
+	}
+}
+
+func run(argv []string) error {
+	if len(os.Args) == 1 {
+		return fmt.Errorf("Usage: %s cmd...\n", os.Args[0])
+	}
+	var ns, name string
+	if ns = os.Getenv("NAMESPACE"); ns == "" {
+		return fmt.Errorf("environment variable NAMESPACE is empty")
+	}
+	if name = os.Getenv("JOB_NAME_SAFE"); name == "" {
+		return fmt.Errorf("environment variable JOB_NAME_SAFE is empty")
+	}
+	client, err := loadClient(ns)
+	if err != nil {
+		return err
+	}
+	if err := execCmd(os.Args); err != nil {
+		return fmt.Errorf("failed to execute wrapped command: %v\n", err)
+	}
+	if _, err := os.Stat(DIR); err != nil {
+		if os.IsNotExist(err) {
+			return nil
+		}
+		return fmt.Errorf("failed to stat directory %q: %v", DIR, err)
+	}
+	secret, err := util.SecretFromDir(DIR)
+	if err != nil {
+		return fmt.Errorf("failed to generate secret: %v\n", err)
+	}
+	secret.Name = name
+	if _, err := util.UpdateSecret(client, secret); err != nil {
+		return fmt.Errorf("failed to update secret: %v\n", err)
+	}
+	return nil
+}
+
+func loadClient(namespace string) (coreclientset.SecretInterface, error) {
+	config, err := util.LoadClusterConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load cluster config: %v", err)
+	}
+	client, err := coreclientset.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create client: %v", err)
+	}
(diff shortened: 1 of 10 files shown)


pkg/steps/multi_stage.go
```diff
@@ -20,6 +20,7 @@ import (
 
 const (
 	multiStageTestLabel = "ci.openshift.io/multi-stage-test"
+	secretMountPath     = "/var/run/secrets/ci.openshift.io/multi-stage"
```
bbguimaraes: /shrug

bbguimaraes:
> If someone writes data into that directory, how do we backfill that data into the secret? Will this be easier than using a persistent volume?

We don't. One of the design decisions for E2E tests was that test steps would not need to share data, all communication should be done through the test cluster.

The secret in that case is only used to expose the `kubeconfig` that grants access to the test cluster and to store data needed to tear it down. Only one step in the test modifies the secret, it uses a single `curl` command to do so (it can be done with `oc`, if available), and it is likely to be written once and shared via references or workflows.

Because of that, I don't think persistent volumes are necessary. We could make it easier to transfer data from the container to the secret (as previously discussed), but it doesn't seem worth it. It can be added later if we determine it is necessary.

stevekuznetsov:
@bbguimaraes you mean `curl` in the actual job configuration? That should not be in the domain of the test. We should consider a sidecar or something to do the upload for people. I don't think that users should be using the API using `curl` in their test to get this functionality

bbguimaraes:
> @bbguimaraes you mean `curl` in the actual job configuration?

Yes, the last command in (my version of) the `installer` step is a PUT to the API server to update the secret.

> That should not be in the domain of the test. We should consider a sidecar or something to do the upload for people. I don't think that users should be using the API using `curl` in their test to get this functionality

A sidecar would require either equally intrusive changes to the test or cooperation from `ci-operator` similar to what is done with artifacts (*shudder*). A wrapper for the test command, as you've suggested in the past, would be better suited. Given that it would be completely independent of this PR, which is only concerned with creating the secret and adding the mounts, can we agree to merge it and add a separate task to create and use that tool?

stevekuznetsov:
Why not pull the data back in `ci-operator` and create the secret from there, a la pushing config up to GCS?

bbguimaraes:
As decided in private discussion, the secret should be updated automatically from the contents of a directory.

Rebased, added a tool that does that, integrated it with multi-stage tests, tested the E2E suite. The PR is now `XL`, but https://github.com/openshift/ci-tools/pull/171/commits/70bfb69535bb3e06929559377bcd281cddd9026c is the rebased version of the old commit and half the other changes are unit and integration test updates.

cmd/secret-wrapper/main.go
```diff
@@ -0,0 +1,94 @@
...
+
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const (
+	DIR = "/tmp/secret"
+)
+
+func main() {
+	if err := run(os.Args); err != nil {
```
stevekuznetsov: Why not use `entrypoint`?

bbguimaraes: There isn't much overlap other than the very simple function that executes a program, and we would still need a separate container to create the secret when `entrypoint` finishes.

cmd/secret-wrapper/main.go
```diff
@@ -0,0 +1,94 @@
...
+	"os/signal"
+	"syscall"
+
+	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const (
+	DIR = "/tmp/secret"
```
stevekuznetsov: nit: this is not idomatic Go

bbguimaraes: I'll admit it took me a long time to realize the capitalization was the problem (had to `grep` for `const`) =)

cmd/secret-wrapper/main.go
```diff
@@ -0,0 +1,94 @@
...
+	}
+	if name = os.Getenv("JOB_NAME_SAFE"); name == "" {
+		return fmt.Errorf("environment variable JOB_NAME_SAFE is empty")
+	}
+	client, err := loadClient(ns)
+	if err != nil {
+		return err
+	}
+	if err := execCmd(os.Args); err != nil {
+		return fmt.Errorf("failed to execute wrapped command: %v\n", err)
```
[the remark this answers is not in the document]
bbguimaraes: Bad refactoring.

bbguimaraes:
> I worry there are on-exit bugs we've solved in entrypoint that will need to be re-solved since we're not reusing it

I completely understand. I had a look at the `git` history when evaluating `entrypoint`, problems were usually either in the interaction with other components, the timeout, or the files it creates, none of which are required in our case — it could be argued those bugs are a reason *not* to use it.

`secret-wrapper` follows kube's [documentation](https://kubernetes.io/docs/concepts/workloads/pods/pod/#termination-of-pods) and should be sufficient for the narrower case it supports.

bbguimaraes:
Added integration tests.

stevekuznetsov:
/hold cancel
(oops)

PR #173, opened 2019-09-20
ci-op-configresolver: add fsnotify and periodic reloaders

AlexNPavel:
This adds both a periodic reloader and an fs event based reloader for the ci-operator-configresolver. The fsnotify reloader uses a new `Coalescer` type that only allows one instance of the reload to run at a time. Any other instance that is started while the reload is still occuring simply waits for the existing instance to finish and then returns. This is useful because if we mount configs via a configmap, changing a single file can result in 10s of events due to the way configmap mounts work, which would result in the reloader running many times. The periodic reloader is included in case the fsnotify reloader manages to run before all the configmap mount changes are finished and makes sure that the configs being served by the resolver are eventually consistent.

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -4,18 +4,15 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
-	"path/filepath"
 	"time"
 
-	"github.com/ghodss/yaml"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/test-infra/prow/interrupts"
 
-	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/load"
 )
 
 const (
@@ -25,14 +22,12 @@ const (
 	variantQuery = "variant"
 )
 
-type filenameToConfig map[string]api.ReleaseBuildConfiguration
-
 type options struct {
-	configs     filenameToConfig
 	configPath  string
 	logLevel    string
 	address     string
 	gracePeriod time.Duration
+	cycle       time.Duration
 }
 
 func gatherOptions() options {
@@ -42,6 +37,7 @@ func gatherOptions() options {
 	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
 	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
 	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
+	fs.DurationVar(&o.cycle, "cycle", time.Minute*2, "Cycle duration for config reload")
 	fs.Parse(os.Args[1:])
 	return o
 }
@@ -51,6 +47,9 @@ func validateOptions(o options) error {
 	if err != nil {
 		return fmt.Errorf("invalid --log-level: %v", err)
 	}
+	if o.cycle == 0 {
+		return fmt.Errorf("invalid cycle: duration cannot equal 0")
+	}
 	return nil
 }
 
@@ -59,7 +58,7 @@ func missingQuery(w http.ResponseWriter, field string) {
 	fmt.Fprintf(w, "%s query missing or incorrect", field)
 }
 
-func resolveConfig(o *options) http.HandlerFunc {
+func resolveConfig(agent load.ConfigAgent) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if r.Method != "GET" {
 			w.WriteHeader(http.StatusNotImplemented)
@@ -89,10 +88,10 @@ func resolveConfig(o *options) http.HandlerFunc {
 			Variant: variant,
 		}
 
-		config, ok := o.configs[info.Basename()]
-		if !ok {
+		config, err := agent.GetConfig(info)
+		if err != nil {
 			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte("config does not exist"))
+			fmt.Fprintf(w, "failed to get config: %v", err)
 			return
 		}
 		jsonConfig, err := json.MarshalIndent(config, "", "  ")
@@ -106,36 +105,11 @@ func resolveConfig(o *options) http.HandlerFunc {
 	}
 }
 
-// loadFilenameToConfig generates a new filenameToConfig map.
-func loadFilenameToConfig(configPath string) (filenameToConfig, error) {
-	log.Debug("Reloading configs")
-	configs := filenameToConfig{}
-	err := filepath.Walk(configPath, func(path string, info os.FileInfo, err error) error {
-		ext := filepath.Ext(path)
-		if info != nil && !info.IsDir() && (ext == ".yml" || ext == ".yaml") {
-			data, err := ioutil.ReadFile(path)
-			if err != nil {
-				return fmt.Errorf("failed to read ci-operator config (%v)", err)
-			}
-
-			var configSpec *api.ReleaseBuildConfiguration
-			if err := yaml.Unmarshal(data, &configSpec); err != nil {
-				return fmt.Errorf("failed to load ci-operator config (%v)", err)
-			}
-
-			if err := configSpec.ValidateAtRuntime(); err != nil {
-				return fmt.Errorf("invalid ci-operator config: %v", err)
-			}
-			log.Debugf("Adding %s to filenameToConfig", filepath.Base(path))
-			configs[filepath.Base(path)] = *configSpec
-		}
-		return nil
-	})
-	if err != nil {
-		return nil, err
+func getGeneration(agent load.ConfigAgent) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.WriteHeader(http.StatusOK)
+		fmt.Fprintf(w, "%d", agent.GetGeneration())
 	}
-	log.Info("Configs reloaded")
-	return configs, nil
 }
 
 func main() {
@@ -147,12 +121,13 @@ func main() {
 	level, _ := log.ParseLevel(o.logLevel)
 	log.SetLevel(level)
 
-	o.configs, err = loadFilenameToConfig(o.configPath)
+	configAgent, err := load.NewConfigAgent(o.configPath, o.cycle)
 	if err != nil {
-		log.Fatalf("Failed to load configs: %s", err)
+		log.Fatalf("Failed to get config agent: %v", err)
 	}
 
-	http.HandleFunc("/config", resolveConfig(&o))
+	http.HandleFunc("/config", resolveConfig(configAgent))
+	http.HandleFunc("/generation", getGeneration(configAgent))
 	interrupts.ListenAndServe(&http.Server{Addr: o.address}, o.gracePeriod)
 	interrupts.WaitForGracefulShutdown()
 }
--- a/pkg/coalescer/coalescer.go
+++ b/pkg/coalescer/coalescer.go
@@ -1,0 +1,46 @@
+package coalescer
+
+import (
+	"sync"
+)
+
+// Coalescer contains a Run function that will run a specified function.
+// If the function is already being run in a separate thread, the Coalescer
+// will simply wait until the function finishes and return nil
+type Coalescer interface {
+	Run() error
+}
+
+// NewCoalescer returns a new Coalescer with the given function set as the runFunc.
+func NewCoalescer(f func() error) Coalescer {
+	return &coalescer{runFunc: f, once: &sync.Once{}}
+}
+
+type coalescer struct {
+	sync.Mutex
+	once    *sync.Once
+	runFunc func() error
+}
+
+// Run the runFunc. If the runFunc is currently being run in another thread, wait
+// and return nil when the the other thread finishes.
+func (c *coalescer) Run() error {
+	var err error
+	var once *sync.Once
+
+	c.Lock()
+	once = c.once
+	c.Unlock()
+	once.Do(func() {
+		defer func() {
+			c.Lock()
+			c.once = &sync.Once{}
+			c.Unlock()
+		}()
+		err = c.runFunc()
+		if err != nil {
+			return
+		}
+	})
+	return err
+}
--- a/pkg/coalescer/coalescer_test.go
+++ b/pkg/coalescer/coalescer_test.go
@@ -1,0 +1,30 @@
+package coalescer
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+)
+
+func TestRun(t *testing.T) {
+	sharedVar := 0
+	wg := &sync.WaitGroup{}
+	c := NewCoalescer(func() error {
+		fmt.Printf("Hello\n")
+		time.Sleep(time.Millisecond * 250)
+		sharedVar++
+		return nil
+	})
+	for n := 0; n < 10; n++ {
+		wg.Add(1)
+		go func() {
+			c.Run()
+			wg.Done()
+		}()
+	}
+	wg.Wait()
+	if sharedVar != 1 {
+		t.Errorf("Shared var should equal 1, but instead equals %d", sharedVar)
+	}
+}
--- a/pkg/load/configAgent.go
+++ b/pkg/load/configAgent.go
@@ -1,0 +1,161 @@
+package load
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"time"
+
+	"github.com/fsnotify/fsnotify"
+	log "github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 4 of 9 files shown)


test/ci-operator-configresolver-integration/run.sh
```diff
@@ -14,11 +14,20 @@ trap "rm -rf ${WORKDIR}" EXIT
...
+ci-operator-configresolver -config tests/configs -log-level debug -cycle 2m &
 PID=$!
 if ! timeout 10s bash -c 2>/dev/null -- "until diff tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'); do sleep 1; done" &>/dev/null; then
     echo "diff: $(diff -ru tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'))"
     kill $PID
+    wait $PID
+    exit 1
+fi
+cp tests/configs2/release-4.2/openshift-installer-release-4.2-golang111.yaml tests/configs/release-4.2/openshift-installer-release-4.2.yaml
+if ! timeout 10s bash -c 2>/dev/null -- "until diff tests/expected/openshift-installer-release-4.2-golang111.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'); do sleep 1; done" &>/dev/null; then
```
bbguimaraes: Suggestion: add a `/generation` (or whatever) endpoint that is incremented after each update and check it here. As it is, we'll block for up to 20s in case of a true positive.

pkg/coalescer/coalescer.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	// inform the waiters that we are done
+	c.Lock()
+	c.loading = false
+	c.loader.L.Lock()
+	c.loader.Broadcast()
+	c.loader.L.Unlock()
+	c.Unlock()
+	return nil
```
bbguimaraes: Looking at the resolver, it seems subsequent invocations don't actually need to wait for the first, is that right?
- Yes:
```go
// sync/atomic
if atomic.CompareAndSwapUint32(&loading, 0, 1) {
    err = f()
    atomic.StoreUint32(&loading, 0)
}
```
(there may be an equivalent in the stdlib, but this works)
- No:
```go
// sync.Once
once.Do(func() { err = f() })
```

AlexNPavel: The resolver itself should work fine even if there are multiple invocations running at the same time and it doesn't actually need to wait. However, I prefer having the call to coalescer to be blocking, so that if a function calls it and the reload is going on, the caller waits until an up to date version of the resolver is ready.

cmd/ci-operator-configresolver/main.go
```diff
@@ -42,6 +37,7 @@ func gatherOptions() options {
 	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
 	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
 	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
+	fs.DurationVar(&o.cycle, "cycle", 0, "cycle to reload config files on")
```
stevekuznetsov: Explain this is time

pkg/load/configAgent.go
```diff
@@ -0,0 +1,162 @@
...
+	a.lock = &sync.Mutex{}
+	configCoalescer := a.getConfigCoalescer()
+	err := configCoalescer.Run()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load configs: %v", err)
+	}
+
+	// periodic reload
+	if a.cycle != 0 {
+		interrupts.Run(func(ctx context.Context) {
```
stevekuznetsov: Use `interrupts.Tick`

pkg/load/configAgent.go
```diff
@@ -0,0 +1,162 @@
...
+		}
+		return nil
+	})
+}
+
+func reloadWatcher(ctx context.Context, w *fsnotify.Watcher, c coalescer.Coalescer, path string) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
stevekuznetsov: Seems like you could close the watcher here for a little cleaner workflow instead of the defer above

pkg/load/configAgent.go
```diff
@@ -0,0 +1,162 @@
...
+				w.Remove(event.Name)
+			}
+			go c.Run()
+			// add new files to be watched; if a watch already exists on a file, the
+			// watch is simply updated
+			if err := populateWatcher(w, path); err != nil {
+				log.Errorf("Failed to update fsnotify watchlist: %v", err)
+			}
+		case err := <-w.Errors:
+			log.Errorf("error: %v", err)
```
stevekuznetsov: Use `logrus.WithError(err).Error("whatever")`

pkg/load/configAgent.go
```diff
@@ -0,0 +1,162 @@
...
+	}
+	a.lock.Lock()
+	a.configs = configs
+	a.generation++
+	a.lock.Unlock()
+	log.Info("Configs reloaded")
+	return nil
+}
+
+func (a *agent) getConfigCoalescer() coalescer.Coalescer {
```
stevekuznetsov: Why is this called out as a method?

AlexNPavel: I originally had more logic there that I moved into `loadFilenameToConfig`. It's unnecessary now, so I'll remove it.

pkg/coalescer/coalescer_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+	"time"
+)
+
+func TestRun(t *testing.T) {
+	sharedVar := 0
+	wg := &sync.WaitGroup{}
+	c := NewCoalescer(func() error {
+		fmt.Printf("Hello\n")
+		time.Sleep(time.Millisecond * 50)
```
stevekuznetsov: This test will flake if the scheduler decides to take more than 50ms to run the later `go func()` calls. Let's make this 250ms or so, long enough to not flake but also not be annoying when run in parallel but

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+	configs    filenameToConfig
+	configPath string
+	cycle      time.Duration
+	generation int
+}
+
+func NewConfigAgent(configPath string, cycle time.Duration) (ConfigAgent, error) {
+	a := &agent{configPath: configPath, cycle: cycle}
+	// mutex to prevent data race on o.configs
+	a.lock = &sync.Mutex{}
```
stevekuznetsov: Why isnt this in the literal declaration for `a`?

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+	configCoalescer := coalescer.NewCoalescer(a.loadFilenameToConfig)
+	err := configCoalescer.Run()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load configs: %v", err)
+	}
+
+	// periodic reload
+	if a.cycle != 0 {
```
stevekuznetsov: What's the use in disabling this?

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+			}
+		}, a.cycle)
+	}
+
+	// fsnotify reload
+	configWatcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, fmt.Errorf("Failed to create new watcher: %v", err)
+	}
+	populateWatcher(configWatcher, a.configPath)
```
stevekuznetsov: Are we ignoring a returned error here?

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+	config, ok := a.configs[info.Basename()]
+	a.lock.Unlock()
+	if !ok {
+		return api.ReleaseBuildConfiguration{}, fmt.Errorf("Could not find config %s", info.Basename())
+	}
+	return config, nil
+}
+
+func (a *agent) GetGeneration() int {
+	return a.generation
```
stevekuznetsov: Why don't we need to lock on this?

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	a.lock.Lock()
+	a.configs = configs
+	a.generation++
+	a.lock.Unlock()
+	log.Info("Configs reloaded")
```
stevekuznetsov: nit: these are usually more useful with a `duration` field

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+		return nil
+	})
+}
+
+func reloadWatcher(ctx context.Context, w *fsnotify.Watcher, c coalescer.Coalescer, path string) {
+	for {
+		select {
+		case <-ctx.Done():
+			w.Close()
```
stevekuznetsov: Does this return an error?

pkg/load/configAgent.go
```diff
@@ -0,0 +1,151 @@
...
+			return
+		case event := <-w.Events:
+			log.Debugf("Received %v event for %s", event.Op, event.Name)
+			// Remove deleted files from watches
+			if event.Op == fsnotify.Remove {
+				log.Debugf("Removing %s from watches", event.Name)
+				w.Remove(event.Name)
```
stevekuznetsov: Can this return an error?

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -14,11 +14,35 @@ trap "rm -rf ${WORKDIR}" EXIT
...
+ci-operator-configresolver -config tests/configs -log-level debug -cycle 2m &
 PID=$!
+# generation should be >= 1. 0 or an error means the config function has not run yet
+if ! timeout 10s bash -c 2>/dev/null -- 'until [[ $(curl http://127.0.0.1:8080/generation 2>/dev/null) -ne 0 ]]; do sleep 1; done'; then
+    echo "Timed out waiting for generation to equal 1"
+    kill $PID
+    wait $PID
+    exit 1
+fi
 if ! timeout 10s bash -c 2>/dev/null -- "until diff tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'); do sleep 1; done" &>/dev/null; then
```
stevekuznetsov: Why does this need to be retried?

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -14,11 +14,35 @@ trap "rm -rf ${WORKDIR}" EXIT
...
     echo "diff: $(diff -ru tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'))"
     kill $PID
+    wait $PID
+    exit 1
+fi
+export currGen=$(curl 'http://127.0.0.1:8080/generation')
+cp tests/configs2/release-4.2/openshift-installer-release-4.2-golang111.yaml tests/configs/release-4.2/openshift-installer-release-4.2.yaml
+if ! timeout 10s bash -c 2>/dev/null -- 'until [[ $(curl http://127.0.0.1:8080/generation 2>/dev/null) -eq $currGen+1 ]]; do sleep 1; done'; then
+    echo "Generation is $(curl http://127.0.0.1:8080/generation)"
+    echo "Timed out waiting for generation to equal $nextGen"
```
stevekuznetsov: What is `$nextGen`?

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -14,11 +14,35 @@ trap "rm -rf ${WORKDIR}" EXIT
...
+    kill $PID
+    wait $PID
+    exit 1
+fi
+if ! timeout 10s bash -c 2>/dev/null -- "until diff tests/expected/openshift-installer-release-4.2-golang111.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'); do sleep 1; done" &>/dev/null; then
```
stevekuznetsov: Why does this need to be retries?

test/ci-operator-configresolver-integration/run.sh
AlexNPavel: An old variable that I meant to change here

pkg/load/configAgent.go
AlexNPavel: I guess? The function signature indicates that it returns an error, but the function only ever returns nil if you look at the function. I'll add a check in case that changes...

pkg/load/configAgent.go
AlexNPavel: Not sure if there is much reason to do so. Should I just add a check in `validateOptions` for the `configresolver` that errors out if the cycle is 0 instead?

stevekuznetsov: Sure, and default to something sensible so we can always run it

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -9,16 +9,43 @@ set -o pipefail
...
-cp -a $ROOTDIR/test/ci-operator-configresolver-integration/ tests
-ci-operator-configresolver -config tests/configs -log-level debug &
+cp -a "$ROOTDIR"/test/ci-operator-configresolver-integration/ tests
+ci-operator-configresolver -config tests/configs -log-level debug -cycle 2m &
 PID=$!
-if ! timeout 10s bash -c 2>/dev/null -- "until diff tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'); do sleep 1; done" &>/dev/null; then
-    echo "diff: $(diff -ru tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2'))"
+# generation should be >= 1. 0 or an error means the config function has not run yet
+# shellcheck disable=SC2016
+if ! timeout 10s bash -c 2>/dev/null -- 'until [[ $(curl http://127.0.0.1:8080/generation 2>/dev/null) -ne 0 ]]; do sleep 1; done'; then
```
stevekuznetsov: nit: unless we care about the time taken to actually run the `curl` we could be a little simpler:

```sh
for (( i = 0; i < 10; i++ )); do
    if [[ "$(curl http://127.0.0.1:8080/generation 2>/dev/null)" -ne 0 ]]; then
        break
    fi
    if [[ "${i}" -eq 10 ]]; then
        echo "[ERROR] Timed out waiting for ci-operator-configresolver to load configuration for the first time"
       exit 1
    fi
done
```

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -9,16 +9,43 @@ set -o pipefail
...
+    echo "Timed out waiting for generation to equal 1"
     kill $PID
+    wait $PID
+    exit 1
+fi
+if ! diff tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2' 2>/dev/null) &>/dev/null; then
```
stevekuznetsov: Can we copy the `diff` formatting from the other tests?

```sh
if ! diff -Naupr "${expected_output_config_dir}" "${generated_output_config_dir}"> "${workdir}/diff"; then
  cat << EOF
ERROR: Incorrect TestGrid config was generated!
ERROR: The following errors were found:

EOF
  cat "${workdir}/diff"
  exit 1
fi
```

test/ci-operator-configresolver-integration/run.sh
```diff
@@ -9,16 +9,43 @@ set -o pipefail
...
+    echo "diff: $(diff -ru tests/expected/openshift-installer-release-4.2.json <(curl 'http://127.0.0.1:8080/config?org=openshift&repo=installer&branch=release-4.2' 2>/dev/null))"
+    kill $PID
+    wait $PID
+    exit 1
+fi
+currGen=$(curl 'http://127.0.0.1:8080/generation')
+export currGen
```
stevekuznetsov: Why do you need to export this?

stevekuznetsov: Oh, because of the child of `timeout`?

AlexNPavel: Yeah. I spent like half an hour trying to get that to work before realizing I needed to export `currGen`.

cmd/ci-operator-configresolver/main.go
```diff
@@ -89,10 +88,10 @@ func resolveConfig(o *options) http.HandlerFunc {
 			Variant: variant,
 		}
 
-		config, ok := o.configs[info.Basename()]
-		if !ok {
+		config, err := agent.GetConfig(info)
+		if err != nil {
 			w.WriteHeader(http.StatusBadRequest)
-			w.Write([]byte("config does not exist"))
+			fmt.Fprintf(w, "failed to get config: %v", err)
```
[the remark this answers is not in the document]
AlexNPavel: That error is more of a bad request issue (like someone typing `opnshift` for the org instead of `openshift`). I'll add a warning log for it just in case there is a problem and we need to debug. The `json.Unmarshal` error on the otherhand I'll make an Error level log.

PR #175, opened 2019-09-22
Staticcheck nits corrections

droslean:

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -57,7 +57,6 @@ func validateOptions(o options) error {
 func missingQuery(w http.ResponseWriter, field string) {
 	w.WriteHeader(http.StatusBadRequest)
 	fmt.Fprintf(w, "%s query missing or incorrect", field)
-	return
 }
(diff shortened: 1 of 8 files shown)


cmd/autoowners/main.go
```diff
@@ -142,14 +142,14 @@ func getOwnersHTTP(orgRepo orgRepo) (httpResult, error) {
...
-					Debug("Not found file in the upstream repo")
-				httpResult.ownersFileExists = false
-				continue
-			} else {
-				return httpResult, err
-			}
+			return httpResult, err
+		}
+
+		if _, ok := err.(*github.FileNotFound); ok {
```
hongkailiu: This would change the logic ...
When the error is with type `FileNotFound` ... we do warning logs only, instead of returning error.
Because we want to keep the caller of `getOwnersHTTP` going for other repos.

During the review, I do think we have a bug in this function.
I will add tests with another PR for this function and ask for your review.

hongkailiu: https://github.com/openshift/ci-tools/pull/176

PR #176, opened 2019-09-23
Fix getOwnersHTTP in autoowners

hongkailiu:
The bug is from the case in which `OWNERS` file DOES exist while
`OWNERS_ALIASES` not. It led to the result of non-existence of
the `OWNERS` file, i.e., `httpResult.ownersFileExists=false`.

The impact of this bug is that we had ignored syncing `OWNERS`
for the repos in the above case.

Test cases have been added to avoid recurrence of this bug in the
feature.

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -136,25 +136,28 @@ func (r httpResult) resolveOwnerAliases() interface{} {
 	}
 }
 
-func getOwnersHTTP(orgRepo orgRepo) (httpResult, error) {
+type FileGetter interface {
+	GetFile(org, repo, filepath, commit string) ([]byte, error)
+}
+
+func getOwnersHTTP(fg FileGetter, orgRepo orgRepo) (httpResult, error) {
 	var httpResult httpResult
 
 	for _, filename := range []string{"OWNERS", "OWNERS_ALIASES"} {
-		data, err := gc.GetFile(orgRepo.Organization, orgRepo.Repository, filename, "")
+		data, err := fg.GetFile(orgRepo.Organization, orgRepo.Repository, filename, "")
 		if err != nil {
 			if _, nf := err.(*github.FileNotFound); nf {
 				logrus.WithField("orgRepo", orgRepo.repoString()).WithField("filename", filename).
 					Debug("Not found file in the upstream repo")
-				httpResult.ownersFileExists = false
 				continue
 			} else {
 				return httpResult, err
 			}
 		}
 
-		httpResult.ownersFileExists = true
 		switch filename {
 		case "OWNERS":
+			httpResult.ownersFileExists = true
 			simple, err := repoowners.LoadSimpleConfig(data)
 			if err != nil {
 				logrus.WithError(err).Error("Unable to load simple config.")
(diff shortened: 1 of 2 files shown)


cmd/autoowners/main_test.go
```diff
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"fmt"
 	"reflect"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/github"
```
droslean: nit: split the test-infra and apimachinery libraries. Add a line between them.

cmd/autoowners/main.go
```diff
@@ -363,14 +362,14 @@ func main() {
 	if err := secretAgent.Start([]string{o.githubToken}); err != nil {
 		logrus.WithError(err).Fatalf("Error starting secrets agent.")
 	}
-	gc = github.NewClient(secretAgent.GetTokenGenerator(o.githubToken), secretAgent.Censor, github.DefaultGraphQLEndpoint, github.DefaultAPIEndpoint)
+	gc := github.NewClient(secretAgent.GetTokenGenerator(o.githubToken), secretAgent.Censor, github.DefaultGraphQLEndpoint, github.DefaultAPIEndpoint)
```
droslean: this comment review is not related to your changes in this PR, but it worths commenting. You don't have to `Start` any secretAgent since this tool is not going to be used as a daemon. Just create a closure that returns the bytes of your key and pass it to the `NewClient`. You can do something similar with the `Censor` too.

cmd/autoowners/main_test.go
hongkailiu: Aren't the rules?

* build-in pkgs
* pkgs from site_1
* pkgs from site_n
* local pkgs

So `apimachinery` and `test-infra` are from `k8s.io` and no need for the line in between?

cmd/autoowners/main.go
hongkailiu: I have to admit this is a copy/paste from autobumper at upstream.
Is `secretAgent`'s task to ensure the reload if secret changes?

Each run is about 30 mins for syncing owners for all the folders.
What about the `secret` changed during the run?

Another reason as you pointed out, is that with agent i get Censor for free.

cmd/autoowners/main_test.go
droslean: We tend to split also the repositories, not only the org. Not so important though.

hongkailiu: Maybe I do not know the rules at all ... :)

Why do we have a line between `apimachinery` and `client-go` but not `api` and `apimachinery`?

https://github.com/openshift/ci-tools/blob/b5e3576894e5b3aab0d8609e38024a5d1977d160/cmd/ci-operator/main.go#L26-L32

cmd/autoowners/main.go
droslean: I might add a minimum secretAgent functionality in upstream's API for cases like this. I didn't also compare the resources that the binary will use with the secretAgent's Start and without.

cmd/autoowners/main_test.go
droslean: in some part of the code, we have it and in another part, we don't. :)

petr-muller: I think sometimes this is a matter of personal preference and that is not harmful. The only solid rule we have is to separate standard library, 3rd party libs and local imports, but sometimes it makes sense to separate a part of 3rd party libs to its separate block, especially when the 3rd party import block is long.

hongkailiu: Make sense.
I will keep those in mind from now on.

PR #178, opened 2019-09-23
Allow ci-operator to build/clone from private repositories 

droslean:
This patch adds a way to inject a private ssh key to ci-operator builds and let it clone from private repositories as well.

- [X] Add a flag that points to ssh key's path
- [X] Refactor secret creation logic
- [X] Inject ssh key secret to `source` step.
    - [X] Update clonerefs dockerfile.
    - [X] Add `--key-files` flag to clonerefs options. 
    - [X] Add test cases that use the ssh secret.
- [X] Inject ssh key secret to `git source` step
- [X] Inject ssh key secret to `rpm injection` step


pkg/steps/source.go
```diff
@@ -37,22 +38,29 @@ const (
...
-ENV GOPATH=%s
-RUN git submodule update --init
-`, api.PipelineImageStream, fromTag, gopath, workingDir, gopath)
+func sourceDockerfile(fromTag api.PipelineImageStreamTagReference, workingDir string, sshSecret *coreapi.Secret) string {
+	var dockerCommands []string
+	dockerCommands = append(dockerCommands, fmt.Sprintf("FROM %s:%s", api.PipelineImageStream, fromTag))
+	dockerCommands = append(dockerCommands, "ADD ./app.binary /clonerefs")
+
+	if sshSecret != nil {
+		dockerCommands = append(dockerCommands, fmt.Sprintf("RUN echo %s > %s", string(sshSecret.Data[coreapi.SSHAuthPrivateKey]), sshKeyPath))
```
bbguimaraes: I'll do a full review later, but this caught my eye while I was skimming the changes: we don't want to record the private key in plain text in the builds/logs. See:

https://docs.okd.io/latest/dev_guide/builds/build_inputs.html#using-secrets-during-build

droslean: Oops, yea you are right. So this will become more complicated. Btw, the link you commend above is not for this case. I already add the source secret for builds, but that Dockerfile is different, because we need to inject that ssh key to the clonerefs container and let `cloneref` to use it. I guess we will need to inject a shared Volume in that build, etc....

cmd/ci-operator/main.go
```diff
@@ -1212,3 +1210,65 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+
+	if err != nil {
+		return fmt.Errorf("could not create secret %s: %v", secret.Name, err)
+	}
+	log.Printf("Created secret %s", secret.Name)
+	return nil
+}
+
+func convertToValidSecretName(secretName string) string {
+	re := regexp.MustCompile(`[^a-z0-9]([^-a-z0-9]*[^a-z0-9])?(\.[^a-z0-9]([^-a-z0-9]*[^a-z0-9])?)*`)
```
stevekuznetsov: We should use upstream validators and mutators for this

stevekuznetsov: Or just reject invalid names at runtime and not try to change anything here

pkg/defaults/defaults.go
```diff
@@ -120,13 +123,13 @@ func FromConfig(
 			if err != nil {
 				return nil, nil, fmt.Errorf("unable to access image stream tag on remote cluster: %v", err)
 			}
-			step = steps.SourceStep(*rawStep.SourceStepConfiguration, config.Resources, buildClient, srcClient, imageClient, artifactDir, jobSpec, dryLogger)
+			step = steps.SourceStep(*rawStep.SourceStepConfiguration, config.Resources, buildClient, srcClient, imageClient, artifactDir, jobSpec, dryLogger, sshSecret)
 		} else if rawStep.ProjectDirectoryImageBuildStepConfiguration != nil {
 			step = steps.ProjectDirectoryImageBuildStep(*rawStep.ProjectDirectoryImageBuildStepConfiguration, config.Resources, buildClient, imageClient, imageClient, artifactDir, jobSpec, dryLogger)
 		} else if rawStep.ProjectDirectoryImageBuildInputs != nil {
-			step = steps.GitSourceStep(*rawStep.ProjectDirectoryImageBuildInputs, config.Resources, buildClient, imageClient, artifactDir, jobSpec, dryLogger)
+			step = steps.GitSourceStep(*rawStep.ProjectDirectoryImageBuildInputs, config.Resources, buildClient, imageClient, artifactDir, jobSpec, dryLogger, getSecretName(sshSecret))
```
stevekuznetsov: I'm not a fan of passing `""` to let the delegate function choose what to do. Why shouldn't the `GitSource` step do the secret creation?

pkg/steps/source.go
stevekuznetsov: No, we should use the secret injection mechanism linked by Bruno to inject the secrets into this build, too. The `clonerefs` options just need to reference it.

pkg/steps/source.go
```diff
@@ -37,22 +38,29 @@ const (
...
+	sshKeyPath = "/sshprivatekey"
 )
 
 var (
 	JobSpecAnnotation = fmt.Sprintf("%s/%s", CiAnnotationPrefix, "job-spec")
 )
 
-func sourceDockerfile(fromTag api.PipelineImageStreamTagReference, workingDir string) string {
-	return fmt.Sprintf(`
-FROM %s:%s
```
stevekuznetsov: Would prefer not to go away from this string formatting

pkg/steps/source.go
```diff
@@ -591,3 +609,17 @@ func trimLabels(labels map[string]string) map[string]string {
 	}
 	return labels
 }
+
+func getSourceSecretFromName(secretName string) *coreapi.LocalObjectReference {
```
stevekuznetsov: I don't really see the utility in having these as functions

pkg/steps/source_test.go
```diff
@@ -90,14 +90,12 @@ func TestCreateBuild(t *testing.T) {
 						ServiceAccount: "builder",
 						Source: buildapi.BuildSource{
 							Type: buildapi.BuildSourceDockerfile,
-							Dockerfile: strP(`
```
stevekuznetsov: Let's not make changes we don't need to make.

cmd/ci-operator/main.go
```diff
@@ -1212,3 +1210,65 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
 	return eventBroadcaster.NewRecorder(
 		templatescheme.Scheme, coreapi.EventSource{Component: namespace})
 }
+
+func getSSHSecretFromPath(sshKeyPath string) (*coreapi.Secret, error) {
+	secret := &coreapi.Secret{Data: make(map[string][]byte)}
+	secret.Type = coreapi.SecretTypeSSHAuth
+	secret.Name = convertToValidSecretName(filepath.Base(sshKeyPath))
+
+	files, err := ioutil.ReadDir(sshKeyPath)
```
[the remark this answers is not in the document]
droslean: yes I noticed that also. I will address it. We should point directly to the ssh-key file.

cmd/ci-operator/main.go
```diff
@@ -1212,3 +1210,65 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+	if err != nil {
+		return nil, fmt.Errorf("could not read dir %s for secret: %v", sshKeyPath, err)
+	}
+
+	for _, f := range files {
```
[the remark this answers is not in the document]
droslean: Yes exactly. I will address this. We should point to the file.

pkg/steps/source.go
droslean: this is a helper function to check to avoid panics when accessing the `secret.Name` when secret is nil.

pkg/steps/source.go
droslean: Yep, I will address this.

cmd/ci-operator/main.go
```diff
@@ -695,30 +705,18 @@ func (o *options) initializeNamespace() error {
...
-			for k, v := range secret.Data {
-				existing.Data[k] = v
-			}
-			if _, err := client.Secrets(o.namespace).Update(existing); err != nil {
-				return fmt.Errorf("could not update secret %s: %v", secret.Name, err)
-			}
-			log.Printf("Updated secret %s", secret.Name)
-			continue
+	if o.sshSecret != nil {
+		if err := createORUpdateSecret(o.namespace, client, o.sshSecret); err != nil {
```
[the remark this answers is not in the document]
droslean: it is not intentional. Thanks, I will correct this

cmd/ci-operator/main.go
```diff
@@ -1212,3 +1210,65 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+		secret.Data[coreapi.SSHAuthPrivateKey], err = ioutil.ReadFile(path)
+		if err != nil {
+			return nil, fmt.Errorf("could not read file %s for secret: %v", path, err)
+		}
+	}
+	return secret, nil
+}
+
+func createORUpdateSecret(namespace string, client *coreclientset.CoreV1Client, secret *coreapi.Secret) error {
+	_, err := client.Secrets(namespace).Create(secret)
```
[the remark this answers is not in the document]
droslean: Nice point. Maybe it's better to create the secret plus the hash from their contents, and we won't have to update anything.

cmd/ci-operator/main.go
droslean: Yea, my point here is to convert any not-allowed char to `-`. However, I will try to check if there is an upstream validator.

pkg/steps/source.go
droslean: yea the difference is only an `\n` both at end and beginning. I will address this and not make changes to the tests too.

pkg/defaults/defaults.go
droslean: the step should care only if there is a secret name. If so, it can inject the `LocalObjectReference` to the build. Although, if the step will do the secret creation, it means that we have to pass the ssh key path and it can be `""` as well if there isn't any.

petr-muller:
Looks good to me in the current form. Have you thought about how do want to test this?

droslean:
> Looks good to me in the current form. Have you thought about how do want to test this?

I will test this by trying to run a job against my private repo, and I will pass my private key to check it. 
I have already tested only the `src-build` if it can clone my private repo, and everything works fine.

However, there are few more things that I want to address, such as adding test cases and maybe include the ssh key to the integration tests if it's necessary.

Another issue is that how can I avoid doing this ssh config hacking in the dockerfile. This is supposed to be done from `clonerefs` but its not. Either I am doing something wrong or there is a bug upstream. However this can be addressed in an follow-up.

petr-muller:
> Another issue is that how can I avoid doing this ssh config hacking in the dockerfile. This is supposed to be done from clonerefs but its not. Either I am doing something wrong or there is a bug upstream. However this can be addressed in an follow-up.

If clonerefs is supposed to handle the case but doesn't let's fix it upstream. This is not urgent, we can afford to fix it properly.

droslean:
> If clonerefs is supposed to handle the case but doesn't let's fix it upstream. This is not urgent, we can afford to fix it properly.

I found a way how to tell `clonerefs` to add `github.com` to the `known_hosts` file, but I don't trust it since we need to hardcode some IP addresses and it turns out that is more code than the approach I already did.

pkg/steps/source.go
```diff
@@ -38,22 +38,40 @@ const (
 
 	ProwJobIdLabel = "prow.k8s.io/id"
 
-	gopath = "/go"
+	gopath           = "/go"
+	sshPrivateKey    = "/sshprivatekey"
+	githubSshKeyScan = "github.com,140.82.118.4,140.82.113.3,192.30.253.113 ssh-rsa AAAAB3NzaC1yc2EAAAABIwAAAQEAq2A7hRGmdnm9tUDbO9IDSwBK6TbQa+PXYPCPy6rbTrTtw7PHkccKrpp0yVhp5HdEIcKr6pLlVDBfOLX9QUsyCOV0wzfjIJNlGEYsdlLJizHhbn2mUjvSAHQqZETYP81eFzLQNnPHt4EVVUh7VfDESU84KezmD5QlWpXLmvU31/yMf+Se8xhHTvKSCZIFImWwoG6mbUoWf9nzpIoaSjB+weqqUUmpaaasXVal72J+UX2B+2RPW3RcT0eOzQgqlJL3RKrTJvdsjE3JEAvGq3lGHSZXy28G3skua2SmVi/w4yCE6gbODqnTWlg7+wC604ydGXA8VJiS5ap43JXiUFFAaQ=="
```
stevekuznetsov: This needs to be a configuration for `plank` and should exist outside of this PR

cmd/ci-operator/main.go
```diff
@@ -1212,3 +1228,34 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, namespace string) rec
...
+	secret.Name = fmt.Sprintf("ssh-%s", getHashFromBytes(b))
+
+	// Secret.Data["ssh-privatekey"] is required on SecretTypeSSHAuth type.
+	// https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5466-L5470
+	secret.Data[coreapi.SSHAuthPrivateKey] = b
+
+	return secret, nil
+}
+
+func getHashFromBytes(b []byte) string {
```
stevekuznetsov: In general I do not like these methods. They are used once and only once, they are not broken out for testing and they break up the flow of the original code. 

We should inline them into the original method unless we are very specifically breaking them out to test them. In both cases -- `getHashFromBytes` and `getSecretName` -- I do not think that the logic inside requires tests.

droslean: Yea, it makes sense. I will address that.

pkg/steps/source.go
droslean: opened [openshift/release/pull/5177](https://github.com/openshift/release/pull/5177)

pkg/steps/source.go
droslean: I don't want to pass empty LocalObjectReference, otherwise, I will have to update all the integration tests adding the empty object. I would prefer having this helper function to pass nil when there is no secret

pkg/steps/source.go
```diff
@@ -37,22 +38,40 @@ const (
...
+		dockerCommands = append(dockerCommands, fmt.Sprintf("COPY ./%s %s", coreapi.SSHAuthPrivateKey, sshPrivateKey))
+	}
+
+	dockerCommands = append(dockerCommands, fmt.Sprintf("RUN umask 0002 && /clonerefs && find %s/src -type d -not -perm -0775 | xargs -r chmod g+xw", gopath))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("WORKDIR %s/", workingDir))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("ENV GOPATH=%s", gopath))
+	dockerCommands = append(dockerCommands, "RUN git submodule update --init")
+
+	// After the clonerefs command, we don't need the secret anymore.
+	// We don't want to let the key keep existing in the image's layer.
```
petr-muller: Isn't the key present in an intermediate layer though? If so, is it a problem (are we pushing that image somewhere publicly available)?

droslean: All the build inputs are deleted and are not are available while building. Therefore we need to `ADD` the key somewhere to be used from `clonerefs`. If we don't delete it, the keys will be available in the image layer and anyone can access it if the image will be pulled.

droslean: > Users should normally remove their input secrets from the final application image so that the secrets are not present in the container running from that image. However, the secrets will still exist in the image itself in the layer where they were added. This removal should be part of the Dockerfile itself.

ref: https://docs.openshift.com/container-platform/3.5/dev_guide/builds/build_inputs.html#using-secrets-docker-strategy

petr-muller: I'm not saying it should not be deleted. I'm wondering if deleting it from the outermost layer is actually enough to protect it: the bits are still present in an intermediate layer. Unless I'm missing some OpenShift build smartness, but this was a gotcha with Docker.

droslean: The initial secret will be injecting as a build input. After the build, the bytes won't be available. I guess this is some Openshift builds magic....

petr-muller: I want to to test this a bit: this is sensitive area, let's be 100% sure. Also @bbguimaraes mentioned in Slack something that supports this should be fine.

pkg/steps/source.go
stevekuznetsov: You only use it in one place, though, so why not just do the `if` logic there? That's what I mean.

droslean: I am using it in `source` `git_source` and `rpm_injection`. That's why I made this a function.  However, If you think that it's ok, I could put the `if` logic to those places.

pkg/steps/source.go
stevekuznetsov: Let's ask DevEx team on this please

pkg/steps/source.go
stevekuznetsov: No, if we're reusing then we should keep it

pkg/steps/source.go
```diff
@@ -109,38 +129,41 @@ func (s *sourceStep) Run(ctx context.Context, dry bool) error {
...
-	return handleBuild(s.buildClient, createBuild(s.config, s.jobSpec, clonerefsRef, s.resources), dry, s.artifactDir, s.dryLogger)
+	return handleBuild(s.buildClient, createBuild(s.config, s.jobSpec, clonerefsRef, s.resources, s.sshSecretName), dry, s.artifactDir, s.dryLogger)
 }
 
-func createBuild(config api.SourceStepConfiguration, jobSpec *api.JobSpec, clonerefsRef coreapi.ObjectReference, resources api.ResourceConfiguration) *buildapi.Build {
+func createBuild(config api.SourceStepConfiguration, jobSpec *api.JobSpec, clonerefsRef coreapi.ObjectReference, resources api.ResourceConfiguration, sshSecretName string) *buildapi.Build {
 	var refs []prowapi.Refs
 	if jobSpec.Refs != nil {
+		if len(sshSecretName) > 0 {
+			jobSpec.Refs.CloneURI = fmt.Sprintf("ssh://git@github.com/%s/%s.git", jobSpec.Refs.Org, jobSpec.Refs.Repo)
```
bbguimaraes: Mutating the JobSpec this deep into the execution stack is very surprising. Is it necessary or could we just copy `Refs` and `ExtraRefs`? If it is, we should make the change in `main.go`.

pkg/defaults/defaults.go
```diff
@@ -120,13 +121,13 @@ func FromConfig(
...
 		} else if rawStep.ProjectDirectoryImageBuildStepConfiguration != nil {
 			step = steps.ProjectDirectoryImageBuildStep(*rawStep.ProjectDirectoryImageBuildStepConfiguration, config.Resources, buildClient, imageClient, imageClient, artifactDir, jobSpec, dryLogger)
 		} else if rawStep.ProjectDirectoryImageBuildInputs != nil {
-			step = steps.GitSourceStep(*rawStep.ProjectDirectoryImageBuildInputs, config.Resources, buildClient, imageClient, artifactDir, jobSpec, dryLogger)
+			step = steps.GitSourceStep(*rawStep.ProjectDirectoryImageBuildInputs, config.Resources, buildClient, imageClient, artifactDir, jobSpec, dryLogger, sshSecretName)
 		} else if rawStep.RPMImageInjectionStepConfiguration != nil {
-			step = steps.RPMImageInjectionStep(*rawStep.RPMImageInjectionStepConfiguration, config.Resources, buildClient, routeGetter, imageClient, artifactDir, jobSpec, dryLogger)
+			step = steps.RPMImageInjectionStep(*rawStep.RPMImageInjectionStepConfiguration, config.Resources, buildClient, routeGetter, imageClient, artifactDir, jobSpec, dryLogger, sshSecretName)
```
bbguimaraes: Can `GitSourceStep` and `RPMImageInjectionStep` ever use the key?

droslean: Yes. For example `build_root.project_image`, it will need to access a `Dockerfile` in the private repo.

bbguimaraes: Then we need to change this as well: https://github.com/openshift/ci-tools/pull/178/files#diff-abdbc02ab75ad52d862e0625802eb296L35.

pkg/steps/source.go
droslean: Yea we shouldn't mutate jobspec in this level. I addressed that.

pkg/defaults/defaults.go
droslean: addressed!

bbguimaraes: We're still changing `RPMImageInjectionStep`, that one [can't possibly use the secret](https://github.com/openshift/ci-tools/blob/6902e9ce6a95261f639739672c4b3763065246ea/pkg/steps/rpm_injection.go#L15-L18).

droslean: on second look, I don't even think that we have to add any secret to the `RPMImageInjectionStep`

droslean: i removed the commit.

PR #181, opened 2019-09-23
Grab base config from one commit back

stevekuznetsov:
When a postsubmit is triggered from a merge, `SHA^1` will refer to the
previous state of the branch before the merge, regardless of how many
commits have been merged in one action. Therefore, this is a safe way
to grab the previous state of the repository. This will work unless
someone manually pushes commits to the repository, which shouldn't
happen, so this is a reasonably resilient, albeit best-effort approach.

/assign @petr-muller @dros

### DIFF
--- a/cmd/config-change-trigger/main.go
+++ b/cmd/config-change-trigger/main.go
@@ -92,7 +92,7 @@ func main() {
 	}
 
 	prConfig := config.GetAllConfigs(o.releaseRepoPath, logger)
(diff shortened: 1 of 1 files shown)


bbguimaraes:
I'm surprised Prow doesn't expose that information from the push event. It's Github-specific, but we could parse `base_link`:

```yaml
  refs:
    base_link: https://github.com/openshift/release/compare/c7cbc0a93dbd...d75ce7822649
```

PR #186, opened 2019-09-25
Stop setting redundant flags

hongkailiu:
Required by https://github.com/openshift/release/pull/5171


hongkailiu:
> We could do this but I was hoping to centralize this logic upstream so we don't copy it in every binary :)

Do you mean when NewGithubClient ... and the client does the init and save the user data as a field?

stevekuznetsov:
Yes. I'm working on it, you don't need to worry about it for now.

hongkailiu:
> Yes. I'm working on it, you don't need to worry about it for now.

ACK.

PR #187, opened 2019-09-27
ci-operator-configresolver: add metrics

AlexNPavel:
Adds 4 metrics to configresolver:
- HTTP response size
- Response time
- Error rate (sorted by error)
- Config reload time

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -6,10 +6,13 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	log "github.com/sirupsen/logrus"
 	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/pjutil"
 
 	"github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/load"
@@ -29,6 +32,46 @@ type options struct {
 	gracePeriod time.Duration
 	cycle       time.Duration
 }
+
+var (
+	configresolverMetrics = struct {
+		httpRequestDuration *prometheus.HistogramVec
+		httpResponseSize    *prometheus.HistogramVec
+		errorRate           *prometheus.CounterVec
+		configReloadTime    prometheus.Histogram
+	}{
+		httpRequestDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    "configresolver_http_request_duration_seconds",
+				Help:    "http request duration in seconds",
+				Buckets: []float64{0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 20},
+			},
+			[]string{"status"},
+		),
+		httpResponseSize: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    "configresolver_http_response_size_bytes",
+				Help:    "http response size in bytes",
+				Buckets: []float64{16384, 32768, 65536, 131072, 262144, 524288, 1048576, 2097152, 4194304, 8388608, 16777216, 33554432},
+			},
+			[]string{"status"},
+		),
+		errorRate: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "configresolver_error_rate",
+				Help: "number of errors, sorted by label/type",
+			},
+			[]string{"error"},
+		),
+		configReloadTime: prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "configresolver_config_reload_duration_milliseconds",
+				Help:    "config reload duration in milliseconds",
+				Buckets: []float64{0.1, 0.2, 0.3, 0.4, 0.5, 0.75, 1, 2.5, 5, 7.5, 10},
+			},
+		),
+	}
+)
 
 func gatherOptions() options {
 	o := options{}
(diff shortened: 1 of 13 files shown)


cmd/ci-operator-configresolver/main.go
```diff
@@ -53,31 +96,48 @@ func validateOptions(o options) error {
...
+	return http.StatusBadRequest, []byte(fmt.Sprintf("%s query missing or incorrect", field))
 }
 
 func resolveConfig(agent load.ConfigAgent) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
+		t := time.Now()
+		// get int pointer
+		statusCode := new(int)
+		var body []byte
+		defer func() {
```
stevekuznetsov: I think I'd prefer chaining/nesting two `http.HandlerFunc` so that each is simple. Let's look at `traceHandler` in `deck` upstream.

cmd/ci-operator-configresolver/main.go
```diff
@@ -120,14 +191,16 @@ func main() {
...
-	configAgent, err := load.NewConfigAgent(o.configPath, o.cycle)
+	configAgent, err := load.NewConfigAgent(o.configPath, o.cycle, configresolverMetrics.errorRate, &configresolverMetrics.configReloadTime)
 	if err != nil {
 		log.Fatalf("Failed to get config agent: %v", err)
 	}
 
 	http.HandleFunc("/config", resolveConfig(configAgent))
 	http.HandleFunc("/generation", getGeneration(configAgent))
 	interrupts.ListenAndServe(&http.Server{Addr: o.address}, o.gracePeriod)
+	health.ServeReady()
```
stevekuznetsov: Are we serving metrics? I would have expected a call to `metrics.ExposeMetrics`

pkg/load/configAgent.go
```diff
@@ -112,7 +122,9 @@ func (a *agent) loadFilenameToConfig() error {
 	a.configs = configs
 	a.generation++
 	a.lock.Unlock()
-	log.WithField("duration", time.Since(startTime)).Info("Configs reloaded")
+	duration := time.Since(startTime)
+	(*a.reloadMetrics).Observe(float64(duration.Nanoseconds() / 1e6))
```
stevekuznetsov: Why do you need to dereference the pointer?

pkg/load/configAgent.go
```diff
@@ -145,16 +157,19 @@ func reloadWatcher(ctx context.Context, w *fsnotify.Watcher, c coalescer.Coalesc
...
 			}
 			go c.Run()
 			// add new files to be watched; if a watch already exists on a file, the
 			// watch is simply updated
 			if err := populateWatcher(w, path); err != nil {
+				a.recordError("failed to update watcher")
 				log.WithError(err).Error("Failed to update fsnotify watchlist")
 			}
 		case err := <-w.Errors:
+			a.recordError(fmt.Sprintf("received fsnotify error: %v", err))
```
stevekuznetsov: We should avoid label values that change -- it's easy to overdo the cardinality in Prometheus and crash it.

pkg/load/configAgent.go
AlexNPavel: That method only works on the non-dereferenced value. Otherwise: `type *prometheus.Histogram has no field or method Observe`

cmd/ci-operator-configresolver/main.go
```diff
@@ -90,14 +171,18 @@ func resolveConfig(agent load.ConfigAgent) http.HandlerFunc {
...
 		if err != nil {
+			recordError("config not found")
 			w.WriteHeader(http.StatusBadRequest)
 			fmt.Fprintf(w, "failed to get config: %v", err)
+			log.WithError(err).Warning("failed to get config")
 			return
 		}
 		jsonConfig, err := json.MarshalIndent(config, "", "  ")
 		if err != nil {
+			recordError("failed to unmarshal config")
```
petr-muller: unmarshal -> marshal

petr-muller: unmarshal -> marshal

cmd/ci-operator-configresolver/main.go
```diff
@@ -90,14 +171,18 @@ func resolveConfig(agent load.ConfigAgent) http.HandlerFunc {
...
 			w.WriteHeader(http.StatusInternalServerError)
 			fmt.Fprintf(w, "failed to marshal config to JSON: %v", err)
+			log.WithError(err).Errorf("failed to unmarshal config to JSON")
```
petr-muller: unmarshal -> marshal

pkg/load/configAgent.go
```diff
@@ -61,11 +64,16 @@ func NewConfigAgent(configPath string, cycle time.Duration) (ConfigAgent, error)
 		return nil, fmt.Errorf("Failed to populate watcher: %v", err)
 	}
 	interrupts.Run(func(ctx context.Context) {
-		reloadWatcher(ctx, configWatcher, configCoalescer, a.configPath)
+		reloadWatcher(ctx, configWatcher, a, configCoalescer, a.configPath)
```
petr-muller: Does it make sense to still pass `a.configPath` separately now when full `agent` needs to passed in?

pkg/load/configAgent.go
```diff
@@ -94,10 +102,12 @@ func (a *agent) loadFilenameToConfig() error {
 		if info != nil && !info.IsDir() && (ext == ".yml" || ext == ".yaml") {
 			configSpec, err := Config(path)
 			if err != nil {
+				a.recordError("failed to load config")
 				return fmt.Errorf("failed to load ci-operator config (%v)", err)
 			}
 
 			if err := configSpec.ValidateAtRuntime(); err != nil {
+				a.recordError("invalid config")
```
petr-muller: "invalid ci-operator config" would carry some context and is reasonably future-proof

cmd/ci-operator-configresolver/main.go
```diff
@@ -30,6 +35,63 @@ type options struct {
...
+		errorRate: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "configresolver_error_rate",
+				Help: "number of errors, sorted by label/type",
+			},
+			[]string{"error"},
+		),
+		configReloadTime: prometheus.NewHistogram(
+			prometheus.HistogramOpts{
+				Name:    "configresolver_config_reload_duration_milliseconds",
```
stevekuznetsov: nit: would be nice to be consistent and have this in seconds, too

pkg/load/configAgent.go
stevekuznetsov: I'm not sure this will do what you want. Why not define the metrics you want in this package as `var` and do the registration as normal? Instead of trying to pass them through

AlexNPavel: It now declares the `reloadMetrics` in the `configAgent`'s file. It still passes in the `errorRate` metric, since both the `main.go` and `configAgent.go` use it (and it's a pointer, so it doesn't have the same problem)

PR #188, opened 2019-09-30
ci-operator multi-stage: add cluster profile

bbguimaraes:
When the test has a non-empty profile:
- check that a secret with a pre-defined name exists
- mount the contents of that secret in every pod
- assume a `kubeconfig` key will be added to the test's secret and set
  `$KUBECONFIG` to point to it in every pod

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -19,14 +19,16 @@ import (
 )
 
 const (
-	multiStageTestLabel = "ci.openshift.io/multi-stage-test"
-	secretMountPath     = "/var/run/secrets/ci.openshift.io/multi-stage"
+	multiStageTestLabel     = "ci.openshift.io/multi-stage-test"
+	clusterProfileMountPath = "/var/run/secrets/ci.openshift.io/cluster-profile"
+	secretMountPath         = "/var/run/secrets/ci.openshift.io/multi-stage"
 )
 
 type multiStageTestStep struct {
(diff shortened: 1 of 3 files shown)


pkg/steps/multi_stage.go
```diff
@@ -231,6 +251,27 @@ func addSecret(secret string, pod *coreapi.Pod) {
...
+		},
+	})
+	container := &pod.Spec.Containers[0]
+	container.VolumeMounts = append(container.VolumeMounts, coreapi.VolumeMount{
+		Name:      volumeName,
+		MountPath: clusterProfileMountPath,
+	})
+	container.Env = append(container.Env, coreapi.EnvVar{
+		Name:  "CLUSTER_TYPE",
+		Value: strings.Split(string(profile), "-")[0],
```
petr-muller: Ouch. Is this (rather arbitrary) dependency of cluster type on a pattern of the cluster profile name documented somewhere? I'm almost tempted to ask if we should have a `(*api.ClusterProfile).toClusterType()` for this to avoid burying this dependency here.

bbguimaraes: I don't know if we'll gain much in terms of overall maintenance, but I like the idea of having that in `api`. Added.

petr-muller: I imagined some poor soul creating a `aws_awesome` cluster profile and the following debugging session resulting in discovering this line :)

PR #191, opened 2019-10-02
testgrid-config-generator: Support more informing jobs

smarterclayton:
The doc for release informing jobs didn't make it clear that
we allow jobs from periodics that aren't part of release config.

Add support to the generator to read periodics that have the label
`ci.openshift.io/release-type=informing` and add them to dashboards
based on the existing `job-release` label (already in use by other
components like ci-chat-bot and release-controller). These become
part of the appropriate informing dashboard.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -12,7 +12,9 @@ import (
 	"sort"
 
 	"github.com/GoogleCloudPlatform/testgrid/config"
+	jc "github.com/openshift/ci-tools/pkg/jobconfig"
 	"github.com/sirupsen/logrus"
+	prowconfig "k8s.io/test-infra/prow/config"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 	"sigs.k8s.io/yaml"
@@ -21,9 +23,13 @@ import (
 type options struct {
 	releaseConfigDir  string
 	testGridConfigDir string
+	prowJobConfigDir  string
 }
 
 func (o *options) Validate() error {
+	if o.prowJobConfigDir == "" {
+		return errors.New("--prow-jobs-dir is required")
+	}
 	if o.releaseConfigDir == "" {
 		return errors.New("--release-config is required")
 	}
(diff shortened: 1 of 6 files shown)


smarterclayton:
https://github.com/openshift/release/pull/5266 enables these

/assign @stevekuznetsov

cmd/testgrid-config-generator/main.go
```diff
@@ -128,6 +141,17 @@ func main() {
 		logrus.Fatalf("Invalid options: %v", err)
 	}
 
+	informingPostSubmits := make(map[string]prowconfig.Periodic)
```
stevekuznetsov: `informingPeriodics`?

cmd/testgrid-config-generator/main.go
```diff
@@ -171,6 +195,11 @@ func main() {
 				blocking.add(job.ProwJob.Name)
 			}
 		}
+		for _, p := range informingPostSubmits {
+			if p.Labels["job-release"] == version {
```
stevekuznetsov: We likely want a presubmit in `release` that enforces that you set this if you set it to be informing, and maybe also if you commit to release-controller config that you set the infroming/blocking label, too?

pkg/jobconfig/files.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/openshift/ci-tools/pkg/promotion"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
stevekuznetsov: :(

smarterclayton:
Added integration test, no place to add unit tests and the integration is pretty accurate.

cmd/testgrid-config-generator/main.go
smarterclayton: Agree

smarterclayton:
Feedback addressed, tagging so we can review upgrade health before GA

PR #192, opened 2019-10-02
Collapse job configuration configmaps for master branch

stevekuznetsov:
Now that we zip all of our files, space constraints are not relevant any
longer and we can just shard simply by branch.

/assign @hongkailiu

### DIFF
--- a/pkg/jobconfig/files.go
+++ b/pkg/jobconfig/files.go
@@ -53,10 +53,6 @@ func (i *Info) ConfigMapName() string {
 	if i.Type == "periodics" && i.Branch == "" {
 		return fmt.Sprintf("job-config-%s", promotion.FlavorForBranch(""))
 	}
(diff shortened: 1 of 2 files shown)


petr-muller:
Huh how is this not a breaking change?

hongkailiu:
/lgtm

/hold

@petr-muller Follow up on breaking change job?

petr-muller:
> Follow up on breaking change job?
@hongkailiu I was expecting this to change the generated jobs, but I did not realize this only collapses *job* config shards (which are *not* used in job definitions), not ci-operator config shards (which are).

Job config shards are AFAIK only used by Prow components, so as long as we are sure we mount this shard for them and also fix config-updater [here](https://github.com/openshift/release/blob/master/core-services/prow/02_config/_plugins.yaml#L1221) this should be fine.

PR #193, opened 2019-10-03
testgrid-config-generator: Correct some faulty assumptions

smarterclayton:
The publish steps inside the release config are not good indicators of ocp or okd. Instead, the naming conventions chosen by the release should be used.  `-openshift-origin-` means `okd` in a job, and `X.Y.Z-0.ci` for release name also means `okd`.  The okd release streams are downstream of CI and can be ignored for now.

Fix a bug where the periodics were added to both dashboards by checking for the correct convention. Also revert the run.sh script to be correct and add a generate mode.

/assign @stevekuznetsov

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -9,7 +9,9 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	"sort"
+	"strings"
 
 	"github.com/GoogleCloudPlatform/testgrid/config"
 	jc "github.com/openshift/ci-tools/pkg/jobconfig"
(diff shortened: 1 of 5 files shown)


cmd/testgrid-config-generator/main.go
```diff
@@ -170,21 +163,27 @@ func main() {
...
-			version = releaseConfig.Publish.Tag.TagRef.Name
-		} else {
-			logrus.Infof("could not determine publish destination for config at %v", path)
+		case strings.HasSuffix(releaseConfig.Name, ".ci"):
+			product = "okd"
+		default:
+			logrus.Infof("release is not recognized: %s", releaseConfig.Name)
 			return nil
 		}
+		m := reVersion.FindStringSubmatch(releaseConfig.Name)
```
stevekuznetsov: Do we have to? Ugh.

smarterclayton: Yeah, the publish tags aren't guaranteed to match releases, names are

smarterclayton:
Tag publishing doesn't have anything to do with release versioning - we have examples where we would publish to 4.3 from other places, and OKD is just accidental.  Publishing is a pipeline thing, not identifying what release this is.

Names define what releases things are.

smarterclayton:
The regex is less fragile than the publishing thing, because the publishing thing should be able to change at will to target other locations.

stevekuznetsov:
yuck/10

/lgtm
/approve

PR #204, opened 2019-10-10
Lease client

bbguimaraes:
POC implementation of a lease client and its use by multi-stage tests.

TODO:

- [x] `cmd/ci-operator/main.go:391`: set `owner` properly.  Is
  `$JOB_NAME_SAFE`, used by the template implementation, enough?)
- [x] `pkg/steps/lease.go:41`: should — and if so, how — errors (communication
  failures, the lease being changed by external processes, etc.) when updating
  the leases be treated?
- [x] polish once an initial design is agreed upon

```
$ git diff --shortstat master -- `find ! \( -name vendor -prune \) -type f ! -name '*_test.go'`
 7 files changed, 215 insertions(+), 4 deletions(-)
```


cmd/ci-operator/main.go
```diff
@@ -236,6 +239,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.BoolVar(&opt.verbose, "v", false, "Show verbose output.")
 
 	// what we will run
+	flag.StringVar(&opt.leaseServer, "lease-server", "", "")
```
stevekuznetsov: What is this? Add doc

cmd/ci-operator/main.go
```diff
@@ -398,13 +406,30 @@ func (o *options) Run() error {
...
-	buildSteps, postSteps, err := defaults.FromConfig(o.configSpec, o.jobSpec, o.templates, o.writeParams, o.artifactDir, o.promote, o.clusterConfig, o.targets.values, dryLogger, secretName)
+	buildSteps, postSteps, err := defaults.FromConfig(o.configSpec, o.jobSpec, o.templates, o.writeParams, o.artifactDir, o.promote, o.clusterConfig, o.leaseClient, o.targets.values, dryLogger, secretName)
 	if err != nil {
 		return fmt.Errorf("failed to generate steps from config: %v", err)
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
-
+	if o.leaseClient != nil {
+		defer func() {
```
stevekuznetsov: Isn't there some crazy on-exit code in ci-operator? We should use that

cmd/ci-operator/main.go
```diff
@@ -398,13 +406,30 @@ func (o *options) Run() error {
...
+			}
+		}()
+		go func() {
+			t := time.NewTicker(30 * time.Second)
+			for {
+				select {
+				case <-ctx.Done():
+					t.Stop()
+				case <-t.C:
+					o.leaseClient.Heartbeat()
```
stevekuznetsov: I thought there was a blocking acquire call that we could use?

pkg/steps/lease.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (s *leaseStep) Name() string                               { return s.wrapped.Name() }
+func (s *leaseStep) Description() string                        { return s.wrapped.Description() }
+func (s *leaseStep) Requires() []api.StepLink                   { return s.wrapped.Requires() }
+func (s *leaseStep) Creates() []api.StepLink                    { return s.wrapped.Creates() }
+func (s *leaseStep) Provides() (api.ParameterMap, api.StepLink) { return s.wrapped.Provides() }
+
+func (s *leaseStep) Run(ctx context.Context, dry bool) error {
+	log.Printf("Acquiring lease for %q", s.leaseType)
+	// TODO handle unexpected lease release?
```
stevekuznetsov: in what sense

cmd/ci-operator/main.go
bbguimaraes: See item 3 in the TODO list =)

cmd/ci-operator/main.go
bbguimaraes: This is a safety net in case the resources are not released normally (see `leaseStep.Run` and `leaseStep.release`). Not a good one, and arguably unnecessary, because the lease has timeouts anyway, but at least we get a message in the log.

In successful and unsuccessful executions, `defer` is sufficient.

I don't think `ci-operator`'s handling of interrupted executions allows us to do better (I had to discard some design choices because of that), see:

https://github.com/openshift/ci-tools/blob/d6ac3d9f96a4945c7c34381bc2cc1d6bc032ccdc/cmd/ci-operator/main.go#L408-L416

In case of an interrupt, all lease steps will execute a `release`, so we would be racing with them here (for 10 seconds, before `handler`'s `os.Exit(1)`).

cmd/ci-operator/main.go
bbguimaraes: This is only responsible for updating existing leases. The acquire process is done by `lease.Client.Acquire`.

pkg/steps/lease.go
bbguimaraes: Just a paranoid check, see item 2 in the TODO list.

bbguimaraes: In this particular place in the code: what should happen to the test if we cannot (for all possible reasons) update the lease?

pkg/lease/client.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+func (c *client) Heartbeat() error          { return c.boskos.UpdateAll(leasedState) }
+func (c *client) Release(name string) error { return c.boskos.ReleaseOne(name, freeState) }
+
+func (c *client) ReleaseAll() error {
+	if !c.boskos.HasResource() {
+		return nil
+	}
+	log.Println("error: Not all leases were released, releasing them now")
```
petr-muller: I don't think this is the proper place to log this error - if we can easily do `ReleaseAll`, why complain about it in library code? If the caller sets a rule for itself that each rule needs to be individually released, why not, but I don't see a reason why the lib should do that.

droslean: Yea, let's return the error here.

pkg/lease/fake.go
```diff
@@ -0,0 +1,34 @@
...
+	call := "acquire " + rtype
+	c.Calls = append(c.Calls, call)
+	if c.Failures.Has(call) {
+		return "", fmt.Errorf("injected failure %q", call)
+	}
+	return rtype + strconv.Itoa(len(c.Calls)-1), nil
+}
+
+func (c *FakeClient) Release(name string) error {
+	call := "release " + name
```
droslean: appending the name with a space between them doesn't seem right? Is this intentional?

pkg/lease/client.go
bbguimaraes: So are you proposing to also expose `HasResource` to allow the caller to do the check?

Would it be better if the method had a name that made it clearer that we are checking for leaks?

pkg/lease/fake.go
bbguimaraes: See `TestError` and `TestAcquireRelease` in `pkg/steps/lease_test.go`.

bbguimaraes:
One clarification since there seems to be some confusion about the purpose of `lease.Client.ReleaseAll`: there are three possible cases.

1. A bug-free execution. Lease steps acquire and release leases as they start and finish. `ReleaseAll` is a no-op.
2. Steps for whatever reason don't release their leases. *At the end of the execution*, `ReleaseAll` warns and releases them.
3. An interrupted execution. The `context` held by the lease steps is cancelled, they try to release their leases in the 10s period `ci-operator` gives us. `ReleaseAll` is not called.

stevekuznetsov:
- `$JOB_NAME_SAFE` should be a good choice
 - dropping over dead when something unexpected happens with the lease seems good to me

bbguimaraes:
I could not find a good reference for the purpose of the `owner` field, but
looking at code my guess is:

1.  Prevent owners from affecting resources that are not theirs.  Doesn't seem
    very useful since we update and release by name.  `$JOB_NAME_SAFE` would
    not help either, because it is not unique across jobs.
2.  Aggregate metrics.  This depends on how we want to group them.

One annoyance is that the client has an associated owner, so we would have to
create one instance for each test we are executing.

pkg/lease/client.go
petr-muller: > So are you proposing to also expose `HasResource` to allow the caller to do the check?

Depends on that is the API's contract: if it imposes "all leases need to eventually be explicitly released", then no, but that contract also seems quite strong for library that has easy `ReleaseAll` and so can't really leak.

> Would it be better if the method had a name that made it clearer that we are checking for leaks?

I think this would be best: make it more clear that releases should be explicit and `ReleaseAll` is really the last resort call that is usually not expected to really find anything not released. Maybe drop the severity of the message to "warn".

bbguimaraes:
Updated, will squash after review.  I had to rebase to make Prow happy, new
commits are:

- e0cf5dfd
  - small fixes
  - owner in Boskos is `namespace + job name hash`
  - use fake client in dry-run mode
- 9ed4e809
  - pass a cancellation function when a lease is acquired to allow the client
    to cancel the test if the update fails
  - release in `main.go` when interrupted to avoid needing an extra go routine
    for each lease step
  - because there is now list of active leases, the client's `ReleaseAll`
    method can return that list so the check for leaks happens on the caller
    side
- 2d5d7b48
  - add a `boskosClient` interface and test the client itself

bbguimaraes:
*Sigh*.

/me downloads go 1.13

pkg/lease/client.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+type client struct {
+	sync.RWMutex
+	boskos boskosClient
+	cancel map[string]context.CancelFunc
+}
+
+func (c *client) Acquire(rtype string, cancel context.CancelFunc) (string, error) {
+	c.Lock()
```
stevekuznetsov: technically you only need to lock around `c.cancel` update, right?

bbguimaraes: I missed that after the transition from `ReleaseAll` to `ReleaseOne`. Changed `Release` as well.

Also fixed `Heartbeat` to use a write lock since it now also mutates the client.

bbguimaraes:
More conflicts in `vendor/`, rebased again. Only 9ae5c97 is new.

bbguimaraes:
Cleaned up the history, this is ready for a final review.

pkg/lease/client.go
```diff
@@ -0,0 +1,99 @@
...
+	return &client{
+		boskos: boskos,
+		cancel: make(map[string]context.CancelFunc),
+	}
+}
+
+type client struct {
+	sync.RWMutex
+	boskos boskosClient
+	cancel map[string]context.CancelFunc
```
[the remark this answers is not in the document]
bbguimaraes: Added comments to all public identifiers.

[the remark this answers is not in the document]
bbguimaraes: For some reason, Github is showing an outdated version of the code, is this comment (added last week) enough?

https://github.com/openshift/ci-tools/pull/204/files#diff-594c5d067525ac389bcbca2d8795a9cbR57-R60

PR #207, opened 2019-10-10
ci-operator: implement retrieving config from configresolver

AlexNPavel:
This allows us to pass in the configresolver's address and the org-repo-branch + variant of a project config in addition to a config file. This currently still requires passing a config via env var or file and the config pulled from the configresolver will only be checked against the normal config to make sure that values match. If the values don't match, the error will be printed to the log, but will not be fatal. This will allow us to test this in production without breaking everything. If it doesn't have any errors, we can start using the configresolver as the main way of getting configs.

Full migration to the configresolver will require an update to prowgen as well to add the flags required for the ci-operator command.


pkg/load/load.go
```diff
@@ -32,6 +47,48 @@ func Config(path string) (*api.ReleaseBuildConfiguration, error) {
...
+			logrus.WithError(err).Error("Failed to read configresolver response body")
+			return configSpec, nil
+		}
+		configSpecHTTP := &api.ReleaseBuildConfiguration{}
+		err = json.Unmarshal(data, configSpecHTTP)
+		if err != nil {
+			logrus.Errorf("failed to unmarshal config from configresolver: invalid configuration: %v\nvalue:\n%s", err, string(data))
+			return configSpec, nil
+		}
+		if !reflect.DeepEqual(configSpec, configSpecHTTP) {
```
petr-muller: Just a note: If we enable this in production, we'll see this failing for rehearsals running over a changed config (where a "production" resolver will have the master config, but the job itself will get the changed one).

pkg/load/load.go
```diff
@@ -32,6 +47,48 @@ func Config(path string) (*api.ReleaseBuildConfiguration, error) {
 	if err := yaml.Unmarshal([]byte(raw), configSpec); err != nil {
 		return nil, fmt.Errorf("invalid configuration: %v\nvalue:\n%s", err, raw)
 	}
+
+	// If configresolver address info is provided, GET a copy of the config from there
+	if info != nil {
+		req, err := http.NewRequest("GET", fmt.Sprintf("http://%s/config", info.Address), nil)
```
petr-muller: Personally I would consider putting the gory http request details into a function for readability, but I do not insist on it.

test/ci-operator-integration/multi-stage/run.sh
```diff
@@ -29,4 +30,48 @@ if ! diff "${EXPECTED}" "${OUT}"; then
     exit 1
 fi
 
+# Run test with ci-operator-configresolver
+ci-operator-configresolver -config ${TEST_CONFIG_DIR} -log-level debug -cycle 2m &
```
petr-muller: plz quote the `${TEST_CONFIG_DIR}`

pkg/load/load.go
stevekuznetsov: I don't think we want to prefix with `http://` here

pkg/load/load.go
stevekuznetsov: Yeah rehearsals would need to be still loading from disk from the mutated configmap we pass in, that's a known gap

pkg/load/load.go
```diff
@@ -32,6 +47,48 @@ func Config(path string) (*api.ReleaseBuildConfiguration, error) {
 	if err := yaml.Unmarshal([]byte(raw), configSpec); err != nil {
 		return nil, fmt.Errorf("invalid configuration: %v\nvalue:\n%s", err, raw)
 	}
+
+	// If configresolver address info is provided, GET a copy of the config from there
```
stevekuznetsov: If you factor this out into a smaller function, you should be able to test it with `httptest` mock server without too much issue.

pkg/load/load.go
AlexNPavel: Without `http://`, error: parse 127.0.0.1:8080/config: first path segment in URL cannot contain colon

stevekuznetsov: I was thinking the user would pass it in with the flag -- what happens if we want to serve this between clusters and use a secure route? With this implementation we'd need to change the code, but if we pass `--config-resolver-url https://whatever.com` we would just change runtime parameters

AlexNPavel: Yeah, didn't initially realize that. Fixing it now

pkg/load/load_test.go
```diff
@@ -574,6 +577,102 @@ func TestConfig(t *testing.T) {
 	}
 }
 
+func correctHandler(t *testing.T, jsonConfig []byte) http.Handler {
```
stevekuznetsov: These should be inlined into the test cases you are using them in, as they are validating specific test data

AlexNPavel: Fixed

PR #213, opened 2019-10-14
pj-rehearse: remove configresolver flags from rehearsals

AlexNPavel:
~~Add a flag to `ci-operator` to enable a "rehearse" mode, which simply disables use of the configresolver by preventing the creation of the `load.ResolverInfo` that `load.Config()` uses for the configresolver.~~

~~`pj-rehearse` has been updated to make use of this as well, adding the `--rehearse` flag to all rehearsal jobs that use the `ci-operator` command.~~

The way this is handled has been redone. Now `pj-rehearse` looks at the `Args` for all rehearsal jobs that use the `ci-operator` commands and removes all configresolver configuration args.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -623,11 +623,40 @@ func (e *Executor) waitForJobs(jobs sets.String, selector string) (bool, error) 
 	}
 }
 
+func removeConfigResolverFlags(args []string) []string {
+	newArgs := []string{}
+	ignoreNext := false
+	for _, arg := range args {
+		if ignoreNext {
+			ignoreNext = false
+			continue
+		}
+		ignore := false
+		for _, ignoredArg := range []string{"-resolver-address", "-org", "-repo", "-branch", "-variant"} {
+			// Handle both single and double dash forms of go flags
+			if strings.Contains(arg, ignoredArg) {
+				ignore = true
+			}
+			if !strings.Contains(arg, "=") {
+				ignoreNext = true
+			}
+		}
+		if !ignore {
+			newArgs = append(newArgs, arg)
+		}
+	}
+	return newArgs
+}
+
 func (e *Executor) submitRehearsals() ([]*pjapi.ProwJob, error) {
 	var errors []error
 	var pjs []*pjapi.ProwJob
 
 	for _, job := range e.presubmits {
+		// Remove configresolver flags from ci-operator jobs
+		if len(job.Spec.Containers) > 0 && len(job.Spec.Containers[0].Command) > 0 && job.Spec.Containers[0].Command[0] == "ci-operator" {
+			job.Spec.Containers[0].Args = removeConfigResolverFlags(job.Spec.Containers[0].Args)
+		}
 		created, err := e.submitPresubmit(job)
 		if err != nil {
 			e.loggers.Job.WithError(err).Warn("Failed to execute a rehearsal presubmit")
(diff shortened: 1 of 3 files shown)


cmd/ci-operator/main.go
```diff
@@ -280,6 +282,9 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.StringVar(&opt.branch, "branch", "", "Branch of the project (used by configresolver)")
 	flag.StringVar(&opt.variant, "variant", "", "Variant of the project's ci-operator config (used by configresolver)")
 
+	// flag to disable configresolver for rehearse
+	flag.BoolVar(&opt.rehearse, "rehearse", false, "Run in rehearse mode (disable configresolver calls)")
```
stevekuznetsov: I am not a huge fan of test logic leaking into production code. Couldn't we just remove the org/repo/branch/URL flags when doing a rehearsal?

AlexNPavel: That was one of the other ideas I had about how to handle this. I thought that doing it this way would be a bit cleaner, but either way works.

AlexNPavel: Changed it so now it removes the 5 flags the configure the configresolver instead.

AlexNPavel:
/hold

I want to add some unit tests before merging this.

AlexNPavel:
/hold cancel

The tests have been updated

pkg/rehearse/jobs.go
```diff
@@ -628,6 +628,23 @@ func (e *Executor) submitRehearsals() ([]*pjapi.ProwJob, error) {
...
+			for _, arg := range job.Spec.Containers[0].Args {
+				ignore := false
+				for _, ignoredArg := range []string{"-resolver-address", "-org", "-repo", "-branch", "-variant"} {
+					// Handle both single and double dash forms of go flags
+					if strings.HasPrefix(arg, ignoredArg) || strings.HasPrefix(arg, fmt.Sprintf("-%s", ignoredArg)) {
+						ignore = true
+					}
+				}
+				if !ignore {
+					newArg = append(newArg, arg)
```
petr-muller: This does not handle the `--org value` case (where the value is passed as a following arg) which would be a pitfall. It looks like we do not use that form in our jobs but I still think we should handle that correctly to avoid surprises.

pkg/rehearse/jobs.go
```diff
@@ -628,6 +628,23 @@ func (e *Executor) submitRehearsals() ([]*pjapi.ProwJob, error) {
...
 
 	for _, job := range e.presubmits {
+		// Remove configresolver flags from ci-operator jobs
+		if len(job.Spec.Containers) > 0 && len(job.Spec.Containers[0].Command) > 0 && job.Spec.Containers[0].Command[0] == "ci-operator" {
+			var newArg []string
+			for _, arg := range job.Spec.Containers[0].Args {
+				ignore := false
+				for _, ignoredArg := range []string{"-resolver-address", "-org", "-repo", "-branch", "-variant"} {
+					// Handle both single and double dash forms of go flags
+					if strings.HasPrefix(arg, ignoredArg) || strings.HasPrefix(arg, fmt.Sprintf("-%s", ignoredArg)) {
```
stevekuznetsov: Couldn't we just do `strings.Contains`?

pkg/rehearse/jobs.go
```diff
@@ -628,6 +628,31 @@ func (e *Executor) submitRehearsals() ([]*pjapi.ProwJob, error) {
 	var pjs []*pjapi.ProwJob
 
 	for _, job := range e.presubmits {
```
stevekuznetsov: Could we get a function that takes `[]args` and spits out post-processed args, so that we can unit test it?

PR #200, opened 2019-10-07
Fix failures when no update of image is made

hongkailiu:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/periodic-prow-image-autobump/64

/cc @openshift/openshift-team-developer-productivity-test-platform

PR #218, opened 2019-10-16
Fix failures when no update of image is made (2)

hongkailiu:
The problem is not fixed by [1] since we still found such failures,
(see [2]).

The reason is that the images map returned from bumper pkg does not
reflect the files that have been modifed.

We implement this by a `git-status` call.

[1]. https://github.com/openshift/ci-tools/pull/200
[2]. https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/periodic-prow-image-autobump/85

### DIFF
--- a/cmd/autobumper/main.go
+++ b/cmd/autobumper/main.go
@@ -4,6 +4,8 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"os/exec"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


hongkailiu:
Seems working this time.
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/periodic-prow-image-autobump/86

Will bring it to upstream.

PR #221, opened 2019-10-17
Make sure injected RPM repos are enabled

dulek:
For some reason the repos injected after RPMs are built in the build
process are not enabled. Moreover, on those images yum-config-manager is
disabled as well. In order to enable the repos hacks like those need to
be used:

```
RUN sed -i -e 's/enabled \?= \?0/enabled = 1/' /etc/yum.repos.d/*
```
(See openshift/kuryr-kubernetes@eee5fe3 for reference)

This commit solves that by making sure those injected repos have
`enabled = 1` set.

### DIFF
--- a/pkg/steps/rpm_injection.go
+++ b/pkg/steps/rpm_injection.go
@@ -14,7 +14,7 @@ import (
 
 func rpmInjectionDockerfile(from api.PipelineImageStreamTagReference, repo string) string {
 	return fmt.Sprintf(`FROM %s:%s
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
There was a reason for them not to be enabled by default ... @smarterclayton do you remember?

dulek:
> There was a reason for them not to be enabled by default ... @smarterclayton do you remember?

I don't know what the reason might have been, but if so - what was the correct way to have those repos enabled then? I think those are useless if disabled and my idea with `sed` is really hacky?

stevekuznetsov:
You can enable repos at runtime using `dnf` or `yum` with `--enablerepo`

dulek:
> You can enable repos at runtime using `dnf` or `yum` with `--enablerepo`

That would still mean I need to maintain two Dockerfiles - one for upstream builds with `--enablerepo built` and second one for downstream where (I think?) there is no "built" repo?

stevekuznetsov:
I can't actually remember how this used to work. @smarterclayton do you? The content of this patch (always-on) is sure to be wrong, though. Sorry.

dulek:
Okay, so closing this and keeping my workaround.

PR #223, opened 2019-10-18
Support templates with env. var.

hongkailiu:

follow up https://github.com/openshift/release/pull/5493


cmd/applyconfig/applyconfig.go
```diff
@@ -139,8 +140,20 @@ func (c *configApplier) asGenericManifest() error {
 	return err
 }
 
-func (c configApplier) asTemplate() error {
-	ocProcessCmd := makeOcCommand(ocProcess, c.path, c.user)
+func (c configApplier) asTemplate(params []templateapi.Parameter) error {
```
hongkailiu: /hold

need to check no leaking in the output of the env. vars which are secret in the log.

hongkailiu: /hold cancel

risk is limited as discussed with @petr-muller 

I will work on the logger with censor

cmd/applyconfig/applyconfig.go
```diff
@@ -83,8 +83,9 @@ func isStandardConfig(filename string) bool {
 		!isAdminConfig(filename)
 }
 
-func makeOcCommand(cmd command, path, user string) *exec.Cmd {
+func makeOcCommand(cmd command, path, user string, args1 ...string) *exec.Cmd {
```
petr-muller: let's call `args1` `additional_args` or something...

cmd/applyconfig/applyconfig.go
```diff
@@ -139,8 +140,20 @@ func (c *configApplier) asGenericManifest() error {
 	return err
 }
 
-func (c configApplier) asTemplate() error {
-	ocProcessCmd := makeOcCommand(ocProcess, c.path, c.user)
+func (c configApplier) asTemplate(params []templateapi.Parameter) error {
+	var args []string
+	for _, param := range params {
+		envValue := os.Getenv(param.Name)
+		if len(param.Generate) > 0 {
```
petr-muller: We should do this check before the `Getenv`

cmd/applyconfig/applyconfig.go
```diff
@@ -139,8 +140,20 @@ func (c *configApplier) asGenericManifest() error {
...
+			continue
+		}
+		if len(envValue) > 0 {
+			args = append(args, []string{"-p", fmt.Sprintf("%s=%s", param.Name, envValue)}...)
+		} else {
+			args = append(args, []string{"-p", fmt.Sprintf("%s=%s", param.Name, param.Value)}...)
```
petr-muller: Is this branch even needed? Isn't `param.Value` used by default?

cmd/applyconfig/applyconfig.go
hongkailiu: Done.

cmd/applyconfig/applyconfig.go
hongkailiu: Done.

cmd/applyconfig/applyconfig.go
hongkailiu: Done.

cmd/applyconfig/applyconfig_test.go
```diff
@@ -105,7 +107,7 @@ func TestMakeOcCommand(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			cmd := makeOcCommand(tc.cmd, tc.path, tc.user)
 			if !reflect.DeepEqual(cmd.Args, tc.expected) {
-				t.Errorf("Command differs from expected:\n%s", diff.ObjectReflectDiff(tc.expected, cmd.Args))
+				t.Errorf("Command differs from expected:\n%s", cmp.Diff(tc.expected, cmd.Args))
```
stevekuznetsov: Why did we change this?

cmd/applyconfig/applyconfig_test.go
```diff
@@ -270,11 +293,18 @@ func TestAsTemplate(t *testing.T) {
 		},
 	}
 
+	original := os.Getenv("image")
+	if err := os.Setenv("image", "docker.io/redis"); err != nil {
```
stevekuznetsov: This is not threadsafe, we can't  be doing that in parallel test

cmd/applyconfig/applyconfig_test.go
petr-muller: `// DEPRECATED: use github.com/google/go-cmp/cmp.Diff`, I assume (https://github.com/kubernetes/apimachinery/blob/master/pkg/util/diff/diff.go#L70)

hongkailiu: Thanks Petr.

cmd/applyconfig/applyconfig_test.go
hongkailiu: I have to admit that I have not considered parallel testing when writing the code.
Thanks, @bbguimaraes

PR #224, opened 2019-10-18
Enable censoring for applyConfig's logger

hongkailiu:


cmd/applyconfig/applyconfig.go
```diff
@@ -239,6 +242,55 @@ func applyConfig(rootDir, cfgType string, process processFn) error {
...
+}
+
+func getCensoringFormatter(f logrus.Formatter) logrus.Formatter {
+	secrets := sets.NewString()
+	//those strings do not look like valid secrets
+	noSecrets := sets.NewString("", ":", "1")
+	for _, e := range os.Environ() {
+		pair := strings.SplitN(e, "=", 2)
+		if noSecrets.Has(pair[1]) {
+			logrus.WithField("key", pair[0]).WithField("value", pair[1]).Warn("ignoring censoring value to make log useful ...")
```
hongkailiu: This is from the test on my PC:

```
$  ./applyconfig --config-dir /tmp/test_ac
WARN[0000] ignoring censoring value to make log useful ...  key=MANPATH value=":"
WARN[0000] ignoring censoring value to make log useful ...  key=LOADEDMODULES value=
WARN[0000] ignoring censoring value to make log useful ...  key=SHLVL value=1

```

It indicates that any secret in `sets.NewString("", ":", "1")` will be leaked.
Those are not acceptable as secret anyway.

cmd/applyconfig/applyconfig.go
```diff
@@ -239,6 +242,55 @@ func applyConfig(rootDir, cfgType string, process processFn) error {
 	return nil
 }
 
+type censoringFormatter struct {
```
stevekuznetsov: Why can't we import this from `test-infra`?

cmd/applyconfig/applyconfig.go
stevekuznetsov: Why do we need an approach that is any different from the upstream `bytes.Replace`?

cmd/applyconfig/applyconfig.go
```diff
@@ -239,6 +242,55 @@ func applyConfig(rootDir, cfgType string, process processFn) error {
...
+		}
+	}
+	return content
+}
+
+func getCensoringFormatter(f logrus.Formatter) logrus.Formatter {
+	secrets := sets.NewString()
+	//those strings do not look like valid secrets
+	noSecrets := sets.NewString("", ":", "1")
+	for _, e := range os.Environ() {
```
petr-muller: Huh this looks excessive to me... Have you considered adding the value to the list of items to be censored only after we actually know we read it and will use it as a template parameter?

hongkailiu: That was my original try but gave up after talking to you via [slack](https://coreos.slack.com/archives/GB7NB0CUC/p1571403302269400?thread_ts=1571403013.266700&cid=GB7NB0CUC).

Will try to use the params only.

cmd/applyconfig/applyconfig.go
hongkailiu: In `test-inra`, the censoring logger is associated with an agent, which I do not need/have in the context of `applyConfig`.

cmd/applyconfig/applyconfig.go
hongkailiu: I probably wont need this when fulfil https://github.com/openshift/ci-tools/pull/224/files#r336988964

cmd/applyconfig/applyconfig.go
petr-muller: Sorry for the confusion... I meant _"whatever we read"_ in that post like "whatever we actually read and use from env var" :(

hongkailiu: Done.

@petr-muller 
I am thinking about the comments from @stevekuznetsov above:

I think the censoring logger here is more generic than the one used in `secretAgent` of `test-infra`.
Do you think it is worth a try at upstream (with necessary public class and function) so that `secretAgent` and `applyConfig` can use the shared code via `import`?

petr-muller: Yeah, definitely. There's even a good place for it already, I think: https://github.com/kubernetes/test-infra/tree/master/prow/logrusutil

cmd/applyconfig/applyconfig.go
```diff
@@ -239,6 +244,57 @@ func applyConfig(rootDir, cfgType string, process processFn) error {
 	return nil
 }
 
+type censoringFormatter struct {
+	sync.RWMutex
+	secrets  sets.String
+	delegate logrus.Formatter
+}
+
+func (f *censoringFormatter) Format(entry *logrus.Entry) ([]byte, error) {
```
stevekuznetsov: Feels bad to copy this. Can't we factor out the difference in behavior (impl of how to censor) and reuse the logging bits

hongkailiu: Understood.
Will work on this for upstream today, following https://github.com/openshift/ci-tools/pull/224#discussion_r337089805

cmd/applyconfig/logrusutil/logrusutil.go
```diff
@@ -0,0 +1,47 @@
+package logrusutil
```
hongkailiu: Remove this when https://github.com/kubernetes/test-infra/pull/14889 gets merged.

cmd/applyconfig/applyconfig.go
```diff
@@ -10,9 +10,15 @@ import (
 	"path/filepath"
 	"regexp"
 	"strings"
+	"sync"
 
 	templateapi "github.com/openshift/api/template/v1"
 	templatescheme "github.com/openshift/client-go/template/clientset/versioned/scheme"
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/ci-tools/cmd/applyconfig/logrusutil"
```
hongkailiu: Use `"k8s.io/test-infra/prow/logrusutil"` from https://github.com/kubernetes/test-infra/pull/14889

/hold

hongkailiu:
presubmit is gone ... need to fix the breaking change.

hongkailiu:
local test passed:

```
$ github_client_secret=aaa ./applyconfig --config-dir /tmp/test_ac
INFO[0000] Applying standard config in directory: /tmp/test_ac 
INFO[0000] oc process -f /tmp/test_ac/ttt.yaml -p github_client_secret=CENSORED: OK 
INFO[0001] oc apply -f - --dry-run: OK                  
INFO[0001] Success!                  
```

cmd/applyconfig/applyconfig.go
```diff
@@ -119,14 +124,14 @@ func (c commandExecutor) runAndCheck(cmd *exec.Cmd, action string) ([]byte, erro
 
 	if output, err = cmd.Output(); err != nil {
 		if exitError, ok := err.(*exec.ExitError); ok {
-			fmt.Printf("ERROR: %s: failed to %s\n%s\n", pretty, action, exitError.Stderr)
+			logrus.Errorf("%s: failed to '%s' '%s'", pretty, action, exitError.Stderr)
```
petr-muller: I do not like the removing of separator between the action and stderr (which can be quite long)

hongkailiu: I got the impression that  `\n` is not line break in logrus.
https://github.com/sirupsen/logrus/issues/608

petr-muller: If that is true, then the output will be quite unreadable (the Stderr will very likely contain multiple lines) and I wonder what do we actually gain by switching to logrus...

hongkailiu: The censor is the reason we switched.
Let me try to get an example.

hongkailiu: line breaker seems fine ...

```
$  ./applyconfig --config-dir /tmp/test_ac
INFO[0000] Applying standard config in directory: /tmp/test_ac 
ERRO[0000] oc apply -f /tmp/test_ac/test.yaml --dry-run: failed to 'apply' 'error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:

  1. Via the command-line flag --config
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config

To view or setup config directly use the 'config' command.
' 

```

So created https://github.com/openshift/ci-tools/pull/238
:laughing:

PR #225, opened 2019-10-19
[prowgen] Introduce prowgen configuration files.

droslean:
Now the config contains only one field `private: bool`.
When we detect this in any repo's ci-op config folder, we generate jobs with the bot's secret mounted and add the `ssh-key-path` argument to ci-operator's command.

### DIFF
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -16,6 +16,10 @@ import (
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
 )
 
+const (
+	prowgenConfigFile = ".config.prowgen"
+)
+
 func readCiOperatorConfig(configFilePath string, info Info) (*cioperatorapi.ReleaseBuildConfiguration, error) {
 	data, err := ioutil.ReadFile(configFilePath)
 	if err != nil {
@@ -34,6 +38,11 @@ func readCiOperatorConfig(configFilePath string, info Info) (*cioperatorapi.Rele
 	return configSpec, nil
 }
 
+// Prowgen holds the information of the prowgen's configuration file.
+type Prowgen struct {
+	Private bool `json:"private,omitempty"`
+}
+
 // DataWithInfo describes the metadata for a CI Operator configuration file
 type Info struct {
 	Org    string
@@ -43,6 +52,8 @@ type Info struct {
 	Variant string
 	// Filename is the full path to the file on disk
 	Filename string
+
+	ProwgenConfig Prowgen
 }
 
 // Basename returns the unique name for this file in the config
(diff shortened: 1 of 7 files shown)


pkg/config/load.go
```diff
@@ -34,6 +38,10 @@ func readCiOperatorConfig(configFilePath string, info Info) (*cioperatorapi.Rele
 	return configSpec, nil
 }
 
+type ProwgenConfig struct {
+	Private bool `json:"private,omitempty"`
```
stevekuznetsov: This will be `config.ProwgenConfig` -- maybe rename to `config.Prowgen`?

pkg/config/load.go
```diff
@@ -43,6 +51,8 @@ type Info struct {
 	Variant string
 	// Filename is the full path to the file on disk
 	Filename string
+
+	ProwgenConfig *ProwgenConfig
```
stevekuznetsov: There's no need to use a pointer here, a zero-value struct will have a `false` boolean and will work fine.

bbguimaraes:
Isn't this information relevant to `ci-operator` as well? Why a separate configuration file?

petr-muller:
> Isn't this information relevant to ci-operator as well?

It is not now; ci-operator needs the actual credentials passed via command line arg. I guess we *could* do something like a `credentials: openshift-bot` field for ci-operator config, which would imply some (named by convention?) `openshift-bot-github-ssh-key` secret exists in the cluster. ci-op could fetch its value instead of reading it from a file, and prowgen would use the field to generate an appropriate mount config.

Pros: Less new things
Cons: Binds ci-op config with a single arbitrary secret/account

Can't say I'm not tempted...

@droslean @stevekuznetsov WDYT?

bbguimaraes:
Responding to the cons pointed out in an existing design document:

- `prowgen`-specific field being added to `ci-operator`'s configuration: that seems a bit backwards. It is `prowgen` specific because the logic added to `ci-operator` was "I was given an ssh key (by a flag added by `prowgen`), so this is a private repository". It could just as well have been "this repository is private according to the configuration, I will verify that I was also given an ssh key".
- `private` is at minimum a repository-level configuration: that is true for other fields (`canonical_go_repository` comes to mind).

I am not completely opposed to this PR, but I don't see enough reason to add a new configuration file just for this case.

petr-muller:
> private is at minimum a repository-level configuration: that is true for other fields (canonical_go_repository comes to mind).

True, but that does not make it less undesirable.

> I am not completely opposed to this PR, but I don't see enough reason to add a new configuration file just for this case.

The good reason is the ability to simply set this org-wide (feature not present in this PR but still desired) vs having an additional config line in each of the private org's repo/branch config file.

petr-muller:
LGTM except for what @stevekuznetsov said, especially the test yaml amount.

After some consideration, I personally still think this is a good way to go forward.

pkg/config/load.go
```diff
@@ -90,12 +101,23 @@ func InfoFromPath(configFilePath string) (*Info, error) {
 		branch = branch[:i]
 	}
 
+	pConfig := Prowgen{}
+	b, err := ioutil.ReadFile(filepath.Join(configSpecDir, prowgenConfigFile))
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("prowgen config found but couldn't read the file: %v", err)
+	}
+
+	if err := yaml.Unmarshal(b, &pConfig); err != nil {
```
petr-muller: I think we should be more explicit here, and not even attempt to `Unmarshal` when `ReadFile` failed. It is true that it will probably work, because `b` is probably empty (but is not documented as such), and `yaml.Unmarshal` works on empty string, but I do not like relying on undocumented behavior.

droslean: addressed

pkg/config/load.go
```diff
@@ -90,12 +101,23 @@ func InfoFromPath(configFilePath string) (*Info, error) {
 		branch = branch[:i]
 	}
 
+	pConfig := Prowgen{}
+	b, err := ioutil.ReadFile(filepath.Join(configSpecDir, prowgenConfigFile))
+	if err != nil && !os.IsNotExist(err) {
+		return nil, fmt.Errorf("prowgen config found but couldn't read the file: %v", err)
+	} else {
```
petr-muller: This should be `else if err != nil {` otherwise it is the same like before, would `Unmarshal(b, ...)` even if it has whatever value when `ReadFile` failed with NotExist.

droslean: thanks nice catch :)

PR #226, opened 2019-10-20
cmd/blocking-issue-creator: Use "Failed to ..." for all failing errors

wking:
This makes it easier to grep for things which might eventually cause an terminal failure [like][1]:

```
level=fatal msg="Could not publish merge blocking issues." error="<nil>"
```

You can already find those by grepping for `level=error`, since we only use error logging before setting `failed=true`, which you can confirm with:

```console
$ git grep -1 'logger.*Error\|failed = true' cmd/blocking-issue-creator/
```

But "there's already some consistency" shouldn't block us from adding the additional consistency I'm adding here ;).

[1]: https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/periodic-openshift-release-merge-blockers/692


stevekuznetsov:
We can also update the log viewer to highlight these lines like in https://github.com/openshift/release/pull/5510 for better dev UX looking at the job.

PR #232, opened 2019-10-23
Add a unit test for isTemplate of applyConfig

hongkailiu:

### DIFF
--- a/cmd/applyconfig/applyconfig_test.go
+++ b/cmd/applyconfig/applyconfig_test.go
@@ -457,6 +457,50 @@ spec:
     name: hook
 `),
 		},
+		{
+			//https://docs.openshift.com/container-platform/4.2/openshift_images/using-templates.html#templates-writing_using-templates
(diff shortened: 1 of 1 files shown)


hongkailiu:
/hold 

give it a try when https://jira.coreos.com/browse/DPTP-623 is done.

cmd/applyconfig/applyconfig_test.go
```diff
@@ -457,6 +457,50 @@ spec:
     name: hook
 `),
 		},
+		{
+			//https://docs.openshift.com/container-platform/4.2/openshift_images/using-templates.html#templates-writing_using-templates
+			name: "template in openshift doc in the doc is a template",
+			contents: []byte(`apiVersion: v1
```
hongkailiu: Will pass when `apiVersion: template.openshift.io/v1`

hongkailiu: I think we need to allow for `v1` as well.

PR #234, opened 2019-10-24
[WIP]Align k8s.io versions with test-infra

hongkailiu:

https://jira.coreos.com/browse/DPTP-623


hongkailiu:
TODO:
upgrade openshift/{api,client-go} as well.

https://github.com/openshift/ci-tools/blob/9f0782d1144317f46e62a96a9bb6b984f2c73bad/go.mod#L26-L27

Asked #forum-apiserver via slack
https://coreos.slack.com/archives/CB48XQ4KZ/p1571918611058100

bbguimaraes:
It would be useful to document how to determine these version (in the file/commit message) since I imagine this will have to be done constantly.

Does it come directly from https://github.com/kubernetes/test-infra/blob/master/go.mod?

PR #236, opened 2019-10-24
Define a pull_request_policy that allows gen of always_run:false

smarterclayton:
always_run is hard to maintain across a thousand jobs, so we should
bring it into the config. Use a policy rather than a boolean so we
can evolve over time (and prevent configs we don't really want to be
frequently used, like optional: true, always_run: false).

Plumb into generation.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -479,15 +479,18 @@ func generateJobBase(name, prefix string, info *config.Info, label jc.ProwgenLab
 	return base
 }
 
-func generatePresubmitForTest(name string, info *config.Info, label jc.ProwgenLabel, podSpec *kubeapi.PodSpec, rehearsable bool, pathAlias *string) *prowconfig.Presubmit {
+func generatePresubmitForTest(name string, alwaysRun bool, runIfChanged string, info *config.Info, label jc.ProwgenLabel, podSpec *kubeapi.PodSpec, rehearsable bool, pathAlias *string) *prowconfig.Presubmit {
 	if len(info.Variant) > 0 {
 		name = fmt.Sprintf("%s-%s", info.Variant, name)
 	}
 	base := generateJobBase(name, presubmitPrefix, info, label, podSpec, rehearsable, pathAlias)
 	return &prowconfig.Presubmit{
 		JobBase:   base,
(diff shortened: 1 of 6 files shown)


smarterclayton:
/assign @petr-muller 
/assign @droslean 

for review since Steve said you were working in here on related stuff

smarterclayton:
In the future this could add `pull_request_if_changed: <regex>` which causes policy to be `pull_request_policy: OnSourceChange`

smarterclayton:
Added run_if_changed support.

droslean:
configResolver integration tests need to be updated as well

pkg/api/types.go
```diff
@@ -330,6 +330,27 @@ type TestStepConfiguration struct {
...
+	// * Always: this test is run on every pull request
+	// * OnRequest: this test will not be run until a user
+	//     explicitly requests it via /test NAME, at which
+	//     point it will be required for merge.
+	// * RunIfChanged: this test will be run if the regex
+	//     in pull_request_run_if_changed matches a change
+	//     in the PR.
+	//
+	// The default value is Always
+	PullRequestPolicy string `json:"pull_request_policy"`
```
petr-muller: Is there a reason why these should not be `omitempty`? We would generate a ton of useless empty strings when determinizing and generating configs for release branches.

cmd/ci-operator-prowgen/main.go
```diff
@@ -563,7 +566,16 @@ func generateJobs(
 		}
 
 		if element.Cron == nil {
-			presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(element.As, info, label, podSpec, true, configSpec.CanonicalGoRepository))
+			var alwaysRun bool
+			switch element.PullRequestPolicy {
+			case "Always", "":
+				alwaysRun = true
+			case "OnRequest":
```
petr-muller: ` RunIfChanged` makes this more complicated, no? `RunIfChanged` implies `alwaysRun = false`, and also nonempty `PullRequestRunIfChanged` and empty `PullRequestPolicy` together imply `alwaysRun = false` too.

petr-muller:
>I thought this might be a fit for the drop-in config

I don't think so - these are tied to individual tests/jobs, drop-ins are repo/org level.

PR #239, opened 2019-10-25
Read and cache prowgen configs in prowgen

petr-muller:
Proof-of-concept quality: haven't invested time into writing tests (but I tested it with both repo-level and org-level drop-ins), error handling and code cleanups, but I will do so if the main idea gets an ACK.

Compared to https://github.com/openshift/ci-tools/pull/235, where an attempt to read Prowgen drop-ins together with ci-op configs lead to the need of a global state, this puts the responsibility for prowgen drop-ins into prowgen itself. Prowgen configs are read and stored into a new struct that wraps the generic `config.Info`, leading to a larger blast radius, but the actual change in logic is isolated into the `generateJobsToDir` method. In order to not leak the details about the directory structure from the responsibility of `pkg/config/load.go` code, the `config.Info` structure was augmented with information about locations, so Prowgen does not need to compute them.

@droslean PTAL


cmd/ci-operator-prowgen/main.go
```diff
@@ -595,11 +599,57 @@ func generateJobs(
...
+		}
+	}
+
+	return pConfig // , nil
+}
+
 // generateJobsToDir returns a callback that knows how to generate prow job configuration
 // into the dir provided by consuming ci-operator configuration
 func generateJobsToDir(dir string, label jc.ProwgenLabel) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
+	cache := map[string]*Config{}
```
droslean: This can live in the `prowgenInfo`

cmd/ci-operator-prowgen/main.go
```diff
@@ -595,11 +599,57 @@ func generateJobs(
...
 	return func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
-		return jc.WriteToDir(dir, info.Org, info.Repo, generateJobs(configSpec, info, label))
+		orgRepo := fmt.Sprintf("%s/%s", info.Org, info.Repo)
+		pInfo := &prowgenInfo{Info: *info, config: Config{Private: false}}
+		orgConfig, ok := cache[info.Org]
+		if !ok {
+			cache[info.Org] = readProwgenConfig(filepath.Join(info.OrgPath, prowgenConfigFile))
```
droslean: we really need to catch an error while trying to read the file.

petr-muller:
> This refactoring makes sense. I've left a couple of nits, but in general, it looks good.

OK so I'll finalize it and clean it up a bit.

> The only thing I don't like is that we `hacky` assume that the org dir is the parent directory, which this will always true since it is designed like that, so no problems.

I'd like to point out that the assumption is already there - in the same piece of code we infer the org name from the name of the directory (https://github.com/openshift/ci-tools/pull/239/files#diff-08b796968aefd12ccbb1e9a0959b41f2R82), so while it's not maybe not ideal, it's not new. Plus, I've kept the assumption localized in a single place (the loading code).

cmd/ci-operator-prowgen/main.go
petr-muller: I don't understand. The cache is only needed in the returned function (a closure), to build a specific `prowgenInfo`. Having a pointer to the cache in `prowgenInfo` is possible, but IMHO is not needed for anything.

petr-muller: I don't understand. The cache is only needed in the returned function (a closure), to build a specific prowgenInfo. Having a pointer to the cache in prowgenInfo is possible, but IMHO is not needed for anything.

droslean: Yes, you are right. Since we are doing everything serial here, caching here makes sense.

cmd/ci-operator-prowgen/main.go
petr-muller: Definitely, I ignored it only to crank out the PoC. Errors are now propagated. I have considered eating them and only issuing a warning, but decided not to, because nbody cares about warnings.

petr-muller:
At first I was like "WTF, I ran the test locally,", but then I realized I did that before I needed to rebase. The rebase brought in the removing the hack where we generated a private-specific decoration_config for private jobs, but my expected test jobs still had that.

PR #242, opened 2019-10-29
ci-operator: Import release image stream to correct namespace

smarterclayton:
We should not be importing these images into the release
source namespace, but instead to the job spec namespace.


smarterclayton:
Same as #241

PR #245, opened 2019-10-30
types: consolidate api.LiteralTestStep & steps/types.TestStep

AlexNPavel:
This PR moves the structs from `steps/types` into `api/types.go`. The types in `steps/types` are equivalent to the types for `MultiStageTestConfiguration` with the `Reference`s, `Chain`s, and `Workflow`s removed. By moving these types into `api` and adding a new `literal_steps` test type alongside the `steps` type that the standard `MultiStepTestConfiguration` has, we can hand the ci-operator a resolved config that does not have references and `nil` values that it may need to handle in the multi-step configurations otherwise.


AlexNPavel:
/hold
I need to add more config file validation for `literal_steps`

PR #246, opened 2019-10-30
pkg/load/configAgent: make some functions more generic

AlexNPavel:
This PR refactors `pkg/load/configAgent.go` to make it more generic and make implementing a similar `registryAgent.go` cleaner.

- Renames `agent` to `configAgent`
- Renames `reloadTimeMetric` to `configReloadTimeMetric`
- Makes `reloadWatcher` more generic by not requiring a `configAgent` in the function signature
- Move generic agent helpers to their own file

### DIFF
--- a/pkg/load/agent_utils.go
+++ b/pkg/load/agent_utils.go
@@ -1,0 +1,52 @@
+package load
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/fsnotify/fsnotify"
+	log "github.com/sirupsen/logrus"
+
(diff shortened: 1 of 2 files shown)


pkg/load/agent_utils.go
```diff
@@ -0,0 +1,52 @@
+package load
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/fsnotify/fsnotify"
+	log "github.com/sirupsen/logrus"
```
petr-muller: nit: do we need the `log` name? most other places are fine with using the `logrus` name...

PR #247, opened 2019-10-31
pkg/api/config: add validation tests for literal_steps

AlexNPavel:
Adds validation checks for the new `literal_steps` TestConfigurationType


AlexNPavel:
This is being implemented as part of a larger PR now

PR #248, opened 2019-11-01
configresolver: add support for registry resolution

AlexNPavel:
This adds support for registry resolution by the config resolver as well as fully migrates from using `MultiStageTestConfiguration` internally to `MultiStageTestConfigurationLiteral`, which is what `steps` gets resolved to by the `registryAgent`.

Short summary of changes:
- Adds `registryAgent`
- Adds `registryAgent` calls/support to configresolver
- Changes internal usage of `MultiStageTestConfiguration` to `MultiStageTestConfigurationLiteral` both in code and in the ci-operator integration tests
- Adds missing validation of `literal_steps` type in `pkg/api/config.go`
- Updates configresolver integration tests to include registry resolution


droslean:
@AlexNPavel It would be better to have each related changes to their commit. This will make the review proccess easier.

AlexNPavel:
@droslean Yeah, this grew a lot bigger than I initially expected. I'll make some PRs to implement smaller parts of this first.

/hold

test/ci-operator-integration/multi-stage/configs/master/openshift-hyperkube-master.yaml
```diff
@@ -38,7 +38,7 @@ tag_specification:
   namespace: ci
 tests:
 - as: multi-stage
-  steps:
+  literal_steps:
```
stevekuznetsov: This seems odd -- why? Users should never set this

test/ci-operator-integration/multi-stage/run.sh
```diff
@@ -30,8 +31,8 @@ if ! diff "${EXPECTED}" "${OUT}"; then
     exit 1
 fi
 
-# Run test with ci-operator-configresolver
-ci-operator-configresolver -config "${TEST_CONFIG_DIR}" -log-level debug -cycle 2m &
+# Run test with ci-operator-configresolver; registry isn't actually being used here, but needs to be set...
```
stevekuznetsov: Why not use it?

test/ci-operator-integration/multi-stage/configs/master/openshift-hyperkube-master.yaml
AlexNPavel: This is part of the `ci-operator`'s integration tests, so this is supposed to be a config that the ci-operator would get from the resolver (or pj-rehearse when we update that to support registry), which would have converted the `steps` to `literal_steps` during resolution. With the other changes made to multistage, this section would simply be ignored if it was just `steps`.

test/ci-operator-integration/multi-stage/run.sh
AlexNPavel: This test part of the `ci-operator-integration` tests, so we're just testing that the ci-operator can communicate with the configresolver here. We actually test proper registry support in the `ci-operator-configresolver-integration` tests. In this case, the result shouldn't be any different with or without a populated registry, as all the steps in the confgi used here are literal steps.

test/ci-operator-integration/multi-stage/configs/master/openshift-hyperkube-master.yaml
stevekuznetsov: Ah, so this is something served by the registry? I was expecting this file to be read from disk in this test if the aim of the test is not to use the resolver.

stevekuznetsov:
I think I'm looking for the full, end-to-end user story where they choose a mutli-stage run not by `steps` but by using a resolution feature

AlexNPavel:
The PR has been updated with more comprehensive tests

PR #252, opened 2019-11-05
Added the ability to attach custom prow metadata to ci-operator jobs.

meowfaceman:
If Prow jobs generate a file called "custom-prow-metadata.json", it will
be captured and written to the metadata.json file which is parsed in
Spyglass. This will allow custom metadata to be displayed in Spyglass
for Prow job runs.


cmd/ci-operator/main_test.go
```diff
@@ -61,3 +72,148 @@ func TestSanitizeMessage(t *testing.T) {
...
+	err = ioutil.WriteFile(filepath.Join(testArtifactDirectory, "e-ignore1.txt"), []byte(""), os.FileMode(0644))
+	if err != nil {
+		t.Errorf("One or more of the empty *ignore files failed to write. (%v)", err)
+		return
+	}
+
+	o := createOptions(tempDir)
+	o.writeMetadataJSON()
+
+	verifyMetadata(t, tempDir)
```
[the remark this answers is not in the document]
droslean: In addition, we should pass the values in that func and use it in a multi test cases.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
 	if err != nil {
 		return err
 	}
+
+	// Try to find the custom prow metadata file. We assume that there's only one. If there's more than one,
+	// we'll just use the first one that we find.
+	var customProwMetadataFile string
+	filepath.Walk(o.artifactDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return filepath.SkipDir
```
petr-muller: This is imho unnecessary strict and could cause a proper file being skipped. Just return nil?

https://golang.org/pkg/path/filepath/#Walk:
> If the function returns SkipDir when invoked on a non-directory file, Walk skips the remaining files in the containing directory.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
+		}
+
+		// If the prow metadata file is set, just return, since we've already found the file
+		if customProwMetadataFile != "" {
+			return filepath.SkipDir
+		}
+
+		var fileName = info.Name()
```
petr-muller: Extra variable seems unnecessary.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
+	var customProwMetadataFile string
+	filepath.Walk(o.artifactDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return filepath.SkipDir
+		}
+
+		// If the prow metadata file is set, just return, since we've already found the file
+		if customProwMetadataFile != "" {
+			return filepath.SkipDir
+		}
```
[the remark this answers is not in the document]
meowfaceman: This isn't how I read https://golang.org/pkg/path/filepath/#WalkFunc: it seems like SkipDir will skip the rest of the directory's contents, but could potentially descend into other directories. Can you clarify? I'm a bit of a go noob.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
+		// Turns out Unmarshaling into an map is not destructive, so if we call it twice it will
+		// effectively merge these two JSON objects.
+		var mergedJSON map[string]interface{}
+		json.Unmarshal(prowMetadataJSON, &mergedJSON)
+
+		var customJSONFile []byte
+		if customJSONFile, err = ioutil.ReadFile(customProwMetadataFile); err != nil {
+			return err
+		}
+		json.Unmarshal(customJSONFile, &mergedJSON)
```
petr-muller: This order would overwrite ci-operator's metadata with custom ones on key conflict, which is undesirable.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
+	if customProwMetadataFile != "" {
+		log.Printf("Found custom prow metadata.")
+
+		// Turns out Unmarshaling into an map is not destructive, so if we call it twice it will
+		// effectively merge these two JSON objects.
+		var mergedJSON map[string]interface{}
+		json.Unmarshal(prowMetadataJSON, &mergedJSON)
+
+		var customJSONFile []byte
+		if customJSONFile, err = ioutil.ReadFile(customProwMetadataFile); err != nil {
```
petr-muller: Have you made sure Prow is able to handle if there is some deeper hierarchy under the key? Ignoring such key is fine but choking is not.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
+			return err
```
petr-muller: This would make failure to read custom JSON to not emit ci-operator's metadata. It's desirable to make the run fail, but the standard metadata needs to be emitted.

cmd/ci-operator/main.go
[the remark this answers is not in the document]
droslean: oh yes. it seems that `Walk` will continue if it get an `filepath.SkipDir` error.

cmd/ci-operator/main.go
```diff
@@ -846,10 +849,56 @@ func (o *options) writeMetadataJSON() error {
...
+		}
+		json.Unmarshal(customJSONFile, &mergedJSON)
+
+		if data, err = json.MarshalIndent(mergedJSON, "", "  "); err != nil {
+			return nil
```
petr-muller: Maybe we should try to emit at least the standard ci-operato metadata if marshalling the merged ones fail (probably super paranoid, feel free to ignore).

cmd/ci-operator/main.go
meowfaceman: Good catch. This is probably the right way to go.

petr-muller:
> Only one concern. Those changes will assume that the metadata file is already in the artifacts directory. This is useful if we run the ci-operator locally, otherwise, if we need to pass this in our jobs we need to change a lot of stuf

@droslean Huh, can you give an example of a problem that you see with this? I don't understand the _if we need to pass this in our jobs we need to change a lot of stuff_ part.

I think the concept here is fine - find a specific artifact in the artifact directory, and if it exists, merge its content into the metadata that ci-operator emits on its own. If not, then emit just ci-operator's metadata like before.

droslean:
@petr-muller For example [templates/openshift/installer/cluster-launch-installer-e2e.yaml#L105-L107](https://github.com/openshift/release/blob/master/ci-operator/templates/openshift/installer/cluster-launch-installer-e2e.yaml#L105-L107)

In our templates, we are just using an `EmptyDir` for the artifacts. There is no way for the ci-operator to find something there when it will run.

meowfaceman:
> In our templates, we are just using an `EmptyDir` for the artifacts. There is no way for the ci-operator to find something there when it will run.

In that case will we even write out prow metadata? As is if the artifactDir is empty we'll skip writing anything out.

petr-muller:
> In our templates, we are just using an EmptyDir for the artifacts. There is no way for the ci-operator to find something there when it will run.

Oh, right - I think there's a misunderstanding, but you are actually right. I understood (@meowfaceman , correct?)  the goal of the PR is to allow prowjobs to generate (as a part of the test workload) an artifact that would be recongnized and added to the metadata. But the good catch is that `writeMetadataJSON` is actually called **before** the workload: https://github.com/openshift/ci-tools/blob/6e76ddcfe7a0975f3b2270229230db2cdea937f5/cmd/ci-operator/main.go#L496

At that time, like Nikos says, there's nothing.

droslean:
@meowfaceman Can you let us know, why we need this feature? We need to understand the use cases first.

meowfaceman:
Well crap. That makes things harder.

@droslean I'm on the OSD CI/CD team working on https://github.com/openshift/osde2e. Basically, we want people to be able to look at jobs and inspect, in Spyglass, what versions of OSD were used during a test. This way people can validate what upgrades occurred and whether the test was successful without the need to delve into the logs. This is part of a greater effort add eventual automation to OSD verification.

droslean:
@meowfaceman Thanks. Can you also explain to us how your team is using `ci-operator`?

meowfaceman:
We're generating our jobs here: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/osde2e/openshift-osde2e-master.yaml

@stevekuznetsov recommended that we use the config.yaml approach as opposed to generating our own job.yaml files. Perhaps ci-operator is too far outside of what we're trying to accomplish?

droslean:
So. Since you are using `ci-operator` in this way, in our cluster, Your feature will never be useful. You need to find a way also to inject the custom metadata to the volume.

meowfaceman:
Yeah, sounding more and more like we need to move away from ci-operator -- sounds like it's not a good fit for us. Thanks for the clarification.

meowfaceman:
After talking with @petr-muller , looks like this could be viable with some minor changes. Will update the PR.

stevekuznetsov:
Just looking at this now -- updating `metadata.json` would work for now but one of the design goals of `spyglass` was that people could write their own lenses to render artifacts into the overview, instead of _having_ to put everything into the metadata and slamming everything into a table. That's more work than updating the metadata and could be a good follow-up/long-term solution but should not be a blocker for this proces..

meowfaceman:
So I've added in a second call into writeMetadataJSON at the end of a job (by my understanding), as well as a catch to prevent that write if no custom metadata was generated by the workload. Let me know if you guys think this is on the right track.

cmd/ci-operator/main.go
```diff
@@ -824,6 +833,36 @@ func (o *options) writeMetadataJSON() error {
 		return nil
 	}
 
+	// Try to find the custom prow metadata file. We assume that there's only one. If there's more than one,
+	// we'll just use the first one that we find.
+	var customProwMetadataFile string
+	filepath.Walk(o.artifactDir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
```
droslean: This metadata file will be generated from the job, right? Can we find another way of detecting it? Maybe we can allocate a specific env var for the custom metadata file, instead of trying to find it. Something like `$CUSTOM_PROW_METADATA_FILE`.

@stevekuznetsov @petr-muller Any thoughts on that?

petr-muller: One thing that makes it hard is the indirection layer in the template. `ci-operator` does not know the name of the directory where the template will put its artifacts.

Another thing is that ci-operator can run multiple targets, and each target can (in theory) create different artifacts, so having a single location would be tricky. So I think we'll eventually want to handle multiple of these anyways, which seems fine to me to do incrementally and half-ass it with the "first found, first used" approach here.

cmd/ci-operator/main.go
```diff
@@ -824,6 +833,34 @@ func (o *options) writeMetadataJSON() error {
...
+			return filepath.SkipDir
+		}
+
+		return nil
+	})
+
+	metadataJSONPath := filepath.Join(o.artifactDir, "metadata.json")
+
+	// If the metadata JSON exists and there's no custom prow metadata, then skip the second write.
+	_, existErr := os.Stat(metadataJSONPath)
```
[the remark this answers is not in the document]
meowfaceman: Found https://gist.github.com/mastef/05f46d3ab2f5ed6a6787 which helped me understand what was going on here. Looks like with this pattern the thing to do is just check if the err != nil.

meowfaceman:
> I would suggest explicitly targeting a specific field or fields to allow for merging.

I think this either blows away our use case or will result in a very long "extra info" field that users will have to parse by eye. The latter is doable, though I wouldn't prefer it, because it makes for less computer-readable data that would be harder to scrape later should we need to do it.

We could potentially iterate through the custom metadata and explicitly remove any field that corresponds with the prowResultMetadata struct using reflect. That way we're guaranteed not to have any merging conflicts. @stevekuznetsov thoughts? In Java reflection was always kind of iffy business, so not sure how things are in go, but I wouldn't be surprised if it was the same.

stevekuznetsov:
What is even allowed for metadata presentation? Why would it be less useful to have a `extraFields map[string]string` that the jobs could set?

meowfaceman:
> What is even allowed for metadata presentation? Why would it be less useful to have a `extraFields map[string]string` that the jobs could set?

That's fair. If we could present some arbitrary map, that would make sense. That being said, does spyglass support nested JSON objects as metadata? I don't see the "repos" field in https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/logs/periodic-ci-openshift-osde2e-master-e2e-int-4.1/199 despite it being set in https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-openshift-osde2e-master-e2e-int-4.1/199/artifacts/metadata.json.

stevekuznetsov:
I don't know the Spyglass code that well, so I would have to dig through it to figure that out. Like I said, throwing things into the table is kind of a hack and one of the core design principles behind Spyglass was to allow custom, federated rendering.

stevekuznetsov:
The code for rendering the metadata table is here: https://github.com/kubernetes/test-infra/blob/master/prow/spyglass/lenses/metadata/lens.go

meowfaceman:
I'll take a look. Just to make sure this is clear, the current code *does* preserve the original prow metadata values even if they're present in the custom JSON. I can add tests to verify this, though.

stevekuznetsov:
I know, but why let a user provide any fields and bother having code that needs to be tested for that fact when we just need to inject one field? :)

meowfaceman:
@stevekuznetsov Well well well! It looks like there's already a mechanism for this baked into testgrid by adding in a field called "metadata" with a sub object underneath. I'll reorient this code to shove all of this into the Metadata field instead of merging it with the base metadata object and take your advice about refactoring the writeMetadataJSON function.

cmd/ci-operator/main.go
```diff
@@ -817,15 +826,74 @@ type prowResultMetadata struct {
...
+		return customMetadataErr
+	}
+
+	return nil
+}
+
+func (o *options) findCustomMetadataFile() (customProwMetadataFile string) {
+	// Try to find the custom prow metadata file. We assume that there's only one. If there's more than one,
+	// we'll just use the first one that we find.
+	filepath.Walk(o.artifactDir, func(path string, info os.FileInfo, err error) error {
```
stevekuznetsov: This returns an error you're ignoring -- I would suggest looking to see if your editor can highlight that

cmd/ci-operator/main.go
```diff
@@ -846,15 +914,27 @@ func (o *options) writeMetadataJSON() error {
...
-	}
-	if o.dry {
-		log.Printf("metadata.json:\n%s", string(data))
-		return nil
+	return m
+}
+
+func (o *options) parseCustomMetadata(customProwMetadataFile string) (customMetadata map[string]string, err error) {
+	// Merge custom metadata into resulting JSON
+	if customProwMetadataFile != "" {
```
stevekuznetsov: nit: you can check to see if this is true at the call-site and just don't call this function when there's nothing to do

cmd/ci-operator/main.go
```diff
@@ -846,15 +914,27 @@ func (o *options) writeMetadataJSON() error {
...
+		log.Printf("Found custom prow metadata.")
+
+		// Turns out Unmarshaling into an map is not destructive, so if we call it twice it will
+		// effectively merge these two JSON objects.
+		if customJSONFile, readingError := ioutil.ReadFile(customProwMetadataFile); readingError != nil {
+			log.Printf("Error while reading custom prow metadata.(%v)", readingError)
```
stevekuznetsov: nit: we use `": %v"` for errors, not parentheses

meowfaceman:
So AFAIK, I've addressed most of the major concerns. The current review:

- Cannot produce any collisions with the generated prow metadata
- Stores the metadata in the metadata field, which is required by spyglass
- Resolved all idiomatic go issues raised
- Catches all generated errors (:crossed_fingers:) and checks them
- Verifies the expected double write of the metadata file
- Overall is better engineered than the first pass.

Let me know what you guys think.

cmd/ci-operator/main_test.go
```diff
@@ -61,3 +76,204 @@ func TestSanitizeMessage(t *testing.T) {
...
+	if hasCustomMetadata && !info.ModTime().After(originalModTime) {
+		return fmt.Errorf("file was not modified after the second function call when it should have due to the presence of custom metadata")
+	} else if !hasCustomMetadata && info.ModTime().After(originalModTime) {
+		return fmt.Errorf("file was modified after the second function call when it should not have due to the absence of custom metadata")
+	}
+
+	return nil
+}
+
+func createOptions(tempDir string, org string, repo string, extraOrg string, extraRepo string, prowJobID string, namespace string) *options {
```
stevekuznetsov: I'm not a fan of methods like these. Test should, when possible, contain all the logic that makes them tick in the `Test*` method itself. If you're worried about vertical space in your editor, take up some horizontal space instead:

```go
o := &options{
	artifactDir: tempDir,
	namespace: namespace,
	jobSpec: &api.JobSpec{JobSpec: downwardapi.JobSpec{
			ProwJobID: prowJobID,
			Refs: &prowapi.Refs{Org:  org, Repo: repo},
			ExtraRefs: []prowapi.Refs{{Org:  extraOrg, Repo: extraRepo}},
		},
	},
}
```

cmd/ci-operator/main_test.go
```diff
@@ -61,3 +76,204 @@ func TestSanitizeMessage(t *testing.T) {
...
+					},
+				},
+				ProwJobID: prowJobID,
+			},
+		},
+		namespace: namespace,
+	}
+}
+
+func verifyMetadataFields(metadataFile string, org string, repo string, extraOrg string, extraRepo string, prowJobID string, namespace string, customMetadata map[string]string) error {
```
stevekuznetsov: Similarly, I am not a fan of custom validation methods either. We should try to present the literal data that we want to validate. Whether that's raw JSON strings we expect to exist on disk or literal `struct`s once they're loaded, we should validate by either `ioutil.ReadFile()` and `string(expected) == string(read)` or load, then unmarshal and use `reflect.DeepEqual` to validate expected structures. 

When presenting a diff to a user, use `diff.ObjectReflectDiff` or `StringDiff` for rich and useful diffs on complex objects.

cmd/ci-operator/main_test.go
```diff
@@ -61,3 +76,204 @@ func TestSanitizeMessage(t *testing.T) {
...
+			return fmt.Errorf("error marshalling custom metadata: %v", err)
+		}
+		err = ioutil.WriteFile(filepath.Join(testArtifactDirectory, "custom-prow-metadata.json"), []byte(testJSON), os.FileMode(0644))
+		if err != nil {
+			return fmt.Errorf("unable to create custom metadata file: %v", err)
+		}
+	}
+
+	// Write a bunch of empty files that should be ignored
+	err = ioutil.WriteFile(filepath.Join(testArtifactDirectory, "a-ignore1.txt"), []byte(""), os.FileMode(0644))
```
stevekuznetsov: Maybe some of these should contain valid custom metadata JSON?

cmd/ci-operator/main_test.go
```diff
@@ -61,3 +76,204 @@ func TestSanitizeMessage(t *testing.T) {
...
+	err = ioutil.WriteFile(filepath.Join(testArtifactDirectory, "b-ignore1.txt"), []byte(""), os.FileMode(0644))
+	err = ioutil.WriteFile(filepath.Join(testArtifactDirectory, "d-ignore1.txt"), []byte(""), os.FileMode(0644))
+	err = ioutil.WriteFile(filepath.Join(testArtifactDirectory, "e-ignore1.txt"), []byte(""), os.FileMode(0644))
+	if err != nil {
+		return fmt.Errorf("one or more of the empty *ignore files failed to write: %v", err)
+	}
+
+	// Sleep for one second to guarantee we get a different modified time of the second write
+	time.Sleep(1 * time.Second)
```
stevekuznetsov: Most of our unit tests run on the order of milliseconds or microseconds so something like this can have an outsize impact.

At least on my machine, `stat` gives me modification times down to the nanosecond, so we should be able to reduce this wait to 10ms or so. 

```
$ stat /tmp/
  File: /tmp/
  Size: 760       	Blocks: 0          IO Block: 4096   directory
Device: 2dh/45d	Inode: 22654       Links: 31
Access: (1777/drwxrwxrwt)  Uid: (    0/    root)   Gid: (    0/    root)
Context: system_u:object_r:tmp_t:s0
Access: 2019-11-07 10:09:19.414764485 -0800
Modify: 2019-11-07 11:42:24.296412280 -0800
Change: 2019-11-07 11:42:24.296412280 -0800
 Birth: -
```

However, adding time dependencies to a test is often buggy as you have no control over the kernel deciding to sleep your thread for whatever amount of time and in rare cases on time boundaries `time.After()` is not guaranteed to work. If we can validate that the code did something by checking for a changed field and not use time, that would be preferrable.

meowfaceman: In practice without the sleep the mod time before == the mod time after regardless of the write, I'm guessing due to CPU scheduling. I'm not sure how else to generically check that the file has or hasn't been modified other than adding a "revision" field or something to the metadata JSON itself. This is doable, it just sounds like it's not preferable.

meowfaceman: Thinking on it, I'll add in the revision. Should be straightforward.

cmd/ci-operator/main_test.go
meowfaceman: FYI, I elected to use `go-cmp/cmp.Diff` as `diff.ObjectReflectDiff` comments note that it's deprecated in favor of go-cmp's version.

PR #254, opened 2019-11-06
[WIP]Enable prowgen to support build clusters

hongkailiu:
Currently support only non-e2e presubmits.

LMK if the idea sounds good.

/hold

### DIFF
--- a/GENERATOR.md
+++ b/GENERATOR.md
@@ -48,7 +48,6 @@ following fields will not be overwritten if they are already present:
  - `optional`
  - `max_concurrency`
  - `skip_report`
- - `cluster`
 
 ## Postsubmits
 
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -525,6 +525,28 @@ func generatePeriodicForTest(name string, info *prowgenInfo, label jc.ProwgenLab
 	}
 }
 
+var (
+	clusterMap = map[string]string{
+		"openshift/ci-ns-ttl-controller":           "ci/api-build01-ci-devcluster-openshift-com:6443",
+		"openshift/ci-secret-mirroring-controller": "ci/api-build01-ci-devcluster-openshift-com:6443",
+	}
+)
+
+func getBuildClusterForPresubmit(org, repo string, e2e bool) string {
+	if e2e {
+		return ""
+	}
+	return clusterMap[fmt.Sprintf("%s/%s", org, repo)]
+}
+
+func getBuildClusterForPostsubmit(org, repo string) string {
+	return ""
+}
+
+func getBuildClusterForPeriodic(org, repo string) string {
+	return ""
+}
+
 // Given a ci-operator configuration file and basic information about what
 // should be tested, generate a following JobConfig:
 //
@@ -543,9 +565,11 @@ func generateJobs(
 
 	for _, element := range configSpec.Tests {
 		var podSpec *kubeapi.PodSpec
+		e2e := false
 		if element.ContainerTestConfiguration != nil {
 			podSpec = generateCiOperatorPodSpec(info, element.Secret, element.As)
 		} else {
+			e2e = true
 			var release string
 			if c := configSpec.ReleaseTagConfiguration; c != nil {
 				release = c.Name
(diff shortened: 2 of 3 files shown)


hongkailiu:
Maybe  `cluster` at root level to indicate all jobs are in that cluster?
I think that is less intrusive to the existing fields.

petr-muller:
I don't like this. The users (who write ci-op configs) do not care where the jobs run and IMO it should be transparent to them.

To work in the build cluster, the various image source/destination fields (`tag_specification`, `base_images` etc) will need the `cluster:` stanza (which I don't like for the same reason but at least that is something we already have) - I think Prowgen could act on the presence of these.

hongkailiu:
> I don't like this. The users (who write ci-op configs) do not care where the jobs run and IMO it should be transparent to them.

Maybe some user care. See https://jira.coreos.com/browse/DPTP-620. A user could bring their own cluster and say _I want to use my own cluster for build/test_.

If users do not care the cluster, just ignore it (we could use a default value for it: as today, the default value is api.ci while in the future it could be build03 ... or even do a roundrobin to the clusters in our build farm).


> To work in the build cluster, the various image source/destination fields (`tag_specification`, `base_images` etc) will need the `cluster:` stanza (which I don't like for the same reason but at least that is something we already have) - I think Prowgen could act on the presence of these.

Currently, prowgen does not even allow cluster as prowjob param which is a valid one from prow's point of view. We need at very least prowgen to go easy with this param (instead of failing the job when seeing cluster in the job definition).

petr-muller:
> > I don't like this. The users (who write ci-op configs) do not care where the jobs run and IMO it should be transparent to them.
> 
> Maybe some user cares. See [DPTP-620](https://jira.coreos.com/browse/DPTP-620). A user could bring their own cluster and say _I want to use my own cluster for build/test of my repos_.

Fair enough, but that's 0.000001% use case. What's the common case? What jobs will mostly run on the build cluster?

> If users do not care the cluster, just ignore it (we could use a default value for it: as today, the default value is api.ci while in the future it could be build03 ... or even do a roundrobin to the clusters in our build farm).
> 
> > To work in the build cluster, the various image source/destination fields (`tag_specification`, `base_images` etc) will need the `cluster:` stanza (which I don't like for the same reason but at least that is something we already have) - I think Prowgen could act on the presence of these.
> 
> Currently, prowgen does [not even allow cluster](https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/5797/pull-ci-openshift-release-master-generated-config/11438#1:build-log.txt%3A165) as prowjob param which is a valid one from prow's point of view. We need at very least prowgen to go easy with this param (instead of failing the job when seeing cluster in the job definition).

We could make Prowgen to not care about that field, just like we do it for `always_run:` etc.?

This boils down to a question: so which jobs should run on a build cluster? I lived under impression that our middle-term goal is "ideally all that can". If we make the plumbing work, then those that have 'cluster:' in their `tag_specification`, `base_images`, `promotion` can be run on a build cluster, so Prowgen may as well act on that.

hongkailiu:
> We could make Prowgen to not care about that field, just like we do it for `always_run:` etc.?

Ack. This is enough to unblock me. Will work on this today.

> This boils down to a question: so which jobs should run on a build cluster? I lived under impression that our middle-term goal is "ideally all that can". If we make the plumbing work, then those that have 'cluster:' in their `tag_specification`, `base_images`, `promotion` can be run on a build cluster, so Prowgen may as well act on that.

We do not need the answer right now. But since we started the discussion:

If I understand this at all:
`tag_specification`, `base_images`, `promotion`: those are ci-operator stuff ... prow does/need NOT know any of them.

prowgen: from ci-config to prowjobs: how to generate `cluster` of a prowjob? Answering this would answer _which jobs should run on a build cluster?_

In my option:
* user rule: if user defines a cluster to run in ci-config, prowgen uses it.
* prowgen rule1: act on existing cluster param (in `base_images` ...). I think conceptually it is not right: a build needs an image from `clusterX` does not necessarily mean the build has to happen on `clusterX`.
* prowgen rule2: 99.999...% users do not care. Then prowgen defines a cluster (which is configurable).

stevekuznetsov:
> prowgen rule1: act on existing cluster param (in base_images ...). I think conceptually it is not right: a build needs an image from clusterX does not necessarily mean the build has to happen on clusterX.

I read it the other way -- _if_ a configuration already has cluster specified for the images it pulls, it is safe for ProwGen to put it on _any_ cluster

hongkailiu:
> > prowgen rule1: act on existing cluster param (in base_images ...). I think conceptually it is not right: a build needs an image from clusterX does not necessarily mean the build has to happen on clusterX.
> 
> I read it the other way -- _if_ a configuration already has cluster specified for the images it pulls, it is safe for ProwGen to put it on _any_ cluster

This makes sense.

petr-muller:
> I read it the other way -- if a configuration already has cluster specified for the images it pulls, it is safe for ProwGen to put it on any cluster

That's exactly what I meant.

hongkailiu:
Before we migrate other jobs ... about `prowgen rule1`

Maybe `ci-operator` uses the default value of cluster when it is not defined by user?

hongkailiu:
Temporary blocked.
Want to go through e2e test first.

PR #261, opened 2019-11-07
Specify the full name of ci-operator image

hongkailiu:

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -182,7 +182,7 @@ func generatePodSpec(info *prowgenInfo, secret *cioperatorapi.Secret) *kubeapi.P
 		ServiceAccountName: "ci-operator",
 		Containers: []kubeapi.Container{
 			{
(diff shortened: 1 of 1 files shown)


hongkailiu:
/hold 

@petr-muller @stevekuznetsov 

In order to run pods on any build cluter ... I think the full name of image is required.
What is the best way of doing this?

petr-muller:
I think this is acceptable (explicit is good). The alternative would be to keep an ImageStream tracking these images on each build cluster, which feels "cleaner". /shrug

hongkailiu:
> I think this is acceptable (explicit is good). The alternative would be to keep an ImageStream tracking these images on each build cluster, which feels "cleaner". /shrug

ImageStream sounds like a better idea. Let me give it a try with auto schedule.
https://docs.openshift.com/container-platform/3.11/architecture/core_concepts/builds_and_image_streams.html#image-streams

hongkailiu:
/close
in favor of https://github.com/openshift/release/pull/5825

PR #262, opened 2019-11-07
prowgen: Generate private jobs as `hidden: true`

petr-muller:
This is a low-cost way how to make sure private and public jobs are shown in the correct Deck instance. This way we do not need to sync repos to the `hidden_repos` config for deck, or implement `hidden_orgs` yet.

We may eventually decide to implement `hidden_orgs` anyway, as it's the most clean solution, but this saves us from it being urgent.

/cc @droslean @openshift/openshift-team-developer-productivity-test-platform


petr-muller:
Expecting this to fail `breaking changes` for the testing private job.

PR #267, opened 2019-11-13
[WIP] pj-rehearse: implement rehearsing of change registry steps

AlexNPavel:
Early WIP implementation of step registry rehearsal support in pj-rehearse. Description will be updated after some cleanup and implementation of updated integration tests

### DIFF
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -7,6 +7,8 @@ import (
 	"strings"
 
 	"github.com/openshift/ci-tools/pkg/api"
+	registryutil "github.com/openshift/ci-tools/pkg/registry"
+
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -265,28 +267,6 @@ func GetChangedRegistrySteps(path, baseRev string, refs map[string]api.LiteralTe
 	return trimmedChanges, nil
 }
 
-// unrollChains expands/unrolls all chains recursively into references.
-// Borrowed from pkg/resolver/resolver.go
-func unrollChains(input []api.TestStep, fullRegistry registry) (unrolledSteps []api.TestStep, errs []error) {
-	for _, step := range input {
-		if step.Chain != nil {
-			chain, ok := fullRegistry.chains[*step.Chain]
-			if !ok {
-				return []api.TestStep{}, []error{fmt.Errorf("unknown step chain: %s", *step.Chain)}
-			}
-			// handle nested chains
-			chain, err := unrollChains(chain, fullRegistry)
-			if err != nil {
-				errs = append(errs, err...)
-			}
-			unrolledSteps = append(unrolledSteps, chain...)
-			continue
-		}
-		unrolledSteps = append(unrolledSteps, step)
-	}
-	return
-}
-
 // getNestedChainsFromSteps returns the names of all chains nested inside other chains in the provided RegistryChanges
 func getNestedChainsFromSteps(chains []api.TestStep, fullRegistry registry) []string {
 	// make a set of chain names to prevent duplicates
@@ -442,7 +422,7 @@ func trimRegistryChanges(changes []RegistryStep, fullRegistry registry) []Regist
 		steps := chain.Steps
 		chainSteps = append(chainSteps, steps...)
 	}
-	unrolledChainSteps, errs := unrollChains(chainSteps, fullRegistry)
+	unrolledChainSteps, errs := registryutil.UnrollChains(chainSteps, fullRegistry.chains)
 	if len(errs) != 0 {
 		// handle this
 	}
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -231,9 +231,9 @@ func loadRegistryStep(filename string, fullRegistry registry) (RegistryStep, err
 
 // GetChangedRegistrySteps identifies all registry components (refs, chains, and workflows) that changed.
 // After all changes have been identified, the list gets "trimmed": all changed refs that are included in
-// changed refs are removed from the list since the refs will be tested when the chain gets tested. This
-// is done for changed chains included in changed workflows. This helps reduce the number of rehearsals
-// that are performed.
+// changed chains are removed from the list since the refs will be tested when the chain gets tested. This
+// is also done for changed chains and refs included in changed workflows. This helps reduce the number of
+// rehearsals that are performed.
 func GetChangedRegistrySteps(path, baseRev string, refs map[string]api.LiteralTestStep, chains map[string][]api.TestStep, workflows map[string]api.MultiStageTestConfiguration) ([]RegistryStep, error) {
 	revChanges, err := getRevChanges(path, RegistryPath, baseRev, true)
 	if err != nil {
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/sirupsen/logrus"
 
 	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
@@ -36,12 +37,28 @@ const (
 	ClusterProfilePrefix = "cluster-profile-"
 	// StagingNamespace is the staging namespace in api.ci
 	StagingNamespace = "ci-stg"
+	// RegistryPath is the path to the multistage step registry
+	RegistryPath = "ci-operator/step_registry"
 )
 
 // ReleaseRepoConfig contains all configuration present in release repo (usually openshift/release)
 type ReleaseRepoConfig struct {
 	Prow       *prowconfig.Config
 	CiOperator ByFilename
+}
+
+// RegistryStep includes the 3 types of components of the multistage registry as pointers
+// and is used to store all registry changes using a single struct that can be put into a slice.
+type RegistryStep struct {
+	Reference *api.RegistryReference
+	Chain     *api.RegistryChain
+	Workflow  *api.RegistryWorkflow
+}
+
+type registry struct {
+	refs      map[string]api.LiteralTestStep
+	chains    map[string][]api.TestStep
+	workflows map[string]api.MultiStageTestConfiguration
 }
 
 func git(repoPath string, args ...string) (string, error) {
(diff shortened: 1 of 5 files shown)


pkg/config/release.go
```diff
@@ -163,6 +183,340 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+	}
+	if change == nil {
+		return RegistryChanges{}, fmt.Errorf("could not find component")
+	}
+	return *change, nil
+}
+
+// GetChangedRegistrySteps identifies all registry components (refs, chains, and workflows) that changed.
+// After all changes have been identified, the list gets "trimmed": all changed refs that are included in
+// changed refs are removed from the list since the refs will be tested when the chain gets tested. This
```
stevekuznetsov:
```
all changed refs that are included in changed refs
```
?

pkg/config/release.go
```diff
@@ -163,6 +183,340 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+			}
+			registryChanges = append(registryChanges, change)
+		}
+	}
+	trimmedChanges := trimRegistryChanges(registryChanges, fullRegistry)
+	return trimmedChanges, nil
+}
+
+// unrollChains expands/unrolls all chains recursively into references.
+// Borrowed from pkg/resolver/resolver.go
```
stevekuznetsov: why does it need to be borrowed?

pkg/config/release.go
```diff
@@ -163,6 +183,340 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+		}
+		unrolledSteps = append(unrolledSteps, step)
+	}
+	return
+}
+
+// getNestedChainsRecursive returns the names of all chains nested inside other chains in the provided RegistryChanges
+func getNestedChainsRecursive(chains []api.TestStep, fullRegistry registry) []string {
+	// make a set of chain names to prevent duplicates
+	nestedChains := make(map[string]bool)
```
stevekuznetsov: use `sets.String`

pkg/config/release.go
```diff
@@ -163,6 +183,340 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+			unrolledSteps = append(unrolledSteps, chain...)
+			continue
+		}
+		unrolledSteps = append(unrolledSteps, step)
+	}
+	return
+}
+
+// getNestedChainsRecursive returns the names of all chains nested inside other chains in the provided RegistryChanges
+func getNestedChainsRecursive(chains []api.TestStep, fullRegistry registry) []string {
```
stevekuznetsov: nit: it's not necessary or useful to tell people this is recursive in the method name

pkg/config/release.go
```diff
@@ -163,6 +183,340 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
 	return ret, nil
 }
 
+// generateCommandToSteps creates a map that maps command filenames to the step that refers to them
```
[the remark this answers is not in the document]
AlexNPavel: This command is now removed as it is not necessary with registry validation

pkg/config/release.go
```diff
@@ -163,6 +183,340 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+func generateCommandToSteps(fullRegistry registry) map[string]string {
+	var commandToSteps map[string]string
+	for name, step := range fullRegistry.refs {
+		commandToSteps[step.Commands] = name
+	}
+	return commandToSteps
+}
+
+func getStepsFromCommands(filename, sha string, cToS map[string]string) (string, error) {
```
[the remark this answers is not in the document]
AlexNPavel: This command is now removed as it is not necessary with registry validation

pkg/config/release.go
AlexNPavel: Currently, there is an `unrollChains` function that is almost identical in `pkg/registry`, but it is not exported and it is a function of the internal registry type in `pkg/registry`. Looking at the function, it doesn't actually modify the internal `registry` type that it operates on and it only needs the map of chains, so it could be easily exported.

I'll export that function and make the modifications needed so it doesn't need to be duplicated.

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
 	return ret, nil
 }
 
+func loadRegistryStep(filename string, fullRegistry registry) (RegistryStep, error) {
+	if strings.HasSuffix(filename, "-commands.sh") {
```
stevekuznetsov: nit: this could be 

```go
filename = strings.Replace(filename, "-commands.sh", "-ref.yaml")
```

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+// After all changes have been identified, the list gets "trimmed": all changed refs that are included in
+// changed chains are removed from the list since the refs will be tested when the chain gets tested. This
+// is also done for changed chains and refs included in changed workflows. This helps reduce the number of
+// rehearsals that are performed.
+func GetChangedRegistrySteps(path, baseRev string, refs map[string]api.LiteralTestStep, chains map[string][]api.TestStep, workflows map[string]api.MultiStageTestConfiguration) ([]RegistryStep, error) {
+	revChanges, err := getRevChanges(path, RegistryPath, baseRev, true)
+	if err != nil {
+		return nil, err
+	}
+	var registryChanges []RegistryStep
```
stevekuznetsov: this seems misplaced? or should it have had a load line? Why is there an error check afterword?

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to load registry: %v", err)
+	}
+	fullRegistry := registry{refs: refs, chains: chains, workflows: workflows}
+	seenReference := make(map[string]bool)
```
stevekuznetsov: This should still be a `sets.String`

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+		for _, step := range steps {
+			// ignore steps that aren't chains
+			if step.Chain == nil {
+				continue
+			}
+			// add nested chain to set
+			nestedChains.Insert(*step.Chain)
+			// recursively check for nested chains and add those to the nested chain set
+			for _, chain := range getNestedChainsFromSteps(fullRegistry.chains[*step.Chain], fullRegistry) {
+				nestedChains.Insert(chain)
```
stevekuznetsov: all of these could be

```go
nestedChains.Insert(*step.Chain, getNestedChainsFromSteps(fullRegistry.chains[*step.Chain], fullRegistry)...)
```

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+			}
+		}
+	}
+	// convert set into slice
```
stevekuznetsov: you *could* return `nestedChains.List()` but I suggest you return the set since you can use it's methods later

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+	combinedChains := append(chains, resolvedWFChains...)
+	nestedChains := getNestedChains(combinedChains, fullRegistry)
+	var trimmedChanges []RegistryStep
+	// we still want a []api.TestStep of non-nested chains for later steps
+	chains = []api.RegistryChain{}
+	for _, change := range changes {
+		if change.Chain != nil {
+			ignore := false
+			for _, nested := range nestedChains {
+				if change.Chain.As == nested {
```
stevekuznetsov: All of these types of checks are cleaners with sets and `.Contains()` or `ContainsAny()`

pkg/config/release.go
```diff
@@ -163,6 +183,278 @@ func GetChangedTemplates(path, baseRev string) ([]ConfigMapSource, error) {
...
+	}
+	// get references that exist in changed chains
+	var usedReferences []string
+	for _, step := range unrolledChainSteps {
+		if step.Reference != nil {
+			usedReferences = append(usedReferences, *step.Reference)
+		}
+	}
+	usedReferences = append(usedReferences, wfRefs...)
+	// remove references that are used in changed chains and workflows
```
stevekuznetsov: You can rebuild the slice to actually remove things instead of redeclaring a new slice and creating it -- https://github.com/golang/go/wiki/SliceTricks#delete

pkg/registry/resolver.go
```diff
@@ -101,15 +101,16 @@ func (r *registry) process(steps []api.TestStep) (literalSteps []api.LiteralTest
 	return
 }
 
-func (r *registry) unrollChains(input []api.TestStep) (unrolledSteps []api.TestStep, errs []error) {
+// UnrollChains expands/unrolls all chains recursively into references.
```
stevekuznetsov: You don't unroll into references, you unroll into steps

pkg/rehearse/jobs.go
```diff
@@ -396,6 +396,64 @@ func AddRandomJobsForChangedTemplates(templates []config.ConfigMapSource, toBeRe
 	return rehearsals
 }
 
+func getPresubmitForRegistryStep(step config.RegistryStep, configs config.ByFilename, prConfigPresubmits map[string][]prowconfig.Presubmit) (string, *prowconfig.Presubmit) {
+	for _, ciopConfig := range configs {
+		tests := ciopConfig.Configuration.Tests
+		repoPresubmits := prConfigPresubmits[ciopConfig.Info.Repo]
+		for _, test := range tests {
+			if test.MultiStageTestConfiguration != nil {
+				jobPrefix := fmt.Sprintf("%s-ci-%s-%s-%s-", "pull", ciopConfig.Info.Org, ciopConfig.Info.Repo, ciopConfig.Info.Branch)
```
stevekuznetsov: Can we DRY this out with the logic that actually creates the names to ensure they stay in sync?

pkg/rehearse/jobs.go
```diff
@@ -396,6 +396,64 @@ func AddRandomJobsForChangedTemplates(templates []config.ConfigMapSource, toBeRe
 	return rehearsals
 }
 
+func getPresubmitForRegistryStep(step config.RegistryStep, configs config.ByFilename, prConfigPresubmits map[string][]prowconfig.Presubmit) (string, *prowconfig.Presubmit) {
+	for _, ciopConfig := range configs {
```
stevekuznetsov: You get random access to the elements of a map so this function will return a random output every time -- we want it to be deterministic so that when you /retest it picks the same one

pkg/rehearse/jobs.go
```diff
@@ -396,6 +396,64 @@ func AddRandomJobsForChangedTemplates(templates []config.ConfigMapSource, toBeRe
...
+				jobPrefix := fmt.Sprintf("%s-ci-%s-%s-%s-", "pull", ciopConfig.Info.Org, ciopConfig.Info.Repo, ciopConfig.Info.Branch)
+				jobName := fmt.Sprintf("%s%s", jobPrefix, test.As)
+				if step.Workflow != nil && test.MultiStageTestConfiguration.Workflow != nil && step.Workflow.As == *test.MultiStageTestConfiguration.Workflow {
+					for _, presubmit := range repoPresubmits {
+						if presubmit.Name == jobName {
+							return ciopConfig.Info.Repo, &presubmit
+						}
+					}
+					// continue to check other tests
+					continue
```
stevekuznetsov: The control flow here is a bit odd. I suggest using a `switch` to make the algorithm more clear -- you have already pruned the diff so each change is unique, so every job you look at can match only one of the changed items (workflow,chain,ref)

PR #268, opened 2019-11-13
pkg/load: add validation during registry loading

AlexNPavel:
Add validation during registry loading. This makes the registry contents much more consistent and prevents naming collisions. It should also significantly reduce the complexity of implementing rehearsals for the changes in the step_registry.

This also adds error logging for registry loading errors that occur in the reload watcher and fixes a bug in the configresolver integration tests.

### DIFF
--- a/pkg/load/agent_utils.go
+++ b/pkg/load/agent_utils.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/fsnotify/fsnotify"
+	"github.com/sirupsen/logrus"
 	log "github.com/sirupsen/logrus"
 
 	"github.com/openshift/ci-tools/pkg/coalescer"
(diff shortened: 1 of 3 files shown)


pkg/load/agent_utils.go
```diff
@@ -37,7 +38,12 @@ func reloadWatcher(ctx context.Context, w *fsnotify.Watcher, root string, errFun
 			return
 		case event := <-w.Events:
 			log.Tracef("Received %v event for %s", event.Op, event.Name)
-			go c.Run()
+			go func() {
+				err := c.Run()
+				if err != nil {
+					logrus.WithError(err).Errorf("Coalescer function failed")
```
stevekuznetsov: Do we care to track this in metrics?

AlexNPavel: We probably should. That function already has `errFunc`, which is the func to report the error to the resolver metrics, so that should be easy to implement.

AlexNPavel:
@stevekuznetsov We don't yet have a presubmit that loads the registry. The configresolver does have metrics for the registry though, it just needs to be added to grafana (that being said, we don't provide it a registry as one doesn't exist yet, so I don't know if we're actually getting metrics at the moment)

PR #272, opened 2019-11-15
prowgen: Generate promoting postsubmit on only additional_images

petr-muller:
Previously, when `additional_images` were present but `images` were not,
prowgen did not create a promoting postsubmit.


petr-muller:
Expecting breaking-changes to fail, https://github.com/openshift/release/pull/5937 is the followup that needs to merge after new prowgen is built.

petr-muller:
I have tested that running `ci-operator --target=[images]` is okay even for configs with no images, so these postsubmits should be fine.

PR #273, opened 2019-11-15
Validate flags in the configresolver, add a validator mode

stevekuznetsov:
Running the configresolver without config or registry directories is
going to fail anyway so we should validate those on startup to make it
clear to the user. A new `--validate-only` flag allows running the
binary to validate config on disk but not running the server.

/assign @AlexNPavel @bbguimaraes

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -34,6 +34,7 @@ type options struct {
 	address      string
 	gracePeriod  time.Duration
 	cycle        time.Duration
+	validateOnly bool
 }
 
 type traceResponseWriter struct {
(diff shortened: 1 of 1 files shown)


cmd/ci-operator-configresolver/main.go
```diff
@@ -106,6 +108,20 @@ func validateOptions(o options) error {
...
+		if _, err := os.Stat(o.configPath); err != nil && os.IsNotExist(err) {
+			return fmt.Errorf("--config points to a nonexistent directory: %v", err)
+		}
+	}
+	if o.registryPath == "" {
+		return fmt.Errorf("--registry is required")
+	} else {
+		if _, err := os.Stat(o.registryPath); err != nil && os.IsNotExist(err) {
+			return fmt.Errorf("--registry points to a nonexistent directory: %v", err)
+		}
```
bbguimaraes: Don't we already check this when creating the config/registry agents?

stevekuznetsov: Yeah but you get it as some semi-related error that's harder to interpret. This is useful for users since it tells them that the flag they passed was bad. Right now you see this:

```
FATA[0000] Failed to get registry agent: Failed to load registry: failed to load ci-operator registry (lstat : no such file or directory) 
```

bbguimaraes: That's covered by the `if` part of the conditional, isn't it? That is fine.

If we get a bad error message when there's a problem reading the configuration/registry ("the file/directory doesn't exist" being one of the possibilities, which can also happen on reloads), we should fix that instead of adding this kind of code to every user.

PR #279, opened 2019-11-16
pkg/registry: add registry graph

AlexNPavel:
This adds a registry graph, which allows a user to identify all ancestors or descendants of a registry component as well as detect any cycles in the chains.

Will be used in pj-rehearse and greatly simplify #267

### DIFF
--- a/pkg/registry/graph.go
+++ b/pkg/registry/graph.go
@@ -7,8 +7,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-// Type declarations
-
 // Type identifies the type of registry element a Node refers to
 type Type int
 
@@ -82,8 +80,6 @@ type workflowNodeByName map[string]*workflowNode
 type chainNodeByName map[string]*chainNode
 type referenceNodeByName map[string]*referenceNode
 
-// Set functions
-
 func (set workflowNodeSet) insert(node *workflowNode) {
 	set[node] = sets.Empty{}
 }
@@ -120,14 +116,9 @@ func (set referenceNodeSet) list() []*referenceNode {
 	return res
 }
 
-// Interface Functions
-// Name function
-
 func (n *nodeWithName) Name() string {
 	return n.name
 }
-
-// Type functions
 
 func (*workflowNode) Type() Type {
 	return Workflow
@@ -141,8 +132,6 @@ func (*referenceNode) Type() Type {
 	return Reference
 }
 
-// ParentNames functions
-
 func (n *nodeWithParents) ParentNames() sets.String {
 	parents := sets.NewString()
 	for parent := range n.workflowParents {
@@ -154,9 +143,7 @@ func (n *nodeWithParents) ParentNames() sets.String {
 	return parents
 }
 
-func (*workflowNode) ParentNames() sets.String { return nil }
-
-// ChildrenNames functions
+func (*workflowNode) ParentNames() sets.String { return sets.NewString() }
 
 func (n *nodeWithChildren) ChildrenNames() sets.String {
 	children := sets.NewString()
@@ -169,9 +156,7 @@ func (n *nodeWithChildren) ChildrenNames() sets.String {
 	return children
 }
 
-func (*referenceNode) ChildrenNames() sets.String { return nil }
-
-// AncestorNames functions
+func (*referenceNode) ChildrenNames() sets.String { return sets.NewString() }
 
 func (n *nodeWithParents) AncestorNames() sets.String {
 	ancestors := n.ParentNames()
@@ -181,9 +166,7 @@ func (n *nodeWithParents) AncestorNames() sets.String {
 	return ancestors
 }
 
-func (*workflowNode) AncestorNames() sets.String { return nil }
-
-// DescendantNames function
+func (*workflowNode) AncestorNames() sets.String { return sets.NewString() }
 
 func (n *nodeWithChildren) DescendantNames() sets.String {
 	descendants := n.ChildrenNames()
@@ -193,10 +176,7 @@ func (n *nodeWithChildren) DescendantNames() sets.String {
 	return descendants
 }
 
-func (*referenceNode) DescendantNames() sets.String { return nil }
-
-// Struct helper functions
-// addChild functions
+func (*referenceNode) DescendantNames() sets.String { return sets.NewString() }
 
 func (n *workflowNode) addChainChild(child *chainNode) {
 	n.chainChildren.insert(child)
@@ -238,7 +218,7 @@ func newNodeWithChildren() nodeWithChildren {
 
 func hasCycles(node *chainNode, ancestors sets.String, traversedPath []string) error {
 	if ancestors == nil {
-		ancestors = make(sets.String)
+		ancestors = sets.NewString()
 	}
 	if ancestors.Has(node.name) {
 		return fmt.Errorf("Cycle detected: %s is an ancestor of itself; traversedPath: %v", node.name, append(traversedPath, node.name))
@@ -249,10 +229,7 @@ func hasCycles(node *chainNode, ancestors sets.String, traversedPath []string) e
 			continue
 		}
 		// get new copy of ancestors and traversedPath so the root node's set isn't changed
-		ancestorsCopy := sets.NewString()
-		for _, ancestor := range ancestors.List() {
-			ancestorsCopy.Insert(ancestor)
-		}
+		ancestorsCopy := sets.NewString(ancestors.UnsortedList()...)
 		traversedPathCopy := append(traversedPath[:0:0], traversedPath...)
 		traversedPathCopy = append(traversedPathCopy, node.name)
 		if err := hasCycles(child, ancestorsCopy, traversedPathCopy); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/registry/graph.go
```diff
@@ -0,0 +1,176 @@
...
+
+import (
+	"fmt"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Component contains the name of a registry component and pointers for
+// either a reference, chain, or workflow
```
stevekuznetsov: We know we have strong one-way relationships (a child must be a chain or ref, a parent must be a chain or workflow and workflows must be roots) so I think we could do better with typing here.

pkg/registry/graph.go
```diff
@@ -0,0 +1,176 @@
...
+		if err := hasCycles(child, ancestorsCopy, traversedPathCopy); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// HasCycles detects any cycles in the descendants of the provided root node
+func HasCycles(root *Node) error {
+	return hasCycles(root, sets.NewString(), []string{})
```
stevekuznetsov: This assumes a tree, not a forest

AlexNPavel: In `NewGraph`, a `root` node with an empty name field is created whose children are all nodes that don't have any parents after everything is resolved. If we pass the `root` node to `HasCycles`, it should be able to traverse the entire graph and find all cycles. Another option that would work would be to pass the `NodesByName` type, but either way should work.

stevekuznetsov: Why should we prefer a synthetic root with no content versus allowing a forest?

pkg/registry/graph.go
AlexNPavel: We need to have a quick way to access any node by name if we want to be able to get all parents/descendants quickly, and being able to put any of these into a single type that can be put into a map makes this simple. Otherwise, we would need to have 3 maps again, one for each of refs, chains, and workflows to be able to find the nodes we want quickly.

One thing we could do is just have the name and an enum for the type instead of 3 pointers. This would simplify the `Component` type, but would require us to have access to the refs, chains, and workflow maps if we need to access the contents of one of the components (vs the current type, which just points to the contents directly).

pkg/registry/graph.go
AlexNPavel: It doesn't really matter, that's just how I first thought to do it (I've worked with trees a lot more than graphs/forests so this made sense to me). We could just not have a single root and use only the `NodesByName` map to access the graph's contents.

stevekuznetsov: IMO synthetic valueless roots for forests often get mistaken by someone who didn't write the implementation in the first place so I'd suggest not doing that.

pkg/registry/graph.go
stevekuznetsov: I think specifically my feedback here is really similar to the internal/external types. We are using a strongly typed language and all we do by fudging these sorts of declarations is we make possible in code strictly forbidden combinations. This can lead to programmer bugs and frustration when implicitly invalid combinations either happen in practice, become explicitly possible and need refactoring, etc. If you think that there is no simple alternative to this type, I think it's reasonable to fudge it but in my opinion I'd prefer to invest some nontrivial amount of effort creating a tight type system that doesn't let us mess something up down the road. Of course if the upfront development cost is great, then it's a crapshoot.

AlexNPavel:
This PR has been mostly rewritten. All node types have more explicit values and have strong relationships.

pkg/registry/graph.go
```diff
@@ -0,0 +1,295 @@
...
+const (
+	Workflow NodeType = iota
+	Chain
+	Reference
+)
+
+// Node is an interface that allows a user to identify ancestors and descendants of a step registry element
+type Node interface {
+	Name() string
+	GetNodeType() NodeType
```
stevekuznetsov: super nit: this could be `Type()`

pkg/registry/graph.go
```diff
@@ -0,0 +1,295 @@
...
+	GetAncestorNames() sets.String
```
stevekuznetsov: super nit: no need for the verb in the method name, could be `DescendatNames()`

pkg/registry/graph.go
stevekuznetsov: See the comment on stuttering here https://blog.golang.org/package-names (although this specific case in this PR is not a package stutter, but the concept is the same)

pkg/registry/graph.go
```diff
@@ -0,0 +1,295 @@
...
+	nodeWithChildren
+}
+
+type referenceNode struct {
+	nodeWithName
+	nodeWithParents
+}
+
+// Verify that all node types implement Node
+var _ = Node(&workflowNode{})
```
stevekuznetsov: Nit: this is usually written as 

```go
var _ Node = &workflowNode{}
```

pkg/registry/graph.go
```diff
@@ -0,0 +1,326 @@
+package registry
+
+import (
+	"fmt"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+// Type declarations
```
stevekuznetsov: nit: probably not useful comments?

pkg/registry/graph.go
```diff
@@ -0,0 +1,326 @@
...
+
+// Type identifies the type of registry element a Node refers to
+type Type int
+
+const (
+	Workflow Type = iota
+	Chain
```
stevekuznetsov: @bbguimaraes is this the weird Go thing where this `const` is not of the type you'd expect?

pkg/registry/graph.go
```diff
@@ -0,0 +1,326 @@
...
+	for parent := range n.workflowParents {
+		parents.Insert(parent.Name())
+	}
+	for parent := range n.chainParents {
+		parents.Insert(parent.Name())
+	}
+	return parents
+}
+
+func (*workflowNode) ParentNames() sets.String { return nil }
```
stevekuznetsov: Is `nil` functionally equivalent to an empty set?

pkg/registry/graph.go
```diff
@@ -0,0 +1,326 @@
...
+func newNodeWithChildren() nodeWithChildren {
+	return nodeWithChildren{
+		chainChildren:     make(chainNodeSet),
+		referenceChildren: make(referenceNodeSet),
+	}
+}
+
+func hasCycles(node *chainNode, ancestors sets.String, traversedPath []string) error {
+	if ancestors == nil {
+		ancestors = make(sets.String)
```
stevekuznetsov: nit: `sets.NewString()`

pkg/registry/graph.go
```diff
@@ -0,0 +1,326 @@
...
+	for child := range node.chainChildren {
+		if child.Type() != Chain {
+			continue
+		}
+		// get new copy of ancestors and traversedPath so the root node's set isn't changed
+		ancestorsCopy := sets.NewString()
+		for _, ancestor := range ancestors.List() {
+			ancestorsCopy.Insert(ancestor)
+		}
+		traversedPathCopy := append(traversedPath[:0:0], traversedPath...)
```
stevekuznetsov: :wizardry:

pkg/registry/graph.go
```diff
@@ -0,0 +1,326 @@
...
+	if ancestors.Has(node.name) {
+		return fmt.Errorf("Cycle detected: %s is an ancestor of itself; traversedPath: %v", node.name, append(traversedPath, node.name))
+	}
+	ancestors.Insert(node.name)
+	for child := range node.chainChildren {
+		if child.Type() != Chain {
+			continue
+		}
+		// get new copy of ancestors and traversedPath so the root node's set isn't changed
+		ancestorsCopy := sets.NewString()
```
stevekuznetsov: super nit: could this be

```go
ancestorsCopy := sets.NewString(ancestors.UnsortedList()...)

Comment (stevekuznetsov) [reply pkg/registry/graph.go]: (unsorted list saves performance where we don't care)

File: pkg/registry/graph.go [hunk 12]
```diff
@@ -0,0 +1,326 @@
...
+	nodesByName := make(NodeByName)
+	// References can only be children; load them so they can be added as children by workflows and chains
+	referenceNodes := make(referenceNodeByName)
+	for name := range stepsByName {
+		node := &referenceNode{
+			nodeWithName:    newNodeWithName(name),
+			nodeWithParents: newNodeWithParents(),
+		}
+		referenceNodes[name] = node
+		nodesByName[name] = Node(node)
```
stevekuznetsov: nit: the duck type will make the cast implicit so you don't need to specify it

pkg/registry/graph.go
AlexNPavel: Would it be better to just make a `for` loop to clone the slice? It's slower but more readable. I don't think we'll get datasets so large that performance here will be much of an issue.

stevekuznetsov: I think it's fine to leave as is

stevekuznetsov: If we cared about performance we'd also switch all the `List()` to `UnsortedList()` and sort post-facto in tests or on display etc where we care

pkg/registry/graph.go
bbguimaraes: Usually yes, but `iota` seems to makes it work: https://play.golang.org/p/2VVuVZZZJz5.

PR #280, opened 2019-11-17
Update with custom metadata during CI job failure.

meowfaceman:
Custom prow metadata will not be added to the metadata file if a job
fails at present. This has been fixed.

Note: I've noticed that the updated metadata is not currently being written out when job errors occur, which isn't really that useful, as it prevents otherwise useful metadata from being written that could help debug/diagnose issues. As a result, I've moved to treat this more like the junit writing mechanism. Please feel free to critique.


meowfaceman:
cc @stevekuznetsov as he was pretty involved in the feedback with the original review.

PR #287, opened 2019-11-18
convert metadata revision to string and refactor the tests

droslean:
after #252 the jobs couldn't report to testgrid.  This patch is fixing that.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -245,7 +245,6 @@ func bindOptions(flag *flag.FlagSet) *options {
 	opt := &options{
 		idleCleanupDuration: 1 * time.Hour,
 		cleanupDuration:     12 * time.Hour,
-		metadataRevision:    0,
 	}
 
 	// command specific options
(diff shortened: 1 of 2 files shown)


droslean:
closing... Expecting changes from testgrid side to support both `int` and `string` revisions. 

/close

droslean:
@meowfaceman FYI

meowfaceman:
Sorry about that -- that's an unexpected problem here. Thanks for following up on this.

meowfaceman:
Sorry, so thinking on this: what's the problem, exactly? I would have figured testgrid could have handled this. We're definitely still seeing results here: https://testgrid.k8s.io/redhat-osd-int#osd-int-4.2

droslean:
testgrid magic :D

petr-muller:
> testgrid magic :D

I would be quite interested in more detailed reason too... How was this diagnosed?

stevekuznetsov:
Michelle was able to look at the (internal) logs for the process that reads these files and see the error.

droslean:
> I would be quite interested in more detailed reason too... How was this diagnosed?

@petr-muller Testgrid was able to find the other job because it had another revision that it could read.
check [finished.json](https://storage.googleapis.com/origin-ci-test/logs/periodic-ci-openshift-osde2e-master-e2e-int-4.2/287/finished.json)

PR #286, opened 2019-11-18
Default artifacts directory, set env

stevekuznetsov:
When a user runs a test they should be able to set their artifacts
directory to something other than the default, but they should not need
to do anything special to get the default. All good tests create
artifacts so this should not make much of a difference for the amount of
containers we actually run.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @bbguimaraes 
/cc @openshift/openshift-team-developer-productivity-test-platform

PR #288, opened 2019-11-18
Expose shared directory for users in test continers

stevekuznetsov:
depends on #286 
/assign @bbguimaraes 
How were you expecting that users use this dir if it was not exposed via some var?

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -22,6 +22,7 @@ const (
 	multiStageTestLabel     = "ci.openshift.io/multi-stage-test"
 	clusterProfileMountPath = "/var/run/secrets/ci.openshift.io/cluster-profile"
 	secretMountPath         = "/var/run/secrets/ci.openshift.io/multi-stage"
+	secretMountEnv          = "SHARED_DIR"
 )
 
 type multiStageTestStep struct {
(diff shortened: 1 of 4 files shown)


bbguimaraes:
-0 on the necessity of this. I modeled those directories after `/var/run/secrets/kubernetes.io`. No one consults an environment variable to access that path, it is a standard.

stevekuznetsov:
In what context do people do that? I've personally never seen that before.

bbguimaraes:
I'm probably answering the wrong question, but e.g.:

https://github.com/kubernetes/test-infra/blob/affacb471620f11b9745c4e3cbf4cd3ef0d0f30f/prow/kube/client.go#L271-L301

It's even what the documentation recommends:

https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#directly-accessing-the-rest-api-1

bbguimaraes:
My comment was regarding

>  How were you expecting that users use this dir if it was not exposed via some var?

I'm not strictly against this change.

/lgtm
/hold

test/ci-operator-integration/multi-stage/run.sh
```diff
@@ -2,7 +2,8 @@
 set -euo pipefail
 
 WORKDIR="$(mktemp -d)"
-trap 'rm -rf "${WORKDIR}"' EXIT
+trap 'echo "${WORKDIR}"' EXIT
```
droslean: Is this a debugging leftover?

stevekuznetsov: Nice catch, fixed

Issue #284, opened 2019-11-18
ci-operator: additional images for publishing are not built if they are not used


stevekuznetsov:
See [this configuration](https://github.com/openshift/release/blob/a96eb16981761d553b89b7ed76688e72515372a6/ci-operator/config/openshift/origin/openshift-origin-master.yaml#L188-L191) from Origin:

```
tests:
- as: images-artifacts
  commands: '# noop, just to force the building the `artifacts` image'
  container:
    from: artifacts
```

We should build this in the `[images]` target, if we don't.

@tnozicka

stevekuznetsov:
For context, this was added in https://github.com/openshift/release/commit/eac0009b21528e8f269be8318dee44375c214d4d -- hopefully @tnozicka can give more context here.

/cc @openshift/openshift-team-developer-productivity-test-platform

petr-muller:
This is currently worked-around in Prowgen which generates jobs with additional explicit `--target` for all additional images, so this currently affects only the cases when someone runs `ci-operator --promote` outside of the generated jobs.

stevekuznetsov:
So we can just remove this fake presubmit from the origin config?

stevekuznetsov:
If we do that we can close this issue.

petr-muller:
I believe we can, and it should only cause the presubmit to be removed, but promoting postsubmit should stay the same, explicitly building `artifacts` so it can promote it.

stevekuznetsov:
Will the `images` presubmit build the image?

petr-muller:
Good point. It won't, but can be quite easily made to, I think.

stevekuznetsov:
OK, I would expect it to and I think that would allow us to close this issue and remove the fake presubmit.

PR #289, opened 2019-11-19
prowgen: build all promoted images in `images` presubmits

petr-muller:
Previously, it was possible to not attempt to build all promoted images
in any of the PR presubmits. Make the `images` presubmit target all
images that are promoted, including those in `additional_images`.

Resolves: https://github.com/openshift/ci-tools/issues/284

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -235,7 +235,7 @@ func generatePodSpec(info *prowgenInfo, secret *cioperatorapi.Secret) *kubeapi.P
 	}
 }
 
-func generateCiOperatorPodSpec(info *prowgenInfo, secret *cioperatorapi.Secret, target string, additionalArgs ...string) *kubeapi.PodSpec {
+func generateCiOperatorPodSpec(info *prowgenInfo, secret *cioperatorapi.Secret, targets []string, additionalArgs ...string) *kubeapi.PodSpec {
 	for _, arg := range additionalArgs {
 		if !strings.HasPrefix(arg, "--") {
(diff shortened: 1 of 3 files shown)


petr-muller:
This is expected to be a breaking change, followup PR is https://github.com/openshift/release/pull/5995

petr-muller:
/hold

The semantic check on openshift/release has some assumptions which are no longer valid, and will need to be adjusted.

petr-muller:
/hold cancel

I changed the generator so the `--target=[images]` is always passed, even if there are no `images:` in the ci-op config - it does not hurt anything and it makes everything a bit more explicit (an `...-images` job building, for example, only `--target=src` could be quite confusing). This made the semantic check happy, but I still relaxed its `ci-op jobs should only have one `--target` check in https://github.com/openshift/release/pull/6000.

petr-muller:
> Why should we prefer this approach to redefining what constitutes `[images]`?

Confused, I understood that's what we agreed on https://github.com/openshift/ci-tools/issues/284#issuecomment-555267260.

For the reasons: I think this is just less invasive, more explicit and it follows what was done earlier (we do multiple targets for the same reason in postsubmits already). `[images]` always meant just _"everything in `images:`"_, changing its meaning to _"everything that would be promoted"_ may have some unexpected consequences, although I'm not immediately aware of any.

stevekuznetsov:
Ah, makes sense. I thought `[images]` already meant something else -- but that's the release-flavor.

/lgtm
/approve

stevekuznetsov:
/hold 

Merge when we have the `openshift/release` PR in flight.

petr-muller:
> Merge when we have the `openshift/release` PR in flight.

I filed https://github.com/openshift/release/pull/5995 together with this one

PR #291, opened 2019-11-20
pj-rehearse: resolve multi-stage configurations

AlexNPavel:
Resolve inlined ci-operator configs in pj-rehearse.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -32,7 +32,6 @@ import (
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/pjutil"
 
-	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/registry"
 )
@@ -230,7 +229,7 @@ func filterJobSpec(spec *v1.PodSpec, allowVolumes bool) error {
 // of the needed config file passed to the job as a direct value. This needs
 // to happen because the rehearsed Prow jobs may depend on these config files
 // being also changed by the tested PR.
-func inlineCiOpConfig(container v1.Container, ciopConfigs config.ByFilename, loggers Loggers) error {
+func inlineCiOpConfig(container v1.Container, ciopConfigs config.ByFilename, resolver registry.Resolver, loggers Loggers) error {
 	for index := range container.Env {
 		env := &(container.Env[index])
 		if env.ValueFrom == nil {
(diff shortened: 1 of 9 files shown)


AlexNPavel:
The implementation for this is ready for review, but it needs more tests to verify it works correctly. I will also be moving the function that removes the configresolver flags from rehearsals to the new `resolveInlineCiOpConfig` function.

AlexNPavel:
This PR is ready for review

pkg/rehearse/jobs.go
```diff
@@ -337,10 +341,56 @@ func (jc *JobConfigurer) ConfigurePresubmitRehearsals(presubmits config.Presubmi
 	return rehearsals
 }
 
+func resolveInlineCiOpConfig(container v1.Container, resolver registry.Resolver, loggers Loggers) error {
```
stevekuznetsov: I was hoping we could DRY this out with `configFromResolver` -- some abstraction around `func(org, repo, branch string) ReleaseBuildConfiguration` with either a local resolution or HTTP request as a backing

AlexNPavel: I think that would be very messy, because you would then need to provide a resolver and either a map of configs or path to configs for local resolution, all of which you don't have when doing remote resolution. The handling of the 2 cases shares no common code, so it would be unreasonable to handle both in the same function IMO.
There is a lot of shared resolution code in the confgresolver's `main.go` and in the `jobs.go` files though. I'll DRY those functions out.

test/pj-rehearse-integration/expected.yaml
```diff
@@ -158,7 +158,7 @@
       prow.k8s.io/refs.pull: "1234"
       prow.k8s.io/refs.repo: release
       prow.k8s.io/type: presubmit
-    name: 1e57006d-b7c8-11e9-a628-54e1ad07d68a
+    name: c2cc7272-0ca1-11ea-a57f-8c1645feb4f4
```
droslean: Please don't do these changes, since we are already ignore them in the script.

test/pj-rehearse-integration/expected.yaml
```diff
@@ -258,19 +258,19 @@
...
-        sha: 2a2c6326002a2df6099cb3cadf145f338ad3e137
+        sha: 368830202d143898f8d9e84bd4900d83adc1b9b6
       repo: release
     report: true
     rerun_auth_config: {}
     rerun_command: /test pj-rehearse
     type: presubmit
   status:
-    startTime: "2019-06-18T09:30:17Z"
+    startTime: "2019-11-21T20:59:08Z"
```
droslean: and startTime changes as well..

pkg/rehearse/jobs.go
stevekuznetsov: Hmm. Could we improve that with factoring and better interfaces? The problem we're solving here in rehearsal is the same as that we have in the registry, no? We want to take a configuration from the repo and return the fully resolved data. Sure, in the registry we first get the org/repo/branch as parameters but everything downstream of that is the same, right? I think the tricky bit with rehearsals is getting the tool to have as little bespoke logic as possible. Almost all bugs we've had in rehearsal are places where we do something here, and only here. If the data transformation is the same (roughly `func(data userConfig) (resolvedConfig, error)`) I would love it to be shared.

PR #297, opened 2019-11-25
ci-operator: Update the embedded template to the latest

smarterclayton:
Picks up changes to unify under CLUSTER_VARIANT


smarterclayton:
Tagging, needed for some testing of on-demand variant jobs.

PR #300, opened 2019-11-27
Add a group auto updater tool

droslean:
Adding a new tool that updates users in groups in an Openshift cluster.

Currently, the tool has the ability to get the users that need to be updated from a peribolos config.


cmd/groups-auto-updater/main.go
```diff
@@ -0,0 +1,176 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/config/org"
+
+	userV1 "github.com/openshift/client-go/user/clientset/versioned/typed/user/v1"
+
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+// Config holds the information of all group names and their configuration
```
stevekuznetsov: This seems like a lot of complexity. Why? We need a tool that takes four(three?) things as input:
 - the peribolos configuration path and the org to copy users from
 - the name of the group to update and a path to a kubeconfig to do it with (or just run it with in-cluster config)

So why not three flags?

cmd/groups-auto-updater/main.go
```diff
@@ -0,0 +1,176 @@
...
+type Config struct {
+	Groups map[string]GroupConfig `json:"groups"`
+}
+
+func (c *Config) validate() error {
+	for group, config := range c.Groups {
+		if config.FromPeribolosConfig != nil {
```
stevekuznetsov: For instance, what is the use of this? In what case do we care to configure this but not configure _how_?

cmd/groups-auto-updater/main.go
```diff
@@ -0,0 +1,176 @@
...
+	return users, nil
+}
+
+func updateGroup(groupName string, groupConfig GroupConfig, groupV1Interface userV1.GroupInterface) error {
+	g, err := groupV1Interface.Get(groupName, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("couldn't get group from cluster: %v", err)
+	}
+
+	if groupConfig.FromPeribolosConfig != nil {
```
stevekuznetsov: Or, this ... why does our tool have to handle the case where we are explicitly configured to do nothing?

cmd/groups-auto-updater/test_files/test_peribolos_config.yaml
```diff
@@ -0,0 +1,39 @@
...
+  test-org:
+    admins:
+    - admin1
+    - admin2
+    - admin3
+    - admin4
+    members:
+    - member1
+    - member2  
+    billing_email: test@redhat.com
```
stevekuznetsov: nit: we don't need to set any fields we don't read, right? so just admins and members

cmd/groups-auto-updater/main_test.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/openshift/client-go/user/clientset/versioned/fake"
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/diff"
+	coretesting "k8s.io/client-go/testing"
+)
+
+const (
+	configFile      = "test_files/test_config.yaml"
```
stevekuznetsov: This looks like an integration test, but written in Go. IMO we should have two types of tests:

1. unit tests, in Go, here, that verify that we call the client correctly or do filtering correctly or move data around correctly (implementation details that are important to get right but not important _how_)
2. integration tests, in shell, that use the binary as a black box and ensure the output is correct

cmd/groups-auto-updater/main.go
droslean: Yea, let's keep it simple for now. I will remove the config and add the flags.

cmd/groups-auto-updater/main_test.go
droslean: Yea I wanted to write a low-level unit test here. But I will remove the config, therefore the tests can be simplified as well.

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,123 @@
...
+	g, err := groupV1Interface.Get(group, metav1.GetOptions{})
+	if err != nil {
+		return fmt.Errorf("couldn't get group from cluster: %v", err)
+	}
+
+	users, err := getUsersFromPeribolosConfig(peribolosConfig, org)
+	if err != nil {
+		return fmt.Errorf("couldn't get users from peribolos config: %v", err)
+	}
+	g.Users = users.List()
```
stevekuznetsov: If you need a list, return it from `getUsers`

droslean:
> Can we get an integration test that uses the binary after it's built? We may need to re-use the dry-logger.

Sure, I will add the integration tests in this PR

cmd/group-auto-updater/main_test.go
```diff
@@ -0,0 +1,53 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/diff"
+
+	v1 "github.com/openshift/api/user/v1"
+	"github.com/openshift/client-go/user/clientset/versioned/fake"
+)
+
+func TestUpdateGroup(t *testing.T) {
+	peribolosConfig := "test_files/test_peribolos_config.yaml"
```
petr-muller: I think the fact that we cannot inline these 10 lines of YAML exposes a smell - IMO the updating logic should not read the filesystem (=take a path to the config as an input) but should  receive a parsed instance and work over that. Then even this test can be simpler and not include hitting a filesystem.

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,123 @@
...
+
+	b, err := ioutil.ReadFile(pConfig)
+	if err != nil {
+		return nil, fmt.Errorf("could not read peribolos configuration file: %v", err)
+	}
+
+	var peribolosConfig org.FullConfig
+	if err := yaml.Unmarshal(b, &peribolosConfig); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal peribolos config: %v", err)
+	}
```
petr-muller: Like said in the test file, I think this should be done in `main()` and the updating logic should just receive a parsed instance.

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,123 @@
...
+	g.Users = users
+
+	if _, err := groupV1Interface.Update(g); err != nil {
+		return fmt.Errorf("couldn't update group in the cluster: %v", err)
+	}
+
+	return nil
+}
+
+func main() {
```
petr-muller: I feel like we are missing a paragraph somewhere that would document what this tool is supposed to do and why it exists.

droslean: I'll add a readme file.

cmd/group-auto-updater/main_test.go
droslean: Inlining them in the code sounds good.

cmd/group-auto-updater/main.go
droslean: Nice catch. That was a leftover from the previous version of the tool. I will address it. Thanks.

cmd/group-auto-updater/main.go
petr-muller: I don't think a full README is actually necessary, all I would like was like two or three sentences in a GoDoc.

cmd/group-auto-updater/main_test.go
```diff
@@ -0,0 +1,82 @@
...
+	if _, err := userV1Client.Groups().Create(baseGroup); err != nil {
+		t.Fatalf("%v", err)
+	}
+
+	users, err := getUsersFromPeribolosConfig(peribolosConfigPath, testOrg)
+	if err != nil {
+		t.Fatalf("coulnd't get users from peribolos config: %v", err)
+	}
+
+	if err := updateGroup(testGroup, users, userV1Client.Groups()); err != nil {
```
petr-muller: We should also check that the API calls were actually made (at least the update one).

cmd/group-auto-updater/main_test.go
```diff
@@ -0,0 +1,82 @@
...
+	"k8s.io/apimachinery/pkg/api/equality"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/diff"
+
+	v1 "github.com/openshift/api/user/v1"
+	"github.com/openshift/client-go/user/clientset/versioned/fake"
+)
+
+func TestUpdateGroup(t *testing.T) {
```
petr-muller: I think this test is still trying to accomplish too much - it's a hybrid between a unit test and an integration test, and the result is that it tests multiple things and is unnecessarily complicated. Any time when a **unit** test needs to hit a filesystem, it is usually a very good indicator that some refactoring is in order.

The tool essentially does three things: 

1. Reads Peribolos config: we do not need to unit-test this (we only use other packages for this)
2. From Peribolos config read a list of logins to put into the group. This is very simple functionality and we probably would do without unit testing that (it is just reading two fields from the peribolos config), although it's easy and would not hurt.
3. Take a list of logins, a group and a client, and update the group accordingly. This is the `updateGroup` method and it should definitely be unit tested. For unit testing it, you do not even need Peribolos config: the method's input is just a list of logins. Thorough unit tests would also test error scenarios like that the method can handle error conditions gracefully.

In addition to unit tests, I think at least one integration test is always good - we have these in bash and over real(ish) input - like @stevekuznetsov [suggested](https://github.com/openshift/ci-tools/pull/300#pullrequestreview-323962872), we should use the dry-logger here to check that the tools issues expected API calls.

droslean: Yea integration tests using the drylogger would be nice. I will add them in this PR, and I will make the unit tests simplier

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,143 @@
...
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.group, "group", "", "The group that will be updated in the cluster.")
+	fs.StringVar(&o.peribolosConfig, "peribolos-config", "", "Peribolos configuration file")
+	fs.StringVar(&o.org, "org", "", "Org from peribolos configuration")
+
+	fs.BoolVar(&o.dryRun, "dry-run", false, "Print the generated group without updating it")
+	fs.BoolVar(&o.determinizeOutput, "determinize-output", false, "Determinize dry run's output by ordering the created object and removing the unnecessary fields")
```
stevekuznetsov: This smells like "test mode" -- why is it wrong to sort fields always and why can't we ignore fields that don't matter in our call to `diff` in the tests?

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,143 @@
...
+
+	userV1Client, err := userV1.NewForConfig(clusterConfig)
+	if err != nil {
+		return nil, fmt.Errorf("could not create user openshift client: %v", err)
+	}
+
+	return userV1Client, nil
+}
+
+func getUsersFromPeribolosConfig(peribolosConfig org.FullConfig, pOrg string) []string {
```
stevekuznetsov: This is only called once, why does it need it's own function?

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,143 @@
...
+	users := sets.NewString()
+	users.Insert(peribolosConfig.Orgs[pOrg].Admins...)
+	users.Insert(peribolosConfig.Orgs[pOrg].Members...)
+	return users.List()
+}
+
+func updateGroup(group string, users []string, groupV1Interface userV1.GroupInterface, dryLogger *steps.DryLogger, dryRun bool) error {
```
stevekuznetsov: Similarly, this is called once and doesn't do anything complicated -- lets just inline it into `main()`

pkg/steps/testing.go
```diff
@@ -122,6 +126,15 @@ func (dl *DryLogger) AddImageStreamTag(ist *imageapi.ImageStreamTag) {
 	dl.AddObject(ist.DeepCopyObject())
 }
 
+func (dl *DryLogger) AddGroup(group *userv1.Group) {
+	if dl.determinizeOutput {
+		if len(group.ObjectMeta.GetAnnotations()) > 0 {
```
stevekuznetsov: Why?

test/group-auto-updater-integration/run.sh
```diff
@@ -0,0 +1,43 @@
...
+
+GROUP="openshift-priv-admins"
+PERIBOLOS_CONFIG="${TEST_ROOT}/data/peribolos-config.yaml"
+ORG="test-org"
+
+
+compare_to_expected() {
+  local expected="${TEST_ROOT}/data/expected.yaml"
+  local output="$1"
+  diff -u "$expected" "$output"
```
stevekuznetsov: We use `-Naupr` everywhere else, why not here?

droslean:
> couldn't get group from cluster" error="groups.user.openshift.io \"openshift-priv-admins\" is forbidden: User \"system:serviceaccount:ci-op-jgmwmc7i:default\" cannot get groups.user.openshift.io at the cluster scope: no RBAC policy matched

@stevekuznetsov @petr-muller I could use your advice on this one. The tool tries first to `get` the group and then update it, and of course, in order to do that, it needs permissions. So the integration tests are failing because the test is running under `ci-operator`, and I don't feel comfortable to update `ci-operator` to generate more rbacs for the jobs. 

Shall we make the tool not to try to `get` anything first but instead create the `group` as new each time and then to try to `update` it? or 
Make ci-operator to generate `get` `groups` rbacs? or
Get the integration tests out of `ci-operator`?

petr-muller:
@droslean My thoughts: The integration test should run the test in `--dry-run` mode. Dry-run mode means we can attempt the `Get` but not even attempt the `Update`. I think it's reasonable behavior for `--dry-run` to attempt to `Get` a group, and if it does not succeed, assume the group does not exist and continue as if it would need to be created (but of course, never issue the actual Create calls and).

droslean:
@petr-muller Thanks. So it means that it is appropriate to allow `get` `groups`

petr-muller:
@droslean No, I don't think so - I think that the `--dry-run` mode in the tool should handle both having `get` permissions and not having them. If it does, then get it and behave as in update, If not, then assume it would need to be created. Then we can run the test in a job even without any rights.

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,130 @@
...
+	userV1Client, err := getUserV1Client()
+	if err != nil {
+		logger.WithError(err).Fatal("could not get user client")
+	}
+
+	users := sets.NewString()
+	users.Insert(peribolosConfig.Orgs[o.org].Admins...)
+	users.Insert(peribolosConfig.Orgs[o.org].Members...)
+
+	g := &v1.Group{}
```
stevekuznetsov: nit: this should be a literal initialization

```go
g := &v1.Group{
    Name: o.group,
    Users: users.List(),
}
```

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,130 @@
...
+	g.Name = o.group
+	g.Users = users.List()
+
+	if o.dryRun {
+		dryLogger.AddObject(g.DeepCopyObject())
```
stevekuznetsov: This isn't technically correct -- in the case where `Get` succeeds we do not create `g` but update `group` -- please log the actual thing we're doing below instead of assuming this here

droslean: addressed

cmd/group-auto-updater/main.go
droslean: addressed

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,133 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: o.group},
+		Users:      users.List(),
+	}
+
+	group, err := userV1Client.Groups().Get(o.group, metav1.GetOptions{})
+	if err == nil {
+		group.Users = users.List()
+		dryLogger.AddObject(group.DeepCopyObject())
+
+		if _, err := userV1Client.Groups().Update(group); err != nil && !o.dryRun {
```
stevekuznetsov: But if we are dry-running, we should not actually do the `Update()` or `Create()` right? :)

droslean: oops

cmd/group-auto-updater/main.go
```diff
@@ -0,0 +1,136 @@
...
+	userV1Client, err := getUserV1Client()
+	if err != nil {
+		logger.WithError(err).Fatal("could not get user client")
+	}
+
+	users := sets.NewString()
+	users.Insert(peribolosConfig.Orgs[o.org].Admins...)
+	users.Insert(peribolosConfig.Orgs[o.org].Members...)
+
+	g := &v1.Group{
```
stevekuznetsov: This is getting pretty convoluted ... something simpler:


```go
	var action func(*v1.Group) (*v1.Group, error)
	var group *v1.Group


	if existing, err := userV1Client.Groups().Get(o.group, metav1.GetOptions{}); err == nil {
	    group = existing
	    action = userV1Client.Groups().Update
	} else if err != nil && (kerrors.IsNotFound(err) || kerrors.IsUnauthorized(err) && o.dryRun) {
	    group = &v1.Group{ObjectMeta: metav1.ObjectMeta{Name: o.group}}
	    action = userV1Client.Groups().Create
	} else {
		logger.WithError(err).Fatal("couldn't get group from cluster")
	}

	group.Users = users.List()
	if o.dryRun {
		dryLogger.AddObject(g.DeepCopyObject())
		if err := dryLogger.Log(); err != nil {
			logger.WithError(err).Fatal("error while parsing dry logger's objects")
		}
	} else if _, err := action(group); err != nil {
		logger.WithError(err).Fatal("couldn't sync group to the cluster")	
	}
```

PR #301, opened 2019-11-27
Add Dockerfile for ca-cert-issuer

hongkailiu:
https://jira.coreos.com/browse/DPTP-619

### DIFF
--- a/images/ca-cert-issuer/Dockerfile
+++ b/images/ca-cert-issuer/Dockerfile
@@ -1,0 +1,9 @@
+FROM centos:7
+LABEL maintainer="hongkliu@redhat.com"
+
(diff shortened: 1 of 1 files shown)


hongkailiu:
> Why? Why not just add a `BuildConfig` in `core-services`?

HAHA ... you know what I have finished BC definition ... but i realized the feeling that @petr-muller prefer ci-operator build than `bc` when i did the ipi image.

OK ... switching back.

stevekuznetsov:
A Dockerfile here and builds every time this repository change makes sense when the content that is in the build depends on this repository or changes with it. The build you had proposed here is static and entirely unrelated to this repository, so it does not belong here. @hongkailiu

PR #303, opened 2019-11-30
[prowgen] Generate private jobs by using OAuth token instead of SSH 

droslean:
Since `clonerefs` now supports OAuth authentication, we want to get rid of the SSH configuration and use an OAuth token instead.

Note: OAuth requires a `cloneURI` with a scheme included. 

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -56,8 +56,9 @@ exec ci-operator \
     --template=/tmp/%[1]s
 `
 
-	sshKeyPath    = "/usr/local/github-ssh-credentials-openshift-bot"
-	sshSecretName = "github-ssh-credentials-openshift-bot"
+	oauthTokenPath  = "/usr/local/github-credentials"
+	oauthSecretName = "github-credentials-openshift-ci-robot-private-git-cloner"
+	oauthKey        = "oauth"
 
 	prowgenConfigFile = ".config.prowgen"
 )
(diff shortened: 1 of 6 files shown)


droslean:
/hold 

This PR needs also changes in the realese repo.

droslean:
Hold until changing `ci-operator` to support OAuth token.

cmd/ci-operator-prowgen/main.go
```diff
@@ -57,9 +57,8 @@ exec ci-operator \
     --target=%[1]s \
     --template=/tmp/%[1]s
 `
-
-	sshKeyPath    = "/usr/local/github-ssh-credentials-openshift-bot"
-	sshSecretName = "github-ssh-credentials-openshift-bot"
+	oauthSecretName = "oauth-token"
+	oauthKey        = "oauth"
```
petr-muller: Let's use a better-named, appropriately scoped token for this purpose. I'll prep it tomorrow together with a `populate...` script change.

droslean: I initially opened this PR to add the OAuth authentication but then I realize that I need make `ci-operator` first to support this. 

Those names are supposed to point to `oauth-token` secret in `ci` namespace and the `oauth` is the key. 
However the approach now is changed, since we only need to add those under `volumes/volumeMounts` and not in `DecorationConfig`

petr-muller: > When you update the secret, take care also the prow components that are using that.

That's the point, I do not want to change prow components; let them use whatever secrets they are using: they need all kinds of permissions. But here I want to use a properly named secret containing a token with properly scoped permissions. The jobs only need to clone code, so the token only needs read rights on repos.

droslean: Yea, that makes sense. We need a new token for cloning as you said. Thanks for pointing this out.

petr-muller: Once https://github.com/openshift/release/pull/6339 is merged we can use the following in Prowgen

Secret name: `github-credentials-openshift-ci-robot-private-git-cloner`, item `oauth`

stevekuznetsov: We should not be using the `oauth-token` secret

```
$ grep 'oauth-token' ci-operator/populate-secrets-from-bitwarden.sh
$
```

All places you see it used need to change to use the appropriate `github-credentials-` secret, please.

cmd/ci-operator-prowgen/main.go
```diff
@@ -56,8 +56,9 @@ exec ci-operator \
     --template=/tmp/%[1]s
 `
 
-	sshKeyPath    = "/usr/local/github-ssh-credentials-openshift-bot"
-	sshSecretName = "github-ssh-credentials-openshift-bot"
+	oauthTokenPath  = "/usr/local/github-oauth-credentials-openshift-bot"
```
petr-muller: This still mentions `openshift-bot`, I think this can easily be called just `github-oauth-credentials`

PR #304, opened 2019-12-02
[ci-operator] Add OAuth token authentication support while cloning

droslean:


pkg/steps/source.go
```diff
@@ -129,27 +139,33 @@ func (s *sourceStep) Run(ctx context.Context, dry bool) error {
...
+}
+
+func determineCloneURI(sshSecretName, oauthSecretName, org, repo string) string {
+	if len(sshSecretName) > 0 {
+		return fmt.Sprintf("ssh://git@github.com/%s/%s.git", org, repo)
+	}
+	if len(oauthSecretName) > 0 {
+		return fmt.Sprintf("https://github.com/%s/%s.git", org, repo)
+	}
+	return ""
```
stevekuznetsov: This is not a valid URI and will cause a crash if we allow it

pkg/steps/source.go
```diff
@@ -129,27 +139,33 @@ func (s *sourceStep) Run(ctx context.Context, dry bool) error {
 		return fmt.Errorf("could not resolve clonerefs source: %v", err)
 	}
 
-	return handleBuild(s.buildClient, createBuild(s.config, s.jobSpec, clonerefsRef, s.resources, s.sshSecretName), dry, s.artifactDir, s.dryLogger)
+	return handleBuild(s.buildClient, createBuild(s.config, s.jobSpec, clonerefsRef, s.resources, s.sshSecretName, s.oauthSecretName), dry, s.artifactDir, s.dryLogger)
+}
+
+func determineCloneURI(sshSecretName, oauthSecretName, org, repo string) string {
```
stevekuznetsov: Passing flags as arguments is a smell: https://martinfowler.com/bliki/FlagArgument.html

pkg/steps/source.go
droslean: The value is already empty if it reaches that point.

pkg/steps/source.go
droslean: yes this is smelly. I will take a look.

droslean:
> Do we have no integration test that uses SSH or OAuth?

No, we have only the base and the mult_stage integration tests that both of them are not using any of SSH or OAuth. But we do have unittests that test those cases.

pkg/steps/source.go
droslean: Those args are not related to the flags. The flags are pointing to the file that contains the secret.
Those are the names of the generated secret. The names are a hash code related to their contents. 

I guess we need a better way to determine the cloneURI, since we don't initially allow both secrets

pkg/steps/source.go
stevekuznetsov: That was not the behavior before and we should not allow it.

pkg/steps/source.go
stevekuznetsov: I am saying that passing an argument to a method to use it in an `if` statement is smelly

pkg/steps/source.go
droslean: I wonder why the tests passed. I'll check.

cmd/ci-operator/main.go
```diff
@@ -270,6 +272,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.Var(&opt.templatePaths, "template", "A set of paths to optional templates to add as stages to this job. Each template is expected to contain at least one restart=Never pod. Parameters are filled from environment or from the automatic parameters generated by the operator.")
 	flag.Var(&opt.secretDirectories, "secret-dir", "One or more directories that should converted into secrets in the test namespace. If the directory contains a single file with name .dockercfg or config.json it becomes a pull secret.")
 	flag.StringVar(&opt.sshKeyPath, "ssh-key-path", "", "A path of the private ssh key that is going to be used to clone a private repository.")
+	flag.StringVar(&opt.oauthTokenPath, "oauth-token", "", "A path of the OAuth token that is going to be used to clone a private repository.")
```
petr-muller: I think we should call this `oauth-token-path` (or just `token-path`) for consistency with `--ssh-key-path`.

cmd/ci-operator/main.go
```diff
@@ -1425,27 +1437,35 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, authClient *authclien
...
-	secret.Data[coreapi.SSHAuthPrivateKey] = b
+	if cloneAuthType == steps.CloneAuthTypeSSH {
+		secret.Name = fmt.Sprintf("ssh-%s", hash)
+		secret.Type = coreapi.SecretTypeSSHAuth
+		// Secret.Data["ssh-privatekey"] is required on SecretTypeSSHAuth type.
+		// https://github.com/kubernetes/api/blob/master/core/v1/types.go#L5466-L5470
+		secret.Data[coreapi.SSHAuthPrivateKey] = data
+	} else if cloneAuthType == steps.CloneAuthTypeOAuth {
+		secret.Type = coreapi.SecretTypeOpaque
+		secret.Name = fmt.Sprintf("oauth-%s", getHashFromBytes(data))
```
petr-muller: No need to hash the data again, we already have it in `hash`

cmd/ci-operator/main.go
```diff
@@ -1425,27 +1437,35 @@ func eventRecorder(kubeClient *coreclientset.CoreV1Client, authClient *authclien
...
+		secret.Data[steps.OauthSecretKey] = []byte(strings.TrimSpace(string(data)))
```
petr-muller: the `TrimSpace` seems misplaced here. IMO, we should `TrimSpace` immediately after we `ReadFile` (even before we hash) and just use the cleaned string later

pkg/steps/source.go
```diff
@@ -41,33 +41,61 @@ const (
...
+	CloneAuthTypeSSH   CloneAuthType = "SSH"
+	CloneAuthTypeOAuth CloneAuthType = "OAuth"
+)
+
+type CloneAuthConfig struct {
+	Secret *coreapi.Secret
+	Type   CloneAuthType
+}
+
+func (c *CloneAuthConfig) determineCloneURI(org, repo string) string {
```
petr-muller: maybe just `get` instead of `determine`? it's a simple choice, no fancy process :)

pkg/steps/source.go
```diff
@@ -41,33 +41,61 @@ const (
...
 
-	if len(sshSecretName) > 0 {
+	if cloneAuthConfig != nil && cloneAuthConfig.Type == CloneAuthTypeSSH {
 		dockerCommands = append(dockerCommands, fmt.Sprintf("ADD %s /etc/ssh/ssh_config", sshConfig))
 		dockerCommands = append(dockerCommands, fmt.Sprintf("COPY ./%s %s", coreapi.SSHAuthPrivateKey, sshPrivateKey))
 	}
 
+	if cloneAuthConfig != nil && cloneAuthConfig.Type == CloneAuthTypeOAuth {
+		dockerCommands = append(dockerCommands, fmt.Sprintf("COPY ./%s %s", OauthSecretKey, oauthToken))
+	}
```
petr-muller: nit: I think that something like this would better convey the actual logic. Pls consider.

```go
if cloneAuthConfig != nil {
  switch cloneAuthConfig.Type {
  case CloneAuthTypeSSH:
    ...
  case CloneAuthTypeOAuth:
    ...
  }
}
```

cmd/ci-operator/main.go
droslean: well, I know how that it seems misplaced, but we have already broken prow once because of this. It seems that we need to Trim it otherwise the `ReadFile` will append a newline in the value. 
See [prow/config/secret/secret.go#L45](https://github.com/kubernetes/test-infra/blob/master/prow/config/secret/secret.go#L45)

droslean: But your concern is correct. I will move this.

petr-muller: Yeah, that's what I meant - we need to `TrimSpace`, just not here but earlier.

droslean: I am not sure if trimming space in the ssh key will cause any problems. I will validate that and then address your comment.

petr-muller: Should not hurt. One more thing - `bytes` has `TrimSpace` too, no need to convert it back and forth.

https://golang.org/pkg/bytes/#TrimSpace

PR #306, opened 2019-12-04
Replace svc bokos with its route

hongkailiu:
https://jira.coreos.com/browse/DPTP-670

require https://github.com/kubernetes/test-infra/pull/15458/files and prow-image bumping

/hold

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -379,7 +379,32 @@ func generatePodSpecTemplate(info *prowgenInfo, secret *cioperatorapi.Secret, re
 	// TODO add to all templates when they are migrated
 	// TODO expose boskos (behind an oauth proxy) so it can be used by build clusters
 	if needsLeaseServer {
-		container.Args = append(container.Args, "--lease-server=http://boskos")
+		if info.Org == "openshift" && info.Repo == "ci-secret-mirroring-controller" {
+			container.Args = append(container.Args, "--lease-server=https://boskos-ci.svc.ci.openshift.org")
+			container.Args = append(container.Args, "--lease-server-username=ci")
+			container.Args = append(container.Args, "--lease-server-password-file=/etc/boskos/password")
+			container.VolumeMounts = append(container.VolumeMounts, kubeapi.VolumeMount{
+				Name:      "boskos",
+				MountPath: "/etc/boskos",
+				ReadOnly:  true,
+			})
+			podSpec.Volumes = append(podSpec.Volumes, kubeapi.Volume{
+				Name: "boskos",
+				VolumeSource: kubeapi.VolumeSource{
+					Secret: &kubeapi.SecretVolumeSource{
+						SecretName: "boskos-credentials",
+						Items: []kubeapi.KeyToPath{
+							{
+								Key:  "password",
+								Path: "password",
+							},
+						},
+					},
+				},
+			})
+		} else {
+			container.Args = append(container.Args, "--lease-server=http://boskos")
+		}
 	}
 	container.VolumeMounts = append(container.VolumeMounts, kubeapi.VolumeMount{Name: "cluster-profile", MountPath: clusterProfilePath})
 	if len(template) > 0 {
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -212,15 +212,17 @@ type options struct {
 	idleCleanupDuration time.Duration
 	cleanupDuration     time.Duration
 
-	inputHash     string
-	secrets       []*coreapi.Secret
-	sshSecret     *coreapi.Secret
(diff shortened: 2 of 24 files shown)


pkg/steps/clusterinstall/template.go
```diff
@@ -110,6 +110,12 @@ objects:
     - name: cluster-profile
       secret:
         secretName: ${JOB_NAME_SAFE}-cluster-profile
+    - name: boskos
+      secret:
+        secretName: boskos-credentials
+        items:
```
stevekuznetsov: Why do you need to call out the password?

hongkailiu: It contains other items as well ... we need only the password part here for this pod

```bash
$ oc get secret -n ci boskos-credentials -o yaml | yq -r '.data | keys[]'
ci.htpasswd
password
username

```

hongkailiu:
It uses the image in our own registry.
Need to figure out where we generate it.

https://github.com/openshift/ci-tools/blob/5a3523c521bb2d36cc93be82a1c227f156f2dec4/pkg/steps/clusterinstall/template.go#L123

hongkailiu:
@stevekuznetsov 
There are other files (of jobs and templates) in `release` repo that contains `http://boskos` ... Do we need to change too? Or We need to `prowgen` them after this PR merges?

One exception is `core-services/prow/03_deployment/boskos_reaper.yaml:23:        - --boskos-url=http://boskos` ... we might not need to change it because `reaper` runs on the same cluster.

pkg/steps/clusterinstall/template.go
stevekuznetsov: I'm concerned that this is about ten times more complicated than every other way that we load secrets into any deployment. Why don't we have one file, maybe JSON or YAML, that has the username and password, add it to a known path and give that to the client library? Right now, this secret mounting is special because:

 - you have to know that there is other data you don't want to expose
 - you have to know which parts to expose as a file
 - you have to know how to expose the username as an environment var

stevekuznetsov:
Jobs using `http://boskos` or `http://boskos.ci` will need to be updated with `prowgen` but may need to be updated manually if they are manual. Agreed that deployments do not need to change.

bbguimaraes:
FYI this is all going away with https://jira.coreos.com/browse/DPTP-677, which should land today (with https://github.com/openshift/release/pull/6255 + a future PR here to remove the container from this template).

All templates will use [the native client](https://github.com/openshift/ci-tools/blob/master/pkg/lease/client.go). It is configured using an argument:

https://github.com/openshift/ci-tools/blob/594d99eb0c953cd81e70eebae4547c965e38319f/cmd/ci-operator/main.go#L259

that is generated by `prowgen`:

https://github.com/openshift/ci-tools/blob/594d99eb0c953cd81e70eebae4547c965e38319f/cmd/ci-operator-prowgen/main.go#L379-L383

bbguimaraes:
> Jobs using `http://boskos` or `http://boskos.ci` will need to be updated with `prowgen` but may need to be updated manually if they are manual. Agreed that deployments do not need to change.

The script in https://github.com/openshift/release/pull/6127 and https://github.com/openshift/release/pull/6241 can be used to update most of the non-generated jobs.

hongkailiu:
/test images

going to push image to my own repo ... and test with it in rehearsal

pkg/steps/clusterinstall/template.go
hongkailiu: I was trying to minimize the number of secrets by aggregating the sensitive information into a single secret if they are kind of related. In this case, it is all about boskos basic auth set up (both server side and client sides). Maybe it is totally unnecessary.

I used env. var. because of multi-occurrence. With the current k8s-API way, that reason is no longer valid and env. vars are removed. And username (ci) is in the src control now and I hope that is OK. Password file is mounted as the unknown path is given to ci-operator.

@stevekuznetsov 

See if the current one is better.

hongkailiu:
the breaking change should be fixed by https://github.com/openshift/release/pull/6264

hongkailiu:
I cannot explain this:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/6017/rehearse-6017-pull-ci-openshift-ci-secret-mirroring-controller-master-e2e-aws-build01/25

```
2019/12/05 20:13:24 Acquiring lease for "aws-quota-slice"
2019/12/05 20:13:54 No custom metadata found and prow metadata already exists. Not updating the metadata.
2019/12/05 20:13:56 Ran for 4m21s
2019/12/05 20:13:56 Submitted failure event to sentry (id=200329a188624c7b8e05b6a33215c2cf)
error: could not run steps: step e2e-aws-build01 failed: failed to acquire lease: status 504 Gateway Time-out, status code 504
```

I tried to `curl` inside the pod and it works.

PR #308, opened 2019-12-04
pkg: make and use reloadWatcher designed for configmaps

AlexNPavel:
This PR changes reload handling for configresolver instances that mount configs and registry via configmaps. Due to a very large number of file changes that occur whenever a configmap is updated, the configresolver reloads excessively when run in cluster and uses a very large amount of CPU. Instead of watching all files, the `reloadWatcher` instead just watches for a `CREATE` event to occur on the `..data` directory/symlink, which is the last step that occurs when a configmap is updated. This should drastically reduce the number of reloads and CPU usage of the configresolver run in a cluster.


pkg/watcher/watcher.go
```diff
@@ -0,0 +1,62 @@
...
+		for {
+			select {
+			case <-ctx.Done():
+				if err := w.Close(); err != nil {
+					errFunc(err)
+					//logrus.WithError(err).Error("Failed to close fsnotify watcher")
+				}
+				return
+			case event := <-w.Events:
+				if event.Name == "..data" && event.Op == fsnotify.Create {
```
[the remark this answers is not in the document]
AlexNPavel: The steps that occur during a configmap update are:
1. A new timestamped directory is created with a `..` prefix
2. New symlinks (that point to `..data/<filename>`) are created if necessary
3. A `..data_tmp` symlink that points to the timestamped directory is made
4. `..data_tmp` is moved to `..data`, in a way that results in a `CREATE` event for `..data`.
5. Remove stale (deleted) symlinks

Since the last important step is the `CREATE` event on `..data`, we can run reloads when `..data` gets a `CREATE` event. The newly added integration tests simulate part of this behavior.

cmd/ci-operator-configresolver/main.go
```diff
@@ -99,6 +100,7 @@ func gatherOptions() options {
 	fs.DurationVar(&o.cycle, "cycle", time.Minute*2, "Cycle duration for config reload")
 	fs.BoolVar(&o.validateOnly, "validate-only", false, "Load the config and registry, validate them and exit.")
 	fs.BoolVar(&o.flatRegistry, "flat-registry", false, "Disable directory structure based registry validation")
+	fs.BoolVar(&o.configMapMount, "cm-mount", false, "If set, use more efficient reload watcher designed for configmaps")
```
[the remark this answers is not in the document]
AlexNPavel: Updated the flag name to make it more related to what it does

droslean:
Since this got more complicated, I don't think that this is the correct approach.  I think it would be best to monitor the configMaps directly with `informers` Instead of mounting them and watch the files. That would save a lot of CPU.

@stevekuznetsov WDYT?

cmd/ci-operator-configresolver/main.go
```diff
@@ -99,6 +100,7 @@ func gatherOptions() options {
 	fs.DurationVar(&o.cycle, "cycle", time.Minute*2, "Cycle duration for config reload")
 	fs.BoolVar(&o.validateOnly, "validate-only", false, "Load the config and registry, validate them and exit.")
 	fs.BoolVar(&o.flatRegistry, "flat-registry", false, "Disable directory structure based registry validation")
+	fs.BoolVar(&o.configMapReloader, "cm-reloader", false, "If set, use more efficient reload watcher designed for configmaps")
```
stevekuznetsov: Why not always do this if there exists a `..data` directory?

cmd/ci-operator-configresolver/main.go
```diff
@@ -128,6 +130,9 @@ func validateOptions(o options) error {
 	if o.validateOnly && o.flatRegistry {
 		return errors.New("--validate-only and --flat-registry flags cannot be set simultaneously")
 	}
+	if o.configMapReloader && !o.flatRegistry {
+		return errors.New("If cm-reloader is set, flat-registry must also be set")
```
stevekuznetsov: Similarly, couldn't we default `--flat-registry` if we detect we're reading from a configmap?

pkg/load/configAgent.go
```diff
@@ -66,18 +70,42 @@ func NewConfigAgent(configPath string, cycle time.Duration, errorMetrics *promet
 		}
 	}, a.cycle)
 
-	// fsnotify reload
-	configWatcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		return nil, fmt.Errorf("Failed to create new watcher: %v", err)
+	reloaderFuncs := []func(ctx context.Context){}
+	if configMapReload {
```
stevekuznetsov: Instead of assuming a structure, couldn't we handle both cases with a recursive traversal that considers the full subtree handled if it finds a `..data` element?

pkg/load/registryAgent.go
```diff
@@ -61,18 +62,31 @@ func NewRegistryAgent(registryPath string, cycle time.Duration, errorMetrics *pr
...
-	registryWatcher, err := fsnotify.NewWatcher()
-	if err != nil {
-		return nil, fmt.Errorf("Failed to create new watcher: %v", err)
-	}
-	err = populateWatcher(registryWatcher, a.registryPath)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to populate watcher: %v", err)
+	reloaderFunc := func(ctx context.Context) {}
+	if configMapMount {
+		errFunc := func(error) {
```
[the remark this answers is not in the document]
stevekuznetsov: More like, a large amount of `New(Registry|Config)Agent` functions are duplicated, could we DRY it out?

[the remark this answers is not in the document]
AlexNPavel: We could. The 2 agents share 5 variables with only 1-2 new variables specific to the agent's type. The `recordError`, `GetGeneration`, and all but the first 2 lines of `New...Agent` are the same between the 2 agents. I can make a new type just called `Agent` that handles all of the stuff in common.

cmd/ci-operator-configresolver/main.go
AlexNPavel: I assume so, but I prefer to be more explicit about the flags/args instead of silently enabling them. Either way would work.

pkg/load/registryAgent.go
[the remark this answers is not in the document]
AlexNPavel: I will do this in a follow-up PR, as this PR is already quite large on its own.

AlexNPavel:
The PR has been reworked so the registry and config agents automatically determine if there is a configmap mount and there is a helper function that adds the watchers more cleanly.

pkg/load/configAgent.go
```diff
@@ -51,6 +51,8 @@ func init() {
 
 // NewConfigAgent returns a ConfigAgent interface that automatically reloads when
 // configs are changed on disk as well as on a period specified with a time.Duration.
+// If configMapMount is true, NewConfigAgent assumes that all files in configPath are
```
stevekuznetsov: Old godoc

pkg/load/registryAgent.go
```diff
@@ -61,18 +61,33 @@ func NewRegistryAgent(registryPath string, cycle time.Duration, errorMetrics *pr
...
-	registryWatcher, err := fsnotify.NewWatcher()
+	cmMount, err := containsCMMount(a.registryPath)
 	if err != nil {
-		return nil, fmt.Errorf("Failed to create new watcher: %v", err)
+		return nil, fmt.Errorf("Failed to walk directory %s: %v", a.registryPath, err)
 	}
-	err = populateWatcher(registryWatcher, a.registryPath)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to populate watcher: %v", err)
+	if cmMount {
```
stevekuznetsov: Can we DRY out the code between these two agent init functions? It looks very similar

PR #309, opened 2019-12-05
test/ci-operator-*: improve configresolver tests

AlexNPavel:
This improves the `ci-operator-configresolver`'s tests in a variety of ways.

1. Checks that the generation is greater than `currGen` instead of `currGen+1`. This was always failing previously
2. Report error when `"${i}" -eq 9` instead of `"${i}" -eq 10`. Due to the way bash for loops work, `i` only reaches 9 in the loop, so this case was never hit previously, which is why the previous test bug wasn't detected.
3. Use `rsync -avh --quiet --delete --inplace` instead of deleting all files in `registry` and then copying files from `registry2` into the newly emptied directory. This prevents the registry from being reloaded excessively during the updates and prevents the configresolver reload errors.
4. Instead of storing `$PID`, use `$(jobs -p)` to get all background jobs. This prevents the test from indefinitely stalling if the configresolver crashes during the test.
5. Redirect all output to a logfile and only print it on error. This cleans up the output considerably.
6. Check output file for errors. We didn't check for errors previously, so the configresolver could error silently.
7. Add some new progress output since all the configresolver's output is now gone.
8. Add `[INFO]` and `[ERROR]` prefixes to all bash echoes.

### DIFF
--- a/test/ci-operator-configresolver-integration/run.sh
+++ b/test/ci-operator-configresolver-integration/run.sh
@@ -15,71 +15,98 @@ cd "$WORKDIR"
 # copy registry to tmpdir to allow tester to modify registry
 cp -a "$ROOTDIR"/test/ci-operator-configresolver-integration/ tests
 cp -a "$ROOTDIR"/test/multistage-registry multistage-registry
-ci-operator-configresolver -config tests/configs -registry multistage-registry/registry -log-level debug -cycle 2m &
-PID=$!
+ci-operator-configresolver -config tests/configs -registry multistage-registry/registry -log-level debug -cycle 2m &> output.log &
+echo "[INFO] Started configresolver"
 for (( i = 0; i < 10; i++ )); do
     if [[ "$(curl http://127.0.0.1:8081/healthz/ready 2>/dev/null)" == "OK" ]]; then
         break
     fi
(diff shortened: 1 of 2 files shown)


test/ci-operator-configresolver-integration/run.sh
```diff
@@ -15,71 +15,98 @@ cd "$WORKDIR"
...
     exit 1
 fi
-currGen=$(curl 'http://127.0.0.1:8080/configGeneration')
+echo "[INFO] Got correct resolved config from resolver"
+currGen=$(curl 'http://127.0.0.1:8080/configGeneration' 2>/dev/null)
 export currGen
 cp tests/configs2/release-4.2/openshift-installer-release-4.2-golang111.yaml tests/configs/release-4.2/openshift-installer-release-4.2.yaml
 for (( i = 0; i < 10; i++ )); do
-    if [[ "$(curl http://127.0.0.1:8080/configGeneration 2>/dev/null)" -gt $currGen+1 ]]; then
+    if [[ "$(curl http://127.0.0.1:8080/configGeneration 2>/dev/null)" -gt $currGen ]]; then
```
stevekuznetsov: Amazed arithmetic works like this bare inside of a `[[`...

PR #310, opened 2019-12-05
SDCICD-99: Add multi-secret support in ci-operator

jeefy:
This adds support for attaching multiple secrets to a prowjob build using the ci-operator.

It adds the `secrets` option to the test config, while maintaining the original `secret` config (so we don't break literally every job) but does note that we should deprecate the singular-option in favor of `secrets`

A config will throw a validation failure if both `secret` and `secrets` are set. It will also fail if a config has two secrets of the same name. In the background, setting `secret` will just append the contents to `secrets` then nil-out the `secret` attribute.

This is a first pass at supporting this. In the future we should go through and try to update all the configs to use the new attribute, but until then supporting both as this does seems reasonable. 

/cc @meowfaceman


jeefy:
https://i.kym-cdn.com/entries/icons/original/000/027/475/Screen_Shot_2018-10-25_at_11.02.15_AM.png

Looking at what I did. :)

pkg/steps/pod_test.go
```diff
@@ -3,7 +3,7 @@ package steps
 import (
 	"testing"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
jeefy: I think this was vscode auto-resolving something. Can undo if needed.

cmd/ci-operator-prowgen/main.go
```diff
@@ -607,8 +617,11 @@ func generateJobs(
 
 	for _, element := range configSpec.Tests {
 		var podSpec *kubeapi.PodSpec
+		if element.Secret != nil {
+			element.Secrets = append(element.Secrets, element.Secret)
+		}
```
petr-muller: This does not look correct. We move `Secret` to `Secrets` here for each test, but then we do it again in `generatePodSpecRandom` and `generatePodSpecTemplate`? This should be only done once.

pkg/api/config.go
```diff
@@ -146,15 +146,33 @@ func validateTestStepConfiguration(fieldRoot string, input []TestStepConfigurati
...
+
+		if test.Secret != nil && test.Secrets != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("test.Secret and test.Secrets cannot both be set"))
+		}
+
 		if test.Secret != nil {
+			test.Secrets = append(test.Secrets, test.Secret)
+		}
+
+		secretNames := make(map[string]bool)
```
petr-muller: We can use `sets.String` here, it's even already imported in this module.

pkg/steps/pod_test.go
petr-muller: Please undo.

CONFIGURATION.md
```diff
@@ -317,16 +320,22 @@ of the images in the pipeline.
...
 resource quantity semantics (e.g. `1Gi` or `500M`).
 
-# `tests.secret` 
+# `test.secrets`
+`Secrets` is an array of secrets. This  enables users to mount multiple secrets
+inside test container. It is users responsibility to make sure secret is available
+in the temporary namespace. This can be done by providing flag `--secret-dir`
+to ci-operator in prow configuration.
+
+# `tests.secret` Deprecated
```
droslean: Since this is not being used anymore, the configuration will break either way. Lets remove it and open a PR to release repo as a follow-up to fix the ci-operator configurations.

petr-muller: The config should not break, `secret` is internally turned into `secrets` with a single element, which seems fine to me. Here we can migrate to new config "properly" (merge this change, then migrate to new config, then remove the old code), there's no reason to do it via a breaking change.

pkg/steps/pod_test.go
jeefy: Will do

cmd/ci-operator-prowgen/main.go
jeefy: I agree. I took a sort of "look for any reference of singular-Secret and handle it" in case `generatePodSpecRandom` and `generatePodSpecTemplate` were called elsewhere, but AFAICT they aren't. I'll clean this up next.

jeefy:
@petr-muller PTAL :)

hongkailiu:
@jeefy and @petr-muller 

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/6519/rehearse-6519-periodic-ci-openshift-osde2e-master-addon-prow-operator-test/1

Probably we need include index on the secret name?

PR #314, opened 2019-12-06
Log boskos acquiring time

hongkailiu:
Follow up https://github.com/openshift/release/pull/6269

Before we figure out why boskos takes that long, we log the time in the client side.
We could add metric/panels on the sever side too.

### DIFF
--- a/pkg/lease/client.go
+++ b/pkg/lease/client.go
@@ -3,7 +3,9 @@ package lease
 import (
 	"context"
 	"fmt"
+	"log"
 	"sync"
+	"time"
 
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
This will just spam the users, though, who are reading CI Operator output. We should instead add request metrics to `boskos`

hongkailiu:
> This will just spam the users, though, who are reading CI Operator output. We should instead add request metrics to `boskos`

Understood.

Creating a card for this.

/close

hongkailiu:
https://jira.coreos.com/browse/DPTP-678

PR #316, opened 2019-12-09
pkg/lease: use AcquireWait instead of Acquire

bbguimaraes:
Many resource types have very low maximum values and rely on waiting for
extended periods of time until one is available.

The hard-coded timeout period used is based on the original templates
(https://github.com/openshift/release/pull/5802).


bbguimaraes:
Tested more thoroughly, found no issues. Merging because it is impacting some teams.

/hold cancel

PR #323, opened 2019-12-10
lease: use acquire with priority

stevekuznetsov:
This was another regression between the usage in `boskosctl` and the
in-house lease adaptation.

/assign @bbguimaraes


bbguimaraes:
I saw it when I implemented, why do we favor this over FIFO?

/lgtm

PR #327, opened 2019-12-12
Update Configuration.MD

eisraeli:
Rephrase the base_images description in order to make it less confusing.

### DIFF
--- a/CONFIGURATION.md
+++ b/CONFIGURATION.md
@@ -130,9 +130,9 @@ override specific components for specific tests at non-standard levels when many
 streams and one tag are used to assemble a release.
 
 # `base_images`
-`base_images` are parents for images built from the repository, but are not built
-from the repository. The field is a mapping from pipeline image name to remote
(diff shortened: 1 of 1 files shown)


CONFIGURATION.md
```diff
@@ -130,9 +130,9 @@ override specific components for specific tests at non-standard levels when many
 streams and one tag are used to assemble a release.
 
 # `base_images`
-`base_images` are parents for images built from the repository, but are not built
-from the repository. The field is a mapping from pipeline image name to remote
-`ImageStream` specification. A common base image might be an operating system:
+`base_images` are parent images.
```
petr-muller: I feel like we are losing information with this change (parent images for what?). I agree that the original sentence is not ideal. 

Maybe "... are external images that serve as parents for images built from the repository"?

eisraeli: @petr-muller 
Yeah, I agree. That sounds better.

eisraeli:
@petr-muller It should be more clear now without losing any information. Thanks.

PR #336, opened 2019-12-18
Specify pull secret in build steps

hongkailiu:
openshift/installer repo has builds which need accessimages in api.ci.

The default pull secret by build is from SA `builder` on the same cluster.

We use the pull secrets created with https://github.com/openshift/release/pull/6478 which contains oauth for both registries on api.ci and build01.

Verified by
https://github.com/openshift/release/pull/6420#issuecomment-567329232

### DIFF
--- a/pkg/steps/rpm_injection.go
+++ b/pkg/steps/rpm_injection.go
@@ -7,6 +7,7 @@ import (
 	buildapi "github.com/openshift/api/build/v1"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	routeclientset "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	coreapi "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/openshift/ci-tools/pkg/api"
@@ -26,6 +27,7 @@ type rpmImageInjectionStep struct {
 	artifactDir string
 	jobSpec     *api.JobSpec
 	dryLogger   *DryLogger
+	pullSecret  *coreapi.Secret
 }
 
 func (s *rpmImageInjectionStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
@@ -52,6 +54,7 @@ func (s *rpmImageInjectionStep) Run(ctx context.Context, dry bool) error {
 		},
 		"",
 		s.resources,
+		s.pullSecret,
 	), dry, s.artifactDir, s.dryLogger)
 }
 
@@ -77,7 +80,7 @@ func (s *rpmImageInjectionStep) Description() string {
 	return "Inject an RPM repository that will point at the RPM server"
 }
 
-func RPMImageInjectionStep(config api.RPMImageInjectionStepConfiguration, resources api.ResourceConfiguration, buildClient BuildClient, routeClient routeclientset.RoutesGetter, istClient imageclientset.ImageStreamTagsGetter, artifactDir string, jobSpec *api.JobSpec, dryLogger *DryLogger) api.Step {
+func RPMImageInjectionStep(config api.RPMImageInjectionStepConfiguration, resources api.ResourceConfiguration, buildClient BuildClient, routeClient routeclientset.RoutesGetter, istClient imageclientset.ImageStreamTagsGetter, artifactDir string, jobSpec *api.JobSpec, dryLogger *DryLogger, pullSecret *coreapi.Secret) api.Step {
 	return &rpmImageInjectionStep{
 		config:      config,
 		resources:   resources,
@@ -87,5 +90,6 @@ func RPMImageInjectionStep(config api.RPMImageInjectionStepConfiguration, resour
 		artifactDir: artifactDir,
 		jobSpec:     jobSpec,
 		dryLogger:   dryLogger,
+		pullSecret:  pullSecret,
 	}
 }
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -44,6 +44,8 @@ const (
 	oauthToken    = "/oauth-token"
 
 	OauthSecretKey = "oauth-token"
+
+	PullSecretName = "regcred"
 )
 
 type CloneAuthType string
@@ -148,6 +150,7 @@ type sourceStep struct {
 	jobSpec            *api.JobSpec
 	dryLogger          *DryLogger
 	cloneAuthConfig    *CloneAuthConfig
+	pullSecret         *coreapi.Secret
 }
 
 func (s *sourceStep) Inputs(ctx context.Context, dry bool) (api.InputDefinition, error) {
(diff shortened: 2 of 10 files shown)


hongkailiu:
working on rehearsals. Will fix the test later.

hongkailiu:
"image-registry.openshift-image-registry.svc:5000"

Need add auth for internal svc dns: 
```
Pulling image image-registry.openshift-image-registry.svc:5000/ci-op-jx0is4sl/pipeline:root ...
```
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/6420/rehearse-6420-pull-ci-openshift-installer-master-images/5#1:build-log.txt%3A41

hongkailiu:
/hold

retesting with the latest image.

hongkailiu:
/hold cancel

https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/6420/rehearse-6420-pull-ci-openshift-installer-master-images/7

pkg/steps/source.go
```diff
@@ -293,6 +297,11 @@ func buildFromSource(jobSpec *api.JobSpec, fromTag, toTag api.PipelineImageStrea
 			},
 		},
 	}
+	if _, err := secretClient.Secrets(jobSpec.Namespace).Get(PullSecretName, meta.GetOptions{}); err == nil {
+		build.Spec.Strategy.DockerStrategy.PullSecret = getSourceSecretFromName(PullSecretName)
+	} else if errors.IsNotFound(err) {
+		log.Printf("Cannot find pull secret '%s', ignoring set it up in the build: %v\n", PullSecretName, err)
```
stevekuznetsov: Why is this just a log? Won't it make the build fail?

hongkailiu: OK ... here is the logic:

Only when `--image-import-pull-secret` is provided to `ci-operator`, we have the secret created. For now we have it provided only for migrated jobs to build01.

So before we migrate all jobs, we will keep the backward-compatibility?

stevekuznetsov: So shouldn't we be able to tell when someone does or does not pass that flag, and only do this if they pass it?

hongkailiu: My first version was like that ... passing the option.pullsecretpath all the way to the changed function.
But i think the current one is better. If it passes the arg, we will create. Then we can get it by client.

hongkailiu: Updated.

hongkailiu:
Latest code:
https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/6420/rehearse-6420-pull-ci-openshift-installer-master-images/8

pkg/steps/rpm_injection.go
```diff
@@ -5,11 +5,11 @@ import (
 	"fmt"
 
 	buildapi "github.com/openshift/api/build/v1"
+	"github.com/openshift/ci-tools/pkg/api"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	routeclientset "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
+	coreapi "k8s.io/api/core/v1"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	"github.com/openshift/ci-tools/pkg/api"
```
droslean: lets keep the order here.

pkg/steps/source.go
```diff
@@ -293,6 +296,9 @@ func buildFromSource(jobSpec *api.JobSpec, fromTag, toTag api.PipelineImageStrea
 			},
 		},
 	}
+	if pullSecret != nil {
```
droslean: There is no reason to pass the whole *coreapi.Secret, just only to check if its nil or not. Is there any other reason why we can't use a bool instead?

pkg/steps/source_test.go
```diff
@@ -685,7 +685,7 @@ RUN rm -f /oauth-token
 
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			if actual, expected := createBuild(testCase.config, testCase.jobSpec, testCase.clonerefsRef, testCase.resources, testCase.cloneAuthConfig), testCase.expected; !equality.Semantic.DeepEqual(actual, expected) {
+			if actual, expected := createBuild(testCase.config, testCase.jobSpec, testCase.clonerefsRef, testCase.resources, testCase.cloneAuthConfig, nil), testCase.expected; !equality.Semantic.DeepEqual(actual, expected) {
```
droslean: It's better to have a test case for this.

pkg/steps/rpm_injection.go
hongkailiu: Updated.

pkg/steps/source.go
hongkailiu: Maybe pointer/address is fine here?
The original arg for the program is `--pull-secret-path` which is used to init the secret object.
I feel passing the secret (pointer) through makes more sense than the path: such as an invalid path.

pkg/steps/source_test.go
hongkailiu: Will work on this.

hongkailiu: Test case updated.

stevekuznetsov:
Also would be nice to see this play out in an integration test

hongkailiu:
> Instead of changing the basic test, why not add a new test that shows off what happens. Changing a current test case to do something new and different lowers our test coverage instead of improving it.

Done. Working on I.T. now.

hongkailiu:
> Also would be nice to see this play out in an integration test

Done.

test/ci-operator-integration/base/run.sh
```diff
@@ -35,6 +39,7 @@ run_test() {
         --determinize-output \
         --namespace "${TEST_NAMESPACE}" \
         --config "${TEST_CONFIG}" \
+        --image-import-pull-secret "${PULL_SECRET_PATH}" \
```
stevekuznetsov: What you've done here is the same as what you did in the Go tests. Each invocation of `run_test` in this script is analogous to a test case, and when you add this arg to every single possible test invocation, you're muddying the waters. A good test will test something in a targeted manner. Imagine what happens when someone makes a change that breaks this functionality -- they will run the tests and see every single integration test fail. Then, they'll read what the tests are doing and try to figure out why their change had that effect. Instead, let's add a new test case here that specifically sets this flag so if someone breaks it in the future, they can see one isolated test case (that's clearly labeled as testing this feature) fail.

hongkailiu: Ack. Will add a new case.

hongkailiu: I guess a new case will be a new folder under `./test/ci-operator-integration`

hongkailiu: Done.

PR #345, opened 2020-01-02
Set LC_ALL in artifacts so that aws cli would convert unicode correctly

vrutkovs:
* Set LC_ALL so that py2-based `aws ec2 get-console-output` wouldn't crash with UnicodeEncodeError
* Store only text (json by default) in output file

Backport of https://github.com/openshift/release/pull/6536

/cc @wking

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -717,6 +717,8 @@ objects:
         value: test
       - name: HOME
         value: /tmp
+      - name: LC_ALL
+        value: en_US.UTF-8
       command:
       - /bin/bash
       - -c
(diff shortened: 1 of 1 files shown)


wking:
/lgtm

Although I'm not an approver over here.

wking:
Or maybe [I am][1] :p.

[1]: https://github.com/openshift/ci-tools/commit/893caf869445ae4b58ba4238225b06a34e7f5a8c#diff-4775ec094d07c9d17dffabc5bf6f08f3R8

wking:
/retitle Set LC_ALL in artifacts so that aws cli would convert unicode correctly

PR #353, opened 2020-01-08
Implement ci-secret-bootstrap

hongkailiu:
https://issues.redhat.com/browse/DPTP-639

/assign @stevekuznetsov

### DIFF
--- a/cmd/ci-secret-bootstrap/README.md
+++ b/cmd/ci-secret-bootstrap/README.md
@@ -3,6 +3,8 @@
 This tool extends the [populate-secrets-from-bitwarden.sh](https://github.com/openshift/release/blob/c8c89d08c56c653b91eb8c7580657f7ce522253f/ci-operator/populate-secrets-from-bitwarden.sh)
 to support mirroring secrets cross Kubernetes/OpenShift-clusters.
 
+## Args and config.yaml
+
 We use `--kubeconfig` to specify the path to a [kube config](https://kubernetes.io/docs/concepts/configuration/organize-cluster-access-kubeconfig/)
 that the tool will load and use it to access clusters for writing secrets.
 
@@ -10,25 +12,24 @@ It expects a configuration like the one below which specifies the mapping from t
 in BitWarden and the targeting secret.
 
 ```yaml
-secrets:
 - from:
     key-name-1:
-      bw-item: item-name-1
+      bw_item: item-name-1
       field: field-name-1
     key-name-2:
-      bw-item: item-name-1
+      bw_item: item-name-1
       field: field-name-2
     key-name-3:
-      bw-item: item-name-1
+      bw_item: item-name-1
       attachment: attachment-name-1
     key-name-4:
-      bw-item: item-name-2
+      bw_item: item-name-2
       field: field-name-1
     key-name-5:
-      bw-item: item-name-2
+      bw_item: item-name-2
       attachment: attachment-name-1
     key-name-6:
-      bw-item: item-name-3
+      bw_item: item-name-3
       attachment: attachment-name-2
   to:
     - cluster: default
@@ -58,3 +59,14 @@ And then the secret will be populated to
 
 * the `secret` `prod-secret-1` in `namespace-1` on the `default` cluster, and
 * the `secret` `prod-secret-2` in `namespace-2` on the `build01` cluster.
+
+
+## Run
+
+```bash
+$ echo -n "bw_password" > /tmp/bw_password 
+$ ci-secret-bootstrap --bw-password-path=/tmp/bw_password -bw-user kerberos_id@redhat.com --kubeconfig <path_to_kubeconfig_file> --config <path_to_config.yaml>
+
+```
+
+where `kubeconfig` contains the `contexts` for the `default` cluster and the `build01` cluster.
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
@@ -16,8 +17,10 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/test-infra/prow/logrusutil"
 
 	"github.com/openshift/ci-tools/pkg/bitwarden"
 	"github.com/openshift/ci-tools/pkg/util"
@@ -64,6 +67,7 @@ func (o *options) validateOptions() error {
 
 func (o *options) completeOptions() error {
 	bytes, err := ioutil.ReadFile(o.bwPasswordPath)
+	secrets.addSecrets(string(bytes))
 	if err != nil {
 		return err
 	}
@@ -245,6 +249,32 @@ func printSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
 		}
 	}
 	return nil
+}
+
+type secretGetter struct {
+	sync.RWMutex
+	secrets sets.String
+}
+
+func (g *secretGetter) addSecrets(newSecrets ...string) {
+	g.Lock()
+	defer g.Unlock()
+	g.secrets.Insert(newSecrets...)
+}
+
+func (g *secretGetter) getSecrets() sets.String {
+	g.RLock()
+	defer g.RUnlock()
+	return g.secrets
+}
+
+var (
+	secrets *secretGetter
+)
+
+func init() {
+	secrets = &secretGetter{secrets: sets.NewString()}
+	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, secrets.getSecrets))
 }
 
 func main() {
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -1,0 +1,277 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
+	"sort"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	coreapi "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime/serializer/json"
+	"k8s.io/client-go/kubernetes/scheme"
+	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"github.com/openshift/ci-tools/pkg/bitwarden"
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+type options struct {
+	dryRun         bool
+	kubeConfigPath string
+	configPath     string
+	bwUser         string
+	bwPasswordPath string
+
+	bwPassword     string
+	secretsGetters map[string]coreclientset.SecretsGetter
+	config         []secretConfig
+}
+
+func parseOptions() options {
+	var o options
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
+	fs.StringVar(&o.kubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
+	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
+	fs.StringVar(&o.bwUser, "bw-user", "", "Username to access BitWarden.")
+	fs.StringVar(&o.bwPasswordPath, "bw-password-path", "", "Path to a password file to access BitWarden.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Errorf("cannot parse args: %q", os.Args[1:])
+	}
+	return o
+}
+
+func (o *options) validateOptions() error {
+	if o.bwUser == "" {
+		return fmt.Errorf("--bw-user is empty")
+	}
+	if o.bwPasswordPath == "" {
+		return fmt.Errorf("--bw-password-path is empty")
+	}
+	if o.configPath == "" {
+		return fmt.Errorf("--config is empty")
+	}
+	return nil
+}
+
+func (o *options) completeOptions() error {
+	bytes, err := ioutil.ReadFile(o.bwPasswordPath)
+	if err != nil {
+		return err
+	}
+	o.bwPassword = strings.TrimSpace(string(bytes))
+
+	kubeConfigs, _, err := util.LoadKubeConfigs(o.kubeConfigPath)
+	if err != nil {
+		return err
+	}
+
+	bytes, err = ioutil.ReadFile(o.configPath)
+	if err != nil {
+		return err
+	}
+	var config []secretConfig
+	err = yaml.Unmarshal(bytes, &config)
+	if err != nil {
+		return err
+	}
+	o.config = config
+
+	o.secretsGetters = map[string]coreclientset.SecretsGetter{}
+	for i, secretConfig := range o.config {
+		for j, secretContext := range secretConfig.To {
+			if o.secretsGetters[secretContext.Cluster] == nil {
+				kc, ok := kubeConfigs[secretContext.Cluster]
+				if !ok {
+					return fmt.Errorf("config[%d].to[%d]: failed to find cluster context %q in the kubeconfig", i, j, secretContext.Cluster)
+				}
+				client, err := coreclientset.NewForConfig(&kc)
+				if err != nil {
+					return err
+				}
+				o.secretsGetters[secretContext.Cluster] = client
+			}
+		}
+	}
+
+	return o.validateCompletedOptions()
+}
+
+func (o *options) validateCompletedOptions() error {
+	if o.bwPassword == "" {
+		return fmt.Errorf("--bw-password-file was empty")
+	}
+	if len(o.config) == 0 {
+		return fmt.Errorf("--config was empty")
+	}
+	toMap := map[string]map[string]string{}
+	for i, secretConfig := range o.config {
+		if len(secretConfig.From) == 0 {
+			return fmt.Errorf("config[%d].from is empty", i)
+		}
+		if len(secretConfig.To) == 0 {
+			return fmt.Errorf("config[%d].to is empty", i)
+		}
+		for key, bwContext := range secretConfig.From {
+			if key == "" {
+				return fmt.Errorf("config[%d].from: empty key is not allowed", i)
+			}
+			if bwContext.BWItem == "" {
+				return fmt.Errorf("config[%d].from[%s]: empty value is not allowed", i, key)
+			}
+			if bwContext.Field == "" && bwContext.Attachment == "" {
+				return fmt.Errorf("config[%d].from[%s]: either field or attachment needs to be non-empty", i, key)
+			}
+			if bwContext.Field != "" && bwContext.Attachment != "" {
+				return fmt.Errorf("config[%d].from[%s]: cannot use both field and attachment", i, key)
+			}
+		}
+		for j, secretContext := range secretConfig.To {
+			if secretContext.Cluster == "" {
+				return fmt.Errorf("config[%d].to[%d].cluster: empty value is not allowed", i, j)
+			}
+			if secretContext.Namespace == "" {
+				return fmt.Errorf("config[%d].to[%d].namespace: empty value is not allowed", i, j)
+			}
+			if secretContext.Name == "" {
+				return fmt.Errorf("config[%d].to[%d].name: empty value is not allowed", i, j)
+			}
+
+			if toMap[secretContext.Cluster] == nil {
+				toMap[secretContext.Cluster] = map[string]string{secretContext.Namespace: secretContext.Name}
+			} else if toMap[secretContext.Cluster][secretContext.Namespace] != secretContext.Name {
+				toMap[secretContext.Cluster][secretContext.Namespace] = secretContext.Name
+			} else {
+				return fmt.Errorf("config[%d].to[%d]: secret %v listed more than once in the config", i, j, secretContext)
+			}
+		}
+	}
+	return nil
+}
+
+type bitWardenContext struct {
+	BWItem     string `json:"bw_item"`
+	Field      string `json:"field,omitempty"`
+	Attachment string `json:"attachment,omitempty"`
+}
+
+type secretContext struct {
+	Cluster   string `json:"cluster"`
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
+type secretConfig struct {
+	From map[string]bitWardenContext `json:"from"`
+	To   []secretContext             `json:"to"`
+}
+
+func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[string][]*coreapi.Secret, error) {
+	secretsMap := map[string][]*coreapi.Secret{}
+	for _, secretConfig := range config {
+		data := make(map[string][]byte)
+		for key, bwContext := range secretConfig.From {
+			var value []byte
+			var err error
+			if bwContext.Field != "" {
+				value, err = bwClient.GetFieldOnItem(bwContext.BWItem, bwContext.Field)
+			} else {
+				value, err = bwClient.GetAttachmentOnItem(bwContext.BWItem, bwContext.Attachment)
+			}
+			if err != nil {
+				return nil, err
+			}
+			data[key] = value
+		}
+		for _, secretContext := range secretConfig.To {
+			secret := &coreapi.Secret{
+				Data: data,
+				ObjectMeta: meta.ObjectMeta{
+					Name:      secretContext.Name,
+					Namespace: secretContext.Namespace,
+					Labels:    map[string]string{"ci.openshift.org/auto-managed": "true"},
+				},
+			}
+			secretsMap[secretContext.Cluster] = append(secretsMap[secretContext.Cluster], secret)
+		}
+	}
+	return secretsMap, nil
+}
+
+func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secretsMap map[string][]*coreapi.Secret) error {
+	for cluster, secrets := range secretsMap {
+		for _, secret := range secrets {
+			secretsGetter := secretsGetters[cluster]
+			if _, err := secretsGetter.Secrets(secret.Namespace).Create(secret); err != nil {
+				if !kerrors.IsAlreadyExists(err) {
+					return err
+				}
+				if _, err := secretsGetter.Secrets(secret.Namespace).Update(secret); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func printSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
+	var clusters []string
+	for cluster := range secretsMap {
+		clusters = append(clusters, cluster)
+	}
+	sort.Strings(clusters)
+	for _, cluster := range clusters {
+		if _, err := fmt.Fprintln(w, fmt.Sprintf("###%s###", cluster)); err != nil {
+			return err
+		}
+		for _, secret := range secretsMap[cluster] {
+			if _, err := fmt.Fprintln(w, "---"); err != nil {
+				return err
+			}
+			s := json.NewSerializerWithOptions(json.DefaultMetaFactory, scheme.Scheme,
+				scheme.Scheme, json.SerializerOptions{Yaml: true, Pretty: true, Strict: true})
+			if err := s.Encode(secret, w); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func main() {
+	o := parseOptions()
+	if err := o.validateOptions(); err != nil {
+		logrus.WithError(err).Fatal("Invalid arguments.")
+	}
+	if err := o.completeOptions(); err != nil {
+		logrus.WithError(err).Fatal("Failed to complete options.")
+	}
+	bwClient, err := bitwarden.NewClient(o.bwUser, o.bwPassword)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to get Bitwarden client.")
+	}
+
+	secretsMap, err := constructSecrets(o.config, bwClient)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to get secrets.")
+	}
+
+	if o.dryRun {
+		if err := printSecrets(secretsMap, os.Stdout); err != nil {
+			logrus.WithError(err).Fatalf("Failed to print secrets on dry run.")
+		}
+	} else {
+		if err := updateSecrets(o.secretsGetters, secretsMap); err != nil {
+			logrus.WithError(err).Fatalf("Failed to update secrets.")
+		}
+	}
+}
--- a/cmd/ci-secret-bootstrap/main_test.go
+++ b/cmd/ci-secret-bootstrap/main_test.go
@@ -1,0 +1,1174 @@
+package main
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"reflect"
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	coreapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes/fake"
+	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+
+	"github.com/openshift/ci-tools/pkg/bitwarden"
+)
+
+func TestParseOptions(t *testing.T) {
+	testCases := []struct {
+		name     string
+		given    []string
+		expected options
+	}{
+		{
+			name:  "basic case",
+			given: []string{"cmd", "--dry-run=false", "--bw-user=username", "--bw-password-path=/tmp/bw-password", "--config=/tmp/config"},
+			expected: options{
+				bwUser:         "username",
+				bwPasswordPath: "/tmp/bw-password",
+				configPath:     "/tmp/config",
+			},
+		},
+		{
+			name:  "with kubeconfig",
+			given: []string{"cmd", "--dry-run=false", "--bw-user=username", "--bw-password-path=/tmp/bw-password", "--config=/tmp/config", "--kubeconfig=/tmp/kubeconfig"},
+			expected: options{
+				bwUser:         "username",
+				bwPasswordPath: "/tmp/bw-password",
+				configPath:     "/tmp/config",
+				kubeConfigPath: "/tmp/kubeconfig",
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			oldArgs := os.Args
+			defer func() { os.Args = oldArgs }()
+			os.Args = tc.given
+			actual := parseOptions()
+			if actual.dryRun != tc.expected.dryRun {
+				t.Errorf("%q: (dryRun) actual differs from expected:\n%s", tc.name, cmp.Diff(actual.dryRun, tc.expected.dryRun))
+			}
+			if actual.bwUser != tc.expected.bwUser {
+				t.Errorf("%q: (bwUser) actual differs from expected:\n%s", tc.name, cmp.Diff(actual.bwUser, tc.expected.bwUser))
+			}
+			if actual.bwPasswordPath != tc.expected.bwPasswordPath {
+				t.Errorf("%q: (bwPasswordPath) actual differs from expected:\n%s", tc.name, cmp.Diff(actual.bwPasswordPath, tc.expected.bwPasswordPath))
+			}
+			if actual.kubeConfigPath != tc.expected.kubeConfigPath {
+				t.Errorf("%q: (kubeConfigPath) actual differs from expected:\n%s", tc.name, cmp.Diff(actual.kubeConfigPath, tc.expected.kubeConfigPath))
+			}
+		})
+	}
+}
+
+func TestValidateOptions(t *testing.T) {
+	testCases := []struct {
+		name     string
+		given    options
+		expected error
+	}{
+		{
+			name: "basic case",
+			given: options{
+				bwUser:         "username",
+				bwPasswordPath: "/tmp/bw-password",
+				configPath:     "/tmp/config",
+			},
+		},
+		{
+			name: "empty bw user",
+			given: options{
+				bwPasswordPath: "/tmp/bw-password",
+				configPath:     "/tmp/config",
+			},
+			expected: fmt.Errorf("--bw-user is empty"),
+		},
+		{
+			name: "empty bw user password path",
+			given: options{
+				bwUser:     "username",
+				configPath: "/tmp/config",
+			},
+			expected: fmt.Errorf("--bw-password-path is empty"),
+		},
+		{
+			name: "empty config path",
+			given: options{
+				bwUser:         "username",
+				bwPasswordPath: "/tmp/bw-password",
+			},
+			expected: fmt.Errorf("--config is empty"),
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual := tc.given.validateOptions()
+			equalError(t, tc.expected, actual)
+		})
+	}
+}
+
+const (
+	configContent = `---
+- from:
+    key-name-1:
+      bw_item: item-name-1
+      field: field-name-1
+    key-name-2:
+      bw_item: item-name-1
+      field: field-name-2
+    key-name-3:
+      bw_item: item-name-1
+      attachment: attachment-name-1
+    key-name-4:
+      bw_item: item-name-2
+      field: field-name-1
+    key-name-5:
+      bw_item: item-name-2
+      attachment: attachment-name-1
+    key-name-6:
+      bw_item: item-name-3
+      attachment: attachment-name-2
+  to:
+    - cluster: default
+      namespace: namespace-1
+      name: prod-secret-1
+    - cluster: build01
+      namespace: namespace-2
+      name: prod-secret-2
+`
+	configContentWithTypo = `---
+- from:
+    key-name-1:
+      bw_item: item-name-1
+      field: field-name-1
+    key-name-2:
+      bw_item: item-name-1
+      field: field-name-2
+    key-name-3:
+      bw_item: item-name-1
+      attachment: attachment-name-1
+    key-name-4:
+      bw_item: item-name-2
+      field: field-name-1
+    key-name-5:
+      bw_item: item-name-2
+      attachment: attachment-name-1
+    key-name-6:
+      bw_item: item-name-3
+      attachment: attachment-name-2
+  to:
+    - cluster: default
+      namespace: namespace-1
+      name: prod-secret-1
+    - cluster: bla
+      namespace: namespace-2
+      name: prod-secret-2
+`
+	kubeConfigContent = `---
+apiVersion: v1
+clusters:
+- cluster:
+    server: https://api.ci.openshift.org:443
+  name: api-ci-openshift-org:443
+- cluster:
+    server: https://api.build01.ci.devcluster.openshift.com:6443
+  name: api-build01-ci-devcluster-openshift-com:6443
+contexts:
+- context:
+    cluster: api-build01-ci-devcluster-openshift-com:6443
+    namespace: ci
+    user: system:serviceaccount:ci:tool/api-build01-ci-devcluster-openshift-com:6443
+  name: build01
+- context:
+    cluster: api-ci-openshift-org:443
+    namespace: ci
+    user: system:serviceaccount:ci:tool/api-ci-openshift-org:443
+  name: default
+current-context: default
+kind: Config
+preferences: {}
+users:
+- name: system:serviceaccount:ci:tool/api-ci-openshift-org:443
+  user:
+    token: token1
+- name: system:serviceaccount:ci:tool/api-build01-ci-devcluster-openshift-com:6443
+  user:
+    token: token2
+`
+)
+
+var (
+	configDefault = rest.Config{
+		Host:        "https://api.ci.openshift.org:443",
+		BearerToken: "token1",
+	}
+	configBuild01 = rest.Config{
+		Host:        "https://api.build01.ci.devcluster.openshift.com:6443",
+		BearerToken: "token2",
+	}
+
+	defaultConfig = []secretConfig{
+		{
+			From: map[string]bitWardenContext{
+				"key-name-1": {
+					BWItem: "item-name-1",
+					Field:  "field-name-1",
+				},
+				"key-name-2": {
+					BWItem: "item-name-1",
+					Field:  "field-name-2",
+				},
+				"key-name-3": {
+					BWItem:     "item-name-1",
+					Attachment: "attachment-name-1",
+				},
+				"key-name-4": {
+					BWItem: "item-name-2",
+					Field:  "field-name-1",
+				},
+				"key-name-5": {
+					BWItem:     "item-name-2",
+					Attachment: "attachment-name-1",
+				},
+				"key-name-6": {
+					BWItem:     "item-name-3",
+					Attachment: "attachment-name-2",
+				},
+			},
+			To: []secretContext{
+				{
+					Cluster:   "default",
+					Namespace: "namespace-1",
+					Name:      "prod-secret-1",
+				},
+				{
+					Cluster:   "build01",
+					Namespace: "namespace-2",
+					Name:      "prod-secret-2",
+				},
+			},
+		},
+	}
+)
+
+func TestCompleteOptions(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test")
+	if err != nil {
+		t.Errorf("Failed to create temp dir")
+	}
+	defer func() {
+		if err := os.RemoveAll(dir); err != nil {
+			t.Errorf("Failed to remove temp dir")
+		}
+	}()
+
+	bwPasswordPath := filepath.Join(dir, "bwPasswordPath")
+	configPath := filepath.Join(dir, "configPath")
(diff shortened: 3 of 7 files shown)


cmd/ci-secret-bootstrap/bitwarden/bitwarden.go
```diff
@@ -0,0 +1,29 @@
...
+
+type Item struct {
+	ID          string        `json:"id"`
+	Name        string        `json:"name"`
+	Fields      []Field       `json:"fields"`
+	Attachments []Attachments `json:"attachments"`
+}
+
+type BitwardenClient interface {
+	Login(username, passward string) error
```
stevekuznetsov: Mis-spelling of `password`

cmd/ci-secret-bootstrap/bitwarden/cli.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (c *bitwardenCliClient) GetAttachment(itemID, AttachmentID string, raw bool) (string, error) {
+	return "", nil
+}
+
+func (c *bitwardenCliClient) Logout() error {
+	return nil
+}
+
+func newBitwardenCliClient() *bitwardenCliClient {
```
stevekuznetsov: Return the interface, not the type, and export the constructor

cmd/ci-secret-bootstrap/bitwarden/bitwarden.go
```diff
@@ -0,0 +1,29 @@
+package bitwarden
```
stevekuznetsov: This should be `pkg/bitwarden`

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func parseOptions() options {
+	var o options
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
+	fs.StringVar(&o.kubeconfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
+	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
+	fs.StringVar(&o.bwUser, "bw-user", "", "Username to access BitWarden.")
+	fs.StringVar(&o.bwPassword, "bw-password", "", "Password to access BitWarden.")
```
stevekuznetsov: This needs to be a file, not a plain text password

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Errorf("cannot parse args: '%s'", os.Args[1:])
+	}
+	return o
+}
+
+func (o *options) validateOptions() error {
+	for _, arg := range []string{o.bwUser, o.bwPassword, o.configPath} {
+		if arg == "" {
+			return fmt.Errorf("the mandatory arg '%s' is empty", arg)
```
stevekuznetsov: This will print `"the mandatory arg '' is empty"`

stevekuznetsov: Nit: use `%q`

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,111 @@
...
+		}
+	}
+	return nil
+}
+
+func (o *options) completeOptions() error {
+	kcPath := o.kubeconfigPath
+	if kcPath == "" {
+		kcPath = filepath.Join(homedir.HomeDir(), ".kube", "config")
```
stevekuznetsov: This is implicit in the loading rules upstream, you do not need it

cmd/ci-secret-bootstrap/bitwarden/bitwarden.go
hongkailiu: caused by so many `bitWARDens`

cmd/ci-secret-bootstrap/bitwarden/bitwarden.go
hongkailiu: Done.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: Removed.
IIRC, I had to to this. But will test it later.

cmd/ci-secret-bootstrap/main.go
hongkailiu: was trying to be start but failed.

cmd/ci-secret-bootstrap/bitwarden/cli.go
hongkailiu: Not sure if I did as you expected. Please check the current version.

cmd/ci-secret-bootstrap/README.md
```diff
@@ -13,22 +13,22 @@ in BitWarden and the targeting secret.
 secrets:
 - from:
     key-name-1:
-      bw-item: item-name-1
+      bwItem: item-name-1
```
petr-muller: Can we keep the lower case with dashes? Pretty much all our YAML uses that format.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+	"io/ioutil"
+	coreapi "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/rest"
+	"os"
+	"strings"
```
petr-muller: nit: imports need cleanup

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+		return err
+	}
+	o.config = config
+
+	return validateConfig(kubeConfigs, config)
+}
+
+type Config struct {
+	Secrets []SecretConfig `json:"secrets"`
+}
```
petr-muller: should this be a `type Secrets []SecretConfig`? Do you expect to add more fields here?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+
+type BWContext struct {
+	BWItem     string `json:"bwItem"`
+	Field      string `json:"field"`
+	Attachment string `json:"attachment"`
```
petr-muller: `Field` and `Attachment` are mutually exclusive, right? So these should be omitempty, and we should validate we have exactly non-empty one.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+type SecretContext struct {
+	Cluster   string `json:"cluster"`
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
```
petr-muller: We should validate to prevent overwriting a single destination from multiple sources.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/openshift/ci-tools/pkg/bitwarden"
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+type options struct {
+	dryRun         bool
+	kubeConfigPath string
+	configPath     string
+	bwUser         string
+	bwPasswordPath string
```
petr-muller: This is not that related to the actual code, but BW credentials are a immensely sensitive secret: leaking it means leaking all secrets. Are we planning to run this tool just manually, or will we deploy it somewhere where we'll need to put also the credentials?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}()
+
+	for _, secretConfig := range o.config.Secrets {
+		data := make(map[string][]byte)
+		for key, bwContext := range secretConfig.From {
+			if bwContext.Field != "" {
+				fieldValue, err := bwClient.GetField(bwContext.BWItem, bwContext.Field)
+				if err != nil {
+					logrus.WithError(err).Fatalf("Failed to get field value '%s' in item '%s'", bwContext.Field, bwContext.BWItem)
```
petr-muller: AFAIK `Fatalf` will cause the deferred `Logout` to not run

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+				},
+			}
+			kc := o.kubeConfigs[secretContext.Cluster]
+			client, err := coreclientset.NewForConfig(&kc)
+			if err != nil {
+				logrus.WithError(err).Fatalf("could not get core client for cluster '%s'", secretContext.Cluster)
+			}
+			_, err = client.Secrets(secretContext.Namespace).Create(secret)
+			if err != nil {
+				if kerrors.IsAlreadyExists(err) {
```
petr-muller: Inverting the condition and turning it into a guard clause would allow you to reduce some indentation depth

```
if !kerrors.IsAlreadyExists(err) {
  error out
}
if _, err := client.Secrets(secretContext.Namespace).Update(secret); err != nil {
  error out
}
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := o.completeOptions(); err != nil {
+		logrus.WithError(err).Fatal("Failed to complete options.")
+	}
+	bwClient := bitwarden.NewBitwardenClient()
+	if err := bwClient.Login(o.bwUser, o.bwPassword); err != nil {
+		logrus.WithError(err).Fatal("Failed to login Bitwarden.")
+	}
+	defer func() {
+		if err := bwClient.Logout(); err != nil {
+			logrus.WithError(err).Fatal("Failed to login Bitwarden.")
```
petr-muller: Failed to log**out**?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,182 @@
...
+		}
+	}()
+
+	for _, secretConfig := range o.config.Secrets {
+		data := make(map[string][]byte)
+		for key, bwContext := range secretConfig.From {
+			if bwContext.Field != "" {
+				fieldValue, err := bwClient.GetField(bwContext.BWItem, bwContext.Field)
```
petr-muller: You can make this all less verbose by only getting the value in the conditional and do the error handling, conversion and assignment just once:

```
var value string
var err error
if field {
  value, err = GetField()
} else {
  value, err = GetAttachment()
}
if err != nil {
  error out
}
data[key] = []byte(value)
```

cmd/ci-secret-bootstrap/main.go
hongkailiu: Validation has listed in TODO.
https://github.com/openshift/ci-tools/blob/abf6fe6b64a51e1e2f3157f5513b08585ee500d4/cmd/ci-secret-bootstrap/main.go#L117-L119

Is `omitempty` only for serialization?
I think we do only deserialization here.
I am adding it anyway just in case.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: yes, noticed that `panic` is called inside `Fatal`.
But I feel `logout` (or defer) is kind of best effort thing.

cmd/ci-secret-bootstrap/main.go
hongkailiu: Smart!
Updated.

One comment: Removing the else-part (for reduction of code complication) replies on the Fatal/Panic to keep the logic right.

cmd/ci-secret-bootstrap/README.md
hongkailiu: Updated as discussed over slack.
We choose our tools' style over k8s' one.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: Will do it today.

cmd/ci-secret-bootstrap/main.go
Thought about this too and chose the other way.

cmd/ci-secret-bootstrap/main.go
hongkailiu: Did not finalize this with Steve.
In my opinion, we should keep BW from our jobs and only run it locally/manually.
BW is too risky and hard to recover (James indicates that giving a user readonly access is not possible):
- rotation all is hard
- no git-blame or git-reset for BW

cmd/ci-secret-bootstrap/main.go
hongkailiu: Will do today.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,247 @@
...
+	config, err := loadConfig(o.configPath)
+	if err != nil {
+		return err
+	}
+	o.config = config
+
+	return o.validateCompletedOptions()
+}
+
+func (o *options) validateCompletedOptions() error {
```
stevekuznetsov: Please use the cascading prefix pattern to identify exactly which field is invalid. For instance, instead of

```
found empty key whose value is 'thing'`
```

You should use:

```
secrets[9].from.thing: value cannot be empty
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,247 @@
...
+	if o.bwPassword == "" {
+		return fmt.Errorf("found empty file: '%s'", o.bwPasswordPath)
```
stevekuznetsov: A quoted value is `%q`

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,247 @@
...
+			}
+			if secretContext.Name == "" {
+				return fmt.Errorf("found empty name value: '%v'", secretContext)
+			}
+			if _, ok := o.kubeConfigs[secretContext.Cluster]; !ok {
+				return fmt.Errorf("failed to find cluster context in the kubeconfig: '%s'", secretContext.Cluster)
+			}
+			secretContextString := fmt.Sprintf("%v", secretContext)
+			if toSet.Has(secretContextString) {
+				return fmt.Errorf("secret context '%s' showed more than once in the config", secretContextString)
```
stevekuznetsov: If this is the case, can you `map[cluster]map[namespace]name` ?

stevekuznetsov: (don't use lists when you want maps)

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,247 @@
...
+			}
+			toSet.Insert(secretContextString)
+		}
+	}
+	return nil
+}
+
+type Config = []SecretConfig
+
+type BWContext struct {
```
stevekuznetsov: nit: `s/BW/BitWarden/g`

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,247 @@
...
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
+type SecretConfig struct {
+	From map[string]BWContext `json:"from"`
+	To   []SecretContext      `json:"to"`
+}
+
+func loadConfig(path string) (Config, error) {
```
stevekuznetsov: Why is this a function? Is it called more than once?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,247 @@
...
+				if _, err := secretsGetter.Secrets(secretContext.Namespace).Update(secret); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func getSecretGetter(kubeConfigs map[string]rest.Config, cluster string) (coreclientset.SecretsGetter, error) {
```
stevekuznetsov: Why is this a function? Accessing a map seems pretty straightforward

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -0,0 +1,1148 @@
...
+			if tc.expected == nil {
+				assert.Nil(t, actual)
+			} else {
+				assert.EqualError(t, actual, tc.expected.Error())
+			}
+		})
+	}
+}
+
+func getGetFakeSecretGetter(clients map[string]*fake.Clientset) func(map[string]rest.Config, string) (coreclientset.SecretsGetter, error) {
```
stevekuznetsov: Couldn't you just set a different set of `clients`?

stevekuznetsov: The actual logic for choosing which client should not be mocked out -- this code will not test, for instance, that the production code actually chooses the correct client to use for each secret

pkg/bitwarden/bitwarden.go
```diff
@@ -0,0 +1,29 @@
...
+}
+
+type Item struct {
+	ID          string        `json:"id"`
+	Name        string        `json:"name"`
+	Fields      []Field       `json:"fields"`
+	Attachments []Attachments `json:"attachments"`
+}
+
+type BitwardenClient interface {
```
stevekuznetsov: This sutters. `Client` is fine

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -0,0 +1,1148 @@
...
+				assert.Nil(t, actual)
+			} else {
+				assert.EqualError(t, actual, tc.expected.Error())
+			}
+			actualSecretsOnDefault, err := fkcDefault.CoreV1().Secrets("").List(metav1.ListOptions{})
+			assert.Nil(t, err)
+			assert.Equal(t, tc.expectedSecretsOnDefault, actualSecretsOnDefault.Items)
+
+			actualSecretsOnBuild01, err := fkcBuild01.CoreV1().Secrets("").List(metav1.ListOptions{})
+			assert.Nil(t, err)
```
stevekuznetsov: Why? Just write Go code -- `if err != nil { t.Errorf() }`

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -0,0 +1,1148 @@
...
+			assert.Equal(t, tc.expectedSecretsOnBuild01, actualSecretsOnBuild01.Items)
```
stevekuznetsov: Use `diff.ObjectReflectDiff` and `reflect.DeepEqual` like we do in other tests. There's no utility to using new dependencies here.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: Removed.

pkg/bitwarden/bitwarden.go
hongkailiu: Done.

cmd/ci-secret-bootstrap/main.go
hongkailiu: Done. With some workaround to make the `run` function testable.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,240 @@
...
+	fs.StringVar(&o.bwPasswordPath, "bw-password-path", "", "Path to a password file to access BitWarden.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Errorf("cannot parse args: '%s'", os.Args[1:])
+	}
+	return o
+}
+
+func (o *options) validateOptions() error {
+	if o.bwUser == "" {
+		return fmt.Errorf("the mandatory arg bw-user is empty")
```
stevekuznetsov: `s/mandatory arg bw-user/--bw-user/` -- these are flags, not args, anyway

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,240 @@
...
+	for i, secretConfig := range o.config {
+		if len(secretConfig.From) == 0 {
+			return fmt.Errorf("found empty 'From'")
+		}
+		if len(secretConfig.To) == 0 {
+			return fmt.Errorf("found empty 'To'")
+		}
+		for key, bwContext := range secretConfig.From {
+			if key == "" {
+				return fmt.Errorf("config[%d].From: empty key is not allowed", i)
```
stevekuznetsov: Not `From`, `from` -- the user is giving you a YAML where `from` is the key they incorrectly set

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+			if bwContext.BWItem == "" {
+				return fmt.Errorf("found empty bw_item value for key '%s'", key)
```
stevekuznetsov: I literally mean every single error message needs a prefix telling the user which field was wrong

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,240 @@
...
+	bwPasswordPath string
+
+	bwPassword    string
+	secretClients map[string]secretClient
+	config        config
+}
+
+type secretClient struct {
+	secretsGetter coreclientset.SecretsGetter
+	config        rest.Config
```
stevekuznetsov: What is this for? You should be able to just use the client?

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: Really tried. The other one is for test.
Once it changes to getter, I cannot do the `equal` on them because unimported fields.

hongkailiu: Ignore my non-sense.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/config.yaml
```diff
@@ -0,0 +1,96 @@
...
+  to:
+    - cluster: default
+      namespace: ci
+      name: mirror.openshift.com
+    - cluster: default
+      namespace: ocp
+      name: mirror.openshift.com
+    - cluster: build01
+      namespace: ocp
+      name: mirror.openshift.com
```
hongkailiu: @stevekuznetsov 

Should I throw an error on this case?
Same content from BW is mapped to 2 secrets in the same cluster?
The sync of secrets inside a cluster can be done with ci-secret-mirroring-controller.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -0,0 +1,1420 @@
...
+				},
+			},
+			expectedClusters: []string{"build01", "default"},
+		},
+		{
+			name: "production case",
+			given: options{
+				bwUser:         "username",
+				bwPasswordPath: bwPasswordPath,
+				configPath:     "config.yaml",
```
hongkailiu: This one checks the production `config.yaml`.

It will result in the fact that each time we change on the production config, we need to update the test.
I think it is a "not-too-bad" burden because it gives us a chance to double check the changes on correctness of production config.

cmd/ci-secret-bootstrap/config.yaml
stevekuznetsov: The mirroring controller was built in order to allow a user who does not have access to production namespaces to  create a secret somewhere else and have it mirrored in, allowing them to manage it. That use-case is distinct from happening to have two secrets on the cluster that are from the same data. This should not error and we should be OK to populate from BitWarden.

cmd/ci-secret-bootstrap/main_test.go
stevekuznetsov: Please. no -- that sounds like a massive pain in the ass and will break people working on `ci-tools`

hongkailiu: Removed the case although I cannot see why it will break people working on `ci-tools`.

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+		return fmt.Errorf("bw-login succeeded with empty '.data.raw'")
+	}
+	// should never happen
+	return fmt.Errorf("bw-login failed without run's error")
+}
+
+func (c bitwardenCliClient) listItems() ([]Item, error) {
+	var items []Item
+	if c.sd.session == "" {
+		return items, fmt.Errorf("not yet login to Bitwarden")
```
stevekuznetsov: Make it impossible for this to happen in the program, instead of checking for this at run-time.

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+		return items, err
+	}
+	return items, nil
+}
+
+func (c bitwardenCliClient) GetField(itemName, filedName string) (string, error) {
+	if c.sd.session == "" {
+		return "", fmt.Errorf("not yet login to Bitwarden")
+	}
+	if len(c.sd.savedItems) == 0 {
```
stevekuznetsov: make `listItems()` encapsulated and opaque -- a caller should never need to know that `listItems()` has caching behavior

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+		if _, err := c.listItems(); err != nil {
+			return "", err
+		}
+	}
+	for _, i := range c.sd.savedItems {
```
stevekuznetsov: for instance -- callers should *always* call `listItems()` and should never access this field

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+		return items, err
+	}
+	err = json.Unmarshal(out, &items)
+	if err != nil {
+		return items, err
+	}
+	return items, nil
+}
+
+func (c bitwardenCliClient) GetField(itemName, filedName string) (string, error) {
```
stevekuznetsov: This is `GetFieldOnItem`

cmd/ci-secret-bootstrap/config.yaml
hongkailiu: It does the some overlapping jobs with

https://github.com/openshift/release/blob/master/core-services/secret-mirroring/_mapping.yaml#L171-L176

pkg/bitwarden/cli.go
stevekuznetsov: Or refactor your code to chain -- `GetItem(itemName).GetField(fieldName)` / `GetItem(itemName).GetAttachment(fieldName)`

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+			return "", err
+		}
+	}
+	for _, i := range c.sd.savedItems {
+		if i.Name == itemName {
+			for _, attachment := range i.Attachments {
+				if attachment.FileName == attachmentName {
+					//bw --session "${BW_SESSION}" get attachment "${attachment_id}" --itemid "${item_id}" --raw
+					out, err := c.run("bw", "--session", c.sd.session, "get", "attachment", attachment.ID, "--itemid", itemName, "--raw")
+					return string(out), err
```
stevekuznetsov: Can we keep operating on `[]byte`?

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+				}
+			}
+		}
+	}
+	return "", fmt.Errorf("not found attachment %q in item %q", attachmentName, itemName)
+}
+
+func (c bitwardenCliClient) Logout() error {
+	_, err := c.run("bw", "logout", "--response")
+	if err != nil {
```
stevekuznetsov: nit: you don't use `err` outside of this context so 


```go
if _, err := c.run(/*...*/); err != nil {

}
```

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+
+func run(cmd string, args ...string) ([]byte, error) {
+	if err := validate(cmd, args...); err != nil {
+		return nil, err
+	}
+	logrus.WithField("cmd", cmd).WithField("args", args).Info("running command ...")
+	return exec.Command(cmd, args...).Output()
+}
+
+func validate(cmd string, args ...string) error {
```
stevekuznetsov: For every check in this function, you have written the wrong abstraction in `run`. Make it _impossible_ for another developer to call `run` incorrectly, using the structure of your abstraction.

hongkailiu:
Saved.
[config.yaml.zip](https://github.com/openshift/ci-tools/files/4056960/config.yaml.zip)

cmd/ci-secret-bootstrap/config.yaml
```diff
@@ -0,0 +1,96 @@
+- from:
```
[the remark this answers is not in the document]
hongkailiu: Removed.
Will be in release repo after this PR.

cmd/ci-secret-bootstrap/README.md
```diff
@@ -58,3 +59,15 @@ And then the secret will be populated to
 
 * the `secret` `prod-secret-1` in `namespace-1` on the `default` cluster, and
 * the `secret` `prod-secret-2` in `namespace-2` on the `build01` cluster.
+
+
+## Run
+
+Run the tool with the assumption that `kubeconfig` is available at `${HOME}/.kube/config` with `contexts` for the `default`
```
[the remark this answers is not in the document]
hongkailiu: Updated.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	o.config = config
+
+	return o.validateCompletedOptions(kubeConfigs)
+}
+
+func (o *options) validateCompletedOptions(kubeConfigs map[string]rest.Config) error {
+	o.secretsGetters = map[string]coreclientset.SecretsGetter{}
+	if o.bwPassword == "" {
+		return fmt.Errorf("found empty file: %q", o.bwPasswordPath)
```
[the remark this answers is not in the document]

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	if len(o.config) == 0 {
+		return fmt.Errorf("found empty config")
```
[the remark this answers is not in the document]

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,264 @@
...
+				return fmt.Errorf("config[%d].to[%d].namespace: empty value is not allowed", i, j)
+			}
+			if secretContext.Name == "" {
+				return fmt.Errorf("config[%d].to[%d].name: empty value is not allowed", i, j)
+			}
+			if _, ok := kubeConfigs[secretContext.Cluster]; !ok {
+				return fmt.Errorf("config[%d].to[%d]: failed to find cluster context %q in the kubeconfig", i, j, secretContext.Cluster)
+			}
+			if o.dryRun {
+				o.secretsGetters[secretContext.Cluster] = fake.NewSimpleClientset().CoreV1()
```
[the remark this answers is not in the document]
hongkailiu: Moved.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,264 @@
...
+				toMap[secretContext.Cluster][secretContext.Namespace] = secretContext.Name
+			} else {
+				toMap[secretContext.Cluster] = map[string]string{secretContext.Namespace: secretContext.Name}
+			}
+		}
+	}
+	return nil
+}
+
+type config = []secretConfig
```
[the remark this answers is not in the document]
hongkailiu: Removed.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,264 @@
...
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
+type secretConfig struct {
+	From map[string]bitWardenContext `json:"from"`
+	To   []secretContext             `json:"to"`
+}
+
+func run(secretsGetters map[string]coreclientset.SecretsGetter, config config, bwClient bitwarden.Client, dryRun bool) error {
```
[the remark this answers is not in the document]

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,161 @@
...
+	Data    struct {
+		Raw string `json:"raw"`
+	} `json:"data"`
+}
+
+var (
+	defaultCliRun = run
+)
+
+type sessionData struct {
```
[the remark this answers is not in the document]
hongkailiu: Flat it out.

pkg/bitwarden/cli.go
hongkailiu: Embedded the logic into the constructor.

pkg/bitwarden/cli.go
hongkailiu: Done.

pkg/bitwarden/cli.go
hongkailiu: Done.

pkg/bitwarden/cli.go
hongkailiu: Hopefully it has a better shape now.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,264 @@
...
+				o.secretsGetters[secretContext.Cluster] = fake.NewSimpleClientset().CoreV1()
+			} else {
+				kc := kubeConfigs[secretContext.Cluster]
+				client, err := coreclientset.NewForConfig(&kc)
+				if err != nil {
+					return err
+				}
+				o.secretsGetters[secretContext.Cluster] = client
+			}
+			if toMap[secretContext.Cluster] != nil {
```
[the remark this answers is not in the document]

pkg/bitwarden/cli.go
hongkailiu: Done.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,297 @@
...
+				return fmt.Errorf("config[%d].to[%d].namespace: empty value is not allowed", i, j)
+			}
+			if secretContext.Name == "" {
+				return fmt.Errorf("config[%d].to[%d].name: empty value is not allowed", i, j)
+			}
+			if _, ok := kubeConfigs[secretContext.Cluster]; !ok {
+				return fmt.Errorf("config[%d].to[%d]: failed to find cluster context %q in the kubeconfig", i, j, secretContext.Cluster)
+			}
+
+			kc := kubeConfigs[secretContext.Cluster]
```
stevekuznetsov: This is still here and still not validation. Validation should do validation. Create your clients elsewhere.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,297 @@
...
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
+type secretConfig struct {
+	From map[string]bitWardenContext `json:"from"`
+	To   []secretContext             `json:"to"`
+}
+
+func getSecrets(config []secretConfig, bwClient bitwarden.Client) (map[string][]*coreapi.Secret, error) {
```
stevekuznetsov: You are not getting secrets here, you are creating them

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,297 @@
...
+					Labels:    map[string]string{"ci.openshift.org/auto-managed": "true"},
+				},
+			}
+			secretsMap[secretContext.Cluster] = append(secretsMap[secretContext.Cluster], secret)
+		}
+	}
+	return secretsMap, nil
+}
+
+func applySecrets(secretsGetters map[string]coreclientset.SecretsGetter, secretsMap map[string][]*coreapi.Secret) error {
```
stevekuznetsov: You are not applying secrets here, you are Updating them. IMO we should use Replace. Update can fail if someone else touches them but we know we're the only source of truth here.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,297 @@
...
+				if _, err := secretsGetter.Secrets(secret.Namespace).Update(secret); err != nil {
+					return err
+				}
+			}
+		}
+	}
+	return nil
+}
+
+func dryRunDump(secretsGetters map[string]coreclientset.SecretsGetter, w io.Writer) error {
```
stevekuznetsov: In what case do we want to dump the current secrets on the cluster?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,297 @@
...
+	}()
+
+	secretsMap, err := getSecrets(o.config, bwClient)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to get secrets.")
+	}
+
+	if o.dryRun {
+		for cluster := range o.secretsGetters {
+			o.secretsGetters[cluster] = fake.NewSimpleClientset().CoreV1()
```
stevekuznetsov: Oh my. No. Let's not use a test fake here. Just print the things.

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,153 @@
...
+import (
+	"encoding/json"
+	"fmt"
+	"os/exec"
+	"sync"
+
+	"github.com/sirupsen/logrus"
+)
+
+type bitwardenCliClient struct {
```
stevekuznetsov: This type stutters, too. `bitwarden.cliClient` is fine

pkg/bitwarden/bitwarden.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// ItemGetter works on an item from BitWarden
+type ItemGetter interface {
+	GetField(filedName string) (string, error)
+	GetAttachment(attachmentName string) ([]byte, error)
+}
+
+// NewBitwardenClient generates a BitWarden client
+func NewBitwardenClient(username, password string) (Client, error) {
```
stevekuznetsov: NewClient

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,153 @@
...
+func (c *bitwardenCliClient) GetItem(itemName string) (ItemGetter, error) {
+	var itemGetter cliItemGetter
+	for _, i := range c.savedItems {
+		if i.Name == itemName {
+			itemGetter.item = i
+			itemGetter.client = c
+			return &itemGetter, nil
+		}
+	}
+	return nil, fmt.Errorf("failed to find item with name %s", itemName)
```
stevekuznetsov: What do I see if I `GetItem().GetField()` and `GetItem()` returns `nil`?

stevekuznetsov: I imagine a nil pointer panic?

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,153 @@
...
+	return fmt.Errorf("bw-login failed without run's error")
+}
+
+func (c *bitwardenCliClient) GetItem(itemName string) (ItemGetter, error) {
+	var itemGetter cliItemGetter
+	for _, i := range c.savedItems {
+		if i.Name == itemName {
+			itemGetter.item = i
+			itemGetter.client = c
+			return &itemGetter, nil
```
stevekuznetsov: nit: this could just be

```go
return &itemGetter{
    item: i,
    client: c,
}, nil
```

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,153 @@
...
+type bkCmdType string
+
+const (
+	bwLoginCmd  bkCmdType = "login"
+	bwLogoutCmd bkCmdType = "logout"
+	bwListCmd   bkCmdType = "list"
+	bwGetCmd    bkCmdType = "get"
+)
+
+func (c *bitwardenCliClient) runBWCmd(cmdType bkCmdType, args ...string) ([]byte, error) {
```
stevekuznetsov: This still seems off. Why isn't there `bitwardenCliClient.GetAttachment` ? Why is it the `cliItemGetter` that knows how to use the `cliClient` to get an attachment?

Why don't you refactor this so the client (that exposes `GetAttachment` and `GetItem` etc) is in a separate package from the CLI implementation, that should help you separate concerns.

cmd/ci-secret-bootstrap/main.go
hongkailiu: Moved to `completeOptions`.

cmd/ci-secret-bootstrap/main.go
hongkailiu: renamed as discussed..

cmd/ci-secret-bootstrap/main.go
hongkailiu: Renamed.
Keep the update call and see if we hit issues (as discussed).

cmd/ci-secret-bootstrap/main.go
hongkailiu: Removed.

pkg/bitwarden/cli.go
hongkailiu: Done.

pkg/bitwarden/bitwarden.go
hongkailiu: Done.

pkg/bitwarden/cli.go
hongkailiu: Back to 

```
GetFieldOnItem(itemName, fieldName string) (string, error)
GetAttachmentOnItem(itemName, attachmentName string) ([]byte, error)
```

pkg/bitwarden/cli.go
hongkailiu: Function is deleted.

pkg/bitwarden/cli.go
hongkailiu: The function is deleted.

hongkailiu:
\lint

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,285 @@
...
+	Namespace string `json:"namespace"`
+	Name      string `json:"name"`
+}
+
+type secretConfig struct {
+	From map[string]bitWardenContext `json:"from"`
+	To   []secretContext             `json:"to"`
+}
+
+func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[string][]*coreapi.Secret, error) {
```
hongkailiu: @stevekuznetsov 

The reason that the return is `map[string][]*coreapi.Secret` instead of `[]*coreapi.Secret` is that I need a place to save the cluster that this secret belongs to.

Just found an alternative
https://github.com/kubernetes/apimachinery/blob/44bd77c24ef93cd3a5eb6fef64e514025d10d44e/pkg/apis/meta/v1/types.go#L266-L270

I think we could use it. 
What do you think?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,285 @@
...
+			if secretContext.Name == "" {
+				return fmt.Errorf("config[%d].to[%d].name: empty value is not allowed", i, j)
+			}
+
+			if toMap[secretContext.Cluster] == nil {
+				toMap[secretContext.Cluster] = map[string]string{secretContext.Namespace: secretContext.Name}
+			} else if toMap[secretContext.Cluster][secretContext.Namespace] != secretContext.Name {
+				toMap[secretContext.Cluster][secretContext.Namespace] = secretContext.Name
+			} else {
+				return fmt.Errorf("config[%d].to[%d]: secret %v showed more than once in the config", i, j, secretContext)
```
stevekuznetsov: nit: `listed more than once`

cmd/ci-secret-bootstrap/main.go
stevekuznetsov: Map is fine

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,285 @@
...
+	secretsMap := map[string][]*coreapi.Secret{}
+	for _, secretConfig := range config {
+		data := make(map[string][]byte)
+		for key, bwContext := range secretConfig.From {
+			var value []byte
+			var field string
+			var err error
+			if bwContext.Field != "" {
+				field, err = bwClient.GetFieldOnItem(bwContext.BWItem, bwContext.Field)
+				value = []byte(field)
```
stevekuznetsov: If you want a `[]byte` make the `GetFieldOnItem` return it :)

cmd/ci-secret-bootstrap/main.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+	}
+	return secretsMap, nil
+}
+
+func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secretsMap map[string][]*coreapi.Secret) error {
+	for cluster, secrets := range secretsMap {
+		for _, secret := range secrets {
+			secretsGetter := secretsGetters[cluster]
+			_, err := secretsGetter.Secrets(secret.Namespace).Create(secret)
```
stevekuznetsov: nit: 

```go
if _, err := func(); err != nil {

}
```

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,118 @@
...
+type cliClient struct {
+	username string
+	password string
+	sync.Mutex
+	session    string
+	savedItems []Item
+	run        func(cmd string, args ...string) ([]byte, error)
+}
+
+func newCliClient(username, password string) (Client, error) {
```
stevekuznetsov: Why do you need `newCliClient` and `NewClient`?

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,118 @@
...
+func (c *cliClient) loginAndListItems() error {
+	c.Lock()
+	defer c.Unlock()
+	output, err := c.run("bw", "login", c.username, c.password, "--response")
+	if err != nil {
+		return err
+	}
+	r := bwLoginResponse{}
+	if err := json.Unmarshal(output, &r); err != nil {
+		//TODO need to censor if we plan this on a job
```
stevekuznetsov: Should just use the censoring logger straight away

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,118 @@
...
+		return err
+	}
+	if r.Success {
+		raw := r.Data.Raw
+		if raw != "" {
+			c.session = raw
+			if len(c.savedItems) > 0 {
```
stevekuznetsov: When is this the case?

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,118 @@
...
+			return nil
+		}
+		// should never happen
+		return fmt.Errorf("bw-login succeeded with empty '.data.raw'")
+	}
+	// should never happen
+	return fmt.Errorf("bw-login failed without run's error")
+}
+
+func (c *cliClient) Logout() error {
```
stevekuznetsov: Why do we need this?

pkg/bitwarden/cli.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+	return "", fmt.Errorf("failed to find field %s in item %s", fieldName, itemName)
+}
+
+func (c *cliClient) GetAttachmentOnItem(itemName, attachmentName string) ([]byte, error) {
+	for _, item := range c.savedItems {
+		if itemName == item.Name {
+			for _, attachment := range item.Attachments {
+				if attachment.FileName == attachmentName {
+					//bw --session "${BW_SESSION}" get attachment "${attachment_id}" --itemid "${item_id}" --raw
```
stevekuznetsov: Can remove this comment

pkg/bitwarden/cli_test.go
```diff
@@ -0,0 +1,561 @@
...
+		run                func(cmd string, args ...string) ([]byte, error)
+		expectedSession    string
+		expectedSavedItems []Item
+		expectedError      error
+	}{
+		{
+			name:     "basic case",
+			username: "u",
+			password: "p",
+			run: func(cmd string, args ...string) ([]byte, error) {
```
stevekuznetsov: I would much rather see a simple test fake that has a map of args to responses instead of this

pkg/bitwarden/cli_test.go
```diff
@@ -0,0 +1,561 @@
...
+        "url": "https://cdn.bitwarden.net/attachments/111/222"
+      }
+    ],
+    "revisionDate": "2019-04-04T03:43:19.503Z"
+  }
+]`), nil
+				}
+				return nil, fmt.Errorf("unexpected call")
+			},
+			expectedSession: "not-going-to-tell-you==",
```
stevekuznetsov: This test is at the wrong level of abstraction. `NewClient()` can continue to work perfectly while the underlying cache implementation changes. Furthermore, if there is something _other_ than a CLI client, it may be implemented in a different way altogether. This should be a test for the `loginAndListItems()`

pkg/bitwarden/cli_test.go
```diff
@@ -0,0 +1,561 @@
...
+func TestGetFieldOnItem(t *testing.T) {
+	testCases := []struct {
+		name        string
+		client      *cliClient
+		itemName    string
+		fieldName   string
+		expected    string
+		expectedErr error
+	}{
+		{
```
stevekuznetsov: Think about your formatting. This test case is large and spans 50 lines., which makes it hard to read. Try something like this:


```go
		{
			name: "basic case",
			client: &cliClient{savedItems: []Item{{
						ID:   "id1", Name: "unsplash.com",
						Fields: []Field{{Name:  "API Key", Value: "value1"},{Name:  "no name",Value: "value2"}},
				},},
			},
			itemName:  "unsplash.com",
			fieldName: "API Key",
			expected:  "value1",
		},
```

Furthermore, you can just have the list of items in your test case and create the client from that list at runtime, further making this test case simpler.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: done.

cmd/ci-secret-bootstrap/main.go

pkg/bitwarden/cli.go
hongkailiu: `NewClient`: the constructor for the client interface. out-of-pkg caller uses it.
`newCliClient`: the constructor for the cli-implementation. Used by `NewClient`. I guess you want me to use the `cliClient{}` directly. I think hiding loginAndList() call in `newCliClient` looks nicer.

One day, if we switch to e.g., rest-implementation, whose constructor would be `newRestClient` and `NewClient` will call it.

pkg/bitwarden/cli.go
hongkailiu: Removed.

pkg/bitwarden/cli.go
hongkailiu: Removed.

pkg/bitwarden/cli.go
hongkailiu: Removed.

pkg/bitwarden/cli_test.go
hongkailiu: Stole your testing code.

pkg/bitwarden/cli_test.go
hongkailiu: Done.

pkg/bitwarden/cli_test.go
hongkailiu: Fix the client for all test cases.

pkg/bitwarden/cli.go
hongkailiu: Added a commit for protecting `bw-password`.

PR #355, opened 2020-01-08
configresolver: add webreg package for registry help pages 

AlexNPavel:
Add a web UI providing documentation and other useful information to users about the test step registry.


pkg/load/load.go
```diff
@@ -86,10 +86,11 @@ func configFromResolver(info *ResolverInfo) (*api.ReleaseBuildConfiguration, err
 
 // Registry takes the path to a registry config directory and returns the full set of references, chains,
 // and workflows that the registry's Resolver needs to resolve a user's MultiStageTestConfiguration
-func Registry(root string, flat bool) (references map[string]api.LiteralTestStep, chains map[string][]api.TestStep, workflows map[string]api.MultiStageTestConfiguration, err error) {
-	references = map[string]api.LiteralTestStep{}
-	chains = map[string][]api.TestStep{}
-	workflows = map[string]api.MultiStageTestConfiguration{}
+func Registry(root string, flat bool) (references registry.ReferenceMap, chains registry.ChainMap, workflows registry.WorkflowMap, documentation map[string]string, err error) {
```
petr-muller: `references+chains+workflows+documentation` starts to resemble a struct

pkg/webreg/webreg.go
```diff
@@ -0,0 +1,455 @@
...
+	"sort"
+	"strings"
+	"time"
+
+	"github.com/alecthomas/chroma/formatters/html"
+	"github.com/alecthomas/chroma/lexers"
+	"github.com/alecthomas/chroma/styles"
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/load"
+	"github.com/openshift/ci-tools/pkg/registry"
```
petr-muller: nit: separate local imports from third-party

pkg/webreg/webreg.go
```diff
@@ -0,0 +1,455 @@
...
+	name := path.Base(req.URL.Path)
+
+	page := template.Must(template.New("referencePage").Funcs(
+		template.FuncMap{
+			"syntaxedSource": func(source string) template.HTML {
+				formatted, err := syntaxBash(source)
+				if err != nil {
+					logrus.Errorf("Failed to format source file: %v", err)
+				}
+				return template.HTML(formatted)
```
petr-muller: We should not rely on `formatted` on error, should probably put some safe value there.

cmd/ci-operator-configresolver/main.go
```diff
@@ -273,6 +274,7 @@ func main() {
 	http.HandleFunc("/config", handleWithMetrics(resolveConfig(configAgent, registryAgent)))
 	http.HandleFunc("/configGeneration", handleWithMetrics(getConfigGeneration(configAgent)))
 	http.HandleFunc("/registryGeneration", handleWithMetrics(getRegistryGeneration(registryAgent)))
+	http.HandleFunc("/help/", handleWithMetrics(webreg.MainPageHandler(registryAgent, configAgent)))
```
stevekuznetsov: This is going to be out landing page so `/` might be nicer for UX

pkg/load/configAgent.go
```diff
@@ -21,14 +21,15 @@ import (
 // memory and retrieve them when provided with a config.Info.
 type ConfigAgent interface {
 	GetConfig(config.Info) (api.ReleaseBuildConfiguration, error)
+	GetFilenameToConfig() FilenameToConfig
```
stevekuznetsov: nit: naming after types can be fragile, this is really `GetAll()` or `GetAllByFilename()`?

pkg/load/configAgent.go
```diff
@@ -75,27 +76,31 @@ func (a *configAgent) recordError(label string) {
...
 	if !ok {
 		return api.ReleaseBuildConfiguration{}, fmt.Errorf("Could not find config %s", info.Basename())
 	}
 	return config, nil
 }
 
+func (a *configAgent) GetFilenameToConfig() FilenameToConfig {
+	a.lock.RLock()
+	defer a.lock.RUnlock()
+	return a.configs
```
stevekuznetsov: Is this threadsafe or do you need to copy out the map?

pkg/load/load.go
```diff
@@ -86,10 +86,11 @@ func configFromResolver(info *ResolverInfo) (*api.ReleaseBuildConfiguration, err
...
+	references = registry.ReferenceMap{}
```
stevekuznetsov: nit: naming types after implementation specifics can also be fragile -- why not `registry.ReferencesByName`?

pkg/webreg/webreg.go
```diff
@@ -0,0 +1,591 @@
...
+<tbody>
+	{{ range .Post }}
+		{{ tableEntryForStep . }}
+	{{ end }}
+</tbody>
+</table>
+<h2 id="source">TODO: Graph</h2>
+`
+
+type Components struct {
```
stevekuznetsov: I'm surprised we had to declare new types for this -- why couldn't the existing ones be used?

stevekuznetsov: Does that just move formatting/processing to Go code that creates these types instead of the template? I might prefer the latter if the logic is not too complex just to keep the types flat

AlexNPavel: This was mainly done to reduce the number of functions I was using in the templates. This was more useful when I was testing. We could move more stuff into functions used by the templates to avoid new types.

AlexNPavel:
I've redone a lot of the PR to handle more of the work inside the HTML templates instead of go code and removed the new structs that I had before. There are also many style improvements and some bug fixes.

Graphs will be added in a future PR, as this PR is already very large.

PR #356, opened 2020-01-09
Prevent rehearsing hidden jobs

droslean:
/cc @petr-muller

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -166,6 +166,11 @@ func filterPresubmits(changedPresubmits map[string][]prowconfig.Presubmit, logge
 		for _, job := range jobs {
 			jobLogger := logger.WithFields(logrus.Fields{"repo": repo, "job": job.Name})
 
+			if job.Hidden {
+				jobLogger.Warn("hidden jobs are not allowed to be rehearsed")
+				continue
+			}
+
 			if !hasRehearsableLabel(job.Labels) {
 				jobLogger.Warnf("job is not allowed to be rehearsed. Label %s is required", jobconfig.CanBeRehearsedLabel)
 				continue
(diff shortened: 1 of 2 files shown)


pkg/diffs/diffs.go
```diff
@@ -105,7 +106,7 @@ func GetChangedPresubmits(prowMasterConfig, prowPRConfig *prowconfig.Config, log
 			masterJob := masterJobs[repo][job.Name]
 			logFields := logrus.Fields{logRepo: repo, logJobName: job.Name}
 
-			if job.Agent == string(pjapi.KubernetesAgent) {
+			if job.Agent == string(pjapi.KubernetesAgent) && !job.Hidden {
```
petr-muller: I don't think the check belongs here. We have the two methods (`filterPresubmits` and `filterPeriodics`) whose job is to remove the presubmits that should be rehearsed but for some reason can't be, that seems like a much better place.

pkg/diffs/diffs.go
```diff
@@ -290,7 +291,7 @@ func GetChangedPeriodics(prowMasterConfig, prowPRConfig *prowconfig.Config, logg
 	masterPeriodicsPerName := getPeriodicsPerName(prowMasterConfig.JobConfig.AllPeriodics())
 
 	for name, job := range getPeriodicsPerName(prowPRConfig.JobConfig.AllPeriodics()) {
-		if job.Agent == string(pjapi.KubernetesAgent) {
+		if job.Agent == string(pjapi.KubernetesAgent) && !job.Hidden {
```
petr-muller: (same as in `GetChangedPresubmits`)

pkg/rehearse/jobs.go
```diff
@@ -605,6 +605,10 @@ func pickTemplateJob(presubmits map[string][]prowconfig.Presubmit, templateFile,
 				continue
 			}
 
+			if job.Hidden {
+				continue
+			}
```
petr-muller: For clarity, I'll just comment that (as opposed to the previous comments for `GetChanged*` methods) this hunk should stay - it's in our interest to pick up a job that won't be later filtered out by the `filter*` methods.

This also made me spot a probable bug, btw: `pickTemplateJob` will probably happily pick a job without an `rehearsable` label, which will later be filtered out because of that. Looks like we should extract that whole logic (in `filter` methods ) that decide whether a job is reheasable or not and also use it here to check that we won't pick an unrehearsable job, so that we avoid duplicating that code.

droslean: Nice catch. It seems that I missed updating this part when I added the rehearse label.

PR #357, opened 2020-01-09
Add a tool for syncing git content between mirrors

petr-muller:
This tool automatically syncs code from which "official" images are
built from their public git repo locations to their mirrors in private
GitHub organization. It is intended to be run in a Prow periodic job
with an interval of less than an hour.

Both source and destination are queried with `git ls-remote`: if they are
already in sync, no further git operations are done. When the destination is
detected to be an empty repo, full `git fetch` is done against a source,
and FETCH_HEAD is then pushed to the destination. When the destination
is a non-empty repository, `git fetch --depth X` is done against the
source and the resulting shallow branch tip is attempted to be pushed.
When this fails, X is exponentially increased and the fetch is retried.
When a threshold is hit, the shallow branch tip is fully fetched with
`--unshallow` and the push is retried again.

Currently the tool does not fail when the destination repository does not exist
at all: this should allow running against full openshift/release while
repository mirrors are created in the private org.

### DIFF
--- a/cmd/private-org-sync/main.go
+++ b/cmd/private-org-sync/main.go
@@ -1,0 +1,472 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"math"
+	"net/url"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/errorutil"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/logrusutil"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+type options struct {
+	configDir string
+	tokenPath string
+	targetOrg string
+
+	org  string
+	repo string
+
+	gitDir  string
+	confirm bool
+}
+
+func (o *options) validate() []error {
+	var errs []error
+	if o.configDir == "" {
+		errs = append(errs, fmt.Errorf("--config-dir is required"))
+	}
+
+	if o.targetOrg == "" {
+		errs = append(errs, fmt.Errorf("--target-org is required"))
+	}
+	if o.org != "" && o.targetOrg == o.org {
+		errs = append(errs, fmt.Errorf("--only-org cannot be equal to --target-org"))
+	}
+
+	if o.repo != "" {
+		if o.org != "" {
+			errs = append(errs, fmt.Errorf("--only-org cannot be used together with --only-repo"))
+		}
+		items := strings.Split(o.repo, "/")
+		if len(items) != 2 {
+			errs = append(errs, fmt.Errorf("--only-repo must have org/repo format"))
+		}
+		if items[0] == o.targetOrg {
+			errs = append(errs, fmt.Errorf("repo passed in --repo-only must have org different from --target-org"))
+		}
+	}
+
+	if o.tokenPath == "" {
+		errs = append(errs, fmt.Errorf("--token-path is required"))
+	}
+	return errs
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.tokenPath, "token-path", "", "Path to token to use when pushing to GitHub.")
+	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
+	fs.StringVar(&o.targetOrg, "target-org", "", "Name of the org holding repos into which the git content should be mirrored")
+
+	fs.StringVar(&o.org, "only-org", "", "Mirror only repos that belong to this org")
+	fs.StringVar(&o.repo, "only-repo", "", "Mirror only a single repo")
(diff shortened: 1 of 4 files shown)


petr-muller:
I'm aware this is probably quite hard to review: if needed, I think I can still reap some optimizations that make the operation faster, and make it do simple fetch-push for now, and submit the optimizations as followups. Let me know if this is desired.

petr-muller:
> However, it will makes more sense to run this in a postsubmit instead of periodic.

@droslean: Well spotted.

Syncing the content in the postsubmit was one of the options we (well, mostly I) considered to implement the whole "things should sync to another org" story. It has some merits: that way you only get triggered when there is something new to sync, you know exactly what to sync and Prow would fetch you the git content, so the only thing remaining would be to find a destination and push. The problem is that it moves complexity somewhere else: you would need to implement something that makes sure every repo, every "official" branch has such a postsubmit (this is not that hard with Prowgen). Such system is also less resilient against failures: when a postsubmit would fail to sync for any reason, you either rely on another postsubmit fixing the situation, or you need to monitor individual failures (even flakes!) and fix them. With a periodic, there is a single place that syncs everything, and if something fails, next iteration can reconcile everything into a desired state, so you only need to fix persistent failures. Also we would have many new jobs (lot of boilerplate YAML).

So all this made me to follow to go the way of having a tool that run and reconciles everything to the desired state at once, and run that tool periodically. We do the same concept for other tasks (like branch fast-forwarding).

cmd/private-org-sync/main.go
```diff
@@ -0,0 +1,469 @@
...
+	logger := g.logger.WithFields(mirrorFields)
+	logger.Info("Syncing content between locations")
+
+	logger.Debug("Initializing git repository")
+	if _, exitCode, err := g.git(logger, repoDir, "init", "--bare"); err != nil || exitCode != 0 {
+		logger.WithField("exit-code", exitCode).WithError(err).Error("Failed to initialize local git directory")
+		return fmt.Errorf("failed to initialize local git directory")
+	}
+
+	srcRemote := fmt.Sprintf("%s-%s", src.org, src.repo)
```
stevekuznetsov: Why do you need to configure the `git` to have remotes? `git fetch`, `git ls-remote` and `git push` all can just take the remote as an arg

petr-muller: Good question, looks like I misplaced the comment with the explanation (I swear I wrote it but it got lost in the revisions). Without the remote, the local repo does not serve as a cache:

```console
$ git init --bare
Initialized empty Git repository in /home/afri/Temporary/bare-origin/
$ time git fetch https://github.com/openshift/origin.git --depth 2 master
remote: Enumerating objects: 67292, done.
remote: Counting objects: 100% (67292/67292), done.
remote: Compressing objects: 100% (44239/44239), done.
remote: Total 67292 (delta 22115), reused 44537 (delta 17109), pack-reused 0
Receiving objects: 100% (67292/67292), 114.47 MiB | 10.84 MiB/s, done.
Resolving deltas: 100% (22115/22115), done.
From https://github.com/openshift/origin
 * branch                master     -> FETCH_HEAD
7.18user 1.32system 0:25.63elapsed 33%CPU (0avgtext+0avgdata 88496maxresident)k
0inputs+238144outputs (0major+30789minor)pagefaults 0swaps
$ time git fetch https://github.com/openshift/origin.git --depth 2 master
remote: Enumerating objects: 67292, done.
remote: Counting objects: 100% (67292/67292), done.
remote: Compressing objects: 100% (44239/44239), done.
remote: Total 67292 (delta 22115), reused 44537 (delta 17109), pack-reused 0
Receiving objects: 100% (67292/67292), 114.47 MiB | 10.80 MiB/s, done.
Resolving deltas: 100% (22115/22115), done.
From https://github.com/openshift/origin
 * branch                master     -> FETCH_HEAD
8.88user 1.33system 0:26.67elapsed 38%CPU (0avgtext+0avgdata 332716maxresident)k
0inputs+238144outputs (0major+81533minor)pagefaults 0swaps
$ rm -rf *; git init --bare
Initialized empty Git repository in /home/afri/Temporary/bare-origin/
$ git remote add oo https://github.com/openshift/origin.git
$ time git fetch oo --depth 2 master
remote: Enumerating objects: 67292, done.
remote: Counting objects: 100% (67292/67292), done.
remote: Compressing objects: 100% (44239/44239), done.
remote: Total 67292 (delta 22115), reused 44537 (delta 17109), pack-reused 0
Receiving objects: 100% (67292/67292), 114.47 MiB | 10.67 MiB/s, done.
Resolving deltas: 100% (22115/22115), done.
From https://github.com/openshift/origin
 * branch                master     -> FETCH_HEAD
 * [new branch]          master     -> oo/master
7.20user 1.45system 0:27.39elapsed 31%CPU (0avgtext+0avgdata 88852maxresident)k
16inputs+238152outputs (0major+29544minor)pagefaults 0swaps
$ time git fetch oo --depth 2 master
remote: Total 0 (delta 0), reused 0 (delta 0), pack-reused 0
From https://github.com/openshift/origin
 * branch                master     -> FETCH_HEAD
0.13user 0.04system 0:01.88elapsed 9%CPU (0avgtext+0avgdata 88112maxresident)k
0inputs+16outputs (0major+9670minor)pagefaults 0swaps
```

But you are right, this definitely needs explanation.

petr-muller:
>Will this work if we add a PV to the job to hold git state between runs?

@stevekuznetsov yes it should, that's what the `--git-dir` is for, although I do not expect the performance gain from that to be that great.

cmd/private-org-sync/main.go
petr-muller: Added an explanatory comment to the code.

PR #359, opened 2020-01-10
fix retry around template instance ready check (in case it is called …

gabemontero:
…before the templates instance is created)

Please see https://coreos.slack.com/archives/CBN38N3MW/p1578667979391000 for my analysis of a race condition on the template instance ready check in ci-tools.

A quick synopsis of my analysis of https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_cluster-samples-operator/211/pull-ci-openshift-cluster-samples-operator-master-e2e-aws-upgrade/335

(though I've had a lot of other similar failures over the last few days)

- I see that the e2e-aws-upgrade template instance does exist in my PR's namespace, where it was created at creationTimestamp: "2020-01-10T14:47:14Z"
- In the prow log, it starts waiting for the instance at 2020/01/10 14:47:09 Waiting for template instance to be ready ... so yeah a few seconds before it was created
- the existing for loop actual punts if it cannot find the templateinstance on the first pass

@openshift/openshift-team-developer-productivity-platform PTAL

or if that is not the right team, @petr-muller @bbguimaraes @droslean @hongkailiu @stevekuznetsov PTAL

thanks

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -382,21 +382,28 @@ func isPodCompleted(podClient coreclientset.PodInterface, name string) (bool, er
 }
 
 func waitForTemplateInstanceReady(templateClient templateclientset.TemplateInstanceInterface, instance *templateapi.TemplateInstance) (*templateapi.TemplateInstance, error) {
-	for {
-		ready, err := templateInstanceReady(instance)
-		if err != nil {
-			return nil, fmt.Errorf("could not determine if template instance was ready: %v", err)
+	var actualErr error
+	err := wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
+		ready, actualErr := templateInstanceReady(instance)
+		if actualErr != nil {
+			log.Printf("could not determine if template instance was ready: %v", actualErr)
+			return false, nil
 		}
 		if ready {
-			return instance, nil
+			return true, nil
 		}
 
-		time.Sleep(2 * time.Second)
-		instance, err = templateClient.Get(instance.Name, meta.GetOptions{})
-		if err != nil {
-			return nil, fmt.Errorf("unable to retrieve existing template instance: %v", err)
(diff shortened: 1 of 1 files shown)


gabemontero:
@openshift/openshift-team-developer-experience FYI as well ... a fix for the recurring CI issue I mentioned in scrum today

gabemontero:
> Thanks, Gabe

np likewise thanks for the quick turnaround Steve

pkg/steps/template.go
```diff
@@ -382,21 +382,28 @@ func isPodCompleted(podClient coreclientset.PodInterface, name string) (bool, er
...
-		if err != nil {
-			return nil, fmt.Errorf("unable to retrieve existing template instance: %v", err)
+		instance, actualErr = templateClient.Get(instance.Name, meta.GetOptions{})
+		if actualErr != nil {
+			log.Printf("unable to retrieve existing template instance: %v", actualErr)
+			return false, nil
 		}
+		return true, nil
+	})
+	if err == nil {
```
stevekuznetsov: nit: you could drop this check and just 


```go
return instance, actualErr
``` 

The caller has to check `err == nil` on their end anyway

pkg/steps/template.go
```diff
@@ -382,21 +382,28 @@ func isPodCompleted(podClient coreclientset.PodInterface, name string) (bool, er
 }
 
 func waitForTemplateInstanceReady(templateClient templateclientset.TemplateInstanceInterface, instance *templateapi.TemplateInstance) (*templateapi.TemplateInstance, error) {
-	for {
-		ready, err := templateInstanceReady(instance)
-		if err != nil {
-			return nil, fmt.Errorf("could not determine if template instance was ready: %v", err)
+	var actualErr error
+	err := wait.PollImmediate(2*time.Second, 1*time.Minute, func() (bool, error) {
```
stevekuznetsov: We would loop forever in the past, right? Might want to set this timeout for ten minutes

PR #363, opened 2020-01-10
fix retry around template instance ready check (in case it is called …

gabemontero:
…before the templates instance is created)

another attempt at the issue originally cited in https://github.com/openshift/ci-tools/pull/359

/assign @stevekuznetsov

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -382,21 +382,23 @@ func isPodCompleted(podClient coreclientset.PodInterface, name string) (bool, er
 }
 
 func waitForTemplateInstanceReady(templateClient templateclientset.TemplateInstanceInterface, instance *templateapi.TemplateInstance) (*templateapi.TemplateInstance, error) {
-	for {
-		ready, err := templateInstanceReady(instance)
-		if err != nil {
-			return nil, fmt.Errorf("could not determine if template instance was ready: %v", err)
+	var actualErr error
+	err := wait.PollImmediate(2*time.Second, 10*time.Minute, func() (bool, error) {
+		instance, actualErr = templateClient.Get(instance.Name, meta.GetOptions{})
+		if actualErr != nil {
+			return false, nil
 		}
+		ready := false
+		ready, actualErr = templateInstanceReady(instance)
 		if ready {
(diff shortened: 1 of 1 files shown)


pkg/steps/template.go
```diff
@@ -382,21 +382,23 @@ func isPodCompleted(podClient coreclientset.PodInterface, name string) (bool, er
...
 func waitForTemplateInstanceReady(templateClient templateclientset.TemplateInstanceInterface, instance *templateapi.TemplateInstance) (*templateapi.TemplateInstance, error) {
-	for {
-		ready, err := templateInstanceReady(instance)
-		if err != nil {
-			return nil, fmt.Errorf("could not determine if template instance was ready: %v", err)
+	var actualErr error
+	err := wait.PollImmediate(2*time.Second, 10*time.Minute, func() (bool, error) {
+		instance, actualErr = templateClient.Get(instance.Name, meta.GetOptions{})
+		if actualErr != nil {
+			return false, nil
```
stevekuznetsov: We never fall out on perma-failing on this get anymore, except for the timeout. That's probably fine?

pkg/steps/template.go
```diff
@@ -382,21 +382,23 @@ func isPodCompleted(podClient coreclientset.PodInterface, name string) (bool, er
...
-		if err != nil {
-			return nil, fmt.Errorf("unable to retrieve existing template instance: %v", err)
+			return true, nil
 		}
+		return false, nil
+	})
+	if err == nil {
+		return instance, nil
 	}
+	return nil, actualErr
```
gabemontero: it seemed better to no return an instance ptr where it existed but was not ready ... the calling code does not care, but I was worried about possible future use

pkg/steps/template.go
gabemontero: I thought it was and in the spirit of retrying as long as possible

gabemontero: and before you were looping forever if the ready check did not return an error but ready was always false

gabemontero:
per slack discussion with @stevekuznetsov we've agreed to cancel the hold

/hold cancel

PR #364, opened 2020-01-11
steps: Update cluster upgrade template to latest

smarterclayton:
Picks up changes to azure placement and exiting the test container
if setup fails with exit code 0 instead of 1 (so we don't double
report errors).


smarterclayton:
Synced to master, tagging

PR #365, opened 2020-01-11
testgrid: Gather more testgrid results for jobs that run less frequently

smarterclayton:
Many periodics are on 24h or 48h intervals, so test grid's default of 7
days doesn't show much info. For those jobs, use a heuristic to try and get
100 test results based on the configured interval of the periodic, between
7d (the default) and 60d (a reasonable max).

Release controller jobs are not changed because they are run at intervals
outside of our control.

Also:

testgrid: Exclude the test container summary junit from testgrid

The test container junit (generated by ci-operator) is already summarized
by the overall test result (if we fail test, we fail the overall run), but
the summary page for test grid will count it as part of the flake rate and
failure percentage. A downside of this change is that if the test container
breaks test grid won't show a specific red box (but prow would), and also
that we can't see the time the full test suite takes to run in the table
graph, which is an acceptable tradeoff for making the test summary clearer.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -6,12 +6,14 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"math"
 	"os"
 	"path"
 	"path/filepath"
 	"regexp"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/GoogleCloudPlatform/testgrid/pb/config"
 	jc "github.com/openshift/ci-tools/pkg/jobconfig"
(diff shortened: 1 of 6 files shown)


smarterclayton:
/assign @stevekuznetsov 

I haven’t pushed the test-infra PR yet, a good number of periodically are assigned 50 or 33 days of retention (12h or 8h)

smarterclayton:
Will regen against head later and link

smarterclayton:
https://github.com/kubernetes/test-infra/pull/15894 has the generation, please tag both.

smarterclayton:
Dashboards merged, merging this.

PR #369, opened 2020-01-14
private-org-sync: fail-fast when destination does not exist

petr-muller:
This saves hundreds of useless git executions in the current state when `openshift-priv` is empty.


petr-muller:
oops, tests

/meow

petr-muller:
/hold 

need to fix tests

PR #373, opened 2020-01-15
appplyconfig: use --validate only with --dry-run

stevekuznetsov:


petr-muller:
/hold

I believe we still need to weed out the failures in  existing config: https://github.com/openshift/ci-tools/pull/372#issue-363139996

petr-muller:
We agreed `--validate` does not bring enough value to compensate for the problems we have with it.

/close

PR #376, opened 2020-01-16
pkg/registry/graph: fix panics for invalid regsitry components

AlexNPavel:
Fix panics for registry components that include a chain or reference that does not actually exist in the registry. This was caught by the graph, but through a panic. Now the graph detects the problem and returns a proper error message.

### DIFF
--- a/pkg/registry/graph.go
+++ b/pkg/registry/graph.go
@@ -264,6 +264,9 @@ func NewGraph(stepsByName ReferenceMap, chainsByName ChainMap, workflowsByName W
 		nodesByName[name] = node
 		for _, step := range chain {
 			if step.Reference != nil {
+				if _, exists := referenceNodes[*step.Reference]; !exists {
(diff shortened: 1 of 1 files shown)


pkg/registry/graph.go
```diff
@@ -264,6 +264,9 @@ func NewGraph(stepsByName ReferenceMap, chainsByName ChainMap, workflowsByName W
 		nodesByName[name] = node
 		for _, step := range chain {
 			if step.Reference != nil {
+				if referenceNodes[*step.Reference] == nil {
```
stevekuznetsov: nit: idiomatic Go for map key existence is `if _, exists := myMap[key]; !exists {`

AlexNPavel: Fixed

stevekuznetsov:
```
$ git grep TestNewGraph | wc -l
0
```

PR #378, opened 2020-01-17
Improve the secret-bootstrap tool

hongkailiu:
* Load kubeconfig from default e.g., $HOME/.kube/config if
`--kubeconfig` is not set.
* Log standard err when bw cmd failed
* Add logout to avoid "you have already logged in" error.

### DIFF
--- a/pkg/bitwarden/cli.go
+++ b/pkg/bitwarden/cli.go
@@ -19,9 +19,14 @@ type cliClient struct {
 }
 
 func newCliClient(username, password string) (Client, error) {
-	return newCliClientWithRun(username, password, func(args ...string) (bytes []byte, err error) {
+	return newCliClientWithRun(username, password, func(args ...string) ([]byte, error) {
+		// bw-password is protected, session in args is not
 		logrus.WithField("args", args).Info("running bw command ...")
-		return exec.Command("bw", args...).Output()
+		out, err := exec.Command("bw", args...).CombinedOutput()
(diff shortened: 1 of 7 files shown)


pkg/util/client_test.go
```diff
@@ -56,11 +56,6 @@ func TestLoadKubeConfigs(t *testing.T) {
 		expectedContext string
 		expectedError   error
 	}{
-		{
-			name: "empty string",
```
hongkailiu: It would fail if `$HOME/.kube/config` exists.

pkg/bitwarden/cli.go
```diff
@@ -19,9 +20,19 @@ type cliClient struct {
 }
 
 func newCliClient(username, password string) (Client, error) {
-	return newCliClientWithRun(username, password, func(args ...string) (bytes []byte, err error) {
+	return newCliClientWithRun(username, password, func(args ...string) ([]byte, error) {
+		// bw-password is protected, session in args is not
 		logrus.WithField("args", args).Info("running bw command ...")
-		return exec.Command("bw", args...).Output()
```
stevekuznetsov: Couldn't this have been

```go
out, err := exec.Command("bw", args...).CombinedOutput()
if err != nil {
    logrus.WithError(err).Errorf("bw cmd failed: %v", out)
}
return out, errr
```

hongkailiu:
/hold
testing locally

PR #382, opened 2020-01-17
Detect diff. between BW items and secrets

hongkailiu:
https://issues.redhat.com/browse/DPTP-645

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -10,9 +10,11 @@ import (
 	"strings"
 
 	"github.com/ghodss/yaml"
+	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/serializer/json"
@@ -31,6 +33,7 @@ type options struct {
 	configPath     string
 	bwUser         string
 	bwPasswordPath string
+	force          bool
 
 	bwPassword     string
 	secretsGetters map[string]coreclientset.SecretsGetter
@@ -45,6 +48,7 @@ func parseOptions() options {
 	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
 	fs.StringVar(&o.bwUser, "bw-user", "", "Username to access BitWarden.")
 	fs.StringVar(&o.bwPasswordPath, "bw-password-path", "", "Path to a password file to access BitWarden.")
+	fs.BoolVar(&o.force, "force", false, "If true, update the secrets even if existing one differs from Bitwarden items instead of existing with error. Default false.")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Errorf("cannot parse args: %q", os.Args[1:])
 	}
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -212,19 +215,26 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
...
+				if !kerrors.IsNotFound(err) {
 					return err
 				}
-				if _, err := secretsGetter.Secrets(secret.Namespace).Update(secret); err != nil {
+				if _, err := secretsGetter.Secrets(secret.Namespace).Create(secret); err != nil {
 					return err
 				}
 			}
+			if !force && !equality.Semantic.DeepEqual(existingSecret.Data, secret.Data) {
+				return fmt.Errorf("found unsynchronized secret: %s:%s/%s", cluster, secret.Namespace, secret.Name)
```
stevekuznetsov: You should paste the diff here

cmd/ci-secret-bootstrap/main.go
```diff
@@ -212,19 +215,26 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
...
-func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secretsMap map[string][]*coreapi.Secret) error {
+func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secretsMap map[string][]*coreapi.Secret, force bool) error {
 	for cluster, secrets := range secretsMap {
 		for _, secret := range secrets {
 			logrus.Infof("handling secret: %s:%s/%s", cluster, secret.Namespace, secret.Name)
 			secretsGetter := secretsGetters[cluster]
-			if _, err := secretsGetter.Secrets(secret.Namespace).Create(secret); err != nil {
-				if !kerrors.IsAlreadyExists(err) {
+			existingSecret, err := secretsGetter.Secrets(secret.Namespace).Get(secret.Name, meta.GetOptions{})
+			if err != nil {
```
stevekuznetsov: Couldn't this be simplified a bit

```go
if kerrors.IsNotFound(err) {
    if _, err := secretsGetter.Secrets(secret.Namespace).Create(secret); err != nil {
        return err
    }
} else if err != nil {
    return err
}
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -212,24 +216,40 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
...
 				}
+			} else if err != nil {
+				return err
+			}
+			if !force && !equality.Semantic.DeepEqual(secret.Data, existingSecret.Data) {
+				logrus.Errorf("actual %s:%s/%s differs the expected:\n%s", cluster, secret.Namespace, secret.Name,
+					cmp.Diff(bytesMapToStringMap(secret.Data), bytesMapToStringMap(existingSecret.Data)))
+				return fmt.Errorf("found unsynchronized secret: %s:%s/%s", cluster, secret.Namespace, secret.Name)
+			}
+			if _, err := secretsGetter.Secrets(secret.Namespace).Update(secret); err != nil {
```
droslean: If we hit `kerrors.IsNotFound(err)` you create the ticket above and there is no reason to `Update` it again.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -212,24 +216,40 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
...
-				if _, err := secretsGetter.Secrets(secret.Namespace).Update(secret); err != nil {
+			existingSecret, err := secretsGetter.Secrets(secret.Namespace).Get(secret.Name, meta.GetOptions{})
+			if kerrors.IsNotFound(err) {
+				if _, err := secretsGetter.Secrets(secret.Namespace).Create(secret); err != nil {
 					return err
 				}
+			} else if err != nil {
+				return err
+			}
+			if !force && !equality.Semantic.DeepEqual(secret.Data, existingSecret.Data) {
```
droslean: Lets do that right after we Get the secret.
 And we want to update only if that passes right?

It seems that you need something like:

```go
			if existingSecret, err := secretsGetter.Secrets(secret.Namespace).Get(secret.Name, meta.GetOptions{}); err == nil {
				if !force && !equality.Semantic.DeepEqual(secret.Data, existingSecret.Data) {
					// logrus....
					return fmt.Errorf("found unsynchronized secret: %s:%s/%s", cluster, secret.Namespace, secret.Name)
				}

				if _, err := secretsGetter.Secrets(secret.Namespace).Update(secret); err != nil {
					return err
				}

			} else if kerrors.IsNotFound(err) {
				if _, err := secretsGetter.Secrets(secret.Namespace).Create(secret); err != nil {
					return err
				}
			} else
				return err
			}
```

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: Updated with more fix from Nikos.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
hongkailiu: Good catch. Updated.

PR #391, opened 2020-01-20
Timeout after 30m of destroying a cluster

hongkailiu:
Follow up https://coreos.slack.com/archives/CBN38N3MW/p1579539221077700

### DIFF
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -59,7 +59,7 @@ EOF
 done
 
 for workdir in $( find /tmp/deprovision -mindepth 1 -type d | shuf ); do
-  openshift-install --dir "${workdir}" --log-level debug destroy cluster
+  timeout 30m openshift-install --dir "${workdir}" --log-level debug destroy cluster
 done
 
 echo "Deprovision finished successfully"


cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -59,7 +59,7 @@ EOF
 done
 
 for workdir in $( find /tmp/deprovision -mindepth 1 -type d | shuf ); do
-  openshift-install --dir "${workdir}" --log-level debug destroy cluster
+  timeout 30m openshift-install --dir "${workdir}" --log-level debug destroy cluster
```
hongkailiu: This would exit the script.

Should we ignore the error and continue with the `for-loop`?

jstuever: I would say so, yes. The goal is to attempt to delete all clusters even if one fails. The `other` pruner should catch those that are not deleting here.

petr-muller: @hongkailiu I think it's fine to abort with a failure, it's essentially the same like `openshift-install` exiting unsuccessfully, and we run the tool often enough so that the shuffler makes it clean the remaining clusters eventually.

hongkailiu: ack. Updated.

hongkailiu: Removed again.

We want to see the alerts of failing jobs.

PR #392, opened 2020-01-20
Revertd92df0facdccf541f126d679750e214bd33c7c9d

hongkailiu:
Follow up https://coreos.slack.com/archives/CEKNRGF25/p1579544330056200

Some check passed without running the test pod from the template.

/assign @petr-muller @gabemontero


gabemontero:
yep that was the old form .... I also approve

reader's digest from the slack thread referenced ... @petr-muller said he would tomorrow look at investigating the readiness stuff we found today and re-introducing the fix

I'll report if I start seeing the problem I saw again if need be in the interm

petr-muller:
/lgtm

Yeah tomorrow I'll try to re-include this with a fix to the problem we saw and discussed in the Slack thread above

PR #393, opened 2020-01-21
ci-operator: fixes in template-handling logic

petr-muller:
Another shot at the issue described in https://github.com/openshift/ci-tools/pull/359, essentially a re-revert of https://github.com/openshift/ci-tools/pull/392 with additional fixes.

Previously, it was possible for `waitForTemplateInstanceReady` to exit
with `instance, nil` (no error) even when `templateInstanceReady`
returned an error. This then likely caused the cases where ci-operator
treat this operation as success and never run any tests (it continued to
wait until no containers were running, which was true for the template
that failed to instantiate).

Fix this so that the error from `waitForTemplateInstanceReady` is
propagated to the polling method return code. This way the polling ends
only either on timeout (returning an error), or when
`templateInstanceReady` returns ready=true. In that case the error is
still propagated outside and we return it. We never exit the polling on
the error when `Get`-ing the template, and its errors are never
propagated out. If `Get` never succeeds, we end on timeout and the error
path is triggered anyway.

The code allowed some streamlining, I think it is now more clear about
what it is supposed to do.

I considered fixing the issue by making `templateInstanceReady` return
`false` on errors, meaning the polling should continue, but I think
that semantics would not be correct. A template that failed to
instantiate is an error condition and there is no point in waiting for
it further.

/cc @gabemontero @stevekuznetsov @bbguimaraes

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -138,7 +138,7 @@ func (s *templateExecutionStep) Run(ctx context.Context, dry bool) error {
 	}
 
 	log.Printf("Waiting for template instance to be ready")
-	instance, err = waitForTemplateInstanceReady(s.templateClient.TemplateInstances(s.jobSpec.Namespace), instance)
+	instance, err = waitForTemplateInstanceReady(s.templateClient.TemplateInstances(s.jobSpec.Namespace), s.template.Name)
 	if err != nil {
 		return fmt.Errorf("could not wait for template instance to be ready: %v", err)
 	}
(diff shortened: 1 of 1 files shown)


petr-muller:
Yeah, if you're not in a great hurry to get this back in, I'll happily invest some effort into some unit tests.

gabemontero:
> Yeah, if you're not in a great hurry to get this back in, I'll happily invest some effort into some unit tests.

There are enough other things blocking me anyway.  Let's give this some more time and see if you get the cycles to add the unit tests (wouldn't bee too bad I would think ... even you you have to focus the new test just on the waitForTemplateInstance method).

Based on how things progress we can reconvene in a few days if need be.  Otherwise we'll look at your unit tests in the next couple of days if you get the cycles.

PR #395, opened 2020-01-22
ipi-deprovision: calculate GCE TTL from creation time

stevekuznetsov:


cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -32,11 +31,15 @@ EOF
   done
 done
 
+echo "Searching for clusters with a TTL of ${CLUSTER_TTL}"
```
jstuever: this is redundant.

PR #400, opened 2020-01-23
pkg/steps/clusterinstall/template: Run must-gather before tests too

wking:
This will help @ironcladlou debug [rhbz#1785457][1], and might be generally useful for other folks looking into update failures.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1785457

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -334,6 +334,10 @@ objects:
           setup-google-cloud-sdk
         fi
 
+        echo "Running must-gather..."
+        mkdir -p /tmp/artifacts/pre-test-must-gather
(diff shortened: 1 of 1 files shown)


pkg/steps/clusterinstall/template.go
```diff
@@ -334,6 +334,10 @@ objects:
           setup-google-cloud-sdk
         fi
 
+        echo "Running must-gather..."
+        mkdir -p /tmp/artifacts/pre-test-must-gather
+        oc --insecure-skip-tls-verify adm must-gather --dest-dir /tmp/artifacts/pre-test-must-gather >/tmp/artifacts/pre-test-must-gather/must-gather.log
```
wking: not sure we actually need `insecure-skip-tls-verify`.  I'd expect the kubeconfig to include the custom cert.  But whatever, seems orthogonal to this PR, so I'm just rolling forward with the same `must-gather` call we use during `teardown`.

wking:
#402 is carrying and extending this with some iptables gathering.

PR #402, opened 2020-01-24
Add a pre-upgrade save for iptables and must-gather

abhat:


ironcladlou:
Supercedes https://github.com/openshift/ci-tools/pull/402?

PR #403, opened 2020-01-24
ci-operator multi-stage: create service account

bbguimaraes:
This addresses two problems: allowing the `secret-wrapper` to update its
Secret and allowing tests to manipulate a limited set of objects in the
test namespace.

Template tests are "allowed" to create any kind of resource because they
are evaluated by the ci-operator ServiceAccount, an admin in the test
namespace.  Multi-stage tests currently use the same account as other
tests (`default`), so their set of allowed operations is much more
limited.

This commit creates:

- a ServiceAccount specific to each test
- a Role that can
  - list/create RoleBindings
  - get/list/create the test-specific Secret
  - get imagestream/layers
- a RoleBinding that assigns the Role above and the
  `system:image-puller` ClusterRole to the test's ServiceAccount

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -8,11 +8,13 @@ import (
 	"strings"
 
 	coreapi "k8s.io/api/core/v1"
+	rbacapi "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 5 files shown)


pkg/steps/multi_stage.go
```diff
@@ -158,6 +171,54 @@ func (s *multiStageTestStep) Provides() (api.ParameterMap, api.StepLink) {
 }
 func (s *multiStageTestStep) SubTests() []*junit.TestCase { return s.subTests }
 
+func (s *multiStageTestStep) setupRBAC() error {
+	labels := map[string]string{multiStageTestLabel: s.name}
+	m := meta.ObjectMeta{Name: s.name, Labels: labels}
```
bbguimaraes: There is the potential for conflict here: the test cannot be named `default`, `ci-operator`, `default`, or `deployer`. These are all `create` calls and the test will fail with an obvious message, so I don't think it is a significant problem.

PR #404, opened 2020-01-24
Fix the bugs of ci-secret-bootstrap found in test

hongkailiu:
* bw-get-attachment saves the attachment in a file, the output
is the file path. We need to read that file. In the fix, we specify
the path with `--output` of bw-cmd.
* Consider the default value of secret type to fix [1].

1. error log

```
time="2020-01-24T19:42:41Z" level=fatal msg="Failed to update secrets." error="cannot change secret type from \"Opaque\" to \"\" (immutable file): default:ci/sentry-dsn"
```

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -233,6 +233,9 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
 			data[key] = value
 		}
 		for _, secretContext := range secretConfig.To {
+			if secretContext.Type == "" {
+				secretContext.Type = coreapi.SecretTypeOpaque
+			}
 			secret := &coreapi.Secret{
 				Data: data,
 				ObjectMeta: meta.ObjectMeta{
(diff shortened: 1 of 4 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -233,14 +233,18 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
 			data[key] = value
 		}
 		for _, secretContext := range secretConfig.To {
+			t := secretContext.Type
+			if t == "" {
+				t = coreapi.SecretTypeOpaque
```
droslean: Why not changing the `secretContext.Type` directly? Is it going to be used in somewhere else?

pkg/bitwarden/cli.go
```diff
@@ -93,12 +95,29 @@ func (c *cliClient) GetFieldOnItem(itemName, fieldName string) ([]byte, error) {
...
 
-func (c *cliClient) GetAttachmentOnItem(itemName, attachmentName string) ([]byte, error) {
+func (c *cliClient) GetAttachmentOnItem(itemName, attachmentName string) (bytes []byte, retErr error) {
+	file, err := ioutil.TempFile("", "attachmentName")
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err := os.Remove(file.Name()); err != nil {
+			retErr = err
```
droslean: defer will run after the function returns. `retErr` will never get that value.

hongkailiu: It will work (return `retErr = err`) as we discussed in slack.

cmd/ci-secret-bootstrap/main.go

hongkailiu:
/hold

Let me pass the local test first.

hongkailiu:
Local run passed.

/hold cancel

PR #406, opened 2020-01-28
webreg: add job search page

AlexNPavel:
This PR adds a page where users can search for jobs they are interested in.

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -32,6 +32,7 @@ const (
 type options struct {
 	configPath   string
 	registryPath string
+	prowPath     string
 	logLevel     string
 	address      string
 	uiAddress    string
@@ -95,6 +96,7 @@ func gatherOptions() options {
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	fs.StringVar(&o.configPath, "config", "", "Path to config dirs")
 	fs.StringVar(&o.registryPath, "registry", "", "Path to registry dirs")
+	fs.StringVar(&o.prowPath, "prow-config", "", "Path to prow config")
 	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
 	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
 	fs.StringVar(&o.uiAddress, "ui-address", ":8082", "Address to run the registry UI on")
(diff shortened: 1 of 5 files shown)


AlexNPavel:
Test version of PR accessible at: https://ci-tools-406.svc.ci.openshift.org/

cmd/ci-operator-configresolver/main.go
```diff
@@ -128,6 +130,13 @@ func validateOptions(o options) error {
 			return fmt.Errorf("--registry points to a nonexistent directory: %v", err)
 		}
 	}
+	if o.prowPath == "" {
```
stevekuznetsov: I don't think we can do this, because there's no way to incrementally deploy this without blowing up the tool. Adding a nonexistent flag before we bump will break, too. Choose an incremental process.

cmd/ci-operator-configresolver/main.go
```diff
@@ -128,6 +130,13 @@ func validateOptions(o options) error {
...
+		return fmt.Errorf("--prow-config is required")
+	} else {
+		if _, err := os.Stat(o.prowPath); err != nil && os.IsNotExist(err) {
+			return fmt.Errorf("--prow-config points to a nonexistent file/directory: %v", err)
```
stevekuznetsov: Why don't we care to show off any other types of errors?

pkg/webreg/webreg.go
```diff
@@ -639,21 +711,142 @@ func jobHandler(regAgent load.RegistryAgent, confAgent load.ConfigAgent, w http.
...
-	testMap := make(map[string][]string)
+func getAllMultiStageTests(confAgent load.ConfigAgent, jobAgent *prowConfig.Agent) *Jobs {
+	jobs := &Jobs{}
 	configs := confAgent.GetAll()
-	for filename, config := range configs {
+	allRepos := jobAgent.Config().AllRepos
+	for filename, releaseConfig := range configs {
 		name := strings.TrimSuffix(filename, ".yaml")
-		for _, test := range config.Tests {
+		var org, repo, branch string
```
stevekuznetsov: Do we not already have this logic somewhere? I would be surprised.

cmd/ci-operator-configresolver/main.go
AlexNPavel: That's how the `--registry` and `--config` flags work as well. I can change them to print other errors as well.

pkg/webreg/webreg.go
AlexNPavel: Not in code I've written, since I haven't used a slice of `org/repo` strings to identify the org/repo/branch of a config file before. Maybe somewhere else in the code. I'll see if I can find anything.

cmd/ci-operator-configresolver/main.go
AlexNPavel: I've added a fix for this. It breaks the job table (which is just empty without the flag), but everything else still works.

PR #407, opened 2020-01-28
lease: log when acquiring, lower timeout

stevekuznetsov:
The CI NS TTL controller will remove namespaces that are inactive for
more than an hour. Therefore, it does not make sense for CI Operator to
wait for more than an hour while acquiring a lease, or the underlying
namespace will be removed.

/assign @bbguimaraes


petr-muller:
In the land where we're hitting the boskos cap quite often I wonder if we should prolong the namespace reaping timeout instead?

PR #413, opened 2020-02-03
ci-operator: don't verify resolved config if not using resolver

AlexNPavel:
Currently, ci-operator is failing for handwritten prowjob tests because the ci-operator expects all configs to be fully resolved even when not using the resolver. This PR changes ci-operator to use `ValidateAtRuntime` instead of `ValidateResolved` when not using the resolver.

In the future, the `ValidateResolved` check should be run when a multistage config is run in case a handwritten job tries to load a multistage config.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -333,7 +333,14 @@ func (o *options) Complete() error {
 	}
 	o.configSpec = config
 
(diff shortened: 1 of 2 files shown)


AlexNPavel:
/close

We are fixing this by adding the configresolver flags to all prowjobs, including non-generated ones

PR #416, opened 2020-02-04
Add tool to mirror ci-operator configs to a single private repo

droslean:
This is a simple tool that syncs ci-operator configuration files from a single or multiple organization to a single organization and generates a prowgen config file to mark the org as private.

### DIFF
--- a/cmd/ci-op-configs-mirror/main.go
+++ b/cmd/ci-op-configs-mirror/main.go
@@ -1,0 +1,130 @@
+package main
+
+import (
+	"bytes"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+)
+
+const (
+	prowgenConfigFile = ".config.prowgen"
+)
+
+type options struct {
+	dryRun         bool
+	configDir      string
+	destinationDir string
+	fromOrg        string
+	toOrg          string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
+	fs.StringVar(&o.destinationDir, "destination-path", "", "Path of the destination of the generated ci-operator configurations")
+
+	fs.StringVar(&o.fromOrg, "from-org", "", "Mirror the ci-operator configuration files from a single organization, if specified")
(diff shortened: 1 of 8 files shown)


cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
+	fs.StringVar(&o.destinationDir, "destination-path", "", "Path of the destination of the generated ci-operator configurations")
+
+	fs.StringVar(&o.fromOrg, "from-org", "", "Mirror the ci-operator configuration files from a single organization, if specified")
```
petr-muller: I think we should make our lives simpler (by behaving consistently everywhere) here and just use the same logic like in `config-brancher` or `private-org-sync` and by default include all configs that promote official images, possibly only provide the options to limit that set to a single repo or an org.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+	fromOrg        string
+	toOrg          string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
+	fs.StringVar(&o.destinationDir, "destination-path", "", "Path of the destination of the generated ci-operator configurations")
```
petr-muller: Why do we need separate input and destination? I would expect us to be confined inside `ci-operator/config`, both input and output. Is this really needed?

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+	prowgenConfigFullPath := filepath.Join(o.destinationDir, o.toOrg, prowgenConfigFile)
+	if o.dryRun {
+		dryBuff.WriteString(fmt.Sprintf("%s:\n%v", prowgenConfigFullPath, string(data)))
+		fmt.Printf("%s", dryBuff)
+	} else {
+		if err := ioutil.WriteFile(prowgenConfigFullPath, data, 0755); err != nil {
+			logrus.WithError(err).Fatal("error while writing to file")
+		}
+	}
```
petr-muller: Do we even need this? I don't think so: we have the org-wide drop-in inside `ci-operator/config/openshift-priv`, that should be enough, IMO it's not necessary to handle it with the tool.

petr-muller:
I mistakenly approved this, sorry :D

cmd/ci-op-configs-mirror/main.go
droslean: Yea we can just remove it or we can manage it by a `--private` flag. Let's remove it for now.

cmd/ci-op-configs-mirror/main.go
droslean: Yea you are right. I'll address this as well

PR #418, opened 2020-02-04
Add tool to mirror ci-operator configs to a single repo

droslean:
This is a simple tool that syncs ci-operator configuration files from a single or multiple organization to a single organization with the ability to generate a prowgen config file to mark the org as private.

### DIFF
--- a/cmd/ci-op-configs-mirror/main.go
+++ b/cmd/ci-op-configs-mirror/main.go
@@ -1,0 +1,97 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+type options struct {
+	configDir string
+	toOrg     string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
+	fs.StringVar(&o.toOrg, "to-org", "", "Name of the organization which the ci-operator configuration files will be mirrored")
+
+	fs.Parse(os.Args[1:])
+	return o
+}
+
+func (o *options) validate() error {
+	if len(o.configDir) == 0 {
+		return errors.New("--config-path is not defined")
+	}
+
+	if len(o.toOrg) == 0 {
+		return errors.New("--to-org is not defined")
+	}
+	return nil
+}
+
+func main() {
+	o := gatherOptions()
+	if err := o.validate(); err != nil {
+		logrus.WithError(err).Fatal("Invalid option")
+	}
+
+	go func() {
+		interrupts.WaitForGracefulShutdown()
+		os.Exit(1)
+	}()
+
+	callback := func(rbc *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
+		logger := logrus.WithFields(logrus.Fields{"org": repoInfo.Org, "repo": repoInfo.Repo, "branch": repoInfo.Branch})
+
+		if repoInfo.Org == o.toOrg {
+			return nil
+		}
+
+		if !promotion.BuildsOfficialImages(rbc) {
+			logger.Warn("Skipping...")
+			return nil
+		}
+		logger.Info("Processing...")
+
+		if rbc.CanonicalGoRepository == nil {
+			rbc.CanonicalGoRepository = strP(fmt.Sprintf("github.com/%s/%s", repoInfo.Org, repoInfo.Repo))
+		}
+
+		repoInfo.Org = o.toOrg
+
+		rbc.PromotionConfiguration.Disabled = true
+		rbc.PromotionConfiguration.Name = fmt.Sprintf("%s-private", rbc.PromotionConfiguration.Name)
+
+		dataWithInfo := config.DataWithInfo{
+			Configuration: *rbc,
+			Info:          *repoInfo,
+		}
+
+		if err := dataWithInfo.CommitTo(o.configDir); err != nil {
+			logger.WithError(err).Fatal("couldn't create the configuration file")
+		}
+
+		return nil
+	}
+
+	if err := config.OperateOnCIOperatorConfigDir(o.configDir, callback); err != nil {
+		logrus.WithError(err).Fatal("error while generating the ci-operator configuration files")
+	}
+}
+
+func strP(str string) *string {
+	return &str
+}
--- a/test/ci-op-configs-mirror-integration/run.sh
+++ b/test/ci-op-configs-mirror-integration/run.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+WORKDIR="$( mktemp -d )"
+trap 'rm -rf "${WORKDIR}"' EXIT
+
+TEST_ROOT="$( dirname "${BASH_SOURCE[0]}")"
+INPUT_FILES="${TEST_ROOT}/data/input"
+OUTPUT_FILES="${TEST_ROOT}/data/output"
+
+cp -r ${INPUT_FILES} ${WORKDIR}
+
+echo "[INFO] Running ci-op-configs-mirror..."
+if ! ci-op-configs-mirror --to-org "super-priv" --config-path "${WORKDIR}/input" 2> "${WORKDIR}/ci-op-configs-mirror-stderr.log"; then
+  echo "ERROR: ci-op-configs-mirror failed:"
+  cat "${WORKDIR}/ci-op-configs-mirror-stderr.log"
+  exit 1
+fi
+
+echo "[INFO] Validating generated ci-operator configuration files"
+if ! diff -Naupr "${WORKDIR}/input" "${OUTPUT_FILES}"; then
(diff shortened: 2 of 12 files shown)


cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+		if err := ioutil.WriteFile(fileFullPath, data, 0755); err != nil {
+			logger.WithError(err).Fatal("error while writing to file")
+		}
+
+		return nil
+	}
+
+	if err := config.OperateOnCIOperatorConfigDir(o.configDir, callback); err != nil {
+		if err != nil {
```
stevekuznetsov: This is double-checking the `err`?

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+		os.Exit(1)
+	}()
+
+	prowgenConfig := config.Prowgen{Private: true}
+
+	dryBuff := &bytes.Buffer{}
+	callback := func(rbc *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
+		logger := logrus.WithFields(logrus.Fields{"org": repoInfo.Org, "repo": repoInfo.Repo})
+
+		if len(o.fromOrg) > 0 && repoInfo.Org != o.fromOrg {
```
stevekuznetsov: I don't think this is the correct selection logic. We don't want to mirror `openshift/ci-tools` for instance.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+	if err := o.validate(); err != nil {
+		logrus.WithError(err).Fatal("Invalid option")
+	}
+
+	go func() {
+		interrupts.WaitForGracefulShutdown()
+		os.Exit(1)
+	}()
+
+	prowgenConfig := config.Prowgen{Private: true}
```
stevekuznetsov: nit: move this to where it's used

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+	data, err := yaml.Marshal(prowgenConfig)
+	if err != nil {
+		logrus.WithError(err).Fatal("couldn't marshal prowgen configuration")
+	}
+
+	prowgenConfigFullPath := filepath.Join(o.destinationDir, o.toOrg, prowgenConfigFile)
+	if o.dryRun {
+		dryBuff.WriteString(fmt.Sprintf("%s:\n%v", prowgenConfigFullPath, string(data)))
+		fmt.Printf("%s", dryBuff)
```
stevekuznetsov: I have a feeling this will be generally unreadable. Why do we need a dry-run for this tool?

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+			return nil
+		}
+
+		logger.Info("Processing...")
+		data, err := yaml.Marshal(rbc)
+		if err != nil {
+			logger.WithError(err).Fatal("couldn't marshal ci-operator's configuration")
+		}
+
+		filename := fmt.Sprintf("%s-%s-%s.yaml", o.toOrg, repoInfo.Repo, repoInfo.Branch)
```
stevekuznetsov: There is a helper for this transformation from an appropriate `*config.Info`

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+		orgRepoDestinationDir := filepath.Join(o.destinationDir, o.toOrg, repoInfo.Repo)
+		fileFullPath := filepath.Join(orgRepoDestinationDir, filename)
+
+		if o.dryRun {
+			dryBuff.WriteString(fmt.Sprintf("%s:\n%v\n", fileFullPath, string(data)))
+			return nil
+		}
+
+		if err := os.MkdirAll(orgRepoDestinationDir, 0755); err != nil {
```
stevekuznetsov: There should be already code to write config given an `*config.Info` in the config-brancher.

cmd/ci-op-configs-mirror/main.go
droslean: oops

cmd/ci-op-configs-mirror/main.go
droslean: In that case, I think it's better to add an `--ignore-repo` flag.

cmd/ci-op-configs-mirror/main.go
droslean: Its the filename followed by its contents. It seems readable and also it's being used for the integration tests as well.

cmd/ci-op-configs-mirror/main.go
petr-muller: I think we should make our lives simpler (by behaving consistently everywhere) here and just use the same logic like in config-brancher or private-org-sync and by default include all configs that promote official images.

Possibly only provide the options to limit that set to a single repo or an org, but I don't think that's necessary.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if o.private {
+		prowgenConfig := config.Prowgen{Private: true}
+		data, err := yaml.Marshal(prowgenConfig)
+		if err != nil {
+			logrus.WithError(err).Fatal("couldn't marshal prowgen configuration")
+		}
+
+		prowgenConfigPath := filepath.Join(o.toOrg, config.ProwgenFile)
+		if o.dryRun {
```
stevekuznetsov: My comment on this remains

test/ci-op-configs-mirror-integration/run.sh
```diff
@@ -0,0 +1,27 @@
...
+
+WORKDIR="$( mktemp -d )"
+trap 'rm -rf "${WORKDIR}"' EXIT
+
+TEST_ROOT="$( dirname "${BASH_SOURCE[0]}")"
+INPUT_FILES="${TEST_ROOT}/data/input"
+EXPECTED="${TEST_ROOT}/expected.yaml"
+
+echo "[INFO] Running ci-op-configs-mirror in dry-mode..."
+if ! ci-op-configs-mirror --to-org super-priv --config-path "${INPUT_FILES}" --dry-run --private 2> "${WORKDIR}/ci-op-configs-mirror-stderr.log" > "${WORKDIR}/ci-op-configs-mirror-stdout.log"; then
```
petr-muller: I'm not a great fan of this special output for testing. This tool does not interact with anything external, it's just producing files, so it does not need anything like `--dry-run` at all. The integration tests can let it generate the files and then compare the full tree with `diff -Naupr`, just like [repo-init](https://github.com/openshift/ci-tools/blob/master/test/repo-init-integration/run.sh#L77) or [prowgen](https://github.com/openshift/ci-tools/blob/master/test/prowgen-integration/run.sh#L27) tests are doing it.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+			dryBuff.WriteString(fmt.Sprintf("%s:\n%v", prowgenConfigPath, string(data)))
+			fmt.Printf("%s", dryBuff)
+		} else {
+			if err := ioutil.WriteFile(filepath.Join(o.configDir, prowgenConfigPath), data, 0755); err != nil {
+				logrus.WithError(err).Fatal("error while writing to file")
+			}
+		}
+	}
```
petr-muller: I think there is no reason why the tool should be doing this. We already have that file (https://github.com/openshift/release/tree/master/ci-operator/config/openshift-priv) IMO we don't need an 1) additional option 2) additional code. It just increases complexity and adds some coupling.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+const (
+	privatePromotionIS = "4.4-private"
```
petr-muller: Huh, I know this is just a temporary measure, but either 1) name this something that clearly communicates it's not a real destination for promotion (it should definitely not contain any version substring `4.4`) or 2) dynamically create the name by appending the `-private` suffix to whatever name is in the original. This way it's confusing.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -0,0 +1,130 @@
...
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
+	fs.StringVar(&o.toOrg, "to-org", "", "Name of the organization which the ci-operator configuration files will be mirrored")
+
+	fs.BoolVar(&o.dryRun, "dry-run", false, "Whether to actually generate the new ci-operator configuration files")
+	fs.BoolVar(&o.private, "private", false, "Generate a prowgen configuration file for a private organization")
```
petr-muller: See other comments, I do not think we should even have these.

cmd/ci-op-configs-mirror/main.go
droslean: Ok, I'll remove it. Less code is always better!

test/ci-op-configs-mirror-integration/run.sh
droslean: I am ok with generating the files. I will address this.

cmd/ci-op-configs-mirror/main.go
droslean: I will remove the dry run completely.

PR #431, opened 2020-02-10
Migrate openshift/jenkins-openshift-login-plugin to build01

hongkailiu:
required by https://github.com/openshift/release/pull/7088

### DIFF
--- a/pkg/migrate/migrate.go
+++ b/pkg/migrate/migrate.go
@@ -2,6 +2,7 @@ package migrate
 
 import (
 	"fmt"
+	"regexp"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 )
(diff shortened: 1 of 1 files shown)


pkg/migrate/migrate.go
```diff
@@ -12,12 +13,22 @@ const (
 
 var (
 	migratedRepos = sets.NewString(
-		"openshift/jenkins-openshift-login-plugin/master",
+		"openshift/jenkins-openshift-login-plugin/*",
```
droslean: Why not pass the regexes via a flag rathen that hardcoding them into the code?

hongkailiu: This is used only in the migrating phase and will be removed very soon.

PR #433, opened 2020-02-11
webreg: add help page

AlexNPavel:
This PR adds a help page for the test step registry web UI as well as some small descriptions of what each test component is above the respective components table.

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -77,6 +77,7 @@ td {
 </head>
 <body>
 <div class="container">
+<p>Navigation: <a href="/">Step Registry</a> | <a href="/search">Jobs and Search</a> | <a href="/help">Help</a></p>
 `
 
 const htmlPageEnd = `
@@ -87,46 +88,32 @@ const htmlPageEnd = `
 `
 
 const errPage = `
-<h1><a href="/">Openshift CI Step Registry</a></h1>
 {{ . }}
 `
 
-const mainPage = `<h1>Openshift CI Step Registry</h1>
-<p>Not sure what the step registry is or how to use it? Visit the <a href="/help">help page</a>.</p>
-<form action="/search">
-  <div>
-    Search for job: <input type="search" id="search" name="job" placeholder="Job Name">
-	<button>Search</button>
-  </div>
-</form>
+const mainPage = `<h1>Step Registry</h1>
 {{ template "workflowTable" .Workflows }}
 {{ template "chainTable" .Chains }}
 {{ template "referenceTable" .References}}
-{{ template "jobTable" .Jobs }}
 `
 
-const workflowListPage = `
-<h2><a href="/">Openshift CI Step Registry</a> &gt; Workflows
+const workflowListPage = `<h2>Step Registry &gt; Workflows</h2>
 {{ template "workflowTable" .Workflows }}
 `
 
-const chainListPage = `
-<h2><a href="/">Openshift CI Step Registry</a> &gt; Chains
+const chainListPage = `<h2>Step Registry &gt; Chains</h2>
 {{ template "chainTable" .Chains }}
 `
 
-const referenceListPage = `
-<h2><a href="/">Openshift CI Step Registry</a> &gt; References
+const referenceListPage = `<h2>Step Registry &gt; References</h2>
 {{ template "referenceTable" .References }}
 `
 
-const jobListPage = `
-<h2><a href="/">Openshift CI Step Registry</a> &gt; Jobs
+const jobListPage = `<h2>Jobs</h2>
 {{ template "jobTable" .Jobs }}
 `
 
-const referencePage = `
-<h1><a href="/">Openshift CI Step Registry</a></h1>
+const referencePage = `<h1>Step Registry</h1>
 <h2><a href="/references">References</a> &gt; <nobr>{{ .As }}</nobr></h2>
 <p id="documentation">{{ .Documentation }}</p>
 <h2 id="image">Container image used for this step: <span style="font-family:monospace">{{ .From }}</span></h2>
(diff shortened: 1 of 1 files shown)


AlexNPavel:
Running version of updated webui in this PR accessible at: https://ci-tools-433.svc.ci.openshift.org/

pkg/webreg/webreg.go
```diff
@@ -248,6 +250,7 @@ const templateDefinitions = `
 
 {{ define "chainTable" }}
 	<h2 id="chains">Chains</h2>
+	<p>Chains are registry components that allow users to string together multiple steps under one name.</p>
```
alvaroaleman: What about `to string together multiple references as its steps`?  It is not quite obvious that a `Reference` is the same as a `step` IMHO

pkg/webreg/webreg.go
```diff
@@ -328,6 +333,195 @@ const templateDefinitions = `
...
+  documentation: |-
+    The IPI deprovision step chain contains all the individual steps necessary to deprovision an OpenShift cluster.
+</pre>
+
+<h2>Workflow:</h2>
+<p>
+A workflow is the highest level component of the step registry. It is almost
+identical to the syntax of the ci-operator config for multistage tests and
+defines an entire test from start to finish. It has 4 basic components: a
+cluster_type string (eg: <code>aws</code>, <code>azure4</code>,
```
alvaroaleman: Should be `cluster_profille` and IMHO in code blocks. Is `s/azure4/azure/` I think?

pkg/webreg/webreg.go
petr-muller: Yeah, this ref/reference/step makes understanding this all a bit harder for me too.

pkg/webreg/webreg.go
```diff
@@ -328,6 +333,195 @@ const templateDefinitions = `
...
 `
 
+const helpPage = `
+<h1><a href="/">Openshift CI Step Registry</a></h1>
+<h1>What is the Multi-Stage Test and the Test Step Registry?</h1>
+
+<p>
+The multistage test style in the ci-operator is a modular test design that
+allows users to create new tests by combining smaller, individual test steps.
+These individual tests can be put into a shared registry that other tests can
```
petr-muller: It feels like the sentence should start "These individual steps" instead. If the intention is really to speak about "tests", then I suggest to remove the _individual_ that makes it connected to the "individual test step" from the previous sentence.

pkg/webreg/webreg.go
```diff
@@ -328,6 +333,195 @@ const templateDefinitions = `
...
+updated individually to fix bugs or add new features. It also reduces the
+chances of a mistake when copying a feature from one test workflow to
+another.
+</p>
+
+<p>
+To understand how the multistage tests and registry work, we must first talk
+about the 3 components of the test registry and how to use those components
+to create a test.
+</p>
```
petr-muller: I think we should have a three-item bullet list here with a one-sentence summary for each item, before we go to the gory details of individual items.

- Reference: A reference is the lowest level component in the test step registry, describes an individual test step.
- Chain: A chain is a registry component that specifies multiple steps to be run. Any item of the chain can be either a reference or another chain.
- Workflow: A workflow is the highest level component of the step registry. It contains three chains: pre, test, post.

Or something similar.

pkg/webreg/webreg.go
```diff
@@ -328,6 +333,195 @@ const templateDefinitions = `
...
+    requests:
+      cpu: 1000m
+      memory: 100Mi
+  documentation: |-
+	The IPI configure step generates the install-config.yaml file based on the cluster profile and optional input files.
+</pre>
+
+<p>
+Note: the shell script file must follow the naming convention described later
+in this help page.
```
petr-muller: An anchor link would be helpful.

pkg/webreg/webreg.go
```diff
@@ -328,6 +333,195 @@ const templateDefinitions = `
...
+
+<p>
+A reference may be referred to in chains, workflows, and ci-operator configs.
+</p>
+
+<h2>Chain:</h2>
+<p>
+A chain is a registry component that specifies multiple steps to be run.
+Steps are run in the order that they are written. Steps specified by a chain
+can be either references and other chains. Example of a chain:
```
petr-muller: Can items in chains fail to execute? Are the remaining items in the chain executed if one fails?

pkg/webreg/webreg.go
```diff
@@ -328,6 +333,195 @@ const templateDefinitions = `
...
+In this example, the ci-operator config simply specifies the desired cluster
+profile and the <code>origin-e2e</code> workflow shown in the example for the
+<code>Workflow</code> section above.
+</p>
+
+<p>
+Since the ci-operator-config and workflows share the same fields, it is
+possible to override fields specified in a workflow. In cases where both the
+workflow and a ci-operator config specify the same field, the ci-operator config’s
+field has priority (i.e. the value from the ci-operator config is used).
```
petr-muller: IMO this could use an example too.

pkg/webreg/webreg.go
AlexNPavel: Yup, will change to `cluster_profile`. For azure, the profile is `azure4`: https://github.com/openshift/ci-tools/blob/master/pkg/api/types.go#L511

stevekuznetsov:
IMO we might want to bite the bullet and do a top bar or side bar with step content, jobs and search, and help -- adding more and more single lines to the top of the landing page is starting to feel like a hack.

pkg/webreg/webreg.go
AlexNPavel: Added an example for that. Also added comments for fields and syntax highlighting.

stevekuznetsov:
Not sure if this was ready to review but I don't really see the index of "how to" articles on the help page or how to navigate to one?

AlexNPavel:
> Not sure if this was ready to review but I don't really see the index of "how to" articles on the help page or how to navigate to one?

I was originally thinking to make those other pages in separate PRs to prevent this one from getting too large. But either way works.

I've added the other pages (and a help page index) to this PR.

stevekuznetsov:
Styling with the JS looks amazing

/lgtm

PR #434, opened 2020-02-11
ci-op-configs-mirror: Use private namespace and imagestream on Promotion config

droslean:
follow-up of https://github.com/openshift/release/pull/7114
/cc @petr-muller


petr-muller:
/hold

We'll need this but let's wait with merging after we merge all the prep work (https://github.com/openshift/release/pull/7114) and are 100% about the names etc, so we don't have useless change churn in the repo.

PR #436, opened 2020-02-11
rehearse: add options to ignore templates, profiles and registry

petr-muller:
The use case is the redhat-operator-ecosystem/release repo which is
supposed to only contain job config, and all other config items should
be taken from openshift/release repo. This means that templates etc are
never changed in PRs to that repo, so we can safely ignore all these.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -30,6 +30,10 @@ type options struct {
 	debugLogPath string
 	metricsPath  string
 
+	noTemplates       bool
+	noRegistry        bool
+	noClusterProfiles bool
+
 	releaseRepoPath string
 	rehearsalLimit  int
 }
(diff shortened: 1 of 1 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -183,39 +191,60 @@ func rehearseMain() int {
...
+			logger.WithError(err).Error("could not create step registry graph")
+			return gracefulExit(o.noFail, misconfigurationOutput)
+		}
+		changedRegistrySteps, err = config.GetChangedRegistrySteps(o.releaseRepoPath, jobSpec.Refs.BaseSHA, graph)
+		if err != nil {
+			logger.WithError(err).Error("could not get step registry differences")
+			return gracefulExit(o.noFail, misconfigurationOutput)
+		}
+	} else {
+		graph, err = registry.NewGraph(refs, chains, workflows)
```
alvaroaleman: This means that we create a graph out of nothing, as all of `refs`, `chains` and `workflows` are nil, are you sure that is what you want? If yes, please add a comment.

alvaroaleman: And nit: The err check could reside outside of the branches

bbguimaraes:
Code LGTM, a few questions:
- is this purely an optimization?
- do we need separate flags instead of just one?
- can we add an integration test?

cmd/pj-rehearse/main.go
petr-muller: Yeah, I basically need an empty graph, analogic to what I would get if I had an empty step-registry directory. The option should make rehearsal assume there were no changes in the registry (in that case, rehearsal tool does not need the registry at all). It's straightforward for templates and cluster profiles, not so much for registry unfortunately. 

For the error check, I think I can't extract it because in the first branch it's unfortunately in the middle and guards the `GetChangedRegistryStep()` from being called with possibly invalid graph.

petr-muller:
>     * is this purely an optimization?

No. The use case is to make rehearsals work in https://github.com/redhat-operator-ecosystem/release repository. That repository is supposed to contain a partial config (ci-operator configs and prow job configs) for stuff that will eventually run on our Prow. That repo currently does not have step registry, cluster profiles nor templates and the rehearsal tool is failing on that.

Alternatives I considered:
- Put empty directory placeholders in that repo. I did that initially with registry (to make another job paass), but I would need two more dirs like that and that feels like clutter. 
- Make rehearsals not fail on failing to load these structures. This does not feel like the right thing to do in openshift/release though.

>     * do we need separate flags instead of just one?

The complexity does not seem to be that much higher, and the flexibility of being able to gradually enable what we'll eventually need in r-o-e/release seems to be valuable (I expect we may need step-registry, less likely cluster profiles and hopefully templates never).

>     * can we add an integration test?

Yes. It's quite an effort for what seems to be a minor feature (that's why I did not make one initially) but tests never hurt.

PR #443, opened 2020-02-12
ci-operator: add flag to resolve steps

bbguimaraes:
Upstreaming a simpler version of a patch I've had for months on my machine.
This requires a local copy of `openshift/release`, but is useful to test
changes to the registry locally (@petr-muller: FYI).


petr-muller:
Thanks, this will be very useful!

PR #448, opened 2020-02-13
vendor: remove vendor and use go mod directly

AlexNPavel:
Since go 1.13, go modules are automatically used instead of vendor if a `go.mod` file exists in the project. Thus, this project hasn't been using the vendor directory for months now (since Oct. 11 2019 to be more precise), and it should be safe to remove.

This also specifically sets `GOPROXY` to golang 1.13's default. Fedora changes this to `direct` due to potential privacy concerns of having the proxy set by default. However, the proxy speeds up the downloading of modules dramatically. It also fixes a bug with tags/versioning with the `openshift/api` library when using `direct`, so the proxy must be set in the Makefile to build properly on Fedora until that's fixed.


AlexNPavel:
I was wrong. Looks like the golang images for origin-release set `GOFLAGS='-mod=vendor'`. So vendor is actually enabled in our api.ci builds (even though that's likely not the case on most of the developer's workstations...). Will close for now.

/close

PR #449, opened 2020-02-13
Autoconfigbrancher: Add option to self-approve

alvaroaleman:
This allows the autoconfigbrancher to self-approve its PR and hence avoid distracting humans.

We repeat the `BotName()` and `FindIssues` GitHub api call, because `bumper.UpdatePullRequest` doesn't allow to specificy labels that should be set, which I could fix upstream in a follow-up.

### DIFF
--- a/cmd/autoconfigbrancher/main.go
+++ b/cmd/autoconfigbrancher/main.go
@@ -13,6 +13,8 @@ import (
 	"k8s.io/test-infra/experiment/autobumper/bumper"
 	"k8s.io/test-infra/prow/config/secret"
 	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/labels"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


cmd/autoconfigbrancher/main.go
```diff
@@ -205,4 +209,48 @@ func main() {
...
+	if err != nil {
+		return fmt.Errorf("bot name: %v", err)
+	}
+
+	issues, err := gc.FindIssues("is:open is:pr archived:false in:title author:"+me+" "+matchTitle, "updated", false)
+	if err != nil {
+		return fmt.Errorf("find issues: %v", err)
+	} else if len(issues) == 0 {
+		logrus.Info("No reusable issues found")
+		return nil
```
petr-muller: Can we be super-paranoid and also bail out on >1? That would mean we have an accidental match which we probably do not want to auto-approve.

alvaroaleman: Updated

PR #451, opened 2020-02-13
ci-op-configs-mirror: change promotion config naming configuration

droslean:
/cc @petr-muller

### DIFF
--- a/cmd/ci-op-configs-mirror/main.go
+++ b/cmd/ci-op-configs-mirror/main.go
@@ -15,6 +15,10 @@ import (
 	"github.com/openshift/ci-tools/pkg/promotion"
 )
 
+const (
+	privatePromotionNamespace = "ocp-private"
+)
+
 type options struct {
 	configDir string
 	toOrg     string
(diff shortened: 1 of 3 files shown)


cmd/ci-op-configs-mirror/main.go
```diff
@@ -73,7 +77,8 @@ func main() {
 		repoInfo.Org = o.toOrg
 
 		rbc.PromotionConfiguration.Disabled = true
-		rbc.PromotionConfiguration.Name = fmt.Sprintf("%s-private", rbc.PromotionConfiguration.Name)
+		rbc.PromotionConfiguration.Name = fmt.Sprintf("%s-priv", rbc.PromotionConfiguration.Name)
```
petr-muller: Given we are in a separate namespace now, I *think* we may not even need to change the IS names?

droslean: I would really like to have this, just in case.  Also, if we keep the same name, this can lead to confusion.

petr-muller: /shrug

PR #452, opened 2020-02-14
breaking-changes check: check breaking r-o-e/release

petr-muller:
Typically we'll break both repos so checking openshift/release is usually enough, but checking both repos does not hurt anything.


petr-muller:
The test is failing because the r-o-e/release is currently out of sync (https://github.com/redhat-operator-ecosystem/release/pull/10)

PR #456, opened 2020-02-17
Censor seesion in the log of ci-secret-bootstrap

hongkailiu:
https://issues.redhat.com/browse/DPTP-880

Local test results:

```bash
$ ci-secret-bootstrap -bw-user aaa@redhat.com --bw-password-path /tmp/_bw_password --config ./core-services/ci-secret-bootstrap/_config.yaml 
INFO[0000] Parsed kubeconfig context: default           
INFO[0000] Parsed kubeconfig context: build01           
INFO[0000] running bw command ...                        args="[login aaa@redhat.com CENSORED --response]"
INFO[0003] running bw command ...                        args="[--session CENSORED list items]"
INFO[0003] running bw command ...                        args="[--session CENSORED get attachment
```

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -68,16 +68,13 @@ func (o *options) validateOptions() error {
 	return nil
 }
 
-func (o *options) completeOptions() error {
+func (o *options) completeOptions(secrets *sets.String) error {
 	bytes, err := ioutil.ReadFile(o.bwPasswordPath)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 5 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -68,16 +68,26 @@ func (o *options) validateOptions() error {
 	return nil
 }
 
+var (
+	secrets = sets.NewString()
```
stevekuznetsov: We should not use global variables. If you're threading through a closure to the client anyway, just update a local variable in `main()`.

PR #459, opened 2020-02-18
Change all imagestream tag references in the generated ci-op configuration files.

droslean:
/cc @petr-muller 

While writing this, I realize that some of the `base_images` and some of the `build_root` are using images from different imagestreams as well. (ex. ocp/openshift)

We need to make sure that we have those as well.

### DIFF
--- a/cmd/ci-op-configs-mirror/main.go
+++ b/cmd/ci-op-configs-mirror/main.go
@@ -17,6 +17,7 @@ import (
 
 const (
 	privatePromotionNamespace = "ocp-private"
+	ocpNamespace              = "ocp"
 )
 
 type options struct {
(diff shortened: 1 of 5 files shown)


cmd/ci-op-configs-mirror/main.go
```diff
@@ -74,6 +74,22 @@ func main() {
 			rbc.CanonicalGoRepository = strP(fmt.Sprintf("github.com/%s/%s", repoInfo.Org, repoInfo.Repo))
 		}
 
+		if rbc.ReleaseTagConfiguration != nil {
+			rbc.ReleaseTagConfiguration.Name = fmt.Sprintf("%s-priv", rbc.ReleaseTagConfiguration.Name)
+			rbc.ReleaseTagConfiguration.Namespace = privatePromotionNamespace
```
petr-muller: This will change *all* namespaces to the private namespace, right? We only need to change the `ocp` namespace references, no others.

cmd/ci-op-configs-mirror/main.go
```diff
@@ -74,12 +75,23 @@ func main() {
 			rbc.CanonicalGoRepository = strP(fmt.Sprintf("github.com/%s/%s", repoInfo.Org, repoInfo.Repo))
 		}
 
-		repoInfo.Org = o.toOrg
+		if rbc.ReleaseTagConfiguration != nil {
+			filterReleaseTagConfiguration(rbc.ReleaseTagConfiguration)
```
petr-muller: readability:

1. we don't really filter anything, that name is confusing. We are transforming, changing or generating something new, but not filtering.
2. it stutters (I admit it's hard to avoid here)

I'm also not a fan of abusing side effects (modifying the struct by passing the method a pointer). Ideally I'd like see something like this:

```go
rbc.ReleaseTagConfiguration = privateReleaseTag(rbc.ReleaseTagConfiguration)
...
```

cmd/ci-op-configs-mirror/main.go
```diff
@@ -74,12 +75,23 @@ func main() {
...
+		}
+
+		if rbc.BuildRootImage.ImageStreamTagReference != nil {
```
petr-muller: Can we dereference a nil here if `BuildRootImage` is nil?

cmd/ci-op-configs-mirror/main_test.go
```diff
@@ -0,0 +1,173 @@
+package main
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/openshift/ci-tools/pkg/api"
```
petr-muller: nit: separate local (ci-tools) from third-party(apimachinery)

cmd/ci-op-configs-mirror/main.go
droslean: addressed

cmd/ci-op-configs-mirror/main.go
droslean: addressed

cmd/ci-op-configs-mirror/main_test.go
droslean: addressed

PR #464, opened 2020-02-20
prowgen: temporarily rename private postsubmits

petr-muller:
Revert this after DPTP-823

For the CVE effort we expect many builds to be broken for some time
so we are changing the name to not be covered by the regexes that
openshift org build cops are watching.

/cc @droslean


petr-muller:
Expecting `breaking-changes` to fail: https://github.com/openshift/release/pull/7254 is the followup.

PR #470, opened 2020-02-21
Wait pipeline image ready before build src

hongkailiu:
https://issues.redhat.com/browse/DPTP-826

### DIFF
--- a/pkg/steps/input_image_tag.go
+++ b/pkg/steps/input_image_tag.go
@@ -4,13 +4,17 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"time"
 
 	imageapi "github.com/openshift/api/image/v1"
-	"github.com/openshift/ci-tools/pkg/api"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	coreapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/util"
 )
 
 // inputImageTagStep will ensure that a tag exists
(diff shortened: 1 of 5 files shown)


pkg/steps/source.go
```diff
@@ -257,6 +258,18 @@ func buildFromSource(jobSpec *api.JobSpec, fromTag, toTag api.PipelineImageStrea
...
+	log.Printf("waiting for importing %s ...", from.Name)
+	if err := wait.Poll(10*time.Second, 5*time.Minute, func() (bool, error) {
+		pipeline, err := imageClient.ImageStreams(jobSpec.Namespace).Get(api.PipelineImageStream, meta.GetOptions{})
+		if err != nil {
+			return false, err
+		}
+		_, exists := util.ResolvePullSpec(pipeline, string(fromTag), true)
+		return exists, nil
+	}); err != nil {
+		log.Printf("could not resolve imagestream %s: %v", from.Name, err)
```
hongkailiu: @stevekuznetsov 

Should it `panic()` here?

hongkailiu:
/hold

let me rehearse with the built image.

hongkailiu:
Tested the code by https://prow.svc.ci.openshift.org/view/gcs/origin-ci-test/pr-logs/pull/openshift_release/7224/rehearse-7224-pull-ci-integr8ly-cloud-resource-operator-master-vendor/7#1:build-log.txt%3A14

bbguimaraes:
Feel free to ignore for now if this is blocking more important work, but can we:
- use a watch instead of polling?
- poll/watch the specific `ImageStreamTag` instead of the entire `ImageStream`?
- use/refactor the `Done` method (which is currently dead code, as already mentioned elsewhere)?
- add a test?

petr-muller:
> Feel free to ignore for now if this is blocking more important work, but can we:

I'd prefer these to not be ignored. We have enough tech debt in ci-operator already and should really not create more, IMO. The items you posted are all relevant.

hongkailiu:
> * use a watch instead of polling?
> * poll/watch the specific `ImageStreamTag` instead of the entire `ImageStream`?

Asked dev-ex: `isTag` create is not a synced call. We need to check the `status.tags` of the `is` (not `isTag` but i did not ask why) as we do in `util.ResolvePullSpec`. What we do in `ResolvePullSpec` is existing code. I hope i do not introduce more debts by reusing it.

Watch could be working too (also asked dev-ex). But they use `wait.pull` [in their code](https://github.com/openshift/origin/blob/master/test/extended/util/framework.go#L294). 

> * use/refactor the `Done` method (which is currently dead code, as already mentioned elsewhere)?

I do not have answer for this. What is the plan for `Done` method? Do we want to remove it or use it more properly?

I could implement import-checking in the `Done` method and call it in `run`. However, it does not help if the plan is to remove it.

> * add a test?

Fixing the existing test already is a test case for the changed code.
Do you want me to add more tests? The one that I can think of

* the status is not ready until for example 30 secs after the isTag is created

Let me know if you want me to add the above case, or other cases that I missed.

bbguimaraes:
I'm fine with polling and the test. As I said, those changes are not strictly necessary, and migrating to the build server is orders of magnitude more important at the moment.

If we're not able to use `Done` unchanged, then I see no reason to mess with it.

PR #476, opened 2020-02-24
Add code for configuring new template "gcp nested virt custom test image"

dhiller:


dhiller:
This should go in before openshift/release#7290 if that's good.

/hold

PR #479, opened 2020-02-25
Use upstream boskos client retrying

alvaroaleman:
/assign @bbguimaraes 

In contrast to our downstream implementation, it retries all types of requests, not just update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,8 @@ update-vendor:
 		-w /go/src/github.com/openshift/ci-tools \
 		-e GO111MODULE=on \
 		-e GOPROXY=https://proxy.golang.org \
-		golang:1.13 go mod vendor
+		golang:1.13 \
+		/bin/bash -c "go mod tidy && go mod vendor"
 .PHONY: update-vendor
 
 validate-vendor:
--- a/go.mod
+++ b/go.mod
@@ -35,6 +35,6 @@ require (
 	k8s.io/api v0.17.2
 	k8s.io/apimachinery v0.17.2
 	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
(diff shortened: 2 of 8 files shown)


bbguimaraes:
LGTM for the other calls, but we currently configure the (downstream) client to retry updates 60 times before failing, which, with 30s between each update, gives a total timeout of 30m.

Doesn't the upstream client retry for just `1 + 4 + 9 + 25 == 30s` (+ the duration of each request) in total?

alvaroaleman:
> LGTM for the other calls, but we currently configure the (downstream) client to retry updates 60 times before failing, which, with 30s between each update, gives a total timeout of 30m.
> 
> Doesn't the upstream client retry for just 1 + 4 + 9 + 25 == 30s (+ the duration of each request) in total?

Err yeah, but that sounds like a complicated "ignore all update failures" in disguise? Why do we bother tracking them if we have such a high timeout?

bbguimaraes:
Tests take forever and we don't want to abort them because of some transient failure in Boskos. I don't remember exactly where the 30m timeout came from, but it was based on what was already used:

https://github.com/kubernetes/test-infra/commit/0b9d575f6ee0cfcb570708a60b28046d5d826565#diff-563f645b3178add19d07cb5a0871a2fd

alvaroaleman:
@bbguimaraes okay, I guess that sanity of the timeout is an orthogonal discussion anyways. I've removed the second commit, now this Pr changes the timeout on `/update` from `extremely high` to `even higher than that`, due to the two leveled retrying. But I guess thats fine.

bbguimaraes:
We call `Heartbeat` every 30s and all our tests acquire a single lease, so as long as the request doesn't take too long, this shouldn't increase the timeout significantly. I agree the discussion about whether 30m is an acceptable timeout is independent of this change.

/lgtm

PR #490, opened 2020-02-27
[ci-op-configs-mirror] Enable promotion

droslean:
this PR removes the `disable: true` in the promotion configuration on the generated ci-operator configs for the private repositories.

/cc @petr-muller


petr-muller:
Let's wait with this until we have greater confidence about builds in the private org.

PR #493, opened 2020-02-28
Autoowners: Remove non-org members

alvaroaleman:
This PR extends the autoowners tool to filter out entries from the owners file that are not org members in order to produce an owners file that is valid.

Note that I explicitly choose not to use [`trigger.TrustedUser`](https://github.com/kubernetes/test-infra/blob/18fae0a00a2cd5be7ee45563d0b0f940c0e86397/prow/plugins/trigger/trigger.go#L155) because it does a set of api calls per user, which would result in a gigantic token usage. This means that:

* We filter out entries as invalid that are valid due to being a collaborator 
* We filter out entries as invalid due to being member of the trusted org (which we haven't configured, so should be irrelevant)

/assign stevekuznetsov @petr-muller

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -108,13 +108,13 @@ type httpResult struct {
 }
 
 // resolveOwnerAliases computes the resolved (simple or full config) format of the OWNERS file
-func (r httpResult) resolveOwnerAliases() interface{} {
+func (r httpResult) resolveOwnerAliases(cleaner ownersCleaner) interface{} {
 	if !r.simpleConfig.Empty() {
 		return SimpleConfig{
 			Config: repoowners.Config{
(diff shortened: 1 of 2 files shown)


cmd/autoowners/main.go
```diff
@@ -404,3 +409,28 @@ func main() {
...
+	if err != nil {
+		return nil, fmt.Errorf("listOrgMembers failed: %w", err)
+	}
+
+	membersSet := sets.String{}
+	for _, member := range members {
+		membersSet.Insert(member.Login)
+	}
+
+	return func(unfilteredMembers []string) []string {
```
stevekuznetsov: nit:

```go
return func(unfilteredMembers []string) []string {
    return sets.NewString(unfilteredMembers...).Difference(membersSet).List()
}
```

alvaroaleman: Intersection, not Difference, that returns the items that are in membersSet and not in unfilteredMembers

PR #495, opened 2020-02-28
[ci-op-configs-mirror] Update only integration imagestreams to private ones

droslean:
/cc @petr-muller 

This PR adds an extra check to update only integration imagestreams instead. Also, update the `base_rpm_images` as well.

### DIFF
--- a/cmd/ci-op-configs-mirror/main.go
+++ b/cmd/ci-op-configs-mirror/main.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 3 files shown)


cmd/ci-op-configs-mirror/main.go
```diff
@@ -87,6 +88,10 @@ func main() {
 			privateBaseImages(rbc.BaseImages)
 		}
 
+		if len(rbc.BaseRPMImages) > 0 {
+			privateBaseImages(rbc.BaseRPMImages)
```
stevekuznetsov: nit: functions with side-effects are hard to follow and we should prefer not to use them, instead using functions with clear data flow

cmd/ci-op-configs-mirror/main.go
```diff
@@ -144,3 +149,10 @@ func privatePromotionConfiguration(promotion *api.PromotionConfiguration) {
 func strP(str string) *string {
 	return &str
 }
+
+func isIntegrationImageStream(name string) bool {
```
stevekuznetsov: Technically you need to walk every configuration we care about and see which image streams we promote to, right? Not everyone promotes to something prefixed with `4.`

petr-muller: It should be sufficient to use the same condition that is used in `promotesOfficialImages` check - we only mirror configs that promtoe into official imagestreams.

But in general yes, any reference to any IS that is a promotion target of the private org config should be changed.

PR #503, opened 2020-03-02
Add cluster for base_rpm_images

hongkailiu:
This PR is used to generate https://github.com/openshift/release/pull/7415

We missed the case base_rpm_images before.

### DIFF
--- a/cmd/config-migrator/main.go
+++ b/cmd/config-migrator/main.go
@@ -68,11 +68,17 @@ func generateMigratedConfigs(input config.DataWithInfo) []config.DataWithInfo {
 		return nil
 	}
 
-	var newBaseImages map[string]api.ImageStreamTagReference
-	for k, baseImage := range futureConfig.BaseImages {
-		if newBaseImages == nil {
-			newBaseImages = map[string]api.ImageStreamTagReference{}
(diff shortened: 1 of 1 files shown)


cmd/config-migrator/main.go
```diff
@@ -68,6 +68,18 @@ func generateMigratedConfigs(input config.DataWithInfo) []config.DataWithInfo {
 		return nil
 	}
 
+	var newBaseRPMImages map[string]api.ImageStreamTagReference
+	for k, baseRPMImage := range futureConfig.BaseRPMImages {
+		if newBaseRPMImages == nil {
```
alvaroaleman: Just initialize it in L17_

hongkailiu: L71 right?

I did not do it there because i wanna tell between nil vs emptyMap. Only nil will make the yaml serialization work for `omitempty`? Let me test it.

hongkailiu: You are right.
EmpytMap works too.

PR updated.

hongkailiu:
> A tiny test maybe?

This tool is a temporary tool for changing ci-op's config (in a batch mode).
I think it will be removed when our migration is done.
Could i skip the test implementation for the above reasons/excuses?

PR #504, opened 2020-03-02
ci-operator: Do not overwrite TTL annotations unless explicit

smarterclayton:
When the user does not specify a TTL argument, don't overwrite an
existing annotation value on the namespace. This allows a later
invocation to inherit an earlier invocation's values, but still
override if necessary.


smarterclayton:
/assign @stevekuznetsov 

as discussed on slack

PR #506, opened 2020-03-02
Added a libvirt profile for remote s390x testing.

jaypoulz:
This patch was modeled after #425, so it tightly couples the architecture to the cluster profile. Some consideration was made to leaving this profile as a generic "libvirt" profile, but the current x86_64 tests use the AWS profile. Also, it would be useful in terms of tracking metrics to ensure that Z/s390x can be filtered out.


jaypoulz:
@bbguimaraes @droslean Can I get one of you to review my work at your earliest convenience?

I'm under a tight deadline for deliverables and this is preventing me from making progress on testing https://github.com/openshift/release/pull/7161

PR #509, opened 2020-03-02
ci-operator: make build functions interruptible

bbguimaraes:
/hold


bbguimaraes:

PR #510, opened 2020-03-02
ci-operator: make pod functions interruptible

bbguimaraes:
/hold


bbguimaraes:

bbguimaraes:
/hold

The jobs are failing and retesting will only exacerbate the problems in the CI clusters.

PR #511, opened 2020-03-02
ci-operator: make other functions interruptible

bbguimaraes:
/hold


bbguimaraes:

bbguimaraes:
@stevekuznetsov: PTAL again, this now includes the timeout increase in image importing.

/retest

PR #520, opened 2020-03-04
Reintegrate nested virt enablement into custom-test-image template

dhiller:
We add two variables:
* bool switch for enable-nested-virt on/off
* nested-virt custom image

Prerequisite for openshift/release#7457

/cc @davidvossel @sallyom


dhiller:
/cc @petr-muller 

we decided to put the changes for enabling nested-virt into the custom-test-image template again. So we removed i.e. OpenshiftInstallerGCPNestedVirtCustomTestImageClusterTestConfiguration. Sorry for the noise...

PR #523, opened 2020-03-05
secret-wrapper: update secret unconditionally

bbguimaraes:


bbguimaraes:
We are the only users ([via `ci-operator`](https://github.com/openshift/ci-tools/blob/1c28bd8abe336458f6ab825747edc42cfc0f29ae/pkg/steps/multi_stage.go#L289-L312)), so a flag isn't really necessary.

As for the change in behavior, this was requested by users and is required for some scenarios. I don't see a downside in always copying the files or any potential for breakage.

PR #525, opened 2020-03-05
prowgen: duplicate tests on build01 that don't use quota

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @hongkailiu @petr-muller

PR #527, opened 2020-03-06
Revert "prowgen: duplicate tests on build01 that don't use quota"

droslean:
Reverts openshift/ci-tools#525

/hold


alvaroaleman:
cc @stevekuznetsov , this broke the config validation for o/release it seems

alvaroaleman:
/hold
to make sure breaking changes actually passes now

PR #526, opened 2020-03-05
deps: bump prow

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

Generated with

```
GO111MODULE=on go get k8s.io/test-infra
GO111MODULE=on go mod vendor
```

PR #529, opened 2020-03-06
Revert "deps: bump prow"

droslean:
Reverts openshift/ci-tools#526


alvaroaleman:
/hold
Until we are certain this actually fixes things.

PR #533, opened 2020-03-07
test/pj-rehearse: fix an incorrect job

AlexNPavel:
This PR fixes a job in the `pj-rehearse` integration tests that was incorrect. It was using the outdated and unused `ci-operator-configs` configmap, which our `IsCiOpConfigCM` regex did not match against. This resulted in `CONFIG_SPEC` not being inlined. This has been changed to the `ci-operator-misc-configs` configmap, which is used when a configmap is not a master or release branch config.

The job was also referencing a non-existent `ci-operator` config which would cause the `pj-rehearse` to fail and had `openshift/origin` as the org-repo instead of `super/duper`. This has also been fixed.


AlexNPavel:
Also found an incorrect `extra_refs: base_ref`

PR #534, opened 2020-03-07
prowgen,pj-rehearse: do not add CONFIG_SPEC to generated jobs

AlexNPavel:
This PR removes generation of the `CONFIG_SPEC` env var from `ci-operator-prowgen`. It also updates `pj-rehearse` to not depend on `CONFIG_SPEC` being set and instead use other methods of determining what config file needs to be inlined.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -104,6 +104,11 @@ func CompletePrimaryRefs(refs pjapi.Refs, jb prowconfig.JobBase) *pjapi.Refs {
 	return &refs
 }
 
+func getTrimmedBranch(branches []string) string {
+	return strings.TrimPrefix(strings.TrimSuffix(branches[0], "$"), "^")
+
+}
+
 func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber int, refs *pjapi.Refs) (*prowconfig.Presubmit, error) {
 	var rehearsal prowconfig.Presubmit
 	deepcopy.Copy(&rehearsal, source)
@@ -114,7 +119,7 @@ func makeRehearsalPresubmit(source *prowconfig.Presubmit, repo string, prNumber 
 	var context string
 
 	if len(source.Branches) > 0 {
-		branch = strings.TrimPrefix(strings.TrimSuffix(source.Branches[0], "$"), "^")
+		branch = getTrimmedBranch(source.Branches)
 		if len(repo) > 0 {
 			orgRepo := strings.Split(repo, "/")
 			jobOrg := orgRepo[0]
@@ -228,9 +233,14 @@ func hasRehearsableLabel(labels map[string]string) bool {
 // of the needed config file passed to the job as a direct value. This needs
 // to happen because the rehearsed Prow jobs may depend on these config files
 // being also changed by the tested PR.
-func inlineCiOpConfig(container v1.Container, ciopConfigs config.ByFilename, resolver registry.Resolver, loggers Loggers) error {
+func inlineCiOpConfig(container v1.Container, ciopConfigs config.ByFilename, resolver registry.Resolver, info config.Info, loggers Loggers) error {
+	configSpecSet := false
+	// replace all ConfigMapKeyRef mounts with inline config maps
 	for index := range container.Env {
 		env := &(container.Env[index])
+		if env.Name == "CONFIG_SPEC" {
+			configSpecSet = true
+		}
 		if env.ValueFrom == nil {
 			continue
 		}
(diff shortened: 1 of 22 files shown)


AlexNPavel:
This currently is based on the branch from PR #533 since it needs the fix from there to function properly. Will rebase against master once #533 gets merged.

AlexNPavel:
/hold
To make sure this doesn't get merged before 533

pkg/rehearse/jobs.go
```diff
@@ -303,7 +360,18 @@ func (jc *JobConfigurer) ConfigurePeriodicRehearsals(periodics config.Periodics)
 	filteredPeriodics := filterPeriodics(periodics, jc.loggers.Job)
 	for _, job := range filteredPeriodics {
 		jobLogger := jc.loggers.Job.WithField("target-job", job.Name)
-		if err := jc.configureJobSpec(job.Spec, jc.loggers.Debug.WithField("name", job.Name)); err != nil {
+		variant := ""
+		if newVariant, ok := job.Labels[jobconfig.ProwJobLabelVariant]; ok {
```
stevekuznetsov: nit: this is not really a new variant, maybe `variantLabel` or something could be a more appropriate name

pkg/rehearse/jobs.go
```diff
@@ -303,7 +360,18 @@ func (jc *JobConfigurer) ConfigurePeriodicRehearsals(periodics config.Periodics)
...
+			variant = newVariant
+		}
+		var orgrepo, branch string
+		if len(job.ExtraRefs) != 0 {
```
stevekuznetsov: This is *almost* an `Info{}` -- should be a smell that passing these around together but not in a structure is not the most appropriate ;)

pkg/rehearse/jobs.go
```diff
@@ -320,16 +388,21 @@ func (jc *JobConfigurer) ConfigurePresubmitRehearsals(presubmits config.Presubmi
...
-			rehearsal, err := makeRehearsalPresubmit(&job, repo, jc.prNumber, jc.refs)
+			jobLogger := jc.loggers.Job.WithFields(logrus.Fields{"target-repo": orgrepo, "target-job": job.Name})
+			rehearsal, err := makeRehearsalPresubmit(&job, orgrepo, jc.prNumber, jc.refs)
 			if err != nil {
 				jobLogger.WithError(err).Warn("Failed to make a rehearsal presubmit")
 				continue
 			}
 
-			if err := jc.configureJobSpec(rehearsal.Spec, jc.loggers.Debug.WithField("name", job.Name)); err != nil {
+			variant := ""
```
stevekuznetsov: DRY the variant getting, a simple helper with a test would be nice

pkg/rehearse/jobs_test.go
```diff
@@ -248,16 +292,22 @@ func TestInlineCiopConfig(t *testing.T) {
 	}
 }
 
-func makeTestingPresubmit(name, context, branch string) *prowconfig.Presubmit {
+func makeTestingPresubmit(name, context, branch, variant string) *prowconfig.Presubmit {
```
stevekuznetsov: nit: I don't like helper functions like this in tests as they begin to accrue their own complex logic. Static test data is simpler and harder to mess up

pkg/rehearse/jobs_test.go
```diff
@@ -431,14 +481,14 @@ func TestExecuteJobsErrors(t *testing.T) {
 	}{{
 		description: "fail to Create a prowjob",
 		jobs: map[string][]prowconfig.Presubmit{targetOrgRepo: {
-			*makeTestingPresubmit("job1", "ci/prow/job1", "master"),
+			*makeTestingPresubmit("job1", "ci/prow/job1", "master", ""),
```
stevekuznetsov: Passing empty strings to a function and checking if they are not empty inside is smelly -- can we make a `makeTestPresubmitWithVariant` or similar?

pkg/rehearse/jobs_test.go
```diff
@@ -35,12 +35,56 @@ import (
...
+	"github.com/openshift/ci-tools/pkg/jobconfig"
 	"github.com/openshift/ci-tools/pkg/load"
 	"github.com/openshift/ci-tools/pkg/registry"
 )
 
 const testingRegistry = "../../test/multistage-registry/registry"
 
+// configFiles contains the info needed to allow inlineCiOpConfig to successfully inline
+// CONFIG_SPEC and not fail
+var configFiles = config.ByFilename{
```
stevekuznetsov: This is a global and if someone does not know that they cannot change it, you could get bugs in test from modification. We could wrap this in a producer method and give each consumer their own version

pkg/rehearse/jobs_test.go
AlexNPavel: Looks like none of the tests needed the variant flags anyway? I'll just revert it back to how it was before. We may need to add some tests for variants in a followup PR though.

PR #537, opened 2020-03-09
Private prow configs mirroring tool

droslean:
This tool detects the configuration of all of the org/repos that promotes official images and injects an `openshift-priv/repo` version of the config.

Affected configuration: 

`config.branch-protection`
`config.context_options`
`config.tide.merge_method`
`config.tide.queries`
`config.tide.pr_status_base_urls`
`config.plank.default_decoration_configs`
`config.plank.job_url_prefix_config`
`plugins.approve`
`plugins.lgtm`
`plugins.plugins`
`plugins.bugzilla`

### DIFF
--- a/cmd/private-prow-configs-mirror/main.go
+++ b/cmd/private-prow-configs-mirror/main.go
@@ -1,0 +1,436 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"path/filepath"
+	"runtime"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	prowconfig "k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/plugins"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+const (
+	openshiftPrivOrg = "openshift-priv"
+)
+
+type options struct {
+	releaseRepoPath string
+}
+
+type orgReposWithOfficialImages map[string]sets.String
+
+func (o orgReposWithOfficialImages) isOfficialRepo(org, repo string) bool {
+	if _, ok := o[org]; ok {
+		if o[org].Has(repo) {
+			return true
+		}
+	}
+	return false
+}
+
+func (o *orgReposWithOfficialImages) isOfficialRepoFull(orgRepo string) bool {
+	if orgRepoList := strings.Split(orgRepo, "/"); len(orgRepoList) == 2 {
+		return o.isOfficialRepo(orgRepoList[0], orgRepoList[1])
+	}
+	return false
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.releaseRepoPath, "release-repo-path", "", "Path to a openshift/release repository directory")
+	fs.Parse(os.Args[1:])
+	return o
+}
+
+func (o *options) validate() error {
+	if len(o.releaseRepoPath) == 0 {
+		return errors.New("--release-repo-path is not defined")
+	}
+	return nil
+}
+
+func loadProwPlugins(pluginsPath string) (*plugins.Configuration, error) {
+	agent := plugins.ConfigAgent{}
+	if err := agent.Load(pluginsPath, false); err != nil {
+		return nil, err
+	}
+	return agent.Config(), nil
+}
+
+func updateProwConfig(configFile string, config prowconfig.ProwConfig) error {
+	data, err := yaml.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("could not marshal Prow configuration: %v", err)
+	}
+
+	return ioutil.WriteFile(configFile, data, 0644)
+}
+
+func updateProwPlugins(pluginsFile string, config *plugins.Configuration) error {
+	data, err := yaml.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("could not marshal Prow configuration: %v", err)
+	}
+
+	return ioutil.WriteFile(pluginsFile, data, 0644)
+}
+
+func privateOrgRepo(repo string) string {
+	return fmt.Sprintf("%s/%s", openshiftPrivOrg, repo)
+}
+
+func getOrgReposWithOfficialImages(releaseRepoPath string) (orgReposWithOfficialImages, error) {
+	ret := make(orgReposWithOfficialImages)
+
+	callback := func(c *api.ReleaseBuildConfiguration, i *config.Info) error {
+		if !promotion.BuildsOfficialImages(c) {
+			return nil
+		}
+
+		if _, ok := ret[i.Org]; !ok {
+			ret[i.Org] = sets.NewString(i.Repo)
+		} else {
+			ret[i.Org].Insert(i.Repo)
+		}
+
+		return nil
+	}
+
+	if err := config.OperateOnCIOperatorConfigDir(filepath.Join(releaseRepoPath, config.CiopConfigInRepoPath), callback); err != nil {
+		return ret, fmt.Errorf("error while operating in ci-operator configuration files: %v", err)
+	}
+
+	return ret, nil
+}
+
+func injectPrivateBranchProtection(branchProtection prowconfig.BranchProtection, orgRepos orgReposWithOfficialImages) {
+	privateOrg := prowconfig.Org{
+		Repos: make(map[string]prowconfig.Repo),
+	}
+
+	logrus.Info("Processing...")
+	for orgName, orgValue := range branchProtection.Orgs {
+		for repoName, repoValue := range orgValue.Repos {
+			if orgRepos.isOfficialRepo(orgName, repoName) {
+				logrus.WithField("repo", repoName).Info("Found")
+				privateOrg.Repos[repoName] = repoValue
+			}
+		}
+
+	}
+
+	if len(privateOrg.Repos) > 0 {
+		branchProtection.Orgs[openshiftPrivOrg] = privateOrg
+	}
+}
+
+func injectPrivateTideOrgContextPolicy(contextOptions prowconfig.TideContextPolicyOptions, orgRepos orgReposWithOfficialImages) {
+	logrus.Info("Processing...")
+	privateOrgRepos := make(map[string]prowconfig.TideRepoContextPolicy)
+
+	for orgName, orgValue := range contextOptions.Orgs {
+		for repoName, repoValue := range orgValue.Repos {
+			if orgRepos.isOfficialRepo(orgName, repoName) {
+				logrus.WithField("repo", repoName).Info("Found")
+				privateOrgRepos[repoName] = repoValue
+			}
+		}
+	}
+
+	if len(privateOrgRepos) > 0 {
+		contextOptions.Orgs[openshiftPrivOrg] = prowconfig.TideOrgContextPolicy{Repos: privateOrgRepos}
+	}
+}
+
+func injectPrivateReposTideQueries(tideQueries []prowconfig.TideQuery, orgRepos orgReposWithOfficialImages) {
+	logrus.Info("Processing...")
+
+	for index, tideQuery := range tideQueries {
+		repos := sets.NewString(tideQuery.Repos...)
+
+		for _, orgRepo := range tideQuery.Repos {
+			if orgRepos.isOfficialRepoFull(orgRepo) {
+				repo := strings.Split(orgRepo, "/")[1]
+
+				logrus.WithField("repo", repo).Info("Found")
+				repos.Insert(privateOrgRepo(repo))
+			}
+		}
+
+		tideQueries[index].Repos = repos.List()
+	}
+}
+
+func injectPrivateMergeType(tideMergeTypes map[string]github.PullRequestMergeType, orgRepos orgReposWithOfficialImages) {
+	logrus.Info("Processing...")
+
+	for orgRepo, value := range tideMergeTypes {
+		if orgRepos.isOfficialRepoFull(orgRepo) {
+			repo := strings.Split(orgRepo, "/")[1]
+
+			logrus.WithField("repo", repo).Info("Found")
+			tideMergeTypes[privateOrgRepo(repo)] = value
+		}
+	}
+}
+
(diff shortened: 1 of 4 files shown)


cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+				if _, ok := orgs[orgName].repos[repoName]; ok {
+					logger.WithField("repo", repoName).Info("Found")
+					privateOrg.Repos[repoName] = repoValue
+				}
+			}
+		}
+	}
+
+	if len(privateOrg.Repos) > 0 {
+		logger.WithField("value", asYAML(privateOrg)).Info("Generating openshift-priv org.")
```
stevekuznetsov: I do not think it's useful to put the entire YAML into a log line like this

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+func injectPrivateTideOrgContextPolicy(contextOptions prowconfig.TideContextPolicyOptions, orgs orgsWithOfficialImages) {
+	logger := logrus.WithField("component", "config.context_options")
+	logger.Info("Processing...")
+	privateOrgRepos := make(map[string]prowconfig.TideRepoContextPolicy)
+	for orgName, orgValue := range contextOptions.Orgs {
+		if _, ok := orgs[orgName]; ok {
+			for repoName, repoValue := range orgValue.Repos {
+				if _, ok := orgs[orgName].repos[repoName]; ok {
```
stevekuznetsov: Better yet, write your types to describe what they are going to do. My earlier suggestions were based on your names, which made it sound like you needed to know the list of images that were promoted. As of yet I'm seeing this access pattern commonly, so you want a type that makes simple and easy to answer the question "is this org/repo(branch) promoting and do I therefore need to care about it?"

stevekuznetsov: So, for instance, if you had:

```go
type officialBranchesByOrgRepo map[string]map[string]sets.String
```

You could have 

```go
func (o *officialBranchesByOrgRepo) isOfficialRepo(org, repo string) bool {
    if repos, ok := o[org]; ok {
        _, isOfficial := o[org]
        return isOfficial
    }
    return false
}

func (o *officialBranchesByOrgRepo) isOfficialBranch(org, repo, branch string) bool {
    if ok := o.isOfficialRepo(org,repo); ok  {
        _, isOfficial := o[org][repo][branch]
        return isOfficial
    }
    return false
}
```

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+// In addition, it generates a repo. if it contains an extra non-common field.
+func mergeCommonPrivatePlugins(plugins map[string][]string, repeatedValues sets.String, logger *logrus.Entry) {
+	privateOrgPlugins := sets.NewString()
+
+	for repo, values := range plugins {
+		if strings.HasPrefix(repo, openshiftPrivOrg) {
+			repoLevelPlugins := sets.NewString()
+
+			for _, p := range values {
+				if repeatedValues.Has(p) {
```
stevekuznetsov: This is a set intersection, we can use the helper

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+					privateOrgPlugins.Insert(p)
+					continue
+				}
+				repoLevelPlugins.Insert(p)
```
stevekuznetsov: This is a set difference

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+		logger.WithField("value", asYAML(privateOrgPlugins.List())).Info("Generating openshift-priv org.")
+		plugins[openshiftPrivOrg] = privateOrgPlugins.List()
+	}
+}
+
+func getRepeatedValues(values []string) sets.String {
+	temp := sets.NewString()
+	repeated := sets.NewString()
+
+	for _, value := range values {
```
stevekuznetsov: This should use set intersections

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+func updateProwConfig(configFile string, config prowconfig.ProwConfig) error {
+	data, err := yaml.Marshal(config)
+	if err != nil {
+		return fmt.Errorf("could not marshal Prow configuration: %v", err)
+	}
+
+	return ioutil.WriteFile(configFile, data, 0644)
+}
+
+func updateProwPlugins(pluginsFile string, config *plugins.Configuration) error {
```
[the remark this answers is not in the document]
petr-muller: I did not have time today to do a full review, but I like to see the code structured like this. The `main()` then contains a nice skimmable overview of all the modifications that are needed, and the details of each individual actions are hidden in implementation, separate from each other.

If you're not feeling really strongly about this, I'd like this structure to be kept.

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+type orgWithOfficialImages struct {
+	repos map[string]repo
+}
+
+type repo struct {
+	branches sets.String
+}
+type orgsWithOfficialImages map[string]orgWithOfficialImages
```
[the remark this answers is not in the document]
droslean: Each org can have multiple repos. Either way, I'll find a way to simplify this.

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,460 @@
...
+
+	return ioutil.WriteFile(pluginsFile, data, 0644)
+}
+
+func privateOrgRepo(repo string) string {
+	return fmt.Sprintf("%s/%s", openshiftPrivOrg, repo)
+}
+
+func asYAML(v interface{}) string {
+	b, err := yaml.Marshal(v)
```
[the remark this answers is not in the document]
droslean: You are right, I'll remove this.

cmd/private-prow-configs-mirror/main.go
droslean: Can't use set intersections here.  My purpose here is to grab a set of values that exist more than once in the `allPrivate []string`

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,437 @@
...
+	if _, ok := o[org]; ok {
+		if o[org].Has(repo) {
+			return true
+		}
+	}
+	return false
+}
+
+func (o *orgReposWithOfficialImages) isOfficialRepoFull(orgRepo string) bool {
+	if orgRepoList := strings.Split(orgRepo, "/"); len(orgRepoList) > 1 {
```
petr-muller:
```suggestion
	if orgRepoList := strings.Split(orgRepo, "/"); len(orgRepoList) == 2 {
```

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+
+	return ioutil.WriteFile(pluginsFile, data, 0644)
+}
+
+func privateOrgRepo(repo string) string {
+	return fmt.Sprintf("%s/%s", openshiftPrivOrg, repo)
+}
+
+func getOrgReposWithOfficialImages() (map[string]sets.String, error) {
```
petr-muller: We defined a type `orgReposWithOfficialImages`, we should use it in the signature

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,437 @@
...
+	ret := make(map[string]sets.String)
```
petr-muller: and here too

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		logrus.WithError(err).Fatal("couldn't get current directory")
+	}
+
+	configs := config.GetAllConfigs(currentDir, logrus.NewEntry(logrus.New()))
+	pluginsConfig, err := loadProwPlugins(o.prowPlugins)
+	if err != nil {
+		logrus.WithError(err).Fatal("could not load Prow plugin configuration")
+	}
+	prowConfig := configs.Prow.ProwConfig
```
petr-muller: This will NPD if that config failed to be read, see the doc for `GetAllConfigs`.

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,437 @@
...
+	if err := o.validate(); err != nil {
+		logrus.WithError(err).Fatal("Invalid option")
+	}
+
+	go func() {
+		interrupts.WaitForGracefulShutdown()
+		os.Exit(1)
+	}()
+
+	currentDir, err := os.Getwd()
```
petr-muller: This does not match the expectations given by options of the program... I tried to test and I got a crash:

```console
$ go run ./cmd/private-prow-configs-mirror/ --ci-op-config-path ../release/ci-operator/config/ --prow-config=../release/core-services/prow/02_config/_config.yaml --prow-plugins ../release/core-services/prow/02_config/_plugins.yaml
ERRO[0000]load.go:159 github.com/openshift/ci-tools/pkg/config.OperateOnCIOperatorConfigSubdir.func1() Failed to walk CI Operator configuration dir  error="lstat /home/remote/pmuller/Projects/RH/go/src/github.com/openshift/ci-tools/ci-operator/config: no such file or directory" source-file=/home/remote/pmuller/Projects/RH/go/src/github.com/openshift/ci-tools/ci-operator/config
WARN[0000] failed to load ci-operator configuration from release repo  error="lstat /home/remote/pmuller/Projects/RH/go/src/github.com/openshift/ci-tools/ci-operator/config: no such file or directory"
WARN[0000] failed to load Prow configuration from release repo  error="stat /home/remote/pmuller/Projects/RH/go/src/github.com/openshift/ci-tools/core-services/prow/02_config/_config.yaml: no such file or directory"
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x13ebd94]

goroutine 1 [running]:
main.main()
	/home/remote/pmuller/Projects/RH/go/src/github.com/openshift/ci-tools/cmd/private-prow-configs-mirror/main.go:410 +0x484
exit status 2
```

Don't assume to run in the correct cwd... The inputs and ouputs need to be more systematic (and documented). At the very least the path to relesae repo needs to be a param. We need full paths to the configs anyway (because we need to write to them), so I think we should just read the configs directly ourselves and not use `GetAllConfigs()` helper.

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,437 @@
...
+			if orgRepos.isOfficialRepo(orgName, repoName) {
+				logrus.WithField("repo", repoName).Info("Found")
+				privateOrg.Repos[repoName] = repoValue
+			}
+		}
+
+	}
+
+	if len(privateOrg.Repos) > 0 {
+		logrus.WithField("value", privateOrg).Info("Generating openshift-priv org.")
```
petr-muller: Given `privateOrg` is a deep object, I think logging it verbatim is not very useful... Personally I think we don't even need this log at all (in all of these methods).

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,441 @@
...
+
+	callback := func(c *api.ReleaseBuildConfiguration, i *config.Info) error {
+		if !promotion.BuildsOfficialImages(c) {
+			return nil
+		}
+
+		if _, ok := ret[i.Org]; !ok {
+			ret[i.Org] = sets.NewString(i.Repo)
+		} else {
+			ret[i.Org].Insert(i.Repo)
```
bbguimaraes: The value returned by the first `ret[i.Org]` can be used here to avoid a second look-up.

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,441 @@
...
+func mergeCommonPrivatePlugins(plugins map[string][]string, privateReposWithValues map[string][]string, repeatedValues sets.String) {
+	privateOrgPlugins := sets.NewString()
+
+	for repoName, repoValues := range privateReposWithValues {
+		repoLevelPlugins := sets.NewString()
+		valuesSet := sets.NewString(repoValues...)
+
+		if len(repeatedValues) > 0 {
+			commonValues := valuesSet.Intersection(repeatedValues)
+			privateOrgPlugins.Insert(commonValues.List()...)
```
stevekuznetsov: Use a `Union()` in these cases

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,441 @@
...
+
+			if diffValues := valuesSet.Difference(repeatedValues); len(diffValues) > 0 {
+				repoLevelPlugins.Insert(diffValues.List()...)
```
stevekuznetsov: Use a union with the difference, does not matter if the difference is empty

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,441 @@
...
+// and merges them into the org `openshift-priv` level.
+// In addition, it generates a repo. if it contains an extra non-common field.
+func mergeCommonPrivatePlugins(plugins map[string][]string, privateReposWithValues map[string][]string, repeatedValues sets.String) {
+	privateOrgPlugins := sets.NewString()
+
+	for repoName, repoValues := range privateReposWithValues {
+		repoLevelPlugins := sets.NewString()
+		valuesSet := sets.NewString(repoValues...)
+
+		if len(repeatedValues) > 0 {
```
stevekuznetsov: If there are none, why call this function at all?

droslean: If there are none we need only to create the repo level, otherwise we will merge the common values under the org level (openshift-priv)

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,441 @@
...
+
+const (
+	openshiftPrivOrg = "openshift-priv"
+)
+
+type options struct {
+	realeseRepoPath string
+}
+
+type orgReposWithOfficialImages map[string]sets.String
```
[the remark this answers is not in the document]
droslean: It's being used as a type in all the functions below, Of course, we can just replace them with just a `map[string]sets.String` instead but I would like to keep it this way to make the code more readable.

cmd/private-prow-configs-mirror/main.go
droslean: we need to update the value directly. Either way, a second lookup will happen when we will update the actual value.

droslean:
I will open a follow-up PR to include this tool run in the `autoconfigbrancher` pipeline as well.

cmd/private-prow-configs-mirror/main.go
```diff
@@ -0,0 +1,436 @@
...
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+const (
+	openshiftPrivOrg = "openshift-priv"
+)
+
+type options struct {
+	realeseRepoPath string
```
petr-muller:
```suggestion
	releaseRepoPath string
```

droslean:
> There still seem to be some assumption about directory locations:

@petr-muller Fixed. Thanks!

PR #538, opened 2020-03-09
Add the CVP trigger tool skeleton

petr-muller:
CVP Trigger tool will be used by the CVP pipeline to parametrize and
trigger the verification jobs for optional operator artifacts built
internally in RH.

Right now the tool reads Prow config, finds a specified periodic job
specification, creates a Prowjob, submits it to the cluster and waits
until the job finishes.

In the future we expect this tool to perform some job parametrization.

/cc @openshift/openshift-team-developer-productivity-test-platform  @dirgim

### DIFF
--- a/cmd/cvp-trigger/main.go
+++ b/cmd/cvp-trigger/main.go
@@ -1,0 +1,148 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/rest"
+
+	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	pjclientset "k8s.io/test-infra/prow/client/clientset/versioned"
+	prowconfig "k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/pjutil"
+
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const (
+	prowConfigPathOption = "prow-config-path"
+	jobConfigPathOption  = "job-config-path"
+	periodicOption       = "periodic"
+)
+
+type options struct {
+	confirm bool
+
+	prowConfigPath string
+	jobConfigPath  string
+
+	periodic string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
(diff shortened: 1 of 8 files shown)


cmd/cvp-trigger/main.go
```diff
@@ -0,0 +1,149 @@
...
+		fmt.Printf(string(jobAsYAML))
+		os.Exit(0)
+	}
+
+	logger := logrus.WithFields(pjutil.ProwJobFields(created))
+	logger.Info("submitted the prowjob, waiting for its result")
+
+	selector := fields.SelectorFromSet(map[string]string{"metadata.name": created.Name})
+
+	for {
```
stevekuznetsov: If you exit on error from the Watch why do you need this outer `for`?

cmd/cvp-trigger/main.go
```diff
@@ -0,0 +1,149 @@
...
+	jobConfigPath  string
+
+	periodic string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.BoolVar(&o.dry, "dry-run", true, "Whether to actually submit the job to Prow")
```
stevekuznetsov: nit: better UX is to make the user provide a positive, so make a user pass `--dry-run=true` for a program that defaults to doing something and `--confirm=true` for something that defaults to doing nothing

cmd/cvp-trigger/main.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	pjclient, err := rehearse.NewProwJobClient(clusterConfig, config.ProwJobNamespace, o.dry)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to create a ProwJob client")
+	}
+
+	prowjob := pjutil.NewProwJob(pjutil.PeriodicSpec(*selected), nil, nil)
+	logrus.WithFields(pjutil.ProwJobFields(&prowjob)).Info("submitting a new prowjob")
+	created, err := pjclient.Create(&prowjob)
```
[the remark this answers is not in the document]
petr-muller: Nope, `rehearse.NewProwJobClient` is smart and creates a fake client when `dry`. But you are right, I guess I could avoid even using that dependency and simply conditionally `Create`.

cmd/cvp-trigger/main.go
petr-muller: I followed what the rehearsals are doing, and I vaguely recollect the reason why we did that was that sometimes the result channel closed prematurely.

cmd/cvp-trigger/main.go
petr-muller: Yep, I had a real dry first, then changed it because I like the "safe by default" pattern. I'll change this to `--confirm` with a `false` default.

PR #543, opened 2020-03-10
make GetAllConfigs to return an error

droslean:
/cc @petr-muller


petr-muller:
/close

Discussed on slack, we don't want to do this. This design is intentional: it's used in rehearsals which needs to survive when Prow config is wrong in the baseline, so that PRs that fix the wrong baselines can be properly tested.

PR #545, opened 2020-03-11
ci-operator,prowgen: don't set redundant ci-operator flags

AlexNPavel:
This PR allows the `ci-operator` to use Prow's downward API to determine the `org`, `repo`, and `branch` that is passed to the `configresolver` when retrieving the config. It also changes the loading priority to the following:

1. If `--config` is set, use file.
2. If `CONFIG_SPEC` is set, use environment variable.
3. Use configresolver. Use `jobSpec.Refs` and `jobSpec.ExtraRefs` to determine `org`, `repo`, and `branch`. If `org`, `repo`, and/or `branch` are not available via the refs and extra_refs, use flags to get unset `org`, `repo`, and `branch`.

The variant for the resolver still gets passed via the `--variant` flag, and now the `--resolverAddress` defaults its value instead of being set by prowgen.

Prowgen has also been updated to not set flags for jobs other than `--variant`.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -354,7 +354,7 @@ func (o *options) Complete() error {
 	jobSpec.BaseNamespace = o.baseNamespace
 	o.jobSpec = jobSpec
 
-	info := getResolverInfo(jobSpec, o)
+	info := o.getResolverInfo(jobSpec)
 
 	config, err := load.Config(o.configSpecPath, o.registryPath, info)
 	if err != nil {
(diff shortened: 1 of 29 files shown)


cmd/ci-operator/main.go
```diff
@@ -1495,3 +1483,49 @@ func getPullSecretFromFile(filename string) (*coreapi.Secret, error) {
...
+		Repo:    refs.Repo,
+		Branch:  refs.BaseRef,
+		Variant: o.variant,
+	}
+	if len(info.Org) > 0 && len(info.Repo) > 0 && len(info.Branch) > 0 {
+		return info
+	}
+
+	// if org, repo, or branch could not be identified, try extra refs
+	for _, ref := range jobSpec.ExtraRefs {
```
alvaroaleman: This code allows `refs` to get filled from a mixture of different sources. IMHO if we filled any part of it from one source, we must return and then have a check if everything we need is filled, rather than starting to mix.

Also, please do empty checks for strings via `s == ""`, because that tells the reader of your code that this is a string.

cmd/ci-operator/main.go
```diff
@@ -1495,3 +1483,49 @@ func getPullSecretFromFile(filename string) (*coreapi.Secret, error) {
...
+		}
+		if len(info.Branch) == 0 {
+			info.Branch = ref.BaseRef
+		}
+		if len(info.Org) > 0 && len(info.Repo) > 0 && len(info.Branch) > 0 {
+			return info
+		}
+	}
+
+	// if org, repo, or branch could not be identified in any refs, try flags
```
bbguimaraes: This is the opposite of what I would expect, flags usually override whatever value would be used if they were not present.

cmd/ci-operator/main.go
```diff
@@ -147,6 +147,8 @@ const (
 	annotationIdleCleanupDurationTTL = "ci.openshift.io/ttl.soft"
 	// annotationCleanupDurationTTL is the annotation for requesting namespace cleanup after the namespace has been active
 	annotationCleanupDurationTTL = "ci.openshift.io/ttl.hard"
+	// configResolverAddress is the default configresolver address in api.ci
+	configResolverAddress = "http://ci-operator-configresolver-ci.svc.ci.openshift.org"
```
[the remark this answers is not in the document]
AlexNPavel: It changes the value from being hardcoded into prowgen's default job args into ci-operator's default flag value.

test/ci-operator-integration/multi-stage/expected/hyperkube.json
```diff
@@ -4,14 +4,14 @@
     "kind": "Build",
     "metadata": {
       "annotations": {
-        "ci.openshift.io/job-spec": "{\"type\":\"presubmit\",\"job\":\"pull-ci-openshift-release-master-ci-operator-integration\",\"buildid\":\"0\",\"prowjobid\":\"uuid\",\"refs\":{\"org\":\"openshift\",\"repo\":\"ci-tools\",\"base_ref\":\"master\",\"base_sha\":\"af8a90a2faf965eeda949dc1c607c48d3ffcda3e\",\"pulls\":[{\"number\":1234,\"author\":\"droslean\",\"sha\":\"538680dfd2f6cff3b3506c80ca182dcb0dd22a58\"}]}}"
+        "ci.openshift.io/job-spec": "{\"type\":\"presubmit\",\"job\":\"pull-ci-openshift-release-master-ci-operator-integration\",\"buildid\":\"0\",\"prowjobid\":\"uuid\",\"refs\":{\"org\":\"openshift\",\"repo\":\"installer\",\"base_ref\":\"release-4.2\",\"base_sha\":\"af8a90a2faf965eeda949dc1c607c48d3ffcda3e\",\"pulls\":[{\"number\":1234,\"author\":\"droslean\",\"sha\":\"538680dfd2f6cff3b3506c80ca182dcb0dd22a58\"}]}}"
```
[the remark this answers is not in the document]
AlexNPavel: Since we're using the `JOB_SPEC` to identify the org, repo, and branch that ci-operator uses for the configresolver, I updated the `JOB_SPEC` in the test to use the correct values for those in the refs.

cmd/ci-operator/main.go
```diff
@@ -1495,3 +1483,38 @@ func getPullSecretFromFile(filename string) (*coreapi.Secret, error) {
...
+func (o *options) getResolverInfo(jobSpec *api.JobSpec) *load.ResolverInfo {
+	// address and variant can only be set via options
+	info := &load.ResolverInfo{
+		Address: o.resolverAddress,
+		Variant: o.variant,
+	}
+
+	allRefs := jobSpec.ExtraRefs
+	if jobSpec.Refs != nil {
+		allRefs = append([]prowapi.Refs{*jobSpec.Refs}, allRefs...)
```
stevekuznetsov: Actually -- in the case of a generated periodic, `jobSpec.Refs` will be `nil` but `jobSpec.ExtraRefs` will not and we _do_ want to use the `jobSpec.ExtraRefs` there. Please fix in a follow-up

PR #554, opened 2020-03-13
ci-operator: log the configresolver fetch if we do it

stevekuznetsov:

/assign @AlexNPavel @bbguimaraes

### DIFF
--- a/pkg/load/load.go
+++ b/pkg/load/load.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
+	"log"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
```
 [INFO] Running test with ci-operator-configresolver
parse error: Invalid numeric literal at line 1, column 8
ERROR: ci-operator failed. 
```

WTF?

/retest

bbguimaraes:
> ```
>  [INFO] Running test with ci-operator-configresolver
> parse error: Invalid numeric literal at line 1, column 8
> ERROR: ci-operator failed. 
> ```

You're messing with the dry-run output, which we use in integration tests. I think `log` instead of `fmt` is what we use in all other places because it goes to `stderr`.

stevekuznetsov:
Fixed

PR #559, opened 2020-03-16
Populate testgrid dashboardtab description from k8s.io/description annotation on prowjob

bparees:


cmd/testgrid-config-generator/main.go
```diff
@@ -278,7 +281,8 @@ func main() {
 				}
 			}
 
-			informing.add(p.Name, daysOfResults)
+			fmt.Printf("period job annotations: %s: %v\n", p.Name, p.Annotations)
```
smarterclayton: drop this, I think we're good to go.  let's check with sam and Jessica though on which annotation we use (I think it's k8s.io, I want to be positive)

bparees: whoops, debug. removed.

bparees: @spadgett says they only look at `openshift.io/description` and that's on projects.

smarterclayton: hrm.

bparees:
@smarterclayton passing tests

PR #563, opened 2020-03-16
fix generator for testgrid file bug template

bparees:


bparees:
/hold cancel

@smarterclayton let's go ahead and merge this.  `<link>` doesn't work, but it sounds like there's no way to actually get the link to the job today unfortunately.  This fixes the rest of the BZ url at least.

bparees:
corresponding testgrid changes are here: https://github.com/kubernetes/test-infra/pull/16805

smarterclayton:
Cross link your change to test-infra here

bparees:
done

PR #567, opened 2020-03-17
keep the 'approve' value in the private plugins repo level

droslean:
considering https://github.com/openshift/release/pull/7712 merged, the fix seems to work.

`$ checkconfig --config-path ./core-services/prow/02_config/_config.yaml  --job-config-path ./ci-operator/jobs/  --plugin-config ./core-services/prow/02_config/_plugins.yaml  --strict --exclude-warning long-job-names --exclude-warning mismatched-tide-lenient`
```
{"component":"unset","file":"/home/nmoraitis/go/src/k8s.io/test-infra/prow/cmd/checkconfig/main.go:325","func":"main.main","level":"info","msg":"checkconfig passes without any error!","time":"2020-03-17T16:10:14+01:00"}
```

/cc @petr-muller 

### DIFF
--- a/cmd/private-prow-configs-mirror/main.go
+++ b/cmd/private-prow-configs-mirror/main.go
@@ -28,7 +28,8 @@ import (
 )
 
 const (
-	openshiftPrivOrg = "openshift-priv"
+	openshiftPrivOrg   = "openshift-priv"
+	approvePluginValue = "approve"
 )
 
 type options struct {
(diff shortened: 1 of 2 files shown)


petr-muller:
/hold 

I do not think this is an appropriate fix.  There is no reason why approve should be treated in some way special. The logic for computing which plugins need to be enabled where needs to be correct, then no hacks like this would be needed.

droslean:
> I do not think this is an appropriate fix. There is no reason why approve should be treated in some way special. The logic for computing which plugins need to be enabled where needs to be correct, then no hacks like this would be needed.

We already treat the `approve` in a special way. Either way the `plugins` values interact with the `tide.queries`. If we exclude the `appove` from the org wide level, we need to include it in every repo that needs it. 

This is what we are doing already for the public repos as well. If we want to take another approach we need also to update the [checkconfig](https://github.com/kubernetes/test-infra/tree/master/prow/cmd/checkconfig) tool as well.

PR #568, opened 2020-03-17
prowgen: add expose option to prowgen config (DPTP-972)

ray-harris:
https://issues.redhat.com/browse/DPTP-972

### DIFF
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -21,8 +21,15 @@ var ProwgenFile = ".config.prowgen"
 
 // Prowgen holds the information of the prowgen's configuration file.
 type Prowgen struct {
+	// Private indicates that the openshift-ci-operator's github credentials should
+	// be made available via kube secrets and that hidden: true should
+	// be added to the job to prevent the prow job's logs from being public.
 	Private bool `json:"private,omitempty"`
-	Expose  bool `json:"expose,omitempty"`
+	// Expose indicates that, when the Private field is true, that the github
+	// credentials should be made available, but "hidden: true" should not be
+	// added so that the job's logs will be publicly available. This field
+	// has no effect if Private is false.
+	Expose bool `json:"expose,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


pkg/config/load.go
```diff
@@ -22,6 +22,7 @@ var ProwgenFile = ".config.prowgen"
 // Prowgen holds the information of the prowgen's configuration file.
 type Prowgen struct {
 	Private bool `json:"private,omitempty"`
+	Expose  bool `json:"expose,omitempty"`
```
stevekuznetsov: Could we get some GoDoc for this? What does this feature do?

ray-harris: Added

pkg/config/load.go
```diff
@@ -21,7 +21,15 @@ var ProwgenFile = ".config.prowgen"
 
 // Prowgen holds the information of the prowgen's configuration file.
 type Prowgen struct {
+	// Private indicates that the openshift-ci-operator's github credentials should
```
stevekuznetsov: I think I'd word this slightly differently:

```go
// Private indicates that generated jobs should be marked as hidden 
// from display in deck and that they should mount appropriate git credentials
// to clone the repository under test
```

pkg/config/load.go
```diff
@@ -21,7 +21,15 @@ var ProwgenFile = ".config.prowgen"
...
+	// be made available via kube secrets and that hidden: true should
+	// be added to the job to prevent the prow job's logs from being public.
 	Private bool `json:"private,omitempty"`
+	// Expose indicates that, when the Private field is true, that the github
```
stevekuznetsov:
```go
// Expose declares that jobs should not be hidden from view if they are private.
// This field has no effect if private is not set.
```

pkg/config/load.go
ray-harris: Done

pkg/config/load.go
ray-harris: Done

PR #569, opened 2020-03-17
[private-prow-configs-mirror] Fix the private plugins generation 

droslean:
/cc @petr-muller 

This PR refactors the implementation of generating the plugins for the private repos.
Added unit tests for the get common values function.

This implementation includes also the empty public repos that inherit their values from the org wide level config.

### DIFF
--- a/cmd/private-prow-configs-mirror/main.go
+++ b/cmd/private-prow-configs-mirror/main.go
@@ -290,73 +290,51 @@ func injectPrivateBugzillaPlugin(bugzillaPlugins plugins.Bugzilla, orgRepos orgR
 }
 
 func injectPrivatePlugins(plugins map[string][]string, orgRepos orgReposWithOfficialImages) {
-	logrus.Info("Processing...")
+	privateRepoPlugins := make(map[string][]string)
+	for org, repos := range orgRepos {
 
-	var allPrivate []string
+		for repo := range repos {
+			values := sets.NewString()
+			values.Insert(plugins[org]...)
 
-	privateReposWithValues := make(map[string][]string)
-
-	for orgRepo, value := range plugins {
-		if orgRepos.isOfficialRepoFull(orgRepo) {
-			newValue := value
-
-			// We want to append the org level values, because the private repo
-			// can exist in different org with different configuration.
-			org := strings.Split(orgRepo, "/")[0]
-			if _, ok := plugins[org]; ok {
-				newValue = append(newValue, plugins[org]...)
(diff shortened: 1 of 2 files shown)


cmd/private-prow-configs-mirror/main.go
```diff
@@ -290,73 +290,55 @@ func injectPrivateBugzillaPlugin(bugzillaPlugins plugins.Bugzilla, orgRepos orgR
 }
 
 func injectPrivatePlugins(plugins map[string][]string, orgRepos orgReposWithOfficialImages) {
-	logrus.Info("Processing...")
-
-	var allPrivate []string
-
 	privateReposWithValues := make(map[string][]string)
```
petr-muller: There's no reason to talk about "values" here, we are dealing with "plugins" here, so let's use that term everywhere (`privateRepoPlugins` instead of `privateReposWithValues`, `commonPlugins` instead of `commonValuesSet`, `getCommonPlugins` instead of `getCommonValues` etc).

cmd/private-prow-configs-mirror/main.go
```diff
@@ -290,73 +290,55 @@ func injectPrivateBugzillaPlugin(bugzillaPlugins plugins.Bugzilla, orgRepos orgR
...
+			count++
+			continue
 		}
+
+		ret = ret.Intersection(valuesSet)
+		count++
 	}
 
-	return repeated
+	return ret
```
petr-muller: This can be IMO simplified as:

```
	var common sets.String
	for _, v := range privateReposWithValues {
		vs := sets.NewString(v...)
		if common == nil {
		 	common = vs
		} else {
			common = common.Intersection(vs)
		}
	}
	
	return common
```

droslean: I had it like this but the problem with this approach is that if the intersection becomes nil in the middle of the iteration we will get the current values instead. That's why I used the counter instead.

petr-muller: How can the intersection become nil in the middle? Intersection seems to returns an empty set (=empty map), not nil when the intersection is empty....

petr-muller: I was curious so I tested, and it seems to behave correctly: https://play.golang.org/p/c0QmcLhxKk4

droslean: I will address this. The nil check will just do the trick
Thanks.

PR #570, opened 2020-03-17
Applyconfig: Allow passing multiple config dirs

alvaroaleman:
Needed so we do not have to duplicate everything for app.ci.

/assign @petr-muller @hongkailiu

### DIFF
--- a/cmd/applyconfig/applyconfig.go
+++ b/cmd/applyconfig/applyconfig.go
@@ -17,6 +17,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/logrusutil"
 )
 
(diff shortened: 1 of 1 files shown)


cmd/applyconfig/applyconfig.go
```diff
@@ -302,7 +303,7 @@ func init() {
 
 func main() {
 	o := gatherOptions()
-	var adminErr, standardErr error
+	var adminErrs, standardErrs []error
```
bbguimaraes: These don't seem to be used directly, they could be replaced by a single `bool`.

cmd/applyconfig/applyconfig.go
```diff
@@ -316,9 +317,11 @@ func main() {
 			return apply(o.kubeConfig, o.context, path, o.user.val, !o.confirm)
 		}
 
-		adminErr = applyConfig(o.directory, "admin", f)
-		if adminErr != nil {
-			logrus.Errorf("There were failures while applying admin config")
+		for _, dir := range o.directories.Strings() {
+			if err := applyConfig(dir, "admin", f); err != nil {
+				adminErrs = append(adminErrs, err)
+				logrus.WithError(err).Error("There were failures while applying admin config")
```
bbguimaraes: Do the errors returned by `applyConfig` have enough context to determine which file/directory? Would including `dir` in the log be helpful?

cmd/applyconfig/applyconfig.go
alvaroaleman: Done

cmd/applyconfig/applyconfig.go
alvaroaleman: Yeah, when we walk the dir we log each folder and we also log the `oc` command, so I think that should be enough

PR #571, opened 2020-03-18
Autoowners: Default Reviewers to Approvers

alvaroaleman:
This is done to make sure there is never an empty Reviewers list, as
that makes blunderbuss traverse up until it finds reviewers rather than
defaulting to the approvers.

/assign @petr-muller


alvaroaleman:
@petr-muller ping

PR #577, opened 2020-03-19
Determinize job config: Set api.ci cluster if cluster is unset

alvaroaleman:
I do realize that mutating the jobs means the tool does more than what would expect it to do, but its the one thing we run over all jobs. Maybe rename it?

/assign @petr-muller

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,9 @@ format:
 	gofmt -s -w $(shell go list -f '{{ .Dir }}' ./... )
 .PHONY: format
 
+update-integration:
+	UPDATE=true make integration
+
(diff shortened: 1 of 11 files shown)


petr-muller:
/test all

Yeah, I guess we can re-purpose this tool to "make jobs right". I'm not sure about a good name for it, though. Any ideas?

alvaroaleman:
> Yeah, I guess we can re-purpose this tool to "make jobs right". I'm not sure about a good name for it, though. Any ideas?

What about `sanitize-prow-jobs`?

Would you be okay if I do that in follow-ups because it will be a bit fiddly to do in a way that doesn't break o/release (I guess we termporarily have to publish it under two names)?

alvaroaleman:
/retest
works on my machine?

alvaroaleman:
Needs https://github.com/openshift/release/pull/7815 so we actually have an `api.ci` cluster.
/hold

PR #578, opened 2020-03-19
cvp-trigger: describe interface to common CVP jobs

petr-muller:
/cc @stevekuznetsov @dirgim @robnester-rh 

Please note that the mentioned job is still in a pending PR https://github.com/openshift/release/pull/7710, but this describes my current intentions.

### DIFF
--- a/cmd/cvp-trigger/README.md
+++ b/cmd/cvp-trigger/README.md
@@ -1,4 +1,28 @@
 # CVP Trigger
 
 CVP Trigger tool will be used by the CVP pipeline to parametrize and trigger
-the verification jobs for optional operator artifacts built internally in RH. 
+the verification jobs for optional operator artifacts built internally in RH.
+
+## Triggered Job Interface
+
+The tool should work by loading Prow and Job configuration, finding specific job
+configuration structs and modifying those structs so that their specified
+containers are executed with the specified environmental variables having the
+desired values.
+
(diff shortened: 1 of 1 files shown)


cmd/cvp-trigger/README.md
```diff
@@ -1,4 +1,28 @@
...
+`cpaas-cvp-optional-operator-common-tests` Prowjob and modify its
+`.Spec.Containers[0].Env` list to have the following environmental variables set:
+
+- `OCP_VERSION`: Required. A string specifying the desired minor OCP version
+  which will be used when provisioning the ephemeral testing cluster. If
+  `RELEASE_IMAGE_LATEST` is not specified, an ephemeral release payload will be
+  created, corresponding to the collection of OCP images built from HEADs of
+  appropriate release branches of OCP components (that is, latest development
+  version of that minor release).
+- `CLUSTER_TYPE`: Required. One of `aws` strings, specifying the cloud platform
```
robnester-rh: Just wondering, does this mean that the `CLUSTER_TYPE` should be validated that it starts with 'aws'?

petr-muller: Nope, right now it needs to be just `aws` exactly, but as I add more glue for other cloud platforms (it's short-term), additional values like `gcp` and `azure` will become permissible.

PR #580, opened 2020-03-20
ci-operator: identify ci-operator specific workloads

stevekuznetsov:
Users may want to detect when their workload is being executed by
ci-operator. Today, they can tell that they are executing in Prow by
checking $CI. This change sets $OPENSHIFT_CI so that they can
furthermore detect that they are running under ci-operator specifically.


stevekuznetsov:
WTF, go 1.13.5 locally and neither `go fmt` nor `go test` fail locally

PR #582, opened 2020-03-20
webreg: change layout of workflow table to better fit step names

AlexNPavel:
Due to the long names of some steps, the workflow table on the main page for the step registry web UI became very poorly formatted. This PR combines the `Pre`, `Test`, and `Post` columns into a single `Steps` column to alleviate the issue.


AlexNPavel:
A running version of this PR will be available here when it finishes building: https://server-ci-tools-582.svc.ci.openshift.org/

PR #584, opened 2020-03-20
hack/pr-deploy: build images faster

AlexNPavel:
`pr-deploy` is useful to generate an example of changes to webreg, but the builds are very slow. Part of this is because it is running `make` install to build the binaries, which builds all of the ci-tools binaries instead of just the resolver. This is unnecessary for the `pr-deploy` command, so this changes that to `go install ./cmd/ci-operator-configresolver/...`

/cc @stevekuznetsov

### DIFF
--- a/hack/pr-deploy.yaml
+++ b/hack/pr-deploy.yaml
@@ -92,7 +92,7 @@ objects:
         FROM api.ci.openshift.org/openshift/release:golang-1.13
 
         COPY . .
-        RUN make install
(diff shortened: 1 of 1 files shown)


AlexNPavel:
I ran `pr-deploy` for this PR to compare it to the last 2 PRs I just made.
PR 582 build time for `binaries`: 11m30s
PR 583 build time for `binaries`: 9m47s
This PR build time for `binaries`: 2m12s

PR #587, opened 2020-03-21
add libvirt template resources to deprovision script

sallyom:
Adds teardown steps for nested-libvirt-in-gcp clusters as a backup if ever the teardown container fails to run.


sallyom:
AFAICT this is no longer needed, closing

PR #591, opened 2020-03-24
Autoconfigbrancher: Fix determinize-prow-job call arguments

alvaroaleman:

### DIFF
--- a/cmd/determinize-prow-jobs/main.go
+++ b/cmd/determinize-prow-jobs/main.go
@@ -34,10 +34,20 @@ func bindOptions(flag *flag.FlagSet) *options {
 
 func determinizeJobs(prowJobConfigDir string) error {
 	errChan := make(chan error)
+
+	var errs []error
+	errReadingDone := make(chan struct{})
+	go func() {
+		for err := range errChan {
+			errs = append(errs, err)
+		}
+		close(errReadingDone)
+	}()
+
 	wg := sync.WaitGroup{}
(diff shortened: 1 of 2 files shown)


cmd/determinize-prow-jobs/main.go
```diff
@@ -34,10 +34,20 @@ func bindOptions(flag *flag.FlagSet) *options {
 
 func determinizeJobs(prowJobConfigDir string) error {
 	errChan := make(chan error)
+
+	var errs []error
```
alvaroaleman: The previous positioning of this created a deadlock, because the channel is unbuffered and we only started reading from it (thus freeing space) after we walked the filepath, but that doesn't continue when we are blocked writing to the channel in the syncrhonous part

cmd/determinize-prow-jobs/main.go
```diff
@@ -34,10 +34,20 @@ func bindOptions(flag *flag.FlagSet) *options {
...
+			errs = append(errs, err)
+		}
+		close(errReadingDone)
+	}()
+
 	wg := sync.WaitGroup{}
 	if err := filepath.Walk(prowJobConfigDir, func(path string, info os.FileInfo, err error) error {
 		if err != nil {
-			fmt.Fprintf(os.Stderr, "Failed to walk file/directory '%s'", path)
+			errChan <- fmt.Errorf("Failed to walk file/directory '%s'", path)
```
alvaroaleman: This meant that the tool exited with rc0 if we had an error here

PR #593, opened 2020-03-24
ci-operator-checkconfig: disallow duplicate promotions

stevekuznetsov:
It should never be possible for separate source configurations to
publish the same image tag.

### DIFF
--- a/cmd/ci-operator-checkconfig/main.go
+++ b/cmd/ci-operator-checkconfig/main.go
@@ -3,9 +3,14 @@ package main
 import (
 	"flag"
 	"fmt"
+	"os"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
-	"os"
+	"github.com/openshift/ci-tools/pkg/steps/release"
 )
 
 func main() {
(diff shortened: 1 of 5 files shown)


cmd/ci-operator-checkconfig/main.go
```diff
@@ -18,11 +22,47 @@ func main() {
...
+				}
+				formatted = append(formatted, identifier)
+			}
+			dupes = append(dupes, fmt.Errorf("output tag %s/%s:%s is promoted from more than one place: %v", tag.Namespace, tag.Name, tag.Tag, strings.Join(formatted, ", ")))
+		}
+	}
+	if len(dupes) > 0 {
+		fmt.Println("non-unique image publication found: ")
+		for _, dupe := range dupes {
+			fmt.Println(dupe)
```
alvaroaleman: Maybe prefix with `ERROR: ` so we get highlighting?

cmd/ci-operator-checkconfig/main.go
```diff
@@ -18,11 +22,47 @@ func main() {
 		os.Exit(1)
 	}
 
+	seen := map[api.ImageStreamTagReference][]*config.Info{}
 	if err := config.OperateOnCIOperatorConfigDir(configDir, func(configuration *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
 		// validation is implicit, so we don't need to do anything
+		// but record the images we saw for future validation
+		if configuration.PromotionConfiguration == nil {
```
alvaroaleman: I think we want to keep this functionality, so it would be great to have a basic test for it :)

stevekuznetsov: As a unit test for the function passed to `OperateOn..`?

alvaroaleman: Well, then we only test half of it. Wouldn't it make sense to factor out the whole of it so we can actually test it?

cmd/ci-operator-checkconfig/main.go
```diff
@@ -18,11 +22,47 @@ func main() {
...
+			return nil
+		}
+		tags, _ := release.ToPromote(*configuration.PromotionConfiguration, configuration.Images, sets.NewString())
```
smarterclayton: Also need to include all RawSteps.  And there may be rename shenanigans in RawSteps that relabel images (like artifacts) that we probably need to negative test for (are you promoting something that doesn't exist).

I've caught people promoting `src` via `additionalImages` - that was never really intended as a use case.  We can allow it for now, but in the future I reserve the right to slap hands.

pkg/defaults/defaults.go
```diff
@@ -2,6 +2,7 @@ package defaults
 
 import (
 	"fmt"
+	"k8s.io/apimachinery/pkg/util/sets"
```
petr-muller: nit: third-party import

pkg/steps/release/promote.go
```diff
@@ -173,6 +157,47 @@ func (s *promotionStep) Run(ctx context.Context, dry bool) error {
...
+	for dst, src := range config.AdditionalImages {
+		tagsByDst[dst] = src
+		names.Insert(dst)
+	}
+
+	if config.NamePrefix == "" {
+		return tagsByDst, names
+	}
+
+	namesByDst := map[string]string{}
```
petr-muller: nit: on L162, you `make`, here you `{}`  - make it consistent

cmd/ci-operator-checkconfig/main.go
petr-muller: I've seen people promote `src` this way in tests-like repositories to allow other repos to use some kind of shared tests where a build was not necessary... I thought that was exactly the intended use case? At least that's what we recommended to people in the past.

cmd/ci-operator-checkconfig/main.go
petr-muller: Personally I would be fine with unit-testing the function passed to `OperateOn`.

bbguimaraes: Agreed with @alvaroaleman, factoring out seems like the best option here.

cmd/ci-operator-checkconfig/main.go
stevekuznetsov: @smarterclayton why do we need to check `RawSteps`? For what? Promotion will not promote them unless it's identified using the `ToPromote` logic, so...

stevekuznetsov:
Updated. I don't think the value of an integration test here is high (we're calling a library that is used in every tool, many of which already have integration tests), so I factored out the logic and unit-tested it. I don't touch Raw Steps, because there is no reason to.

pkg/steps/release/promote.go
stevekuznetsov: ah, other was preeexisting

cmd/ci-operator-checkconfig/main.go
smarterclayton: AdditionalImages can reference RawSteps output, so promote validation needs to check if the additional images actually exists (unless you think validation already covers that).

smarterclayton: As discussed on slack I'll add that validation next (additionalImages must reference a real name).

This is lGTM then

PR #598, opened 2020-03-25
webreg: number lines and make them linkable

AlexNPavel:
This PR adds line numbers printed next to the lines as well as lines that can be linked to and be highlighted.

The `chroma` package does not currently have support for clickable links which leads to some limitations.
- To link to a line, you manually append `line%d` to the address
- This doesn't work for the help pages, as all code blocks would have the same id prefixes. Clickable lines would solve this problem as we could make unique prefixes for each block.

/cc @stevekuznetsov


AlexNPavel:
Running example will be available here: https://server-ci-tools-598.svc.ci.openshift.org/

PR #600, opened 2020-03-25
Pass optional operator installation parameters to steps

petr-muller:
The parameters correspond to the inputs in:
https://steps.svc.ci.openshift.org/registry/optional-operators-subscribe

Right now, this only passes the same information passed into ci-operator
via environmental variables (similar how the `RELEASE_IMAGE_*` ones are
consumed). In the future, some images built by ci-operator itself will
be used to populate these parameters.

/assign  @stevekuznetsov

### DIFF
--- a/pkg/steps/optional_operators.go
+++ b/pkg/steps/optional_operators.go
@@ -1,0 +1,109 @@
+package steps
+
+import (
+	"fmt"
+	"strings"
+
+	coreapi "k8s.io/api/core/v1"
+)
+
+const (
+	ooIndex            = "OO_INDEX"
+	ooPackage          = "OO_PACKAGE"
+	ooChannel          = "OO_CHANNEL"
+	ooInstallNamespace = "OO_INSTALL_NAMESPACE"
+	ooTargetNamespaces = "OO_TARGET_NAMESPACES"
+)
+
+// optionalOperator is the information needed by the optional operator installation
+// steps to be able to subscribe and install an optional operator
+type optionalOperator struct {
+	// Index is the pullspec of an index image
+	Index string
+	// Package is the name of the operator package to be installed
+	Package string
+	// Channel is name of the operator channel to track
+	Channel string
+	// Namespace is the name of the namespace into which the operator and catalog
+	// will be installed (optional)
+	Namespace string
+	// TargetNamespaces is a list of namespaces the operator will target. If empty,
+	// all namespaces will be targeted.
+	TargetNamespaces []string
+}
+
+// getter is a subset of api.Parameters
+type getter interface {
+	Get(name string) (string, error)
+}
+
+// resolveOptionalOperator extracts an optionalOperator instance from the
+// given parameters. If no optional operator-related parameters are set, returns
+// nil
+func resolveOptionalOperator(parameters getter) (*optionalOperator, error) {
+	var err error
+	var oo optionalOperator
+
+	required := map[string]*string{
+		ooIndex:   &oo.Index,
+		ooPackage: &oo.Package,
+		ooChannel: &oo.Channel,
+	}
+
+	for param, valuePointer := range required {
+		value, err := parameters.Get(param)
+		if err != nil {
+			return nil, err
+		}
+		*valuePointer = value
+	}
+
+	if oo.Namespace, err = parameters.Get(ooInstallNamespace); err != nil {
+		oo.Namespace = ""
+	}
+	if targetNamespaces, err := parameters.Get(ooTargetNamespaces); err != nil || targetNamespaces == "" {
+		oo.TargetNamespaces = nil
+	} else {
+		oo.TargetNamespaces = strings.Split(targetNamespaces, ",")
+	}
+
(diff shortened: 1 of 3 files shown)


pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,107 @@
...
+	ooPackage          = "OO_PACKAGE"
+	ooChannel          = "OO_CHANNEL"
+	ooInstallNamespace = "OO_INSTALL_NAMESPACE"
+	ooTargetNamespaces = "OO_TARGET_NAMESPACES"
+)
+
+// optionalOperator is the information needed by the optional operator installation
+// steps to be able to subscribe and install an optional operator
+type optionalOperator struct {
+	// The pullspec of an index image
```
stevekuznetsov: Nit: `godoc` expects this to be 

```go
// Index is the pullspec...
```

pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,107 @@
...
+	Index string
+	// The name of the operator package to be installed
+	Pkg string
+	// The name of the operator channel to track
+	Channel string
+	// The namespace into which the operator and catalog will be installed (optional)
+	Namespace string
+	// A comma-separated list of namespaces the operator will target. If empty,
```
stevekuznetsov: Comma-separated? This is a slice?

pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,107 @@
...
+	ooTargetNamespaces = "OO_TARGET_NAMESPACES"
+)
+
+// optionalOperator is the information needed by the optional operator installation
+// steps to be able to subscribe and install an optional operator
+type optionalOperator struct {
+	// The pullspec of an index image
+	Index string
+	// The name of the operator package to be installed
+	Pkg string
```
stevekuznetsov: Please add vowels

pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,107 @@
...
+		oo.Namespace = ""
+	}
+	if targetNamespaces, err := parameters.Get(ooTargetNamespaces); err != nil || targetNamespaces == "" {
+		oo.TargetNamespaces = nil
+	} else {
+		oo.TargetNamespaces = strings.Split(targetNamespaces, ",")
+	}
+
+	if oo.Index != "" || oo.Pkg != "" || oo.Channel != "" || oo.Namespace != "" || len(oo.TargetNamespaces) > 0 {
+		if oo.Index == "" {
```
stevekuznetsov:
```go
for param, value := range map[string]string{
  ooIndex: oo.Index,
  ooPackage: oo.Pkg,
  ooChannel: oo.Channel,
} {
    if value == "" {
        return  nil, fmt.Errorf("at least one of optional operator parameters OO_* is set, but not the required parameter %s", parameter)
    }
}
```

pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,107 @@
...
+			return nil, fmt.Errorf("at least one of optional operator paramaters OO_* is set, but not the required parameter %s", ooIndex)
```
stevekuznetsov: sp: paramaters

pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,107 @@
...
+	Get(name string) (string, error)
+}
+
+// resolveOptionalOperator extracts an optionalOperator instance from the
+// given parameters
+func resolveOptionalOperator(parameters getter) (*optionalOperator, error) {
+	var err error
+	var oo optionalOperator
+
+	if oo.Index, err = parameters.Get(ooIndex); err != nil {
```
stevekuznetsov:
```go
for param, valuePointer := range map[string]*string{
  ooIndex: &oo.Index,
  ooPackage: &oo.Pkg,
  ooChannel: &oo.Channel,
} {
    if value, err := parameters.Get(parameter); err != nil {
        return nil, err
    }
    valuePointer = value
}
```

?

pkg/steps/optional_operators.go
petr-muller: Too much copy pasta from the step

pkg/steps/optional_operators.go
petr-muller: This one feels like borderline too smart for what it really solves, but after some deliberation I did it your way :)

petr-muller:
@stevekuznetsov thanks for review, addressed!

pkg/steps/optional_operators.go
```diff
@@ -0,0 +1,111 @@
...
+		oo.Namespace = ""
+	}
+	if targetNamespaces, err := parameters.Get(ooTargetNamespaces); err != nil || targetNamespaces == "" {
+		oo.TargetNamespaces = nil
+	} else {
+		oo.TargetNamespaces = strings.Split(targetNamespaces, ",")
+	}
+
+	if oo.Index != "" || oo.Package != "" || oo.Channel != "" || oo.Namespace != "" || len(oo.TargetNamespaces) > 0 {
+		for param, value := range map[string]string{
```
stevekuznetsov: Couldn't we collapse the two maps into one, declare it as a var at the top and re-use here?

petr-muller: Done.

PR #606, opened 2020-03-26
cvp-trigger: complete cvp-trigger functionality

robnester-rh:


cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,165 @@ import (
...
 }
 
+// getClusterConfig returns a config from calling util.LoadClusterConfig()
+func getClusterConfig() (*rest.Config, error) {
+	var clusterConfig *rest.Config
+	clusterConfig, err := util.LoadClusterConfig()
+	if err != nil {
+		return nil, err
+	}
+	return clusterConfig, nil
```
petr-muller: simple `return clusterConfig, err` is equivalent, nothing should rely on the first returned value when err is not `nil` anyway

robnester-rh: I usually return `nil` just to make sure I don't cheat and that it's painfully obvious that we're in an error situation. I can amend to just do a `return clusterConfig, err` if you would prefer, however.

stevekuznetsov: IMO this should just be `util.LoadClusterConfig()` -- the wrapper adds no utility, calling code that attempts to use the config if `err != nil` is malformed

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,165 @@ import (
...
+	var selectedJob *prowconfig.Periodic
+	for _, job := range config.AllPeriodics() {
+		if job.Name == jobName {
+			selectedJob = &job
+			break
+		}
+	}
+
+	if selectedJob == nil {
+		return nil, fmt.Errorf("failed to find the job")
```
stevekuznetsov: You likely want the name of the job in this error message

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,165 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"sort"
+	"strings"
 
 	"github.com/ghodss/yaml"
+	"github.com/openshift/ci-tools/pkg/util"
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
```
stevekuznetsov: What problem is this solving?

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,165 @@ import (
...
+		return nil, fmt.Errorf("failed to find the job")
+	}
+
+	prowjob := pjutil.NewProwJob(pjutil.PeriodicSpec(*selectedJob), nil, nil)
+	return &prowjob, nil
+}
+
+// addEnvVarsToJobSpec adds the given key/value sets as EnvVars in the Env var list of the
+// .Spec.PodSpec.Containers[0].Env
+func addEnvVarsToJobSpec(job *pjapi.ProwJob, envVars map[string]string) {
```
stevekuznetsov: nit: functions with side effects present challenges in control flow and testing. Consider a non-mutating approach.

test/cvp-trigger-integration/run.sh
```diff
@@ -5,23 +5,28 @@ set -o nounset
...
+expected="$datadir/$job_name.expected"
 
 update=false
 if [[ ${1:-} == "--update" ]]; then
   update=true
 fi
 
 echo "[INFO] Running CVP trigger"
-cvp-trigger --prow-config-path="$prow_config" --job-config-path="$job_config" --periodic="$job" >"$workdir/$job.output"
+cvp-trigger --bundle-image-ref="$bundle_image_ref" --index-image-ref="$index_image_ref" --prow-config-path="$prow_config_path" --job-config-path="$job_config_path" --job-name="$job_name" --ocp-version="$ocp_version" --package-name="$package_name" --channel="$channel" --dry-run>"$workdir/$job_name.output"
```
stevekuznetsov: nit: prefer `"${var}"` over `"$var"`

robnester-rh: Will update.

cmd/cvp-trigger/main.go
robnester-rh: I leveraged this because it allows for an interface which improves testing in my opinion. We can provide a `memMapOs` object in lieu of writing to the testing system's file-system and have greater flexibility for things such as validation for existence of paths, etc.

cmd/cvp-trigger/main.go
robnester-rh: Will update to include the job name in the error message, and call `util.LoadClusterConfig()` directly in main() instead.

cmd/cvp-trigger/main.go
robnester-rh: Would the preference be to do this in line in `main()` instead?

stevekuznetsov: Or return a copy of the job that has the appended variables. If there are no other callers, though, I generally prefer to inline work (unless it's split out specifically to allow for unit testing)

robnester-rh: The reason I'd split out the inline calls were mainly to allow for unit testing. It doesn't seem like there's much of a focus on code coverage in the cmds I see here (not intended as a dig, at all, just to be clear). I generally try to keep `main()` primarily limited to logging and calls to other functions. I'm content to follow any conventions / standards, however, as there's a lot of existing go here.

cmd/cvp-trigger/main.go
```diff
@@ -80,20 +201,33 @@ func main() {
...
+		logrus.WithField("job-name", o.jobName).Fatal(err)
 	}
 
-	if selected == nil {
-		logrus.WithField("job-name", o.periodic).Fatal("failed to find the job")
+	// Add flag values to inject as ENV var entries in the prowjob configuration
+	envVars := map[string]string{
+		"OCP_VERSION":  o.ocpVersion,
+		"CLUSTER_TYPE": "aws",
+		"OO_INDEX":     o.indexImageRef,
```
stevekuznetsov, line 213: These have constants in the repo, let's import and use them

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,190 @@ import (
...
 	}
+	sort.Strings(keys)
 
+	for _, k := range keys {
+		job.Spec.PodSpec.Containers[0].Env = append(job.Spec.PodSpec.Containers[0].Env, v1.EnvVar{Name: k, Value: envVars[k]})
+	}
+}
+
+// writeOutputFile writes the returned JSON to the specified file
+func writeOutputFile(json []byte, path string) error {
```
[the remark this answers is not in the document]
robnester-rh: ack. This was just a placeholder for writing some output and I'm not sure what the returned data will look like and whether it will need to be manipulated.

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,190 @@ import (
...
 
-	if o.periodic == "" {
-		return fmt.Errorf("required parameter %s was not provided", periodicOption)
+	for k := range envVars {
+		keys = append(keys, k)
 	}
+	sort.Strings(keys)
 
+	for _, k := range keys {
+		job.Spec.PodSpec.Containers[0].Env = append(job.Spec.PodSpec.Containers[0].Env, v1.EnvVar{Name: k, Value: envVars[k]})
```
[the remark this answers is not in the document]
robnester-rh: Thanks for the heads-up on that. I've removed this function and replaced it with a call to `decorate.KubeEnv()`

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,190 @@ import (
...
+	packageName      string
+	prowConfigPath   string
+	releaseImageRef  string
+	targetNamespaces string
+	dryRun           bool
+}
 
-	prowConfigPath string
-	jobConfigPath  string
+var fileSystem = afero.NewOsFs()
```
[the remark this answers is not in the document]
robnester-rh: `fileSystem` is available for substitution in testing. In our test we set `filesystem = afero.NewMemMapFs()` so that we're able to create arbitrary paths for path existence validations.

cmd/cvp-trigger/main.go
```diff
@@ -4,69 +4,190 @@ import (
...
+	_, err = f.Write(json)
+	_ = f.Sync()
 	return nil
 }
 
 func main() {
-	o := gatherOptions()
-	if err := o.validate(); err != nil {
+	o.gatherOptions()
+	_ = fs.Parse(os.Args[1:])
```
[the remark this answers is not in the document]
robnester-rh: Perhaps I'm mistaken, but since we're setting `var fs = flag.NewFlagSet(os.Args[0], flag.ExitOnError)` this automatically calls `os.Exit(2)` on error state. As such, there's no error to handle in this scenario.

cmd/cvp-trigger/main.go
robnester-rh: I see the following:
``` go
const (
	ooIndex            = "OO_INDEX"
	ooPackage          = "OO_PACKAGE"
	ooChannel          = "OO_CHANNEL"
	ooInstallNamespace = "OO_INSTALL_NAMESPACE"
	ooTargetNamespaces = "OO_TARGET_NAMESPACES"
)
```
 in `pkg/steps/optional_operators.go` but they are un-exported. 
Should I make a change to export these?

cmd/cvp-trigger/main.go
```diff
@@ -80,20 +201,33 @@ func main() {
...
+	// Add flag values to inject as ENV var entries in the prowjob configuration
+	envVars := map[string]string{
+		"OCP_VERSION":  o.ocpVersion,
+		"CLUSTER_TYPE": "aws",
+		"OO_INDEX":     o.indexImageRef,
+		"OO_PACKAGE":   o.packageName,
+		"OO_CHANNEL":   o.channel,
+	}
+	if o.releaseImageRef != "" {
+		envVars["RELEASE_IMAGE_LATEST"] = o.releaseImageRef
```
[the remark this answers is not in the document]
robnester-rh: I see the following:

``` golang
const (
	multiStageTestLabel     = "ci.openshift.io/multi-stage-test"
	clusterProfileMountPath = "/var/run/secrets/ci.openshift.io/cluster-profile"
	secretMountPath         = "/var/run/secrets/ci.openshift.io/multi-stage"
	secretMountEnv          = "SHARED_DIR"
	clusterProfileMountEnv  = "CLUSTER_PROFILE_DIR"
	initialReleaseEnv       = "RELEASE_IMAGE_INITIAL"
	latestReleaseEnv        = "RELEASE_IMAGE_LATEST"
)
```
in `pkg/steps/mutli-stage.go` but they are un-exported.
Should I make a change to export these?

cmd/cvp-trigger/main.go
petr-muller: I wouldn't mind exporting these, but I'm wondering if we should centralize all these env names which can be set from outside (these, `RELEASE_*` ones, I don't know if there are more) to a single place.

petr-muller:
The increments here are getting harder to review, perhaps we should wrap this up and build the still missing functionality in incremental, smaller PRs?

robnester-rh:
@petr-muller I'm fine with submitting this for "final" review and looking at the constant usage, which I think is the last component of the current implementation that would change, at a later point. If you think this is advisable, let me know and I'll remove the WIP label.
/cc @dirgim 
/cc @stevekuznetsov

stevekuznetsov:
@robnester-rh sorry -- missed your comment, will do a final pass

cmd/cvp-trigger/main.go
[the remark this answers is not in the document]
stevekuznetsov: Sure, but nothing is guaranteeing that someone does not change that call. It's not idiomatic to ignore errors and it's strictly more correct to handle it, so let's try to do that.

cmd/cvp-trigger/main.go
```diff
@@ -136,13 +262,57 @@ func main() {
 			if !ok {
 				logrus.WithField("object-type", fmt.Sprintf("%T", event.Object)).Fatal("received an unexpected object from Watch")
 			}
+
+			prowJobArtifactsURL = fmt.Sprintf("%s/%s/logs/%s/%s",
```
stevekuznetsov, line 266: This is not correct. Use `gcsupload.PathsForJob`

cmd/cvp-trigger/main.go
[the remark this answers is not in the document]
stevekuznetsov: Yes.

cmd/cvp-trigger/main.go
stevekuznetsov: Let's just export them for now.

cmd/cvp-trigger/main.go
```diff
@@ -1,72 +1,183 @@
...
+	Status       pjapi.ProwJobState `json:"status"`
+	ArtifactsURL string             `json:"prowjob_artifacts_url"`
+	URL          string             `json:"prowjob_url"`
 }
 
-func gatherOptions() options {
-	o := options{}
-	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+func (p *prowjobResult) toJSON() ([]byte, error) {
+	j, err := json.MarshalIndent(p, "", "    ")
```
stevekuznetsov, line 66: This method should just be `return json.MarshalIndent(...` and at that point I don't think there's any utility in a `toJSON` method at all

robnester-rh: Updated to `return json.MarshalIndent(...`
Method exists to allow for unit testing with a fake result.

cmd/cvp-trigger/main_test.go
```diff
@@ -1,56 +1,572 @@
...
+
+type fakeJobResult struct {
+	err error
+}
+
+func (v *fakeJobResult) toJSON() ([]byte, error) {
+	if v.err != nil {
+		return nil, v.err
+	}
+	return []byte(`null`), v.err
```
stevekuznetsov, line 23: Similarly, you could always just return this

cmd/cvp-trigger/main.go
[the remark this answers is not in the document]
robnester-rh: updated to handle error

cmd/cvp-trigger/main.go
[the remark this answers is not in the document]
robnester-rh: I've updated these to be exported

cmd/cvp-trigger/main.go
robnester-rh: I've updated these to be exported

cmd/cvp-trigger/main_test.go
```diff
@@ -1,56 +1,572 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := writeResultOutput(tt.args.prowjobResult, tt.args.outputPath); (err != nil) != tt.wantErr {
+				t.Errorf("writeResultOutput() error = %v, wantErr %v", err, tt.wantErr)
 			}
 		})
 	}
 }
+
+//func Test_prowjobResult_toJson(t *testing.T) {
```
stevekuznetsov, line 528: Let's not check in commented-out dead code, please

cmd/cvp-trigger/main.go
```diff
@@ -136,13 +265,64 @@ func main() {
 			if !ok {
 				logrus.WithField("object-type", fmt.Sprintf("%T", event.Object)).Fatal("received an unexpected object from Watch")
 			}
+			spec := downwardapi.NewJobSpec(prowJob.Spec, prowJob.Status.BuildID, prowJob.Name)
```
stevekuznetsov, line 268: Do we have a unit test for all this logic?

robnester-rh: We do now.

cmd/cvp-trigger/main_test.go
```diff
@@ -1,55 +1,653 @@
...
+	}
+	tests := []struct {
+		name string
+		args args
+		want string
+	}{
+		{
+			name: "Returns artifacts URL when we have .Spec.Ref",
+			args: args{
+				prowJob: &prowJobWithSpecRef,
```
stevekuznetsov, line 633: Why do we need named variables for `prowJobWithSpecRef` and `specWithRef`? Couldn't we just inline a `&ProwJob{}` literal in these test cases?

PR #608, opened 2020-03-26
ci-operator: expose which resource was leased to steps

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @bbgui

PR #609, opened 2020-03-26
webreg: document leases and quota

stevekuznetsov:

/assign @bbguimaraes

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -116,6 +116,7 @@ td {
         <div class="dropdown-menu" aria-labelledby="navbarDropdown">
           <a class="dropdown-item" href="/help">Getting Started</a>
           <a class="dropdown-item" href="/help/ci-operator">CI Operator Overview</a>
+          <a class="dropdown-item" href="/help/leases">Leases and Quota</a>
           <a class="dropdown-item" href="/help/adding-components">Adding and Changing Content</a>
           <a class="dropdown-item" href="/help/examples">Examples</a>
         </div>
@@ -814,27 +815,27 @@ The following environment variables will be available to commands in a step:
 	<th>When is it Present?</th>
   </tr>
   <tr>
-    <td style="white-space: nowrap"><code>${SHARED_DIR}<code></td>
+    <td style="white-space: nowrap"><code>${SHARED_DIR}</code></td>
 	<td>Directory on the step's filesystem where files shared between steps can be read.</td>
 	<td>Always.</td>
   </tr>
   <tr>
-    <td style="white-space: nowrap"><code>${CLUSTER_PROFILE_DIR}<code></td>
+    <td style="white-space: nowrap"><code>${CLUSTER_PROFILE_DIR}</code></td>
 	<td>Directory on the step's filesystem where credentials and configuration from the cluster profile are stored.</td>
 	<td>When the test as defined in a <code>ci-operator</code> configuration file sets a <code>cluster_profile</code>.</td>
   </tr>
   <tr>
-    <td style="white-space: nowrap"><code>${KUBECONFIG}<code></td>
+    <td style="white-space: nowrap"><code>${KUBECONFIG}</code></td>
 	<td>Path to <code>system:admin</code> credentials for the ephemeral OpenShift cluster under test.</td>
 	<td>After an ephemeral cluster has been installed.</td>
   </tr>
   <tr>
-    <td style="white-space: nowrap"><code>${RELEASE_IMAGE_INITIAL}<code></td>
+    <td style="white-space: nowrap"><code>${RELEASE_IMAGE_INITIAL}</code></td>
 	<td>Image pull specification for the initial release payload snapshot when the test began to run.</td>
 	<td>Always.</td>
   </tr>
   <tr>
-    <td style="white-space: nowrap"><code>${RELEASE_IMAGE_LATEST}<code></td>
+    <td style="white-space: nowrap"><code>${RELEASE_IMAGE_LATEST}</code></td>
 	<td>Image pull specification for the ephemeral release payload used to install the ephemeral OpenShift cluster.</td>
 	<td>Always.</td>
   </tr>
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Shown at https://server-ci-tools-609.svc.ci.openshift.org/help/leases

stevekuznetsov:
Depends on https://github.com/openshift/ci-tools/pull/608

stevekuznetsov:
/cc @openshift/openshift-team-developer-productivity-test-platform @bparees @jwforres @jupierce 

This is ready for review.

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+
+<p>
+The process for granting a <i>lease</i> on a <i>resource</i> from the available pool generally follows this workflow:
+</p>
+
+<ul>
+  <li>a client (<i>lessee</i>) requests a <i>lease</i> on an available <i>resource</i></li>
+  <li>the server (<i>lessor</i>) grants the <i>lease</i>, if possible, or places the client in a FIFO queue to wait for the next available <i>resource</i></li>
+  <li>the client emits a heartbeat while the <i>lease</i> is under active use</li>
+  <li>the client relinquishes the <i>lease</i> once the it is no longer in use</li>
```
jwforres: "once it is"

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+{{ yamlSyntax (index . "dynamicBoskosConfig") }}
+
+<p>
+If it is not clear exactly how many concurrent jobs can share the cloud provider at once, the convention is to set the
+<code>min-count</code> and <code>max-count</code> to <code>1000</code>, to effectively leave jobs unlimited and allow
+for investigation.
+</p>
+
+<p>
+In addition to registering the volume of concurrent jobs that are allowed to against a new cloud platform, it is required
```
jwforres: "are allowed against"

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
           memory: 2Gi
 `
 
+const quotasAndLeasesPage = `<h2 id="title"><a href="#title">How are Cloud Quota and Aggregate Concurrency Limits Handled?</a></h2>
+<p>
+Jobs that interact with an Infrastructure-as-a-Service (IaaS) cloud provider use credentials shared across the broader
```
alvaroaleman: I would reverse how you ordered the info in this passage and start it off with `In order to ensure that` to immediately clarify what this is doing. And generally, keep it shorter, something like `In order to remain within quota limits for cloud resources, we provide a shared locking system that limits the number of concurrent users for a given account`.

I only read this passage because I wanted to write that there is no high-level problem description and wanted to make sure I didn't miss it.

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+CI platform. Therefore, all jobs interacting with a specific IaaS will use API quota for these cloud providers from a
+shared pool. In order to ensure that our job throughput for a provider remains within the aggregate limit imposed by
+shared quota, jobs acquire leases for slices of the quota before they run and only relinquish them once all actions are
+completed. This document will describe the mechanism used to provide leases of quota slices to jobs, how jobs determine
```
alvaroaleman: `s/will describe/describes/`, always use present tense

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+which quota to ask for, how available leases can be configured and how current usage can be investigated.
```
alvaroaleman: `investigated` sounds like there is a lot of effort involved, how about `monitored`?

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+</p>
+
+<table class="table">
+  <tr>
+    <th style="white-space: nowrap">Term</th>
+    <th>Definition</th>
+  </tr>
+  <tr>
+    <td style="white-space: nowrap"><i>resource</i></td>
+    <td>An item which may be leased to clients. In the context of CI jobs, resources are slices of the larger cloud quota.</td>
```
alvaroaleman: I think the `In the context of CI jobs` part is irrelevant, this is not a document that describes a usecase other than CI

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+  </tr>
+  <tr>
+    <td style="white-space: nowrap"><i>lease</i></td>
+    <td>A binding between a <i>resource</i> and a client. When a lease is active, the underlying <i>resource</i> is not available for other clients.</td>
+  </tr>
+</table>
+
+<p>
+The process for granting a <i>lease</i> on a <i>resource</i> from the available pool generally follows this workflow:
```
alvaroaleman: I think it would be better to mark `lease` and `resource` rather than using italic

alvaroaleman: Also again, shorten: `The process for granting a lease follows the following workflow:`

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+</p>
+
+<ul>
+  <li>a client (<i>lessee</i>) requests a <i>lease</i> on an available <i>resource</i></li>
+  <li>the server (<i>lessor</i>) grants the <i>lease</i>, if possible, or places the client in a FIFO queue to wait for the next available <i>resource</i></li>
+  <li>the client emits a heartbeat while the <i>lease</i> is under active use</li>
+  <li>the client relinquishes the <i>lease</i> once the it is no longer in use</li>
+  <li>the server places the <i>resource</i> into the available pool for future clients to request</li>
```
alvaroaleman: places the resource _back_ into the available pool

bparees:
modulo @jwforres' corrections it looks ok to me.

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+</ul>
+
+<p>
+If a client fails to emit a heartbeat for long enough while the client holds a <i>lease</i>, the server will forcibly
+relinquish the <i>lease</i> and return the <i>resource</i> to the available pool for other clients. This mechanism
+ensures that clients which crash or otherwise fail to remain responsive cannot exhaust <i>resources</i> by holding a
+<i>lease</i> indefinitely.
```
alvaroaleman: unrelated to the doc, but can ci-op properly cope with the case that its resource is gone, e.G. because it got starved or sth? Or do we workaround that by having high timeouts?

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+</p>
+
+<h3 id="admins"><a href="#admins">Directions for Cloud Administrators</a></h3>
+<p>
+An administrator of a cloud platform will interact with the leasing server in order to configure the aggregate limit on 
+jobs for the platform or inspect the current settings and usage. Care must be taken when configuring the leasing server
+in order to ensure that jobs are well-behaved against the cloud provider APIs.
+</p>
+
+<h4 id="adding"><a href="#adding">Adding A New Type Of Cloud Quota</a></h4>
```
alvaroaleman: I would stick to the term `resource`, i.E. `Adding a new type of resource`

alvaroaleman: Also, the `A` should be lowercase

pkg/webreg/webreg.go
```diff
@@ -1228,6 +1229,138 @@ const imageExampleLiteral = `- as: repo-e2e
...
+for investigation.
+</p>
+
+<p>
+In addition to registering the volume of concurrent jobs that are allowed to against a new cloud platform, it is required
+that the leasing server is configured to reap leases which have not seen a recent heartbeat. This is done by adding the
+name of the resource type to the <a href="https://github.com/openshift/release/blob/master/core-services/prow/03_deployment/boskos_reaper.yaml#L27">reaper's configuration.</a>
+</p>
+
+<h5 id="static"><a href="#static">Configuration for Static, Limited Resources</a></h5>
```
alvaroaleman: Dynamic resources are just as limited. We should make very clear here when this is required and when not, I don't think it has much to do if we are working against a `large cloud api`. IIUC its only about needing specific names for each individual resources to derive some data for it.

`Configuration when using boskos leases to derive data from for your build` maybe?

pkg/webreg/webreg.go
stevekuznetsov: We've used italics for special words elsewhere, I prefer that to code since these are not code concepts/config

pkg/webreg/webreg.go
stevekuznetsov: So you mean the following case:

 - ci-op gets a lease
 - gets preempted by the kernel or something for 30 minutes
 - comes back and its lease has been reaped?

It would likely fail on the next heartbeat

stevekuznetsov:
Thanks all, updated

PR #610, opened 2020-03-27
Applyconfig: Log command

alvaroaleman:
Helps tremendously with debugging.

We set a censoring formatter for logrus in init, so this should be fine.


alvaroaleman:
Yes, the command is already being logged

PR #611, opened 2020-03-27
cvp-trigger: document the architecture of the OO tests

petr-muller:
Not sure if this is the best location for this documentation, but I haven't found any better. It describes how the `tool ↔ job ↔ ci-operator config ↔ workflow` are wired together, so that this knowledge is no longer only in my head. The implementation details can (and probably will) change while we are further iterating on this feature; I **commit** to maintain this section up-to-date until the feature settles.

I'd especially like the DPTP to review this so that I know if it makes sense to people who might need to build on this. Comments by CPaaS folks are also welcome though!

/cc @dirgim @robnester-rh


bbguimaraes:
Same.

PR #616, opened 2020-03-30
webreg: cross-reference lease doc in env doc

stevekuznetsov:

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -821,7 +821,12 @@ The following environment variables will be available to commands in a step:
   </tr>
   <tr>
     <td style="white-space: nowrap"><code>${SHARED_DIR}</code></td>
-    <td>Directory on the step's filesystem where files shared between steps can be read.</td>
+    <td>Directory on the step's filesystem where files shared between steps can be read and written.</td>
+    <td>Always.</td>
+  </tr>
+  <tr>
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
https://server-ci-tools-616.svc.ci.openshift.org/help#env

pkg/webreg/webreg.go
```diff
@@ -844,6 +849,11 @@ The following environment variables will be available to commands in a step:
     <td>Image pull specification for the ephemeral release payload used to install the ephemeral OpenShift cluster.</td>
     <td>Always.</td>
   </tr>
+  <tr>
+    <td style="white-space: nowrap"><code>${LEASED_RESOURCE}</code></td>
+    <td>The name of the <i>resource</i> leased to grant access to cloud quota. See <a href="./help/leases">the documentation</a>.</td>
+    <td>When the test requires a <i>lease</i>.</td>
```
alvaroaleman: I am always going to find the italic clunky to read, it looks like it was supposed to stress something when its supposed to indicate that we are talking about a specific thing we defined.

Its btw also not in line with what the openshift docs do, they do use code style for this situation, see e.G. here: https://docs.openshift.com/container-platform/4.3/operators/olm-adding-operators-to-cluster.html

stevekuznetsov: They seem to use bold, or code, or both. I'll sync with them on this one

PR #604, opened 2020-03-26
Check empty cluster in ci-op's config

hongkailiu:
https://issues.redhat.com/browse/DPTP-1010

/hold

require https://github.com/openshift/release/pull/7951

PR #622, opened 2020-03-30
Revert "Revert "Merge pull request #604 from hongkailiu/cluster_stanza""

hongkailiu:
It was reverted because the integration tests are broken when there is a kubeconfig in $HOME folder with api.ci context (our laptop env.) because it still loads the config. It works on CI because there is no such config in the pod.

We add a commit in this PR to fix that by not loading the config in $HOME folder.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -450,16 +450,21 @@ func (o *options) Complete() error {
 		o.templates = append(o.templates, template)
 	}
 
-	clusterConfig, err := util.LoadClusterConfig()
-	if err != nil {
-		return fmt.Errorf("failed to load cluster config: %v", err)
+	o.clusterConfig = &rest.Config{}
+	if !o.dry {
+
+		clusterConfig, err := util.LoadClusterConfig()
+		if err != nil {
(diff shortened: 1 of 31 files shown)


cmd/ci-operator/main.go
```diff
@@ -450,16 +450,21 @@ func (o *options) Complete() error {
 		o.templates = append(o.templates, template)
 	}
 
-	clusterConfig, err := util.LoadClusterConfig()
-	if err != nil {
-		return fmt.Errorf("failed to load cluster config: %v", err)
-	}
+	o.clusterConfig = &rest.Config{}
```
hongkailiu: I know in principle, it should be fine if we use `nil`.
However, we use this config to init clients.

https://github.com/openshift/ci-tools/blob/a0e763cba29a22c2b09fd51236c991a1fb2fbcb5/pkg/defaults/defaults.go#L68-L79

And we usually do not check if the client is nil before using it.
Really not easy with nil clusterConfig.

hongkailiu:
Run the integration tests 6 times ... all green

https://prow.svc.ci.openshift.org/pr-history/?org=openshift&repo=ci-tools&pr=622

PR #623, opened 2020-03-31
add variant prefix in secret names

droslean:

after #614 merged we change the value of the `JOB_NAME_SAFE` env var. Unfortunately, this is being used in the template as the name of the secret for mounting, but `ci-operator` creates the secret without the variant prefix.

This PR will resolve this issue. Related JIRA https://issues.redhat.com/browse/DPTP-1019

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -300,7 +300,13 @@ func generatePodSpecOthers(info *prowgenInfo, release string, test *cioperatorap
 	case cioperatorapi.ClusterProfileVSphere:
 		targetCloud = "vsphere"
 	}
-	clusterProfilePath := fmt.Sprintf("/usr/local/%s-cluster-profile", test.As)
+
+	clusterProfileName := fmt.Sprintf("%s-cluster-profile", test.As)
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -419,6 +419,10 @@ func (o *options) Complete() error {
 			return fmt.Errorf("failed to generate secret %s: %v", name, err)
 		}
 		secret.Name = name
+		if len(o.variant) > 0 {
+			secret.Name = fmt.Sprintf("%s-%s", o.variant, name)
+		}
```
petr-muller: This does not seem to be a correct thing to do - the `o.variant` is only used to specify which config to obtain from the configresolver, otherwise ci-operator has no concept of a "variant" internally, and I think we should keep it that way.

If something that ci-operator creates needs to match to whatever `JOB_NAME_SAFE` is, let's base its name directly on `JOB_NAME_SAFE`, not on something that affects it.

droslean: in this case, we don't need to update `ci-operator` at all but instead `prowgen` should add the correct secret name including the variant in the `--secret-dir` flag

PR #624, opened 2020-03-31
rehearse: infer ci-operator config from parameters when needed

petr-muller:
Improve pj-rehearse to try harder in cases when it needs to rehearse a
job (usually a periodic) which is not apparently tied to any specific
ci-operator config, but is configured to run `ci-operator` with explicit
`--org/--repo/--branch/--variant` options. We were already removing
these options, so we can as well save their values, and use them in case
when the normally used ones are not available (previously we would
simply fail with a "ci-operator config file --.yaml was not found"
message).

The method that removes the args from containers running `ci-operator`
was rewritten to better accomodate the need to save the values,
otherwise the changes are quite simple.

### DIFF
--- a/pkg/config/load_test.go
+++ b/pkg/config/load_test.go
@@ -85,6 +85,53 @@ func TestExtractRepoElementsFromPath(t *testing.T) {
 	}
 }
 
+func TestInfo_IsComplete(t *testing.T) {
+	testCases := []struct {
+		name        string
+		info        Info
+		expectError string
+	}{
+		{
+			name: "All required members -> complete",
+			info: Info{Org: "organization", Repo: "repository", Branch: "branch"},
+		},
+		{
+			name:        "Missing org -> incomplete",
+			info:        Info{Repo: "repository", Branch: "branch"},
+			expectError: "missing item: organization",
+		},
+		{
+			name:        "Missing repo -> incomplete",
+			info:        Info{Org: "organization", Branch: "branch"},
+			expectError: "missing item: repository",
+		},
+		{
+			name:        "Missing branch -> incomplete",
+			info:        Info{Org: "organization", Repo: "repository"},
+			expectError: "missing item: branch",
+		},
+		{
+			name:        "Everything missing -> incomplete",
+			expectError: "missing items: branch, organization, repository",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			err := tc.info.IsComplete()
+			if err == nil && tc.expectError != "" {
+				t.Errorf("%s: expected error '%s', got nil", tc.expectError, tc.name)
+			}
+			if err != nil {
+				if tc.expectError == "" {
+					t.Errorf("%s: unexpected error %s", tc.name, err.Error())
+				} else if err.Error() != tc.expectError {
+					t.Errorf("%s: expected error '%s', got '%s", tc.name, tc.expectError, err.Error())
+				}
+			}
+		})
+	}
+}
+
 func TestInfo_Basename(t *testing.T) {
 	testCases := []struct {
 		name     string
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -276,6 +276,9 @@ func inlineCiOpConfig(container *v1.Container, ciopConfigs config.ByFilename, re
 	}
 	// inline CONFIG_SPEC for all ci-operator jobs
 	if container.Command != nil && container.Command[0] == "ci-operator" {
+		if err := info.IsComplete(); err != nil {
+			return fmt.Errorf("could not infer which ci-operator config this job uses: %v", err)
+		}
 		filename := info.Basename()
 		loggers.Debug.WithField(logCiopConfigFile, filename).Debug("Rehearsal job uses ci-operator config ConfigMap, needed content will be inlined")
 		ciopConfig, ok := ciopConfigs[filename]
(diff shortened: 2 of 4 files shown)


pkg/config/load_test.go
```diff
@@ -85,6 +85,47 @@ func TestExtractRepoElementsFromPath(t *testing.T) {
...
+		{
+			name:     "Everything missing -> incomplete",
+			expected: false,
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			complete := tc.info.IsComplete()
+			if complete != tc.expected {
+				t.Errorf("%s: expected %t, got %t", tc.name, tc.expected, complete)
```
alvaroaleman: Nit: There is no need to include the tests name in the error, `go test` also does this properly when using subtests (`t.Run()`)

pkg/rehearse/jobs.go
```diff
@@ -276,6 +276,9 @@ func inlineCiOpConfig(container *v1.Container, ciopConfigs config.ByFilename, re
 	}
 	// inline CONFIG_SPEC for all ci-operator jobs
 	if container.Command != nil && container.Command[0] == "ci-operator" {
+		if !info.IsComplete() {
+			return fmt.Errorf("could not infer which ci-operator config this job uses")
```
alvaroaleman: In general, it feels like it would be using what is missing, not only that something is missing

pkg/rehearse/jobs.go
```diff
@@ -821,29 +832,51 @@ func (e *Executor) waitForJobs(jobs sets.String, selector string) (bool, error)
...
+func removeConfigResolverFlags(args []string) ([]string, config.Info) {
+	var usedConfig config.Info
+	toConfig := map[string]*string{
+		"org":     &usedConfig.Org,
+		"repo":    &usedConfig.Repo,
+		"branch":  &usedConfig.Branch,
+		"variant": &usedConfig.Variant,
+	}
+	var newArgs []string
+	for len(args) > 0 {
```
alvaroaleman: err...What about `for idx := range args { current := args[idx]` to allow understanding this in less than 30 seconds?

pkg/config/load_test.go
petr-muller: I'm doing this intentionally; I know the name is included but it's butchered, including it here keeps it always readable,

pkg/rehearse/jobs.go
petr-muller: The need to eat an additional arg for `--arg value` forms in the middle of the block makes simple iteration messy, you need to add hacks like `skipNext = true`, which I think result in even less readable code.

pkg/rehearse/jobs.go
petr-muller: Err, I'm not sure I get what this feedback suggests. Is the error message confusing? Or the logic? 

Adding this check actually prevents us from using something that is missing -- the `filename := info.Basename()` below returns a useless "--.yaml"-like filename.

alvaroaleman: Which of org, repo or branch is empty

pkg/rehearse/jobs.go
alvaroaleman: I think having code in a loop body that mutates the iterable that is being looped over is pretty unexpected and hence hard to read

stevekuznetsov: We will ask new hires to explain this method for their interview

petr-muller: It's a well-known pattern in parsing (this use case is similar), you consume a "stream" of tokens and process them... I think even most of actual parameter parsers work exactly like this (for exactly the same reason).

Do you think adding a comment that explains why a strange loop is used would help, or are you adamant in that the iteration should be your way?

pkg/rehearse/jobs.go
petr-muller: oh, so you suggest that the error should contain what is missing?

alvaroaleman: yes, exactly, I personaly would just make the `IsComplete` return an error that contains that info (or nil if everything exists)

pkg/rehearse/jobs.go
alvaroaleman: the consuming stream part yes, the manipulating stream during process no

petr-muller: But consuming is exactly what the loop does, it consumes items from the beginning of the slice until it is empty. Some iterations consume a single item, some consume two. There's no more manipulation than that.

petr-muller:
Addressed all feedback. I attempted to address Alvaro's concern about the `removeConfigResolverFlags` method by adding a comment and refactoring the code so it contains less (duplicate) syntax and more descriptive names.

pkg/config/load_test.go
```diff
@@ -85,6 +85,49 @@ func TestExtractRepoElementsFromPath(t *testing.T) {
 	}
 }
 
+func TestInfo_IsComplete(t *testing.T) {
```
stevekuznetsov: 50% of the code in `IsComplete()` is about forming the correct error message -- we should validate those

petr-muller: OK

pkg/config/load_test.go
```diff
@@ -85,6 +85,53 @@ func TestExtractRepoElementsFromPath(t *testing.T) {
...
+	testCases := []struct {
+		name        string
+		info        Info
+		expectError string
```
stevekuznetsov, line 92: nit: a little simpler would have just been `expectError error` and use `errors.New()` but OK

PR #625, opened 2020-03-31
rename ci-op-configs-mirror to ci-operator-config-mirror

droslean:


droslean:
It seems that https://github.com/openshift/release/pull/8041 should land first

PR #628, opened 2020-04-01
hack: import hack/lib, restructure integration tests

stevekuznetsov:
Our bash in this repository does not follow common conventions.
Furthermore, our integration tests often achieve common goals in
uncommon ways. This patch includes the `hack/lib` utilities written
originally for Origin in order to standardize how we work with shell in
this repository. Immediate benefits of this include:

 - standardized logging and artifact placement
 - no need to create or manage temporary directories in scripts
 - out-of-the-box support for jUnit from Go tests as well as shell
 - built-in support for updating golden files in integration tests
 - less overhead for writing integration suites


alvaroaleman:
/hold
```
make: *** No rule to make target `check-breaking-changes'.  Stop.
```

PR #629, opened 2020-04-02
rehearse: fix excessive rehearsals for variants

petr-muller:
The `Info.JobName()` method did not return a name including a variant
(although a variant is one of the member fields of the struct), which
made rehearsal mistakenly treat more jobs as affected by a variant
ci-operator config change. By not matching agains job name including a variant,
it incorrectly selected all non-variant jobs as affected.

Fix `JobName()` together with Prowgen, which was actually depending on
this incorrect behavior, and appended the variant itself. Write few new
GoDocs and tests.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -499,26 +499,21 @@ func generateJobBase(name, prefix string, info *prowgenInfo, label jc.ProwgenLab
 }
 
 func generatePresubmitForTest(name string, info *prowgenInfo, label jc.ProwgenLabel, podSpec *kubeapi.PodSpec, rehearsable bool, pathAlias *string) *prowconfig.Presubmit {
-	if len(info.Variant) > 0 {
-		name = fmt.Sprintf("%s-%s", info.Variant, name)
-	}
+	shortName := info.Info.TestName(name)
 	base := generateJobBase(name, jobconfig.PresubmitPrefix, info, label, podSpec, rehearsable, pathAlias)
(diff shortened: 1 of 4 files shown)


petr-muller:
This is the affected part in pj-rehearse:

https://github.com/openshift/ci-tools/blob/d1971b95442a592d9cc3cccaceb34efe8dec3274/pkg/diffs/diffs.go#L206-L212

petr-muller:
/hold

Dammit, looks like I missed a place that appends a variant by itself, and that also means we are missing a test.

petr-muller:
FUUUUU

petr-muller:
/meow

Finally ready for review

cmd/ci-operator-prowgen/main.go
```diff
@@ -499,26 +499,21 @@ func generateJobBase(name, prefix string, info *prowgenInfo, label jc.ProwgenLab
 }
 
 func generatePresubmitForTest(name string, info *prowgenInfo, label jc.ProwgenLabel, podSpec *kubeapi.PodSpec, rehearsable bool, pathAlias *string) *prowconfig.Presubmit {
-	if len(info.Variant) > 0 {
-		name = fmt.Sprintf("%s-%s", info.Variant, name)
-	}
+	shortName := info.Info.TestName(name)
```
stevekuznetsov, line 502: what's the difference?

stevekuznetsov: ah, ok, this should be a no-op

PR #630, opened 2020-04-03
Parralelize secret fetching in ci-secret-bootstrapper

alvaroaleman:

Went with the dumbest possible implementation of just using a locking and waitgroups.

### DIFF
--- a/cmd/ci-secret-bootstrap/main_test.go
+++ b/cmd/ci-secret-bootstrap/main_test.go
@@ -1011,7 +1011,7 @@ func TestConstructSecrets(t *testing.T) {
 					"a-id-3-2": "attachment-name-3-2-value",
 				},
 			),
-			expectedError: fmt.Errorf("failed to find field field-name-1 in item item-name-1"),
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -1089,14 +1089,27 @@ func TestConstructSecrets(t *testing.T) {
...
 		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			actual, acutalError := constructSecrets(tc.config, tc.bwClient)
-			equalError(t, tc.expectedError, acutalError)
+			actual, actualError := constructSecrets(tc.config, tc.bwClient)
+			equalError(t, tc.expectedError, actualError)
+			if actualError != nil {
```
alvaroaleman, line 1100: This PR introduces a change in behaviour in that we continue to try to get everything even when there was an error. I like it because it allows for quicker debugging when a config change introduces more than one issue

alvaroaleman:
> Wouldn't the even dumber thing to do here be to use the sync map?

Its more code, because then we have to type assert on every access and at the end do a loop that gets all objects from the sync map and makes a properly typed map out of that

PR #631, opened 2020-04-03
Simplify 2 commands in integration tests

hongkailiu:
```
#with -a, -r is there too.
man cp
       -a, --archive
              same as -dR --preserve=all

       -R, -r, --recursive
              copy directories recursively

# we redirect the output to /dev/null
# and thus should not matter much if or not we suppress the output
man pgrep
       -c, --count
              Suppress normal output; instead print a count of matching processes.  When count does not match anything, e.g.
              returns zero, the command will return non-zero value.

       -P, --parent ppid,...
              Only match processes whose parent process ID is listed.
```

The real reason that I want this PR: it is more mac-os friendly with it.

### DIFF
--- a/test/repo-init-integration/run.sh
+++ b/test/repo-init-integration/run.sh
@@ -13,7 +13,7 @@ trap 'rm -rf ${WORKDIR}' EXIT
 
 cd "$WORKDIR"
 # copy config to tmpdir to allow tester to modify config
-cp -ar "$ROOTDIR"/test/repo-init-integration/input/* .
+cp -a "$ROOTDIR"/test/repo-init-integration/input/* .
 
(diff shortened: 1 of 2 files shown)


hongkailiu:
/hold

watching the rehearsal

stevekuznetsov:
Does MacOS support `pgrep --parent`? I prefer that over `-P`. Please note #628 is in flight so let's try not to refactor too much

hongkailiu:
> Does MacOS support `pgrep --parent`? I prefer that over `-P`. Please note #628 is in flight so let's try not to refactor too much

Tried. Hit error.

/hold

Wait for https://github.com/openshift/ci-tools/pull/628 lands first.

PR #632, opened 2020-04-03
cmd/testgrid-config-generator: Add details to the README

wking:
The previous invocation example was not sufficient to explain how to adjust the job labels and controller configs to acheive a particular change to the TestGrid configuration.  Hopefully the explanation I'm adding allows folks to do that without having to work through the Go source.

### DIFF
--- a/cmd/testgrid-config-generator/README.md
+++ b/cmd/testgrid-config-generator/README.md
@@ -6,5 +6,28 @@ This utility updates the TestGrid configuration to include promotion gates and i
 $ testgrid-config-generator -testgrid-config path/to/k8s.io/test-infra/config/testgrids/openshift -release-config path/to/openshift/release/core-services/release-controller/_releases -prow-jobs-dir path/to/openshift/release/ci-operator/jobs
 ```
 
+This will walk `-release-config` looking for release files that end with `.json`.
+For each of those release files, it will:
+
+* Select a stream based on the `name`.
+    `name` that match `*.ci`, `*.nightly`, or `stable-4.*` are added to the `ocp` stream.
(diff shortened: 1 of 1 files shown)


bparees:
this looks reasonable in terms of summarizing the current state, but i would really like to see a goal state of:  testgrid dashboards are independent of release-controller behavior.  Or driven off a common set of fields.

if a job is annotated informing, it shows up in an informing dashboard
if a job is annotated blocking, it shows up in a blocking dashboard

(and ideally, those annotations also drive what shows up in release-controller)

@openshift/openshift-team-developer-productivity-test-platform

stevekuznetsov:
@bparees not sure what you mean -- do you want to move the source of truth from a `release-controller` config into an annotation on jobs? Why is that better?

wking:
> this looks reasonable in terms of summarizing the current state...

Can we land this as it stands then, and adjust the docs when we pivot the code?

bparees:
can you add a step about actually committing the testgrid repo changes and submitting the PR?

stevekuznetsov:
I was very close to automating that bit :(

bparees:
> @bparees not sure what you mean -- do you want to move the source of truth from a release-controller config into an annotation on jobs? Why is that better?

Yes.  Because what release controller cares about and what testgrid should expose are not necessarily directly correlated and should not have to be tied together.

wking:
> can you add a step about actually committing the testgrid repo changes and submitting the PR?

Done with d123a669 -> f53986ee.  Once @stevekuznetsov automates that, he can drop the line from the docs.

bparees:
/lgtm

moves the ball forward, clayton and I are discussing the grander restructuring of this config/logic.

PR #634, opened 2020-04-06
Parameterize autoowners

dhiller:
Add parameters for github (org, repo, team) and directories
(config subdirs, target subdirectory).

Fixes kubevirt/project-infra#406

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -29,7 +29,8 @@ const (
 	githubOrg   = "openshift"
 	githubRepo  = "release"
 	githubLogin = "openshift-bot"
-	githubTeam  = "openshift/openshift-team-developer-productivity-test-platform"
+
+	defaultPRAssignee = "openshift/openshift-team-developer-productivity-test-platform"
 
 	configSubDirs      = "jobs,config,templates"
 	targetSubDirectory = "ci-operator"
@@ -51,15 +52,14 @@ func (orgRepo orgRepo) repoString() string {
 	return fmt.Sprintf("%s/%s", orgRepo.Organization, orgRepo.Repository)
 }
 
-func loadRepos(configRootDir string, blacklist sets.String, sourceSubDirs []string, githubOrg string, githubRepo string) ([]orgRepo, error) {
+func loadRepos(configRootDir string, blacklist sets.String, configSubDirs []string, githubOrg string, githubRepo string) ([]orgRepo, error) {
 	var orgRepos []orgRepo
-	configSubDirectories := make([]string, 0, len(sourceSubDirs))
-	for _, sourceSubDir := range sourceSubDirs {
+	configSubDirectories := make([]string, 0, len(configSubDirs))
+	for _, sourceSubDir := range configSubDirs {
 		configSubDirectories = append(configSubDirectories, filepath.Join(configRootDir, sourceSubDir))
 	}
-	jobs := configSubDirectories[0]
 
-	orgDirs, err := ioutil.ReadDir(jobs)
+	orgDirs, err := ioutil.ReadDir(configSubDirectories[0])
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


cmd/autoowners/main.go
```diff
@@ -286,10 +299,15 @@ func parseOptions() options {
...
 	fs.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
+	fs.StringVar(&o.githubOrg, "github-org", githubOrg, "The GitHub org to use.")
+	fs.StringVar(&o.githubRepo, "github-repo", githubRepo, "The GitHub repo to use.")
+	fs.StringVar(&o.githubTeam, "github-team", githubTeam, "The GitHub team to use.")
 	fs.StringVar(&o.gitName, "git-name", "", "The name to use on the git commit. Requires --git-email. If not specified, uses the system default.")
 	fs.StringVar(&o.gitEmail, "git-email", "", "The email to use on the git commit. Requires --git-name. If not specified, uses the system default.")
 	fs.StringVar(&o.assign, "assign", githubTeam, "The github username or group name to assign the created pull request to.")
 	fs.StringVar(&o.targetDir, "target-dir", "", "The directory containing the target repo.")
+	fs.StringVar(&o.targetSubDirectory, "target-subdir", targetSubDirectory, "The sub-directory of the target repo where the configurations are stored.")
+	fs.StringVar(&o.configSubDirs, "config-subdirs", configSubDirs, "The comma-separated list of sub-directories where configuration is stored.")
```
stevekuznetsov, line 310: Use `test-infra/prow/flagutil/strings.go#Strings`

cmd/autoowners/main.go
```diff
@@ -286,10 +299,15 @@ func parseOptions() options {
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the pull request with github client")
 	fs.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
+	fs.StringVar(&o.githubOrg, "github-org", githubOrg, "The GitHub org to use.")
```
stevekuznetsov, line 302: remove `github-` prefixes

stevekuznetsov: What does it mean to "use" an org? The doc text could be improved

cmd/autoowners/main.go
```diff
@@ -48,12 +51,13 @@ func (orgRepo orgRepo) repoString() string {
...
 	var orgRepos []orgRepo
-	operatorRoot := filepath.Join(dir, "ci-operator")
-	jobs := filepath.Join(operatorRoot, "jobs")
-	config := filepath.Join(operatorRoot, "config")
-	templates := filepath.Join(operatorRoot, "templates")
+	configSubDirectories := make([]string, 0, len(sourceSubDirs))
+	for _, sourceSubDir := range sourceSubDirs {
+		configSubDirectories = append(configSubDirectories, filepath.Join(configRootDir, sourceSubDir))
+	}
+	jobs := configSubDirectories[0]
```
petr-muller, line 60: This looks fishy to me. It basically means that the first item in the list has some special meaning, but this is not documented anywhere and even if it were, I still think it's a smell.

dhiller:
@petr-muller valid point, this method should be tested for at least the two use cases (open shift/release vs. kubevirt/project-infra) it would then have. Will have a look tomorrow on what I can do about that.

dhiller:
@petr-muller I added tests for the two use cases, this covers at least the happy cases. WDYT?

cmd/autoowners/main.go
dhiller: Logically it has a meaning, as 
* if it's not present, an error is returned
* if the org and repo dirs are not present in this directory, it is skipped
Maybe we should just make that clear in the documentation that the first directory name in the `configSubDirs` list must contain all the org/repo dirs, or else it will be skipped

cmd/autoowners/main.go
```diff
@@ -390,7 +406,11 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to change to root dir")
 	}
 
-	if err := pullOwners(gc, o.targetDir, sets.NewString(o.blacklist.Strings()...)); err != nil {
+	var configSubDirectories = o.configSubDirs.Strings()
+	if len(o.configSubDirs.Strings()) == 0 {
```
petr-muller, line 410: nit: `if len(configSubDirectories)` is the same thing and IMO shows the flow better.

cmd/autoowners/main.go
```diff
@@ -390,7 +406,11 @@ func main() {
...
+		configSubDirectories = strings.Split(configSubDirs, ",")
+	}
+	if err := pullOwners(gc, filepath.Join(o.targetDir, o.targetSubDirectory), sets.NewString(o.blacklist.Strings()...), configSubDirectories, o.githubOrg, o.githubRepo); err != nil {
```
petr-muller, line 413: nit: this is becoming hard to read with this many params. maybe extract the `Join` and `NewString` to named variables so that the call itself reads simpler?

petr-muller:
And thank you for the test coverage Dan!

hongkailiu:
@petr-muller 

I should pay my debts of test coverage.
(It always got lost in my list when it works.)
Do you want me to create a card on this?

dhiller:
@petr-muller addressed review comments, PTAL! Thanks for the feedback!

PR #636, opened 2020-04-06
[private-org-peribolos-sync] Tool to sync the peribolos config for the private org

droslean:
This tool generates the `peribolos[org].Repos` for the given organization by listing all the repositories that promote official images plus the ones that are included in the `--whitelist` flag. 

This is a part of the workload described in https://issues.redhat.com/browse/DPTP-753

### DIFF
--- a/cmd/private-org-peribolos-sync/README.md
+++ b/cmd/private-org-peribolos-sync/README.md
@@ -1,0 +1,8 @@
+# Private org peribolos sync
+
+This tool generates a mapping table of peribolos repository configurations for the given private
+organization. 
+It walks through the release repository path, given by `--release-repo-path`, and detects which of the repositories
+are promoting official images. The repositories that are specified in `--include-repo`, will be included if they exist in the release repository's path as well. Furthermore, it will get the required information for each of them from GitHub
+and will generate its peribolos configuration.
+Finally, the tool will update the peribolos configuration given by `--peribolos-config`.
--- a/cmd/private-org-peribolos-sync/main.go
+++ b/cmd/private-org-peribolos-sync/main.go
@@ -1,0 +1,231 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"k8s.io/test-infra/prow/config/org"
+	"k8s.io/test-infra/prow/config/secret"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+type gitHubClient interface {
+	GetRepo(owner, name string) (github.FullRepo, error)
+}
+
+type options struct {
+	peribolosConfig string
+	destOrg         string
+	releaseRepoPath string
+	whitelist       flagutil.Strings
+	github          flagutil.GitHubOptions
+
+	whitelistByOrg map[string]sets.String
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.peribolosConfig, "peribolos-config", "", "Peribolos configuration file")
+	fs.StringVar(&o.releaseRepoPath, "release-repo-path", "", "Path to a openshift/release repository directory")
+	fs.StringVar(&o.destOrg, "destination-org", "", "Destination name of the peribolos configuration organzation")
+
+	fs.Var(&o.whitelist, "whitelist", "One or more repositories that do not promote official images and need to be included")
+
+	o.github.AddFlags(fs)
+	fs.Parse(os.Args[1:])
+
+	return o
+}
+
+func validateOptions(o *options) error {
+	var validationErrors []error
+
+	if len(o.releaseRepoPath) == 0 {
+		validationErrors = append(validationErrors, errors.New("--release-repo-path is not specified"))
+	}
+	if len(o.peribolosConfig) == 0 {
+		validationErrors = append(validationErrors, errors.New("--peribolos-config is not specified"))
+	}
+	if len(o.destOrg) == 0 {
+		validationErrors = append(validationErrors, errors.New("--destination-org is not specified"))
+	}
+
+	if len(o.whitelist.Strings()) > 0 {
+		if err := validateWhitelist(o.whitelist.Strings()); err != nil {
+			validationErrors = append(validationErrors, err)
+		} else {
+			o.whitelistByOrg = getWhitelistByOrg(o.whitelist.Strings())
+		}
+	}
+
+	if err := o.github.Validate(false); err != nil {
+		validationErrors = append(validationErrors, err)
+	}
+
+	return kerrors.NewAggregate(validationErrors)
+}
+
+func validateWhitelist(whitelist []string) error {
+	for _, repo := range whitelist {
+		if len(strings.Split(repo, "/")) != 2 {
+			return fmt.Errorf("--whitelist: %s must be in org/repo format", repo)
+		}
+	}
+	return nil
+}
+
+func getWhitelistByOrg(whitelist []string) map[string]sets.String {
+	ret := make(map[string]sets.String)
+
+	for _, repo := range whitelist {
+		orgRepo := strings.Split(repo, "/")
+
+		orgName := orgRepo[0]
+		repoName := orgRepo[1]
+
+		repos, exist := ret[orgName]
+		if !exist {
+			repos = sets.NewString()
+		}
+		ret[orgName] = repos.Insert(repoName)
+
+	}
+	return ret
+}
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(&o)
+	if err != nil {
+		logrus.WithError(err).Fatal("invalid options")
+	}
+	logger := logrus.WithField("destination-org", o.destOrg)
+
+	go func() {
+		interrupts.WaitForGracefulShutdown()
+		os.Exit(1)
+	}()
+
+	b, err := ioutil.ReadFile(o.peribolosConfig)
+	if err != nil {
+		logger.WithError(err).Fatal("could not read peribolos configuration file")
+	}
+
+	var peribolosConfig org.FullConfig
+	if err := yaml.Unmarshal(b, &peribolosConfig); err != nil {
+		logger.WithError(err).Fatal("failed to unmarshal peribolos config")
+	}
+
+	secretAgent := &secret.Agent{}
+	if err := secretAgent.Start([]string{o.github.TokenPath}); err != nil {
+		logrus.WithError(err).Fatal("Error starting secrets agent.")
+	}
+	gc, err := o.github.GitHubClient(secretAgent, false)
+	if err != nil {
+		logger.WithError(err).Fatal("Error getting GitHub client.")
+	}
+
+	orgRepos, err := getReposForPrivateOrg(o.releaseRepoPath, o.whitelistByOrg)
+	if err != nil {
+		logger.WithError(err).Fatal("couldn't get the list of org/repos that promote official images")
+	}
+
+	peribolosRepos := generateRepositories(gc, orgRepos, logger)
+	peribolosConfigByOrg := peribolosConfig.Orgs[o.destOrg]
+	peribolosConfigByOrg.Repos = peribolosRepos
+	peribolosConfig.Orgs[o.destOrg] = peribolosConfigByOrg
+
+	out, err := yaml.Marshal(peribolosConfig)
+	if err != nil {
+		logrus.WithError(err).Fatalf("%s failed to marshal output.", o.peribolosConfig)
+	}
+
+	if err := ioutil.WriteFile(o.peribolosConfig, out, 0666); err != nil {
+		logrus.WithError(err).Fatal("Failed to write output.")
+	}
+}
+
+func generateRepositories(gc gitHubClient, orgRepos map[string]sets.String, logger *logrus.Entry) map[string]org.Repo {
+	peribolosRepos := make(map[string]org.Repo)
+
+	for orgName, repos := range orgRepos {
+		for repo := range repos {
+			logger.WithFields(logrus.Fields{"org": orgName, "repo": repo}).Info("Processing repository details...")
+
+			fullRepo, err := gc.GetRepo(orgName, repo)
+			if err != nil {
+				logger.WithError(err).Fatal("couldn't get repo details")
+			}
+
+			peribolosRepos[fullRepo.Name] = org.PruneRepoDefaults(org.Repo{
+				Description:      &fullRepo.Description,
+				HomePage:         &fullRepo.Homepage,
+				Private:          &fullRepo.Private,
+				HasIssues:        &fullRepo.HasIssues,
+				HasProjects:      &fullRepo.HasProjects,
+				HasWiki:          &fullRepo.HasWiki,
+				AllowMergeCommit: &fullRepo.AllowMergeCommit,
+				AllowSquashMerge: &fullRepo.AllowSquashMerge,
+				AllowRebaseMerge: &fullRepo.AllowRebaseMerge,
+				Archived:         &fullRepo.Archived,
+				DefaultBranch:    &fullRepo.DefaultBranch,
+			})
+		}
+	}
+
+	return peribolosRepos
+}
+
+// getReposForPrivateOrg itterates through the release repository directory and creates a map of
+// repository sets by organization that promote official images or if they are whitelisted by user.
+func getReposForPrivateOrg(releaseRepoPath string, includeRepos map[string]sets.String) (map[string]sets.String, error) {
+	ret := make(map[string]sets.String)
+
+	callback := makeCallback(includeRepos, ret)
+	if err := config.OperateOnCIOperatorConfigDir(filepath.Join(releaseRepoPath, config.CiopConfigInRepoPath), callback); err != nil {
+		return ret, fmt.Errorf("error while operating in ci-operator configuration files: %v", err)
+	}
+
+	return ret, nil
+}
+
+// makeCallback returns a function usable for OperateOnCIOperatorConfigDir that picks the orgs/repos
+// that promotes official images and the one that are in whitelist.
+func makeCallback(includeRepos, orgReposPicked map[string]sets.String) func(*api.ReleaseBuildConfiguration, *config.Info) error {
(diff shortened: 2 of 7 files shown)


cmd/private-org-peribolos-sync/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+This tool generates a mapping table of peribolos repository configurations for the given private
+organization. 
+It walks through the release repository path, given by `--release-repo-path`, and detects which of the repositories
+are promoting official images. The repositories that are specified in `--include-repo`, will be included if they exist in the release repository's path as well. Furthermore, it will get the required information for each of them from GitHub
+and will generate its peribolos configuration.
+Finally, the tool will update the peribolos configuration given by `--peribolos-config`.
+
+```console
+$ private-org-peribolos-sync -h
```
stevekuznetsov, line 11: This will go stale very quickly, rather just not paste it here

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+)
+
+type gitHubClient interface {
+	GetRepo(owner, name string) (github.FullRepo, error)
+}
+
+type stringSlice struct {
```
stevekuznetsov, line 33: There's a utility for this in Prow, please import it instead

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+
+	if len(o.includeRepos.values) > 0 {
+		o.includeReposSet = make(map[string]sets.String)
+		for _, repo := range o.includeRepos.values {
+			orgRepo := strings.Split(repo, "/")
+			if len(orgRepo) != 2 {
+				validationErrors = append(validationErrors, fmt.Errorf("--include-repo: %s must be in org/repo format", repo))
+			} else {
+				if _, ok := o.includeReposSet[orgRepo[0]]; !ok {
+					o.includeReposSet[orgRepo[0]] = sets.NewString([]string{orgRepo[1]}...)
```
stevekuznetsov, line 98:
```go
sets.NewString(orgRepo[1])
```

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+func main() {
+	o := gatherOptions()
+	err := validateOptions(&o)
+	if err != nil {
+		logrus.WithError(err).Fatal("invalid options")
+	}
+
+	go func() {
+		interrupts.WaitForGracefulShutdown()
+		os.Exit(1)
```
stevekuznetsov, line 122: ?

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+		logger.WithError(err).Fatal("Error getting GitHub client.")
+	}
+
+	orgRepos, err := getOrgReposWithOfficialImages(o.releaseRepoPath, o.includeReposSet)
+	if err != nil {
+		logger.WithError(err).Fatal("couldn't get the list of org/repos that promote official images")
+	}
+
+	peribolosRepos := generateRepositories(gc, orgRepos, logger)
+	peribolosConfigByOrg := peribolosConfig.Orgs[o.destOrg]
```
stevekuznetsov, line 152: Something is fishy about this ... why not

```go
peribolosConfig.Orgs[o.destOrg].Repos = peribolosRepos
```

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+	}
+
+	if err := config.OperateOnCIOperatorConfigDir(filepath.Join(releaseRepoPath, config.CiopConfigInRepoPath), callback); err != nil {
+		return ret, fmt.Errorf("error while operating in ci-operator configuration files: %v", err)
+	}
+
+	return ret, nil
+}
+
+// pruneRepoDefaults finds values in org.Repo config that matches the default
```
stevekuznetsov, line 223: This seems like something we need to share with upstream dump code, let's import it?

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+
+	return peribolosRepos
+}
+
+func getOrgReposWithOfficialImages(releaseRepoPath string, includeRepos map[string]sets.String) (map[string]sets.String, error) {
+	ret := make(map[string]sets.String)
+
+	callback := func(c *api.ReleaseBuildConfiguration, i *config.Info) error {
+		if !promotion.BuildsOfficialImages(c) {
+			if repo, ok := includeRepos[i.Org]; !ok && !repo.Has(i.Repo) {
```
stevekuznetsov, line 202: `s/repo/repos/`

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+				return nil
```
stevekuznetsov, line 203: Comment here might be nice
```go
// skip this repo unless it's in the whitelist
```

stevekuznetsov: or godoc on the function

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+		}
+	}
+
+	return peribolosRepos
+}
+
+func getOrgReposWithOfficialImages(releaseRepoPath string, includeRepos map[string]sets.String) (map[string]sets.String, error) {
+	ret := make(map[string]sets.String)
+
+	callback := func(c *api.ReleaseBuildConfiguration, i *config.Info) error {
```
stevekuznetsov, line 200: We should be unit testing this logic

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,254 @@
...
+		validationErrors = append(validationErrors, errors.New("--release-repo-path is not specified"))
+	}
+	if len(o.peribolosConfig) == 0 {
+		validationErrors = append(validationErrors, errors.New("--peribolos-config is not specified"))
+	}
+	if len(o.destOrg) == 0 {
+		validationErrors = append(validationErrors, errors.New("--destination-org is not specified"))
+	}
+
+	if len(o.includeRepos.values) > 0 {
```
stevekuznetsov, line 90: we should be unit testing this

cmd/private-org-peribolos-sync/main.go
droslean: It seems that golang doesn't let us do that.

cmd/private-org-peribolos-sync/main.go
droslean: I was hoping that too but it seems that this code exist in the main package of peribolos. I didn't want to move it and expose it just to use this once.

cmd/private-org-peribolos-sync/main.go
droslean: we should exit with non-zero if we catch an interruption signal, right?

stevekuznetsov: This is not how you are supposed to use this library -- please see the doc and other callers

cmd/private-org-peribolos-sync/main.go
stevekuznetsov: Nope, we should *always* move and import

cmd/private-org-peribolos-sync/main.go
droslean: Thanks, I will check the library out. I was confused because we already use it like this in other tools as well.

droslean:
/hold

Hold until https://github.com/openshift/ci-tools/pull/642
rebase should be needed after the 642 merges.

droslean:
/hold cancel

#642 was merged, This PR has been rebased.

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,233 @@
...
+
+type gitHubClient interface {
+	GetRepo(owner, name string) (github.FullRepo, error)
+}
+
+type options struct {
+	peribolosConfig string
+	destOrg         string
+	releaseRepoPath string
+	includeRepos    flagutil.Strings
```
petr-muller, line 38: I think this should be called a `whitelist` to communicate better what it really is, i.e. a list of repos added on top of the dynamically computed list.

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,233 @@
...
+				Archived:         &fullRepo.Archived,
+				DefaultBranch:    &fullRepo.DefaultBranch,
+			})
+		}
+	}
+
+	return peribolosRepos
+}
+
+func getOrgReposWithOfficialImages(releaseRepoPath string, includeRepos map[string]sets.String) (map[string]sets.String, error) {
```
petr-muller, line 202: The name of this method lies, it does more than that and should be actually called `getOrgReposWithOfficialImagesAndThoseInWhitelist`. That is of course a very bad name, which indicates the code might not be structured the best way.

I suggest to do one of the following:

1) **Either** keep the current structure but rename this to something more generic (`getReposForPrivateOrg`) and document it includes repos from two sources
2) **or** keep the name, remove the whitelist part from the list, and extend the returned map with the whitelisted repos at the callsite, like:

```
reposForPrivateOrg, err = getOrgReposWithOfficialImages(releaseRepoPath string);
...check errors...
reposForPrivateOrg = addWhitelist(reposForPrivateOrg, whitelist)
```

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,233 @@
...
+			if repos, _ := includeRepos[i.Org]; !repos.Has(i.Repo) {
+				return nil
+			}
+		}
+
+		if _, ok := orgReposPicked[i.Org]; !ok {
+			orgReposPicked[i.Org] = sets.NewString(i.Repo)
+		} else {
+			orgReposPicked[i.Org].Insert(i.Repo)
+		}
```
petr-muller, line 229: Could this be a simple `Union` call? `orgReposPicked[i.Org]` should return a zero value of a sets.String when `i.Org` is not in the map, and the zero value should be fine to use in a `Union`...

cmd/private-org-peribolos-sync/main.go
droslean: Option 1 is the best in this case. If we go with option 2 we will end up relying on the user that the whitelisted repo actually exists and we will make unnecessary GitHub calls.
I would prefer to allow picking the repo while we iterate in the release repo.

petr-muller: That sounds good!

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,230 @@
...
+func getWhitelistByOrg(whitelist []string) map[string]sets.String {
+	ret := make(map[string]sets.String)
+
+	for _, repo := range whitelist {
+		orgRepo := strings.Split(repo, "/")
+
+		orgName := orgRepo[0]
+		repoName := orgRepo[1]
+
+		ret[orgName] = ret[orgName].Union(sets.NewString(repoName))
```
stevekuznetsov, line 106: err... `Insert(repoName)`?

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,230 @@
...
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(&o)
+	if err != nil {
+		logrus.WithError(err).Fatal("invalid options")
+	}
+	logger := logrus.WithField("destination-org", o.destOrg)
+
+	sigTerm := make(chan os.Signal, 1)
```
stevekuznetsov, line 119: Why? Just use `prow/interrupts`

cmd/private-org-peribolos-sync/main_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+type fakeGitHub interface {
+	GetRepo(owner, name string) (github.FullRepo, error)
+}
+
+type fakeGitHubClient struct {
+	fakeGitHub
+}
+
+func (f fakeGitHubClient) GetRepo(owner, name string) (github.FullRepo, error) {
```
stevekuznetsov, line 28: Does the upstream `prow/github` not have this fake already?

cmd/private-org-peribolos-sync/main_test.go
```diff
@@ -0,0 +1,277 @@
...
+}
+
+func TestGetWhitelistByOrg(t *testing.T) {
+	testCases := []struct {
+		id        string
+		whitelist []string
+		expected  map[string]sets.String
+	}{
+		{
+			id:        "repos to include, changes expected",
```
stevekuznetsov, line 134: what are changes?

cmd/private-org-peribolos-sync/main_test.go
```diff
@@ -0,0 +1,277 @@
...
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.id, func(t *testing.T) {
+			whitelistByOrg := getWhitelistByOrg(tc.whitelist)
+
+			if !reflect.DeepEqual(whitelistByOrg, tc.expected) {
+				t.Fatal(cmp.Diff(whitelistByOrg, tc.expected))
```
stevekuznetsov, line 153: can we add in the test name and maybe a human message? failure and a diff is really not fun to read

stevekuznetsov: (think about someone breaking this test who has never touched this code before)

cmd/private-org-peribolos-sync/main.go
droslean: It will panic. because there is a chance that the `ret[orgName]` won't exist.

cmd/private-org-peribolos-sync/main.go
droslean: It seems that `prow/interrupts` library is for more complex activities. Here we just want to kill the program when it receives an interruption.

cmd/private-org-peribolos-sync/main_test.go
droslean: surprisingly no!

cmd/private-org-peribolos-sync/main_test.go
droslean: sure.

cmd/private-org-peribolos-sync/main.go
stevekuznetsov: Then don't do anything at all. Your program will exit when it gets `SIGINT` and will do so with a nonzero code.

cmd/private-org-peribolos-sync/main.go
stevekuznetsov: Why does `ret[orgName].Union()` work but `ret[orgName].Insert()` not? This is silly. Just do

```go
repos, exist := ret[orgName]
if !exist {
    repos = sets.NewString()
}
repos.Insert(repoName)
```

The code as written stands out as not normal and makes you wonder why we're doing what we're doing. We want to do a set insertion, so we should do that. By handling the `nil` case explicitly, in code, we make our intent clear to the reader.

cmd/private-org-peribolos-sync/main.go
droslean: It seems that it is not exiting... I guess something is going in the `prow/github` package.

droslean: So it seems that I have to use something like https://github.com/openshift/ci-tools/blob/master/cmd/repo-init/main.go#L89-L92 to actual exit, but this library is not meant to be used that way.

cmd/private-org-peribolos-sync/main.go
droslean: I had this like this in the previous version of the commit, but it seems `Union` does all that in one line.  However, I am willing to rewrite it as you suggested because is more readable.

droslean: addressed

cmd/private-org-peribolos-sync/main_test.go
droslean: I can add this in upstream and use it here in a follow-up.

cmd/private-org-peribolos-sync/main.go
stevekuznetsov: Please look upstream for every single user of that library and copy how they use it :)

stevekuznetsov: The `repo-init` is special since it's interacting with users on the commandline

droslean: I investigated this, and I can't find any reasonable way to use this library in this tool. I can use some insights :)

cmd/private-org-peribolos-sync/main_test.go
droslean: opened https://github.com/kubernetes/test-infra/pull/17164

droslean: addressed. The code is using the upstream fakegithubclient.

cmd/private-org-peribolos-sync/main.go
droslean: With further investigation, I got it. Addressed.

cmd/private-org-peribolos-sync/main.go
stevekuznetsov: The reasoning here is that code is built to be read, not written. There is no utility to us if it is one line or not. Using a side-effect of `Union()` implementation to allow us to do it in one line makes it (maybe) easier to write since there's less code but _definitely_ makes it harder to read. We always want to prefer to make it readable.

cmd/private-org-peribolos-sync/main.go
```diff
@@ -0,0 +1,231 @@
...
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(&o)
+	if err != nil {
+		logrus.WithError(err).Fatal("invalid options")
+	}
+	logger := logrus.WithField("destination-org", o.destOrg)
+
+	interrupts.OnInterrupt(func() {
```
stevekuznetsov, line 123: I do not understand why this is necessary at all. The _default_ behavior of a process that receives `SIGxxx` is to exit with an error code thats `128+SIGxxx`. Literally not having this code do _anything_ will get you the behavior you want, right?

droslean: Exactly. But for some strange reason, the only way to stop the application from running is to `kill -9` the process. When it receives the signal there is log coming from github client that it receives it and it just continues running until it finishes.

petr-muller: @stevekuznetsov AFAIK there's *something* (probalby in gh client by what @droslean writes?) in test-infra that causes the binaries to not simply exit on SIGINT et al, that's why we have these bits in our tools:

https://github.com/openshift/ci-tools/blob/d6f2ea08e5d0e8072549fde2bd87ca825a94df4a/cmd/cvp-trigger/main.go#L73-L76

https://github.com/openshift/ci-tools/blob/d6f2ea08e5d0e8072549fde2bd87ca825a94df4a/cmd/repo-init/main.go#L89-L92

I understand that we may be cargo-culting this snippet around, but without it the behavior is how Nikos describes it.

stevekuznetsov: @petr-muller no, the `repo-init` one is because I wanted to handle SIGINT while handling user input reads from the terminal. Let's dig.

PR #639, opened 2020-04-06
testgrid-config-generator: Include new dashboard type 'broken'

smarterclayton:
The periodic label overrides the release config suggestion. The
generator is restructured to be simpler as a result. A new 'broken'
type is introduced for jobs we deem failures.

This is a result of agreement with release oversight.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -18,7 +18,6 @@ import (
 	"github.com/GoogleCloudPlatform/testgrid/pb/config"
 	jc "github.com/openshift/ci-tools/pkg/jobconfig"
 	"github.com/sirupsen/logrus"
-	prowconfig "k8s.io/test-infra/prow/config"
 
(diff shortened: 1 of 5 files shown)


cmd/testgrid-config-generator/main.go
```diff
@@ -158,37 +157,23 @@ type release struct {
...
 // release-gating and release-informing tests simple.
 //
-// We read the release controller's configuration for all of the release candidates
-// being tested and auto-generate TestGrid configuration for the jobs involved,
-// partitioning them by which release they are using (OKD or OCP), which version they
-// run for and whether or not they are informing or blocking.
+// We read all jobs that are annotated for the grid. The release controller's configuration
+// is used to default those roles but they can be overriden per job. We partition by overall
+// type (blocking, informing, broken), version or generic (generic have no version), and by
+// release type (ocp or okd).
```
bparees, line 168: would be nice to see this summarized in the "updating testgrid" documentation.

PR #651, opened 2020-04-08
rehearse: remove --no-fail parameter entirely

petr-muller:
This was was only useful at the beginning when we wanted to run in a
stealth mode, but not anymore. The `gracefulExit` method only existed to
handle the no-fail decision, so it was removed and partially inlined
in call sites.

This is safe to do after: https://github.com/openshift/release/pull/8219

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -26,7 +26,6 @@ import (
 
 type options struct {
 	dryRun       bool
-	noFail       bool
 	local        bool
 	debugLogPath string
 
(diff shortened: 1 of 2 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -332,18 +344,21 @@ func rehearseMain() int {
...
 	if err != nil {
 		logger.WithError(err).Error("Failed to rehearse jobs")
-		return gracefulExit(o.noFail, rehearseFailureOutput)
+		fmt.Fprintf(os.Stderr, rehearseFailureOutput)
+		return FAILURE
 	}
 	if !success {
 		logger.Error("Some jobs failed their rehearsal runs")
-		return gracefulExit(o.noFail, jobsFailureOutput)
+		fmt.Fprintf(os.Stderr, jobsFailureOutput)
```
stevekuznetsov, line 360: This doesn't follow our normal convention. Why not return an `error` and, if that's not `nil`, have the caller print that to `os.Stderr` and exit non-zero?

petr-muller: Well, I just inlined what was already there /shrug. But I can do it this way, no problem.

petr-muller: Addressed.

PR #652, opened 2020-04-08
rehearse: simplify detection of jobs with changed clusters

petr-muller:
There is no reason why this needs to be in a separate method when
existing methods already iterate over all jobs and check for different
fields.

### DIFF
--- a/pkg/diffs/diffs_test.go
+++ b/pkg/diffs/diffs_test.go
@@ -20,7 +20,12 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobconfig"
 )
 
-var ignoreUnexported = cmpopts.IgnoreUnexported(prowconfig.Presubmit{}, prowconfig.Brancher{}, prowconfig.RegexpChangeMatcher{})
+var ignoreUnexported = cmpopts.IgnoreUnexported(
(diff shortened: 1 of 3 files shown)


pkg/diffs/diffs_test.go
```diff
@@ -20,7 +20,12 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobconfig"
 )
 
-var ignoreUnexported = cmpopts.IgnoreUnexported(prowconfig.Presubmit{}, prowconfig.Brancher{}, prowconfig.RegexpChangeMatcher{})
+var ignoreUnexported = cmpopts.IgnoreUnexported(
+	prowconfig.Presubmit{},
+	prowconfig.Brancher{},
+	prowconfig.RegexpChangeMatcher{},
+	prowconfig.Periodic{},
```
stevekuznetsov, line 27: The whole thing!?!?!

petr-muller: I understand this is how `IgnoreUnexported` is supposed to be used:

> IgnoreUnexported returns an Option that only ignores the immediate unexported fields of a struct

`Periodic` has an unexported `interval time.Duration` member, so I think this usage is correct.

PR #653, opened 2020-04-08
Revert "fix job name safe env var for the variant jobs"

smarterclayton:
This reverts commit 7a4efaafe3cb29db765c791965aa2834f1616573.
This reverts commit fb7981adfcac1401f3466369af1c3a80ace1d7bb.

Variants should not be in the same namespace by design (they are
different versions of the same config and we don't support config
in the same namespace). We should identify whatever the fix is
outside, because a variant using e2e-gcp should look exactly like
the core branch using e2e-gcp.

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -300,13 +300,7 @@ func generatePodSpecOthers(info *prowgenInfo, release string, test *cioperatorap
 	case cioperatorapi.ClusterProfileVSphere:
 		targetCloud = "vsphere"
 	}
-
-	clusterProfileName := fmt.Sprintf("%s-cluster-profile", test.As)
-	if len(info.Variant) > 0 {
-		clusterProfileName = fmt.Sprintf("%s-%s", info.Variant, clusterProfileName)
-	}
-	clusterProfilePath := fmt.Sprintf("/usr/local/%s", clusterProfileName)
-
+	clusterProfilePath := fmt.Sprintf("/usr/local/%s-cluster-profile", test.As)
 	templatePath := fmt.Sprintf("/usr/local/%s", test.As)
 	podSpec := generateCiOperatorPodSpec(info, test.Secrets, []string{test.As})
 	clusterProfileVolume := generateClusterProfileVolume("cluster-profile", fmt.Sprintf("cluster-secrets-%s", targetCloud))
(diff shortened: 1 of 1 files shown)


smarterclayton:
/hold
/assign @droslean 

Variants should be no different than mains - I think whatever broke Tekton is related to them reusing namespaces incorrectly (the hash of the config should be calculated and joined to the namespace).  Did we break hash generation?

smarterclayton:
Ok, found the problem.  Tekton has the following PR that triggered 8 separate jobs:

https://prow.svc.ci.openshift.org/pr-history/?org=openshift&repo=tektoncd-pipeline&pr=364

is a PR to `release-v0.11.0-rc3`.

The tekton repo defines these config definitions:

```
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.0__variant.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.0-rc2__variant.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.0-rc2.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.0-rc3__variant.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.0-rc3.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.0.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.1__variant.yaml
ci-operator/config/openshift/tektoncd-pipeline/openshift-tektoncd-pipeline-release-v0.11.1.yaml
```

Which triggered 8 jobs - 4 each of e2e and images - so looking at images:

```
pull-ci-openshift-tektoncd-pipeline-release-v0.11.0-images
pull-ci-openshift-tektoncd-pipeline-release-v0.11.0-rc3-images
pull-ci-openshift-tektoncd-pipeline-release-v0.11.0-rc3-variant-images
pull-ci-openshift-tektoncd-pipeline-release-v0.11.0-variant-images
```

That is wrong.  Only 

```
pull-ci-openshift-tektoncd-pipeline-release-v0.11.0-rc3-images
pull-ci-openshift-tektoncd-pipeline-release-v0.11.0-rc3-variant-images
```

should have run.

This is because of regex rules - the generated jobs run on release-v0.11.0 branches and ALL substring branches.  So you should not have job configurations for branch `X` and `X-Y`.

The fix is to fix the jobs, this fix is not necessary.  If the team truly needs different definitions, that is a feature request to ci-operator.

smarterclayton:
Basically their release-v0.11.0 and release-v0.11.0-rc3 configs AND commits were the same, which meant they hashed to the same value, which means they ran in the same namespace.  The bigger problem is that nested release jobs like this don't make sense - we just don't support that in DPTP yet (overlapping release branches with promotion).

The easy fix for the team will be to make sure they never have overlapping duplicate ci-operator configs on overlapping branches - if one of the ci-operator configs has different contents, they are hashed appropriately.

The "already exists" is a bug in template creation - you should be able to create identical templates.

/hold cancel

I will revert the release changes.

PR #654, opened 2020-04-08
rehearse: unify how changed jobs are detected

petr-muller:
The comparison logic was duplicated in methods that compared Presubmits
and Periodics, so I extracted the logic to a common method. I also
enhanced the logging to capture the reasons why a job is selected for
rehearsal.

### DIFF
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -23,6 +23,7 @@ import (
 const (
 	logRepo       = "repo"
 	logJobName    = "job-name"
+	logReasons    = "reasons"
 	logCiopConfig = "ciop-config"
 
 	// ConfigInRepoPath is the prow config path from release repo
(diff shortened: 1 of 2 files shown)


pkg/diffs/diffs.go
```diff
@@ -97,27 +98,24 @@ func GetChangedPresubmits(prowMasterConfig, prowPRConfig *prowconfig.Config, log
...
-				} else if masterJob.Optional && !job.Optional {
-					logger.WithFields(logFields).Info(chosenJob)
-					ret.Add(repo, job)
-				} else if !masterJob.AlwaysRun && job.AlwaysRun {
-					logger.WithFields(logFields).Info(chosenJob)
-					ret.Add(repo, job)
-				} else if masterJob.Cluster != job.Cluster {
-					logger.WithFields(logFields).Info(chosenJob)
-					ret.Add(repo, job)
+			var why []string
```
alvaroaleman, line 101: IMHO variables should describe what they contain, not what questions can be asked to which the content of the var is the correct answer, so something like `reasons` for the var name and log key makes IMHO more sense

alvaroaleman: (as in: relation var name var content should be an equasion)

petr-muller: :+1:

pkg/diffs/diffs.go
```diff
@@ -23,6 +23,7 @@ import (
 const (
 	logRepo       = "repo"
 	logJobName    = "job-name"
+	logReasons    = "why"
```
alvaroaleman, line 26: here as well

alvaroaleman:
Just to test it out:
/test ?

petr-muller:
That's goddamn sweet.

pkg/diffs/diffs.go
petr-muller: aye

PR #659, opened 2020-04-09
Fix autoconfigbrancher to generate correct cluster

alvaroaleman:
/assign @stevekuznetsov @petr-muller

PR #660, opened 2020-04-10
Use sanitize-prow-jobs to control jobs' cluster

hongkailiu:
/cc @alvaroaleman @petr-muller 

Tested with

```bash
sanitize-prow-jobs --prow-jobs-dir=./ci-operator/jobs
```

/hold

requires https://github.com/openshift/release/pull/8209/files

### DIFF
--- a/cmd/autoconfigbrancher/main.go
+++ b/cmd/autoconfigbrancher/main.go
@@ -199,7 +199,7 @@ func main() {
 	run(cmd, args...)
 
 	cmd = "/usr/bin/sanitize-prow-jobs"
-	args = []string{"--prow-jobs-dir", "./ci-operator/jobs"}
+	args = []string{"--prow-jobs-dir", "./ci-operator/jobs", "--config-path", "./core-services/sanitize-prow-jobs/_config.yaml"}
 	run(cmd, args...)
 	commitIfNeeded("sanitize-prow-jobs  --prow-jobs-dir ./ci-operator/jobs", author)
 
--- a/cmd/sanitize-prow-jobs/main.go
+++ b/cmd/sanitize-prow-jobs/main.go
@@ -6,22 +6,63 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"sigs.k8s.io/yaml"
 )
 
-const defaultCluster = "api.ci"
-
 type options struct {
 	prowJobConfigDir string
 	configPath       string
 
 	help bool
+}
+
+// Config is the configuration file of this tools, which defines the cluster parameter for each Prow job, i.e., where it runs
+type Config struct {
+	// the cluster context name if no other condition matches
+	Default string `json:"default"`
+	// the cluster context name for non Kubernetes jobs
+	NonKubernetes string `json:"nonKubernetes"`
+	// Groups maps a group of jobs to a cluster
+	Groups map[string]Group `json:"groups"`
+}
+
+//Group is a group of jobs
+type Group struct {
+	// a list of job names
+	Jobs []string `json:"jobs"`
+	// a list of regexes of the file paths
+	Paths []string `json:"paths"`
+
+	PathREs []*regexp.Regexp `json:"-"`
+}
+
+func (config *Config) getClusterForJob(jobBase prowconfig.JobBase, path string) string {
+	if jobBase.Agent != "kubernetes" {
+		return config.NonKubernetes
+	}
+	for cluster, group := range config.Groups {
+		for _, job := range group.Jobs {
+			if jobBase.Name == job {
+				return cluster
+			}
+		}
+	}
+	for cluster, group := range config.Groups {
+		for _, re := range group.PathREs {
+			if re.MatchString(path) {
+				return cluster
+			}
+		}
+	}
+	return config.Default
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
@@ -34,10 +75,42 @@ func bindOptions(flag *flag.FlagSet) *options {
 	return opt
 }
 
-func determinizeJobs(prowJobConfigDir string) error {
-	errChan := make(chan error)
+func loadConfig(configPath string) (*Config, error) {
+	config := &Config{}
+	data, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to read the config file %q", configPath)
+	}
+	err = yaml.Unmarshal(data, config)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to unmarshal the config %q", string(data))
+	}
 
 	var errs []error
+	for cluster, group := range config.Groups {
+		var pathREs []*regexp.Regexp
+		for i, p := range group.Paths {
+			re, err := regexp.Compile(p)
+			if err != nil {
+				errs = append(errs, errors.Wrapf(err, "failed to compile regex config.Groups[%s].Paths[%d] from %q", cluster, i, p))
+				continue
+			}
+			pathREs = append(pathREs, re)
+		}
+		group.PathREs = pathREs
+		config.Groups[cluster] = group
+	}
+
+	if len(errs) > 0 {
+		return nil, utilerrors.NewAggregate(errs)
+	}
+	return config, nil
+}
+
+func determinizeJobs(prowJobConfigDir string, config *Config) error {
+	errChan := make(chan error)
+	var errs []error
+
 	errReadingDone := make(chan struct{})
 	go func() {
 		for err := range errChan {
@@ -73,7 +146,7 @@ func determinizeJobs(prowJobConfigDir string) error {
 				return
 			}
 
-			defaultJobConfig(jobConfig)
+			defaultJobConfig(jobConfig, path, config)
 
 			serialized, err := yaml.Marshal(jobConfig)
 			if err != nil {
@@ -99,28 +172,20 @@ func determinizeJobs(prowJobConfigDir string) error {
 	return utilerrors.NewAggregate(errs)
 }
 
-func defaultJobConfig(jc *prowconfig.JobConfig) {
+func defaultJobConfig(jc *prowconfig.JobConfig, path string, config *Config) {
 	for k := range jc.PresubmitsStatic {
 		for idx := range jc.PresubmitsStatic[k] {
-			defaultJobBase(&jc.PresubmitsStatic[k][idx].JobBase)
+			jc.PresubmitsStatic[k][idx].JobBase.Cluster = config.getClusterForJob(jc.PresubmitsStatic[k][idx].JobBase, path)
 		}
 	}
 	for k := range jc.PostsubmitsStatic {
 		for idx := range jc.PostsubmitsStatic[k] {
-			defaultJobBase(&jc.PostsubmitsStatic[k][idx].JobBase)
+			jc.PostsubmitsStatic[k][idx].JobBase.Cluster = config.getClusterForJob(jc.PostsubmitsStatic[k][idx].JobBase, path)
 		}
 	}
 	for idx := range jc.Periodics {
-		defaultJobBase(&jc.Periodics[idx].JobBase)
+		jc.Periodics[idx].JobBase.Cluster = config.getClusterForJob(jc.Periodics[idx].JobBase, path)
 	}
-}
-
-func defaultJobBase(jb *prowconfig.JobBase) {
-
-	if jb.Cluster == "" || jb.Cluster == "default" {
-		jb.Cluster = defaultCluster
-	}
-
 }
 
 func main() {
@@ -133,11 +198,18 @@ func main() {
 		os.Exit(0)
 	}
 
-	if len(opt.prowJobConfigDir) > 0 {
-		if err := determinizeJobs(opt.prowJobConfigDir); err != nil {
-			logrus.WithError(err).Fatal("Failed to determinize")
-		}
-	} else {
+	if len(opt.prowJobConfigDir) == 0 {
 		logrus.Fatal("mandatory argument --prow-jobs-dir wasn't set")
+	}
+	if len(opt.configPath) == 0 {
+		logrus.Fatal("mandatory argument --config-path wasn't set")
+	}
+
+	config, err := loadConfig(opt.configPath)
+	if err != nil {
+		logrus.WithError(err).Fatalf("Failed to load config from %q", opt.configPath)
+	}
+	if err := determinizeJobs(opt.prowJobConfigDir, config); err != nil {
+		logrus.WithError(err).Fatal("Failed to determinize")
 	}
 }
--- a/cmd/sanitize-prow-jobs/main_test.go
+++ b/cmd/sanitize-prow-jobs/main_test.go
@@ -1,43 +1,184 @@
 package main
 
 import (
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"regexp"
 	"testing"
 
+	"github.com/google/go-cmp/cmp"
 	"k8s.io/test-infra/prow/config"
+	prowconfig "k8s.io/test-infra/prow/config"
 )
 
 func TestDefaultJobConfig(t *testing.T) {
 	jc := &config.JobConfig{
 		PresubmitsStatic: map[string][]config.Presubmit{
-			"a": {{}, {}, {JobBase: config.JobBase{Cluster: "default"}}},
+			"a": {{}, {}, {JobBase: config.JobBase{Agent: "kubernetes", Cluster: "default"}}},
 			"b": {{}, {}},
 		},
 		PostsubmitsStatic: map[string][]config.Postsubmit{
-			"a": {{}, {}, {JobBase: config.JobBase{Cluster: "default"}}},
+			"a": {{}, {}, {JobBase: config.JobBase{Agent: "kubernetes", Cluster: "default"}}},
 			"b": {{}, {}},
 		},
-		Periodics: []config.Periodic{{}, {}, {JobBase: config.JobBase{Cluster: "default"}}},
+		Periodics: []config.Periodic{{}, {}, {JobBase: config.JobBase{Agent: "kubernetes", Cluster: "default"}}},
 	}
 
-	defaultJobConfig(jc)
+	config := &Config{Default: "api.ci"}
+	defaultJobConfig(jc, "", config)
 
 	for k := range jc.PresubmitsStatic {
 		for _, j := range jc.PresubmitsStatic[k] {
-			if j.Cluster != "api.ci" {
+			if j.Agent == "kubernetes" && j.Cluster != "api.ci" {
 				t.Errorf("expected cluster to be 'api.ci', was '%s'", j.Cluster)
 			}
 		}
 	}
 	for k := range jc.PostsubmitsStatic {
 		for _, j := range jc.PostsubmitsStatic[k] {
-			if j.Cluster != "api.ci" {
+			if j.Agent == "kubernetes" && j.Cluster != "api.ci" {
 				t.Errorf("expected cluster to be 'api.ci', was '%s'", j.Cluster)
 			}
 		}
 	}
 	for _, j := range jc.Periodics {
-		if j.Cluster != "api.ci" {
+		if j.Agent == "kubernetes" && j.Cluster != "api.ci" {
 			t.Errorf("expected cluster to be 'api.ci', was '%s'", j.Cluster)
 		}
+	}
+}
+
+var (
+	c = Config{
+		Default:       "api.ci",
+		NonKubernetes: "app.ci",
+		Groups: map[string]Group{
+			"api.ci": {
+				Paths: []string{
+					".*-postsubmits.yaml$",
+					".*openshift/release/.*-periodics.yaml$",
+					".*-periodics.yaml$",
+				},
+				PathREs: []*regexp.Regexp{
+					regexp.MustCompile(".*-postsubmits.yaml$"),
+					regexp.MustCompile(".*openshift/release/.*-periodics.yaml$"),
+					regexp.MustCompile(".*-periodics.yaml$"),
+				},
+				Jobs: []string{
+					"pull-ci-openshift-release-master-build01-dry",
+					"pull-ci-openshift-release-master-core-dry",
+					"pull-ci-openshift-release-master-services-dry",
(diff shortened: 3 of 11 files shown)


cmd/sanitize-prow-jobs/main.go
```diff
@@ -15,27 +17,66 @@ import (
...
 }
 
+type Config struct {
+	Clusters struct {
+		Default      string `json:"default"`
+		Distribution map[string]struct {
+			Jobs   []string `json:"jobs"`
+			Files  []string `json:"files"`
+			Groups []string `json:"groups"`
+		} `json:"distribution"`
```
alvaroaleman, line 34: What about `assignment`? The distribution is more the result of the assignment we do here

cmd/sanitize-prow-jobs/main.go
```diff
@@ -15,27 +17,66 @@ import (
...
+		Groups map[string][]string       `json:"groups"`
+		REMap  map[string]*regexp.Regexp `json:"-"`
```
alvaroaleman, line 36: This is the REMap for the files, right? Why is it not alongside of files but a level higher? Also please name `FileREMap`, otherwise we have to rename it should we ever need another `REMap`

cmd/sanitize-prow-jobs/main.go
```diff
@@ -97,28 +138,64 @@ func determinizeJobs(prowJobConfigDir string) error {
...
-		jb.Cluster = defaultCluster
+func getCluster(name string, path string, config *Config) string {
+	distribution := config.Clusters.Distribution
+	clusters := make([]string, 0, len(distribution))
+	for cluster := range distribution {
+		clusters = append(clusters, cluster)
 	}
+	sort.Strings(clusters)
 
+	//jobs override files which override groups
```
alvaroaleman, line 169: Please add this comment to the type definition

cmd/sanitize-prow-jobs/main.go
```diff
@@ -97,28 +138,64 @@ func determinizeJobs(prowJobConfigDir string) error {
...
 }
 
-func defaultJobBase(jb *prowconfig.JobBase) {
+func defaultJobBase(jb *prowconfig.JobBase, path string, config *Config) {
+	jb.Cluster = getCluster(jb.Name, path, config)
+}
 
-	if jb.Cluster == "" || jb.Cluster == "default" {
-		jb.Cluster = defaultCluster
+func getCluster(name string, path string, config *Config) string {
```
alvaroaleman, line 161: Please rename `getCluster` -> `getClusterForJob` and `name` -> `jobName`.

Would it maybe make sense to make `getCluster` a method of `config` to have them closer together and signal the tight coupling?

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
```diff
@@ -0,0 +1,3 @@
+---
```
alvaroaleman, line 1: Why is this needed?

cmd/sanitize-prow-jobs/main_test.go
```diff
@@ -19,7 +20,20 @@ func TestDefaultJobConfig(t *testing.T) {
 		Periodics: []config.Periodic{{}, {}, {JobBase: config.JobBase{Cluster: "default"}}},
 	}
 
-	defaultJobConfig(jc)
+	config := &Config{Clusters: struct {
+		Default      string `json:"default"`
```
alvaroaleman, line 24: I don't think this needs json tags

cmd/sanitize-prow-jobs/main.go
```diff
@@ -97,28 +138,64 @@ func determinizeJobs(prowJobConfigDir string) error {
...
+			if name == job {
+				return cluster
+			}
+		}
+	}
+	for _, cluster := range clusters {
+		for _, file := range distribution[cluster].Files {
+			re, ok := config.Clusters.REMap[file]
+			if !ok {
+				continue
```
alvaroaleman, line 181: I think this should never happen and hence this is an error or not?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -15,27 +17,66 @@ import (
...
-func determinizeJobs(prowJobConfigDir string) error {
+func determinizeJobs(prowJobConfigDir, configPath string) error {
 	errChan := make(chan error)
-
 	var errs []error
+
+	config := &Config{}
+	data, err := ioutil.ReadFile(configPath)
+	if err != nil {
+		errs = append(errs, err)
```
alvaroaleman, line 57: I think we should return the error here and in the unmarshaling, as all forthcoming errors may be a result of erroring here. Also please wrap them so they are easier to understand and can be grepped for in the codebase

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Renamed.

> Why is it not alongside of files but a level higher?
That was my first try too. I think the current structure is simpler.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
hongkailiu: Removed.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,120 @@ import (
...
-
 type options struct {
 	prowJobConfigDir string
+	configPath       string
 
 	help bool
 }
 
+type Config struct {
+	Clusters struct {
```
stevekuznetsov, line 29: Any reason that these are anonymous types?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,120 @@ import (
...
+		Default    string `json:"default"`
+		Assignment map[string]struct {
+			// Jobs override both files and groups
+			Jobs []string `json:"jobs"`
+			// Files override files
```
stevekuznetsov, line 34: ?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,120 @@ import (
...
 type options struct {
 	prowJobConfigDir string
+	configPath       string
 
 	help bool
 }
 
+type Config struct {
+	Clusters struct {
+		Default    string `json:"default"`
```
stevekuznetsov, line 30: Public-facing config fields should have GoDoc

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,120 @@ import (
...
+	} `json:"clusters"`
+}
+
+func (config *Config) getClusterForJob(jobName string, path string) string {
+	distribution := config.Clusters.Assignment
+	clusters := make([]string, 0, len(distribution))
+	for cluster := range distribution {
+		clusters = append(clusters, cluster)
+	}
+	sort.Strings(clusters)
```
stevekuznetsov, line 49: Why?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,120 @@ import (
...
+
+	//jobs override files which override groups
+	for _, cluster := range clusters {
+		for _, job := range distribution[cluster].Jobs {
+			if jobName == job {
+				return cluster
+			}
+		}
+	}
+	for _, cluster := range clusters {
```
stevekuznetsov, line 59: This smells like a bad configuration structure. It should not be this hard to determine the cluster for a job. Why is it even possible to list a job in multiple lists/groups/regex-matches?

Why not:

```yaml
groups:
- name: foo
  jobs:
  - bar
  cluster: zing
raw:
- jobs:
  - whoa
  cluster: there
files:
- paths:
  - ^whatever.yaml$
  cluster: boogie
```

stevekuznetsov: Or, validate on config load that a job is not in more than one bucket and take it's presence in a bucket as authoritative -- that way if you find it under `cluster.files` or `cluster.groups` you know it belongs in that cluster no matter what.

stevekuznetsov: Said another way, we are building a lot of complexity here to allow `cluster1.groups` to override `cluster2.jobs` which seems silly

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,120 @@ import (
...
 	var errs []error
+	fileREMap := map[string]*regexp.Regexp{}
+	for _, v := range config.Clusters.Assignment {
+		for _, file := range v.Files {
+			if _, ok := fileREMap[file]; ok {
+				continue
+			}
+			re, err := regexp.Compile(file)
+			if err != nil {
+				errs = append(errs, err)
```
stevekuznetsov, line 115: without putting the cluster name, assignmment, and file name into this error it will be impossible to debug

cmd/sanitize-prow-jobs/main.go
```diff
@@ -97,28 +181,24 @@ func determinizeJobs(prowJobConfigDir string) error {
...
 	}
 }
 
-func defaultJobBase(jb *prowconfig.JobBase) {
-
-	if jb.Cluster == "" || jb.Cluster == "default" {
-		jb.Cluster = defaultCluster
-	}
-
+func defaultJobBase(jb *prowconfig.JobBase, path string, config *Config) {
```
stevekuznetsov, line 200: Why do we want a stateful function with a side-effect?

cmd/sanitize-prow-jobs/main_test.go
```diff
@@ -19,7 +20,20 @@ func TestDefaultJobConfig(t *testing.T) {
 		Periodics: []config.Periodic{{}, {}, {JobBase: config.JobBase{Cluster: "default"}}},
 	}
 
-	defaultJobConfig(jc)
+	config := &Config{Clusters: struct {
+		Default    string `json:"default"`
```
stevekuznetsov, line 24: See and *this* is why anonymous structs have a very limited use-case...

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
```diff
@@ -0,0 +1,2 @@
+clusters:
+  default: api.ci
```
stevekuznetsov, line 2: We're going to need something *way* more complicated to test this :)

cmd/sanitize-prow-jobs/main_test.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: No need to do this with the new structure.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main_test.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Removed.

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
hongkailiu: (Tested with the config defined https://github.com/openshift/release/pull/8209/files)

Will add unit tests when the core code goes review.

stevekuznetsov: Unit and integration tests should exist before review, not after -- they are just as important as code.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,134 @@ import (
...
+	for _, g := range config.Clusters.Raw {
+		for _, job := range g.Jobs {
+			if jobName == job {
+				return g.Cluster
+			}
+		}
+	}
+	for _, file := range config.Clusters.Files {
+		for _, p := range file.Paths {
+			if config.Clusters.FileREMap[p].MatchString(path) {
```
stevekuznetsov, line 78: co-locate the regex and the struct that created it

```
if file.regex.MatchString()
```

stevekuznetsov:
Do we even need regex file matching? Can't we achieve the same thing with regex presubmit name matching?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,134 @@ import (
...
 }
 
+// Config defines the structure of configuration file for this tool
+type Config struct {
+	Clusters Clusters
+}
+
+// Clusters defines the cluster parameter for each Prow job, i.e., where it runs
+type Clusters struct {
+	// the cluster context name if no other condition matches
```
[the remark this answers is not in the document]
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,134 @@ import (
...
+// Group defines a name for a group of jobs and where they run
+type Group struct {
+	// the name of the group
+	Name string
+	// the job names of the group
+	UnnamedGroup `json:",inline"`
+}
+
+// UnnamedGroup defines where a group of jobs run
+type UnnamedGroup struct {
```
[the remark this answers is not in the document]
hongkailiu: I chose the one with name to remind us of what the group is about/from.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,134 @@ import (
...
 type options struct {
 	prowJobConfigDir string
+	configPath       string
 
 	help bool
 }
 
+// Config defines the structure of configuration file for this tool
+type Config struct {
+	Clusters Clusters
```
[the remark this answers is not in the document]
hongkailiu: I think it is from https://github.com/openshift/release/pull/8209#pullrequestreview-389839352

hongkailiu:
> Do we even need regex file matching? Can't we achieve the same thing with regex presubmit name matching?

Comparing job names and job file paths, the latter is more regular.
I feel it is easier to express something like, keep the periodics of the release repo on api.ci.

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
hongkailiu: I am not saying it is not as important.
Just feel that it is more efficient after the core code is more stable.
I meant to do it with the current PR, not following up.

hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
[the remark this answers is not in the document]
stevekuznetsov: Maybe it looked different before, but there is no utility anymore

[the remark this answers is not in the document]
stevekuznetsov: This is outstanding @hongkailiu

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,126 @@ import (
...
+}
+
+// Group is where a group of jobs run
+type Group struct {
+	// name of the group
+	Name string
+	// the job names
+	Jobs []string
+	// the cluster where the jobs run
+	Cluster string
```
alvaroaleman, line 50: Why does this field exist? The group should just have a name and a list of jobs. The `Cluster` then specifies the `group(s)`.

Same comment for the `File` struct below.

cmd/sanitize-prow-jobs/main.go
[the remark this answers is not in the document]
hongkailiu: Missed this. On it now.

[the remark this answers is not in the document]
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: So you want to define cluster on groups instead of each group?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -6,36 +6,121 @@ import (
...
 	prowJobConfigDir string
+	configPath       string
 
 	help bool
 }
 
+// Config is the configuration file of this tools, which defines the cluster parameter for each Prow job, i.e., where it runs
+type Config struct {
+	// the cluster context name if no other condition matches
+	Default string
```
stevekuznetsov, line 30: Did we lose the JSON tags?

cmd/sanitize-prow-jobs/main.go
stevekuznetsov:
```yaml
default: "foo"
groups:
  "bar":
    paths:
    - etc
    - what
    jobs:
    - thing
    - other
```

This seems like the simplest approach. no?

cmd/sanitize-prow-jobs/main.go
hongkailiu: "sigs.k8s.io/yaml"

This pkg seems working without json tags for those fields.
(tested with config in https://github.com/openshift/release/pull/8209)

cmd/sanitize-prow-jobs/main.go
hongkailiu: Will do this tonight.

hongkailiu: Asked my question
https://coreos.slack.com/archives/GB7NB0CUC/p1587421714096000

cmd/sanitize-prow-jobs/main.go
stevekuznetsov: Sure, there are defaults when your field names are the same, but please follow the convention we use literally everywhere else -- explicitly set them.

hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -132,7 +199,14 @@ func main() {
 	}
 
 	if len(opt.prowJobConfigDir) > 0 {
-		if err := determinizeJobs(opt.prowJobConfigDir); err != nil {
+		if len(opt.configPath) == 0 {
```
stevekuznetsov, line 202: We should not do this -- always expect a user to give the flag and error if not.

stevekuznetsov: For seamless roll-out, we need a new PR first that adds the flag but does not validate that it is set. Then, we add a PR to add the flag to the job. Then, we land this PR, where we enforce it is set and use it

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
```diff
@@ -0,0 +1,16 @@
+default: api.ci
+nonKubernetes: api.ci
+groups:
```
stevekuznetsov, line 3: you dont need the map for these files

test/repo-init-integration/expected/ci-operator/jobs/org/other/org-other-nonstandard-presubmits.yaml
```diff
@@ -4,7 +4,7 @@ presubmits:
     always_run: true
     branches:
     - nonstandard
-    cluster: api.ci
+    cluster: ci/api-build01-ci-devcluster-openshift-com:6443
```
stevekuznetsov, line 7: Why did a `ci-operator` test change?

hongkailiu: Because we are testing the config in the integration test.
The config tells to run that job in build01.

https://github.com/openshift/ci-tools/blob/0fead13f0eb9f78c23e3fff1c7042b9d345421c4/test/repo-init-integration/input/core-services/sanitize-prow-jobs/_config.yaml#L7-L10

test/repo-init-integration/expected/core-services/sanitize-prow-jobs/_config.yaml
hongkailiu: Removed.

cmd/sanitize-prow-jobs/main.go
hongkailiu: On it now.

hongkailiu: Done.

cmd/autoconfigbrancher/main.go
```diff
@@ -199,7 +199,7 @@ func main() {
 	run(cmd, args...)
```
hongkailiu, line 199: Will revert https://github.com/openshift/ci-tools/pull/659/files
when everything is working with sanitize tool

PR #661, opened 2020-04-13
pj-rehearse: be better at logging

stevekuznetsov:

The semi-machine-read-able logs are hard to parse but this at least fixes areas where we'd not output anything at all when choosing a job.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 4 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -132,9 +133,7 @@ func rehearseMain() error {
 		jobSpec.Refs.Pulls[0].Number = int(time.Now().Unix())
 	}
 	org, repo, prNumber := jobSpec.Refs.Org, jobSpec.Refs.Repo, jobSpec.Refs.Pulls[0].Number
-
-	logger = logrus.WithField(prowgithub.PrLogField, prNumber)
-	logger.Info("Rehearsing Prow jobs for a configuration PR")
+	logger.Infof("Rehearsing Prow jobs for configuration PR %s/%s#%d", org, repo, prNumber)
```
alvaroaleman, line 136: Can we please make these fields and create a logger that we continue using?

stevekuznetsov: I really oppose that. We're not reading this with a machine and my experience reading a log today strongly made me rethink this approach ... I don't need to see the PR number on every line of the log, we only ever operate on one PR at a time.

PR #662, opened 2020-04-14
add Dockerfile for private-org-peribolos-sync tool

droslean:
needs #636 


alvaroaleman:
Why a separate PR?

PR #671, opened 2020-04-15
prowgen: mount secrets at /secrets/secret.Name

AlexNPavel:
Fixes bug where created secret name is different from secret name in podspec, causing a volume mount error.

This will require a regen PR in openshift/release

/cc @stevekuznetsov

### DIFF
--- a/cmd/ci-operator-prowgen/main_test.go
+++ b/cmd/ci-operator-prowgen/main_test.go
@@ -35,7 +35,7 @@ var unexportedFields = []cmp.Option{
 }
 
 func TestGeneratePodSpec(t *testing.T) {
-	testSecret := &ciop.Secret{Name: "test-secret", MountPath: "/usr/local/test-secret"}
+	testSecret := &ciop.Secret{Name: "secret-name", MountPath: "/usr/local/test-secret"}
 	tests := []struct {
 		description    string
 		info           *prowgenInfo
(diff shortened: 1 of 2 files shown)


cmd/ci-operator-prowgen/main_test.go
```diff
@@ -35,7 +35,7 @@ var unexportedFields = []cmp.Option{
 }
 
 func TestGeneratePodSpec(t *testing.T) {
-	testSecret := &ciop.Secret{Name: "test-secret", MountPath: "/usr/local/test-secret"}
+	testSecret := &ciop.Secret{Name: "secret-name", MountPath: "/usr/local/test-secret"}
```
stevekuznetsov, line 38: Why do we need this var?

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -165,7 +166,7 @@ func TestGeneratePodSpec(t *testing.T) {
 						"--promote",
 						"--some=thing",
 						"--target=target",
-						fmt.Sprintf("--secret-dir=%s", testSecret.MountPath),
+						fmt.Sprintf("--secret-dir=/secrets/%s", testSecret.Name),
```
stevekuznetsov, line 169: Let's use string literals everywhere we can -- no sprintf

cmd/ci-operator-prowgen/main_test.go
AlexNPavel: It was already there. It's not really necessary. Removed.

PR #672, opened 2020-04-16
Autoconfigbrancher: use exposed bumper library's function, and refactor

droslean:
This PR removes the local functions that are duplicated with the bumper library's functions and refactors the logic.


alvaroaleman:
/retitle Autoconfigbrancher: use exposed bumper library's function, and refactor

cmd/autoconfigbrancher/main.go
```diff
@@ -155,66 +144,85 @@ func main() {
...
-	run(cmd, args...)
-	commitIfNeeded("private-prow-configs-mirror --release-repo-path .", author)
-
-	cmd = "./hack/ensure_job_cluster.py"
-	args = []string{"-w=true"}
-	run(cmd, args...)
+	for i, step := range steps {
+		commited, err := runAndCommitIfNeeded(stdout, stderr, author, step.command, step.arguments)
+		if err != nil {
+			logrus.WithError(err).Fatalf("step[%d]: failed to run command and commit the changes", i)
```
alvaroaleman, line 209: How exactly is the index in the log message useful here?

cmd/autoconfigbrancher/main.go
```diff
@@ -155,66 +144,85 @@ func main() {
...
 	}
 
-	remoteBranch := "auto-config-brancher"
 	title := fmt.Sprintf("%s by auto-config-brancher job at %s", matchTitle, time.Now().Format(time.RFC1123))
-	cmd = "git"
-	args = []string{"push", "-f", fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", o.githubLogin,
-		string(sa.GetTokenGenerator(o.GitHubOptions.TokenPath)()), o.githubLogin, githubRepo),
-		fmt.Sprintf("HEAD:%s", remoteBranch)}
-	run(cmd, args...)
+	if err := bumper.GitCommitAndPush(fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", o.githubLogin, string(sa.GetTokenGenerator(o.GitHubOptions.TokenPath)()), o.githubLogin, githubRepo), remoteBranch, o.gitName, o.gitEmail, title, stdout, stderr); err != nil {
```
alvaroaleman, line 223: We already committed, I don't think we should do it again. Will this error if there are no changes?

cmd/autoconfigbrancher/main.go
```diff
@@ -89,44 +87,35 @@ func validateOptions(o options) error {
...
+		return false, fmt.Errorf("failed to 'git add .': %v", err)
+	}
 
-func run(cmd string, args ...string) {
-	logrus.WithField("cmd", cmd).WithField("args", args).Info("running command ...")
-	if combinedOutput, err := exec.Command(cmd, args...).CombinedOutput(); err != nil {
-		logrus.WithField("cmd", cmd).Debugf("output is '%s'", string(combinedOutput))
-		logrus.Fatalf("Failed to run command:'%s'", err)
+	commitArgs := []string{"commit", "-m", fullCommand, "--author", author}
+	if err := bumper.Call(stdout, stderr, gitCmd, commitArgs...); err != nil {
```
alvaroaleman, line 114: While we are at it, can we make the whole thing one commit rather than seven?

cmd/autoconfigbrancher/main.go
droslean: well we have a `ci-operator-prowgen` step multiple times. I though that its nice to have information which step failed

cmd/autoconfigbrancher/main.go
droslean: We `commited` but we didn't push the changes in the fork's branch.

cmd/autoconfigbrancher/main.go
droslean: I think it's better to have each command in its commit.

petr-muller: In the past it was useful to see the changes in separate commits, so I'd like to keep it that way, too.

cmd/autoconfigbrancher/main.go
droslean: @alvaroaleman you are right on this one. That will fail.... I going to fix this.

cmd/autoconfigbrancher/main.go
stevekuznetsov: eh, not super useful to know index -- the command that failed will be in the error

PR #676, opened 2020-04-18
Add client-side imagestreamtag wrapper

alvaroaleman:
Needed for https://issues.redhat.com/browse/DPTP-729, without this we need to make an apicall for each imagestreamtag, which doesn't really scale well to the ~80k imagestreamtags we have in api.ci.


pkg/util/imagestreamtagwrapper/imagestreamtagwrapper_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/pmezard/go-difflib/difflib"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	fakectrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/yaml"
+)
+
+func TestGetImageStramTag(t *testing.T) {
```
petr-muller, line 16: nit: `Stram`->`Stream`

pkg/util/imagestreamtagwrapper/imagestreamtagwrapper_test.go
```diff
@@ -0,0 +1,96 @@
...
+				Context:  3,
+			}
+			diffStr, err := difflib.GetUnifiedDiffString(diff)
+			if err != nil {
+				t.Fatalf("failed to get diff: %v", err)
+			}
+
+			if diffStr != "" {
+				t.Fatalf("objects differ:\n---\n%s\n---\n", diffStr)
+			}
```
petr-muller, line 93: Hmmm, the usual `cmp.Diff` can't handle embedded json (the L70 comment) and that's why have this instead, correct?

alvaroaleman: It can handle it insofar as that it correctly reports a diff. But the diff it displays is the diff of two byte arrays and that is pretty useless for humans

PR #680, opened 2020-04-18
Add console URL to ci-op's log

hongkailiu:
https://issues.redhat.com/browse/DPTP-992

/hold

Need RBAC for SA in the release repo.
https://github.com/openshift/release/pull/8410

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -48,6 +48,7 @@ import (
 	buildclientset "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	imageclientset "github.com/openshift/client-go/image/clientset/versioned/typed/image/v1"
 	projectclientset "github.com/openshift/client-go/project/clientset/versioned"
+	routeclientset "github.com/openshift/client-go/route/clientset/versioned/typed/route/v1"
 	templatescheme "github.com/openshift/client-go/template/clientset/versioned/scheme"
 	templateclientset "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
 
@@ -625,7 +626,24 @@ func (o *options) resolveInputs(steps []api.Step) error {
 	// TODO: instead of mutating this here, we should pass the parts of graph execution that are resolved
 	// after the graph is created but before it is run down into the run step.
 	o.jobSpec.Namespace = o.namespace
-	log.Printf("Using namespace %s", o.namespace)
+
+	//If we can resolve the field, use it. If not, don't.
+	var consoleHost string
+	if routeGetter, err := routeclientset.NewForConfig(o.clusterConfig); err != nil {
+		log.Printf("could not get route client for cluster config")
+	} else {
+		if consoleRoute, err := routeGetter.Routes("openshift-console").Get("console", meta.GetOptions{}); err != nil {
+			log.Printf("could not get route console in namespace openshift-console")
+		} else {
+			consoleHost = consoleRoute.Spec.Host
+		}
+	}
+
+	if consoleHost != "" {
+		log.Printf("Using namespace https://%s/k8s/cluster/projects/%s", consoleHost, o.namespace)
+	} else {
+		log.Printf("Using namespace %s", o.namespace)
+	}
 
 	return nil
 }


cmd/ci-operator/main_test.go
```diff
@@ -412,3 +414,138 @@ func TestGetResolverInfo(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := getConsoleURL(tc.cm)
+			equal(t, tc.expected, actual)
+			equalError(t, tc.expectedError, err)
+		})
+	}
+}
+
+func equalError(t *testing.T, expected, actual error) {
+	if expected != nil && actual == nil || expected == nil && actual != nil {
```
alvaroaleman, line 539: Can be simplified as `if (expected == nil ) != (actual == nil)`

cmd/ci-operator/main_test.go
```diff
@@ -412,3 +414,138 @@ func TestGetResolverInfo(t *testing.T) {
...
+			expectedError: fmt.Errorf("nil config map"),
+		},
+		{
+			name: "no data in cm",
+			cm: &coreapi.ConfigMap{
+				TypeMeta: meta.TypeMeta{},
+				ObjectMeta: meta.ObjectMeta{
+					Name:      "console-config",
+					Namespace: "openshift-console",
+				},
```
petr-muller, line 436: This looks like an irrelevant boilerplate; the tested method does not care about these fields, so why set them?

cmd/ci-operator/main_test.go
```diff
@@ -412,3 +414,138 @@ func TestGetResolverInfo(t *testing.T) {
...
+`,
+				},
+			},
+			expected: "https://console-openshift-console.apps.build01.ci.devcluster.openshift.com",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			actual, err := getConsoleURL(tc.cm)
+			equal(t, tc.expected, actual)
```
petr-muller, line 532: Not a fan. IMO brings no benefits over a an inline `if DeepEqual`.

cmd/ci-operator/main.go
```diff
@@ -450,6 +451,7 @@ func (o *options) Complete() error {
 		o.templates = append(o.templates, template)
 	}
 
+	o.clusterConsoleURL = "https://console.svc.ci.openshift.org"
```
petr-muller, line 454: Huh, hardcoding URLs looks fishy. At minimum, this should be named constant.

cmd/ci-operator/main.go
```diff
@@ -463,6 +465,20 @@ func (o *options) Complete() error {
...
 
 		o.clusterConfig = clusterConfig
+
+		coreGetter, err := coreclientset.NewForConfig(clusterConfig)
+		if err != nil {
+			return fmt.Errorf("could not get core client for cluster config: %v", err)
+		}
+		cm, err := coreGetter.ConfigMaps("openshift-console").Get("console-config", meta.GetOptions{})
+		if err != nil {
+			return fmt.Errorf("could not get ConfigMap console-config/openshift-console: %v", err)
```
petr-muller, line 475: If the L478 error is not blocking then this shouln't be either.

cmd/ci-operator/main.go
```diff
@@ -463,6 +465,20 @@ func (o *options) Complete() error {
 		}
 
 		o.clusterConfig = clusterConfig
+
+		coreGetter, err := coreclientset.NewForConfig(clusterConfig)
+		if err != nil {
+			return fmt.Errorf("could not get core client for cluster config: %v", err)
```
petr-muller, line 471: If the L478 error is not blocking then this shouln't be either.

cmd/ci-operator/main.go
```diff
@@ -463,6 +465,20 @@ func (o *options) Complete() error {
 		}
 
 		o.clusterConfig = clusterConfig
+
+		coreGetter, err := coreclientset.NewForConfig(clusterConfig)
```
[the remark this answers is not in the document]
hongkailiu: Done.

cmd/ci-operator/main.go
```diff
@@ -463,6 +465,20 @@ func (o *options) Complete() error {
...
+		if err != nil {
+			return fmt.Errorf("could not get core client for cluster config: %v", err)
+		}
+		cm, err := coreGetter.ConfigMaps("openshift-console").Get("console-config", meta.GetOptions{})
```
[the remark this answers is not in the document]
hongkailiu: Thought about this, too.

Need to figure out the protocol part https or http (after know of `host`).
I guess the route is not bound to the URL as tightly as the config, like i can create another route the mess things up (I know it is unlikely).

cmd/ci-operator/main.go
```diff
@@ -633,11 +649,41 @@ func (o *options) resolveInputs(steps []api.Step) error {
 	// TODO: instead of mutating this here, we should pass the parts of graph execution that are resolved
 	// after the graph is created but before it is run down into the run step.
 	o.jobSpec.Namespace = o.namespace
-	log.Printf("Using namespace %s", o.namespace)
+
+	if len(o.clusterConsoleURL) > 0 {
```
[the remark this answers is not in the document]
hongkailiu: Done.

cmd/ci-operator/main_test.go
hongkailiu: Done.

cmd/ci-operator/main_test.go
hongkailiu: Removed.

cmd/ci-operator/main_test.go
hongkailiu: Done.

cmd/ci-operator/main.go
hongkailiu: A new constant defined.

cmd/ci-operator/main.go
hongkailiu: Done.

cmd/ci-operator/main.go
hongkailiu: Done.

hongkailiu:
@petr-muller 

> One thing I don't like here is that it further (and for very little gain) limits people in running ci-operator locally against a cluster by requiring specific RBACs. 

You mean running it locally with `--dry-run=false`, right?
Did you hit the RBAC error? or yo are complaining on behalf of non-DPTP users?

With non-blocking errors for the whole feature, no limits should be introduced now.

petr-muller:
> @petr-muller
> 
> > One thing I don't like here is that it further (and for very little gain) limits people in running ci-operator locally against a cluster by requiring specific RBACs.
> 
> You mean running it locally with `--dry-run=false`, right?
> Did you hit the RBAC error? or yo are complaining on behalf of non-DPTP users?

Yeah, locally with `--dry-run=false` (which is still the default). I haven't encountered a problem myself; I haven't even tried, maybe my assumption was wrong :). The _ If we can resolve the field, use it. If not, don't._ behavior seems to be a good approach, though.

cmd/ci-operator/main.go
```diff
@@ -625,11 +627,70 @@ func (o *options) resolveInputs(steps []api.Step) error {
...
+func getConsoleURL(cm *coreapi.ConfigMap) (string, error) {
+	if cm == nil {
+		return "", fmt.Errorf("nil config map")
+	}
+	consoleConfigString, ok := cm.Data["console-config.yaml"]
+	if !ok {
+		return "", fmt.Errorf("no key console-config.yaml in the data map")
+	}
+	consoleConfig := &ConsoleConfig{}
+	if err := yaml.Unmarshal([]byte(consoleConfigString), consoleConfig); err != nil {
```
alvaroaleman, line 678: Re the route discussion, I think we can safely assume that the protocol will always be https, hence the route gives us everything we need

hongkailiu: Via `route.spec.host now`.

alvaroaleman:
@stevekuznetsov 

> I don't think we need a flag, or default behavior. If we can resolve the field, use it. If not, don't.

> We already have clientsets and configmap clients, why are we adding controller runtime here? > Please don't.

Currently our codebase is full with a lot of typed clients. Since this PR adds a new client anyways I suggested to use the controller-runtime one with the hope that it will gradually replace all the typed clients over time. WDYT?

stevekuznetsov:
Lets make that separate. All that will achieve right now is a duplicate effort. It's orthogonal to the current issue.

cmd/ci-operator/main.go
```diff
@@ -625,7 +626,24 @@ func (o *options) resolveInputs(steps []api.Step) error {
 	// TODO: instead of mutating this here, we should pass the parts of graph execution that are resolved
 	// after the graph is created but before it is run down into the run step.
 	o.jobSpec.Namespace = o.namespace
-	log.Printf("Using namespace %s", o.namespace)
+
+	//If we can resolve the field, use it. If not, don't.
+	var consoleHost string
+	routeGetter, err := routeclientset.NewForConfig(o.clusterConfig)
+	if err != nil {
+		log.Printf("[WARN] could not get route client for cluster config: %v", err)
```
stevekuznetsov, line 634: Don't use formatters like `[WARN]` and don't spit errors into the developer's log.

cmd/ci-operator/main.go
```diff
@@ -625,7 +626,24 @@ func (o *options) resolveInputs(steps []api.Step) error {
...
+	}
+	consoleRoute, err := routeGetter.Routes("openshift-console").Get("console", meta.GetOptions{})
```
stevekuznetsov, line 636: If the `err` was not `nil`, you should not continue on to these steps

cmd/ci-operator/main.go
```diff
@@ -625,7 +626,24 @@ func (o *options) resolveInputs(steps []api.Step) error {
...
+	if err != nil {
+		log.Printf("[WARN] could not get route console in namespace openshift-console: %v", err)
+	}
+	consoleHost = consoleRoute.Spec.Host
```
stevekuznetsov, line 640: similarly

cmd/ci-operator/main.go
hongkailiu: Removed.

cmd/ci-operator/main.go
hongkailiu: Done.

hongkailiu:
applied the role and the rolebinding manually on api.ci, and then

```
oc --context build01 adm policy who-can get routes console -n openshift-console | grep ci-op
        system:serviceaccount:ci:ci-operator

oc adm policy who-can get routes console -n openshift-console | grep ci-op 
        system:serviceaccount:ci:ci-operator
```

/hold cancel

PR #681, opened 2020-04-18
ci-operator: annotate errors at the top-most levels

stevekuznetsov:
In order to be able to expose a history over time of known, annotated
reasons for failure in the ci-operator process, we need a mechanism to
communicate these errors, with annotations, thoughout the codebase. In
order to allow administrators to apply default policy while also
allowing for specific error causes, a concatentating mechanism is used
in defining the error reason string. This will allow for queries for a
specific subsystem or a symptom that may occur across subsystems by
using regular expression matching on these error strings.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -57,6 +57,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/junit"
 	"github.com/openshift/ci-tools/pkg/lease"
 	"github.com/openshift/ci-tools/pkg/load"
+	"github.com/openshift/ci-tools/pkg/results"
 	"github.com/openshift/ci-tools/pkg/steps"
 	"github.com/openshift/ci-tools/pkg/util"
 )
@@ -178,12 +179,21 @@ func main() {
 	})
 
 	if err := opt.Complete(); err != nil {
+		err = results.ForReason(results.ReasonLoadingArgs).ForError(err)
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		opt.writeFailingJUnit(err)
 		os.Exit(1)
 	}
 
 	if err := opt.Run(); err != nil {
+		// TODO: maybe we can think of a better way to do this,
+		// but it would be awesome to not need to make the results
+		// library aware of this typed error
+		baseErr := err
+		if wrapped, ok := err.(errWroteJUnit); ok {
+			baseErr = wrapped.error
+		}
+		err = results.DefaultReason(baseErr)
 		fmt.Fprintf(os.Stderr, "error: %v\n", err)
 		opt.writeFailingJUnit(err)
 		os.Exit(1)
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
FYI more granular errors will be added over time, this is just the library and some uses.

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Error makes an Error an error
+func (e Error) Error() string {
+	return e.message
+}
+
+// FullReason provides the chain of error reasons, divided by colons
+func (e Error) FullReason() string {
+	reasonedError, ok := e.wrapped.(*Error)
```
alvaroaleman, line 26: lets use `errors.As` here  (requires to implement `Unwrap` on the type). Also, some funcs have value and some have pointer receivers, please stick to one of the two

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+// }
+type Error struct {
+	reason  Reason
+	message string
+	wrapped error
+}
+
+// Error makes an Error an error
+func (e Error) Error() string {
+	return e.message
```
alvaroaleman, line 21: This can lead to non-nil errors that have an empty message. Please print the reason here and only append the `message` if its not an empty string

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+	if !ok {
+		return string(e.reason)
+	}
+	return fmt.Sprintf("%s:%s", e.reason, reasonedError.FullReason())
+}
+
+// FullReason attempts to get the full reason from an error, or uses
+// unknown when it's not something we can do
+func FullReason(err error) string {
+	reasonedError, ok := err.(*Error)
```
alvaroaleman, line 36: `errors.As`

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+		// we enforce that there's some default (if useless) value
+		reason = ReasonUnknown
+	}
+	return &Error{
+		reason: reason,
+	}
+}
+
+// WithError is a builder that adds a child to the Error
+func (e *Error) WithError(err error) *Error {
```
alvaroaleman, line 56: Please call this `SetWrappedError` to clarify that any previously existing error will be overwritten and that the passed in error gets wrapped

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// Errorf is a builder that adds in the main error to an Error
+func (e *Error) Errorf(format string, args ...interface{}) error {
+	e.message = fmt.Sprintf(format, args...)
+	return e
+}
+
+// ForError is a constructor for when a caller does not want to add
+func (e *Error) ForError(err error) error {
```
alvaroaleman, line 68: Its completely unclear from the godoc that and why `WithErrors` sets the wrapped error whereas this sets the `message`. Please call this `SetMessageFromError`. And also, return an `*Error`

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+
+// WithError is a builder that adds a child to the Error
+func (e *Error) WithError(err error) *Error {
+	e.wrapped = err
+	return e
+}
+
+// Errorf is a builder that adds in the main error to an Error
+func (e *Error) Errorf(format string, args ...interface{}) error {
+	e.message = fmt.Sprintf(format, args...)
```
alvaroaleman, line 63: Please call it `SetMessagef`, its unclear from the godocs what this does. Also, why does this return an `error` and not an `*Error`?

pkg/results/error.go
```diff
@@ -0,0 +1,83 @@
...
+func (e *Error) ForError(err error) error {
+	if err == nil {
+		return nil
+	}
+	e.message = err.Error()
+	return e
+}
+
+// DefaultReason is a constructor that adds a reason if needed
+func DefaultReason(err error) error {
```
alvaroaleman, line 77: I don't see the value of this func. It guarantees you there is a reason. We can instead just at callsites iterate through the wrapping and return `ReasonUnknown` if an error chain contains no `Error` from this package

pkg/results/error.go
stevekuznetsov: It's the producer, so you end up with an `error` as you expect

pkg/results/error.go
stevekuznetsov: We are not guaranteed to have an `Error`, this assures that, too

pkg/results/error.go
stevekuznetsov: Only if people are not using the builder chain correctly. That's not the point of this field.

stevekuznetsov:
@alvaroaleman @petr-muller updated to use `errors.{Unwrap|Is|As}()`

pkg/results/error.go
alvaroaleman: Great, the same argument can be made for every single other public function in this package. Please make all of them either return an `error`, `*Error`, or `Error`.

You also completely ignored the first part of my message, without reading the implementation its completely unclear what this does

cmd/ci-operator/main.go
```diff
@@ -178,12 +179,21 @@ func main() {
...
 
 	if err := opt.Run(); err != nil {
+		// TODO: maybe we can think of a better way to do this,
+		// but it would be awesome to not need to make the results
+		// library aware of this typed error
+		baseErr := err
+		if wrapped, ok := err.(errWroteJUnit); ok {
+			baseErr = wrapped.error
+		}
+		err = results.DefaultReason(baseErr)
```
smarterclayton, line 196: This causes rudundant junit reporting to test grid and all our infrastructure - if you see errWroteJunit you don't write a second junit.  I believe you wanted to set a bool and then skip line 188 unless you deliberately wanted to double write junits.

smarterclayton: image

I can't tell whether this PR intended to always double report the junit.  If you did, we now have to hide the duplication in more places - was there a reason not to just use the existing one?

smarterclayton: As discussed on chat now in #735

PR #685, opened 2020-04-21
Validate we don't use the `default` cluster alias and enable gzip

alvaroaleman:
/assign @stevekuznetsov

### DIFF
--- a/cmd/config-shard-validator/main.go
+++ b/cmd/config-shard-validator/main.go
@@ -9,6 +9,7 @@ import (
 	"path/filepath"
 
 	"k8s.io/api/core/v1"
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
 	_ "k8s.io/test-infra/prow/hook"
 	"k8s.io/test-infra/prow/plugins"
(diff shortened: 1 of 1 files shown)


cmd/config-shard-validator/main.go
```diff
@@ -123,6 +124,15 @@ func main() {
 		var matchedMap string
 		logger := logrus.WithField("source-file", pathToCheck.path)
 		for glob, updateConfig := range pcfg.ConfigUpdater.Maps {
+			if _, hasDefaultCluster := updateConfig.Clusters[prowv1.DefaultClusterAlias]; hasDefaultCluster {
+				logger.Errorf("cluster alias %s is not allowed, please explicitly specific the name of the cluster", prowv1.DefaultClusterAlias)
+				foundFailures = true
+			}
+			if updateConfig.GZIP == nil || !*updateConfig.GZIP {
```
stevekuznetsov, line 131: currently we don't zip everything, only jobs -- need to either make this check conditional or fix the config first

cmd/config-shard-validator/main.go
```diff
@@ -123,11 +124,23 @@ func main() {
...
+				logger.Errorf("cluster alias %s is not allowed, please explicitly specific the name of the cluster", prowv1.DefaultClusterAlias)
+				foundFailures = true
+			}
+
 			globLogger := logger.WithField("glob", glob)
 			matches, matchErr := zglob.Match(glob, pathToCheck.path)
 			if matchErr != nil {
 				globLogger.WithError(matchErr).Warn("Failed to check glob match.")
 			}
+			if jobConfigMatch, err := zglob.Match(glob, "ci-operator/config"); err != nil {
```
stevekuznetsov, line 137: `ci-operator/jobs` not `/config`

PR #686, opened 2020-04-22
ci-operator: add optional, best-effort reporting of result

stevekuznetsov:

/assign @ingvagabund @alvaroaleman

### DIFF
--- a/pkg/results/report.go
+++ b/pkg/results/report.go
@@ -1,0 +1,149 @@
+package results
+
+import (
+	"bytes"
+	"crypto/tls"
+	"crypto/x509"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/golang/glog"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+// Options holds the configuration options for connecting to the remote aggregation server
+type Options struct {
+	address  string
+	certFile string
+	keyFile  string
+	caFile   string
+}
+
+// Bind adds flags for the options
+func (o *Options) Bind(flag *flag.FlagSet) {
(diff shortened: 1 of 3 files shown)


pkg/results/report.go
```diff
@@ -0,0 +1,144 @@
...
+type Options struct {
+	address  string
+	certFile string
+	keyFile  string
+	caFile   string
+}
+
+// Bind adds flags for the options
+func (o *Options) Bind(flag *flag.FlagSet) {
+	flag.StringVar(&o.address, "report-address", "", "Address of the aggregate reporting server.")
```
alvaroaleman, line 29: Why? We can just use a specific log line or even junit and find it via ci-search?

alvaroaleman: If we are building an aggregation service I'd like to see some design first

stevekuznetsov: Design has been hashed out pretty well at this stage.

petr-muller: Where?

pkg/results/report.go
```diff
@@ -0,0 +1,144 @@
...
+type Request struct {
+	// JobName is the name of the job for which a result is being reported
+	JobName string `json:"job_name"`
+	// Type is the type of job ("presubmit", "postsubmit", "periodic" or "batch")
+	Type string `json:"type"`
+	// State is "succeeded" or "failed"
+	State string `json:"state"`
+	// Reason is a colon-delimited list of reasons for failure
+	Reason string `json:"reason"`
+}
```
ingvagabund, line 93: Cluster field is missing

stevekuznetsov: Yeah, I thought about that one -- harder to do. Will attempt to do better now.

PR #688, opened 2020-04-22
deprovision a libvirt cluster

clnperez:
this is for the p system currently testing the 4.3 releases. the plan is
to be able to set additional ips if we have different bastions. we could
add an outer for-loop for that in the future if needed.

the journaclctl way of checking for age/existence in time is there because
there is no way to get the age of a libvirt resource unless metadata was
added to the resource upon creation.

Signed-off-by: Christy Norman <christy@linux.vnet.ibm.com>

### DIFF
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -53,6 +53,26 @@ EOF
   echo "will deprovision GCE cluster ${infraID} in region ${region}"
 done
 
+libvirt_p_cluster_age_cutoff="4 hours ago"
+echo "deprovisioning clusters with creation time before ${libvirt_p_cluster_age_cutoff} in libvirt-p"
+for network in $( ssh "${bastion_p}" virsh net-list --all --name | grep -v default ) ; do
+	if $( ssh "${bastion_p}" journalctl -u libvirtd.service --until "${libvirt_p_cluster_age_cutoff}" 2>/dev/null | grep "${network}" ) ; then
+    infraID="${network}"
+    workdir="/tmp/deprovision/${infraID}"
+    mkdir -p "${workdir}"
+    cat <<EOF >"${workdir}/metadata.json"
+{
+  "infraID":"${infraID}",
+  "libvirt":{
+    "uri":"qemu+tcp://${bastion_p}/system"
+  }
+}
+EOF
+  echo "will deprovision libvirt cluster ${infraID} on p host ${bastion_p}"
+  fi
+done
+
+
 for workdir in $( find /tmp/deprovision -mindepth 1 -type d | shuf ); do
   timeout 30m openshift-install --dir "${workdir}" --log-level debug destroy cluster
 done


clnperez:
@stevekuznetsov -- this is a first pass at this. i'm not sure exactly how to go about integrating the bastion with this script. this approach is kind of a guess and might be impossible. thoughts?

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -53,6 +53,24 @@ EOF
   echo "will deprovision GCE cluster ${infraID} in region ${region}"
 done
 
+libvirt_p_cluster_age_cutoff="$(TZ=":America/New_York" date --date="-11 hours" '+%Y-%m-%dT%H:%M%z')"
+echo "deprovisioning clusters with creation time before ${libvirt_p_cluster_age_cutoff} in libvirt-p"
+for network in $(ssh ${bastion_p} virsh net-list --all --name | grep -v default) ; do
+        if [[ ! -z $(ssh ${bastion_p} journalctl -u libvirtd.service --until \"11 hours ago\" 2>/dev/null | grep ${network}) ]] ; then
```
stevekuznetsov, line 59: nit: we're using two spaces as indentation in the rest of the file, can we please follow that format

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -53,6 +53,24 @@ EOF
   echo "will deprovision GCE cluster ${infraID} in region ${region}"
 done
 
+libvirt_p_cluster_age_cutoff="$(TZ=":America/New_York" date --date="-11 hours" '+%Y-%m-%dT%H:%M%z')"
```
stevekuznetsov, line 56: Why should `libvirt` clusters hang around for 11 hours? Is their job timeout not 4?

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -53,6 +53,24 @@ EOF
...
+echo "deprovisioning clusters with creation time before ${libvirt_p_cluster_age_cutoff} in libvirt-p"
+for network in $(ssh ${bastion_p} virsh net-list --all --name | grep -v default) ; do
```
stevekuznetsov, line 58: What sets `"${bastion_p}"`?

cmd/ipi-deprovision/ipi-deprovision.sh
stevekuznetsov, line 59: You are trying to check if the log contains the network? You also want to actually make use of the time cutoff variable:


```sh
if ssh "${bastion_p}" -- journalctl --unit libvirtd.service --until "${libvirt_p_cluster_age_cutoff}" 2>&/dev/null | grep -q "${network}"; then
```

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -53,6 +53,24 @@ EOF
...
+        if [[ ! -z $(ssh ${bastion_p} journalctl -u libvirtd.service --until \"11 hours ago\" 2>/dev/null | grep ${network}) ]] ; then
+                infraID=${network}
+                workdir="/tmp/deprovision/${infraID}"
+                mkdir -p "${workdir}"
+                cat <<EOF >"${workdir}/metadata.json"
+{
```
stevekuznetsov, line 64: Please format this similarly to the others

```json
{
  "infraID":"${infraID}",
  "libvirt": {
    "uri":"qemu+tcp://${bastion_p}/system"
  }
}
```

cmd/ipi-deprovision/ipi-deprovision.sh
stevekuznetsov: nit: quote your variable expansions

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez: oh yah, definitely meant to fix that. i was focusing more on big-picture "will this work" with this submission.

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez: yah, 4's better. i won't get into why i picked 11 to start but will change.

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez: i was hoping that'd be something we could add to the cleanup run. is that possible? i'm not familiar with how these values get set. i noticed that `CLUSTER_TTL` was set somewhere.

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez: yes, that's the only way i can think of to know how old something is (if it was created before a certain time). 

and i can double-check, but i think i tried using that, but iirc the timestamp format's different than what journalctl expects -- so i was just using that var as a logging output. `--until` is so much simpler, it's harder to mess up, so I just went with that. do you mind if i keep that strategy?

clnperez:
@stevekuznetsov thanks for the first review. another thing i'm not sure about with the approach is, will those ssh calls get to the right place? iow -- will the ssh to the bastion ip call journalctl on the libvirt host?

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez:
```
journalctl -u libvirtd --until $libvirt_p_cluster_age_cutoff
Failed to parse timestamp: 2020-04-23T04:48-0400
```

cmd/ipi-deprovision/ipi-deprovision.sh
stevekuznetsov: Sure, we could -- is it a static IP?

cmd/ipi-deprovision/ipi-deprovision.sh
stevekuznetsov: The format of `$libvirt_p_cluster_age_cutoff` is entirely up to you -- I hope we could use it (and format it correctly) -- it is very useful to see the value about to be used printed in the logs.

stevekuznetsov: I'm also way more worried about the Bash on this line than the format of the time -- the `[[ ! -z` is not following best practices, please try out my suggestion.

stevekuznetsov:
```
ssh $host -- $commands
```

Should do what you want, but would be  best to test it :)

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez: re time cutoff format: so can i just set it to `4 hours ago` then, since that's what journalctl is using?
re bash: didn't catch that you were talking about two things there. i'll fix that, thanks.

cmd/ipi-deprovision/ipi-deprovision.sh
clnperez: yah it's static. for the ci it's hard-coded into the release job:
https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-release-4.3-periodics.yaml#L6912

clnperez:
i'm not quite sure how to test it locally. i was assuming that only the prow jobs could ssh to that ip. (i'm talking about the ip of the sshd bastion we set up, in case that wasn't obvious before)

clnperez:
I think that for this to work we would have to modify the sshd bastion to listen on a port that we then include in the ssh tunnel created on the ci host(s). Right now, if I ssh from inside the pod like : 
```
    - /usr/local/bin/mock-nss.sh
    - /bin/sh
    - -c
    - |
      #!/bin/sh
      ssh -T -v root@<internal-pod-ip> -p 2222 -i /.ssh/id_rsa journalctl \
         -u libvirtd.service --until "4 hours ago"
```

I get a verification error: 

```debug1: Reading configuration data /etc/ssh/ssh_config.d/05-redhat.conf
debug1: Reading configuration data /etc/crypto-policies/back-ends/openssh.config
debug1: configuration requests final Match pass
debug1: re-parsing configuration
debug1: Reading configuration data /etc/ssh/ssh_config
debug1: Reading configuration data /etc/ssh/ssh_config.d/05-redhat.conf
debug1: Reading configuration data /etc/crypto-policies/back-ends/openssh.config
debug1: Connecting to 172.16.29.28 [172.16.29.28] port 2222.
debug1: Connection established.
debug1: identity file /.ssh/id_rsa type 0
debug1: identity file /.ssh/id_rsa-cert type -1
debug1: Local version string SSH-2.0-OpenSSH_8.0
debug1: Remote protocol version 2.0, remote software version OpenSSH_7.4
debug1: match: OpenSSH_7.4 pat OpenSSH_7.0*,OpenSSH_7.1*,OpenSSH_7.2*,OpenSSH_7.3*,OpenSSH_7.4*,OpenSSH_7.5*,OpenSSH_7.6*,OpenSSH_7.7* compat 0x04000002
debug1: Authenticating to 172.16.29.28:2222 as 'root'
debug1: SSH2_MSG_KEXINIT sent
debug1: SSH2_MSG_KEXINIT received
debug1: kex: algorithm: curve25519-sha256
debug1: kex: host key algorithm: ecdsa-sha2-nistp256
debug1: kex: server->client cipher: aes256-gcm@openssh.com MAC: <implicit> compression: none
debug1: kex: client->server cipher: aes256-gcm@openssh.com MAC: <implicit> compression: none
debug1: kex: curve25519-sha256 need=32 dh_need=32
debug1: kex: curve25519-sha256 need=32 dh_need=32
debug1: expecting SSH2_MSG_KEX_ECDH_REPLY
debug1: Server host key: ecdsa-sha2-nistp256 SHA256:V3v9eJLf57wfDRc2td4Vv++G53OFIgnFws3350ok7Ek
debug1: checking without port identifier
debug1: read_passphrase: can't open /dev/tty: No such device or address
Host key verification failed.
```

The levels of port-forward and reverse-tunneling and NAT are making my head hurt a little. :D

clnperez:
would like to figure this one out eventually

/remove-lifecycle stale

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -53,6 +53,26 @@ EOF
   echo "will deprovision GCE cluster ${infraID} in region ${region}"
 done
 
+libvirt_p_cluster_age_cutoff="4 hours ago"
+echo "deprovisioning clusters with creation time before ${libvirt_p_cluster_age_cutoff} in libvirt-p"
+for network in $( ssh "${bastion_p}" virsh net-list --all --name | grep -v default ) ; do
+	if $( ssh "${bastion_p}" journalctl -u libvirtd.service --until "${libvirt_p_cluster_age_cutoff}" 2>/dev/null | grep "${network}" ) ; then
```
hamzy, line 59: If I am reading that line correctly, the journal log will continue growing until it gets rotated via settings in /etc/systemd/journald.conf 
I hope those are configured so that a rotation won't occur in the middle of a CI run.

PR #690, opened 2020-04-23
Result aggregator: new server for collecting error messages from ci-operator and exposing error metrics for Prometheus

ingvagabund:
New server for collecting error messages from ci-operator and exposing error metrics for Prometheus

### DIFF
--- a/cmd/result-aggregator/main.go
+++ b/cmd/result-aggregator/main.go
@@ -1,0 +1,147 @@
+package main
+
+import (
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"time"
+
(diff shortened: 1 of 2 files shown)


cmd/result-aggregator/main.go
```diff
@@ -0,0 +1,119 @@
...
+
+		withErrorRate(request)
+
+		w.WriteHeader(http.StatusOK)
+
+		logrus.Infof("Request with %#v request processed", request)
+	})
+}
+
+func main() {
```
stevekuznetsov, line 110: mind using the `interrupts` package from Prow to start the server, provide health + readiness endpoints and serve prom metrics?

ingvagabund: How do I access the metrics with `interrupts`?

PR #696, opened 2020-04-23
ConfigAgent: Allow adding indexers

alvaroaleman:
Needed for the ImageStreamController who needs to index configs by ImageStreamTag

/assign @stevekuznetsov

### DIFF
--- a/pkg/load/agents/configAgent.go
+++ b/pkg/load/agents/configAgent.go
@@ -24,7 +24,12 @@ type ConfigAgent interface {
 	GetConfig(config.Info) (api.ReleaseBuildConfiguration, error)
 	GetAll() FilenameToConfig
 	GetGeneration() int
+	AddIndex(indexName string, indexFunc IndexFn) error
+	GetFromIndex(indexName string, indexKey string) ([]*api.ReleaseBuildConfiguration, error)
 }
+
+// IndexFn can be used to add indexes to the ConfigAgent
+type IndexFn func(api.ReleaseBuildConfiguration) []string
 
 type FilenameToConfig map[string]api.ReleaseBuildConfiguration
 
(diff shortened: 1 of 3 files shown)


pkg/load/agents/configAgent.go
```diff
@@ -35,8 +40,12 @@ type configAgent struct {
 	cycle        time.Duration
 	generation   int
 	errorMetrics *prometheus.CounterVec
+	indexFuncs   map[string]IndexFn
+	indexes      map[string]configIndex
 }
 
+type configIndex map[string][]*api.ReleaseBuildConfiguration
```
stevekuznetsov, line 47: By making these pointers I wonder if we'll run into mutability issues

pkg/load/agents/configAgent_test.go
```diff
@@ -0,0 +1,174 @@
...
+			},
+			configs:  FilenameToConfig{"myfile.yaml": cfg},
+			expected: map[string]configIndex{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.agent.indexes = tc.agent.buildIndexes(tc.configs)
+			if diff := deep.Equal(tc.agent.indexes, tc.expected); diff != nil {
```
stevekuznetsov, line 169: we usually use `cmp.Diff()` -- why is this better?

pkg/load/agents/configAgent.go
alvaroaleman: Even if we returned this by value we had the issues because it contains pointer fields. The thinking was that if we have that problem anyways, we might as well save us the allocation.

pkg/load/agents/configAgent_test.go
alvaroaleman: Updated the pull to use `cmp.Diff`

PR #697, opened 2020-04-24
[Whitelist] allow config to be picked if the org repo exists in the whitelist

droslean:
The whitelist struct is intended to be filled up from a YAML file in the following format
```yaml
whitelist:
  openshift:
    - repo1
    - repo2
  testshift:
    - repo1
    - repo2
    - repo3
```
This file should live in the release repository and the tools can use it while we will control its values from a single place. 

Changes in this PR:

- Introduce `whitelistConfig` in the whitelist package including tests
- Adds the whitelist config (options only) to `private-org-sync` tool
- Adds the whitelist config to `ci-operator-config-mirror` tool. Update integration tests, and add a `disable: true` to the non-ocp promotion configuration 
- Adds the whitelist config (options only) to `private-prow-configs-mirror` tool
- Adds the whitelist config (options only) to `private-org-peribolos-sync` tool, removes the `--whitelist` flag with its related functions and tests. 

/cc @openshift/openshift-team-developer-productivity-test-platform @petr-muller 


pkg/promotion/promotion.go
```diff
@@ -36,6 +43,21 @@ func BuildsOfficialImages(configSpec *cioperatorapi.ReleaseBuildConfiguration) b
 	return RefersToOfficialImage(promotionName, promotionNamespace)
 }
 
+func isWhiteListed(org, repo string, whitelist WhitelistConfig) bool {
```
petr-muller, line 46: This is basically a method, so why not make it one: `func (w *WhitelistConfig) has(org, repo string) bool`

pkg/promotion/promotion.go
```diff
@@ -36,6 +43,21 @@ func BuildsOfficialImages(configSpec *cioperatorapi.ReleaseBuildConfiguration) b
...
+	if whiteRepos, ok := whitelist.Whitelist[org]; ok {
+		if sets.NewString(whiteRepos...).Has(repo) {
```
petr-muller, line 48: Uh, converting the list to a set on each call looks wasteful. Can we have a `Load` method that does it once on load time?

cmd/private-org-sync/main.go
```diff
@@ -24,9 +26,10 @@ import (
 )
 
 type options struct {
-	configDir string
-	tokenPath string
-	targetOrg string
+	configDir     string
+	tokenPath     string
+	targetOrg     string
+	whitelistFile string
```
stevekuznetsov, line 32: Hm -- we're adding a new option that filters the set of config we operate on? Could we follow the pattern set up for `promotion.Options` and `config.Options` that expose their own `OperateOnCIOperatorConfigDir`?

pkg/whitelist/whitelist.go
```diff
@@ -0,0 +1,81 @@
+package whitelist
```
stevekuznetsov, line 1: This should just be `package config` -- no need for a new package

pkg/whitelist/whitelist.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+	return whitelistConfig, nil
+}
+
+// OperateOnCIOperatorConfigDir filters the full set of configurations
+// down to those that exist in the whitelist configuration and build official images
+func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi.ReleaseBuildConfiguration, *config.Info) error, whitelistConfig WhitelistConfig) error {
+	return config.OperateOnCIOperatorConfigDir(configDir, func(configuration *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
+		if !promotion.BuildsOfficialImages(configuration) && !whitelistConfig.IsWhitelisted(info) {
```
stevekuznetsov, line 76: The whitelist should not make any decision about promotion, just the whitelist

pkg/whitelist/whitelist.go
```diff
@@ -0,0 +1,81 @@
...
+		if err := yaml.Unmarshal(bytes, &whitelistConfig); err != nil {
+			return WhitelistConfig{}, errors.New("Couldn't unmarshal whitelist configuration")
+		}
+	}
+	return whitelistConfig, nil
+}
+
+// OperateOnCIOperatorConfigDir filters the full set of configurations
+// down to those that exist in the whitelist configuration and build official images
+func OperateOnCIOperatorConfigDir(configDir string, callback func(*cioperatorapi.ReleaseBuildConfiguration, *config.Info) error, whitelistConfig WhitelistConfig) error {
```
stevekuznetsov, line 74: This should be a method on `(o *Options)` -- if someone is using a tool with a whitelist, it should be impossible for them to forget to add it.

pkg/whitelist/whitelist.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	return nil
+}
+
+func (o *Options) Bind(fs *flag.FlagSet) {
+	fs.StringVar(&o.whitelistFile, "whitelist-file", "", "File of the repository whitelist configuration")
+}
+
+// LoadWhitelist reads the whitelist configuration file and returns a WhitelistConfig object.
+func (o *Options) LoadWhitelist() (WhitelistConfig, error) {
```
stevekuznetsov, line 58: This should be done at validate-time and stored in the options

pkg/whitelist/whitelist.go
droslean: Hm, that would be really tricky.  Unless if we want to add all the repos that build official images in whitelist as well.

droslean:
I think @petr-muller has a point here. We have repositories that don't actually have a ci-operator configuration, and in this case, they will never be whitelisted. However, I believe the current whitelist implementation is fine if we exclude its `OperateOnCIOperatorConfigDir` method and let the tools to make that check, in their callbacks.

stevekuznetsov:
OK, I would prefer us to provide both. Expose a `OperateOnCIOperatorConfigDir` and use it, when that's what we need. Under the hood, re-use the whitelist checking functionality.

droslean:
> OK, I would prefer us to provide both. Expose a OperateOnCIOperatorConfigDir and use it, when that's what we need. Under the hood, re-use the whitelist checking functionality.

Yea that makes sense to implement since `ci-operator-config-mirror` needs that already

petr-muller:
/lgtm
/hold cancel

Thanks Nikos!

stevekuznetsov:
The last rebase here added a commit we did not need.

PR #699, opened 2020-04-24
ipi-deprovision: remove old buckets

stevekuznetsov:
4.2 clusters still leak these, so we need to remove them manually.


wking:
/lgtm

I don't understand the finer details, but if this breaks anything we can always fix it once we have job results to look at ;).

PR #702, opened 2020-04-24
ImageStreamReconciler: Initial working implementation

alvaroaleman:
This PR completes the ImageStreamReconciler to actually work.

Apart from whats in this PR we need:
* ~Some ratelimiting, because currently an imagestream change enqueues all referenced ImageStreams and thats _a lot_ (xref: https://github.com/openshift/ci-tools/pull/704)~ done
* Deployment manifests
* ~Something that triggers periodic reconciliation~ done

/assign @stevekuznetsov


cmd/dptp-controller-manager/main.go
```diff
@@ -74,17 +78,17 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatal("Failed to construct manager")
 	}
+	pjutil.ServePProf()
```
stevekuznetsov, line 81: Will this pull in `interrupts` package

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -185,15 +214,17 @@ type branchReference struct {
 
 func refForIST(ist *imagev1.ImageStreamTag) (*branchReference, error) {
 	imageMetadataLabels := struct {
-		Labels map[string]string `json:"labels"`
+		Config struct {
+			Labels map[string]string `json:"labels"`
+		} `json:"Config"`
 	}{}
 	if err := json.Unmarshal(ist.Image.DockerImageMetadata.Raw, &imageMetadataLabels); err != nil {
```
stevekuznetsov, line 221: There's a type for that metadata and we use it in `project_image.go`

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -116,6 +133,9 @@ func (r *reconciler) Reconcile(req controllerruntime.Request) (controllerruntime
 }
 
 func (r *reconciler) reconcile(req controllerruntime.Request, log *logrus.Entry) error {
```
stevekuznetsov, line 135: Lots of debugging messages here -- do we want to keep those in? If so, maybe publish metrics instead?

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -162,18 +189,20 @@ func (r *reconciler) reconcile(req controllerruntime.Request, log *logrus.Entry)
 	return nil
 }
 
-func (r *reconciler) isPublishedFromCIOperatorRepo(log *logrus.Entry, ist *imagev1.ImageStreamTag) (*cioperatorapi.ReleaseBuildConfiguration, bool) {
```
stevekuznetsov, line 165: (can't comment on the right line, just adding here)

You wrote:

```go
if istHasBuildRunning || err != nil {
	return wrapErrIfNonNil("failed to check if there is a running build for imageStreamTag", err)
}
```

You enter the `if` with two disjoint cases but then exit with two disjoint actions ... why is that better than

```go
if err != nil {
    return fmt.Errorf("failed to check if there is a running build for imageStreamTag", err)
}
if istHasBuildRunning {
    return nil
}
```

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -226,6 +257,8 @@ func (r *reconciler) currentHEADForBranch(br *branchReference) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("fauld to git rev-parse %s: %w", br.branch, err)
 	}
+	// RevParse output contains a trailing newline
+	branchHEADRef = strings.TrimSpace(branchHEADRef)
 
 	return branchHEADRef, nil
 }
```
stevekuznetsov, line 264: i love `getPublishJob` ... lol

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -246,9 +279,13 @@ func (r *reconciler) getPublishJob(br *branchReference, ciOPConfig *cioperatorap
...
 
 func (r *reconciler) isJobRunningForCommit(job *prowconfig.Postsubmit, commitSHA string, istRef *branchReference) (bool, error) {
+	jobNameLabelValue := job.Name
+	// Label values are capped at 63 characters and the job name frequently exceeds that.
```
stevekuznetsov, line 283: Shouldn't the generated job already have the label? Just use that one

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -246,9 +279,13 @@ func (r *reconciler) getPublishJob(br *branchReference, ciOPConfig *cioperatorap
...
+	if len(jobNameLabelValue) > 63 {
+		jobNameLabelValue = job.Name[0:62]
+	}
 	labelSelector := ctrlruntimeclient.MatchingLabels{
-		// Label values are capped at 63 characters and the job name frequently exceeds that.
-		kube.ProwJobAnnotation: job.Name[0:62],
+		kube.ProwJobAnnotation: jobNameLabelValue,
```
stevekuznetsov, line 288: do we seriously call this `annotation` :facepalm:

pkg/controller/image-stream-tag-reconciler/reconciler.go
stevekuznetsov: ah nvm you generate the config, not a pj

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -278,11 +315,12 @@ func (r *reconciler) createBuildForIST(job *prowconfig.Postsubmit, headSHA strin
...
 		BaseRef: istRef.branch,
 		BaseSHA: headSHA,
-	}), nil, nil)
+	}), r.createdProwJobLabels, nil)
 	prowJob.Namespace = r.prowJobNamespace
 
 	if r.dryRun {
-		r.log.Infof("Not creating %s prowjob because dryRun is enabled", prowJob.Spec.Job)
+		serialized, _ := json.Marshal(prowJob)
+		r.log.Infof("Not creating %s prowjob because dryRun is enabled, job: %s", prowJob.Spec.Job, serialized)
```
stevekuznetsov, line 323: If you add it as a `WithField` it will print valid JSON to the command line, whereas this will escape it, harder to copy-paste and read

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -311,3 +349,17 @@ type nonRetriableError struct {
 func (nre nonRetriableError) Error() string {
```
stevekuznetsov, line 349: `func nre(err error) error` seems superfluous

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -311,3 +349,17 @@ type nonRetriableError struct {
...
+func configIndexFn(in cioperatorapi.ReleaseBuildConfiguration) []string {
+	var result []string
+	for _, istRef := range release.PromotedTags(&in) {
+		result = append(result, fmt.Sprintf("%s/%s:%s", istRef.Namespace, istRef.Name, istRef.Tag))
+	}
+	return result
+}
+
+func configIndexKeyForIST(ist *imagev1.ImageStreamTag) string {
+	return ist.Namespace + "/" + ist.Name
```
stevekuznetsov, line 364: Why doesn't this use the tag?

cmd/dptp-controller-manager/main.go
alvaroaleman: Yes, why?

pkg/controller/image-stream-tag-reconciler/reconciler.go
alvaroaleman: I don't understand the comment, which tag do you mean?

pkg/controller/image-stream-tag-reconciler/reconciler.go
alvaroaleman: To be fair, its both an annotation and a label

pkg/controller/image-stream-tag-reconciler/reconciler.go
alvaroaleman: Is that irony? Its somewhat stupid and this should be part of some api package but it seems to work. It creates some implicit assumptions which is a bit unfortunate, so if you can come up with a better approach I am totally open for that.

cmd/dptp-controller-manager/main.go
stevekuznetsov: If you pull it in you may need to wait on the signal there, so this does not hang when it gets SIGINT

pkg/controller/image-stream-tag-reconciler/reconciler.go
stevekuznetsov: No I think your approach is appropriate (but sad we do not have something better)

pkg/controller/image-stream-tag-reconciler/reconciler.go
stevekuznetsov: `ist.Name` is the `name:tag`, disregard

alvaroaleman:
/hold
Needs https://github.com/kubernetes/test-infra/pull/17401 for  `git show-ref`, otherwise I can not check if this works

cmd/dptp-controller-manager/main.go
alvaroaleman: works on my machine

pkg/controller/image-stream-tag-reconciler/reconciler.go
```diff
@@ -46,21 +56,43 @@ func AddToManager(mgr controllerruntime.Manager, opts Options) error {
...
+			"openshift.io/created-by": controllerName,
+		},
+		ignoredGitHubOrganizations: sets.NewString(opts.IgnoredGitHubOrganizations...),
+		createdJobsCounter:         createdJobsCounter,
+	}
+	c, err := controller.New(controllerName, mgr, controller.Options{
+		// Since we watch ImageStreams and not ImageStreamTags as the latter do not support
+		// watch, we create a lot more events the needed. In order to decrease load, we coalesce
+		// and delay all requests after a successful reconciliation for up to an hour.
+		Reconciler: controllerutil.NewReconcileRequestCoalescer(r, time.Hour),
```
stevekuznetsov, line 91: I think we should look at the mean time taken to run a publication job and use that as the delay here, if we can -- does a change from 1h to 10m make a significant change in throughput?

alvaroaleman: Set it to 10mins, now that we use a GitHub client the worqueue depth goes down to zero after a minute or so

pkg/controller/image-stream-tag-reconciler/prowjobreconciler/prowjobreconciler.go
```diff
@@ -0,0 +1,248 @@
...
+	if isJobAlreadyRunning {
+		return nil
+	}
+
+	if err := r.client.Create(r.ctx, pj); err != nil {
+		return fmt.Errorf("failed to create prowjob: %w", err)
+	}
+	r.createdJobsCounter.WithLabelValues(orbc.Org, orbc.Repo, orbc.Branch).Inc()
+
+	if r.dryRun {
```
alvaroaleman, line 163: Oups. I might have created the promotion jobs now :)

pkg/controller/image-stream-tag-reconciler/prowjobreconciler/prowjobreconciler.go
```diff
@@ -0,0 +1,249 @@
...
+
+// Enqueuer allows the caller to Enqueue an OrgRepoBranchCommit
+type Enqueuer func(OrgRepoBranchCommit)
+
+const controllerName = "promotion_job_creator"
+
+func AddToManager(mgr controllerruntime.Manager, config config.Getter, dryRun bool) (Enqueuer, error) {
+	createdJobsCounter := prometheus.NewCounterVec(prometheus.CounterOpts{
+		Namespace: controllerName,
+		Name:      "prowjobs_created",
```
stevekuznetsov, line 47: Can we call this something specific to this controller?

alvaroaleman: it has the `Namespace` set to the `controllerName` so that is the case already

stevekuznetsov: AH

PR #704, opened 2020-04-24
Add a ReconcileRequestCoalescer

alvaroaleman:
This allows to delay ReconcileRequests after a successful Reconciliation
in order to decrease load for controllers that watch objects that change
very frequently.

Needed now for the ImageStreamTagReconciler but may be useful elsewhere.

To give a bit more background: ImageStreamTags do not support watching (https://github.com/openshift/api/issues/601). This means that:
* We have to watch ImageStreams instead
* And enqueue each and every ImageStreamTag in them on every event (basically whenever any changes)
* Since api.ci has some imageStreams that basically permanently get events, we produce _a lot_ of events
* This results in a high load because the ImageStreamTag reconciler has a somewhat expensive reconcile loop where it does a git fetch and some json unmarshaling

Apart from the approach suggested in this PR it would also be possible to filter the update events and check which ImageStreamTag in the ImageSteam actually got changed. This has some drawbacks though:
* The kube api does not guarantee that a given change will be observed, it only guarantees that watchers get any event per change, but that event may not contain the given update
* We want to periodically reconcile which is done internally by generating artifical update events which we would swallow with that approach. We could workaround that as well by creating a ticker that directly enqueues all imageStreamTags once every interval but its obviously a bit more involved.

Considering that we never rebuilt anything on missed events, introducing some delay to reduce load is IMHO good enough.

/assign @petr-muller

### DIFF
--- a/pkg/controller/util/request_coalescer.go
+++ b/pkg/controller/util/request_coalescer.go
@@ -1,0 +1,56 @@
+package util
+
+import (
+	"sync"
+	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// NewReconcileRequestCoalescer returns a reconcile wrapper that will delay new reconcile.Requests
+// after a sucessful reconciliation for coalesceWindow time.
+// A successful reconciliation is defined as as one where all returned values are empty or nil
+func NewReconcileRequestCoalescer(upstream reconcile.Reconciler, coalesceWindow time.Duration) reconcile.Reconciler {
+	return &reconcileRequestCoalescer{
+		lock:                              &sync.RWMutex{},
+		lastSuccessfulReconciliationCache: map[string]time.Time{},
+		coalesceWindow:                    coalesceWindow,
+		upstream:                          upstream,
+		timeSince:                         time.Since,
+	}
+}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/util/request_coalescer.go
```diff
@@ -0,0 +1,56 @@
...
+	// Stamp before acquiring the lock as that might take some time
+	successfulReconcileTime := time.Now()
+	rc.lock.Lock()
+	rc.lastSuccessfulReconciliationCache[r.String()] = successfulReconcileTime
+	rc.lock.Unlock()
+
+	return result, err
+}
+
+func IsReconcileSuccessfull(result reconcile.Result, err error) bool {
```
stevekuznetsov, line 54: Surprised that the library doesn't expose something like this

alvaroaleman: Well this is only needed because we have a massive amount of corner cases in this controller:
* We act on a resource that does not support watch, so we have to watch something else which generates more events than needed
* The number of resources we act on is huge
* Our reconciliation is expensive even if its a no-op

If any  of those three points were different, we wouldn't need this.
/hold cancel

PR #708, opened 2020-04-27
Improve ci-secret-bootstrap error handling

alvaroaleman:


cmd/ci-secret-bootstrap/main.go
```diff
@@ -93,7 +93,8 @@ func (o *options) completeOptions(secrets *sets.String) error {
 
 	kubeConfigs, _, err := util.LoadKubeConfigs(o.kubeConfigPath)
 	if err != nil {
-		return err
+		// We will bail out later on if we don't have the have the right kubeconfigs
+		logrus.WithError(err).Warn("Encountered errors while loading kubeconfigs")
```
hongkailiu, line 97: If the config is wrong, why do we continue using it?

alvaroaleman: It is not wrong, it means that there was an error loading one config from the kubeconfig. We don't know if that config was one this tool cares about at all. We will fail later on if a context we care about does not exist. This tool shouldn't be failing just because there is any context in the kubeconfig it can't use.

In my example I have an env var that references multiple kubeconfigs and if they are loaded together, it works perfectly fine. However this tool defaults to using only one and finds a context referenced in it that is in a different kubeconfig it doesn't know anything about so it bails.

hongkailiu: I think i know ... the wrong config could be with a context that we do not care.

PR #711, opened 2020-04-28
Fix default postsubmit.MaxConcurrency = 1 for promotion jobs

hongkailiu:
Follow up https://coreos.slack.com/archives/GB7NB0CUC/p1588003186030600?thread_ts=1588002752.029100&cid=GB7NB0CUC

Breaking change ... Need to prepare the PR for release change.
Fix tests if breaking.

Update:
Added unit tests and https://github.com/openshift/release/pull/8610

### DIFF
--- a/pkg/jobconfig/files.go
+++ b/pkg/jobconfig/files.go
@@ -2,7 +2,6 @@ package jobconfig
 
 import (
 	"fmt"
-	"github.com/openshift/ci-tools/pkg/config"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -16,6 +15,9 @@ import (
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	prowconfig "k8s.io/test-infra/prow/config"
+
+	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
 )
 
 type ProwgenLabel string
@@ -429,7 +431,9 @@ func mergePresubmits(old, new *prowconfig.Presubmit) prowconfig.Presubmit {
 func mergePostsubmits(old, new *prowconfig.Postsubmit) prowconfig.Postsubmit {
 	merged := *new
 
-	merged.MaxConcurrency = old.MaxConcurrency
+	if _, ok := merged.Labels[cioperatorapi.PromotionJobLabelKey]; !ok {
+		merged.MaxConcurrency = old.MaxConcurrency
+	}
 	if old.Cluster != "" {
 		merged.Cluster = old.Cluster
 	}
--- a/pkg/jobconfig/files_test.go
+++ b/pkg/jobconfig/files_test.go
@@ -508,6 +508,63 @@ func TestMergePostsubmits(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "job with promotion label and MaxConcurrency 1",
+			old: &prowconfig.Postsubmit{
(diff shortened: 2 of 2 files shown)


alvaroaleman:
After we already failed once to implement this correctly, we should definitely add tests for this

pkg/jobconfig/files.go
```diff
@@ -414,7 +414,9 @@ func mergePresubmits(old, new *prowconfig.Presubmit) prowconfig.Presubmit {
 	merged.AlwaysRun = old.AlwaysRun
 	merged.RunIfChanged = old.RunIfChanged
 	merged.Optional = old.Optional
-	merged.MaxConcurrency = old.MaxConcurrency
+	if old.MaxConcurrency != 0 {
```
alvaroaleman, line 417: Sorry, I don't understand this. We need the setting only for postsubmits and postsubmits must have this set to `1` as we only have build jobs as postsubmits. Why the check?

hongkailiu: Is this up to date?
https://github.com/openshift/ci-tools/blob/master/GENERATOR.md#hand-edited-prow-configuration

It did mention both presubmits and postsubmits.
However, periodic is not there.

Our prow-gen sets 1 only for the promotion job. We could have other postsubmits, right?

https://github.com/openshift/ci-tools/blob/9f3e703ba3e72ad422e61e105b95de682cb092a1/pkg/prowgen/prowgen.go#L175-L180

hongkailiu:
After https://github.com/openshift/release/pull/8610 merges, this one should not be breaking any more.

pkg/jobconfig/files.go
alvaroaleman: Right, we could have other postsubmits but I think there is no reason why a postsubmit should ever not run with max_concurrency: 1.

hongkailiu: It is the case for the current jobs.
This returns nothing.
> grep -irn max_ ./ci-operator/jobs | grep -v presubmits.yaml | grep -v "max_concurrency: 1"

Let me do it.

hongkailiu:
/test breaking-changes

```
ll ci-operator/jobs/redhat-operator-ecosystem/playground/redhat-operator-ecosystem-playground-master-postsubmits.yaml
ls: ci-operator/jobs/redhat-operator-ecosystem/playground/redhat-operator-ecosystem-playground-master-postsubmits.yaml: No such file or directory
```

hongkailiu:
I need to change this too.
https://github.com/redhat-operator-ecosystem/playground

pkg/jobconfig/files.go
```diff
@@ -443,7 +447,9 @@ func mergePostsubmits(old, new *prowconfig.Postsubmit) prowconfig.Postsubmit {
 func mergePeriodics(old, new *prowconfig.Periodic) prowconfig.Periodic {
 	merged := *new
 
-	merged.MaxConcurrency = old.MaxConcurrency
+	if old.MaxConcurrency != 0 {
```
stevekuznetsov, line 450: Technically this means that if someone tries to set their job to have no limit on concurrency (which is what `0` means) we will stop them. I think I'd rather check the labels to see if it's a publish job?

hongkailiu: Done.

pkg/jobconfig/files.go
hongkailiu: Not doing as discussed over slack.

hongkailiu:
@alvaroaleman 

No breaking any more.

PR #715, opened 2020-04-28
Enable golangci-lint in CI

alvaroaleman:
Enables golangci-lint with three linters:
* `unused`: Finds variables that are assigned to and then never read from until they go out of scope or are written to again. This finds correctness issues.
* `deadcode`: Pretty much what it says. Doesn't find correctness issues but IMHO good to have to keep things simple
* `gosimple`: Also pretty much what it says. Doesn't find correctness issues either, but keeps the code simple

### DIFF
--- a/hack/lint.sh
+++ b/hack/lint.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+echo "Checking gofmt"
+gofmt -s -l $(go list -f '{{ .Dir }}' ./... ) | grep ".*\.go"; if [ "$$?" = "0" ]; then gofmt -s -d $(shell go list -f '{{ .Dir }}' ./... ); exit 1; fi
+
+echo "Running go vet"
+go vet ./...
+
+echo "Running golangci-lint"
+golangci-lint run --disable-all --enable=unused,deadcode,gosimple ./...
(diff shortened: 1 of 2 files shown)


hack/lint.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if ! which golangci-lint &>/dev/null; then
+  echo "Installing golangci-lint"
+  curl --retry 3 -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.25.1
+fi
```
petr-muller, line 8: Oh please no. At least separate the installation from the execution. I definitely do not want to run random scripts from the internet when I run `make lint`. Make the script fail with suggesting how to install the tool maybe?

alvaroaleman: I did it like that mostly to make the whole thing a self-contained change for CI. Do you prefer it to extend the existing lint job or do you prefer to have a new job for this?

alvaroaleman: (This situation reminds me of why I built inrepoconfig...)

petr-muller: I think it would be great to have it in the existing job.

alvaroaleman: I guess I will go the dance of 1. Use the tool if present 2. Update image 3. Always use the tool then /shrug

stevekuznetsov: I mean our traffic in ci-tools is pretty low volume -- feel free to break us for a couple minutes while images build

PR #719, opened 2020-04-28
private-org-sync: reuse config.Options for filtering

stevekuznetsov:
options: allow opting into --confirm as necessary

---

options: expose log-level for testing

---

private-org-sync: reuse config.Options for filtering

This changes the semantics of the flags for filtering but continues to
support the old flags for the time being.

---

/assign @petr-muller 

I understand that this changes the semantics of the flags, but I'd much rather have one consistent way to filter and I'm not sure why this tool was special before. If we want to make the argument that the way that this tool filtered is how we want it to, we can change the common library to do that, too.

Depends on https://github.com/openshift/ci-tools/pull/717


cmd/private-org-sync/main.go
```diff
@@ -38,42 +38,36 @@ type options struct {
...
-		errs = append(errs, fmt.Errorf("--config-dir is required"))
-	}
 
+	// allow users to pass the previous set of flags
+	if o.Org == "" {
+		o.Org = o.org
+	}
+	if o.Repo == "" {
+		o.Repo = o.repo
+	}
```
petr-muller, line 48: The previous option had a different semantics -- it expected a full `org/repo` option, and it wasn't possible to pass both `--only-repo` and `--only-org` options, so if we want to keep them, they need to be kept this way.

(I don't think we necessarily need to keep the old stuff. Our job only uses `--only-org` so we can keep that for the transitional period and then get rid of it)

PR #721, opened 2020-04-28
Label promototion jobs to allow recognizing them

alvaroaleman:

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -4,6 +4,12 @@ import (
 	"fmt"
 )
 
+const (
+	// PromotionJobLabelKey marks promotion jobs as such. Only its presence is
+	// relevant, its value is not.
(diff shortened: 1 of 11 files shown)


pkg/api/types.go
```diff
@@ -4,6 +4,12 @@ import (
 	"fmt"
 )
 
+const (
+	// PromotionJobLabelKey marks promotion jobs as such. Only its presence is
+	// relevant, its value is not.
+	PromotionJobLabelKey = "ci-operator.openshift.io/is-promotion"
```
hongkailiu, line 10: Maybe define constants of labels in one place?

https://github.com/openshift/ci-tools/blob/d60e3f6613f096975bd621ac868782591fcb1088/pkg/jobconfig/files.go#L23-L26

alvaroaleman: One place sounds good, but this is an API, so these things should be in an api package

PR #735, opened 2020-04-29
ci-operator: Only log errors if they haven't been written

smarterclayton:
The error wrapper prevents writing multiple JUnits and was
accidentally included.


smarterclayton:
/assign @stevekuznetsov 

As on slack

PR #736, opened 2020-04-30
Enable the staticcheck linter

alvaroaleman:


alvaroaleman:
Lint failure is a genuine bug and getting fixed in https://github.com/openshift/ci-tools/pull/702

cmd/repo-init/main.go
```diff
@@ -303,6 +303,9 @@ func fetchWithPrompt(msg string) string {
 // responses to the binary in one moment in testing
 var reader = bufio.NewReader(os.Stdin)
 
+// Agree with the linter that this is confusing code, not sure how to improve
```
stevekuznetsov, line 306: awww cmon it's not _that_ bad

stevekuznetsov: What doesn't it like?

pkg/results/report.go
```diff
@@ -83,7 +83,6 @@ func (o *Options) Reporter(spec *api.JobSpec, consoleHost string) (Reporter, err
 		Certificates: []tls.Certificate{cert},
 		RootCAs:      caCertPool,
 	}
-	tlsConfig.BuildNameToCertificate()
```
stevekuznetsov, line 86: Why?

pkg/webreg/webreg.go
```diff
@@ -1938,6 +1938,7 @@ func getAllMultiStageTests(confAgent agents.ConfigAgent, jobAgent *prowConfig.Ag
 				branchVariant := strings.TrimPrefix(name, fmt.Sprint(dashed, "-"))
 				splitBV := strings.Split(branchVariant, "__")
 				branch = splitBV[0]
+				// nolint:staticcheck
```
stevekuznetsov, line 1941: maybe log in the `if` instead

cmd/repo-init/main.go
alvaroaleman:
```
cmd/repo-init/main.go:328:3: SA4004: the surrounding loop is unconditionally terminated (staticcheck)
		return line
		^
cmd/repo-init/main.go:310:14: SA4008: variable in loop condition never changes (staticcheck)
	for i := 0; i < 5; i++ {
	            ^
```

I am actually not sure I understand the second issue

pkg/results/report.go
alvaroaleman: Because its deprecated and shouldn't be used: https://golang.org/pkg/crypto/tls/#Config.BuildNameToCertificate

pkg/webreg/webreg.go
alvaroaleman: good idea, done

cmd/repo-init/main.go
petr-muller: @alvaroaleman The second issue is the consequence of the first. This is not a real loop, the block never runs more than once (there is no path that does not hit a `break` or `return`, so `i` never gets incremented.

alvaroaleman: oh lol, you are right. I will remove the loop then

alvaroaleman: @stevekuznetsov now I don't understand the purpose at all anymore. Was the `break` statement on line 320 supposed to be a `continue`?

Right now, the `break` results in a no-op assigment to `response` (its never read after assigning it) and a return via the `return ""` that is not as dead as the comment declares it.

stevekuznetsov: Yeah it seems like it should be a `continue` ... but it works as written:

```
$ ~/code/openshift/ci-tools/bin/repo-init --release-repo /home/stevekuznetsov/code/openshift/release/src/github.com/openshift/release
Welcome to the repository configuration initializer.
In order to generate a new set of configurations, some information will be necessary.

Let's start with general information about the repository...
Enter the organization for the repository: 
ERROR: a response is required
Please try again.
Enter the organization for the repository: 
ERROR: a response is required
Please try again.
Enter the organization for the repository: 
ERROR: a response is required
Please try again.
Enter the organization for the repository: 
ERROR: a response is required
Please try again.
Enter the organization for the repository: 
ERROR: a response is required
```

alvaroaleman: that is because for the `org` we use `fetchWithPrompt` which does indeed use `continue`. I fixed up `fetchOrDefaultWithPrompt` to also use continue and now the linter is happy because and the code not buggy.

The follow-up issue now is that `fetchOrDefaultWithPrompt` is used by e.G. `fetchWithPrompt` so in those cases we end up with a total of 25 tries instead of five. But IMHO more tries is not a real problem in any way.

PR #737, opened 2020-04-30
Add flag to secret tool for minimizing the logs

hongkailiu:
`--discard-logrus=true` will be used for automation.
I know we have almost nothing for debugging when `--discard-logrus` is enabled.
But this tool is too special and I can live with only local debugging.
WDYT?


hongkailiu:
/close

it will be a private job

PR #738, opened 2020-04-30
Deflake configresolver integration test

alvaroaleman:


alvaroaleman:
> Isn't there an rsync flag to make the action atomic?

rsync just creates a hidden tempfile in the target dir and then moves that at the end. But the tempfile creation would already trigger the inotify. The idea of using `mv` makes sense though, updated the PR to just read into a tempfile, then use an `mv` to move the file in place.

PR #739, opened 2020-04-30
add a list of conditions when operating in ci-operator configs

droslean:
There are many tools that need to run some operations while iterating in the `ci-operator` configuration files only when specific conditions are true.

With this change, we will be able to pass many conditions to be checked before the operation.

Usage example:
```golang
	callback := func(c *api.ReleaseBuildConfiguration, i *config.Info) error {
		// Whatever...
		return nil
	}
	conditions := func(c *api.ReleaseBuildConfiguration, i *config.Info) bool {
		return promotion.BuildsOfficialImages(c) || whitelistConfig.IsWhitelisted(i)
	}

	if err := config.OperateOnCIOperatorConfigDir(filepath.Join(releaseRepoPath, config.CiopConfigInRepoPath), callback, conditions); err != nil {
		return ret, fmt.Errorf("error while operating in ci-operator configuration files: %v", err)
	}
```

/cc @stevekuznetsov @petr-muller @openshift/openshift-team-developer-productivity-test-platform 

### DIFF
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -191,7 +191,7 @@ func isConfigFile(path string, info os.FileInfo) bool {
 
 // OperateOnCIOperatorConfig runs the callback on the parsed data from
 // the CI Operator configuration file provided
-func OperateOnCIOperatorConfig(path string, callback func(*cioperatorapi.ReleaseBuildConfiguration, *Info) error) error {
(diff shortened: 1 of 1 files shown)


droslean:
/hold

Holding to make sure that this is actualy useable.

droslean:
`lint` job is flaking. Opened https://github.com/openshift/release/pull/8695 to fix it.

PR #741, opened 2020-04-30
hack: add a script to grab a kubeconfig from a running cluster

stevekuznetsov:

/assign @wking @dcbw

### DIFF
--- a/hack/extract-kubeconfig.sh
+++ b/hack/extract-kubeconfig.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+namespace="${1:-}"
+test="${2:-}"
+if [[ -z "${namespace}" || -z "${test}" ]]; then
+	echo "USAGE: $0 <namespace> <test>"
+	exit 1
+fi
+
+echo "Scanning ${namespace} for the \$KUBECONFIG for the ${test} test..."
+output="$( mktemp -d /tmp/kubeconfig.XXXXX )"
+cat <<EOF >"${output}/extract.sh"
+#!/bin/bash
+
+if [[ "${1}" != "${test}" ]]; then
+	# we saw a change to an unrelated secret, nothing to do
+	exit 0
+fi
+
+raw="$( oc get secrets "${test}" --namespace "${namespace}" -o jsonpath="{.data.kubeconfig}" )"
+if [[ -n "${raw}" ]]; then
+	echo -n "${kubeconfig}" | base64 --decode > "${output}/kubeconfig.yaml"
+	echo "\$KUBECONFIG saved to ${output}/kubeconfig.yaml"
+	exit 0
+fi
+echo "No \$KUBECONFIG for the $test test has been created yet, waiting..."
+EOF
+chmod +x "${output}/extract.sh"
+
+oc --namespace "${namespace}" observe secrets -- "${output}/extract.sh" &
+
+while true; do
+	if [[ -s "${output}/kubeconfig.yaml" ]]; then
+		exit 0
+	fi
+	sleep 1
+done
+
+for job in $( jobs -p ); do
+	kill -SIGTERM "${job}"
+	wait "${job}"
+done


hack/extract-kubeconfig.sh
```diff
@@ -0,0 +1,54 @@
...
+
+namespace="${1:-}"
+test="${2:-}"
+if [[ -z "${namespace}" || -z "${test}" ]]; then
+	echo "USAGE: $0 <namespace> <test>"
+	exit 1
+fi
+
+function retrieve() {
+	pods="$( oc get pods --namespace ${namespace} --selector "ci.openshift.io/multi-stage-test=${test}" -o json )"
```
wking, line 15: Why pull this out of a pod, instead of grabbing it from the secret behind `$SHARED_DIR`?

wking, line 15: Why pull this out of a pod, instead of grabbing it from the secret behind `$SHARED_DIR`?

wking: For example, even if we always expect a `test` pod to exist in each pipeline, pulling from the secret would allow us to get the kubeconfig during the install step(s) as well, for folks who wanted to monitor or jostle the cluster before the `test` pod exists.

stevekuznetsov: You know more about this than me apparently <3

stevekuznetsov: Well actually this would be the opposite -- grabbing it from the install pod would allow us to get it while installation was running, waiting for it to be in the secret would require install to already have finished.

stevekuznetsov:
@wking added a check at the very top to just grab it from the secret, but kept the Pod stuff since that's the only way to get it while install is running.

stevekuznetsov:
@bbguimaraes would appreciate your review here, too

hack/extract-kubeconfig.sh
```diff
@@ -0,0 +1,63 @@
...
+	echo "USAGE: $0 <namespace> <test>"
+	exit 1
+fi
+
+echo "Scanning ${namespace} for the \$KUBECONFIG for the $test test..."
+raw="$( oc get secrets --namespace "${namespace}" -o jsonpath="{.items[?(@.metadata.name==\"${test}\")].data.kubeconfig}" )"
+if [[ -n "${kubeconfig}" ]]; then
+	output="$( mktemp /tmp/kubeconfig.XXXXX )"
+	echo -n "${kubeconfig}" | base64 --decode > "${output}"
+	echo "\$KUBECONFIG saved to ${output}"
```
wking, line 19:
```sh
echo "export KUBECONFIG='${output}'"
```

so folks can `eval $(hack/extract-kubeconfig.sh)`?  You'd have to write the rest of the script's output to stderr instead (e.g. the earlier usage message), but it would make it more like `ssh-agent` and similar.

hack/extract-kubeconfig.sh
```diff
@@ -0,0 +1,63 @@
...
+	exit 0
+fi
+
+function retrieve() {
+	pods="$( oc get pods --namespace "${namespace}" --selector "ci.openshift.io/multi-stage-test=${test}" -o json )"
+	for pod in $( jq --raw-output '.items[].metadata.name' <<<"${pods}" ); do
+		kubeconfig="$( jq --raw-output ".items[] | select(.metadata.name==\"${pod}\") | .spec.containers[] | select(.name==\"test\") | .env[] | select(.name==\"KUBECONFIG\") | .value" <<<"${pods}" )"
```
wking, line 26: If this pod fallback is aimed at pulling the content out of a running install pod (with the earlier `get secrets` covering post-install, should we be using `install` or some such for the pod name?  How consistent are pod names?  Do we know that it's always a particular pod name providing the kubeconfig?

wking: oh, wait.  By the time we have `KUBECONFIG` in a pod's environment spec, we have the secret.  So is this just trying to close a race where the secret goes up right after we look for it?

stevekuznetsov: No it looks like the KUBECONFIG is in everyone's env, all the time. So we're getting it from the installing Pod before the upload to the secret.

>should we be using install or some such for the pod name?

no

> How consistent are pod names?

They aren't

> Do we know that it's always a particular pod name providing the kubeconfig?

no

hack/extract-kubeconfig.sh
stevekuznetsov: My `$KUBECONFIG` is a special thing I'm pretty careful with, so I don't know that most people want that? Maybe that's just because I'm logged into N prod clusters usually

stevekuznetsov:
Giving this another day for @bbguimaraes to take a look, then merging with lazy consensus.

bbguimaraes:
Sorry, I missed the notification. A few possible optimizations:

- `KUBECONFIG` will always be set for jobs with a cluster profile, it means "this test will eventually produce a kubeconfig"
  - a non-empty `$KUBECONFIG` will not tell if that particular pod will produce it
  - we can fail as soon as a pod without the environment variable is encountered
- [the file is only moved to $KUBECONFIG at the end of the installation](https://github.com/openshift/release/blob/5abd9a9acd589bdda45981911efda1ec541c8634/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L58-L61)
  - can we look at the installer directory instead (if it always has the same path)?
  - if not, the window between the installation and the pod finishing is probably too small for `oc exec` to work, we should check the secret at the top of every iteration

stevekuznetsov:
ACK, so no utility to do anything other than waiting on the secret

stevekuznetsov:
@wking @bbguimaraes updated, PTAL

alvaroaleman:
/uncc
seems there are sufficient SMEs here

hack/extract-kubeconfig.sh
```diff
@@ -0,0 +1,42 @@
...
+set -o pipefail
+
+namespace="${1:-}"
+test="${2:-}"
+if [[ -z "${namespace}" || -z "${test}" ]]; then
+	echo "USAGE: $0 <namespace> <test>"
+	exit 1
+fi
+
+echo "Scanning ${namespace} for the \$KUBECONFIG for the $test test..."
```
petr-muller, line 14:
```suggestion
echo "Scanning ${namespace} for the \$KUBECONFIG for the ${test} test..."
```

hack/extract-kubeconfig.sh
```diff
@@ -0,0 +1,42 @@
...
+output="$( mktemp -d /tmp/kubeconfig.XXXXX )"
+mkdir -p "${output}"
```
petr-muller, line 16: This looks like a nop, `mktemp` creates the thing by default.

hack/extract-kubeconfig.sh
```diff
@@ -0,0 +1,42 @@
...
+if [[ -n "${raw}" ]]; then
+	echo -n "${kubeconfig}" | base64 --decode > "${output}/kubeconfig.yaml"
+	echo "\$KUBECONFIG saved to ${output}/kubeconfig.yaml"
+	exit 0
+fi
+echo "No \$KUBECONFIG for the $test test has been created yet, waiting..."
+EOF
+chmod +x "${output}/extract.sh"
+
+oc --namespace "${namespace}" observe secrets -- "${output}/extract.sh"
```
petr-muller, line 35: Huh should this run in the background so that the following `while `true` loop has something to wait on? Or am I missing something?

stevekuznetsov: No, I'm dumb

stevekuznetsov:
@petr-muller updated

stevekuznetsov:
```
 error: build error: error creating build container: Error initializing source docker://image-registry.openshift-image-registry.svc:5000/ci-op-txyryh74/pipeline@sha256:9e0c275e0bcb495773b10a18e499985d782810e47b4fce076422acb4bc3da3dd: error pinging docker registry image-registry.openshift-image-registry.svc:5000: Get https://image-registry.openshift-image-registry.svc:5000/v2/: dial tcp 172.30.228.83:5000: i/o timeout 
```

PR #743, opened 2020-05-01
ci-operator: add finer-grained error identifiers

stevekuznetsov:
ci-operator: bubble up multiple errors to the top

We currently flatten all errors from a potentially parallel execution of
steps in the graph, which makes it impossible to report the results of
any individual failure easily. Instead, bubble up all the errors
individually and format them in a flat way just for presentation.

---

results: remove constants for reasons

By definition, each of these reasons is only ever going to occur once in
our code, so there is no re-use happening due to the constants. Adding
them is just grunt work and serves no other purpose.

---

ci-operator: add top-level error identifiers for steps

---

### DIFF
--- a/pkg/steps/write_params.go
+++ b/pkg/steps/write_params.go
@@ -3,6 +3,7 @@ package steps
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/ci-tools/pkg/results"
 	"io/ioutil"
 	"log"
 	"regexp"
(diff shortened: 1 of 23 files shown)


pkg/steps/write_params.go
```diff
@@ -3,6 +3,7 @@ package steps
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/ci-tools/pkg/results"
```
alvaroaleman, line 6: nit: import ordering here and in the file above (I might add a linter for that :P)

stevekuznetsov: Please do, my editor does this for me and I never look

PR #748, opened 2020-05-04
ci-operator: fix an off-by-one bug in error handling

petr-muller:
I think this should fix the false negatives where ci-operator-running jobs end with PASS even when some errors happened during the execution.


droslean:
/hold
Holding for manually testing

PR #753, opened 2020-05-04
[private-org-sync] Include whitelisted org/repo/branches in synching

droslean:
This refactoring stands out the synching logic that was happening inside the callback while iterating in the `ci-operator` configs and includes also the whitelisted org/repo/branches.

/cc @petr-muller @openshift/openshift-team-developer-productivity-test-platform

### DIFF
--- a/cmd/private-org-sync/main.go
+++ b/cmd/private-org-sync/main.go
@@ -461,34 +461,71 @@ func main() {
 		failOnNonexistentDst: o.failOnNonexistentDst,
 	}
 
-	var syncErrors []error
+	var errs []error
+
+	locations, whitelistErrors := getWhitelistedLocations(o.WhitelistOptions.WhitelistConfig.Whitelist, syncer.git)
+	errs = append(errs, whitelistErrors...)
+
 	callback := func(_ *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
-		syncer.logger = config.LoggerForInfo(*repoInfo)
-		source := location{
+		l := location{
 			org:    repoInfo.Org,
 			repo:   repoInfo.Repo,
 			branch: repoInfo.Branch,
 		}
+		locations[l] = struct{}{}
+		return nil
+	}
+
+	if err := config.OperateOnCIOperatorConfigDir(o.configDir, o.makeFilter(callback)); err != nil {
+		errs = append(errs, err)
+	}
+
+	for source := range locations {
+		syncer.logger = config.LoggerForInfo(config.Info{
+			Metadata: api.Metadata{
+				Org:    source.org,
+				Repo:   source.repo,
+				Branch: source.branch,
+			},
+		})
+
 		destination := source
 		destination.org = o.targetOrg
 		gitDir, err := syncer.makeGitDir(source.org, source.repo)
 		if err != nil {
-			syncErrors = append(syncErrors, fmt.Errorf("%s->%s: %v", source.String(), destination.String(), err))
-			return nil
+			errs = append(errs, fmt.Errorf("%s->%s: %v", source.String(), destination.String(), err))
+			continue
 		}
 
 		if err := syncer.mirror(gitDir, source, destination); err != nil {
-			syncErrors = append(syncErrors, fmt.Errorf("%s->%s: %v", source.String(), destination.String(), err))
+			errs = append(errs, fmt.Errorf("%s->%s: %v", source.String(), destination.String(), err))
 		}
-		return nil
 	}
 
-	callback = o.makeFilter(callback)
-	if err := config.OperateOnCIOperatorConfigDir(o.configDir, callback); err != nil || len(syncErrors) > 0 {
-		if err != nil {
-			syncErrors = append(syncErrors, err)
-		}
-		e := utilerrors.NewAggregate(syncErrors)
-		logrus.WithError(e).Fatal("There were failures during git content synchronization")
+	if len(errs) > 0 {
+		logrus.WithError(utilerrors.NewAggregate(errs)).Fatal("There were failures")
 	}
+}
+
+func getWhitelistedLocations(whitelist map[string][]string, git gitFunc) (map[location]struct{}, []error) {
+	var errs []error
+	locations := make(map[location]struct{})
+
+	for org, repos := range whitelist {
+		for _, repo := range repos {
+			remote := fmt.Sprintf("https://github.com/%s/%s", org, repo)
+			logger := logrus.WithField("remote", remote)
+
+			branches, err := getRemoteBranchHeads(logger, git, "", remote)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+
+			for branch := range branches {
+				locations[location{org: org, repo: repo, branch: branch}] = struct{}{}
+			}
+		}
+	}
+	return locations, errs
 }


cmd/private-org-sync/main.go
```diff
@@ -124,9 +124,9 @@ func gitExec(logger *logrus.Entry, dir string, command ...string) (string, int,
 
 type RemoteBranchHeads map[string]string
 
-func getRemoteBranchHeads(logger *logrus.Entry, git gitFunc, repoDir, remote string) (RemoteBranchHeads, error) {
+func getRemoteBranchHeads(logger *logrus.Entry, git gitFunc, remote string) (RemoteBranchHeads, error) {
 	heads := RemoteBranchHeads{}
-	out, exitCode, err := git(logger, repoDir, "ls-remote", "--heads", remote)
+	out, exitCode, err := git(logger, "", "ls-remote", "--heads", remote)
```
stevekuznetsov, line 129: Instead of passing `""` I'd rather we had a `func git()` analogue that did not need a dir

cmd/private-org-sync/main.go
```diff
@@ -461,34 +461,72 @@ func main() {
...
+			Branch: source.branch,
+		})
+
 		destination := source
 		destination.org = o.targetOrg
 		gitDir, err := syncer.makeGitDir(source.org, source.repo)
 		if err != nil {
-			syncErrors = append(syncErrors, fmt.Errorf("%s->%s: %v", source.String(), destination.String(), err))
-			return nil
+			errors = append(errors, fmt.Errorf("%s->%s: %v", source.String(), destination.String(), err))
```
stevekuznetsov, line 499: Don't call this `errors` and please update to an editor that correctly warns that you are shadowing an imported package name with that variable name.

cmd/private-org-sync/main.go
```diff
@@ -461,34 +461,72 @@ func main() {
...
-	callback = o.makeFilter(callback)
-	if err := config.OperateOnCIOperatorConfigDir(o.configDir, callback); err != nil || len(syncErrors) > 0 {
-		if err != nil {
-			syncErrors = append(syncErrors, err)
+	if len(errors) > 0 {
+		logrus.WithError(utilerrors.NewAggregate(errors)).Fatal("There were failures")
+	}
+}
+
+func getWhitelistedLocations(whitelistConfig config.WhitelistConfig, git gitFunc) ([]location, error) {
```
stevekuznetsov, line 513: If you just need the whitelist itself, can we pass that instead of the larger config object?

droslean:
> Sorry, looking at this again ... it's not obvious to me why we need this change. If we have the credentials available to use, why wouldn't we provide them in both places we're asking for the remotes?

Well, the whole refactoring was done to create a different logic to get the remote branches without any repodir. I would prefer now to just pass an empty string as a repodir rather than doing all of this refactoring.

stevekuznetsov:
> Well, the whole refactoring was done to create a different logic to get the remote branches without any repodir.

Why is this something we need?

droslean:
> Why is this something we need?

I was trying to improve the design to address the review comment. Either way, we should do this in a follow-up, if this is necessary.

cmd/private-org-sync/main.go
```diff
@@ -460,35 +460,73 @@ func main() {
 		git:                  gitExec,
 		failOnNonexistentDst: o.failOnNonexistentDst,
 	}
+	var errs []error
+	var locations []location
+
+	whitelist := o.WhitelistOptions.WhitelistConfig.Whitelist
+	if len(whitelist) > 0 {
```
petr-muller, line 467: Can we make `getWhitelistedLocations` return an empty location list on an empty whitelist, so that the logic here can be simpler (without the `len(whitelist) > 0` condition)?

cmd/private-org-sync/main.go
```diff
@@ -460,35 +460,73 @@ func main() {
...
+		whitelistedLocations, err := getWhitelistedLocations(whitelist, syncer.git)
+		if err != nil {
+			errs = append(errs, err)
```
petr-muller, line 470: Hmm, `err` is an aggregate anyway, so would it make sense to return a `[]error` instead from `getWhitelistedLocations` and simply append here? Would make the final aggregate more flat, which I think is better.

cmd/private-org-sync/main.go
```diff
@@ -460,35 +460,73 @@ func main() {
...
-	callback = o.makeFilter(callback)
-	if err := config.OperateOnCIOperatorConfigDir(o.configDir, callback); err != nil || len(syncErrors) > 0 {
-		if err != nil {
-			syncErrors = append(syncErrors, err)
+	if len(errs) > 0 {
+		logrus.WithError(utilerrors.NewAggregate(errs)).Fatal("There were failures")
+	}
+}
+
+func getWhitelistedLocations(whitelist map[string][]string, git gitFunc) ([]location, error) {
```
petr-muller, line 513: I would like to see this tested. There is a git mock already in `main_test.go` which should make testing this not that hard.

cmd/private-org-sync/main.go
```diff
@@ -460,35 +460,73 @@ func main() {
...
+	}
 
-	var syncErrors []error
 	callback := func(_ *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
-		syncer.logger = config.LoggerForInfo(*repoInfo)
-		source := location{
+		locations = append(locations, location{
 			org:    repoInfo.Org,
 			repo:   repoInfo.Repo,
 			branch: repoInfo.Branch,
```
petr-muller, line 479: Hmmm... I am wondering: if we accidentally happen to have a repo both in whitelist and having promoting config, we will end up having these locations in the worklist twice. Do we actually care? The locations will end up being synced twice, but that should not be much of an issue.

But if we want to avoid that, we could make the worklist a set instead of a list.

petr-muller:
> Why is this something we need?

@stevekuznetsov To me the change makes sense. With whitelist, the items that we need to sync git content for are the following:

1. Items from the whitelist. We do not know which branches are important, so we want to sync all of them. So we first ask the source remote for the list of all branches it has and add them all to the worklist.
2. Items for the org/repo/branches that promote official images. Here we know which branches are important, so we only add these to the worklist.

Then we simply sync all items from the worklist. I think that structure makes sense, WDYT?

petr-muller:
@droslean important question: have you tested the thing? Preferably with both types of inputs at the same time?

cmd/private-org-sync/main.go
droslean: Since there is not a good way to merge maps in golang, since the whole concept is like `c` and the map is already a pointer, I can just make the `getWhitelistedLocations` to fill the same map with here to avoid duplicates.

cmd/private-org-sync/main.go
droslean: Well, I don't think that there is a good reason to test that func since it is only iterating in a map and calls `getRemoteBranchHeads`. But instead, I believe that adding unittests for `getRemoteBranchHeads` should cover the required testing.

droslean:
> important question: have you tested the thing? Preferably with both types of inputs at the same time?

Yes I tested it locally without `--confirm`

cmd/private-org-sync/main.go
droslean: Ah this is already being tested in `TestMirror`.

cmd/private-org-sync/main.go
```diff
@@ -461,34 +461,70 @@ func main() {
 		failOnNonexistentDst: o.failOnNonexistentDst,
 	}
 
-	var syncErrors []error
+	var errs []error
+	locations := make(map[location]struct{})
+
+	if whitelistErrors := getWhitelistedLocations(o.WhitelistOptions.WhitelistConfig.Whitelist, syncer.git, locations); len(whitelistErrors) > 0 {
```
petr-muller, line 467: Please no side effects. `locations` should be returned as an output, not received as an input and modified.

cmd/private-org-sync/main.go
```diff
@@ -461,34 +461,70 @@ func main() {
...
+		errs = append(errs, whitelistErrors...)
```
petr-muller, line 468: This doesn't need to be in an if; both an empty slice and a nil slice can be appended like this.

cmd/private-org-sync/main.go
```diff
@@ -461,34 +461,70 @@ func main() {
...
+func getWhitelistedLocations(whitelist map[string][]string, git gitFunc, locations map[location]struct{}) []error {
+	var errs []error
+	for org, repos := range whitelist {
+		for _, repo := range repos {
+			remote := fmt.Sprintf("https://github.com/%s/%s", org, repo)
+			logger := logrus.WithField("remote", remote)
+
+			branches, err := getRemoteBranchHeads(logger, git, "", remote)
+			if err != nil {
+				errs = append(errs, err)
```
petr-muller, line 521: If we have an error here we should not continue and iterate over `branches`

cmd/private-org-sync/main.go
droslean: I would prefer to fill the map directly than merging two different maps, which in golang this is really ugly :/

petr-muller: I don't understand. If you do

`locations, whitelistErrors := getWhitelistedLocations(whitelist, git)` 

then the rest of this method does need to change at all, and we will not have any confusing side effects. But also, why is merging two maps ugly? It's just an easy iteration.

droslean: Yea since the whole thing is not complicated we can use the same map later in the code as well. Addressed

PR #759, opened 2020-05-05
prowgen: resolve configuration

bbguimaraes:
/hold

Requires: https://github.com/openshift/release/pull/8846.


bbguimaraes:
Double hold, `hack/validate-generated-config.sh` will need to be updated and can't use `--from-release-repo`.

petr-muller:
Curious, we does prowgen need to resolve the config?

petr-muller:
/test all

(I need to test something)

stevekuznetsov:
I don't think we will need this

/close

PR #760, opened 2020-05-05
ci-operator: return better errors on lease failures

stevekuznetsov:

/assign @bbguimaraes

### DIFF
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -2,7 +2,7 @@ package steps
 
 import (
 	"context"
-	"fmt"
+	"errors"
 	"log"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
(we need this since the aggregate type does not implement the wrapping functions we use to detect child errors)

pkg/steps/lease.go
```diff
@@ -58,25 +58,33 @@ func (s *leaseStep) SubTests() []*junit.TestCase {
...
+	releaseErr := results.ForReason("releasing_lease").ForError(client.Release(lease))
+
+	// we want a sensible output error for reporting, so we bubble up these individually
+	//if we can, as this is the only step that can have multiple errors
+	if wrappedErr != nil && releaseErr == nil {
+		return wrappedErr
+	} else if wrappedErr == nil && releaseErr != nil {
+		return releaseErr
+	} else {
+		return utilerrors.NewAggregate([]error{wrappedErr, releaseErr})
```
alvaroaleman, line 88: you don't need any of this, utilerrors.NewAggregate filters out nil errors and returns nil if all errors are nil

stevekuznetsov: No, we do ... `NewAggregate` on a list with only one non-`nil` error does not return that error, it returns an aggregate, which does not wrap the error inside of it, and therefore does not expose it in the way we want it to be for the results library.

alvaroaleman: You can handle that at the callsite by asserting on aggregate and then using its `Errors()`

stevekuznetsov: That's going to be super messy and annoying -- we'd need to basically work around the `errors.As` you had me write. If we had a result-annotated error wrapping a aggregate, you end up half of the time doing `errors.As` and half of the time looking at child nodes explicitly and doing typecasting. Ew. This is a smaller wart.

pkg/steps/lease.go
```diff
@@ -58,25 +58,33 @@ func (s *leaseStep) SubTests() []*junit.TestCase {
...
 	client := *s.client
 	if client == nil {
-		return fmt.Errorf("step needs a lease but no lease client provided")
+		return results.ForReason("initializing_client").ForError(errors.New("step needs a lease but no lease client provided"))
 	}
 	ctx, cancel := context.WithCancel(ctx)
 	lease, err := client.Acquire(s.leaseType, ctx, cancel)
 	if err != nil {
-		return fmt.Errorf("failed to acquire lease: %v", err)
+		return results.ForReason("acquiring_Lease").WithError(err).Errorf("failed to acquire lease: %v", err)
```
petr-muller, line 73: https://github.com/openshift/ci-tools/pull/762

PR #762, opened 2020-05-06
lease: fix a typo in a failure reason

petr-muller:


petr-muller:
/test all

Looking at you, GitHub

PR #764, opened 2020-05-06
[private-org-peribolos-sync] include whitelisted repos in peribolos sync

droslean:
/cc @petr-muller @openshift/openshift-team-developer-productivity-test-platform 

### DIFF
--- a/cmd/private-org-peribolos-sync/main.go
+++ b/cmd/private-org-peribolos-sync/main.go
@@ -106,7 +106,7 @@ func main() {
 		logger.WithError(err).Fatal("Error getting GitHub client.")
 	}
 
-	orgRepos, err := getReposForPrivateOrg(o.releaseRepoPath)
+	orgRepos, err := getReposForPrivateOrg(o.releaseRepoPath, o.WhitelistOptions.WhitelistConfig.Whitelist)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/private-org-peribolos-sync/main.go
```diff
@@ -159,8 +159,19 @@ func generateRepositories(gc gitHubClient, orgRepos map[string]sets.String, logg
 
 // getReposForPrivateOrg itterates through the release repository directory and creates a map of
 // repository sets by organization that promote official images.
-func getReposForPrivateOrg(releaseRepoPath string) (map[string]sets.String, error) {
+func getReposForPrivateOrg(releaseRepoPath string, whitelist map[string][]string) (map[string]sets.String, error) {
 	ret := make(map[string]sets.String)
+
+	for org, repos := range whitelist {
```
stevekuznetsov, line 165: You might want to make `map[string]sets.String` into an exported type and expose an `Insert(org, repo string)` function for it to DRY this out, so can share code.

droslean: Every tool is using its maps in a different manner, so there is no reason to expose the logic. Also, we don't need to insert anything in the existing whitelist.

PR #767, opened 2020-05-06
ci-operator-configresolver: serve matching configurations

stevekuznetsov:
At feature-freeze, developers create branches which prefix their working
branch and use Prow's regex matching on branch names to not need to
reconfigure their CI whatsoever. For instance, a `master` branch will be
used to create `master-feature`. We now support the same operation in
the configuration server.

/assign @AlexNPavel

### DIFF
--- a/pkg/load/agents/configAgent.go
+++ b/pkg/load/agents/configAgent.go
@@ -21,6 +21,7 @@ type ConfigAgent interface {
 	// GetMatchingConfig loads a configuration that matches the metadata,
 	// allowing for regex matching on branch names.
 	GetMatchingConfig(metadata api.Metadata) (api.ReleaseBuildConfiguration, error)
+	GetAll() load.ByOrgRepo
 	GetGeneration() int
 	AddIndex(indexName string, indexFunc IndexFn) error
 	GetFromIndex(indexName string, indexKey string) ([]*api.ReleaseBuildConfiguration, error)
@@ -96,7 +97,11 @@ func (a *configAgent) GetMatchingConfig(metadata api.Metadata) (api.ReleaseBuild
 	}
 	var matchingConfigs []api.ReleaseBuildConfiguration
 	for _, config := range repoConfigs {
-		if regexp.MustCompile(config.Metadata.Branch).MatchString(metadata.Branch) && config.Metadata.Variant == metadata.Variant {
+		r, err := regexp.Compile(config.Metadata.Branch)
+		if err != nil {
+			return api.ReleaseBuildConfiguration{}, fmt.Errorf("could not compile regex for %s/%s@%s: %v", metadata.Org, metadata.Repo, config.Metadata.Branch, err)
+		}
+		if r.MatchString(metadata.Branch) && config.Metadata.Variant == metadata.Variant {
 			matchingConfigs = append(matchingConfigs, config)
 		}
 	}
@@ -108,6 +113,12 @@ func (a *configAgent) GetMatchingConfig(metadata api.Metadata) (api.ReleaseBuild
 	default:
 		return api.ReleaseBuildConfiguration{}, fmt.Errorf("found more than one matching config for branch %s on repo %s/%s", metadata.Branch, metadata.Org, metadata.Repo)
 	}
+}
+
+func (a *configAgent) GetAll() load.ByOrgRepo {
+	a.lock.RLock()
+	defer a.lock.RUnlock()
+	return a.configs
 }
 
 func (a *configAgent) GetGeneration() int {
--- a/pkg/load/agents/configAgent.go
+++ b/pkg/load/agents/configAgent.go
@@ -2,6 +2,7 @@ package agents
 
 import (
 	"fmt"
+	"regexp"
 	"sync"
 	"time"
 
@@ -17,8 +18,9 @@ import (
 // ConfigAgent is an interface that can load configs from disk into
 // memory and retrieve them when provided with a config.Info.
 type ConfigAgent interface {
-	GetConfig(metadata api.Metadata) (api.ReleaseBuildConfiguration, error)
-	GetAll() load.FilenameToConfig
+	// GetMatchingConfig loads a configuration that matches the metadata,
+	// allowing for regex matching on branch names.
+	GetMatchingConfig(metadata api.Metadata) (api.ReleaseBuildConfiguration, error)
 	GetGeneration() int
 	AddIndex(indexName string, indexFunc IndexFn) error
 	GetFromIndex(indexName string, indexKey string) ([]*api.ReleaseBuildConfiguration, error)
@@ -29,7 +31,7 @@ type IndexFn func(api.ReleaseBuildConfiguration) []string
 
 type configAgent struct {
 	lock         *sync.RWMutex
-	configs      load.FilenameToConfig
+	configs      load.ByOrgRepo
 	configPath   string
 	cycle        time.Duration
 	generation   int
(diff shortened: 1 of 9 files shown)


pkg/load/load.go
```diff
@@ -36,7 +36,33 @@ const (
...
+	byFilename, err := FromPath(path)
+	if err != nil {
+		return nil, err
+	}
+
+	return PartitionByOrgRepo(byFilename), nil
+}
+
+func PartitionByOrgRepo(byFilename FilenameToConfig) ByOrgRepo {
+	byOrgRepo := map[string]map[string][]api.ReleaseBuildConfiguration{}
```
petr-muller, line 52: Use the `ByOrgRepo` type?

pkg/load/load.go
```diff
@@ -36,7 +36,33 @@ const (
...
+	for _, configuration := range byFilename {
+		if _, exists := byOrgRepo[configuration.Metadata.Org]; !exists {
+			byOrgRepo[configuration.Metadata.Org] = map[string][]api.ReleaseBuildConfiguration{}
+		}
+		if _, exists := byOrgRepo[configuration.Metadata.Org][configuration.Metadata.Repo]; !exists {
+			byOrgRepo[configuration.Metadata.Org][configuration.Metadata.Repo] = []api.ReleaseBuildConfiguration{}
+		}
+		byOrgRepo[configuration.Metadata.Org][configuration.Metadata.Repo] = append(byOrgRepo[configuration.Metadata.Org][configuration.Metadata.Repo], configuration)
```
petr-muller, line 60: semi-nit: I think this method would be more readable if you extracted the org and repo strings to `org` and `repo` variables and used that in the logic. I mean, it's no rocket science, but it *looks* like rocket science :)

pkg/load/load_test.go
```diff
@@ -2,6 +2,7 @@ package load
 
 import (
 	"encoding/json"
+	"github.com/google/go-cmp/cmp"
```
petr-muller, line 5: :trollface:

pkg/load/agents/configAgent_test.go
```diff
@@ -175,3 +175,183 @@ func TestBuildIndexes(t *testing.T) {
...
+			expectedErr: false,
+		},
+		{
+			name: "regex match on branch returns it",
+			input: load.ByOrgRepo{
+				"org": map[string][]api.ReleaseBuildConfiguration{
+					"repo": {{Metadata: api.Metadata{
+						Org:    "org",
+						Repo:   "repo",
+						Branch: "branch",
```
alvaroaleman, line 258: This testcase confuses me, why does it pass, shouldn't the `Branch` be a `branch.*` to have it pass?

pkg/load/load_test.go
alvaroaleman, line 5: import order

pkg/load/agents/configAgent.go
```diff
@@ -82,15 +85,44 @@ func (a *configAgent) recordError(label string) {
...
+	if !exist {
+		return api.ReleaseBuildConfiguration{}, fmt.Errorf("could not find any config for org %s", metadata.Org)
+	}
+	repoConfigs, exist := orgConfigs[metadata.Repo]
+	if !exist {
+		return api.ReleaseBuildConfiguration{}, fmt.Errorf("could not find any config for repo %s/%s", metadata.Org, metadata.Repo)
+	}
+	var matchingConfigs []api.ReleaseBuildConfiguration
+	for _, config := range repoConfigs {
+		if regexp.MustCompile(config.Metadata.Branch).MatchString(metadata.Branch) && config.Metadata.Variant == metadata.Variant {
```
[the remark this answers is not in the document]
stevekuznetsov: Do you want it at load-time? Right now it's at the largest scope that the branch exists

pkg/load/agents/configAgent.go
```diff
@@ -82,15 +85,44 @@ func (a *configAgent) recordError(label string) {
...
+			matchingConfigs = append(matchingConfigs, config)
+		}
+	}
+	switch len(matchingConfigs) {
+	case 0:
+		return api.ReleaseBuildConfiguration{}, fmt.Errorf("could not find any config for branch %s on repo %s/%s", metadata.Branch, metadata.Org, metadata.Repo)
+	case 1:
+		return matchingConfigs[0], nil
+	default:
+		return api.ReleaseBuildConfiguration{}, fmt.Errorf("found more than one matching config for branch %s on repo %s/%s", metadata.Branch, metadata.Org, metadata.Repo)
```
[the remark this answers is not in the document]
stevekuznetsov: No, it's never valid for this to occur (and we validate openshift/release to that extent)

pkg/load/agents/configAgent_test.go
stevekuznetsov: Regex doesn't need to match the full input!

pkg/load/agents/configAgent_test.go
```diff
@@ -175,3 +175,183 @@ func TestBuildIndexes(t *testing.T) {
...
 	}
 }
+
+func TestConfigAgent_GetMatchingConfig(t *testing.T) {
+	var testCases = []struct {
+		name        string
+		input       load.ByOrgRepo
+		meta        api.Metadata
+		expected    api.ReleaseBuildConfiguration
+		expectedErr bool
```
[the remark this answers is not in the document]
stevekuznetsov: I haven't found that to be the case /shrug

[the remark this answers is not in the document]
stevekuznetsov: _which_ error comes out is not part of the expected behavior for this method

pkg/load/load_test.go
stevekuznetsov: the imports were not even right in the first place -.-

pkg/load/agents/configAgent.go
```diff
@@ -17,8 +18,9 @@ import (
 // ConfigAgent is an interface that can load configs from disk into
 // memory and retrieve them when provided with a config.Info.
 type ConfigAgent interface {
-	GetConfig(metadata api.Metadata) (api.ReleaseBuildConfiguration, error)
-	GetAll() load.FilenameToConfig
```
stevekuznetsov, line 21: We had no users

pkg/load/agents/configAgent.go
```diff
@@ -17,8 +18,9 @@ import (
 // ConfigAgent is an interface that can load configs from disk into
 // memory and retrieve them when provided with a config.Info.
 type ConfigAgent interface {
-	GetConfig(metadata api.Metadata) (api.ReleaseBuildConfiguration, error)
```
stevekuznetsov, line 20: There's no utility to this

pkg/load/agents/configAgent.go
AlexNPavel: https://github.com/openshift/ci-tools/blob/master/pkg/webreg/webreg.go#L1851

alvaroaleman: How does this compile then?

pkg/load/agents/configAgent.go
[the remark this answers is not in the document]
alvaroaleman: Just in the very beginning of this func
And also, do not use `MustCompile`, it might panic and that means that a single defunct regex here can kill everything that uses the `configAgent`

pkg/load/agents/configAgent.go
```diff
@@ -29,7 +31,7 @@ type IndexFn func(api.ReleaseBuildConfiguration) []string
 
 type configAgent struct {
 	lock         *sync.RWMutex
-	configs      load.FilenameToConfig
+	configs      load.ByOrgRepo
```
AlexNPavel, line 34: The `findConfigForJob` function in `webreg` will need to be updated (probably need unit tests for some of the helper functions in that pkg): https://github.com/openshift/ci-tools/blob/master/pkg/webreg/webreg.go#L1820

pkg/load/agents/configAgent.go
alvaroaleman: my controller also uses this. Why does this PR compile?

stevekuznetsov: Weird, GoLand did not tell me there were any callers ....

pkg/load/agents/configAgent.go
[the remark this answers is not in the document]
stevekuznetsov: Err, how do you mean at the "top of this func" ... we are compiling the branches on the config, not the branch that gets passed in. This is the largest scope in which any given branch exists.

pkg/load/agents/configAgent.go
stevekuznetsov: No tests ;)

pkg/load/agents/configAgent.go
```diff
@@ -144,7 +172,7 @@ func (a *configAgent) loadFilenameToConfig() error {
 	indexes := a.buildIndexes(configs)
 
 	a.lock.Lock()
-	a.configs = configs
+	a.configs = load.PartitionByOrgRepo(configs)
```
alvaroaleman, line 175: Lets do the `PartitionByOrgRepo` before we acquire the lock

PR #771, opened 2020-05-07
ci-operator-configresolver: serve matching configurations

alvaroaleman:
(Creating a new PR because I can not push to steves branch)
At feature-freeze, developers create branches which prefix their working
branch and use Prow's regex matching on branch names to not need to
reconfigure their CI whatsoever. For instance, a master branch will be
used to create master-feature. We now support the same operation in
the configuration server.

Signed-off-by: Steve Kuznetsov skuznets@redhat.com

/assign @AlexNPavel

This pr has one additional commit that minimizes the exported surface of pkg/load. I didn't get around to make the job retrieval an index, because it requires a bit more refactoring wrt to error handling and I don't have time for that now.
closes #767

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -22,13 +22,6 @@ import (
 	"github.com/openshift/ci-tools/pkg/webreg"
 )
 
-const (
-	orgQuery     = "org"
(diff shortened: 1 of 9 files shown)


stevekuznetsov:
Thanks, Alvaro

PR #775, opened 2020-05-08
Add helper to extract all ImageStreamTags for a test config

alvaroaleman:
/assign @bbguimaraes

### DIFF
--- a/pkg/api/helper/imageextraction.go
+++ b/pkg/api/helper/imageextraction.go
@@ -1,0 +1,59 @@
+package helper
+
+import (
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/types"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/load"
+)
+
+// GetAllTestInputImageStreamTags returns a deduplicated and unsorted list of all ImageStreamTags
+// referenced anywhere as input by a test in the config. It only returns their namespace and name and drops the
+// cluster field, as we plan to remove that.
+func GetAllTestInputImageStreamTags(config load.ByOrgRepo) map[string]types.NamespacedName {
+	result := map[string]types.NamespacedName{}
+	for _, org := range config {
+		for _, repo := range org {
+			for _, cfg := range repo {
+				imageStreamTagReferenceMapIntoMap(cfg.BaseImages, result)
+				imageStreamTagReferenceMapIntoMap(cfg.BaseRPMImages, result)
+				if cfg.BuildRootImage != nil && cfg.BuildRootImage.ImageStreamTagReference != nil {
+					insert(*cfg.BuildRootImage.ImageStreamTagReference, result)
+				}
+
+				for _, rawStep := range cfg.RawSteps {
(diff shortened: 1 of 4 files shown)


pkg/api/helper/imageextraction.go
```diff
@@ -0,0 +1,75 @@
...
+
+				for _, rawStep := range cfg.RawSteps {
+					if rawStep.InputImageTagStepConfiguration != nil {
+						result.Insert(imageStreamTagReferenceToString(rawStep.InputImageTagStepConfiguration.BaseImage))
+					}
+					if rawStep.SourceStepConfiguration != nil {
+						result.Insert(imageStreamTagReferenceToString(rawStep.SourceStepConfiguration.ClonerefsImage))
+					}
+					// TODO: Is this correct?
+					if rawStep.OutputImageTagStepConfiguration != nil {
```
alvaroaleman, line 37: @bbguimaraes do you know?

alvaroaleman:
test failure is a flake, ref: https://github.com/openshift/ci-tools/pull/776
/retest

pkg/api/helper/imageextraction.go
bbguimaraes: These are images that will be built from the configuration, so they should not be included if this PR is part of https://issues.redhat.com/browse/DPTP-1062.

We should rename the method to make it clear that only input ISTs will be included.

pkg/api/helper/imageextraction.go
```diff
@@ -0,0 +1,75 @@
...
+		if split[0] == "" || split[1] == "" {
+			errs = append(errs, fmt.Errorf("namespace or name of item %q is empty", item))
+			continue
+		}
+		if res := strings.Split(split[1], ":"); len(res) != 2 || res[0] == "" || res[1] == "" {
+			errs = append(errs, fmt.Errorf("imagestreamtag name %q is not valid (must be <<name>>:<<tag>>)", split[1]))
+			continue
+		}
+		ret = append(ret, types.NamespacedName{Namespace: split[0], Name: split[1]})
+	}
```
bbguimaraes, line 62: These values are created by functions in this module (unless they were originally empty in the configuration, but that should be validated outside this method), so panicking seems appropriate instead of propagating an error.

Moreover, we could map `imageStreamTagReferenceToString` values directly to `types.NamespacedName` (instead of `struct{}`), then simply collect the values here.

alvaroaleman: > These values are created by functions in this module (unless they were originally empty in the configuration, but that should be validated outside this method), so panicking seems appropriate instead of propagating an error.

The case the error handling is supposed to handle is that we somehow get an invalid config, for example because our validation is buggy. While this should not happen in the first place I would prefer not to blow the whole controller up if it does.

The reason I didn't directly map them to `NamespacedName` is that I need them deduplicated

bbguimaraes: Right, but I'm suggesting using a `map[string]types.NamespacedName` instead of the `map[string]struct{}` (i.e. `sets.String`) we are using. We'd still get unique values.

alvaroaleman:
@bbguimaraes updated, PTAL

PR #778, opened 2020-05-11
prowgen: separate multi-stage generation

bbguimaraes:


pkg/prowgen/prowgen.go
```diff
@@ -454,21 +427,40 @@ func generatePeriodicForTest(name string, info *ProwgenInfo, label jc.ProwgenLab
...
+	case cioperatorapi.ClusterProfileAWS,
+		cioperatorapi.ClusterProfileAzure4,
+		cioperatorapi.ClusterProfileLibvirtS390x,
+		cioperatorapi.ClusterProfileOpenStack,
+		cioperatorapi.ClusterProfileOpenStackVexxhost,
+		cioperatorapi.ClusterProfileOpenStackPpc64le,
+		cioperatorapi.ClusterProfileVSphere:
+	default:
+		s := &ret.VolumeSource.Projected.Sources
+		*s = append(*s, corev1.VolumeProjection{
```
alvaroaleman, line 455: just `append(ret.VolumeSource.Projected.Sources, corev1.VolumeProjection{` is the same as getting a pointer, then instantly dereferencing it or not?

stevekuznetsov: Yeah, `ret.VolumeSource.Projected.Sources = append(ret.VolumeSource.Projected.Sources, ...)` would be the same

bbguimaraes:
Rebased, updated.

PR #788, opened 2020-05-12
PJ rehearse: Allow using dedicated kubeconfig for prowjobs

alvaroaleman:

This PR allows to explicitly set a kubeconfig for prowjobs via a new flag on pj-rehearse.

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
 	prowgithub "k8s.io/test-infra/prow/github"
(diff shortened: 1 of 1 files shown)


alvaroaleman:
/hold
This isn't sufficient

alvaroaleman:
/hold cancel
Should be sufficient now

cmd/pj-rehearse/main.go
```diff
@@ -239,7 +248,7 @@ func rehearseMain() error {
 		prConfig.Prow.ProwJobNamespace = config.StagingNamespace
 	}
 
-	cmClient, err := rehearse.NewCMClient(clusterConfig, prConfig.Prow.ProwJobNamespace, o.dryRun)
+	cmClient, err := rehearse.NewCMClient(clusterConfig, prConfig.Prow.PodNamespace, o.dryRun)
```
petr-muller, line 251: Is this really that simple? If our pods run at one cluster and prowjobs are in another then pj-rehearse need to work with both, no?

How did this even work with jobs running on build01 before? :/

alvaroaleman: Yeah thats what it does, you can now specify a dedicated kubeconfig for prowjobs.

Are you sure it ever worked with build01?

petr-muller: Yeah, sorry, I misread. We always use the cluster config here, and the optionally passed kubeconfig for the pj client. 

From what I'm reading, I would expect this to fail when pj-rehearse runs on a different cluster than the rehearsal (because pj-rehearse always creates this temproary CM on the cluster where it runs itself), but maybe I'm missing something.

But for now this is correct, thanks!

PR #790, opened 2020-05-12
Use app.ci configresolver

alvaroaleman:


alvaroaleman:
Its interesting how this doesn't can be just changed without adjusting any test.

bbguimaraes:
> Its interesting how this doesn't can be just changed without adjusting any test.

We rely on the default value everywhere, and don't usually test argument treatment directly.

/retest

alvaroaleman:
> We rely on the default value everywhere, and don't usually test argument treatment directly.

I guess it mostly confuses me because there is a set of fixtures that has the url. But maybe they are just stale?
/hold cancel

PR #791, opened 2020-05-13
prowgen: add custom secret to custom test image

danielBelenky:
In some cases, we need to use secrets during e2e tests. For example,
when we pull pre-released images from a private registry. To support
this use case, the custom test image template now accepts a
`custom_secret_name` parameter that allows users to specify a name of a
secret to mount to the custom test pod.

For users who don't specify it, a default empty secret will be used and
mounted to their test pod. This will have no effects on the tests pod.

Signed-off-by: Daniel Belenky <dbelenky@redhat.com>

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -773,6 +773,7 @@ type OpenshiftInstallerCustomTestImageClusterTestConfiguration struct {
 	From             string `json:"from"`
 	EnableNestedVirt bool   `json:"enable_nested_virt,omitempty"`
 	NestedVirtImage  string `json:"nested_virt_image,omitempty"`
+	CustomSecretName string `json:"custom_secret_name,omitempty"`
 }
 
 // OpenshiftInstallerGCPNestedVirtCustomTestImageClusterTestConfiguration describes a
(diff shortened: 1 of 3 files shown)


danielBelenky:
/hold

Adding a hole for now, because we need to first create the empty-secret.
Reviews are welcome.

droslean:
These changes won't affect any e2e tests. Prowgen will only add the secret to the job YAML that runs a ci-operator instance, which means that that secret will never make it to the e2e test. Also, in this way we will expose the secret everywhere since the job YAML lives in the public release repo. 

However, there is already a technique to use your own custom secrets.

danielBelenky:
@droslean thanks for your response.

> These changes won't affect any e2e tests. Prowgen will only add the secret to the job YAML that runs a ci-operator instance, which means that that secret will never make it to the e2e test. Also, in this way we will expose the secret everywhere since the job YAML lives in the public release repo.

As I understand it, the env from the container that runs ci-operator is being used to render the template. So this value should end up being used when the operator is running `oc process` on that template. This is at least how I understood it (and I might be wrong).

> However, there is already a technique to use your own custom secrets.

How can I consume my own secret using the custom image template?

droslean:
I am kindly closing this.

As discussed in [slack](https://coreos.slack.com/archives/CBN38N3MW/p1589356370424400) the implementation of the requested feature is in progress. In addition, this PR does not approach any solution or implementation.

Also, the current changes have a big risk in exposing private secrets to the internet.

/close

PR #794, opened 2020-05-13
[ci-operator-config-mirror] allow the tool to delete the existing but deprecated configs 

droslean:
Introduces a map by repo with their data. Two methods will be responsible to clean up and generate the data.
Adds a `--clean` flag, and default it to true.

/cc @petr-muller @stevekuznetsov @openshift/openshift-team-developer-productivity-test-platform 

### DIFF
--- a/cmd/ci-operator-config-mirror/main.go
+++ b/cmd/ci-operator-config-mirror/main.go
@@ -4,7 +4,9 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"path/filepath"
 	"strings"
 
 	"github.com/sirupsen/logrus"
@@ -26,6 +28,8 @@ type options struct {
 
 	configDir string
 	toOrg     string
+
+	clean bool
 }
 
 func gatherOptions() options {
@@ -34,6 +38,8 @@ func gatherOptions() options {
 
 	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
 	fs.StringVar(&o.toOrg, "to-org", "", "Name of the organization which the ci-operator configuration files will be mirrored")
+
+	fs.BoolVar(&o.clean, "clean", true, "If the `to-org` folder already exists, then delete all subdirectories")
 
 	o.WhitelistOptions.Bind(fs)
 	fs.Parse(os.Args[1:])
@@ -51,6 +57,38 @@ func (o *options) validate() error {
 	return o.WhitelistOptions.Validate()
 }
 
+type configsByRepo map[string][]config.DataWithInfo
+
+func (d configsByRepo) cleanDestinationSubdirs(destinationOrgPath string) error {
+	contents, err := ioutil.ReadDir(destinationOrgPath)
+	if err != nil {
+		return fmt.Errorf("failed to read directory %s: %v", destinationOrgPath, err)
+	}
+
+	for _, info := range contents {
+		if !info.IsDir() {
+			continue
+		}
+
+		if err := os.RemoveAll(filepath.Join(destinationOrgPath, info.Name())); err != nil {
+			return fmt.Errorf("couldn't delete dir: %v", err)
+		}
+	}
+
+	return nil
+}
+
+func (d configsByRepo) generateConfigs(configDir string) error {
+	for _, dataWithInfos := range d {
+		for _, dataWithInfo := range dataWithInfos {
+			if err := dataWithInfo.CommitTo(configDir); err != nil {
+				return fmt.Errorf("couldn't create the configuration file: %v", err)
+			}
+		}
+	}
+	return nil
+}
+
 func main() {
 	o := gatherOptions()
 	if err := o.validate(); err != nil {
@@ -61,6 +99,8 @@ func main() {
 		interrupts.WaitForGracefulShutdown()
 		os.Exit(1)
 	}()
+
+	configsByRepo := make(configsByRepo)
 
 	callback := func(rbc *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
 		logger := logrus.WithFields(logrus.Fields{"org": repoInfo.Org, "repo": repoInfo.Repo, "branch": repoInfo.Branch})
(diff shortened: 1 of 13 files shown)


cmd/ci-operator-config-mirror/main.go
```diff
@@ -35,6 +38,8 @@ func gatherOptions() options {
 	fs.StringVar(&o.configDir, "config-path", "", "Path to directory containing ci-operator configurations")
 	fs.StringVar(&o.toOrg, "to-org", "", "Name of the organization which the ci-operator configuration files will be mirrored")
 
+	fs.BoolVar(&o.deprecate, "deprecate-configs", true, "Deletes the existing but deprecated configuration")
```
petr-muller, line 41: I'd like to see an explanation about what a "deprecated" configuration means. "Delete all ci-operator configuration that exist for repos in the target org, but which would not be created by the current ci-operator-config-mirror run"

test/integration/ci-operator-config-mirror.sh
```diff
@@ -10,7 +10,7 @@ expected="${suite_dir}/output"
 os::test::junit::declare_suite_start "integration/ci-operator-config-mirror"
 # This test validates the ci-operator-config-mirror tool
 
-os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual} --whitelist-file ${suite_dir}/whitelist.yaml"
+os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual} --deprecate-configs=false --whitelist-file ${suite_dir}/whitelist.yaml"
```
petr-muller, line 13: We should have a test with `--deprecated-configs=true`, too (as that is the configuration we will want to run).

cmd/ci-operator-config-mirror/main.go
```diff
@@ -106,21 +152,27 @@ func main() {
...
-		if err := dataWithInfo.CommitTo(o.configDir); err != nil {
-			logger.WithError(err).Fatal("couldn't create the configuration file")
-		}
+		})
 
 		return nil
 	}
 
 	if err := config.OperateOnCIOperatorConfigDir(o.configDir, callback); err != nil {
 		logrus.WithError(err).Fatal("error while generating the ci-operator configuration files")
```
petr-muller, line 164: Looks like we no longer actually generate anything in `callback`, so this message needs to be changed.

cmd/ci-operator-config-mirror/main.go
```diff
@@ -51,6 +56,45 @@ func (o *options) validate() error {
 	return o.WhitelistOptions.Validate()
 }
 
+type dataWithInfoByRepo map[string][]config.DataWithInfo
```
petr-muller, line 59: Can we call the type `configsByRepo`? It's simpler and carries better meaning, IMO.

cmd/ci-operator-config-mirror/main.go
```diff
@@ -51,6 +56,45 @@ func (o *options) validate() error {
...
+	}); err != nil {
+		return fmt.Errorf("Failed to walk through the destination dir: %v", err)
+	}
+
+	for _, deprecatedPath := range deprecatedPaths {
+		logrus.WithField("deprecated-path", deprecatedPath).Warn("Removing deprecated repository folder")
+		if err := os.RemoveAll(deprecatedPath); err != nil {
+			return fmt.Errorf("couldn't cleanup repository dir: %v", err)
+		}
+	}
```
petr-muller, line 82: I am puzzled about this... We first build an index of configs by repo (and that index by definition only contains repos for which we have source configs), then delete all subdirectories in the destination directory that we have in that index. Why can't we simply delete all subdirectories in the destination directory? It would be simpler, and would take care of the care where the entire source repo config was removed (or e.g. removed from the whitelist).

So, for example, given this input:
```
.
├── openshift
│   └── api
│       ├── openshift-api-master.yaml
│       └── OWNERS
└── openshift-priv
    ├── api
    │   ├── openshift-priv-api-master.yaml
    │   ├── openshift-priv-api-REMOVED-FROM-SOURCE.yaml
    │   └── OWNERS
    └── stale
        ├── OWNERS
        └── openshift-priv-stale-master.yaml
```

The current code will only clean up the `REMOVED-FROM-SOURCE` config from `openshift-priv/api` but not the entire `openshift-priv/stale` directory which is IMO deprecated too and should be removed.

droslean: We don't want to remove the whole directory, because there is always a chance that only a branch will need to be removed.

droslean:
> I'm wondering if we could just remove all directories under the config/destination-org/ before we mirror the config... It would be much simpler and even cover more cases? (more inline).

`.config.prowgen` :)

cmd/ci-operator-config-mirror/main.go
petr-muller: > We don't want to remove the whole directory, because there is always a chance that only a branch will need to be removed.

Yeah, but in that case the configs for the remaining branches will be normally mirrored from the source configs.

petr-muller:
> `.config.prowgen` :)

@droslean `.config.prowgen` is on org level, that's why I said "all subdirectories". We would delete subdirs and `.config.prowgen` would stay.

droslean:
Yea, I guess that should work.  Let me check that first

cmd/ci-operator-config-mirror/main.go
```diff
@@ -51,6 +56,40 @@ func (o *options) validate() error {
...
+		return nil
+	}); err != nil {
+		return fmt.Errorf("Failed to walk through the destination dir: %v", err)
+	}
+
+	for _, subDir := range subDirs {
+		if err := os.RemoveAll(subDir); err != nil {
+			return fmt.Errorf("couldn't delete dir: %v", err)
+		}
+	}
```
petr-muller, line 77: Looks unnecessarily complicated, the `Walk` is not actually needed (it implies there is a hierarchy, but there is really just a single level of subdirs). It should be enough to `ioutil.ReadDir` and iterate over that, skipping everything that is not a dir.

test/integration/ci-operator-config-mirror.sh
```diff
@@ -10,7 +10,7 @@ expected="${suite_dir}/output"
 os::test::junit::declare_suite_start "integration/ci-operator-config-mirror"
 # This test validates the ci-operator-config-mirror tool
 
-os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual} --whitelist-file ${suite_dir}/whitelist.yaml"
+os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual} --clean=false --whitelist-file ${suite_dir}/whitelist.yaml"
```
petr-muller, line 13: I already wrote this in an earlier review, but I think we should have a test that runs `--clean=true` as that will be the default, desired mode.

droslean: I will address this. Thanks, it seems that I missed that one.

test/integration/ci-operator-config-mirror.sh
```diff
@@ -2,15 +2,23 @@
...
 
 os::test::junit::declare_suite_start "integration/ci-operator-config-mirror"
 # This test validates the ci-operator-config-mirror tool
 
-os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual} --whitelist-file ${suite_dir}/whitelist.yaml"
+os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual} --clean=false --whitelist-file ${suite_dir}/whitelist.yaml"
 os::integration::compare "${actual}" "${expected}"
 
+os::cmd::expect_success "ci-operator-config-mirror --to-org super-priv --config-path ${actual_to_clean} --clean=true --whitelist-file ${suite_dir}/whitelist.yaml"
+os::integration::compare "${actual_to_clean}" "${expected}"
```
petr-muller, line 21: This does not seem to test the functionality though. The input and output `super-priv` are identical, so this test does not show the tool actually removing anything:

```
❯ diff -rq input-to-clean/super-priv/ output/super-priv/

test/integration/ci-operator-config-mirror on  deprecate-configs [$] 
❯ echo $status
0
```

We should have both 1) a full directory in original `super-priv` that gets removed and 2) a single file that gets removed from a directory where other files are kept. This way we actually test that the tool does what it claims.

droslean: addressed

PR #795, opened 2020-05-13
autoowners: Add signoff to git commit message if requested.

### LINKED ISSUE #787: autoowners: add signoff option
There should be an option to use the signoff feature for the created commit. 

The test-infra dco plugin that we use for kubevirt org does not have the option to whitelist the bot explicitly. Therefore dco check fails on the created PR.

We fixed that by amending the commit with a signoff and re-pushing it again. https://github.com/kubevirt/project-infra/pull/429/files#diff-e0a716b1a6d9204c23041c4e82ea7360R178

dhiller:
When the repository the PR is created against is configured to use the [dco plugin](https://github.com/kubernetes/test-infra/blob/master/prow/plugins/dco/dco.go#L80), the check fails because the commit is missing the signoff that is required by dco.

This PR adds the signoff.

Fixes #787

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -34,6 +34,8 @@ const (
 
 	configSubDirs      = "jobs,config,templates"
 	targetSubDirectory = "ci-operator"
+
+	signedOffPattern = "Signed-off-by: %s <%s>"
 )
 
 type SimpleConfig = repoowners.SimpleConfig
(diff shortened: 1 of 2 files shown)


dhiller:
> Could we just use `git commit -S` ?

Not directly. We would have to introduce a [change in test-infra bumper experiment](https://github.com/openshift/ci-tools/blob/66dcd018e6d627c9bc05fd5513c3ab416770e94f/vendor/k8s.io/test-infra/experiment/autobumper/bumper/bumper.go#L161), so that we could add it there.

petr-muller:
What to do here? To me it looks like the bumper lib would really be a better place to put this...

dhiller:
> What to do here? To me it looks like the bumper lib would really be a better place to put this...

True, didn't get to it. Sorry. Maybe just close this?

PR #802, opened 2020-05-14
test: add an e2e test that runs simple commands

stevekuznetsov:

Biggest question is ... will we have the RBAC we need?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,11 +133,8 @@ production-install:
 
 # Run integration tests.
 #
-# Accepts a specific suite to run as an argument.
-#
 # Example:
 #   make integration
-#   make integration SUITE=multi-stage
 integration:
 	# legacy, so we don't break them
 	test/repo-init-integration/run.sh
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
OK, cool. The job fails as expected. The `default` service account in the namespace is not the `ci-operator` service account we usually run the jobs with, so the job itself cannot do `ci-operator` inception. @bbguimaraes @petr-muller do you think there are any sensible options for how to give the correct credentials to this test without the credential proposal?

petr-muller:
> do you think there are any sensible options for how to give the correct credentials to this test without the credential proposal?

Well the credential proposal only gives the multi-stage tests the ability to receive a secret from the `ci` namespace - for container-based tests we can to it even today with [`tests.secrets`](https://github.com/openshift/ci-tools/blob/master/CONFIGURATION.md#testssecrets). So if we can pass sufficient credentials to the test to be able to run ci-operator, we should be able to do that even today.

stevekuznetsov:
@petr-muller interesting.

PR #806, opened 2020-05-15
ci-operator: fix typos

petr-muller:
Noticed while skimming https://github.com/openshift/ci-tools/pull/804

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -323,7 +323,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.BoolVar(&opt.promote, "promote", false, "When all other targets complete, publish the set of images built by this job into the release configuration.")
 
 	// output control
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -323,7 +323,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.BoolVar(&opt.promote, "promote", false, "When all other targets complete, publish the set of images built by this job into the release configuration.")
 
 	// output control
-	flag.StringVar(&opt.artifactDir, "artifact-dir", "", "If set grab artifacts from test and template jobs. Defaults to $ARTIFACTS if set.")
+	flag.StringVar(&opt.artifactDir, "artifact-dir", "", "If set grab artifacts from test and template jobs. Defaults to $ARTIFACTS if not set.")
```
bbguimaraes, line 326: I think the original meant "Defaults to $ARTIFACTS if [$ARTIFACTS is] set". We can shorten it to just "Defaults to $ARTIFACTS", IMO.

petr-muller: Oh, right!

petr-muller:
/close

NVM

PR #807, opened 2020-05-15
ci-operator: annotate namespace as active while acquiring lease

stevekuznetsov:

/assign @bbguimaraes


bbguimaraes:
I think this could be done at a higher level, not here. There is already another point in the execution where we update the annotation.

Merging because it at least eliminates this failure scenario for now.

/lgtm

PR #809, opened 2020-05-15
Limit concurrency on ci-secret-bootstrap

hongkailiu:

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -1,11 +1,13 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 	"io"
 	"io/ioutil"
 	"os"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -13,6 +15,8 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/google/go-cmp/cmp"
 	"github.com/sirupsen/logrus"
+
+	"golang.org/x/sync/semaphore"
 
 	coreapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
@@ -45,6 +49,8 @@ type options struct {
 	bwPassword     string
 	secretsGetters map[string]coreclientset.SecretsGetter
 	config         []secretConfig
+
+	maxConcurrency int
 }
 
 func parseOptions() options {
@@ -59,6 +65,7 @@ func parseOptions() options {
 	fs.BoolVar(&o.force, "force", false, "If true, update the secrets even if existing one differs from Bitwarden items instead of existing with error. Default false.")
 	fs.StringVar(&o.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
 	fs.StringVar(&o.impersonateUser, "as", "", "Username to impersonate")
+	fs.IntVar(&o.maxConcurrency, "concurrency", 0, "Maximum number of concurrent in-flight goroutines to BitWarden.")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Errorf("cannot parse args: %q", os.Args[1:])
 	}
@@ -141,6 +148,10 @@ func (o *options) completeOptions(secrets *sets.String) error {
 			secretConfig.To = to
 			o.config = append(o.config, secretConfig)
 		}
+	}
+
+	if o.maxConcurrency == 0 {
+		o.maxConcurrency = runtime.GOMAXPROCS(0)
 	}
 
 	return o.validateCompletedOptions()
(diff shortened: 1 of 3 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -421,7 +438,8 @@ func main() {
 	})
 	defer logrus.Exit(0)
```
hongkailiu, line 439: Can I remove `defer logrus.Exit(0)`?

If we hit error later, fail the job, and send the alert.

alvaroaleman: I guess? No idea what the purpose of it is

cmd/ci-secret-bootstrap/main.go
```diff
@@ -267,8 +274,18 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
 			keyWG := &sync.WaitGroup{}
```
stevekuznetsov, line 274: You don't need this any longer, you can just acquire the max amount from the semaphore and use that

cmd/ci-secret-bootstrap/main.go
```diff
@@ -267,8 +274,18 @@ func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[str
...
 			for _, key := range keys {
 				keyWG.Add(1)
+				if err := sem.Acquire(ctx, 1); err != nil {
+					errLock.Lock()
+					errs = append(errs, fmt.Errorf("failed to acquire semaphore: %v", err))
```
stevekuznetsov, line 279: Let's use a channel to aggregate errors

cmd/ci-secret-bootstrap/main.go
stevekuznetsov: This was added in 6c0e4425ec5566849c6cf3f86e46dedfe99cedf2 and is needed to ensure that logout always happens on process exit.

stevekuznetsov: (`defer` did not work since `logrus.Fatal` calls `os.Exit` and the `logrus` defer alone did not logout on successful runs)

hongkailiu: Is this the logic?
successful runs (no calls of `logrus.Fatal`), we use `logrus.Exit(0)` to activate `logout`.  

unsuccessful runs: `logrus.Fatal` is called which calls `os.Exit(1)` and `defer logrus.Exit(0)` wont be executed.

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go

cmd/ci-secret-bootstrap/main.go
stevekuznetsov: Yes

cmd/ci-secret-bootstrap/main.go
```diff
@@ -242,12 +248,20 @@ type secretConfig struct {
...
-func constructSecrets(config []secretConfig, bwClient bitwarden.Client) (map[string][]*coreapi.Secret, error) {
+func constructSecrets(ctx context.Context, config []secretConfig, bwClient bitwarden.Client, maxConcurrency int) (map[string][]*coreapi.Secret, error) {
+	sem := semaphore.NewWeighted(int64(maxConcurrency))
 	secretsMap := map[string][]*coreapi.Secret{}
 	secretsMapLock := &sync.Mutex{}
 
+	errChan := make(chan error)
 	var errs []error
-	errLock := &sync.Mutex{}
+	errReadingDone := make(chan struct{})
```
stevekuznetsov, line 258: Accumulate the errors after the works have finished, and you don't need this.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -59,6 +64,7 @@ func parseOptions() options {
 	fs.BoolVar(&o.force, "force", false, "If true, update the secrets even if existing one differs from Bitwarden items instead of existing with error. Default false.")
 	fs.StringVar(&o.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
 	fs.StringVar(&o.impersonateUser, "as", "", "Username to impersonate")
+	fs.IntVar(&o.maxConcurrency, "concurrency", 25, "Maximum number of concurrent in-flight goroutines to BitWarden.")
```
petr-muller, line 67: So, why is 25 a good default?

stevekuznetsov: I would be happy with GOMAXPROCS

cmd/ci-secret-bootstrap/main.go
hongkailiu: My first version was like that (having no `errReadingDone`). But I had hard time to make the tests stable.
Until I found similar code:

https://github.com/openshift/ci-tools/blob/53a01129e7320d7903abc73423b8c95d5cf6a3aa/cmd/sanitize-prow-jobs/main.go#L114-L120

Then all tests are green

```
go test -count=100 -run TestConstructSecrets -v ./cmd/ci-secret-bootstrap/...
```

cmd/ci-secret-bootstrap/main.go
hongkailiu: Blind copying from 

https://github.com/kubernetes/test-infra/blob/e41d9b64ebfd3db18ed9075eb2b7ec796fb215f9/ghproxy/ghproxy.go#L121

May plan is to try to increase the number if 25 works and see if it gets any faster.
Decrease to 1 if 25 does not work. 1 will tell us if it is the concurrent problem at all.


Do you have a better number in mind?

stevekuznetsov: Fine to keep this

cmd/ci-secret-bootstrap/main.go
stevekuznetsov: ?

cmd/ci-secret-bootstrap/main.go
hongkailiu: Took `GOMAXPROCS `.

PR #810, opened 2020-05-15
rehearse: only include targeted test in CONFIG_SPEC

AlexNPavel:
Due to including all tests and resolving all multistage tests in the ci-operator config passed to rehearsals via `CONFIG_SPEC`, the config is starting to become too large to hold in an environment variable for some repos with a large amount of tests. This PR changing the inlining to only include the test that is targeted by the job, if one exists. This reduces the size of the `CONFIG_SPEC` env var considerably.


droslean:
Before going further with this PR, since your are doing it, can we wonder what else we can exclude? We don't need also the `base_images` and `images` to run a specific test, right?

stevekuznetsov:
> Before going further with this PR, since you are doing it, can we wonder what else we can exclude? We don't need also the base_images and images to run a specific test, right?

@droslean not true, tests often depend on input images that are built.

pkg/rehearse/jobs.go
```diff
@@ -282,26 +304,16 @@ func inlineCiOpConfig(container *v1.Container, ciopConfigs config.DataByFilename
...
-			loggers.Job.WithError(err).Error("Failed to marshal ci-operator config file")
+			loggers.Job.WithError(err).Error("Failed to get resolved config for test")
 			return err
 		}
 
 		envs := container.Env
 		env := v1.EnvVar{
 			Name:  "CONFIG_SPEC",
-			Value: string(ciOpConfigContent),
+			Value: ciOpConfigContent,
```
stevekuznetsov, line 316: Could we add a sensible heuristic here to detect that we're likely going to overflow the allowed limit on `exec` and fail when we run the Pod?

pkg/rehearse/jobs.go
```diff
@@ -364,7 +376,9 @@ func (jc *JobConfigurer) ConfigurePeriodicRehearsals(periodics config.Periodics)
 			metadata.Repo = job.ExtraRefs[0].Repo
 			metadata.Branch = job.ExtraRefs[0].BaseRef
 		}
-		if err := jc.configureJobSpec(job.Spec, metadata, jc.loggers.Debug.WithField("name", job.Name)); err != nil {
+		jobNameWithoutTest := metadata.JobName(jobconfig.PeriodicPrefix, "")
```
stevekuznetsov, line 379: Can we add a method to `Metadata` to get the test name instead? This is part of the API.

petr-muller: I think we already have that:

https://github.com/openshift/ci-tools/blob/307b5c731889f13777fa0e361cdb05418ab72237/pkg/api/metadata.go#L39-L46

stevekuznetsov: Not quite ... we have a "test name" meaning the identifier devs use like "unit", then a "test name" as used in `/test foo` like "variant-unit" and a "test name" as in `pull-ci-org-repo-branch-variant-name` ... that function you linked gives you the second from the first, I want the first from the last.

pkg/rehearse/jobs.go
AlexNPavel: That would be quite difficult. The maximum varies based on the environment it's running in, and is the size of all of the command line arguments and all the environment variables combined.

pkg/rehearse/jobs.go
AlexNPavel: The `TestName` method isn't useful for us in the rehearse code. That's just a helper function or `JobName` that takes the `testName` and appends a `-variant` if the metadata contains a variant or just the given `testName` unmodified if there is no variant.

This does make me realize that the code I wrote doesn't account for variants though. I'll make a function in `Metadata` that can handle all of this including variants.

PR #813, opened 2020-05-18
ci-operator: allow custom credentials in steps

stevekuznetsov:

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -438,7 +438,53 @@ func validateLiteralTestStep(fieldRoot string, step LiteralTestStep, seen sets.S
 		ret = append(ret, fmt.Errorf("%s: `commands` is required", fieldRoot))
 	}
 	ret = append(ret, validateResourceRequirements(fieldRoot+".resources", step.Resources)...)
+	ret = append(ret, validateCredentials(fieldRoot, step.Credentials)...)
 	return
+}
+
+func validateCredentials(fieldRoot string, credentials []CredentialReference) []error {
+	var errs []error
+	for i, credential := range credentials {
+		if credential.Name == "" {
+			errs = append(errs, fmt.Errorf("%s.credentials[%d].name cannot be empty", fieldRoot, i))
+		}
+		if credential.Namespace == "" {
+			errs = append(errs, fmt.Errorf("%s.credentials[%d].namespace cannot be empty", fieldRoot, i))
+		}
+		if credential.MountPath == "" {
+			errs = append(errs, fmt.Errorf("%s.credentials[%d].mountPath cannot be empty", fieldRoot, i))
+		} else if !filepath.IsAbs(credential.MountPath) {
+			errs = append(errs, fmt.Errorf("%s.credentials[%d].mountPath is not absolute: %s", fieldRoot, i, credential.MountPath))
+		}
+		for j, other := range credentials[i+1:] {
+			index := i + j + 1
+			if credential.MountPath == other.MountPath {
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s)", fieldRoot, i, index, credential.MountPath))
+				continue
+			}
+			// we can make a couple of assumptions here to improve our check:
+			//  - valid mount paths must be absolute paths
+			//  - given two absolute paths, a relative path between A and B will
+			//    never contain '..' if B is a subdirectory of A
+			relPath, err := filepath.Rel(other.MountPath, credential.MountPath)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] could not check relative path to credentials[%d] (%s)", fieldRoot, i, index, err))
+				continue
+			}
+			if !strings.Contains(relPath, "..") {
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] mounts at %s, which is under credentials[%d] (%s)", fieldRoot, i, credential.MountPath, index, other.MountPath))
+			}
+			relPath, err = filepath.Rel(credential.MountPath, other.MountPath)
+			if err != nil {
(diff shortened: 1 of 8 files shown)


stevekuznetsov:
server is at https://server-ci-tools-813.svc.ci.openshift.org

pkg/api/config.go
```diff
@@ -438,9 +438,27 @@ func validateLiteralTestStep(fieldRoot string, step LiteralTestStep, seen sets.S
...
 	return
 }
 
+func validateCredentials(fieldRoot string, credentials []CredentialReference) []error {
+	var errs []error
+	for i, credential := range credentials {
+		for j, other := range credentials {
+			if i == j {
+				continue
+			}
```
bbguimaraes, line 451:
```suggestion
		for j, other := range credentials[i + 1:] {
```

pkg/api/config.go
```diff
@@ -438,9 +438,27 @@ func validateLiteralTestStep(fieldRoot string, step LiteralTestStep, seen sets.S
...
+			if credential.MountPath == other.MountPath {
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s)", fieldRoot, i, j, credential.MountPath))
+			} else if strings.HasPrefix(credential.MountPath, other.MountPath) {
```
bbguimaraes, line 454: This incorrectly matches `/path_with_a_suffix` and `/path`, doesn't it?

pkg/api/types.go
```diff
@@ -445,6 +445,18 @@ type LiteralTestStep struct {
 	ArtifactDir string `json:"artifact_dir,omitempty"`
 	// Resources defines the resource requirements for the step.
 	Resources ResourceRequirements `json:"resources,omitempty"`
+	// Credentials defines the credentials we'll mount into this step.
+	Credentials []CredentialReference `json:"credentials,omitempty"`
```
bbguimaraes, line 449: Are we planning to replace the existing mechanism for `container` tests as well?

(fine as a separate PR)

pkg/steps/multi_stage.go
```diff
@@ -255,6 +258,33 @@ func (s *multiStageTestStep) createSecret() error {
...
+			// we don't want secrets imported from separate namespaces to collide
+			// but we want to keep them generally recognizable for debugging, and the
+			// chance we get a second-level collision (ns-a, name) and (ns, a-name) is
+			// small, so we can get away with this string prefixing
+			name := fmt.Sprintf("%s-%s", credential.Namespace, credential.Name)
+			if s.dry {
+				s.logger.AddObject(&coreapi.Secret{ObjectMeta: meta.ObjectMeta{Name: name}})
+				continue
+			}
+			raw, err := s.secretClient.Secrets(credential.Namespace).Get(credential.Name, meta.GetOptions{})
```
bbguimaraes, line 274: Does this default to the current namespace if not set?

pkg/steps/multi_stage.go
```diff
@@ -255,6 +258,33 @@ func (s *multiStageTestStep) createSecret() error {
 	return err
 }
 
+func (s *multiStageTestStep) createCredentials() error {
+	log.Printf("Creating multi-stage test credentials for %q", s.name)
+	for _, step := range append(s.pre, append(s.test, s.post...)...) {
+		for _, credential := range step.Credentials {
```
bbguimaraes, line 264: We'll likely have cases where more than one step in a test will use the same secret, so we should gather and deduplicate them before this loop.

bbguimaraes: Also, I had the option to declare credentials at higher levels of the hierarchy (chains, workflows) in the original design document, so we wouldn't need to repeat them in all steps if all of them needed a certain credential.

That can be done in a separate PR, if we think it's a good idea. Even then the implementation here would not change.

pkg/api/config.go
```diff
@@ -438,9 +438,27 @@ func validateLiteralTestStep(fieldRoot string, step LiteralTestStep, seen sets.S
 		ret = append(ret, fmt.Errorf("%s: `commands` is required", fieldRoot))
 	}
 	ret = append(ret, validateResourceRequirements(fieldRoot+".resources", step.Resources)...)
+	ret = append(ret, validateCredentials(fieldRoot, step.Credentials)...)
 	return
 }
 
+func validateCredentials(fieldRoot string, credentials []CredentialReference) []error {
```
bbguimaraes, line 445: Should validate that `name` and `mountPath` are not empty, for good measure.

Any objections to leaving `namespace` empty to get it from `ci-operator`'s current namespace (asked a similar question below)?

pkg/api/types.go
stevekuznetsov: I don't really want to, I wonder if we can make the current container steps just syntactic sugar for a mutli-stage test with one step?

stevekuznetsov:
@bbguimaraes updated

pkg/api/config.go
```diff
@@ -438,9 +438,47 @@ func validateLiteralTestStep(fieldRoot string, step LiteralTestStep, seen sets.S
...
+			//  - given two absolute paths, a relative path between A and B will
+			//    never contain '..' if B is a subdirectory of A
+			relPath, err := filepath.Rel(other.MountPath, credential.MountPath)
+			if err != nil {
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] could not check relative path to credentials[%d] (%s)", fieldRoot, i, index, err))
+				continue
+			}
+			if !strings.Contains(relPath, "..") {
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] mounts at %s, which is under credentials[%d] (%s)", fieldRoot, i, credential.MountPath, index, other.MountPath))
+			}
```
bbguimaraes, line 476: This should be done in reverse as well (try swapping the two credentials in the test).

stevekuznetsov: Ah, *that* was why I had the loop go over everything twice ;)

PR #816, opened 2020-05-19
webreg: doc API for $KUBECONFIG

stevekuznetsov:

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -913,6 +913,16 @@ or changes configuration, a later step could check for that as a pre-condition
 by using <code>oc</code> or the API to view the cluster's configuration.
 </p>
 
+<h5 id="kubeconfig"><a href="#kubeconfig">A Note on <code>$KUBECONFIG</code></a></h5>
+<p>
+In the default execution environment, commands run in steps will be given the
+<code>$KUBECONFIG</code> environment variable to allow them to interact with
+the ephemeral cluster that was created for testing. It is required that any
+steps which execute a cluster installation publish the resulting configuration
+file to <code>$SHARED_DIR/kubeconfig</code> to allow the <code>ci-operator</code>
+to correctly propagate this configuration to subsequent steps.
+</p>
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
server is at https://server-ci-tools-816.svc.ci.openshift.org

pkg/webreg/webreg.go
```diff
@@ -913,6 +913,16 @@ or changes configuration, a later step could check for that as a pre-condition
 by using <code>oc</code> or the API to view the cluster's configuration.
 </p>
 
+<h5 id="kubeconfig"><a href="#kubeconfig">A Note on <code>$KUBECONFIG</code></a></h5>
+<p>
+In the default execution environment, commands run in steps will be given the
+<code>$KUBECONFIG</code> environment variable to allow them to interact with
+the ephemeral cluster that was created for testing. It is required that any
+steps which execute a cluster installation publish the resulting configuration
+file to <code>$SHARED_DIR/kubeconfig</code> to allow the <code>ci-operator</code>
```
wking, line 922: Do we need to expand this?  I'm not clear if `$KUBECONFIG` is set for all steps, or just after we actually have content set.  If it's the former, we might want to say something like:

> In the default execution environment, commands run in all steps will be given the `$KUBECONFIG` environment variable.  For early steps, there will be no file at the given path.  The step creating the cluster must write its kubeconfig to `$KUBECONFIG` to allow the `ci-operator` to correctly propagate this configuration to subsequent steps.

or some such...

stevekuznetsov: Why is that distinction important?

wking: > Why is that distinction important?

If `$KUBECONFIG` is defined for all steps, even the one that installs the cluster, it just feels strange to me to have docs that effectively say:

> For you, oh special installer step, let me say that the value of `$KUBECONFIG` is actually `$SHARED_DIR/kubeconfig`, and you should use that half-expanded form instead.

PR #818, opened 2020-05-19
rehearse,registry: make explicit maps for type in NodeByName

AlexNPavel:
This changes the `NodeByName` type in `pkg/registry` to have explicit `References`, `Chains`, and `Workflows` fields as well as fixes a bug in the graphs handling of nested chains.

When the registry graph was first designed, it was assumed that nobody would make a chain and reference that have identical names (the `as` fields). This was not enforced in the release repo, however, and it has now become convention to have both a reference and a chain called `{i|u}pi-conf-cluster_type` for every `cluster_type` we support. This meant that the graph would overwrite references with the same names as chains. This could potentially cause a rehearsal to not occur for jobs that use just the reference with said name instead of the chain for whatever reason. It is also simply not ideal to overwrite data unintentionally anyway.

This also fixes a bug where the graph would only be able to nest one child chain per parent chain, even if there were multiple child chains. A new unit test has been added for this case.

/cc @stevekuznetsov

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -183,7 +183,7 @@ func rehearseMain() error {
 		affectedJobs = jobs
 	}
 
-	var changedRegistrySteps registry.NodeByName
+	var changedRegistrySteps []registry.Node
 	var refs registry.ReferenceByName
 	var chains registry.ChainByName
 	var workflows registry.WorkflowByName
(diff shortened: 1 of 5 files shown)


AlexNPavel:
/hold

This might be causing problems in the web UI

AlexNPavel:
/hold cancel

That's actually an issue with go 1.14 vs go 1.13 now that I have go 1.14 on my dev laptop. Not an issue with this PR, will work on fixing that issue in a different PR

PR #820, opened 2020-05-20
Enable `make pr-deploy` on app.ci

petr-muller:
We recently stopped updating `ci-operator-*-configs` CMs on `api.ci`,
and now I would like to remove them entirely. `make pr-deploy` needs
`ci-operator-master-configs` on the same cluster where it deploys, so
let's allow to deploy it on app.ci. Alternatively, we can change the
script to fetch the CM from `app.ci` with `--context app.ci` or
something similar.

- I was not able to `oc process` on app.ci without `--as system:admin`,
  so adding it.
- The replacement base images' imagestreams are not (yet) migrated to
  app.ci. Rather than figuring out what needs to move and what not, I
  disabled the replacements. This does not run often enough for this
  optimization to be crucial.

/cc @stevekuznetsov @AlexNPavel

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -175,7 +175,7 @@ update-integration:
 .PHONY: update-integration
 
 pr-deploy:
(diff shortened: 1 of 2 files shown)


petr-muller:
PoC: https://server-ci-tools-820.apps.ci.l2s4.p1.openshiftapps.com

PR #821, opened 2020-05-20
ci-operator: allow steps to reference external images

stevekuznetsov:
A test step from the registry needs to be able to reference an external
image so that it can ship together with the dependencies it needs.
Otherwise, every test configuration that ends up using that test step
will need to update its' `base_images` stanza, which is terrible for UX.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -125,6 +125,7 @@ func FromConfig(
 	var hasReleaseStep bool
 	for _, rawStep := range stepConfigsForBuild(config, jobSpec) {
 		var step api.Step
+		var additional []api.Step
 		var stepLinks []api.StepLink
 		if rawStep.InputImageTagStepConfiguration != nil {
 			srcClient, err := clusterImageStreamClient(imageClient, clusterConfig, rawStep.InputImageTagStepConfiguration.BaseImage.Cluster, kubeconfigs)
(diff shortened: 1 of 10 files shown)


stevekuznetsov:
This is ready for review

stevekuznetsov:
```
 2020/05/23 00:44:22 The namespace in which this test is executing has been deleted, cancelling the test...
Caching blobs under "/var/cache/blobs".
error: error creating buildah builder: Error parsing image configuration: Get https://registry.svc.ci.openshift.org/v2/ci/clonerefs/blobs/sha256:3b03def3a727c227882f402138f3b028e915efc3749fbaeaa9c66777c9f74783: dial tcp [2600:1901:0:f1d::]:443: connect: network is unreachable 
```

???

/retest

stevekuznetsov:
@alvaroaleman don't you laugh at the broken tests ... they aren't my fault!

stevekuznetsov:
Added docs:

https://server-ci-tools-821.apps.ci.l2s4.p1.openshiftapps.com/help#step-from

pkg/defaults/defaults.go
```diff
@@ -177,6 +178,19 @@ func FromConfig(
...
+					if link, ok := subStep.FromImageTag(); ok {
+						config := api.InputImageTagStepConfiguration{
+							BaseImage: *subStep.FromImage,
+							To:        link,
+						}
+						srcClient, err := clusterImageStreamClient(imageClient, clusterConfig, config.BaseImage.Cluster, kubeconfigs)
+						if err != nil {
+							return nil, nil, fmt.Errorf("unable to access image stream tag on remote cluster: %v", err)
+						}
+						additional = append(additional, steps.InputImageTagStep(config, srcClient, imageClient, jobSpec, dryLogger))
```
bbguimaraes, line 191: Does the execution graph handle duplicates or will we start two independent steps?

pkg/webreg/webreg.go
```diff
@@ -774,11 +774,25 @@ step:
...
 </p>
 
-<h4 id="step-from"><a href="#step-from"><code>from</code></a></h4>
+<h4 id="step-image"><a href="#step-image">Configuring The Container Image For a Step</a></h4>
 
 <p>
-The image must be present in the <code>ci-operator</code> configuration file,
-either:
+The container image used to run a test step can be configured in one of two
+ways: by referencing and image tag otherwise present in the configuration or
```
bbguimaraes, line 781:
```suggestion
ways: by referencing an image tag otherwise present in the configuration or
```

pkg/defaults/defaults.go
stevekuznetsov: They will just race to create the object, as is the case already between two different jobs that execute the same steps, so it should not be an issue.

stevekuznetsov:
@bbguimaraes @petr-muller updated

pkg/defaults/defaults.go
petr-muller: I don't think it should break anything, too. I think it's the first time we let things race this way in a single ci-op run, though, so it will be worth keeping the eye on.

PR #824, opened 2020-05-20
Add a testimages distributing controller

alvaroaleman:
This PR adds a controller that will:
* Find all imageStreamTags in the `api.ci` cluster
* Filter out those that are not referenced by any test
* Create a reconciliaton request for all configured build clusters for each filtered ImageStramTag
* Check if the build cluster already has the latest version of the ImageStreamTag
* If not:
    * It will create the target namespace, if it does not yet exist
    * It will upsert an imagePullSecret in that namespace
    * It will create an ImageStreamImport in that namespace


pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+	if err := client.Get(r.ctx, types.NamespacedName{Name: decoded.Namespace}, &corev1.Namespace{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to check if namespace %s exists: %w", decoded.Namespace, err)
+		}
+		if err := client.Create(r.ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: decoded.Namespace}}); err != nil && !apierrors.IsAlreadyExists(err) {
+			return fmt.Errorf("failed to create namespace %s: %w", decoded.Namespace, err)
+		}
+	}
+
+	if err := r.ensureImagePullSecret(decoded.Namespace, client); err != nil {
```
alvaroaleman, line 220: Not sure if its the right thing to have this here. We also have that secret mirroring controller somewhere which could do it as well. The thought was that this easens debugging and testing, because we don't have our business logic scattered around in being partly here and partly done by configuring a generic controller to do the subset of our businesslogic its able to do.

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+	l = logrus.WithField("subcomponent", "test-input-image-stream-tag-filter")
+	return func(nn types.NamespacedName) bool {
+		result, err := ca.GetFromIndex(indexName, nn.String())
+		// Today, GetFromIndex only errors if the index does not exist, so this should
+		// never happen.
+		if err != nil {
+			l.WithField("name", nn.String()).WithError(err).Error("Failed to get configs from index")
+			return false
+		}
+		return len(result) > 0
```
alvaroaleman, line 349: This is not extremely smart or efficient, but the best we can do right now without having to extend the config agent with support for generic post-config-load hooks. IMHO its good enough and I'd prefer to not extend the public api of the config agent.

pkg/api/helper/imageextraction.go
```diff
@@ -38,6 +28,30 @@ func GetAllTestInputImageStreamTags(config load.ByOrgRepo) map[string]types.Name
 	return result
 }
 
+// TestInputImageStreamTagsFromConfig returns all ImageStreamTags referenced anywhere in the config as input.
```
stevekuznetsov, line 31: One of us will be rebasing :(

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -239,18 +229,18 @@ func refForIST(ist *imagev1.ImageStreamTag) (*branchReference, error) {
 	commit := metadata.Config.Labels["io.openshift.build.commit.id"]
 	sourceLocation := metadata.Config.Labels["io.openshift.build.source-location"]
 	if branch == "" {
-		return nil, nonRetriableError{errors.New("imageStreamTag has no `io.openshift.build.commit.ref` label, can't find out source branch")}
+		return nil, controllerutil.TerminalError(errors.New("imageStreamTag has no `io.openshift.build.commit.ref` label, can't find out source branch"))
 	}
 	if commit == "" {
-		return nil, nonRetriableError{errors.New("ImageStreamTag has no `io.openshift.build.commit.id` label, can't find out source commit")}
+		return nil, controllerutil.TerminalError(errors.New("ImageStreamTag has no `io.openshift.build.commit.id` label, can't find out source commit"))
```
stevekuznetsov, line 235: This could/should be split out of this PR?

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+	return &handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(
+		func(mo handler.MapObject) []reconcile.Request {
+			imageStream, ok := mo.Object.(*imagev1.ImageStream)
+			if !ok {
+				logrus.WithField("type", fmt.Sprintf("%T", mo.Object)).Error("Got object that was not an ImageStram")
+				return nil
+			}
+			var requests []reconcile.Request
+			for _, imageStreamTag := range imageStream.Spec.Tags {
+				// Not sure why this happens but seems to be a thing
```
stevekuznetsov, line 118: Can we log other metadata to debug

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+					continue
+				}
+				name := types.NamespacedName{
+					Namespace: mo.Meta.GetNamespace(),
+					Name:      fmt.Sprintf("%s:%s", mo.Meta.GetName(), imageStreamTag.Name),
+				}
+				if !filter(name) {
+					continue
+				}
+				// We have to squeeze both the target cluster name and the imageStreamTag name into a reconcile.Request
```
stevekuznetsov, line 129: Instead of a workqueue per build cluster?

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+func (r *reconciler) reconcile(req reconcile.Request, log *logrus.Entry) error {
+	cluster, decoded, err := decodeRequest(req)
+	if err != nil {
+		return fmt.Errorf("failed to decode request %s: %w", req, err)
+	}
+	// Propagate the cluster field back up
+	*log = *log.WithField("cluster", cluster)
```
stevekuznetsov, line 185: What's with the pointer dereferences here?

pkg/controller/test-images-distributor/test_images_distributor.go
stevekuznetsov: Yeah I think this is fine

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+		referenceSecret.ResourceVersion = resourceVersion
+		if err := client.Update(r.ctx, referenceSecret); err != nil {
+			return fmt.Errorf("failed to update secret %s/%s: %w", referenceSecret.Namespace, referenceSecret.Name, err)
+		}
+	}
+
+	return nil
+}
+
+func (r *reconciler) generateReferencPullSecret(namespace string) *corev1.Secret {
```
stevekuznetsov, line 322: `s/Referenc/Reference/`

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -0,0 +1,364 @@
...
+	m := apihelper.TestInputImageStreamTagsFromConfig(cfg)
+	var result []string
+	for key := range m {
+		result = append(result, key)
+	}
+	return result
+}
+
+func publicURLForImage(potentiallyPrivate string) string {
+	return strings.ReplaceAll(potentiallyPrivate, "docker-registry.default.svc:5000", "registry.svc.ci.openshift.org")
```
stevekuznetsov, line 363: this should realistically be configured with flags?

pkg/controller/test-images-distributor/test_images_distributor.go
alvaroaleman: It sets the value that the pointer we get points to, thus propagating the new field up the logger in `Reconcile` so when we log errors there we have the cluster as a field

pkg/controller/test-images-distributor/test_images_distributor.go
alvaroaleman: yeah, that would be the alternate. I guess my reason for doing it like this is that I think its good if we can limit the total number of parallel imports by setting the worker count. In terms of complexity I would assume doing this vs creating multiple controllers at runtime is probably about the same.

pkg/controller/test-images-distributor/test_images_distributor.go
alvaroaleman: Well, effectively this is a const for us

pkg/controller/promotionreconciler/reconciler.go
alvaroaleman: https://github.com/openshift/ci-tools/pull/830

cmd/dptp-controller-manager/main.go
```diff
@@ -140,16 +178,69 @@ func main() {
...
+
+	if !opts.disabledControllersSet.Has(testImagesDistributorName) {
+		if opts.testImagesDistributorOptions.imagePullSecretPath == "" {
+			logrus.Fatal("The testImagesDistributor requires the --testImagesDistributorOptions.imagePullSecretPath flag to be set ")
+		}
+		data, err := ioutil.ReadFile(opts.testImagesDistributorOptions.imagePullSecretPath)
+		if err != nil {
+			logrus.Fatalf("failed to read --testImagesDistributorOptions.imagePullSecretPath from %s: %v", opts.testImagesDistributorOptions.imagePullSecretPath, err)
+		}
+		// TODO: We should be using testInfras secretAgent here. But it seems the thing can not cope with different types of secrets?
```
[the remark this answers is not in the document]
alvaroaleman: https://coreos.slack.com/archives/GB7NB0CUC/p1590091680196800

hack/run-dptp-controller-manager.sh
```diff
@@ -3,10 +3,33 @@
 set -euo pipefail
 
 cd $(dirname $0)/..
-CGO_ENABLED=0 go build -v -o /tmp/dptp-cm ./cmd/dptp-controller-manager
+
+kubeconfig="$(mktemp)"
+dockercfg=$(mktemp)
+#trap 'rm -f $kubeconfig $dockercfg' EXIT
```
stevekuznetsov, line 9: should not need a comment?

hack/run-dptp-controller-manager.sh
```diff
@@ -3,10 +3,33 @@
...
+kubectl config view >$kubeconfig
+IFS=$'\n'
+for additional_context in $(kubectl --kubeconfig $kubeconfig config get-contexts -o name|egrep -v 'app\.ci|api\.ci|build01'); do
+  kubectl --kubeconfig $kubeconfig config delete-context "$additional_context"
+  kubectl --kubeconfig $kubeconfig config delete-cluster "$additional_context" || true
+done
+export KUBECONFIG=$kubeconfig
+
+# Steve will make this nicer at some point. We need to preserve the `auths[$registryName] = {"auth":"value" }
```
stevekuznetsov, line 18: I will? :)

pkg/controller/test-images-distributor/test_images_distributor.go
stevekuznetsov: I added a const for `svc.ci.openshift.org` -- maybe we can share it?

hack/run-dptp-controller-manager.sh
alvaroaleman: Probably. If not it stays like that :)

alvaroaleman:
@stevekuznetsov updated. In the second commit I moved the const and helpers you created into `pkg/api` and made the typing proper, because those are apis, not `helpers` or `utils`. But I was too lazy to have the discussion on your PR :upside_down_face:

alvaroaleman:
> Where do we load ISTags imported as from_image in a registry step?

Mind if I do that in a follow-up?

PR #825, opened 2020-05-21
[private-org-sync] Add a source/destination merge logic after it fails to push the last retry.

droslean:
This PR does:
- Share the git add remote logic.

- add logic to merge source and destination history
    
    By keeping the current design, the code will still retry to fetch with deeper depth.
    After is will fail to push the unshallow repository, this commit adds the last resort.
    
    - Fetches the destination branch
    - Checkouts to FETCH_HEAD
    - Merge the already fetched source history
    - Attempt to push to $branch to the destination's remote.

/cc @petr-muller @stevekuznetsov @openshift/openshift-team-developer-productivity-test-platform 


cmd/private-org-sync/main.go
```diff
@@ -398,6 +392,89 @@ func (g gitSyncer) mirror(repoDir string, src, dst location) error {
...
+
+	logger.Debug("Initializing git repository")
+	if _, exitCode, err := git(logger, workDir, "init"); err != nil || exitCode != 0 {
+		logger.WithField("exit-code", exitCode).WithError(err).Error("Failed to initialize local git directory")
+		return fmt.Errorf("failed to initialize local git directory")
+	}
+
+	if err := addGitRemote(logger, git, srcOrg, srcRepo, workDir, srcRemote); err != nil {
+		return err
+	}
```
petr-muller, line 432: Quick note: we definitely should not need to create a new working directory and fetch the source. At this stage we already have a working directory with the source fully fetched, all we need to do is to fetch the destination, merge the source to it and push to destination again. This looks like it does more work that is actually needed.

droslean: Unfortunately the problem is that we have a source fully fetched under a bare git repo. I couldn't make that work without breaking the current design.

petr-muller: Can we use a non-bare working repo from the start? Until the merge we are only doing fetches and pushes which should behave identically.

droslean: I tried that, and the whole fetch a depth logic wasn't working. It also generated an empty git folder.  To be honest, I didn't spend much time debugging why, because I didn't intend to change the current design. However, let me try to find a way how to reuse the existing fetched source.

petr-muller: Interesting. At the very worst, we could clone the new non-bare repository from the bare one (that's at least a local operation), but I think we should make it work with a single repo (which needs to be non-bare to allow merges).

droslean:
@jupierce FYI

petr-muller:
/hold 
/assign

Looks good but still IMO does few steps that are not needed. I'd like to run few tests then get back here.

cmd/private-org-sync/main_test.go
```diff
@@ -467,6 +468,71 @@ func TestMirror(t *testing.T) {
...
+					call:     "push --tags --dry-run https://TOKEN@github.com/dest/repo FETCH_HEAD:refs/heads/branch",
+					exitCode: 1,
+					output:   "...Updates were rejected because the remote contains work that you do...",
+				},
+				{call: "remote add dest-repo https://github.com/dest/repo"},
+				{call: "merge org-repo/branch"},
+				{call: "checkout -b branch-priv"},
+				{call: "fetch dest-repo"},
+				{call: "merge dest-repo/branch -m 'Periodic merge from DPTP; pub->priv'"},
+				{call: "push --tags --dry-run https://TOKEN@github.com/dest/repo branch-priv:branch"},
```
petr-muller, line 533: The testcase illustrates how this can be done even simpler: there's no need to make a new named remote nor a named local branch.

```
git fetch https://TOKEN@github.com/dest/repo branch # fetch destination (and only it's branch, not all of them)
git co FETCH_HEAD                                                     # checks out the destination branch so it can be merged into
git merge org-repo/branch                                             # merge the source branch into the current one
git push https://TOKEN@github.com/dest/repo HEAD:branch # push to destination
```

cmd/private-org-sync/main.go
```diff
@@ -398,6 +392,67 @@ func (g gitSyncer) mirror(repoDir string, src, dst location) error {
...
+	if _, exitCode, err := git(logger, repoDir, "remote", "add", remoteName, srcUrl.String()); err != nil || exitCode != 0 {
+		remoteSetupLogger.WithField("exit-code", exitCode).WithError(err).Error("Failed to set up source remote")
+		return fmt.Errorf("failed to set up source remote")
+	}
+
+	return nil
+}
+
+func mergeRemotesAndPush(logger *logrus.Entry, git gitFunc, repoDir, srcRemote, destRemote, org, repo, branch, destUrl string, confirm bool) error {
+	if err := addGitRemote(logger, git, org, repo, repoDir, destRemote); err != nil {
```
petr-muller, line 417: There's no need to make a named remote for this: we can easily use the URL instead in the `fetch/push` commands. We already built the URL in the `mirror` method, in the `destUrl` variable.

And if we do not need to add a named remote, we do not need `addGitRemote` as a new, separate method ;)

cmd/private-org-sync/main.go
```diff
@@ -398,6 +392,67 @@ func (g gitSyncer) mirror(repoDir string, src, dst location) error {
...
+		return err
+	}
+
+	sourceBranch := fmt.Sprintf("%s/%s", srcRemote, branch)
+	if _, _, err := git(logger, repoDir, []string{"merge", sourceBranch}...); err != nil {
```
petr-muller, line 422: This looks very convoluted... we basically need to do three operations: fetch the destination (and only the destination branch), merge the source branch (which we already have from the earlier operations) into it, and push it to the destination... There's no need to merge twice (a `checkout` suffices for what the first merge achiveves here). There's also no need to create a local named branch because you can use the `FETCH_HEAD` and `HEAD` identifiers.

cmd/private-org-sync/main.go
```diff
@@ -398,6 +392,67 @@ func (g gitSyncer) mirror(repoDir string, src, dst location) error {
...
+		return fmt.Errorf("failed to merge %s: %v", sourceBranch, err)
+	}
+
+	// Create another branch, fetch destination and merge
+	privBranch := fmt.Sprintf("%s-priv", branch)
+	if _, _, err := git(logger, repoDir, []string{"checkout", "-b", privBranch}...); err != nil {
+		return fmt.Errorf("failed to checkout to %s: %v", privBranch, err)
+	}
+
+	if _, _, err := git(logger, repoDir, []string{"fetch", destRemote}...); err != nil {
```
petr-muller, line 432: `fetch remote` fetches all branches but we only need one, sot this should be `fetch remote branch`.

cmd/private-org-sync/main.go
petr-muller: Please note that this basically says the same like https://github.com/openshift/ci-tools/pull/825#discussion_r429241984 does, it is just more easily visible on the testcase.

cmd/private-org-sync/main.go
```diff
@@ -351,9 +342,13 @@ func (g gitSyncer) mirror(repoDir string, src, dst location) error {
 		}
 
 		if depth == unshallow {
-			message := "failed to push to destination, no retry possible (already fetched full history)"
-			logger.Error(message)
-			return nil, fmt.Errorf(message)
+			logger.Info("Trying to fetch source and destination full history and perform a merge")
+			destRemoteURI := fmt.Sprintf("https://github.com/%s/%s", dst.org, dst.repo)
```
petr-muller, line 346: Please reuse `destUrl`: there's no point in building the URL twice, plus `destUrl` already contains the necessary token to access the (usually private) repo.

cmd/private-org-sync/main_test.go
```diff
@@ -467,6 +468,69 @@ func TestMirror(t *testing.T) {
...
+					output:   "...Updates were rejected because the remote contains work that you do...",
+				},
+				{call: "rev-parse --is-shallow-repository", output: "true"},
+				{call: "fetch --tags org-repo branch --unshallow"},
+				{
+					call:     "push --tags --dry-run https://TOKEN@github.com/dest/repo FETCH_HEAD:refs/heads/branch",
+					exitCode: 1,
+					output:   "...Updates were rejected because the remote contains work that you do...",
+				},
+				{call: "fetch https://github.com/dest/repo branch"},
```
petr-muller, line 528: The URL needs to have the TOKEN, too.

droslean: addressed

cmd/private-org-sync/main.go
droslean: addressed

petr-muller:
I'm really curious if the `--bare` removal will have an impact on sync performance.

PR #826, opened 2020-05-21
ci-operator: create a new secret instead of copying

stevekuznetsov:
When copying a secret from one spot to another, we need to be careful to
unset fields like the resource version and self-link. Instead, just copy
over the bits we need and create a new secret.

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -276,9 +276,16 @@ func (s *multiStageTestStep) createCredentials() error {
 			if err != nil {
 				return fmt.Errorf("could not read source credential: %v", err)
 			}
-			raw.Namespace = s.jobSpec.Namespace
-			raw.Name = name
-			toCreate[name] = raw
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
To test this, really, we need an e2e test. Once this lands and we have the secret available, I'll write such a test.

pkg/steps/multi_stage.go
```diff
@@ -276,9 +276,16 @@ func (s *multiStageTestStep) createCredentials() error {
 			if err != nil {
 				return fmt.Errorf("could not read source credential: %v", err)
 			}
-			raw.Namespace = s.jobSpec.Namespace
-			raw.Name = name
-			toCreate[name] = raw
+			toCreate[name] = &coreapi.Secret{
```
alvaroaleman, line 279: You could also just replace the embedded ObjectMeta, then you don't have to correctly copy over all fields

stevekuznetsov: You say potato, I say potato

PR #828, opened 2020-05-21
ci-operator: load KUEBCONFIG if set, then use in-cluster

stevekuznetsov:

### DIFF
--- a/pkg/util/client.go
+++ b/pkg/util/client.go
@@ -2,6 +2,7 @@ package util
 
 import (
 	"fmt"
+	"os"
 
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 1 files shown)


pkg/util/client.go
```diff
@@ -10,17 +11,27 @@ import (
...
+		return loadClusterConfig()
+	}
+
+	// otherwise, prefer in-cluster config, then fall back to normal loading
 	clusterConfig, err := rest.InClusterConfig()
 	if err == nil {
 		return clusterConfig, nil
 	}
 
+	return loadClusterConfig()
```
alvaroaleman, line 25: What? If env var is set use that, if not use inCluster, do not show failure, and go back to step 1? This doesn't make sense
/hold

alvaroaleman: Something like this, with an added pre-condition that sets the kubeconfigs flag value to the env var if unset: https://github.com/kubermatic/machine-controller/blob/23656f1f2c518c1e555b12362501b835d36794c3/cmd/machine-controller/main.go#L210

Or just the c-r stuff which does all of it for you.

stevekuznetsov:
Forgot to remove the unused method

stevekuznetsov:
```
 2020/05/21 18:42:08 Executing test validate-vendor
2020/05/21 18:46:40 Container test in pod validate-vendor completed successfully
2020/05/21 18:46:40 error: unable to signal to artifacts container to terminate in pod validate-vendor, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/05/21 18:46:40 error: unable to retrieve artifacts from pod validate-vendor: could not read gzipped artifacts: unable to upgrade connection: container not found ("artifacts") 
```

WTF?

/test validate-vendor

PR #830, opened 2020-05-21
Add re-usable terminal errors

alvaroaleman:
/assign @stevekuznetsov

### DIFF
--- a/pkg/controller/promotionreconciler/reconciler.go
+++ b/pkg/controller/promotionreconciler/reconciler.go
@@ -139,23 +139,18 @@ func (r *reconciler) Reconcile(req controllerruntime.Request) (controllerruntime
 	defer func() { log.WithField("duration", time.Since(startTime)).Trace("Finished reconciliation") }()
 
 	err := r.reconcile(req, log)
-	isNonRetriable := errors.Is(err, nonRetriableError{})
 	if err != nil {
-		log := log.WithError(err).WithField("not-retriable", isNonRetriable)
-		// Degrade non-retriable errors to debug, they most lilely just mean a given imageStreamTag wasn't built
+		log := log.WithError(err)
+		// Degrade terminal errors to debug, they most lilely just mean a given imageStreamTag wasn't built
 		// via ci operator.
-		if isNonRetriable {
+		if controllerutil.IsTerminal(err) {
 			log.Debug("Reconciliation failed")
 		} else {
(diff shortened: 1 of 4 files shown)


pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -139,23 +139,20 @@ func (r *reconciler) Reconcile(req controllerruntime.Request) (controllerruntime
...
-	}
 
-	if !isNonRetriable {
-		// Swallow non-retriable errors to avoid putting the item back into the workqueue
-		err = nil
+		log.WithError(err).Error("Reconciliation failed")
 	}
-	return controllerruntime.Result{}, err
+
+	return controllerruntime.Result{}, controllerutil.SwallowIfTerminal(err)
```
bbguimaraes, line 155: The logic here is a bit all over the place as a result of the refactoring, seems like it could be simplified to:

```go
err := r.reconcile(req, log)
transient := controllerutil.SwallowIfTerminal(err)
if transient == nil {
    log.WithError(err).Debug("Reconciliation failed")
} else {
    log.WithError(err).Error("Reconciliation failed")
}
return controllerruntime.Result{}, transient
```
Perhaps a good candidate for a function in `pkg/controller/error.go`?

alvaroaleman: Updated, PTAL

bbguimaraes: I still don't get why we do `log.Debug/log.Error` twice with (mostly) the same message.

alvaroaleman: The terminal errors for this controller mean that we found an imagestreamtag which doesn't have some labels set which we would need to figure out the job it was built from. Its not actionable, so we log it on debug level.

The transient error means that we encountered an unexpected error. They might require action, so they are logged on error level to make sure they grab our attention.

bbguimaraes: Right, but why do we log it as either `Debug` or `Error` based on that *then again* as `Error` unconditionally (L152)?

alvaroaleman: Ah, that second log was plain wrong, seems I accidentally put it there.

bbguimaraes: Ok, now it makes sense =)

PR #835, opened 2020-05-22
ci-operator: allow requesting resolution of raw config

stevekuznetsov:
ci-operator-configresolver: add an endpoint to resolve raw config

A user may now provide the config resolver with a raw configuration that
they wish to resolve. This allows users to provide configurations for
resolution that may not be present in the organizational tree or are not
identifiable by a (org, repo, branch, variant) tuple.

---

ci-operator: allow requesting resolution of raw config

The CI Opertor now allows for a user to provide a raw, unresolved
configuration for resolution with the registry via --unresolved-config
or $UNRESOLVED_CONFIG. This acts like --config and $CONFIG_SPEC, except
that the user is expected to provide an unresolved configuration, not a
full one.

---


bbguimaraes:
Implementation LGTM. Can we look at the configuration that was passed in and determine if it needs to be resolved instead of adding a separate flag/variable? Or, alternatively, remove the default value for `--resolver-address` and only resolve when either that or `--registry` are set?

stevekuznetsov:
@bbguimaraes I think the simple, explicit approach is sufficient -- what user needs the "resolve sometimes if needed" behavior?

PR #842, opened 2020-05-25
Promotionreconciler: Stop retrying on requests that get a 404 from GH

alvaroaleman:
The 404 means that either:
* The repo/branch was deleted, so this request can not be reconciled
* We do not have access to this repo, which also means we can not
  reconcile it

The PR also removes the `ignoredGitHubOrganizations`, we instead try
everything and just stop retrying on the 404.

Fixes https://github.com/openshift/release/issues/9283

### DIFF
--- a/cmd/dptp-controller-manager/main.go
+++ b/cmd/dptp-controller-manager/main.go
@@ -27,12 +27,7 @@ type options struct {
 	jobConfigPath                 string
 	registryClusterKubeconfigPath string
 	dryRun                        bool
-	promotionreconcilerOptions    promotionreconcilerOptions
 	*flagutil.GitHubOptions
-}
-
-type promotionreconcilerOptions struct {
-	IgnoredGitHubOrganizations flagutil.Strings
 }
 
 func newOpts() (*options, error) {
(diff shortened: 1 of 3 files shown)


cmd/dptp-controller-manager/main.go
```diff
@@ -43,7 +38,7 @@ func newOpts() (*options, error) {
 	flag.StringVar(&opts.configPath, "config-path", "", "Path to the prow config")
 	flag.StringVar(&opts.jobConfigPath, "job-config-path", "", "Path to the job config")
 	flag.StringVar(&opts.registryClusterKubeconfigPath, "registry-cluster-kubeconfig", "", "If set, this kubeconfig will be used to access registry-related resources like Images and ImageStreams. Defaults to the global kubeconfig")
-	flag.Var(&opts.promotionreconcilerOptions.IgnoredGitHubOrganizations, "promotionreconcilerOptions.ignored-github-organization", "GitHub organization to ignore in the imagestreamtagreconciler. Can be specified multiple times")
+	_ = flag.String("promotionreconcilerOptions.ignored-github-organization", "", "deprecated, doesn't do anything. Was used to ignore github organization. We instead attempt all repos and swallow 404 errors we get from github doing so.")
```
droslean, line 41: Let's just remove the whole thing.

alvaroaleman: That will send our deployment into crashlooping: https://github.com/openshift/release/blob/9e4b6fbfe9bc969a66e0b8b5a5a9844900021b44/clusters/app.ci/assets/dptp-controller-manager.yaml#L145

droslean: Can you open a follow-up PR in release repo to remove that as well?

alvaroaleman: Yup, that as the plan

PR #845, opened 2020-05-27
Add openshift-tests run-resourcewatch to ugprades

damemi:
Copied from https://github.com/openshift/release/pull/9198, we want to run this in 4.x upgrade tests


deads2k:
/lgtm

@damemi someone in dptp can probably approve.  As I understand it from https://coreos.slack.com/archives/CBN38N3MW/p1590000428406000?thread_ts=1589915598.371500&cid=CBN38N3MW we cannot rehearse

PR #850, opened 2020-05-27
webreg: allow chains with the same name as a reference be displayed

AlexNPavel:
This PR fixes an issue where chains with the same name as a reference would not appear in the web UI since the web UI only used names as an identifier.


AlexNPavel:
Example of running fix:
Chain: https://server-ci-tools-850.apps.ci.l2s4.p1.openshiftapps.com/chain/ipi-conf-aws
Reference: https://server-ci-tools-850.apps.ci.l2s4.p1.openshiftapps.com/reference/ipi-conf-aws

Current production server, which only has the reference and no way to view the chain: https://ci-operator-configresolver-ui-ci.apps.ci.l2s4.p1.openshiftapps.com/registry/ipi-conf-aws

PR #854, opened 2020-05-28
ci-operator multi-stage: add step parameters

bbguimaraes:
Implementation of the proposal described in:

https://docs.google.com/document/d/14CLXQ4NgHNZHw_FGfmNsqxC1FGgK5qVb6uRc9XIafws

TODO:

- [x] execution
- [x] validation
- [x] registry
- [x] documentation


pkg/api/types.go
```diff
@@ -608,6 +620,8 @@ type MultiStageTestConfiguration struct {
 	// Workflow is the name of the workflow to be used for this configuration. For fields defined in both
 	// the config and the workflow, the fields from the config will override what is set in Workflow.
 	Workflow *string `json:"workflow,omitempty"`
+	// Environment has the values of parameters for the steps.
+	Environment TestEnvironment `json:"env,omitempty"`
```
stevekuznetsov, line 624: Why not call these parameters?

pkg/api/types.go
```diff
@@ -623,8 +637,13 @@ type MultiStageTestConfigurationLiteral struct {
 	// Post is the array of test steps run after the tests finish and teardown/deprovision resources.
 	// Post steps always run, even if previous steps fail.
 	Post []LiteralTestStep `json:"post,omitempty"`
+	// Environment has the values of parameters for the steps.
+	Environment TestEnvironment `json:"env,omitempty"`
 }
 
+// TestEnvironment has the values of parameters for multi-stage tests.
+type TestEnvironment map[string]string
```
stevekuznetsov, line 645: How is this different than the current parameter interface?

pkg/api/types.go
bbguimaraes: Parameters, schmarameters.

I see "parameters" as the abstract concept and "environment" as the implementation. Doesn't make a big difference to me other than "environment" being the name used in the design document.

pkg/api/types.go
bbguimaraes: Hmm, do you mean we should use `pkg/api/parameters.go` to implement this (possibly chaining the existing `params`… parameter)? That interface is intended for dynamic values (and so is more complex), while these parameters are all static.

I don't really see an advantage in extending the parameters passed in. We cannot allow access to the extraneous parameters unless we relax the validation. The current (not present here) validation happens at configuration-resolution time, where we don't yet have information about the dynamic parameters.

bbguimaraes:
Ready for review (note: includes https://github.com/openshift/ci-tools/pull/879, first relevant commit is bb96da8).

pkg/api/config.go
```diff
@@ -502,6 +505,22 @@ func validateCredentials(fieldRoot string, credentials []CredentialReference) []
...
+	for _, param := range params {
+		if param.Default != "" {
+			continue
+		}
+		if _, ok := env[param.Name]; !ok {
+			missing = append(missing, param.Name)
+		}
+	}
+	if missing != nil {
+		return fmt.Errorf("%s: unresolved parameter(s): %s", fieldRoot, missing)
```
petr-muller, line 519: This is user-facing, so I think this error should be much more explanatory and lead the user to the name of the step/workflow/chain that needs this parameter.

pkg/registry/resolver.go
```diff
@@ -69,19 +76,41 @@ func (r *registry) Resolve(name string, config api.MultiStageTestConfiguration)
 	post, errs := r.process(config.Post, sets.NewString(), stack)
 	expandedFlow.Post = append(expandedFlow.Post, post...)
 	resolveErrors = append(resolveErrors, errs...)
-
+	for u := range rec.unused {
+		resolveErrors = append(resolveErrors, stackErrorf(stack, "no step declares parameter %q", u))
```
petr-muller, line 80: So it is an error if I specify a parameter which is not explicitly consumed by any step? That sounds excessive... It also prevents us from e.g. adding a parameter to a step without breaking existing users...

pkg/api/config.go
bbguimaraes: https://github.com/openshift/ci-tools/pull/879 added contextual information to the errors (see `"unresolved test"` in the tests).

A more concrete example:

```
error: failed to load configuration: failed to resolve configuration: Failed resolve MultiStageTestConfiguration: e2e: ipi-aws-pre: ipi-install: unresolved parameter: FOO
```

pkg/registry/resolver.go
bbguimaraes: Don't we *want* to prevent those? The PR adding the parameter will fail validation if the tests are also not updated.

pkg/api/config.go
bbguimaraes: Oh, that's in `Validate`, not (the same error) in the resolver…

The only cases this covers is inline steps, in which case the problem should be pretty obvious.

pkg/registry/resolver.go
petr-muller: I was worried about the `redhat-operator-ecosystem/release` content, but I guess you are right. And it has an advantage of actively stopping the possible confusion that items in `environment` are just propagated everywhere, even if the step/chain/workflow does not declare to have that parameter, which is a good thing.

pkg/api/config.go
petr-muller: Yeah, I realized that after reading the rest of the PR (and of course only after submitting the review)...

bbguimaraes:
https://server-ci-tools-754.apps.ci.l2s4.p1.openshiftapps.com/

bbguimaraes:
Huh…

```
dial tcp: lookup image-registry.openshift-image-registry.svc on 172.30.0.10:53: no such host
```

/retest

bbguimaraes:
Cache… Haven't seen that error anywhere else.

/retest

PR #858, opened 2020-05-28
ci-operator: multi-stage: require release images only without payload

stevekuznetsov:
When we require a full release payload, we are guaranteed to get the
release images, as well. We do not need to require both in this case.


stevekuznetsov:
```
 $ git merge --no-ff bbed36082192ba2cad8516e7cb730ed595afdecc
merge: bbed36082192ba2cad8516e7cb730ed595afdecc - not something we can merge
# Error: exit status 1 
```

???

PR #865, opened 2020-05-28
webreg: support variant jobs in the job list and search

AlexNPavel:
This PR adds support for displaying variant tests in the job list and search pages separately from the standard tests and providing proper hyperlinks to the the individual job pages for variant jobs.

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -346,18 +346,21 @@ const templateDefinitions = `
 						<td rowspan="{{ (repoSpan $repo $containsVariant) }}" style="vertical-align: middle;">{{ $repo.Name }}</td>
 					</tr>
 					{{ range $index, $branch := $repo.Branches }}
-						<tr>
-							{{ $branchLen := len $branch.Variants }}
-							{{ if $containsVariant }}
-								<td rowspan="{{ doubleInc $branchLen }}" style="vertical-align: middle;">{{ $branch.Name }}</td>
-							{{ else }}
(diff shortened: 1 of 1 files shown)


AlexNPavel:
Running example of bug and fix:

Current job page for the `openshift-release-master-{ocp-4.4,ocp-4.5}-e2e-metal-ipi-ipv4` tests: https://ci-operator-configresolver-ui-ci.apps.ci.l2s4.p1.openshiftapps.com/search?job=ipv4

Fixed version: https://server-ci-tools-865.apps.ci.l2s4.p1.openshiftapps.com/search?job=ipv4

stevekuznetsov:
Why not another column>

AlexNPavel:
> Why not another column?

I don't feel like it makes sense to add a column for variant when it isn't commonly used. It would just be an empty space for the vast majority of jobs. Maybe separating variants as separate rows would be clearer though?

stevekuznetsov:
Hm. Conceptually it should be another column, though, since it's just another identifier that can be paired with the previous three. I agree in the non-filtered overview, it would add some white-space, but as soon as someone actually searches for something we could add it or not depending on whether we need it for rendering, and it should be a non-issue?

AlexNPavel:
> Hm. Conceptually it should be another column, though, since it's just another identifier that can be paired with the previous three. I agree in the non-filtered overview, it would add some white-space, but as soon as someone actually searches for something we could add it or not depending on whether we need it for rendering, and it should be a non-issue?

It's possible to do that, but adds a lot of complexity. I'll work on adding support and see how that goes

AlexNPavel:
Updated to have separate column for `Variant`. Extra `Variant` column appears only when at least 1 job in the list contains a variant.

PR #867, opened 2020-05-29
webreg: add example of overriding workflow with literal steps

petr-muller:
I think it is useful to have an explicit example for people who want to migrate from `openshift_installer_src`-like templates.

/cc @bbguimaraes @stevekuznetsov

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -1100,10 +1100,6 @@ multistage test design:
 
 {{ yamlSyntax (index . "configExample1") }}
 
-Example of a <code>ci-operator</code> configuration that overrides a workflow field.
-
-{{ yamlSyntax (index . "configExample2") }}
-
 <p>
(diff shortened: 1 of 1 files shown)


petr-muller:
https://server-ci-tools-867.apps.ci.l2s4.p1.openshiftapps.com/help#config

pkg/webreg/webreg.go
```diff
@@ -1104,6 +1104,11 @@ Example of a <code>ci-operator</code> configuration that overrides a workflow fi
 
 {{ yamlSyntax (index . "configExample2") }}
 
+The configuration can also override a workflow field with a <a href="#step">full literal step</a>
+(not only a reference to a shared step):
+
+{{ yamlSyntax (index . "configExample3") }}
```
stevekuznetsov, line 1110: Can you move this and the example 2 to be beneath "’s field has priority (i.e. the value from the ci-operator configuration is used). "?

petr-muller: Sure, makes sense.

PR #869, opened 2020-05-29
ci-operator: split the import and creation of releases

stevekuznetsov:
The creation of a new release payload from local image sources in an
image stream and the import of an external release are simply two
separate actions, which are executed in entirely unrelated ways. This
commit breaks the previous step that conditionally did both into two,
and chooses which to add to the graph as appropriate. This should be a
no-op for users.

### DIFF
--- a/pkg/steps/release/import_release.go
+++ b/pkg/steps/release/import_release.go
@@ -1,0 +1,401 @@
+package release
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
+	coreapi "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


pkg/steps/release/import_release.go
```diff
@@ -0,0 +1,401 @@
...
+	return results.ForReason("importing_release").ForError(s.run(ctx, dry))
+}
+
+func (s *importReleaseStep) run(ctx context.Context, dry bool) error {
+	_, err := setupReleaseImageStream(s.jobSpec.Namespace, s.saGetter, s.rbacClient, s.imageClient, s.dryLogger, dry)
+	if err != nil {
+		return err
+	}
+
+	streamName := streamNameFor(s.name)
```
petr-muller, line 78: Move this under the dry check for readability.

PR #871, opened 2020-05-30
Implement Prow Jobs dispatcher

hongkailiu:
Implement 

- [X] Query and store the prow job volumes
- [ ] Check if the types of e2e tests
- [ ] Dispatch the prow jobs and save it in the sanitize tool's config.

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -1,0 +1,94 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	prometheusapi "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/logrusutil"
+
+	"github.com/openshift/ci-tools/pkg/dispatcher"
+)
+
+type options struct {
+	prometheusURL          string
+	prometheusUsername     string
+	prometheusPasswordPath string
+	jobVolumesPath         string
+
+	prometheusPassword string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.prometheusURL, "prometheus-url", "https://prometheus-prow-monitoring.apps.ci.l2s4.p1.openshiftapps.com", "The prometheus URL")
+	fs.StringVar(&o.prometheusUsername, "prometheus-username", "", "The Prometheus username.")
+	fs.StringVar(&o.prometheusPasswordPath, "prometheus-password-path", "", "The path to a file containing the Prometheus password")
+	fs.StringVar(&o.jobVolumesPath, "job-volumes-path", filepath.Join(os.TempDir(), "job-volumes.yaml"), "The path to a file containing the job volumes")
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatal("could not parse input")
+	}
+	return o
+}
+
+func (o *options) validate() error {
+	if (o.prometheusUsername == "") != (o.prometheusPasswordPath == "") {
+		return fmt.Errorf("--prometheus-username and --prometheus-password-path must be specified together")
+	}
+	return nil
+}
+
+func (o *options) complete(secrets *sets.String) error {
+	if o.prometheusPasswordPath != "" {
+		bytes, err := ioutil.ReadFile(o.prometheusPasswordPath)
+		if err != nil {
+			return err
+		}
+		o.prometheusPassword = strings.TrimSpace(string(bytes))
+		if o.prometheusPassword == "" {
+			return fmt.Errorf("no content in file: %s", o.prometheusPasswordPath)
+		}
+		secrets.Insert(o.prometheusPassword)
+	}
+	return nil
+}
+
+func main() {
+	secrets := sets.NewString()
+	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, func() sets.String {
+		return secrets
+	}))
+	o := gatherOptions()
+	if err := o.validate(); err != nil {
+		logrus.WithError(err).Fatal("Failed to complete options.")
+	}
+	if err := o.complete(&secrets); err != nil {
+		logrus.WithError(err).Fatal("Failed to complete options.")
+	}
+
+	promClient, err := dispatcher.NewPrometheusClient(o.prometheusURL, o.prometheusUsername, o.prometheusPassword)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to create promethues client.")
+	}
+
+	v1api := prometheusapi.NewAPI(promClient)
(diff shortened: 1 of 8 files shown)


cmd/sanitize-prow-jobs/main.go
```diff
@@ -1,27 +1,43 @@
...
+	"github.com/openshift/ci-tools/pkg/dispatcher"
 )
 
 type options struct {
 	prowJobConfigDir string
 	configPath       string
 
 	help bool
+
+	prometheusURL          string
```
stevekuznetsov, line 34: IMO this would be better off in a standalone (simple) tool instead of adding to a mega-tool

pkg/dispatcher/prometheus_test.go
```diff
@@ -0,0 +1,197 @@
...
+			equalError(t, tc.expectedError, err)
+		})
+	}
+}
+
+func equalError(t *testing.T, expected, actual error) {
+	if (expected == nil) != (actual == nil) {
+		t.Errorf("%s: expecting error \"%v\", got \"%v\"", t.Name(), expected, actual)
+	}
+	if expected != nil && actual != nil && expected.Error() != actual.Error() {
```
stevekuznetsov, line 194: Why is `reflect.DeepEqual()` not sufficient here?

cmd/sanitize-prow-jobs/main.go
hongkailiu: Ha. I actually started with a new tool, called `prow-job-dispatcher` but switched to this.
Maybe this mega-tool only do prometheus query? Nothing about prow?
Let me give it a try.

pkg/dispatcher/prometheus_test.go
hongkailiu: Updated.

hongkailiu: Changed it back.
Need nil check or we need to return in the previous case.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Updated with a new tool.
Where should I put the logic for using job-volume-map?

We could use this tool `dispatch-prow-jobs` to generate the config for `sanitize-prow-jobs`. What you think?

stevekuznetsov: Oh, I see, since `sanitize` now sets the cluster ... yeah a two-step process could make sense. I don't think we should be calling a Prometheus API during `make jobs` for instance

hongkailiu:
I am working on the part of consuming the job volume map.
But that will be another PR.
Since I do not touch the existing sanitize tool, it should not break anything after merge.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Result of slack discussion:
Do the Prom. query and use the answer in the new tool. The new tool will use the sanitize tool’s config too because it needs to know the jobs which have to be on some cluster. And then the new tool will dispatch others jobs and save it to sanitize’s config.

cmd/prow-job-dispatcher/main.go
```diff
@@ -0,0 +1,108 @@
...
+	if err := o.complete(&secrets); err != nil {
+		logrus.WithError(err).Fatal("Failed to complete options.")
+	}
+
+	promClient, err := dispatcher.NewPrometheusClient(o.prometheusURL, o.prometheusUsername, o.prometheusPassword)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to create promethues client.")
+	}
+
+	v1api := v1.NewAPI(promClient)
```
alvaroaleman, line 86: please always alias ambigious imports like `v1`

pkg/dispatcher/prometheus_test.go
alvaroaleman: reflect.Equal does not consider two nil errors to be equal?

cmd/prow-job-dispatcher/main.go
```diff
@@ -0,0 +1,108 @@
...
+	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+	defer cancel()
+
+	var jobVolumes map[string]float64
+	if o.updateJobVolumes {
+		jobVolumes, err = dispatcher.GetJobVolumesFromPrometheus(ctx, v1api)
+		if err != nil {
+			logrus.WithError(err).Fatal("Failed to get job volumes from Prometheus.")
+		}
+		if err = dispatcher.SaveJobVolumes(jobVolumes, o.jobVolumesPath); err != nil {
```
alvaroaleman, line 96: How will this file be used? Wouldn't it make sense to make it update the `/sanitize-prow-jobs/_config.yaml` config rather than introducing yet another config file?

pkg/dispatcher/prometheus.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if len(warnings) > 0 {
+		logrus.WithField("Warnings", warnings).Warn("Got warnings from Prometheus")
+	}
+
+	vector, ok := result.(model.Vector)
+	if !ok {
+		return nil, fmt.Errorf("returned result from Prometheus cannot be cast to vector")
```
alvaroaleman, line 80: please add a type formatter for `result` (`%T`), if we ever hit this someone will be really annoyed at the useless error

pkg/dispatcher/prometheus_test.go
```diff
@@ -0,0 +1,205 @@
...
+	v1 "github.com/prometheus/client_golang/api/prometheus/v1"
+	"github.com/prometheus/common/model"
+)
+
+type prometheusAPIForTest struct {
+	queryFunc func(ctx context.Context, query string, ts time.Time) (model.Value, v1.Warnings, error)
+}
+
+func (prometheusAPI *prometheusAPIForTest) Query(ctx context.Context, query string, ts time.Time) (model.Value, v1.Warnings, error) {
+	return prometheusAPI.queryFunc(ctx, query, ts)
```
alvaroaleman, line 23: Can we make this validate the `query string` parameter?

cmd/prow-job-dispatcher/main.go
hongkailiu: Save the query result to avoid query each time when running this tool.
We can then control the frequency of fresh the volume map.
And we do not need a working prometheus when running this tool.

alvaroaleman: Yeah but whats the point of storing this intermediate state? Just update the `_config` file of the `sanitize-command` with the results or not?

cmd/prow-job-dispatcher/main.go
hongkailiu: Done. It was the default choice of `vscode`.

pkg/dispatcher/prometheus.go
hongkailiu: Done.

hongkailiu: Done.

pkg/dispatcher/prometheus_test.go
hongkailiu: Done.

cmd/prow-job-dispatcher/main.go
hongkailiu: Save the result for the next time to reuse in case that we cannot query prometheus the next time, like prometheus is down, we lose the data of prometheus.
Some old results is better than nothing?

hongkailiu: Removed.

PR #872, opened 2020-06-01
init implementation of the publicize plugin

droslean:
Publicize plugin:

This external prow plugin read a configuration file to determine the linking repositories. 
When a `/publicize` comment will be detected, the plugin will perform a merge between the two repositories.   


droslean:
WIP for tests and for a few minor changes.

cmd/publicize/main.go
```diff
@@ -0,0 +1,174 @@
...
+
+func main() {
+	o := gatherOptions()
+	if err := o.Validate(); err != nil {
+		logrus.Fatalf("Invalid options: %v", err)
+	}
+
+	logrus.SetLevel(logrus.DebugLevel)
+
+	log := logrus.StandardLogger().WithField("plugin", "publicize")
```
stevekuznetsov, line 138: You should be able to use the `ComponentInit()`

cmd/publicize/main.go
```diff
@@ -0,0 +1,174 @@
...
+
+	secretAgent := &secret.Agent{}
+	if err := secretAgent.Start([]string{o.github.TokenPath, o.webhookSecretFile}); err != nil {
+		logrus.WithError(err).Fatal("Error starting secrets agent.")
+	}
+
+	webhookTokenGenerator := secretAgent.GetTokenGenerator(o.webhookSecretFile)
```
stevekuznetsov, line 145: How much of this is shared with other plugins? Can we factor it out into a library in `test-infra`?

cmd/publicize/main.go
```diff
@@ -0,0 +1,174 @@
...
+		gitEmail:              o.gitEmail,
+		ghc:                   githubClient,
+		log:                   log,
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/", serv)
+	externalplugins.ServeExternalPluginHelp(mux, log, helpProvider)
+	httpServer := &http.Server{Addr: ":" + strconv.Itoa(o.port), Handler: mux}
+	defer interrupts.WaitForGracefulShutdown()
```
stevekuznetsov, line 172: Why defer this? Just make it the last line and run it normally

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+	confirm bool
+
+	gitName  string
+	gitEmail string
+
+	ghc github.Client
+	log *logrus.Entry
+}
+
+func (s *server) ServeHTTP(w http.ResponseWriter, r *http.Request) {
```
stevekuznetsov, line 51: All of this should be in a library upstream IMO

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+			}
+		}()
+	default:
+		logrus.Debugf("skipping event of type %q", eventType)
+	}
+	return nil
+}
+
+func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent) error {
+	if !ic.Issue.IsPullRequest() || ic.Action != github.IssueCommentActionCreated {
```
stevekuznetsov, line 84: This should realistically be a generic comment handler and handle all the other cases -- people will want to be able to use a review, etc, like they can with other commands for Prow

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+	num := ic.Issue.Number
+	commentAuthor := ic.Comment.User.Login
+
+	l = l.WithFields(logrus.Fields{
+		github.OrgLogField:  org,
+		github.RepoLogField: repo,
+		github.PrLogField:   num,
+	})
+
+	logger := s.log.WithFields(l.Data)
```
stevekuznetsov, line 103: Why do we have two loggers?

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// Only org members should be able to publicize a pull request.
+	ok, err := s.ghc.IsMember(org, commentAuthor)
```
stevekuznetsov, line 106: We have more complicated checks for other commands like `/lgtm` ... but in general why do we even want this? This comment is going to be in a private repository in a private org. By virtue of someone having access to the comment section, they have privilege to run it

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+	logger.Info("Publicize of PR has been requested.")
+
+	pr, err := s.ghc.GetPullRequest(org, repo, num)
+	if err != nil {
+		return err
+	}
+	baseBranch := pr.Base.Ref
+
+	var prCommits []string
+	commits, err := s.ghc.ListPRCommits(org, repo, num)
```
stevekuznetsov, line 125: Use the git client for git data.

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+	}
+
+	// publicize only merged PRs.
+	if !pr.Merged {
+		resp := "cannot publicize an unmerged PR"
+		logger.Info(resp)
+		return s.ghc.CreateComment(org, repo, num, plugins.FormatICResponse(ic.Comment, resp))
+	}
+
+	logger.Infof("searching bond for %s/%s", org, repo)
```
stevekuznetsov, line 141: Org and repo are already in your logger's fields

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+	bondedOrgRepo, ok := s.bonds[fmt.Sprintf("%s/%s", org, repo)]
+	if !ok {
+		logger.Warn("There is no bond for the current repository.")
```
stevekuznetsov, line 144: `s/bond/public repository configured/`

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+		resp := "cannot publicize because there is no bond for the current pull request"
+		return s.ghc.CreateComment(org, repo, num, plugins.FormatICResponse(ic.Comment, resp))
+	}
+
+	logger.Infof("Trying to merge the PR to destination: %s/%s@%s", org, repo, baseBranch)
+	if err := s.mergeAndPushToRemote(logger, bondedOrgRepo, num, org, repo, baseBranch, s.confirm, prCommits); err != nil {
+		return fmt.Errorf("couldn't merge the pull request and push to the destination: %v", err)
+	}
+
+	return s.ghc.CreateComment(org, repo, num, plugins.FormatICResponse(ic.Comment, "Publicize suceeded."))
```
stevekuznetsov, line 154: This comment could be clearer like

```
Succeeded in pushing these commits to the public repository: <link>
```

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+}
+
+func (s *server) mergeAndPushToRemote(logger *logrus.Entry, bondedOrgRepo string, prNumber int, org, repo, branch string, confirm bool, commits []string) error {
+	// this is already validated
```
stevekuznetsov, line 158: Then pass those as separate strings

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+	bondedOrg := strings.Split(bondedOrgRepo, "/")[0]
+	bondedRepo := strings.Split(bondedOrgRepo, "/")[1]
+
+	bondURIRaw := fmt.Sprintf("https://github.com/%s/%s", bondedOrg, bondedRepo)
+	bondURI, err := url.Parse(bondURIRaw)
```
stevekuznetsov, line 163: This belongs in a GitHub client library

cmd/publicize/server.go
```diff
@@ -0,0 +1,237 @@
...
+		return fmt.Errorf("Failed to create temporary directory for git operations: %v", err)
+	}
+
+	defer func() {
+		if err := os.RemoveAll(gitDir); err != nil {
+			logger.WithError(err).Error("Failed to clean up temporary directory")
+		}
+	}()
+
+	if err := gitExec(logger, gitDir, []string{"init"}...); err != nil {
```
stevekuznetsov, line 189: This belongs in a `git` client library?

cmd/publicize/server.go
droslean: I added this to avoid external users to shoot their legs. I can remove this, and make a check if the current repo is private or not. WDYT?

cmd/publicize/server.go
droslean: Yea, it seems that this func is duplicated in all the plugins... I will check and open a PR to share it.

cmd/publicize/server.go
droslean: I can't use prow's git library for this.  Shall we extend it?

cmd/publicize/server.go
droslean: I just want to construct the uri with the user in order to use it in the git commands.

cmd/publicize/server.go
stevekuznetsov: Yes

stevekuznetsov:
@droslean LMK when the upstream PR(s) to create the external plugin library and extend the `git` client are ready

cmd/publicize/README.md
```diff
@@ -0,0 +1,17 @@
+# Publicize
+
+This tool is an external prow plugin. Its purpose is to merge the commit histories of two repositories.
+It responses in a `/publicize` command only in merged pull requests.
```
petr-muller, line 4: "responds to"

petr-muller: I would really like to see a better, more detailed documentation about what the tool is really doing. Can we get a simple workflow like:

1. User enters `/publicize` in a PR that is already merged
2. The plugin fetches the current remote of (which branch?)
3. The plugin merges what into what
4. The plugin files a PR / directly pushes (what hapepns when a pushj fails?)

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+	"k8s.io/test-infra/prow/logrusutil"
+
+	"k8s.io/test-infra/pkg/flagutil"
+	"k8s.io/test-infra/prow/config/secret"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+)
+
+type Config struct {
+	Bonds []Bond `json:"bonds,omitempty"`
+}
```
petr-muller, line 29: Is there a reason why this is struct rather than `type Config Bonds[]Bond`. Extensibility?

Does it make sense to have empty slice?

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+		if err := bond.validate(); err != nil {
+			errs = append(errs, fmt.Errorf("bond[%d]: %w", index, err))
+		}
+	}
+
+	return kerrors.NewAggregate(errs)
+}
+
+// bond holds the information of which repos are bonded to each other.
+type Bond struct {
```
petr-muller, line 43: Why is this a `From/To` struct and not a map? We should get a lot of validation for free.

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+	Bonds []Bond `json:"bonds,omitempty"`
+}
+
+func (c *Config) validate() error {
+	var errs []error
+	for index, bond := range c.Bonds {
+		if err := bond.validate(); err != nil {
+			errs = append(errs, fmt.Errorf("bond[%d]: %w", index, err))
+		}
+	}
```
petr-muller, line 37: Would it make sense to validate for conficts (one repo appearing in multiple `from`s or multiple `to`s. Would that be a supported case?

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+
+	webhookSecretFile string
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Dry run for testing. Uses API tokens but does not mutate.")
+
+	fs.StringVar(&o.configPath, "config-path", "/etc/config/config.yaml", "Path to publicize configuration.")
```
petr-muller, line 75: I'm a little confused about the usage here. This wouldn't read normal Prow `plugins.yaml`?

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	bytes, err := ioutil.ReadFile(o.configPath)
+	if err != nil {
+		return fmt.Errorf("Couldn't read publicize configuration file: %v", o.configPath)
+	}
+
+	var c Config
+	if err := yaml.Unmarshal(bytes, &c); err != nil {
+		return errors.New("Couldn't unmarshal publicize configuration")
```
petr-muller, line 109: Why eat the error and not returning the actual one (which may contain somehting useful about hwy the unmarshal failed?

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+	}
+
+	if err := c.validate(); err != nil {
+		return err
+	}
+
+	o.bonds = make(map[string]string)
```
petr-muller, line 116: I have a little problem with the term `bonds` correctly describing the actual relation, but I have trouble finding a better one... The relation is directed (one way), so something like "publishing paths' sound better to me... 

@stevekuznetsov native speaker to the rescue?

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+	for _, bond := range c.Bonds {
+		o.bonds[bond.From] = bond.To
+	}
```
petr-muller, line 119: Uh, so this ends up being a map, which means the "repo publishes to multiple destinations" or "multiple repos publish to a single destination" scenario I asked above is not supported. It reiterates the  question of "why isn't this a map in the config right away"?

cmd/publicize/main.go
```diff
@@ -0,0 +1,189 @@
...
+	return nil
+}
+
+func main() {
+	o := gatherOptions()
+	if err := o.Validate(); err != nil {
+		logrus.Fatalf("Invalid options: %v", err)
+	}
+
+	logrus.SetLevel(logrus.DebugLevel)
```
petr-muller, line 134: Don't think we want this in production

cmd/publicize/server.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	bondedOrg := strings.Split(bondedOrgRepo, "/")[0]
+	bondedRepo := strings.Split(bondedOrgRepo, "/")[1]
+
+	logger.Infof("Trying to merge the PR to destination: %s/%s@%s", bondedOrg, bondedRepo, baseBranch)
+	if err := s.mergeAndPushToRemote(bondedOrg, bondedRepo, org, repo, baseBranch); err != nil {
+		logger.WithError(err).Warnf("couldn't merge the pull request and push to the destination: %v", err)
+		s.createComment(ic, fmt.Sprintf("Publicize failed with error: %v", err), logger)
```
petr-muller, line 101: Are we sure that the `err` cannot leak a token possibly embedded in the URL? Are we sure that the error is censored?

cmd/publicize/server.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	bondedOrgRepoLink := fmt.Sprintf("https://github.com/%s", bondedOrgRepo)
+	s.createComment(ic, fmt.Sprintf("Publicize to the repository: [%s/%s](%s) was successful",
+		bondedOrg, bondedRepo, bondedOrgRepoLink), logger)
+}
+
+func (s *server) mergeAndPushToRemote(bondedOrg, bondedRepo string, org, repo, branch string) error {
```
petr-muller, line 110: At this place I really want to see the actual merge direction with words like `from/to` or `source/destination`. I have no idea whether the `bonded` one is the source or the destination :)

cmd/publicize/server.go
```diff
@@ -0,0 +1,154 @@
...
+	defer func() {
+		if err := s.gc.Clean(); err != nil {
+			logrus.WithError(err).Error("Could not clean up git client cache.")
```
petr-muller, line 113: Ok, so dptp-triage would be notified about this. To me this looks like a warning at worst.

petr-muller: Also, is cleaning up the whole client correct? What if we are publicizing multiple PRs simultaneously (I guess rare but possible).

cmd/publicize/server.go
```diff
@@ -0,0 +1,154 @@
...
+		}
+	}()
+
+	repoClient, err := s.gc.ClientFor(bondedOrg, bondedRepo)
```
petr-muller, line 117: What if we are publicizing multiple PRs for a single repo? Won't they race over a single shared git clone?

petr-muller: Note that this is probably not that uncommon - I can imagine a secuity fix being fixed in multiple PRs. For large repos the publicize can take some time (because for merge you need full fetch), and I can imagine the responsible person will just go to all PRs and `/publicize` in all of them.

cmd/publicize/server.go
```diff
@@ -0,0 +1,154 @@
...
+	if err := repoClient.Config("user.name", s.gitName); err != nil {
+		return fmt.Errorf("couldn't set config user.name=%s: %w", s.gitName, err)
+	}
+
+	if err := repoClient.Config("user.email", s.gitEmail); err != nil {
+		return fmt.Errorf("couldn't set config user.name=%s: %w", s.gitEmail, err)
+	}
+
+	mergeMsg := fmt.Sprintf("merge history from %s/%s", bondedOrg, bondedRepo)
+	if merged, err := repoClient.MergeWithStrategy("FETCH_HEAD", "merge", git.MergeOpt{CommitMessage: mergeMsg}); err != nil && !merged {
```
petr-muller, line 139: I don't think this is correct: at this point you basically did:

```
$ git init (and whatever ClientFor is doing internally)
$ git fetch remote branch
$ git merge (...options...) FETCH_HEAD
```

Which means you will always merge to a default branch (`master`), not to the desired one.

cmd/publicize/main.go
droslean: This is needed to allow Config to filled from a YAML file.

cmd/publicize/main.go
droslean: To make the configuration file more readable and more understandable. Otherwise, it would look like

```yaml
bonds:
-  openshift-priv/origin:
     openshift/origin
....
...
```

cmd/publicize/main.go
droslean: Yea, that sounds good to me. I will add that validation as well.

cmd/publicize/main.go
droslean: This will ready the publicize configuration file. Not related to prow.

cmd/publicize/main.go
droslean: Maybe `Pairs` ?

cmd/publicize/main.go
droslean: Answered in: https://github.com/openshift/ci-tools/pull/872#discussion_r454969576

cmd/publicize/server.go
droslean: I was having nightmares with this. I tried also to simulate some errors but I didn't see any leaked token. I will dig this further. Also it would be easier to debug this once I finish with the tests as well.

cmd/publicize/server.go
droslean: Yea that can race. We can just clean the cache when the plugin terminates instead, and just clean the cloned folder here.

cmd/publicize/server.go
droslean: I don't think that this will race. I think the multiple goroutines will clone the same repo but in the different tmp folder. I will dig this to be sure.

cmd/publicize/server.go
droslean: Nice catch. Ok we need to checkout to the desired branched before merging, right?

cmd/publicize/server.go
droslean: It seems that the repos will be cloned in different folder. ref: https://github.com/kubernetes/test-infra/blob/master/prow/git/v2/client_factory.go#L245

cmd/publicize/main.go
petr-muller: Well it would be:

```
bonds:
- openshift-priv/origin: openshift/origin
```

But right, I guess that does not communicate the direction that well. /shrug

cmd/publicize/main.go
petr-muller: Well the description says it's an external Prow plugin. Other external prow plugins seem to have config in that file... But yeah, I guess I understand, this is a *real* external plugin, it does not make sense for the `test-infra` config struct to carry this config.

cmd/publicize/main.go
petr-muller: `pairs` has the same problem, doesn't communicate that the pair is directed...

cmd/publicize/server.go
petr-muller: Yes, and before that make sure that we fetch the corresponding remote branch, so that the checked out branch has a full history.

cmd/publicize/main.go
stevekuznetsov: A list of structs enables us to add in configuration options on a specific bond in the future if we need to, but yes this would be simpler as a mapping

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+	for index, bond := range c.Bonds {
+		if err := bond.validate(); err != nil {
+			errs = append(errs, fmt.Errorf("bond[%d]: %w", index, err))
+		}
+	}
+
+	return kerrors.NewAggregate(errs)
+}
+
+// bond holds the information of which repos are bonded to each other.
```
stevekuznetsov, line 43: You will need to fix up GoDoc

cmd/publicize/main.go
stevekuznetsov: This is outstanding, no?

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+	o.githubEventServerOptions.Bind(fs)
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatalf("cannot parse args: '%s'", os.Args[1:])
+	}
+	return o
+}
+
+func (o *options) Validate() error {
+	if (o.gitEmail == "") != (o.gitName == "") {
+		return errors.New("--git-name and --git-email must be specified")
```
stevekuznetsov, line 94: Why not default them?

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+
+	for _, group := range []flagutil.OptionGroup{&o.github} {
+		if err := group.Validate(o.dryRun); err != nil {
+			return err
+		}
+	}
+
+	bytes, err := ioutil.ReadFile(o.configPath)
+	if err != nil {
+		return fmt.Errorf("Couldn't read publicize configuration file: %v", o.configPath)
```
stevekuznetsov, line 105: Do we want dynamic reloading on file changes?

cmd/publicize/main.go
stevekuznetsov: I'm not convinced we need anything other than a map - do we have any thoughts for how we could need per-pair configuration?

cmd/publicize/main.go
stevekuznetsov: Aren't these just upstream repositories? We could name it that way.

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+		return sets.NewString([]string{string(webhookTokenGenerator()), string(githubTokenGenerator())}...)
+	}
+	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, getter))
+
+	githubClient, err := o.github.GitHubClient(secretAgent, o.dryRun)
+	if err != nil {
+		logrus.WithError(err).Fatal("Error getting GitHub client.")
+	}
+
+	censor := func(content []byte) []byte {
```
stevekuznetsov, line 152: Why is this not from the secret agent?

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+		return githubTokenGenerator()
+	}
+	gitClient, err := o.git.GitClient(githubClient, githubTokenGenerator, censor, o.dryRun)
+	if err != nil {
+		logrus.WithError(err).Fatal("Error getting Git client.")
+	}
+
+	logger := logrus.WithField("component", "publicize")
```
stevekuznetsov, line 160: There's a test-infra logrus util/init, let's use that

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+		bonds:                o.bonds,
+		ghc:                  githubClient,
+		gc:                   gitClient,
+		gitName:              o.gitName,
+		gitEmail:             o.gitEmail,
+		remoteHost:           fmt.Sprintf("https://%s", github.DefaultHost),
+		dry:                  o.dryRun,
+	}
+
+	o.githubEventServerOptions.HmacTokenGenerator = webhookTokenGenerator
```
stevekuznetsov, line 172: These three things seem like something you should get for free, no? What user of the github event server library wouldn't need them?

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+	o.githubEventServerOptions.Metrics = githubeventserver.NewMetrics()
+	o.githubEventServerOptions.Logger = logger
+
+	ges := githubeventserver.NewGitHubEventServer(o.githubEventServerOptions)
```
stevekuznetsov, line 176: Let's not use acronyms for names

cmd/publicize/main.go
```diff
@@ -0,0 +1,188 @@
...
+	ges.RegisterHandleIssueCommentEvent(serv.handleIssueComment)
+	ges.RegisterHelpProvider(helpProvider, logger)
+
+	interrupts.OnInterrupt(func() {
+		ges.GracefulShutdown()
```
stevekuznetsov, line 181: Why doesn't this just take a context? `OnInterrupt` is not how this is supposed to work. You want to `interrupts.Run()` and pass in the `ges.Listen()`, allow `Listen()` to shut down when it gets a cancelled context.

cmd/publicize/server.go
```diff
@@ -0,0 +1,188 @@
...
+	GetPullRequest(org, repo string, number int) (*github.PullRequest, error)
+
+	ListIssueComments(owner, repo string, number int) ([]github.IssueComment, error)
+}
+
+var publicizeRe = regexp.MustCompile(`(?mi)^/publicize\s*$`)
+
+func helpProvider(_ []prowconfig.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	pluginHelp := &pluginhelp.PluginHelp{
+		Description: `The publicize plugin is used for merging a pull request into the bonded repository.`,
```
stevekuznetsov, line 30: Bonded is not a word people use in this context. Let's just call it the upstream.

cmd/publicize/server.go
```diff
@@ -0,0 +1,188 @@
...
+
+	ghc githubClient
+	gc  git.ClientFactory
+
+	remoteHost string
+
+	dry bool
+}
+
+func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent) {
```
stevekuznetsov, line 57: The upstream expects to return an error. Why not this one?

cmd/publicize/server.go
```diff
@@ -0,0 +1,188 @@
...
+	destOrg := strings.Split(destOrgRepo, "/")[0]
+	destRepo := strings.Split(destOrgRepo, "/")[1]
+	sourceRemoteResolver := func() (string, error) {
+		u, err := url.Parse(s.remoteHost)
+		if err != nil {
+			return "", err
+		}
+
+		// If the host isn't a url but its an actual path, return the full path with org and repo included.
+		// This is usefull for testing as well.
```
stevekuznetsov, line 118: This smells. In what case does this branch run during production execution?

cmd/publicize/server.go
stevekuznetsov: Just run the secret agent's censor on the comment body before posting

cmd/publicize/server.go
```diff
@@ -0,0 +1,188 @@
...
+
+	logger.Infof("Trying to merge the PR to destination: %s/%s@%s", destOrg, destRepo, baseBranch)
+	mergeMsg := fmt.Sprintf("merge history from %s/%s", org, repo)
+	if err := s.mergeAndPushToRemote(destOrg, destRepo, sourceRemoteResolver, baseBranch, mergeMsg, s.dry); err != nil {
+		logger.WithError(err).Warnf("couldn't merge the pull request and push to the destination: %v", err)
+		s.createComment(ic, fmt.Sprintf("Publicize failed with error: %v", err), logger)
+		return
+	}
+
+	destOrgRepoLink := fmt.Sprintf("https://github.com/%s/tree/%s", destOrgRepo, baseBranch)
```
stevekuznetsov, line 138: You take a GitHub host as a flag option, but then hard-code this one.

cmd/publicize/server.go
```diff
@@ -0,0 +1,188 @@
...
+	s.createComment(ic, fmt.Sprintf("Publicize to [%s/%s@%s](%s) was successful",
```
stevekuznetsov, line 139: Maybe reword:

```
A pull request at [%s/%s#%d]() was created in the upstream repository to publish this work.
```

droslean: Well, actually it is not creating any pull request, but it direct merges it.

cmd/publicize/main.go
droslean: Yes. I wanted first to address the changes in the configuration file, then extend the validation code.

cmd/publicize/main.go
droslean: Absolutely. I was checking today the upstreams configAgent code to see if I can reuse it.

cmd/publicize/main.go
droslean: Secret agent has `func() []byte` but GitClient wants ` func(content []byte) []byte` /shrug

cmd/publicize/main.go
droslean: well, we need to pass the options t ocreate a github event server and the hmac, metric and logger are empty in the current state, and we rely on the user to pass them.

cmd/publicize/server.go
droslean: That's not true. see: https://github.com/kubernetes/test-infra/blob/master/prow/githubeventserver/githubeventserver.go#L110

cmd/publicize/server.go
droslean: This is a little bit hacky. I will open a follow-up upstream PR to make the `FetchFromRemote` more testable.  Currently, we need to pass a `remoteResolver` and to test this we need a resolver from an actual path.

cmd/publicize/server.go
droslean: I didn't know that this is possible. I'll check that. Thanks.

cmd/publicize/main.go
droslean: This is already handled here: https://github.com/kubernetes/test-infra/blob/82eb19f3809bf7b6343291d43ef23005c25edb98/prow/githubeventserver/githubeventserver.go#L96-L97

So handling the cleanup in `interrupts.OnInterrupt` should be sufficient here, right?

cmd/publicize/main.go
droslean: I am already using the censoring formater.

cmd/publicize/main.go
stevekuznetsov: Secret agent has 

```
prow/config/secret/agent.go:func (a *Agent) Censor(content []byte) []byte {
```

cmd/publicize/main.go
stevekuznetsov: Then you do not need this?

stevekuznetsov: I was expecting you to use ` logrusutil.ComponentInit()`

cmd/publicize/main.go
stevekuznetsov: Right, but if every single user of the github event server will always want these, the constructor should do it for them.

cmd/publicize/main.go
stevekuznetsov: No. Please write the github event server to run with a context.

cmd/publicize/server.go
stevekuznetsov: No, the upstream handlers all return errors. Please make the github event server  do it, too.

droslean: I designed the githubevent server based on `hook`. for example: https://github.com/kubernetes/test-infra/blob/master/prow/hook/events.go#L120

But it will make sense for the handlers to return an error. I will follow-up on this.

cmd/publicize/main.go
droslean: It seems that this overrides the censoring formater. Maybe I missed something here. I will check it.

cmd/publicize/main.go
droslean: Yea I had implemented it this way but it got lost in the reviews. I will open a follow-up PR.

cmd/publicize/main.go
droslean: I will check how `http` handles the context and make the changes.

cmd/publicize/main.go
droslean: https://github.com/kubernetes/test-infra/pull/18426

cmd/publicize/main.go
droslean: https://github.com/kubernetes/test-infra/pull/18428

cmd/publicize/main.go
droslean: It seems that all of the plugins are using that. I will open a follow-up to share this.

cmd/publicize/main.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+	return o
+}
+
+func (o *options) Validate() error {
+	if (o.gitEmail == "") != (o.gitName == "") {
+		return errors.New("--git-name and --git-email must be specified")
+	}
+
+	for _, group := range []flagutil.OptionGroup{&o.github} {
```
stevekuznetsov, line 89: no reason to loop with just one group

cmd/publicize/main.go
```diff
@@ -0,0 +1,210 @@
...
+	o := gatherOptions()
+	if err := o.Validate(); err != nil {
+		logrus.Fatalf("Invalid options: %v", err)
+	}
+
+	configWatchAndUpdate, err := o.getConfigWatchAndUpdate()
+	if err != nil {
+		logrus.WithError(err).Fatal("couldn't get config file watch and update function")
+	}
+	go interrupts.Run(configWatchAndUpdate)
```
stevekuznetsov, line 160: `interrupts.Run()` is not blocking, no need to spawn a Goroutine for it

cmd/publicize/main.go
```diff
@@ -0,0 +1,210 @@
...
+		var c *Config
+		if err := yaml.Unmarshal(bytes, &c); err != nil {
+			return fmt.Errorf("Couldn't unmarshal publicize configuration: %w", err)
+		}
+
+		if err := c.validate(); err != nil {
+			return err
+		}
+
+		o.config = c
```
stevekuznetsov, line 135: You need to lock around this

cmd/publicize/main.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	gitClient, err := o.git.GitClient(githubClient, githubTokenGenerator, secretAgent.Censor, o.dryRun)
+	if err != nil {
+		logrus.WithError(err).Fatal("Error getting Git client.")
+	}
+
+	serv := &server{
+		githubTokenGenerator: githubTokenGenerator,
+		config:               func() *Config { return o.config },
```
stevekuznetsov, line 187: You need to lock around this

cmd/publicize/server.go
stevekuznetsov: That's one layer away. `hook` the binary does not return an error. but each individual handler does. In this function here, you are implementing an individual handler and the server in one place, so I guess it makes sense

cmd/publicize/server.go
```diff
@@ -0,0 +1,191 @@
...
+		return
+	}
+
+	// Only org members should be able to publicize a pull request.
+	ok, err := s.ghc.IsMember(org, commentAuthor)
+	if err != nil {
+		return
+	}
+	if !ok {
+		s.createComment(ic, fmt.Sprintf("only [%s](https://github.com/orgs/%s/people) org members may request a publicize", org, org), logger)
```
stevekuznetsov, line 86: `may request publication of a private pull request`

cmd/publicize/server.go
stevekuznetsov: Is this resolved? I don't want us to have special code in here that's only used in manual testing

cmd/publicize/server.go
stevekuznetsov: Does not seem to be resolved yet?

cmd/publicize/server.go
stevekuznetsov: Outstanding -- if's a commit, then 

```
A merge commit [%s/%s@%s]() was created in the upstream repository to publish this work.
```

cmd/publicize/server.go
droslean: That can break the tests because https://github.com/openshift/ci-tools/pull/872/files#diff-cad94c3357c2c057cd064113902c35a8R120

cmd/publicize/server.go
droslean: This will rely on the git v2 design here. I will do it as a follow up since some investigation is needed.

cmd/publicize/server.go
```diff
@@ -0,0 +1,183 @@
...
+		logger.WithError(err).Warn("error occurred while checking for prerequisites")
+		s.createComment(ic, fmt.Sprintf("%v", err), logger)
+		return
+	}
+
+	destOrgRepo := s.config().Repositories[fmt.Sprintf("%s/%s", org, repo)]
+	destOrg := strings.Split(destOrgRepo, "/")[0]
+	destRepo := strings.Split(destOrgRepo, "/")[1]
+	sourceRemoteResolver := func() (string, error) {
+		return fmt.Sprintf("https://%s@%s/%s/%s", string(s.githubTokenGenerator()), github.DefaultHost, org, repo), nil
```
stevekuznetsov, line 90: Why is this the default host? This needs to be the host that the user provided at startup, with the flag.

cmd/publicize/server.go
```diff
@@ -0,0 +1,183 @@
...
+	if err := s.checkPrerequisites(logger, pr, ic); err != nil {
+		logger.WithError(err).Warn("error occurred while checking for prerequisites")
+		s.createComment(ic, fmt.Sprintf("%v", err), logger)
+		return
+	}
+
+	destOrgRepo := s.config().Repositories[fmt.Sprintf("%s/%s", org, repo)]
+	destOrg := strings.Split(destOrgRepo, "/")[0]
+	destRepo := strings.Split(destOrgRepo, "/")[1]
+	sourceRemoteResolver := func() (string, error) {
```
stevekuznetsov, line 89: Is this not a resolver you can import from upstream?

cmd/publicize/server.go
```diff
@@ -0,0 +1,183 @@
...
+
+	logger.Infof("Trying to merge the PR to destination: %s/%s@%s", destOrg, destRepo, baseBranch)
+	mergeMsg := fmt.Sprintf("merge history from %s/%s", org, repo)
+	if err := s.mergeAndPushToRemote(destOrg, destRepo, sourceRemoteResolver, baseBranch, mergeMsg, s.dry); err != nil {
+		logger.WithError(err).Warnf("couldn't merge the pull request and push to the destination: %v", err)
+		s.createComment(ic, fmt.Sprintf("Publicize failed with error: %v", err), logger)
+		return
+	}
+
+	destOrgRepoLink := fmt.Sprintf("https://%s/%s/tree/%s", github.DefaultHost, destOrgRepo, baseBranch)
```
stevekuznetsov, line 101: This should also be the host that the user passed in

cmd/publicize/server.go
droslean: Nope.

cmd/publicize/server.go
droslean: Yea, it makes sense. I will address this.

droslean: addressed.

cmd/publicize/server.go
droslean: addressed

PR #875, opened 2020-06-01
Add script to revert ci-operator version

alvaroaleman:
As I don't want to search slack next time this is needed.

### DIFF
--- a/hack/revert-ci-operator.sh
+++ b/hack/revert-ci-operator.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+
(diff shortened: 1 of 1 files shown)


hack/revert-ci-operator.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+kubectl config use-context api.ci
```
petr-muller, line 5: It's a nit but let's make the script depend on just `oc`. Either `oc config use-context`, or just run the oc with `--context` in the commands below.

alvaroaleman: Done

PR #884, opened 2020-06-03
Test images distributor: Allow to explicitly configure additional images

alvaroaleman:
Needed for the clonerefs image which is hardcoded in ci-operator (https://github.com/openshift/ci-tools/blob/b2f25a08501c72824dfb7cbe2ab9c314f6869c6b/pkg/defaults/defaults.go#L430-L434) and thus not explicitly referenced by any test


cmd/dptp-controller-manager/main.go
```diff
@@ -60,7 +61,9 @@ func (o *options) addDefaults() {
 }
 
 type testImagesDistributorOptions struct {
-	imagePullSecretPath string
+	imagePullSecretPath          string
+	whiteListedImageStramTagsRaw flagutil.Strings
+	whiteListedImageStreamTags   sets.String
```
petr-muller, line 66: I don't insist on it, but I saw @thiagoalessio 's post on aos-devel about calling this with neutral `allowlist/denylist` terminology instead, and thought it's a good idea. WDYT? (I also want to address this in our other tools where we already have allowlists).

cmd/dptp-controller-manager/main.go
```diff
@@ -86,6 +89,7 @@ func newOpts() (*options, error) {
 	flag.StringVar(&opts.leaderElectionSuffix, "leader-election-suffix", "", "Suffix for the leader election lock. Useful for local testing. If set, --dry-run must be set as well")
 	flag.Var(&opts.enabledControllers, "enable-controller", fmt.Sprintf("Enabled controllers. Available controllers are: %v. Can be specified multiple times. Defaults to %v", allControllers.List(), opts.enabledControllers.Strings()))
 	flag.StringVar(&opts.testImagesDistributorOptions.imagePullSecretPath, "testImagesDistributorOptions.imagePullSecretPath", "", "A file to use for reading an ImagePullSecret that will be bound to all `default` ServiceAccounts in all namespaces that have a test ImageStream on all build clusters")
+	flag.Var(&opts.testImagesDistributorOptions.whiteListedImageStramTagsRaw, "testImagesDistributorOptions.whitelisted-image-stream-tag", "An imagestreamtag that will be distributed even if no test explicitly references it. It must be in namespace/name format (e.G `ci/clonerefs:latest`). Can be passed multiple times.")
```
petr-muller, line 92:
```suggestion
	flag.Var(&opts.testImagesDistributorOptions.whiteListedImageStramTagsRaw, "testImagesDistributorOptions.whitelisted-image-stream-tag", "An imagestreamtag that will be distributed even if no test explicitly references it. It must be in namespace/name:tag format (e.G `ci/clonerefs:latest`). Can be passed multiple times.")
```

cmd/dptp-controller-manager/main.go
```diff
@@ -111,6 +115,22 @@ func newOpts() (*options, error) {
 		}
 	}
 
+	opts.testImagesDistributorOptions.whiteListedImageStreamTags = sets.String{}
+	if vals := opts.testImagesDistributorOptions.whiteListedImageStramTagsRaw.Strings(); len(vals) > 0 {
+		for _, val := range vals {
+			slashSplit := strings.Split(val, "/")
+			if len(slashSplit) != 2 {
+				errs = append(errs, fmt.Errorf("--testImagesDistributorOptions.whitelisted-image-stream-tag value %s was not in namespace/name format", val))
```
petr-muller, line 123:
```suggestion
				errs = append(errs, fmt.Errorf("--testImagesDistributorOptions.whitelisted-image-stream-tag value %s was not in namespace/name:tag format", val))
```

cmd/dptp-controller-manager/main.go
alvaroaleman: Good point, I changed the whole thing to `additional-image-stream-tag`, the whole allow-list name was an artifact of how its implemented which shouldn't play a role in the flag naming

alvaroaleman:
> Hrm, out of curiosity, how was the clonerefs image updated until now on build0

Its because I ran the controller locally yesterday, how else am I supposed to know it works? :upside_down_face: 

PTAL

PR #886, opened 2020-06-04
Bugzilla backporter tool

GeoBK:
The bugzilla backporter tool will be responsible for showing the clones of a given bug. This PR implements the infrastructure for a simple web-interface which would show those details given a bug ID.


cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+	flatRegistry bool
+}
+
+type traceResponseWriter struct {
+	http.ResponseWriter
+	statusCode int
+	size       int
+}
+
+func handleWithMetrics(h http.HandlerFunc) http.HandlerFunc {
```
alvaroaleman, line 79: If we now have the identical code in multiple places we should probabl extract it somewhere in a follow-up and re-use it

GeoBK:
https://issues.redhat.com/browse/DPTP-1234

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
+	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
+	fs.StringVar(&o.uiAddress, "ui-address", ":8082", "Address to run the registry UI on")
+	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
```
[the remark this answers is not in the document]
GeoBK: i am expecting to use it in the next phase, for the front end server

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+		w.Write([]byte(http.StatusText(http.StatusNotFound)))
+	}
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
+	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
+	fs.StringVar(&o.uiAddress, "ui-address", ":8082", "Address to run the registry UI on")
```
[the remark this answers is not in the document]
GeoBK: i am expecting to use it in the next phase, for the front end server

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
GeoBK: i believe im using grace period as a parameter for the server

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+func getAuthToken() []byte {
+	// Auth token commented out for security reasons
+	// Will modify function to use serviceAgent
+	return []byte("")
+}
+
+// Handler returns a function which populates the response with the details of the bug
+// Expects a POST request with field ID
+// Returns bug details in JSON format
+func getBugHandler() http.HandlerFunc {
```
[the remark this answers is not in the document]
GeoBK: I was expecting to keep a front end for that which would make a POST request with the ID in the form.. Is that not what was required?

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+	return nil
+}
+
+func main() {
+	o := gatherOptions()
+	err := validateOptions(o)
+	if err != nil {
+		log.Fatalf("invalid options: %v", err)
+	}
+	level, _ := log.ParseLevel(o.logLevel)
```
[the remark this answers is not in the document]
GeoBK: That is being checked in the validateOptions function, but i'm not exiting at that point. Do you thing it would be a better idea to fail and exit at that point?

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
alvaroaleman: Sure, but you can also re-use the handler and make it do different things based on the method (both is fine)

cmd/bugzilla-backporter/main.go
GeoBK: Do you want me to do it as part of this pr itself? If so where should I move that to? Is pkg/api the right place?

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
alvaroaleman: Okay, I think it would be easier to just have one listener with multiple endpoints, the reason this ended up being split in in the ci-operator-configresolver is that one of the listeners uses plain http and one uses tls.

Also, as a bit of a heads-up, we want the front-end to be super simplistic and just a go template that renders some html/css and no javascript/typescript, check as a sample e.g. here: https://github.com/openshift/ci-tools/blob/master/pkg/webreg/webreg.go#L1895

The reason for this is that we do not have anyone on the team who has experience to speak of in javascript/typescript and by sticking to go, everyone can make and review changes. 

We can rethink this decision should we end up needing something dynamic, but I think it should be fine without.

cmd/bugzilla-backporter/main.go
alvaroaleman: No, not in this pr, as a follow-up. Maybe a new `pkg/instrumentation` would be a good place

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
alvaroaleman: I think I would just avoid doing it twice, maybe rename `validateOptions` to `completeOptions` and set the log level right there (unless the parsing errored, ofc)

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+			return
+		}
+
+		encoded, err := ioutil.ReadAll(r.Body)
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			w.Write([]byte("Could not read bug ID from request body."))
+			return
+		}
+		client := bugzilla.NewClient(getAuthToken, "https://bugzilla.redhat.com/")
```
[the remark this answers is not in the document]
GeoBK: Absolutely. my bad, moving that out of there.

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,187 @@
...
+		h(trw, r)
+		latency := time.Since(t)
+		labels := prometheus.Labels{"status": strconv.Itoa(trw.statusCode), "path": r.URL.EscapedPath()}
+		bzbpMetrics.httpRequestDuration.With(labels).Observe(latency.Seconds())
+		bzbpMetrics.httpResponseSize.With(labels).Observe(float64(trw.size))
+	})
+}
+
+// Needs additional work
+func getAuthToken() []byte {
```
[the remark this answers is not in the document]
GeoBK: I am not sure I understand, could you please explain what you mean by that?

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
GeoBK: i think the intention with the validate options was to leave room to extend it for the rest of the options... but I guess its overkill since I'm not expecting that many options, at least at this point.

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
alvaroaleman: Commandline flags so this can be configured to read the token from a file. It seems we already have some helpers for this in upstream: https://github.com/kubernetes/test-infra/blob/4b5c7c99a851eb427f5c77bd0c8d11526f7b63c4/prow/cmd/hook/main.go#L57-L85

cmd/bugzilla-backporter/main.go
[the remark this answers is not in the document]
GeoBK: Alright sounds good, thank you for the clarification.

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,306 @@
...
+			jsonBugInfo, err := json.MarshalIndent(*bugInfo, "", "  ")
+			if err != nil {
+				w.WriteHeader(http.StatusInternalServerError)
+				fmt.Fprintf(w, "failed to marshal bugInfo to JSON: %v", err)
+				// logger.WithError(err).Errorf("failed to marshal bugInfo to JSON")
+				return err
+			}
+			w.WriteHeader(http.StatusOK)
+			w.Write(jsonBugInfo)
+		} else {
```
alvaroaleman, line 243: The idiomatic way to do thing like this in go is to return early, rather than writing arrow code https://blog.codinghorror.com/flattening-arrow-code/

So here this would mean:
```
if method != "GET" {
  error handling
}
// business logic here
```

rather than
```
if method == "GET" {
 a
 large
 amount
 of
 business
 logic
else {
  error handling
}
```

Because by the time the reader of the code reaches the `else`, they will have forgotten what the condition checks for

GeoBK: Thank you for pointing that out

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,296 @@
...
+
+	"k8s.io/test-infra/prow/bugzilla"
+)
+
+const (
+	//BugIDQuery stores the query for bug ID
+	BugIDQuery = "ID"
+)
+
+const htmlPageStart = `
```
alvaroaleman, line 18: This still needs to be moved to a `var` that gets populated via `template.Must`

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,296 @@
+package backporter
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+	"text/template"
```
alvaroaleman, line 8: I think we should be using `html/template` here, it has some escaping and such

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,296 @@
...
+	}
+}
+
+// GetBugHandler returns a function which populates the response with the details of the bug
+// Returns bug details in JSON format
+func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
+	return func(w http.ResponseWriter, r *http.Request) error {
+		if r.Method != "GET" {
+			w.WriteHeader(http.StatusNotImplemented)
+			w.Write([]byte(http.StatusText(http.StatusNotImplemented)))
```
alvaroaleman, line 213: This is `BadRequest`, `NotImplemented` is a 5XX error implying something on the server side went wrong (here: We didn't implement this) but that is not the case here, its a BadRequest as we are not going to implement it

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,296 @@
...
+		return err
+	}
+	fmt.Fprint(w, htmlPageEnd)
+	return nil
+}
+
+// GetLandingHandler will return a simple bug search page
+func GetLandingHandler() HandlerFuncWithErrorReturn {
+	return func(w http.ResponseWriter, req *http.Request) error {
+		writePage(w, "Home", emptyTemplate, nil)
```
alvaroaleman, line 202: Err, that is a very complicated way to print a string into the body. Simpler would be:

```
if _, err := w.Write([]byte("Home")); err != nil {
  return err
}
```

(tbh I do not see a god reason for `emptyTemplate` to exist. Just write directly where its currently used. )

alvaroaleman: `fmt.Fprintf(w, "format error: %v", err)` also works for simple cases

vendor/k8s.io/test-infra/pkg/flagutil/flagutil.go
```diff
@@ -0,0 +1,34 @@
+/*
```
alvaroaleman, line 1: I don't think this is suppoed to be in this PR

pkg/backporter/backporter.go
GeoBK: This is not a template, it is just a string. Does it still need to be moved in a template.Must? If so, why?

alvaroaleman: ah okay, then its fine :)

vendor/k8s.io/test-infra/pkg/flagutil/flagutil.go
GeoBK: This was generated when I ran the update-vendor/validate-vendor, after merging the changes from the update-test-infra. I ran those commands as the validate-vendor checks were still failing on this pr and thats why I included expected.yaml, BUILD.bazel, flagutil.go and modules.txt. Should i drop just this file or all 4 of the files which were included?

pkg/backporter/backporter.go
GeoBK: The error needs to go in the page and dont think it would be good aesthetically to send just the error message, especially considering that this is not an ajax call but a full page render.

vendor/k8s.io/test-infra/pkg/flagutil/flagutil.go
alvaroaleman: yeah, ideally you can soft reset your branch to the oldest commit in it ~1, create one commit and rebase it onto master.

One problem with merging changes from the basebranch in rather than rebasing is that they might show up in the PR which is confusing.

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,292 @@
...
+
+// HandlerFuncWithErrorReturn allows returning errors to be logged
+type HandlerFuncWithErrorReturn func(http.ResponseWriter, *http.Request) error
+
+type wrapper struct {
+	Bug    *bugzilla.Bug
+	Clones []*bugzilla.Bug
+	Parent *bugzilla.Bug
+	PRs    []bugzilla.ExternalBug
+}
```
petr-muller, line 180: It's not obvious to me what this `wrapper` is (what does it wrap?). A godoc would be helpful, ideally also describing what the members are.

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,292 @@
...
+// GetLandingHandler will return a simple bug search page
+func GetLandingHandler() HandlerFuncWithErrorReturn {
+	return func(w http.ResponseWriter, req *http.Request) error {
+		writePage(w, "Home", emptyTemplate, nil)
+		return nil
+	}
+}
+
+// GetBugHandler returns a function which populates the response with the details of the bug
+// Returns bug details in JSON format
```
petr-muller, line 204: Is this line stale? Just the first one seems totally fine to me, no need for this second one.

pkg/backporter/backporter.go
petr-muller: Oh, so it's a "collection of data for clonesTemplate"! Maybe we should call it `clonesTemplateData` or something like that?

GeoBK: Yup. I can make that change

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,292 @@
...
+}
+
+// GetClonesHandler returns an HTML page with detais about the bug and its clones
+func GetClonesHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
+	return func(w http.ResponseWriter, req *http.Request) error {
+		bugIDStr := req.URL.Query().Get(BugIDQuery)
+		bugID, err := strconv.Atoi(bugIDStr)
+		if bugIDStr == "" || err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			writePage(w, "Error!", emptyTemplate, fmt.Sprintf("%s - query incorrect", BugIDQuery))
```
alvaroaleman, line 245: Nit: Would be nice to return the error

pkg/backporter/backporter_test.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	bug1, err := fake.GetBug(bug1ID)
+	if err != nil {
+		t.Errorf("Error retreiving bug: %v", err)
+	}
+	cloneID, err := fake.CloneBug(bug1)
+	if err != nil {
+		t.Errorf("Error while cloning bug: %v", err)
+	}
+	fmt.Println(cloneID)
```
alvaroaleman, line 164: Leftover? Generally you would do a `t.Log`, it can then be shown by running the tests with the `-v` flag

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,137 @@
...
+		bzbpMetrics.httpRequestDuration.With(labels).Observe(latency.Seconds())
+		bzbpMetrics.httpResponseSize.With(labels).Observe(float64(trw.size))
+	})
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
+	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
```
alvaroaleman, line 93: Nit: Rename to `listen-address`

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+type traceResponseWriter struct {
+	http.ResponseWriter
+	statusCode int
+	size       int
+}
+
+func handleWithMetrics(h backporter.HandlerFuncWithErrorReturn) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
petr-muller, line 74: My IDE says this conversion is unnecessary.

cmd/bugzilla-backporter/main.go
```diff
@@ -0,0 +1,137 @@
...
+	bugzillaClient, err := o.bugzilla.BugzillaClient(secretAgent)
+	if err != nil {
+		log.WithError(err).Fatal("Error getting Bugzilla client.")
+	}
+	health := pjutil.NewHealth()
+	metrics.ExposeMetrics("ci-operator-bugzilla-backporter", prowConfig.PushGateway{}, prowflagutil.DefaultMetricsPort)
+
+	http.HandleFunc("/", handleWithMetrics(backporter.GetLandingHandler()))
+	http.HandleFunc("/getclones", handleWithMetrics(backporter.GetClonesHandler(bugzillaClient)))
+	//Leaving this in here to help with future debugging. This will return bug details in JSON format
```
petr-muller, line 131:
```suggestion
	// Leaving this in here to help with future debugging. This will return bug details in JSON format
```

pkg/backporter/backporter.go
petr-muller: It's not just a nit, the error handling could be better overall: `writePage` returns an error but it's basically never read. Either that method should not return an error (it could just log problems), or its errors should not be ignored.

cmd/bugzilla-backporter/main.go
petr-muller: Same thing is three times in `backporter_test.go`

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,292 @@
...
+
+// Class to hold the UI data for the clones page
+type clonesTemplateData struct {
+	Bug    *bugzilla.Bug          //bug details
+	Clones []*bugzilla.Bug        // List of clones for the bug
+	Parent *bugzilla.Bug          // Root bug if it is a a bug, otherwise holds itself
+	PRs    []bugzilla.ExternalBug // Details of linked PR
+}
+
+//Writes an HTML page, prepends header in htmlPageStart and appends header from htmlPageEnd around tConstructor.
```
petr-muller, line 183: nit: comments should have whitespace between `//` and the text (multiple times in this PR)

pkg/backporter/backporter.go
```diff
@@ -0,0 +1,292 @@
...
+func writePage(w http.ResponseWriter, title string, body *template.Template, data interface{}) error {
+	fmt.Fprintf(w, htmlPageStart, title)
+
+	if err := body.Execute(w, data); err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		fmt.Fprintf(w, "%s: %v", http.StatusText(http.StatusInternalServerError), err)
+		return err
+	}
+	fmt.Fprint(w, htmlPageEnd)
+	return nil
```
petr-muller, line 193: I am not a fan of ignoring the errors from the `Fprintf` here - `writePage is supposed to return an `error` when something goes wrong, why not propagate (or at least log) them?

pkg/backporter/backporter.go
GeoBK: @petr-muller The error returned by error handler is meant to be logged in the handleWithMetrics function in the recordError function. Is that not a good way to log errors?

pkg/backporter/backporter_test.go
GeoBK: leftover. sorry.

pkg/backporter/backporter.go
GeoBK: yup. fixing that now

pkg/backporter/backporter.go
petr-muller: What I meant was the `writePage`'s return value, which is not checked anywhere. So either that function should not return errors at all (and handle them locally, by e.g. just logging), or it should be checked and handled at the callsites. But having `writePage` return errors and ignoring them at callsites seems wrong.

GeoBK: > Nit: Would be nice to return the error

@alvaroaleman I was returning only the error for the empty string, now I have separated both of those conditions out

pkg/backporter/backporter.go
GeoBK: qq - Is this comment specific to this Fprintf or do you mean that I should have done that in general?

cmd/bugzilla-backporter/main.go
GeoBK: Yup. Those conversions are redundant, I'll remove them.

pkg/backporter/backporter.go
GeoBK: @petr-muller Yup. got what you meant now. I'll make sure the returned errors are being used for writePage

vendor/k8s.io/test-infra/pkg/flagutil/flagutil.go
GeoBK: Resolving - Added an extra module, therefore we need these files

pkg/backporter/backporter.go
GeoBK: Ok. I have added error checks inside the error handlers.

alvaroaleman: Still missing an errcheck right here :upside_down_face:

GeoBK: The first time I checked I somehow didnt see that fprint returned an error, which caused this whole thread. I corrected that mistake😓. Thank you for pointing it out

pkg/backporter/backporter_test.go
```diff
@@ -0,0 +1,217 @@
...
+		</tbody>
+	</table>`
+
+const errorSubPage = `Bug#1000 not found`
+
+func unwrapper(h HandlerFuncWithErrorReturn, t *testing.T) http.HandlerFunc {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		err := h(w, r)
+		if err != nil {
+			// Error needs to be checked for the linter, but cant be ignored since the tescases check for error states
```
alvaroaleman, line 61: Sorry I don't understand? What keeps you from doing `_ = h(w, r)` if you do not care about the handler error?

GeoBK: Nothing actually, I can do that.

PR #889, opened 2020-06-04
Use methods for getting/setting the namespace on jobConfig

alvaroaleman:
Before, the namespace was a public property of the jobSpec. This lead to
semi-common wrong usage where it was evaluated before it was set,
resulting in bugs.

This PR changes jobSpec to provide getters and setters instead. The
getter will log an error if the namespace was unset, as this is always a
bug. This allows us to easely identify such wrong usages without making
any tests fail due to a potentially non-fatal issue.

Also fixes a bug where the namespace property was used wrongly as
described above.

### DIFF
--- a/pkg/api/job_spec.go
+++ b/pkg/api/job_spec.go
@@ -4,6 +4,8 @@ import (
 	"crypto/sha256"
 	"encoding/json"
 	"fmt"
+	"log"
+	"runtime/debug"
 
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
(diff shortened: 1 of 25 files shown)


pkg/api/job_spec.go
```diff
@@ -17,13 +19,27 @@ type JobSpec struct {
...
+// at step construction time because its unset there
+func (s *JobSpec) Namespace() string {
+	if s.namespace == "" {
+		log.Println("Warning, namespace accessed before it was set, this is a bug in ci-operator. Stack:")
+		debug.PrintStack()
+	}
+	return s.namespace
+}
+
+func (s *JobSpec) SetNamspace(namespace string) {
```
petr-muller, line 39:
```
func (s *JobSpec) SetNamspace(namespace string)
                  ^^^Namespace^^^
```

...and of course the callsites

pkg/api/job_spec.go
```diff
@@ -17,13 +19,27 @@ type JobSpec struct {
...
 	// if set, any new artifacts will be a child of this object
 	owner *meta.OwnerReference
 }
 
+// Namespace returns the namespace of the job. Must not be evaluated
+// at step construction time because its unset there
+func (s *JobSpec) Namespace() string {
+	if s.namespace == "" {
+		log.Println("Warning, namespace accessed before it was set, this is a bug in ci-operator. Stack:")
+		debug.PrintStack()
```
petr-muller, line 34: Hrm, should we just panic here? Something's likely going to fail later anyway and failing fast would force us to fix it,

alvaroaleman: I thought about this but decided against it because it might not be a reason to abort the test. For example the place where I initially walked on this was that we failed to label the namespace to prevent it from getting deleted.

If it is a reason to fail the test (like with the ci-operator bug the other day) then the test will presumably fail anyways down the line.

pkg/api/job_spec.go
alvaroaleman: Fixed

PR #891, opened 2020-06-05
ci-operator: allow waiting before importing a release

stevekuznetsov:
Today, some jobs are using the following workflow:

 - build an image or set `tag_specification`
 - also provide `$RELEASE_IMAGE_LATEST`

With the previous implementation, two separate steps would create images
to fill out the stable stream and overwrite each other's work. The
effect was that the jobs could end up with images in the stable stream
that were not in the imported release from the env var, as they would be
created by the image steps and not be overwritten.

By allowing an explicit configuration for the import step, we can wait
on these other steps in the cases where we'd do in the past and continue
to provide the same user experience. This is not ideal and we need to
provide a clear way for users to simply import images they are
interested in, since the combination of `tag_specification` and a
`$RELEASE_IAMGE_LATEST` is nonsensical, as they are nominally both means
to configure which release of OpenShift is being tested.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -186,7 +186,7 @@ func FromConfig(
 						return nil, nil, results.ForReason("reading_release").ForError(fmt.Errorf("failed to read input release pullSpec %s: %v", name, err))
 					}
 					log.Printf("Resolved release %s to %s", name, pullSpec)
-					releaseStep = release.ImportReleaseStep(name, pullSpec, config.Resources, podClient, imageClient, saGetter, rbacClient, artifactDir, jobSpec, dryLogger)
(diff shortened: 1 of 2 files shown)


alvaroaleman:
> since the combination of tag_specification and a
$RELEASE_IAMGE_LATEST is nonsensical, as they are nominally both means
to configure which release of OpenShift is being tested.

So we should validate that ppl don't do that and make that validation mandatory once we got rid of this workaround?

/lgtm

stevekuznetsov:
> So we should validate that ppl don't do that and make that validation mandatory once we got rid of this workaround?

@alvaroaleman  yes -- in general I expect *nobody* to be using literal `$RELEASE_IMAGE_foo` variables in the future, since you can use the actual `releases:` configuration to do precisely what you need. Before we generate all release jobs it will remain in this semi-state. We validate in the other direction already (don't ask for `latest` in `releases` if you have `tag_specification`).

PR #892, opened 2020-06-05
Remove cluster field from ci-operator types

alvaroaleman:
This PR removes the cluster field from the CI-Operator types. It will require re-generating the configs and thus potentially break in-flight PRs to o/release.


stevekuznetsov:
Don't we first need a PR to `openshift/release` to remove the use of these fields?

alvaroaleman:
> Don't we first need a PR to openshift/release to remove the use of these fields?

It will fail validation.

We could:
* Remove only the validation in a dedicated PR
* Update o/release
* Then merge this

WDYT?

stevekuznetsov:
Yeah, I think that would be optimal. We already fall back to in-cluster clients without the `cluster` field so removing the validation, then the fields should be a no-op.

stevekuznetsov:
I think we're ready for this now

/lgtm
/hold

alvaroaleman:
ffs
/retest

PR #896, opened 2020-06-08
 ci-operator: refactor step links to be generic on is or istag

stevekuznetsov:
ci-operator: remove step link equality

This method was unused.

---

ci-operator: rename step link method to better reflect usage

The `Matches(other)` function was used to determine if `other` satisfied
the requirements of the link being evaluated. This rename makes that
relationship clear.

---

ci-operator: treat latest and initial stable as an API

---

ci-operator: refactor step links to be generic on is or istag

While it remains important to be able to differentiate at the call-site
on whether we're talking about an internal ImageStreamTag, a release
payload image, or other, it's not important to have a plethora of types
under the hood that do nothing special when we're just matching on name
and tag.

---

/assign @petr-muller @bbguimaraes

### DIFF
--- a/pkg/steps/lease_test.go
+++ b/pkg/steps/lease_test.go
@@ -29,10 +29,12 @@ func (s *stepNeedsLease) Run(ctx context.Context, dry bool) error {
 	return nil
 }
 
-func (stepNeedsLease) Name() string             { return "needs_lease" }
-func (stepNeedsLease) Description() string      { return "this step needs a lease" }
-func (stepNeedsLease) Requires() []api.StepLink { return []api.StepLink{api.ReleaseImagesLink()} }
-func (stepNeedsLease) Creates() []api.StepLink  { return []api.StepLink{api.ImagesReadyLink()} }
+func (stepNeedsLease) Name() string        { return "needs_lease" }
+func (stepNeedsLease) Description() string { return "this step needs a lease" }
+func (stepNeedsLease) Requires() []api.StepLink {
+	return []api.StepLink{api.StableImagesLink(api.LatestStableName)}
+}
+func (stepNeedsLease) Creates() []api.StepLink { return []api.StepLink{api.ImagesReadyLink()} }
 
 func (stepNeedsLease) Provides() (api.ParameterMap, api.StepLink) {
 	return api.ParameterMap{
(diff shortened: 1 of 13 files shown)


pkg/steps/lease_test.go
```diff
@@ -29,10 +29,12 @@ func (s *stepNeedsLease) Run(ctx context.Context, dry bool) error {
...
 }
 
-func (stepNeedsLease) Name() string             { return "needs_lease" }
-func (stepNeedsLease) Description() string      { return "this step needs a lease" }
-func (stepNeedsLease) Requires() []api.StepLink { return []api.StepLink{api.ReleaseImagesLink()} }
-func (stepNeedsLease) Creates() []api.StepLink  { return []api.StepLink{api.ImagesReadyLink()} }
+func (stepNeedsLease) Name() string        { return "needs_lease" }
+func (stepNeedsLease) Description() string { return "this step needs a lease" }
+func (stepNeedsLease) Requires() []api.StepLink {
+	return []api.StepLink{api.StableImagesLink(api.LatestStableName)}
```
stevekuznetsov, line 35: As there's no real generic "release images" intermediary, all of these have changed to explicitly depend on the full presence of the "stable" image stream. This should be a no-op.

pkg/steps/output_image_tag.go
```diff
@@ -72,7 +72,7 @@ func (s *outputImageTagStep) run(ctx context.Context, dry bool) error {
 }
 
 func (s *outputImageTagStep) Requires() []api.StepLink {
-	return []api.StepLink{api.InternalImageLink(s.config.From), api.ReleaseImagesLink()}
```
stevekuznetsov, line 75: This step promotes a `pipeline` ImageStreamTag into something else. I have no idea why this *ever* relied on the `stable` stream existing. The original author isn't of any help in understanding this.

pkg/steps/release/release_images.go
```diff
@@ -214,7 +215,10 @@ func (s *releaseImagesTagStep) Requires() []api.StepLink {
 }
 
 func (s *releaseImagesTagStep) Creates() []api.StepLink {
-	return []api.StepLink{api.ReleaseImagesLink()}
+	return []api.StepLink{
+		api.StableImagesLink(api.InitialStableName),
```
stevekuznetsov, line 219: Technically this is not a no-op ... but we always created both streams here, we just never exposed the `stable-initial` one nor could anyone require it.

stevekuznetsov:
Builds on top of (but is entirely separate from) https://github.com/openshift/ci-tools/pull/895

stevekuznetsov:
@petr-muller rebased

PR #898, opened 2020-06-08
[ci-operator-config-mirror] Add the ability to mirror the configs from a specific organization

droslean:
Introduces the `--only-org` flag. Give the tool the ability to mirror the ci-operator configurations that belong in a specified organization.

### DIFF
--- a/cmd/ci-operator-config-mirror/main.go
+++ b/cmd/ci-operator-config-mirror/main.go
@@ -28,6 +28,7 @@ type options struct {
 
 	configDir string
 	toOrg     string
+	onlyOrg   string
 
 	clean bool
 }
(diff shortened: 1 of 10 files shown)


test/integration/ci-operator-config-mirror.sh
```diff
@@ -10,6 +10,8 @@ actual="${BASETMPDIR}/input"
 actual_to_clean="${BASETMPDIR}/input-to-clean"
 
 expected="${suite_dir}/output"
+expected_only_super="${suite_dir}/output-only-super"
```
stevekuznetsov, line 13: If this is used just once, don't make a var for it

cmd/ci-operator-config-mirror/main.go
```diff
@@ -109,6 +111,10 @@ func main() {
 			return nil
 		}
 
+		if len(o.onlyOrg) > 0 && repoInfo.Org != o.onlyOrg {
```
stevekuznetsov, line 114: nit: you don't need the `len()` check and we should comment that we're skipping like in the next `if`

droslean: We need the len check to make sure that the flag is specified. Otherwise, we will skip all configs. +1 for the log, I will address this.

stevekuznetsov: Err, sure. Could also flip the logic.

PR #900, opened 2020-06-08
Rehearsals: Use ctlrruntimeclient for prowjobs

alvaroaleman:
This PR replaces the typed prowjob client in the reharsal tool with the dynamic one from controller-runtime. The motivation is to avoid needing one typed client per apigroup and the motivation for that in turn is https://issues.redhat.com/browse/DPTP-1244 for which we need an Image client.

/assign @petr-muller


alvaroaleman:
Hm now that I've submitted the PR I realize I did this for the wrong client :upside_down_face: 
Changing this gets rid of the fake prowjob client dependency though

pkg/rehearse/jobs.go
```diff
@@ -817,29 +813,34 @@ func (e *Executor) ExecuteJobs() (bool, error) {
...
-			prowJob, ok := event.Object.(*pjapi.ProwJob)
-			if !ok {
-				return false, fmt.Errorf("received a %T from watch", event.Object)
+	var listErrors []error
+	if err := e.pollFunc(10*time.Second, 4*time.Hour, func() (bool, error) {
+		result := &pjapi.ProwJobList{}
+		// Don't bail out just because one LIST failed
+		if err := e.pjclient.List(context.Background(), result, selector, ctrlruntimeclient.InNamespace(e.namespace)); err != nil {
+			if len(listErrors) > 2 {
+				return false, utilerrors.NewAggregate(append(listErrors, err, errors.New("encountered more than three subsequent errors trying to list")))
```
petr-muller, line 827: I think the "more than three" message is misleading, because we actually bail out on three.

alvaroaleman: Updated

PR #901, opened 2020-06-09
registry: copy step environment when resolving

bbguimaraes:
Different resolutions overwrite each other otherwise.

### DIFF
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -166,13 +166,17 @@ func (r *registry) processStep(step *api.TestStep, seen sets.String, stack []sta
 	}
 	seen.Insert(ret.As)
 	var errs []error
-	for i, e := range ret.Environment {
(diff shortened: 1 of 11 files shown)


pkg/registry/resolver.go
```diff
@@ -166,13 +166,17 @@ func (r *registry) processStep(step *api.TestStep, seen sets.String, stack []sta
...
+		for _, e := range ret.Environment {
+			if v := resolveVariable(e.Name, stack); v != nil {
+				e.Default = *v
+			} else if e.Default == "" {
+				errs = append(errs, stackErrorf(stack, "%s: unresolved parameter: %s", ret.As, e.Name))
+			}
+			env = append(env, e)
 		}
-
+		ret.Environment = env
```
bbguimaraes, line 179: The E2E test adds a lot of noise to the PR, this is the relevant part.

PR #906, opened 2020-06-10
Define AllowSkipOnSuccess and OptionalOnSuccess flags

mgencur:
* allow_skip_on_success is an umbrella flag being set inside the "steps" element,
  it gives a signal to all steps that define optional_on_success to be
skipped
* optional_on_success is being set inside individual step

If both are set to true and all previous `pre` and `test` steps were
successful, the given step will be skipped. If the step is a `post` step it will be skipped in this case even if previous `post` steps failed but all `pre` and `test` steps passed.

### DIFF
--- a/pkg/api/config_test.go
+++ b/pkg/api/config_test.go
@@ -477,6 +477,7 @@ func TestValidateTestSteps(t *testing.T) {
 	// string pointers in golang are annoying
 	myReference := "my-reference"
 	asReference := "as"
+	yes := true
 	for _, tc := range []struct {
 		name  string
 		steps []TestStep
@@ -678,7 +679,7 @@ func TestValidateTestSteps(t *testing.T) {
 				From:              "from",
 				Commands:          "commands",
 				Resources:         resources,
-				OptionalOnSuccess: func(b bool) *bool { return &b }(true)},
+				OptionalOnSuccess: &yes},
 		}},
 		errs: []error{
 			errors.New("test[0]: `optional_on_success` is only allowed for Post steps"),
@@ -702,6 +703,7 @@ func TestValidatePostSteps(t *testing.T) {
 		Requests: ResourceList{"cpu": "1"},
 		Limits:   ResourceList{"memory": "1m"},
 	}
+	yes := true
 	for _, tc := range []struct {
 		name  string
 		steps []TestStep
@@ -716,7 +718,7 @@ func TestValidatePostSteps(t *testing.T) {
 				From:              "from",
 				Commands:          "commands",
 				Resources:         resources,
-				OptionalOnSuccess: func(b bool) *bool { return &b }(true)},
+				OptionalOnSuccess: &yes},
 		}},
 	}} {
 		t.Run(tc.name, func(t *testing.T) {
--- a/pkg/api/config_test.go
+++ b/pkg/api/config_test.go
@@ -669,13 +669,62 @@ func TestValidateTestSteps(t *testing.T) {
 		errs: []error{
 			errors.New("test[1].ref: duplicated name \"as\""),
 		},
+	}, {
+		name: "Test step with forbidden parameter",
+
+		steps: []TestStep{{
+			LiteralTestStep: &LiteralTestStep{
+				As:                "as",
+				From:              "from",
+				Commands:          "commands",
+				Resources:         resources,
+				OptionalOnSuccess: func(b bool) *bool { return &b }(true)},
+		}},
+		errs: []error{
+			errors.New("test[0]: `optional_on_success` is only allowed for Post steps"),
+		},
 	}} {
 		t.Run(tc.name, func(t *testing.T) {
 			seen := tc.seen
 			if seen == nil {
 				seen = sets.NewString()
 			}
-			ret := validateTestSteps("test", tc.steps, seen, nil)
+			ret := validateTestStepsTest("test", tc.steps, seen, nil)
+			if !errListMessagesEqual(ret, tc.errs) {
+				t.Fatal(diff.ObjectReflectDiff(ret, tc.errs))
+			}
+		})
+	}
+}
+
+func TestValidatePostSteps(t *testing.T) {
+	resources := ResourceRequirements{
+		Requests: ResourceList{"cpu": "1"},
+		Limits:   ResourceList{"memory": "1m"},
+	}
+	for _, tc := range []struct {
+		name  string
+		steps []TestStep
+		seen  sets.String
+		errs  []error
+	}{{
+		name: "Valid Post steps",
+
+		steps: []TestStep{{
+			LiteralTestStep: &LiteralTestStep{
+				As:                "as",
+				From:              "from",
+				Commands:          "commands",
+				Resources:         resources,
+				OptionalOnSuccess: func(b bool) *bool { return &b }(true)},
+		}},
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			seen := tc.seen
+			if seen == nil {
+				seen = sets.NewString()
+			}
+			ret := validateTestStepsPost("test", tc.steps, seen, nil)
 			if !errListMessagesEqual(ret, tc.errs) {
 				t.Fatal(diff.ObjectReflectDiff(ret, tc.errs))
 			}
@@ -706,7 +755,7 @@ func TestValidateParameters(t *testing.T) {
 		err:    []error{errors.New("test: unresolved parameter(s): [TEST1]")},
 	}} {
 		t.Run(tc.name, func(t *testing.T) {
-			err := validateLiteralTestStep("test", LiteralTestStep{
+			err := validateLiteralTestStepTest("test", LiteralTestStep{
 				As:       "as",
 				From:     "from",
 				Commands: "commands",
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -548,7 +548,7 @@ type LiteralTestStep struct {
 	// OptionalOnSuccess defines if this step should be skipped as long
 	// as all `pre` and `test` steps were successful and AllowSkipOnSuccess
 	// flag is set to true in MultiStageTestConfiguration.
-	OptionalOnSuccess bool `json:"optional_on_success,omitempty"`
+	OptionalOnSuccess *bool `json:"optional_on_success,omitempty"`
 }
 
 // StepParameter is a variable set by the test, with an optional default.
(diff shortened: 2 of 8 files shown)


pkg/steps/multi_stage.go
```diff
@@ -338,10 +341,14 @@ func (s *multiStageTestStep) runSteps(
 	return utilerrors.NewAggregate(errs)
 }
 
-func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []coreapi.EnvVar) ([]coreapi.Pod, error) {
+func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []coreapi.EnvVar,
+	prevErrs []error) ([]coreapi.Pod, error) {
 	var ret []coreapi.Pod
 	var errs []error
 	for _, step := range steps {
+		if s.allowSkipOnSuccess && step.OptionalOnSuccess && len(prevErrs) == 0 {
```
stevekuznetsov, line 349: Since you don't actually need to know the value of the previous errors, why don't we pass in `hasFailedSteps bool` or similar?

mgencur: It's possible, yeah. But in the run() function I'll need to assign hasFailedSteps more times or call len(errs) more times. When we pass the whole array we can only call (len) once in this common function. WDYT?

I'll also add the multi-step config as you suggested above.

mgencur:
Maybe I should add the AllowSkipOnSuccess field also to `MultiStageTestConfiguration` at https://github.com/openshift/ci-tools/blob/master/pkg/api/types.go#L598 and not only to the literal? That way this option could work also for referenced steps. Not sure if this would required any more changes.

mgencur:
@stevekuznetsov could you take a second look, please?

pkg/api/types.go
```diff
@@ -545,6 +545,10 @@ type LiteralTestStep struct {
 	Credentials []CredentialReference `json:"credentials,omitempty"`
 	// Environment lists parameters that should be set by the test.
 	Environment []StepParameter `json:"env,omitempty"`
+	// OptionalOnSuccess defines if this step should be skipped as long
+	// as all `pre` and `test` steps were successful and AllowSkipOnSuccess
+	// flag is set to true in MultiStageTestConfiguration.
+	OptionalOnSuccess bool `json:"optional_on_success,omitempty"`
```
stevekuznetsov, line 551: Hm so technically with the addition of this here, you add it as a possible value to all `pre` and `test` steps, too, which is not part of the contract that you are describing. We should either make a separate type for `post` steps and only expose it there (my preference) or add validation that disallows it on `pre` and `test` steps.

pkg/steps/multi_stage.go
```diff
@@ -156,13 +158,13 @@ func (s *multiStageTestStep) run(ctx context.Context, dry bool) error {
 		return fmt.Errorf("failed to create RBAC objects: %w", err)
 	}
 	var errs []error
-	if err := s.runSteps(ctx, s.pre, env, true); err != nil {
-		errs = append(errs, fmt.Errorf("%q pre steps failed: %w", s.name, err))
-	} else if err := s.runSteps(ctx, s.test, env, true); err != nil {
-		errs = append(errs, fmt.Errorf("%q test steps failed: %w", s.name, err))
+	if err := s.runSteps(ctx, s.pre, env, true, errs); err != nil {
+		errs = append(errs, fmt.Errorf("%q pre steps failed: %v", s.name, err))
```
stevekuznetsov: You'll need to change this back to `%w`

pkg/steps/multi_stage.go
stevekuznetsov: I would prefer to pass the thing we care about -- the boolean -- instead of letting the downstream callers have access to a list of data they do not need.

test/e2e/multi-stage.sh
```diff
@@ -20,6 +20,7 @@ os::cmd::expect_success "ci-operator --artifact-dir ${BASETMPDIR} --resolver-add
 UNRESOLVED_CONFIG="$( cat "${suite_dir}/config.yaml" )"
 export UNRESOLVED_CONFIG
 os::cmd::expect_success "ci-operator --artifact-dir ${BASETMPDIR} --resolver-address http://127.0.0.1:8080 --target with-references"
+os::cmd::expect_success "ci-operator --artifact-dir ${BASETMPDIR} --resolver-address http://127.0.0.1:8080 --target skip-on-success"
```
stevekuznetsov, line 23: Can we put this underneath or above the `UNRESOLVED_CONFIG` lines so it's not using that feature, as it's unrelated? You can pass the config literally with the flag like the other test cases do.

pkg/api/types.go
mgencur: Hmm. Exposing it in a specific type for `post` phase would be quite some work. The LiteralTestStep is used in 101 places :-D 
To me it makes sense to have this option also for `test` steps. Would it make sense to clarify the contract and then add validation that jsut for `pre` phase where it doesn't make sense? Something like this:
```
OptionalOnSuccess defines if this step should be skipped as long as all steps from previous phases (`pre` or `test`) were successful and AllowSkipOnSuccess flag is set to true in MultiStageTestConfiguration. This option is applicable to `test` and `post` steps.
```

pkg/steps/multi_stage.go
mgencur: OK. Pushed a change to pass only a bool flag.

pkg/api/types.go
petr-muller: I don't understand the case where it makes sense for `test` steps, can you explain? `tests` steps won't be run at all when `pre` ones fail, so having a `test` step optional on success would create a step that never runs?

mgencur: I see :-/ that's probably not what we want.

stevekuznetsov: Did this get resolved? Is it ready for another review?

mgencur: No. This will need more work.

mgencur:
@stevekuznetsov @petr-muller hey, I've added the validation that only allows the given field in `Post` steps. Please let me know how much you like/don't like it. I feel like creating more API types and specific logic for the Post phase would would complicate the overall design too much.

pkg/api/config_test.go
```diff
@@ -669,13 +669,62 @@ func TestValidateTestSteps(t *testing.T) {
...
+	}, {
+		name: "Test step with forbidden parameter",
+
+		steps: []TestStep{{
+			LiteralTestStep: &LiteralTestStep{
+				As:                "as",
+				From:              "from",
+				Commands:          "commands",
+				Resources:         resources,
+				OptionalOnSuccess: func(b bool) *bool { return &b }(true)},
```
stevekuznetsov, line 681: This is neat but generally I think we just do something like the following at the top of the `func Test...`

```go
yes := true

// ...

    OptionalOnSuccess: &yes
```

mgencur: Fixed

mgencur:
@stevekuznetsov I did my best to add the docs but don't know how to check the rendered format. Hopefully this will be alright.

stevekuznetsov:
Deployed a preview of the server at https://server-ci-tools-906.apps.ci.l2s4.p1.openshiftapps.com using 

```
USER=mgencur BRANCH=skip_on_success PULL_REQUEST=906 make pr-deploy
```

pkg/webreg/webreg.go
```diff
@@ -1156,6 +1156,17 @@ The configuration can also override a workflow field with a <a href="#step">full
 
 {{ yamlSyntax (index . "configExample3") }}
 
+<h3 id="allow-skip-on-success"><a href="#allow-skip-on-success">Optional skipping of post steps</a></h3>
```
stevekuznetsov, line 1159: Nit: can we title-case this, make it a subsection and make the phrase more declarative?

```html
 <h2 id="allow-skip-on-success"><a href="#allow-skip-on-success">Options to Change Control Flow</a></h2>
```

pkg/webreg/webreg.go
```diff
@@ -1156,6 +1156,17 @@ The configuration can also override a workflow field with a <a href="#step">full
...
+<p>
+There is an option to skip arbitrary <code>post</code> steps when all <code>test</code>
```
stevekuznetsov, line 1161:
```
<code>ci-operator</code> can be configured to skip some or all <code>post</code> steps when all <code>test</code> steps pass.
```

pkg/webreg/webreg.go
```diff
@@ -1156,6 +1156,17 @@ The configuration can also override a workflow field with a <a href="#step">full
...
+steps pass. For example, it might be useful to skip gathering artifacts when all tests have passed and save
```
stevekuznetsov, line 1162:
```
Skipping a <code>post</code> step when all tests have passed may be useful to skip gathering artifacts and save some time ...
```

pkg/webreg/webreg.go
```diff
@@ -1156,6 +1156,17 @@ The configuration can also override a workflow field with a <a href="#step">full
 
 {{ yamlSyntax (index . "configExample3") }}
 
+<h3 id="allow-skip-on-success"><a href="#allow-skip-on-success">Optional skipping of post steps</a></h3>
+<p>
+There is an option to skip arbitrary <code>post</code> steps when all <code>test</code>
+steps pass. For example, it might be useful to skip gathering artifacts when all tests have passed and save 
+some time at the end of the multistage test. In order to skip a <code>post</code> step, the individual step
```
stevekuznetsov, line 1163: Can we put the ordering the other way?

```
In order to allow steps to be skipped in a test, the <code>allow_skip_on_success</code> field must be set in the<code>steps</code> configuration. Individual <code>post</code> steps opt into being skipped by setting the <code>optional_on_success</code> field.
```

pkg/webreg/webreg.go
```diff
@@ -1335,6 +1346,27 @@ const configExample3 = `tests:
         requests:
           cpu: 100m
           memory: 200Mi`
+const configExample4 = `tests:
+- as: e2e-steps # test name
+  steps:
+    allow_skip_on_success: true
```
stevekuznetsov, line 1352: Could you add a comment to this example here to call this field out?

```
# allows steps to be skipped in this test
```

pkg/webreg/webreg.go
mgencur: Do you want it exactly as you suggested? The "allow-skip-on-success" doesn't really match the title anymore which seems to be more general in your suggestion

mgencur: Anyway. Pushed. Thanks for reviews!

PR #919, opened 2020-06-17
autoowners: allow ignoring full orgs

petr-muller:
I'd like to set up `--ignore-org=openshift-priv` in our autoowners job so that the two jobs stop to flap OWNERS in `openshift-priv/*` repos:

- Here is autoowners creating them: https://github.com/openshift/release/pull/9718/files
- Here is autoconfigbrancher deleting them: https://github.com/openshift/release/pull/9720/files

We do not need OWNERS in `openshift-priv` directories (they are managed purely by automation).

I also took the opportunity to modernize the naming and call the thing `blocklist`.

/cc @hongkailiu @droslean

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -52,7 +52,7 @@ func (orgRepo orgRepo) repoString() string {
 	return fmt.Sprintf("%s/%s", orgRepo.Organization, orgRepo.Repository)
 }
 
-func loadRepos(configRootDir string, blacklist sets.String, configSubDirs []string, githubOrg string, githubRepo string) ([]orgRepo, error) {
+func loadRepos(configRootDir string, blocked blocklist, configSubDirs []string, githubOrg string, githubRepo string) ([]orgRepo, error) {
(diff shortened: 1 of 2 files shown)


cmd/autoowners/main.go
```diff
@@ -380,6 +386,11 @@ func listUpdatedDirectoriesFromGitStatusOutput(s string) ([]string, error) {
 	return directories, nil
 }
 
+type blocklist struct {
+	directories sets.String
```
droslean, line 390: since this represents a repository, let's call it `repo`

petr-muller: No, it's actually directories, not repos... See how it is used: 

https://github.com/openshift/release/blob/94291a3c8fab4ad6eede56341f56f0d7d4965e1e/ci-operator/jobs/infra-periodics.yaml#L635-L670

petr-muller: (I was also quite surprised, I originally called this `repositories`, too)

PR #933, opened 2020-06-19
Remove stale fixtures and add prefix to identify fixtures

alvaroaleman:
This pr removes some stale fixtures whose test doesn't exist anymore or
get renamed and changes the fixture code to prefix them. This will allow
us to add a check if there are fixtures for removed/renamed tests in the
repo to CI.


petr-muller:
:heart:

PR #936, opened 2020-06-22
Bugzilla: Create clones functionality for backporter

GeoBK:
This PR adds the functionality for the "Create Clone" button in the Bugzilla backporter tool.
The create clone button would be used to create a clone with a specific release.
It also adds a test to verify the output.


pkg/backporter/backporter.go
```diff
@@ -370,3 +413,80 @@ func GetClonesHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
 }
+
+// CreateCloneHandler will create a clone of the specified ID and return success/error
+func CreateCloneHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
+	return func(w http.ResponseWriter, req *http.Request) error {
+		req.ParseForm()
+		bugID, err := strconv.Atoi(req.FormValue("ID"))
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			wpErr := writePage(w, "Error", errorTemplate, fmt.Sprintf("Invalid bug id! - Bug#%d : %v", bugID, err))
```
alvaroaleman, line 424: Please don't use exclamation marks in messages, there is no reason to have the software yell at the user

pkg/backporter/backporter.go
```diff
@@ -162,22 +170,49 @@ const clonesTemplateConstructor = `
...
+<div class="alert alert-danger" id="error-banner">
+<a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
+<strong>Error! </strong> <label id ="error-text">{{.}}</label>
+</div>`
 var (
 	clonesTemplate = template.Must(template.New("clones").Parse(clonesTemplateConstructor))
 	emptyTemplate  = template.Must(template.New("empty").Parse("{{.}}"))
+	errorTemplate = template.Must(template.New("error").Parse(errorTemplateConstructor))
+	targetVersions = sets.NewString(
+		"4.7.0",
```
alvaroaleman, line 196: This needs to be configurable as it will change over time. Why did you add the `3.0` and `2.0`?

pkg/backporter/backporter.go
```diff
@@ -370,3 +413,80 @@ func GetClonesHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
+				return wpErr
+			}
+			return err
+		}
+		targetRelease := bugzilla.BugUpdate{
+			TargetRelease: []string{
+				req.FormValue("version"),
+			},
+		}
+		err = client.UpdateBug(cloneID, targetRelease)
```
alvaroaleman, line 459: Nit: Statements like this can be shortened:
```
if err := client.UpdateBug(cloneID, targetRelease); err != nil {
  handle Err
}
```

pkg/backporter/backporter.go
GeoBK: Those were test values - I havent finalized that part yet - I'll try to get them using flags like its being done somewhere upstream

cmd/bugzilla-backporter/main.go
```diff
@@ -131,9 +169,13 @@ func main() {
...
 	health := pjutil.NewHealth()
 	metrics.ExposeMetrics("ci-operator-bugzilla-backporter", prowConfig.PushGateway{}, prowflagutil.DefaultMetricsPort)
-
+	allTargetVersions, err := getAllTargetVersions(o.pluginConfig)
+	if err != nil {
+		log.WithError(err).Fatal("Error parsing plugins configuration.")
+	}
 	http.HandleFunc("/", handleWithMetrics(backporter.GetLandingHandler()))
-	http.HandleFunc("/getclones", handleWithMetrics(backporter.GetClonesHandler(bugzillaClient)))
+	http.HandleFunc("/getclones", handleWithMetrics(backporter.GetClonesHandler(bugzillaClient, allTargetVersions)))
```
stevekuznetsov, line 177: If we want this to be RESTful or RESTish let's consider

```
GET /clone/{id}
PUT|POST /clones/{id}
GET /bug/{id}
```

GeoBK: for PUT|POST /clones - would the ID being in the POST body be ok?

alvaroaleman: Yeah for create this makes sense but would hen be something like `/clones/create` as the URI

GeoBK: Alright, I have made that change

stevekuznetsov: Why not POST to `/clones` ? Why  would we need a separate URI?

GeoBK: So would that just be a POST to /clones - without passing an ID?

cmd/bugzilla-backporter/main.go
```diff
@@ -102,6 +108,36 @@ func gatherOptions() (options, error) {
...
+func getAllTargetVersions(configFile string) (sets.String, error) {
+	// Get the versions from the plugins.yaml file to populate the Target Versions dropdown
+	// for the CreateClone functionality
+	b, err := ioutil.ReadFile(configFile)
+	if err != nil {
+		return nil, err
+	}
+	np := &plugins.Configuration{}
+	if err := yaml.Unmarshal(b, np); err != nil {
+		return nil, err
```
alvaroaleman, line 120: Nit: Can you wrap errors rather then directly returning them so debugging becomes easier? Golang doesn't provide stracktraces, and using the wrapping helps to easely figure out where exactly an error happend.
Sample:
```
return nil, fmt.Errorf("failed to unmarshal yaml: %w")
```

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
 	}
 }
 
+func getClonesTemplateData(bugID int, w http.ResponseWriter, client bugzilla.Client, allTargetVersions sets.String) (*ClonesTemplateData, error) {
+	bug, err := client.GetBug(bugID)
+	if err != nil {
+		w.WriteHeader(http.StatusNotFound)
+		wpErr := writePage(w, "Not Found", errorTemplate, fmt.Sprintf("Bug#%d not found", bugID))
+		if wpErr != nil {
+			http.Error(w, "Error building page.", http.StatusInternalServerError)
```
alvaroaleman, line 313: Not sure if go catches this, but I doubt its valid to send the same header multiple times and you already send it on line 310.

Comment applies to all occurences of this.

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
+			return nil, wpErr
```
alvaroaleman, line 314: I think we want to get both errors here, not only the `writePage` one. This is usually done via `utilerrors`:
```
import utilerrors "k8s.io/apimachinery/pkg/util/errors"
return utilerrors.NewAggregate([]error{err, writePage(w, "Not Found", errorTemplate, fmt.Sprintf("Bug#%d not found", bugID))})
```

(it also filters out nil errors so this will work fine even when `writePage` doesn't return an error)

alvaroaleman: (The comment applies for all the occurences of handling `writeError` after handling other error)

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
+		}
+		return nil, err
+	}
+	parent, err := client.GetRootForClone(bug)
+	if err != nil {
+		w.WriteHeader(http.StatusInternalServerError)
+		wpErr := writePage(w, "Error.", errorTemplate, fmt.Sprintf("Bug#%d Details of parent could not be retrieved : %v", bug.ID, err))
```
alvaroaleman, line 331: tbh I find it a bit weird to have a page titled `Error.` rather than just `Error`? I don't think its common for titles anywhere to have trailing punctuation?

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
+		}
+		return nil, err
+	}
+	// Target versions would be used to populate the CreateClone dropdown
+	targetVersions := sets.NewString(allTargetVersions.List()...)
+	// Remove target versions of the original bug
+	for _, release := range bug.TargetRelease {
+		_, ok := targetVersions[release]
+		if ok {
+			targetVersions.Delete(release)
```
alvaroaleman, line 355: You can just unconditionally do a `targetVersion.Delete`, it will work fine even if the `release` is not in the targetVersions (will just do nothing in that case).

Comment applies to all occurences of this.

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
+	}
+}
+
 // GetClonesHandler returns an HTML page with detais about the bug and its clones
-func GetClonesHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
+func GetClonesHandler(client bugzilla.Client, allTargetVersions sets.String, w http.ResponseWriter, req *http.Request) error {
+	bugIDStr := req.URL.Query().Get(BugIDQuery)
+	if bugIDStr == "" {
+		w.WriteHeader(http.StatusBadRequest)
+		wpErr := writePage(w, "Error.", errorTemplate, fmt.Sprintf("%s - query incorrect", BugIDQuery))
```
alvaroaleman, line 410: Can we reword this to `missing mandatory query arg: BugIDQuery`? The current error message is confusing at best

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
 			w.WriteHeader(http.StatusBadRequest)
-			wpErr := writePage(w, "Error!", emptyTemplate, fmt.Sprintf("%s - query incorrect", BugIDQuery))
+			wpErr := writePage(w, "Error.", errorTemplate, fmt.Sprintf("%s - query incorrect", BugIDQuery))
 			if wpErr != nil {
-				http.Error(w, "Error building page!", http.StatusInternalServerError)
+				http.Error(w, "Error building page.", http.StatusInternalServerError)
 				return wpErr
 			}
-			return fmt.Errorf("%s - query incorrect", BugIDQuery)
+			return fmt.Errorf("Bad request - clones/create can handle only POST requests")
```
alvaroaleman, line 450: error messages never start capitalized unless its a name of something by convention

pkg/backporter/backporter.go
```diff
@@ -278,94 +304,218 @@ func GetBugHandler(client bugzilla.Client) HandlerFuncWithErrorReturn {
...
 			}
 			return err
 		}
-		bug, err := client.GetBug(bugID)
+		bugID, err := strconv.Atoi(req.FormValue("ID"))
 		if err != nil {
-			w.WriteHeader(http.StatusNotFound)
-			wpErr := writePage(w, "Not Found", emptyTemplate, fmt.Sprintf("Bug#%d not found", bugID))
+			w.WriteHeader(http.StatusBadRequest)
+			wpErr := writePage(w, "Error", errorTemplate, fmt.Sprintf("Invalid bug id. - Bug#%d : %v", bugID, err))
```
alvaroaleman, line 466: Please be helpful and tell how it was invalid. Here: Could not parse string as int. Which could be because it wasn't passed as query arg in which case this is even more confusing so maybe it would be a good idea to have a different error in case it wasn't sent

pkg/backporter/backporter_test.go
```diff
@@ -173,7 +145,17 @@ func TestGetClonesHandler(t *testing.T) {
 			},
 			ResCheck{
```
alvaroaleman, line 146: Nit: Tests are easier to write and to grok if they don't have nested attributes so in this case make everything that is part of `ResCheck` an attribute of the test case and then have three `if actualThing != tc.expectedThing { t.Errorf("actual thing does not match expected thing")}`. Golang will not abort the test if you use `Errorf` rather than `Fatalf` so the user will still see all issues in case multiple checks fail.

Also something like this type should not be shared among tests, as it results in the type being a superset of what all tests that use it need.

pkg/backporter/backporter_test.go
```diff
@@ -161,6 +129,10 @@ func TestGetClonesHandler(t *testing.T) {
 	if err != nil {
 		t.Errorf("Error while cloning bug: %v", err)
 	}
+	clone, err := fake.GetBug(cloneID)
+	if err != nil {
+		t.Errorf("Error retreiving clone: %v", err)
```
alvaroaleman, line 134: I don't think the remainder of the test is useful if we fail here, so using `t.Fatalf` to abort it here rather than `Errorf` seems appropriate

pkg/backporter/backporter_test.go
```diff
@@ -199,15 +182,137 @@ func TestGetClonesHandler(t *testing.T) {
...
 	}
 
 }
+
+func TestCreateCloneHandler(t *testing.T) {
+	fake := &bugzilla.Fake{}
+	fake.Bugs = map[int]bugzilla.Bug{}
+	fake.BugComments = map[int][]bugzilla.Comment{}
+
+	bug1Create := &bugzilla.BugCreate{
```
alvaroaleman, line 214: Can we give this a somewhat useful name instead of enumerating, like for example `toBeCloned`?

pkg/backporter/backporter_test.go
```diff
@@ -199,15 +182,137 @@ func TestGetClonesHandler(t *testing.T) {
...
+			req, err := http.NewRequest("POST", "/clones/create", bytes.NewBufferString(formData.Encode()))
+			req.Header.Set("Content-Type", "application/x-www-form-urlencoded; param=value")
+			if err != nil {
+				t.Fatal(err)
+			}
+			rr := httptest.NewRecorder()
+			handler := unwrapper(CreateCloneHandler(fake, allTargetVersions))
+			handler.ServeHTTP(rr, req)
+			if status := rr.Code; status != tc.results.statusCode {
+				t.Errorf("testcase '%v' failed: clonebug returned wrong status code - got %v, want %v", tc, status, tc.results.statusCode)
```
alvaroaleman, line 282: `Fatalf` IMHO

pkg/backporter/backporter.go
GeoBK: I think it should since we are setting it to 200 here (https://github.com/openshift/ci-tools/blob/master/cmd/bugzilla-backporter/main.go#L77) and the last applied header seems to take effect

pkg/backporter/backporter.go
GeoBK: Yeah, you are right

pkg/backporter/backporter_test.go
GeoBK: Alright I have moved all of them into the outer struct and gotten rid of the ResCheck class

pkg/backporter/backporter_test.go
GeoBK: Alright I have switched the Errorf s with Fatalf s where appropriate.

pkg/backporter/backporter_test.go
GeoBK: But since we evaluate multiple testcases in this loop, wouldn't we want it to just register the error and continue in this case?

pkg/backporter/backporter.go
GeoBK: I will use utilerrors, but I dont think it would be ok to put everything in a single statement like - 
return utilerrors.NewAggregate([]error{err, writePage(w, "Not Found", errorTemplate, fmt.Sprintf("Bug#%d not found", bugID))})
since I would want to return an error message back.
Or is that not necessary?

pkg/backporter/backporter.go
GeoBK: Although I do feel that Error! is more appropriate

cmd/bugzilla-backporter/main.go
GeoBK: Instead of wrapping with failed to umarshal would it be better to just put in the function name where it occurred if the objective is to figure out exactly where the error happened?

alvaroaleman: That is essentially what we do but we do it in the caller, not in the callee, otherwise you have to change all errors if you decide to move code from one function in another. Additionally its simpler, because if someone wrote wrapping like 
`failed to do http requst` here you will directly see that this is wrong whereas if the wrapping referred to the function it is in you need to remember the function name which is declared further up.

pkg/backporter/backporter.go
alvaroaleman: I think the traceResponseWriter only uses it for metrics because the golang http server will send a http 200 status if you do not call `WriterHeader`. Also:

```
go doc net/http.ResponseWriter.WriteHeader
[...]
// Only one header may be written.
[...]
```

And even if that wasn't the case I would argue that this must always be a 404. We do not want to confuse clients with a javalike `exception handling exception`, we could write that into the body but we should not mess with the http status if that happens

pkg/backporter/backporter.go
alvaroaleman: The `utilerrors.NewAggregate` returns an error which has a message that contains all non-nil errors or what do you mean?

pkg/backporter/backporter.go
alvaroaleman: That feels to me like windows 98 `Error! no work!`

pkg/backporter/backporter_test.go
alvaroaleman: So the way this works is that everytime you call `t.Run` you start a new (sub-) test. If you then call `t.Fatalf` you only stop the execution of that subtest, not of the parent or the silblings in the loop.

GeoBK: But in this case, I'm pretty sure I want to also check the response body even if the headers differ. Why do you feel otherwise?

pkg/backporter/backporter.go
alvaroaleman: Also looking at the `traceResponseWriter` it doesn't do what its supposed to do and will always increase the metric for the statusCode it was created with rather even if `WriteHeader` was called because it doesn't implement a `WriteHeader` wrapper like here: https://github.com/openshift/ci-tools/blob/2f2c559688e4a27349bae36bf1e38253f52930cc/cmd/ci-operator-configresolver/main.go#L48

Could you please create a different PR that moves the `traceResponseWriter` from `cmd/ci-operator-configresolver` into `pkg/httphelper` and re-uses that in both the `ci-operator-configresolver` and the `bugzilla-backporter` so we don't end up with duplicated code?

GeoBK: Ok

pkg/backporter/backporter_test.go
alvaroaleman: Because the header indicates which codepath we took (e.G. success/failure) and if that is wrong already we can very strongly assume that the response body will be wrong as well

GeoBK: True... but it could also just be that the header passed is a 404 when what was expected was a 500 and that doesn't necessarily call for aborting the tc

alvaroaleman: its not a strong opinion one way or another

pkg/backporter/backporter.go
GeoBK: What I meant was that I wouldn't be able to write the return errors as a single line(like the example you have given), since in the case of an error occurring I would have to write an error message to the body of the response.

GeoBK:
Rewrote the error handling mechanism - so should be ready for review. I'll add in the errors for prometheus once the httphelper package is checked in

pkg/backporter/backporter.go
```diff
@@ -159,214 +168,264 @@ const clonesTemplateConstructor = `
...
+			errorPageHandler(w, err, fmt.Sprintf("clone created - Bug#%d, but failed to specify version for the cloned bug.", cloneID), http.StatusInternalServerError, endpoint, bugID)
+			return
 		}
-
-		wpErr := writePage(w, "Clones", clonesTemplate, wrpr)
-		if wpErr != nil {
-			http.Error(w, "Error building page!", http.StatusInternalServerError)
-			return wpErr
+		// Repopulate the fields of the page with the right data
+		data, err := getClonesTemplateData(bugID, w, client, allTargetVersions, endpoint)
```
alvaroaleman, line 423: err no:
* A function called as part of a handler should not do its own error handling, that is confusing. Error handling is always done by the caller.
* There are very, very few reasons to ever do an `if err != nil { return }`, all of them would require an explaining comment

pkg/backporter/backporter_test.go
```diff
@@ -135,79 +97,207 @@ func TestGetBugHandler(t *testing.T) {
...
-		params  map[string]int
-		results ResCheck
+		name       string
+		params     map[string]int
+		statusCode int
+		data       interface{}
+		tmplt      *template.Template
 	}{
 		{
 			"valid_parameters",
```
alvaroaleman, line 133: Could you please use named arguments, this is pretty hard to read?
```
name: "valid_parameters",
params: map[string]int{"ID": cloneID},
...
```

pkg/backporter/backporter_test.go
```diff
@@ -135,79 +97,207 @@ func TestGetBugHandler(t *testing.T) {
...
+	expectedCloneID := toBeClonedID + 1
+	testcases := []struct {
+		name       string
+		params     map[string]string
+		statusCode int
+		data       interface{}
+		tmplt      *template.Template
+	}{
+		{
+			"valid_parameters",
```
alvaroaleman, line 226: same comment regarding the naming of the properties

pkg/backporter/backporter_test.go
```diff
@@ -135,79 +97,207 @@ func TestGetBugHandler(t *testing.T) {
...
+
+			var buf bytes.Buffer
+			var pageStart string
+			if tc.statusCode == http.StatusOK {
+				pageStart = fmt.Sprintf(htmlPageStart, "Clones")
+				newClone, err := fake.GetBug(expectedCloneID)
+				if err != nil {
+					t.Fatalf("error while fetching clone details from mocked endpoint")
+				}
+				newClone.TargetRelease = []string{
```
alvaroaleman, line 278: All properties of `newClone` are deterministic or not? If yes, can we just put it into the `templateData` and not mutate it during test execution by using a type assertion along the way?

pkg/backporter/backporter_test.go
```diff
@@ -135,79 +97,207 @@ func TestGetBugHandler(t *testing.T) {
...
+				data, ok := tc.data.(ClonesTemplateData)
+				if ok {
+					data.Clones = []*bugzilla.Bug{
+						newClone,
+					}
+				}
+				tc.data = data
+
+			} else {
+				pageStart = fmt.Sprintf(htmlPageStart, "Not Found")
```
alvaroaleman, line 290: This should be part of the testcase itself and not derived from some other property of the test during runtime

cmd/bugzilla-backporter/main.go
```diff
@@ -102,40 +100,76 @@ func gatherOptions() (options, error) {
...
+		return nil, fmt.Errorf("failed to validate file %s : %v", configFile, err)
+	}
+	allTargetVersions := sets.NewString()
+	// Hardcoding with just the "openshift" org here
+	// Could be extended to be configurable in the future to support multiple products
+	// In which case this would have to be moved to the CreateClones function.
+	options := np.Bugzilla.OptionsForRepo("openshift", "")
+	for _, val := range options {
+		if val.TargetRelease != nil {
+			if _, ok := allTargetVersions[*val.TargetRelease]; !ok {
```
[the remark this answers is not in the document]
GeoBK: OK, removed the check

pkg/backporter/backporter.go
```diff
@@ -159,214 +168,264 @@ const clonesTemplateConstructor = `
...
 
+func leanErrorHandler(w http.ResponseWriter, err error, shortErrorMessage string, statusCode int, endpoint string, bugID int) {
+	w.WriteHeader(statusCode)
+	_, fprintfErr := fmt.Fprintf(w, "%s: %v", shortErrorMessage, err)
+	logrus.WithFields(logFieldsFor(endpoint, bugID)).WithError(fmt.Errorf("%s: %v", shortErrorMessage, utilerrors.NewAggregate([]error{err, fprintfErr})))
+}
+
+func errorPageHandler(w http.ResponseWriter, err error, shortErrorMessage string, statusCode int, endpoint string, bugID int) {
+	w.WriteHeader(statusCode)
+	var wpErrWrapper, innerErrWrapper error
```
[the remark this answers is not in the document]
GeoBK: No they aren't, sorry removing them

pkg/backporter/backporter.go
```diff
@@ -159,214 +168,264 @@ const clonesTemplateConstructor = `
...
+	errorTemplate  = template.Must(template.New("error").Parse(errorTemplateConstructor))
 )
 
+func leanErrorHandler(w http.ResponseWriter, err error, shortErrorMessage string, statusCode int, endpoint string, bugID int) {
+	w.WriteHeader(statusCode)
+	_, fprintfErr := fmt.Fprintf(w, "%s: %v", shortErrorMessage, err)
+	logrus.WithFields(logFieldsFor(endpoint, bugID)).WithError(fmt.Errorf("%s: %v", shortErrorMessage, utilerrors.NewAggregate([]error{err, fprintfErr})))
+}
+
+func errorPageHandler(w http.ResponseWriter, err error, shortErrorMessage string, statusCode int, endpoint string, bugID int) {
```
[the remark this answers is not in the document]
GeoBK: I just ended up reusing the lean handler in  writePage since it did what most of what I wanted there, but I have removed its use from the errorPageHandler and have gone with your second suggestion - ie make writePage return an error.

The lean handler is primarily meant to handle the errors for the getBugHanlder which returns a simple JSON and its error were also intended to be similarly simple while the errorPageHandler returns an entire error page and serves a different purpose I feel. Do you still think one needs to be removed?

I'll rename the functions.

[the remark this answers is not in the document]
alvaroaleman: Yeah I think one of them should be removed as its not clear which to use when. I don't think it is very important if we have a nice error page or just the error text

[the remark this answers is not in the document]
GeoBK: Ok. Done

cmd/bugzilla-backporter/main.go
```diff
@@ -92,6 +89,7 @@ func gatherOptions() (options, error) {
 	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
 	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
 	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
+	fs.StringVar(&o.pluginConfig, "plugin-config", "/etc/plugins/plugins.yaml", "Path to plugin config file.")
```
alvaroaleman, line 92: Can you create a pr against openshift/release that adds a mount for this so we can merge this pr and the required config change together?

pkg/backporter/backporter.go
```diff
@@ -159,214 +168,271 @@ const clonesTemplateConstructor = `
...
+		return nil, http.StatusInternalServerError, fmt.Errorf("unable to get clones: %w", err)
+	}
+	// Target versions would be used to populate the CreateClone dropdown
+	targetVersions := sets.NewString(allTargetVersions.List()...)
+	// Remove target versions of the original bug
+	targetVersions.Delete(bug.TargetRelease...)
+	for _, clone := range clones {
+		clonePRs, err := client.GetExternalBugPRsOnBug(clone.ID)
+		// Remove target releases which already have clones
+		targetVersions.Delete(clone.TargetRelease...)
```
alvaroaleman, line 328: always handle error first, this code reads as if the error was unhandled

pkg/backporter/backporter.go
```diff
@@ -159,214 +168,271 @@ const clonesTemplateConstructor = `
...
+		PRs:          prs,
+		CloneTargets: targetVersions.List(),
+	}
+	return &wrpr, http.StatusOK, nil
+}
+
+// ClonesHandler acts as a router for the RESTish calls to clones
+func ClonesHandler(client bugzilla.Client, allTargetVersions sets.String) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != "GET" {
```
alvaroaleman, line 348: Can we move this check into `GetClonesHandler`? Why is it here, are we expecting to handle more verbs here in the future?

cmd/bugzilla-backporter/main.go
GeoBK: I created this PR for that, is this what you meant? https://github.com/openshift/release/pull/10021

pkg/backporter/backporter.go
GeoBK: I had added it for the POST clones which steve had mentioned and left it in there if we were going to move it back. I will remove it since we are sticking with /clones/create

PR #937, opened 2020-06-22
Enable unconvert and errcheck linter

alvaroaleman:
* The unconvert linter checks for unnecessary conversions
* The errcheck linter verifies that all errors are assigned

### DIFF
--- a/cmd/ci-operator-config-mirror/main.go
+++ b/cmd/ci-operator-config-mirror/main.go
@@ -33,7 +33,7 @@ type options struct {
 	clean bool
 }
 
-func gatherOptions() options {
+func gatherOptions() (options, error) {
 	o := options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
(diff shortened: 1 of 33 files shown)


cmd/ci-operator-config-mirror/main.go
```diff
@@ -44,8 +44,10 @@ func gatherOptions() options {
 	fs.BoolVar(&o.clean, "clean", true, "If the `to-org` folder already exists, then delete all subdirectories")
 
 	o.WhitelistOptions.Bind(fs)
-	fs.Parse(os.Args[1:])
-	return o
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		return o, fmt.Errorf("failed to parse flags: %w", err)
+	}
+	return o, nil
```
petr-muller, line 50: I think these are the one we can actually easily ignore because of the `flag.ExitOnError` -- this would make even the callsites simpler.

petr-muller: But I do not insist: if you feel like you already spent enough time on this I'll ack this.

alvaroaleman: I would prefer that, yeah

petr-muller: :+1:

PR #940, opened 2020-06-24
Bugzilla: Add dockerfile for backporter deployment

GeoBK:


images/bugzilla-backporter/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM centos:7
+LABEL maintainer="skuznets@redhat.com"
+
+ADD bugzilla-backporter /usr/bin/bugzilla-backporter
+RUN yum -y install backporter && yum clean all && rm -rf /var/cache/yum
```
GeoBK, line 5: I am trying to install the package which the tool depends on, not sure its necessary though - since I am not installing any of the other dependencies.

droslean: @GeoBK Can you elaborate more in why the tool needs this `backporter` package?

GeoBK: I am not really sure its needed, I was mimicing the format in another file which installed graphviz but its only now that I realize that it was probably an external package it was installing.

GeoBK: @droslean Does it look ok now?

images/bugzilla-backporter/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM centos:7
```
GeoBK, line 1: I also noticed that we use both 7 and 8. Is there a preferred version?

alvaroaleman: Yeah, please use 8. It doesn't make a difference, because it will get replaced with https://github.com/openshift/release/pull/9880/files#diff-2aa47366feea3a8e22a64b1e867062adR62 which resolves to centos:8 but it makes it easier to understand

PR #941, opened 2020-06-24
ci-operator configs: omitempty images.from

petr-muller:
Empty `from` means we do not substitute a `FROM` in the repo's
Dockerfile, which makes sense in some cases.

Hit while building an optional operator bundle image which is built from
scratch.


petr-muller:
Followup in openshift/release: https://github.com/openshift/release/pull/9890

PR #942, opened 2020-06-24
Add registry-replacer

alvaroaleman:
Ref: https://issues.redhat.com/browse/DPTP-1264?focusedCommentId=14174826&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14174826

Could maybe be updated to also create a PR already?

### DIFF
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
-	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
@@ -35,14 +34,18 @@ func main() {
 		*configDir,
 		func(config *api.ReleaseBuildConfiguration, info *config.Info) error {
 			wg.Add(1)
-			go func() {
+			go func(filename string) {
 				defer wg.Done()
-				if err := replacer(githubFileGetterFactory, ioutil.WriteFile)(config, info); err != nil {
+				if err := replacer(
+					githubFileGetterFactory,
+					func(data []byte) error {
+						return ioutil.WriteFile(filename, data, 0644)
+					})(config, info); err != nil {
 					errLock.Lock()
 					errs = append(errs, err)
 					errLock.Unlock()
 				}
-			}()
+			}(info.Filename)
 			return nil
 		},
 	); err != nil {
@@ -59,7 +62,7 @@ func main() {
 // bounds.
 func replacer(
 	githubFileGetterFactory func(org, repo, branch string) githubFileGetter,
-	writer func(name string, data []byte, mode os.FileMode) error,
+	writer func([]byte) error,
 ) func(*api.ReleaseBuildConfiguration, *config.Info) error {
 	return func(config *api.ReleaseBuildConfiguration, info *config.Info) error {
 		if len(config.Images) == 0 {
@@ -98,7 +101,7 @@ func replacer(
 			return nil
 		}
 
-		if err := writer(info.Filename, newConfig, 0644); err != nil {
+		if err := writer(newConfig); err != nil {
 			return fmt.Errorf("faild to write %s: %w", info.Filename, err)
 		}
 
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -83,6 +83,9 @@ func replacer(
 				if config.BaseImages == nil {
 					config.BaseImages = map[string]api.ImageStreamTagReference{}
 				}
+				if _, exists := config.BaseImages[foundTag.String()]; exists {
+					continue
+				}
 				config.BaseImages[foundTag.String()] = api.ImageStreamTagReference{
 					Namespace: foundTag.org,
 					Name:      foundTag.repo,
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -61,7 +61,7 @@ func replacer(
 	writer func(name string, data []byte, mode os.FileMode) error,
 ) func(*api.ReleaseBuildConfiguration, *config.Info) error {
 	return func(config *api.ReleaseBuildConfiguration, info *config.Info) error {
-		if !referencesDockerfile(config.Images) {
+		if len(config.Images) == 0 {
 			return nil
 		}
 
@@ -105,16 +105,6 @@ func replacer(
 	}
 }
 
-func referencesDockerfile(images []api.ProjectDirectoryImageBuildStepConfiguration) bool {
-	for _, image := range images {
-		if image.DockerfilePath != "" {
-			return true
-		}
-	}
-
-	return false
-}
-
 var registryRegex = regexp.MustCompile(`registry\.svc\.ci\.openshift\.org\/[^\s]+`)
 
 type orgRepoTag struct{ org, repo, tag string }
@@ -124,11 +114,12 @@ func (ort orgRepoTag) String() string {
 }
 
 func ensureReplacement(image *api.ProjectDirectoryImageBuildStepConfiguration, getter githubFileGetter) ([]orgRepoTag, error) {
-	if image.DockerfilePath == "" {
-		return nil, nil
+	dockerFilePath := "Dockerfile"
+	if image.DockerfilePath != "" {
+		dockerFilePath = image.DockerfilePath
 	}
 
-	data, err := getter(image.DockerfilePath)
+	data, err := getter(dockerFilePath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get dockerfile %s: %w", image.DockerfilePath, err)
 	}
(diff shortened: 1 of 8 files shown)


alvaroaleman:
/retest

The integration test flakes are yet another race it seems

stevekuznetsov:
> The integration test flakes are yet another race it seems

@alvaroaleman a link to a JIRA tracking this with data to investigate, please?

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+	configDir := flag.String("config-dir", "", "The directory with the ci-operator configs")
+	flag.Parse()
+	if *configDir == "" {
+		logrus.Fatal("--config-dir arg is required")
+	}
+
+	var errs []error
+	errLock := &sync.Mutex{}
+	wg := sync.WaitGroup{}
+	if err := config.OperateOnCIOperatorConfigDir(
```
stevekuznetsov, line 33: Does this need a concurrency limit _a la_ the other tools with ulimit or whatever

alvaroaleman:
> a link to a JIRA tracking this with data to investigate, please

Ill just do it now

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+		logrus.WithError(err).Fatal("Encountered errors")
+	}
+}
+
+// replacer ensures replace directives are in place. It fetches the files via http because using git
+// en masse easily kills a developer laptop whereas the http calls are cheap and can be parallelized without
+// bounds.
+func replacer(
+	githubFileGetterFactory func(org, repo, branch string) githubFileGetter,
+	writer func(name string, data []byte, mode os.FileMode) error,
```
stevekuznetsov, line 61: Could this just be an `io.Writer` since we know at the call-site what the name is, and the mode is static?

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+			return fmt.Errorf("faild to write %s: %w", info.Filename, err)
+		}
+
+		return nil
+	}
+}
+
+func referencesDockerfile(images []api.ProjectDirectoryImageBuildStepConfiguration) bool {
+	for _, image := range images {
+		if image.DockerfilePath != "" {
```
stevekuznetsov, line 110: If it's `""` I think it implies `./Dockerfile`

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+
+		if err := writer(info.Filename, newConfig, 0644); err != nil {
+			return fmt.Errorf("faild to write %s: %w", info.Filename, err)
+		}
+
+		return nil
+	}
+}
+
+func referencesDockerfile(images []api.ProjectDirectoryImageBuildStepConfiguration) bool {
```
stevekuznetsov, line 108: `context_dir` also should be handled here, as that references `context_dir/Dockerfile`

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get dockerfile %s: %w", image.DockerfilePath, err)
+	}
+
+	var toReplace []string
+	for _, line := range bytes.Split(data, []byte("\n")) {
+		if !bytes.Contains(line, []byte("FROM")) {
+			continue
+		}
+		if !bytes.Contains(line, []byte("registry.svc.ci.openshift.org")) {
```
stevekuznetsov, line 141: Is this an optimization to not use regex?

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+		inputs.As = sets.NewString(inputs.As...).Insert(toReplace).List()
+		image.Inputs[orgRepoTag.String()] = inputs
+
+		result = append(result, orgRepoTag)
+	}
+
+	return result, nil
+}
+
+func orgRepoTagFromPullString(pullString string) (orgRepoTag, error) {
```
stevekuznetsov, line 171: I imagine there's a canonical utility for this somewhere but it's simple enough so if we can't find one quickly let's keep our own impl

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,210 @@
...
+	for _, toReplace := range toReplace {
+		orgRepoTag, err := orgRepoTagFromPullString(toReplace)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse string %s as pullspec: %w", toReplace, err)
+		}
+		if image.Inputs == nil {
+			image.Inputs = map[string]api.ImageBuildInputs{}
+		}
+		inputs := image.Inputs[orgRepoTag.String()]
+		inputs.As = sets.NewString(inputs.As...).Insert(toReplace).List()
```
stevekuznetsov, line 162: I imagine for the first pass we might want a non-destructive mode where we don't change an input if it's already set, and in general we dont want to have multiple items in `inputs` that replace the same string (that is, `inputs[*].as[*]` is a unique set)

cmd/registry-replacer/main.go
alvaroaleman: No, the `io.Writer` has no notion of a filename which we need to pass

cmd/registry-replacer/main.go
alvaroaleman: Done

cmd/registry-replacer/main.go
alvaroaleman: I prefer increasing my users ulimit and keeping the tool fast :)

cmd/registry-replacer/main.go
alvaroaleman: Done

cmd/registry-replacer/main.go
alvaroaleman: Yes

cmd/registry-replacer/main.go
alvaroaleman: > I  imagine for the first pass we might want a non-destructive mode where we don't change an input if it's already set

Isn't that the case already since we use the set?

> and in general we dont want to have multiple items in inputs that replace the same string (that is, inputs[*].as[*] is a unique set)

Well if we want that we need validation to that effect which I believe we don't have today

cmd/registry-replacer/main.go
stevekuznetsov: I thought you do know the filename anyway, so you could close over it?

cmd/registry-replacer/main.go
stevekuznetsov: I think the output of this tool could make indeterminate config if we don't have that sort of validation

stevekuznetsov: To be clear I mean that the total set of items in `as` for any input should not overlap

alvaroaleman: I see. I have updated the tool to check for an existing replacement and if found, not create a new one.

cmd/registry-replacer/main.go
alvaroaleman: Oh right, updated it

cmd/registry-replacer/main.go
```diff
@@ -0,0 +1,221 @@
...
+		for idx := range config.Images {
+			foundTags, err := ensureReplacement(&config.Images[idx], githubFileGetterFactory(info.Org, info.Repo, info.Branch))
+			if err != nil {
+				return fmt.Errorf("failed to ensure replacements: %w", err)
+			}
+			for _, foundTag := range foundTags {
+				if config.BaseImages == nil {
+					config.BaseImages = map[string]api.ImageStreamTagReference{}
+				}
+				config.BaseImages[foundTag.String()] = api.ImageStreamTagReference{
```
stevekuznetsov, line 86: Very unlikely but we should also guard against changing this if it's already set somehow. I think we will want the "overwrite existing config to do the right thing" to be an opt-in flag that we will turn on once we alert everyone we're doing it, but before that happens we should make sure we're not ever making a change that could change user's config outside of what we're already doing.

stevekuznetsov:
So we have two classes of user bugs today:

 - an image replacement for `ocp/builder` is configured, matches the `FROM` field in the Dockerfile correctly, and incorrectly changes the Go toolchain and potentially the Go version
 - an image replacement for `ocp/builder` is configured, does not match the `FROM` field in the Dockerfile correctly, and does nothing, making the productized build run with the real builder image

The first can be fixed by removing the input that covers the item we're looking for and replacing it with ours. The latter requires us to search in the opposite direction - find all references in the Dockerfile and remove inputs that do not refer to anything valid.

I don't think either should be fixed by default in this tool today but both will be things we want to do in the future so let's just make sure we're making it easy to do in the future.

cmd/registry-replacer/main.go
alvaroaleman: Updated

PR #944, opened 2020-06-25
Fix pj-reherase integration flakes

alvaroaleman:
/assign @petr-muller 
Should I create a jira task to clean up the tests?


test/integration/pj-rehearse/expected.yaml
```diff
@@ -131,151 +131,6 @@
...
-      vars-origin.yaml
-    vars.yaml: |
-      changed
-  metadata:
-    creationTimestamp: null
-    labels:
-      ci.openshift.org/rehearse-pull: "1234"
-      created-by-pj-rehearse: "true"
-    name: rehearse-cluster-profile-test-profile-47224c86
-    namespace: test-namespace
```
petr-muller, line 278: Do I understand correctly that this is no longer output because dry-run is now always "interfacing" with a single cluster, and this was only output because in the past there were more?

alvaroaleman: Yup, exactly. Because I parallelized the config map creation it started to become indeterministic (in most but not all cases the goroutine we start first finishes first). Furthermore since we don't even keep the info on which cluster the thing gets created (its in all cases a client that just prints the serialized object) its also pretty useless.

PR #952, opened 2020-06-25
ci-operator: DRY out the relationships between step links and env

stevekuznetsov:
The logic for interoperating between environment variables and step
links is currently not common, shared or tested. Before we attempt to
expose such a mechanism to users in config, we need to refactor this to
be self-consistent.

### DIFF
--- a/cmd/cvp-trigger/main.go
+++ b/cmd/cvp-trigger/main.go
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/steps/utils"
 	"os"
 	"path/filepath"
 	"strings"
@@ -210,7 +212,7 @@ func main() {
 		steps.OOChannel: o.channel,
 	}
 	if o.releaseImageRef != "" {
-		envVars[steps.LatestReleaseEnv] = o.releaseImageRef
+		envVars[utils.ReleaseImageEnv(api.LatestStableName)] = o.releaseImageRef
 	}
 	if o.installNamespace != "" {
 		envVars[steps.OOInstallNamespace] = o.installNamespace
--- a/pkg/api/graph.go
+++ b/pkg/api/graph.go
@@ -3,6 +3,7 @@ package api
 import (
 	"context"
 	"fmt"
+	"github.com/google/go-cmp/cmp"
 	"strings"
 )
 
@@ -19,7 +20,7 @@ type Step interface {
 	Description() string
 	Requires() []StepLink
 	Creates() []StepLink
-	Provides() (ParameterMap, StepLink)
+	Provides() ParameterMap
 }
 
 type InputDefinition []string
@@ -165,6 +166,14 @@ func StableImageTagLink(name, tag string) StepLink {
 		name: StableStreamFor(name),
 		tag:  tag,
 	}
+}
+
+func Comparer() cmp.Option {
+	return cmp.AllowUnexported(
+		internalImageStreamLink{},
+		internalImageStreamTagLink{},
+		externalImageLink{},
+	)
 }
 
 func StableStreamFor(name string) string {
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -19,6 +19,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/junit"
 	"github.com/openshift/ci-tools/pkg/results"
+	"github.com/openshift/ci-tools/pkg/steps/utils"
 )
 
 const (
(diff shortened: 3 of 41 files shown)


pkg/api/graph.go
```diff
@@ -19,13 +20,19 @@ type Step interface {
...
 	Creates() []StepLink
-	Provides() (ParameterMap, StepLink)
+	Provides() ParameterMap
 }
 
 type InputDefinition []string
 
 type ParameterMap map[string]func() (string, error)
 
+func (m ParameterMap) AddFrom(other ParameterMap) {
```
stevekuznetsov, line 30: This is unused, will remove

pkg/api/graph.go
```diff
@@ -167,6 +174,14 @@ func StableImageTagLink(name, tag string) StepLink {
 	}
 }
 
+func Comparer() cmp.Option {
```
stevekuznetsov, line 177: I can see an argument for the functions in the env utils package to live in here, so we don't need to export this, WDYT

pkg/steps/multi_stage.go
```diff
@@ -192,7 +187,9 @@ func (s *multiStageTestStep) Requires() (ret []api.StepLink) {
 	if s.profile != "" {
 		needsReleasePayload = true
 		for _, env := range envForProfile {
-			ret = append(ret, s.params.Links(env)...)
+			if link, ok := utils.LinkForEnv(env); ok {
```
stevekuznetsov, line 190: Something here is broken

pkg/steps/output_image_tag_test.go
```diff
@@ -38,8 +38,7 @@ func TestOutputImageStep(t *testing.T) {
 			api.InternalImageLink("configToAs"),
 		},
 		provides: providesExpectation{
-			params: map[string]string{"IMAGE_CONFIGTOAS": "uri://somewhere:configToTag"},
-			link:   api.ExternalImageLink(config.To),
+			params: map[string]string{"IMAGE_CONFIGTOAS": "uri://somewhere@fromImageName"},
```
stevekuznetsov, line 41: I don't think the mock actually updates the spec for the ImageStream when we create the ImageStreamTag, and we can always get away with reference by digest instead of floating tag, so this is both a semi-broken test but also a change in behavior to share the func we use to resolve digests

pkg/steps/pipeline_image_cache.go
```diff
@@ -61,27 +58,13 @@ func (s *pipelineImageCacheStep) Creates() []api.StepLink {
...
 }
 
-func (s *pipelineImageCacheStep) Provides() (api.ParameterMap, api.StepLink) {
+func (s *pipelineImageCacheStep) Provides() api.ParameterMap {
 	if len(s.config.To) == 0 {
-		return nil, nil
+		return nil
 	}
 	return api.ParameterMap{
-		fmt.Sprintf("LOCAL_IMAGE_%s", strings.ToUpper(strings.Replace(string(s.config.To), "-", "_", -1))): func() (string, error) {
```
stevekuznetsov, line 69: I don't understand why this was the _only_ place we exposed a `$LOCAL_IMAGE...`

stevekuznetsov: e.g. not RPMs, etc...

pkg/steps/rpm_server.go
```diff
@@ -381,14 +381,14 @@ func (s *rpmServerStep) rpmRepoURL() (string, error) {
 	return fmt.Sprintf("http://%s", host), nil
 }
 
-func (s *rpmServerStep) Provides() (api.ParameterMap, api.StepLink) {
+func (s *rpmServerStep) Provides() api.ParameterMap {
 	if s.jobSpec.Refs != nil {
 		rpmByOrgAndRepo := strings.Replace(fmt.Sprintf("RPM_REPO_%s_%s", strings.ToUpper(s.jobSpec.Refs.Org), strings.ToUpper(s.jobSpec.Refs.Repo)), "-", "_", -1)
 		return api.ParameterMap{
 			rpmByOrgAndRepo: s.rpmRepoURL,
-		}, api.RPMRepoLink()
```
stevekuznetsov, line 389: I couldn't find any templates using `$RPM_REPO_foo` that weren't hard-coding it

pkg/steps/write_params.go
```diff
@@ -55,15 +55,15 @@ func (s *writeParametersStep) run(dry bool) error {
 }
 
 func (s *writeParametersStep) Requires() []api.StepLink {
-	return s.params.AllLinks()
+	return []api.StepLink{api.AllStepsLink()}
```
stevekuznetsov, line 58: I have no idea WTF this step is, if it's ever actually used, or what. This is technically a change in behavior but should be safe?

pkg/api/graph.go
petr-muller: I think I prefer the current state. Exporting a `Comparer` is not a big deal, and the current layout makes sense to me.

cmd/cvp-trigger/main.go
```diff
@@ -4,6 +4,8 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/steps/utils"
```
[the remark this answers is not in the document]
stevekuznetsov: amazing, new features

pkg/steps/write_params.go
stevekuznetsov: Yeah I think this is a :clayton:

stevekuznetsov:
@petr-muller @bbguimaraes this is ready

PR #954, opened 2020-06-26
Boskos: Directly use app.ci rather than going via api.ci tunnel

alvaroaleman:


pkg/api/domain_test.go
```diff
@@ -0,0 +1,28 @@
+package api
+
+import (
+	"testing"
+)
+
+func TestDomainForService(t *testing.T) {
```
alvaroaleman, line 7: I know I am paranoid :)

petr-muller: I do this too! Mostly so that when someone needs to extend the trivial method or adds a new one, they find an existing file with existing tests. I found people much likely to add new tests that way.

PR #956, opened 2020-06-26
Avoid no-op changes in integration test fixtures

alvaroaleman:
Currently we filter out certain things during diff time. This PR changes
that to instead replace those fields with a static string before writing
the fixtures in order to reduce review overhead for no-op changes.

### DIFF
--- a/hack/lib/integration.sh
+++ b/hack/lib/integration.sh
@@ -25,9 +25,17 @@ function os::integration::compare() {
         fi
     fi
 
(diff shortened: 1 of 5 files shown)


hack/lib/integration.sh
```diff
@@ -25,7 +25,7 @@ function os::integration::compare() {
         fi
     fi
 
-    os::cmd::expect_success "diff -Naupr --ignore-matching-lines 'startTime' --ignore-matching-lines 'name: \w\{8\}\(-\w\{4\}\)\{3\}-\w\{12\}' --ignore-matching-lines 'sha: \w\{40\}' ${actual} ${expected}"
```
stevekuznetsov, line 28: Can you do the replacement in this function so it's not duped everywhere? It can still be written to do the substitution before committing the diff to keep the on-disk representation static.

PR #957, opened 2020-06-26
Prowgen should handle duplicate targets

smarterclayton:
If additionalImages is specified where two output names exist for
the same image, then prowgen can generate duplicate targets which
causes an error on ci-operator parsing (`--target=foo --target=foo`
is invalid).

Fix the generator so that

```
promotion:
  additional_images:
    base: base-7
    base-7: base-7
    base-8: base-8
```

does not generate

```
- --target=base-7
- --target=base-7
- --target=base-8
```

Blocks openshift/release#9949 which needs to publish `base`, `base-7`, and `base-8` from `base-7` and `base-8`

@stevekuznetsov

### DIFF
--- a/pkg/prowgen/prowgen_test.go
+++ b/pkg/prowgen/prowgen_test.go
@@ -253,6 +253,7 @@ func TestGeneratePostSubmitForTest(t *testing.T) {
 func TestGenerateJobs(t *testing.T) {
 	tests := []struct {
 		id       string
+		keep     bool
 		config   *ciop.ReleaseBuildConfiguration
 		repoInfo *ProwgenInfo
 	}{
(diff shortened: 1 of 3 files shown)


pkg/prowgen/prowgen_test.go
```diff
@@ -349,7 +371,9 @@ func TestGenerateJobs(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.id, func(t *testing.T) {
 			jobConfig := GenerateJobs(tc.config, tc.repoInfo, jobconfig.Generated)
-			pruneForTests(jobConfig) // prune the fields that are tested in TestGeneratePre/PostsubmitForTest
+			if !tc.keep {
```
stevekuznetsov, line 374: This smells. IIRC this means a lower-level test handles those specifics.

smarterclayton: The higher level test doesn’t test that pod specs are valid.  So neither unit tests the influence of pod spec.

PR #958, opened 2020-06-27
pkg/steps/artifacts: Wait for running artifacts container

wking:
Avoid races where a slow artifacts container:

```
2020-06-26T20:33:55Z origin-ci-ig-n-2kf1 spec.containers{test} Created container
2020-06-26T20:33:55Z origin-ci-ig-n-2kf1 spec.containers{test} Started container
2020-06-26T20:33:56Z origin-ci-ig-n-2kf1 spec.containers{artifacts} Created container
2020-06-26T20:33:56Z origin-ci-ig-n-2kf1 spec.containers{artifacts} Started container
```

Leads to ci-operator noticing 'test' complete and then failing to kill the not-yet-running 'artifacts' container:

```
2020/06/26 20:33:55 error: unable to signal to artifacts container to terminate in pod endurance-install-aws, triggering deletion: could not run remote command: unable to upgrade connection: container not found ("artifacts")
2020/06/26 20:33:55 error: unable to retrieve artifacts from pod endurance-install-aws: could not read gzipped artifacts: unable to upgrade connection: container not found ("artifacts")
2020/06/26 20:33:56 Container artifacts in pod endurance-install-aws completed successfully
2020/06/26 20:33:56 Container test in pod endurance-install-aws completed successfully
```

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/client-go/kubernetes/scheme"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -214,6 +215,28 @@ func (c podClient) Exec(namespace, pod string, opts *coreapi.PodExecOptions) (re
 type PodClient interface {
 	coreclientset.PodsGetter
 	podCmdExecutor
+}
+
+func waitForContainer(podClient PodClient, ns, name, containerName string) error {
+	logrus.Tracef("Waiting for container %s in pod %s in namespace %s", containerName, name, ns)
+	return wait.PollImmediateInfinite(time.Second, func() (bool, error) {
+		pod, err := podClient.Pods(ns).Get(name, meta.GetOptions{})
+		if err != nil {
+			logrus.WithError(err).Errorf("Waiting for container %s in pod %s in namespace %s", containerName, name, ns)
+			return false, nil
+		}
+
+		for _, container := range pod.Status.ContainerStatuses {
+			if container.Name == containerName {
+				if container.State.Running != nil || container.State.Terminated != nil {
+					return true, nil
+				}
+				break
+			}
+		}
+
+		return false, nil
+	})
 }
 
 func copyArtifacts(podClient PodClient, into, ns, name, containerName string, paths []string) error {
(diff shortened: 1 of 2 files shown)


pkg/steps/artifacts.go
```diff
@@ -216,6 +216,27 @@ type PodClient interface {
 	podCmdExecutor
 }
 
+func waitForContainer(podClient PodClient, ns, name, containerName string) error {
+	logrus.Tracef("Waiting for container %s in pod %s in namespace %s", containerName, name, ns)
```
stevekuznetsov, line 220: nit:

```go
logrus.WithFields(logrus.Fields{
  "namespace": ns,
  "name": name,
  "container": containerName,
}).Trace("Waiting for container to be running.")
```

pkg/steps/artifacts.go
```diff
@@ -216,6 +216,27 @@ type PodClient interface {
...
+		for _, container := range pod.Status.ContainerStatuses {
+			if container.Name == containerName {
+				if container.State.Running != nil || container.State.Terminated != nil {
+					return nil
+				}
+				break
+			}
+		}
+
+		time.Sleep(time.Second)
```
stevekuznetsov, line 236: Can we use `wait.PollImmediate()`

pkg/steps/artifacts.go
wking: My line is cribbed from [this master line][1], and `pkg/steps` has no `WithFields` consumers as of the current master (d9084dd58).  Can I punt on this and pick it up if master pivots `pkg/steps` to `WithFields` before this PR lands?

[1]: https://github.com/openshift/ci-tools/blob/d9084dd58eefba841b7e4e96bd843f6074e11f71/pkg/steps/artifacts.go#L220

pkg/steps/artifacts.go
wking: > Can we use `wait.PollImmediate()`

Pivoted to `PollImmediateInfinite` with 6c2d3168 -> ea588d07, because `copyArtifacts` does not take a `Context` or timeout duration or other limit.  If you want `PollImmediate`, did you have a preference around an internal, hard-coded timeout or plumbing through a context/timeout from a higher level?

bbguimaraes: This should be quick and rarely fail, I don't think sending in the interruption context would be worth the implementation complexity (e.g. we still want to gather artifacts for `post` multi-stage steps even after an interruption). I would still add a reasonably long timeout just so it doesn't get stuck forever, unlikely as that may be.

wking: Pick a reasonably long timeout for me?

bbguimaraes: The containers should start at roughly the same instant, so something like `5m`?

pkg/steps/artifacts.go
stevekuznetsov: It would be faster to fix it, than to type all that out :)

stevekuznetsov:
@wking does not look like you pass tests yet

stevekuznetsov:
@wking any updates on this one?

wking:
Replaced by #1030.

/close

PR #960, opened 2020-06-29
TestImageStreamImports: Use name that passes DNS-1123 validation

alvaroaleman:


droslean:
There is already a library that validates this kind of format. @alvaroaleman wanna check? https://godoc.org/k8s.io/apimachinery/pkg/util/validation

PR #966, opened 2020-06-30
Registry replacer: add pr creation functionality

alvaroaleman:
This PR adds pr-creating abilities to the registry replacer

### DIFF
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -2,10 +2,12 @@ package main
 
 import (
 	"bytes"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"os"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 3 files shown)


cmd/registry-replacer/main.go
```diff
@@ -14,24 +16,73 @@ import (
...
+	selfApprove    bool
+	flagutil.GitHubOptions
+}
+
+func gatherOptions() (*options, error) {
+	o := &options{}
+	flag.StringVar(&o.configDir, "config-dir", "", "The directory with the ci-operator configs")
+	flag.BoolVar(&o.createPR, "create-pr", false, "If the tool should automatically create a PR. Requires --token-file")
+	flag.StringVar(&o.githubUserName, "github-user-name", "openshift-bot", "Name of the github user. Required when --create-pr is set. Does nothing otherwise")
+	flag.BoolVar(&o.selfApprove, "self-approve", false, "If the bot should self-approve its PR.")
```
stevekuznetsov, line 43: Do we want a common options struct to hold these and wrap the interaction with the library?

alvaroaleman: not really worth it at this point, this is about to options, a boolean and a string, I don't think we gain much from sharing that

images/registry-replacer/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM centos:7
+LABEL maintainer="muller@redhat.com"
```
petr-muller, line 2: :trollface:

images/registry-replacer/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM centos:7
```
petr-muller, line 1: Care to bump to 8 given you bumped the applyconfig one?

alvaroaleman: Done

PR #969, opened 2020-07-01
Use upstream prow package for http metrics

GeoBK:
Moving reusable code for metrics handling and HTTP helpers to package httphelper.
This code in configresolver will be used in the bugzilla-backporter tool as well.


GeoBK:
I'll write the tests after a once over @alvaroaleman

cmd/ci-operator-configresolver/main.go
```diff
@@ -39,46 +39,25 @@ type options struct {
...
-}
-
 var (
-	configresolverMetrics = struct {
-		httpRequestDuration *prometheus.HistogramVec
-		httpResponseSize    *prometheus.HistogramVec
-		errorRate           *prometheus.CounterVec
-	}{
-		httpRequestDuration: prometheus.NewHistogramVec(
+	configresolverMetrics = httphelper.Metrics{
```
alvaroaleman, line 43: Can we add a constructor for this that just takes the metrics `namespace` as argument? Something like this
```
func NewMetrics(namespace string) Metrics {
 return Metrics{
  HttpRequestDuration: prometheus.NewHistogramVec(
prometheus.HistogramOpts{
 Namespace: namespace,
 Name: "http_request_duration_seconds",
  Help:    "http request duration in seconds",
Buckets: []float64{0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2}

}
  )
}
```

(weirdly formatted, sorry).

The client lib prefixes the `Name` with `Namespace + "_"` if the latter is set, hence this should not result in a namechange of the metric and we avoid duplicating this code. This is espcially important since the client panics if you use labels that were not defined on the metrics

pkg/httphelper/httphelper_test.go
```diff
@@ -0,0 +1 @@
+package httphelper
```
alvaroaleman, line 1: :trollface:

GeoBK: I'll add stuff in there :P

pkg/httphelper/httphelper.go
```diff
@@ -0,0 +1,57 @@
...
+	HttpResponseSize    *prometheus.HistogramVec
+	ErrorRate           *prometheus.CounterVec
+}
+
+func RecordError(label string, metrics *Metrics) {
+	labels := prometheus.Labels{"error": label}
+	metrics.ErrorRate.With(labels).Inc()
+}
+
+func HandleWithMetrics(h http.HandlerFunc, metrics *Metrics) http.HandlerFunc {
```
GeoBK, line 39: Also, what would be better - leaving it like this or making this function and the RecordErrors function part of the Metrics class?

cmd/ci-operator-configresolver/main.go
GeoBK: And are we ok with stuff like duration buckets being common for different applications?

alvaroaleman: Yes, there is no reason at all for them to be different.

pkg/httphelper/httphelper.go
alvaroaleman: Go doesn't have classes, it only has types. And yeah, making it a method on `*Metrics` makes sense

pkg/httphelper/httphelper_test.go
```diff
@@ -0,0 +1,206 @@
...
+				t.Errorf("mismatch between expected and actual TraceResponseWriter size: expected %d, got %d", len(resp), trw.size)
+			}
+		})
+	}
+}
+
+var metrics = NewMetrics("testnamespace")
+
+var handleMetricsCalls = 0
+var recordErrorCalls = 0
```
petr-muller, line 78: Huh do these *need* to be globals? The `metrics` seems to be shared between the two following tests and the globals are incremented in individual testcases (which is not thread-safe which will surprise anyone who will try to parallelize these tests in the future). Is there a way how to create these separately for each testcase? If yes then we should use it, if not, then this deserves an explanatory paragraph about why these are globals shared from multiple places.

GeoBK: I think they need to be globals, initially I was initializing a metrics variable for each test but I ended up getting an error saying that the same metrics cant be registered multiple times which is why I was forced to make them globals. The call counters were needed to keep track of the number of times the tests were called. I guess I can use closures(?) for the call counters. wdyt?

alvaroaleman: The counter does definitely not need to be a global and should be testcase-local. The metrics currently need to be (binary) global because `NewMetrics` calls `prometheus.MustRegister`. I suggest to add a new constructor that is used in `NewMetrics` for only constructing the metrics without registering  them to allow to also have the metrics test-case local (`testutil.CollectAndCompare` does not require the metric to be registered)

pkg/httphelper/httphelper_test.go
```diff
@@ -0,0 +1,206 @@
...
+			recordErrorCalls++
+			expectedOut := fmt.Sprintf(tc.expectedOut, recordErrorCalls)
+			if err := testutil.CollectAndCompare(metrics.ErrorRate, strings.NewReader(expectedOut)); err != nil {
+				t.Errorf("unexpected metrics for ErrorRate:\n%s", err)
+			}
+
+		})
+	}
+}
+func wrapperOneByteWriter(t *testing.T) http.HandlerFunc {
```
alvaroaleman, line 111: This isn't wrapping anything

pkg/httphelper/httphelper.go
```diff
@@ -0,0 +1,108 @@
...
+}
+
+// NewMetrics is a constructor for Metrics
+func NewMetrics(namespace string) *Metrics {
+	m := &Metrics{
+
+		HTTPRequestDuration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Namespace: namespace,
+				Name:      "http_request_duration_seconds",
```
stevekuznetsov, line 45: We already have https://github.com/kubernetes/test-infra/blob/master/prow/metrics/http.go, let's use it.

pkg/httphelper/httphelper_test.go
GeoBK: Why do you say that isnt it wrapping the function that is being returned? Or do you mean that it doesn't need the "wrapper" since the inner function is anonymous?

pkg/httphelper/httphelper_test.go
GeoBK: So I would be making a new constructor only to allow for testing?

pkg/httphelper/httphelper.go
GeoBK: do you want me to move the tests in http_helper to prow or are they not necessary?

GeoBK: Also the buckets for httpDuration will probably be different from what they were earlier if we use the prow/metric, since it allows only for powers of 2 but right now it is - {0.0005, 0.001, 0.0025, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2}.
Will that be ok?

cmd/ci-operator-configresolver/main.go
```diff
@@ -286,16 +224,35 @@ func main() {
...
+			l("examples"),
+			l("ci-operator"),
+			l("leases"),
+		),
+		l("search"),
+		l("job"),
+		l("reference"),
+		l("chain"),
+		l("workflow"),
+	))
```
GeoBK, line 243: I'm not sure I'm using this the right way - I know I'm missing some fields under search. I'll add them once I get confirmation that this is how it should be used.

pkg/httphelper/httphelper.go
stevekuznetsov: Yes, those are fine.

cmd/ci-operator-configresolver/main.go
stevekuznetsov: Yes, these are correct, literal snippets for parts of the URL path that are static. No variables since this server uses query parameters, not varying paths.

PR #970, opened 2020-07-01
pj-rehearse: Avoid reusing loop iterators in closures 

petr-muller:
Otherwise we do not end up creating all necessary ImageStreams (see
DPTP-1312). Also, adding some logging.

https://github.com/golang/go/wiki/CommonMistakes#using-goroutines-on-loop-iterator-variables

/cc @alvaroaleman @bbguimaraes


bbguimaraes:
Everyone's favorite bug in Go.

/lgtm

PR #971, opened 2020-07-02
Consider a pullspec an input for ImportReleaseStep

petr-muller:
The pullspecs can come from two sources: `releases` stanza in the ci-op
config (which is an input on its own, so in this case its a no-op), and
`RELEASE_IMAGE_` variables. In the second case, we need to treat the
value as an input for the namespace to be different for different
release images.


bbguimaraes:
I wonder if this will fix all cases (is the spec always unique?), but it is at least a step in the right direction and takes care of the current issue.

/lgtm

PR #973, opened 2020-07-02
Conditionally run the resourcewatch during the upgrade tests

retroflexer:
Conditionally run-resourcewatch if it is available, to get the git monitoring tool running during the upgrade tests.


deads2k:
stevek said approval for this file acceptable. There is no rehearse

PR #974, opened 2020-07-02
create namespace for leader election to succeed for resource watcher in upgrade

deads2k:
builds on https://github.com/openshift/ci-tools/pull/973

/assign @damemi

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -322,9 +322,11 @@ objects:
         }
 
         function run-upgrade-tests() {
-          openshift-tests run-resourcewatch --help > /dev/null 2>&1
-          rwsupported=$?
+          rwsupported=0
+          openshift-tests run-resourcewatch --help > /dev/null 2>&1 || rwsupported=$?
           if [[ $rwsupported -eq 0 ]]; then
+            # need to create namespace for lease acquisition
+            oc --kubeconfig=${KUBECONFIG} create namespace ci-operator-monitor || true
             REPOSITORY_PATH=/resourcewatch openshift-tests run-resourcewatch --kubeconfig=${KUBECONFIG} --namespace=ci-operator-monitor &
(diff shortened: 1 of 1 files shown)


deads2k:
found in logs for https://deck-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gcs/origin-ci-test/logs/release-openshift-origin-installer-e2e-gcp-upgrade/1278756056368943104

pkg/steps/clusterinstall/template.go
```diff
@@ -322,13 +322,21 @@ objects:
         }
 
         function run-upgrade-tests() {
-          REPOSITORY_PATH=/resourcewatch openshift-tests run-resourcewatch --kubeconfig=${KUBECONFIG} --namespace=ci-operator-monitor &
-          rw_pid=$!
+          openshift-tests run-resourcewatch --help > /dev/null 2>&1
+          rwsupported=$?
```
retroflexer, line 326: I think these two lines need to be changed. Could you replace the 325-326 with these two lines?
 ```
          rwsupported=0   
          openshift-tests run-resourcewatch --help > /dev/null 2>&1 || rwsupported=$?
```

deads2k: well, this is a pretty bad look. We're hosed again.

I'll try one last tiem to fix it

deads2k:
if this doesn't work, I'm revert the whole lot to prepare for the next disruptive bit.

We have got to find a better way, this is garbage.

pkg/steps/clusterinstall/template.go
retroflexer: We use 'set -e' that is really the trouble maker. I don't like `set -e` at all. Since it fails with any error, we have to invent these kind of workaround that makes the scripts look very ugly. 

I have written lengthy shell scripts and maintained them, while taking care of my own error handling. Never had a problem with scripts or error handling.

PR #975, opened 2020-07-02
Registry replacer: Do not replace if from: is set

alvaroaleman:
We set images.*.from to point to pipelineimagestream:from. Since
the pipelineimagestream is guaranteed to be local, we do not have to add
another replacement.

In order to correctly replace remaining FROM directives in the
Dockerfile, we copy the replacing logic from openshift/builder and run
it as a pre-step before looking what additional replacements are needed.

Corresponding o/release PR: https://github.com/openshift/release/pull/10067


alvaroaleman:
/hold
probably more complicated than that

cmd/registry-replacer/main.go
```diff
@@ -75,6 +75,9 @@ func replacer(
 		}
 
 		for idx := range config.Images {
+			if config.Images[idx].From != "" {
```
stevekuznetsov, line 78: We should determine what field in the `Build` is set with the `images[*].from` and what that behavior is -- it might replace only the "first" `FROM` in the Dockerfile, in which case we stlll need to mess with the others.

alvaroaleman: Done

cmd/registry-replacer/main.go
```diff
@@ -351,3 +358,20 @@ type censor struct {
 func (c *censor) Censor(data []byte) []byte {
 	return bytes.ReplaceAll(data, c.secret, []byte("<< REDACTED >>"))
 }
+
+// applyReplacementsToDockerfile duplicates what the build tools would do
+func applyReplacementsToDockerfile(in []byte, image *api.ProjectDirectoryImageBuildStepConfiguration) ([]byte, error) {
```
stevekuznetsov, line 363: `image` can be `nil`

alvaroaleman: If that is the case we would never get to this code, we have a check in the very beginning:
```
if len(config.Images) == 0 {
    return nil
}
```

PR #978, opened 2020-07-03
Pod step: Fail if pod doesn't start running within 30 minutes

alvaroaleman:
Ideally we would also show the events but that would require replacing the client with one that can get events. I can do that in a follow-up if we deem it necesary. The conditions do at least cover the pull image case properly:
```
  containerStatuses:
  - image: guaranteed-undef
    imageID: ""
    lastState: {}
    name: guaranteed-undef
    ready: false
    restartCount: 0
    started: false
    state:
      waiting:
        message: Back-off pulling image "guaranteed-undef"
        reason: ImagePullBackOff
```

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -683,3 +683,29 @@ func getSourceSecretFromName(secretName string) *coreapi.LocalObjectReference {
 	}
 	return &coreapi.LocalObjectReference{Name: secretName}
 }
+
+func getReasonsForUnreadyContainers(p *coreapi.Pod) string {
+	builder := &strings.Builder{}
+	for _, c := range p.Status.ContainerStatuses {
+		if c.Ready {
+			continue
+		}
+		var reason, message string
+		switch {
+		case c.State.Waiting != nil:
+			reason = c.State.Waiting.Reason
+			message = c.State.Waiting.Message
+		case c.State.Running != nil:
+			reason = c.State.Waiting.Reason
+			message = c.State.Waiting.Message
+		case c.State.Terminated != nil:
+			reason = c.State.Terminated.Reason
+			message = c.State.Terminated.Message
+		default:
+			reason = "unknown"
+			message = "unknown"
+		}
+		_, _ = builder.WriteString(fmt.Sprintf("\n* Container %s is not ready with reason %s and message %s", c.Name, reason, message))
+	}
+	return builder.String()
+}
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -588,6 +588,9 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
 		if event.Type == watch.Deleted {
 			return false, appendLogToError(fmt.Errorf("the pod %s/%s was deleted without completing after %s (failed containers: %s)", pod.Namespace, pod.Name, podDuration(pod).Truncate(time.Second), strings.Join(failedContainerNames(pod), ", ")), podMessages(pod))
 		}
+		if !isPodRunning(pod) && time.Since(pod.CreationTimestamp.Time) > 30*time.Minute {
+			return false, fmt.Errorf("pod didn't start running within 30 minutes: %s", getReasonsForUnreadyContainers(pod))
+		}
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/steps/template.go
```diff
@@ -728,6 +731,15 @@ func podJobIsFailed(pod *coreapi.Pod) bool {
 	return false
 }
 
+func isPodRunning(pod *coreapi.Pod) bool {
+	for _, condition := range pod.Status.Conditions {
```
droslean, line 735: IMHO, there can be a time where the pod was ready but failed after. Searching through pod's condition history to find if there is a `ready` status, shouldn't be the case here. Instead, I think it is better to check only the most recent's time condition.

pkg/steps/source.go
```diff
@@ -683,3 +683,29 @@ func getSourceSecretFromName(secretName string) *coreapi.LocalObjectReference {
...
 }
+
+func getReasonsForUnreadyContainers(p *coreapi.Pod) string {
+	builder := &strings.Builder{}
+	for _, c := range p.Status.ContainerStatuses {
+		if c.Ready {
+			continue
+		}
+		var reason, message string
+		switch {
```
droslean, line 694: nit: to prevent multiple lookups, lets just `switch state := c.State; state {`

pkg/steps/template.go
alvaroaleman: There can only be one ready condition

pkg/steps/source.go
alvaroaleman: sorry, I don't understand. What do we gain from copying the variable?

pkg/steps/template.go
```diff
@@ -728,6 +731,15 @@ func podJobIsFailed(pod *coreapi.Pod) bool {
 	return false
 }
 
+func isPodRunning(pod *coreapi.Pod) bool {
```
stevekuznetsov, line 734: I am vaguely uneasy about this -- I know that there's some weirdness with Pod state when `InitContainers` are running. Can we get some YAML dumps of Pods from production and write a suite of tests or something?

alvaroaleman: Yeah this is only true if all pods are ready after all initContainers have finished. Do we have cases with long-running initcontainers?

stevekuznetsov: I don't know but I surely don't want to find out.

stevekuznetsov: (deploying this as is has the risk of being a highly visible regression in behavior that will surely send some angry users to the triage, so I'd prefer to do the safe thing)

pkg/steps/template.go
```diff
@@ -588,6 +588,9 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
 		if event.Type == watch.Deleted {
 			return false, appendLogToError(fmt.Errorf("the pod %s/%s was deleted without completing after %s (failed containers: %s)", pod.Namespace, pod.Name, podDuration(pod).Truncate(time.Second), strings.Join(failedContainerNames(pod), ", ")), podMessages(pod))
 		}
+		if !isPodRunning(pod) && time.Since(pod.CreationTimestamp.Time) > 15*time.Minute {
+			return false, fmt.Errorf("pod didn't start running within 15 minutes: %s", getReasonsForUnreadyContainers(pod))
+		}
```
bbguimaraes, line 593: Some types of failures will never resolve themselves no matter how long we wait.  We could fail immediately here if we have enough information to identify those.

alvaroaleman: So you are suggesting that we shouldn't use a timeout at all and instead fail on specific conditions, like ImagePullBackoff?

bbguimaraes: I think a hard timeout is still useful, and we cannot avoid having one because some errors (some types of `ImagePullBackoff`, for example) are transient.

If we have enough information in the pod status to determine if the error is transient or not, it would be nice to fail immediately.  If not, this is fine as it is.

alvaroaleman: But do we have any kind of error here that is guaranteed not to be transient?

pkg/steps/template.go
alvaroaleman: I increased the timeout to 30 minutes. It doesn't seem there are any metrics around phase transition time, so we have to rely on guessing.

pkg/steps/template.go
bbguimaraes: IIRC, the registry acting up results in a transient `ImagePullBackoff`, for example.

alvaroaleman: Yeah so that in turn means there is basically no error for which we can say for sure that its terminal, right?

bbguimaraes: That was kind of my question: does the pod information we have at that point allow us to determine whether it is terminal or not?

PR #980, opened 2020-07-03
Pass leaseClient as interface, not as pointer to interface

alvaroaleman:
@bbguimaraes you did this in 78ee4642d why? We do not have a public implementation of this so it should always be nil without needing the (de-)referencing to prevent ppl to accidentally add type info that would make it non-nil?

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -539,7 +539,7 @@ func (o *options) Run() []error {
 
 	dryLogger := steps.NewDryLogger(o.determinizeOutput)
 	// load the graph from the configuration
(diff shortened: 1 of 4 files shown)


alvaroaleman:
/hold
genuine failure it seems

bbguimaraes:
```
$ git cat-file -p 78ee4642d92385c8988fe31eb91987e1fbd07b39
tree 4bf3782229d22dff3a8b87023e665fd6a0aa5b67
parent bd7b9f09affb86fc056fffe65984a41ab4e487ad
author Bruno Barcarol Guimaraes <bbarcaro@redhat.com> 1576252440 +0000
committer Bruno Barcarol Guimaraes <bbarcaro@redhat.com> 1576252681 +0000

pkg/steps/lease: access the client indirectly

The client potentially requires information about the wrapped step, so
it must be created later.
```

bbguimaraes:
I don't quite understand your second question, but the reason it is a pointer is that the actual concrete object (and the pointer to it) is created after the step, as explained in the commit message above.

PR #986, opened 2020-07-08
PJ-Reharse: Deduplicate extraRefs/Refs

alvaroaleman:
Currently, periodics that have openshift/release in extraRefs will fail to rehearse because clonerefs fails with
```
Invalid options: sync config for openshift/release provided more than once
```

This PR fixes that.


droslean:
@alvaroaleman Can you add some description about what problem this change is fixing?

alvaroaleman:
@droslean done

PR #990, opened 2020-07-08
Parallelize externalBugPR calls and add help option

GeoBK:


pkg/backporter/backporter.go
```diff
@@ -299,44 +339,47 @@ func GetBugHandler(client bugzilla.Client) http.HandlerFunc {
...
 	// Remove target versions of the original bug
 	targetVersions.Delete(bug.TargetRelease...)
-	for _, clone := range clones {
-		clonePRs, err := client.GetExternalBugPRsOnBug(clone.ID)
-		if err != nil {
-			return nil, http.StatusInternalServerError, fmt.Errorf("Bug#%d - error occured while retreiving list of PRs : %w", clone.ID, err)
-		}
-
+	numClones := len(clones)
+	PRErrors := make([]error, numClones+1)
```
petr-muller, line 351: Looks like `numClones` is only used here, IMO no need for the variable. Also, looks like this could in fact be a channel of errors? Would save you from the need to book-keep the indexes.

pkg/backporter/backporter.go
```diff
@@ -184,6 +197,32 @@ const clonesTemplateConstructor = `
     </form>
 </div>`
 
+const helpTemplateConstructor = `
+<h2 id="title"><a href="#title">Why a bugzilla-backporter?</a></h2>
+
+<p>
+The <code>Bugzilla Backporter</code> would allow the user to easily understand how different clones
```
petr-muller, line 204: Call the thing consistently either "Bugzilla Backporter" or "bugzilla-backporter" (I think for humans we prefer the former).

petr-muller: Also I don't think we need the conditional - let's just say "backporter allows" instead of "would allow" (in the whole text).

pkg/backporter/backporter.go
```diff
@@ -184,6 +197,32 @@ const clonesTemplateConstructor = `
...
+are related to each other and which release they are targeted for.
+It also allows the user to create clones targeting a specific release.
+</p>
+<h2 id="title"><a href="#title">How to find clones?</a></h2>
+
+<p>
+The user would enter a bugID for the bug whose clones need to be found in the input field on the top right
```
petr-muller, line 211: Maybe use imperative consistenly and tell the user what to do? "_Enter the bug ID into the input field (...) The tool will show a list of all clones..._ etc.

pkg/backporter/backporter.go
GeoBK: Yup, removed the numClones variable and switched to the errgroup suggested by Alvaro in another PR

pkg/backporter/backporter.go
GeoBK: Yes, I will change that

stevekuznetsov:
@GeoBK @alvaroaleman could we get a `make` target similar to `make pr-deploy` for this web server, so that we can view the UI for the PR in question?

pkg/backporter/backporter.go
```diff
@@ -311,40 +356,46 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions
...
-	if err != nil {
-		return nil, http.StatusInternalServerError, fmt.Errorf("unable to fetch root bug: %w", err)
-	}
-
 	clones, err := client.GetAllClones(bug)
 	if err != nil {
 		return nil, http.StatusInternalServerError, fmt.Errorf("unable to get clones: %w", err)
 	}
+	if len(clones) < 1 {
+		return nil, http.StatusInternalServerError, fmt.Errorf("clones list empty")
```
alvaroaleman, line 364: Why is this an error condition? A bug doesn't need to have clones?

pkg/backporter/backporter.go
```diff
@@ -311,40 +356,46 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions
...
 		// Remove target releases which already have clones
 		targetVersions.Delete(clone.TargetRelease...)
-
-		clone.PRs = clonePRs
+		g.Go(func() error {
+			clonePRs, err := client.GetExternalBugPRsOnBug(clone.ID)
+			if err != nil {
+				return fmt.Errorf("Bug#%d - error occured while retreiving list of PRs : %w", clone.ID, err)
+			}
+			clone.PRs = clonePRs
```
alvaroaleman, line 388: This variable is created on line 380 so from what I can see, we assign and never read that back?

pkg/backporter/backporter.go
GeoBK: The current logic for clones should return at least one bug (ie itself) in the clones list, since we have now moved to everything in the table view.

pkg/backporter/backporter.go
GeoBK: We are returning the clones variable, since clone is a pointer to an element in that array, the changes made to it get reflected when the array is returned.

PR #991, opened 2020-07-08
Registry cleaner: Prune replacements

alvaroaleman:
Sample is here and looks correct to me (smoke verified some of the examples): https://github.com/openshift/release/pull/10135


petr-muller:
Code looks fine and I checked some of the proposed outputs, too. Found nothing suspicious.

I think the tool starts to need a better description about all the transformations it is supposed to do.

PR #994, opened 2020-07-09
rehearse: extract CMManager to rehearse package

petr-muller:
The only purpose of CMManager is to support rehearsals, so it makes
sense for it to live in the `rehearse` package. This makes the
assumptions done by the code correct, and allows to simplify names even
more.

The direction of this change is to lessen the importance of the `ConfigMapSource` struct to make decisions about CM names.
This because the only thing that knows the names is the config-updater config for the production CMs, and the rehearse package can then infer a name from that.

### DIFF
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -41,6 +41,25 @@ const (
 	RegistryPath = "ci-operator/step-registry"
 )
 
+type ConfigMapSource struct {
+	PathInRepo, SHA string
+}
+
+func (s ConfigMapSource) Name() string {
+	base := filepath.Base(s.PathInRepo)
+	return strings.TrimSuffix(base, filepath.Ext(base))
+}
+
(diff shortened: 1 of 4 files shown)


pkg/config/release.go
```diff
@@ -41,6 +41,25 @@ const (
 	RegistryPath = "ci-operator/step-registry"
 )
 
+type ConfigMapSource struct {
```
alvaroaleman, line 44: Why do we need this here?

petr-muller: In the current state this is more tied to the config in the openshift/release repo than to the reherarsals: it holds a path in the repository and a SHA, and it is also used by other functions in this package (i.e. the ones that detect "which jobs use a template"), so keeping it here makes sense.

In the followup I want to get rid of this entirely.

PR #996, opened 2020-07-10
rehearse: rework the handling of temporary CMs

petr-muller:
In some cases rehearsals need to create temporary ConfigMaps, and their
handling was previously quite fragile. We tried to infer the ConfigMap
names from the paths and filenames but that was buggy.

The new approach is to reduce assumptions and guesswork. We only assume
the general locations of templates and cluster profiles, and use
config-updater config to determine the names of the CMs involved,
without any guessing.

Additionally, the information about CMs was used by different methods
for different purposes, which lead to some code bloat and unnecessary
complexity. This was reduced by the new `rehearse.ConfigMaps` struct
holding different facets of CM-related data, and only the necessary data
is passed to various methods. For example, the job selection methods now
only receive a set of production CM names, the job configuration methods
receive the mappings between production and rehearsal CMs, etc.

One notable change is the selection of jobs affected by template change:
previously, the code used the filename as a signifier of a job using a
template, but did not care if that filename came from an entirely
different ConfigMap, which is technically incorrect. The code was
changed to the same mechanism like cluster profile one was using:
detecting if the job uses the ConfigMap.

/cc @stevekuznetsov @bbguimaraes @alvaroaleman


pkg/rehearse/configmaps.go
```diff
@@ -91,48 +139,34 @@ func (c *CMManager) createCM(name string, data []updateconfig.ConfigMapUpdate) e
...
-			return nil
-		})
-		if err != nil {
-			return nil, err
-		}
-	}
-	return ret, nil
-}
-
-func (c *CMManager) validateChanges(changes []prowgithub.PullRequestChange) error {
```
petr-muller, line 119: This validation is no longer needed b/c the input the the `Create()` method is now a `ConfigMaps` instance, which cannot be successfully populated by `NewConfigMaps` if any of the changed files are not covered by a `config-updater` pattern.

pkg/rehearse/configmaps.go
```diff
@@ -144,23 +178,20 @@ func replaceSpecNames(namespace string, cfg prowplugins.ConfigUpdater, mapping m
 		}
 		if name, ok := mapping[v.Name]; ok {
 			v.Name = name
+			v.Namespaces = []string{""}
+			ret.Maps[k] = v
```
petr-muller, line 182: This is one tricky change: this prevents us from building a config-updater config (that will ultimately be used to populate the temporary CMs) that updates the production CMs on bugs in the logic. Technically this should no longer happen because our logic is overall fixed, but better safe then sorry.

test/integration/pj-rehearse/expected.yaml
```diff
@@ -702,9 +702,9 @@
           - secret:
               name: cluster-secrets-test-profile
           - configMap:
-              name: rehearse-cluster-profile-test-profile-47224c86
+              name: rehearse-1234-cluster-profile-cluster-profile-test-profile
       - configMap:
-          name: rehearse-template-test-template-4b355dc3
+          name: prow-job-cluster-launch-src
```
petr-muller, line 707: This CM not being changed is the consequence of the change in template substitution logic. This job uses a `test-template.yaml` from a `prow-job-cluster-launch-src` CM, but that CM is not handled at all by our config updater (see the [testing config](https://github.com/openshift/ci-tools/blob/921be048c697211820d3c5afbaec88d8e8273fe9/test/integration/pj-rehearse/master/core-services/prow/02_config/_plugins.yaml#L1-L8). Previously the rehearsal assumed the filenames match, therefore the file should be the same, but in production we should never have such case. New version of rehearse sees that this job does not use any of the config-updater-managed CMs, so it has no reason to do the substitution.

test/integration/pj-rehearse/expected.yaml
```diff
@@ -1059,262 +1059,6 @@
...
-      - author: petr-muller
-        number: 1234
-        sha: test_sha
-      repo: release
-    report: true
-    rerun_command: /test pj-rehearse
-    type: presubmit
-  status:
-    startTime: 2020-06-22T22:25:00Z
-    state: triggered
```
petr-muller, line 1317: These two jobs are the same case: they are not actually using any of the CMs that will be updated by this PR, so new rehearse sees no reason to even rehearse them (see their configs: [1](https://github.com/openshift/ci-tools/blob/921be048c697211820d3c5afbaec88d8e8273fe9/test/integration/pj-rehearse/master/ci-operator/jobs/super/duper/super-duper-master-presubmits.yaml#L196-L247) [2](https://github.com/openshift/ci-tools/blob/921be048c697211820d3c5afbaec88d8e8273fe9/test/integration/pj-rehearse/master/ci-operator/jobs/super/duper/super-duper-master-presubmits.yaml#L144-L195))

petr-muller:
/hold

This should be (I hope) reviewable, but I'd like to test this better against real PRs (it has a potential to be quite invasive), as well as investigate a bit more the impact of stopping of relying on SHAs to name CMs.

alvaroaleman:
Looks much cleaner and just as OOPy :P

petr-muller:
> Looks much cleaner and just as OOPy :P

Hey, I removed one struct with behavior and everything that used it now uses sets instead!

pkg/rehearse/configmaps.go
```diff
@@ -32,6 +33,53 @@ const (
...
+func tempConfigMapName(purpose, source string, prNumber int) string {
+	// Object names can't be too long so we truncate the hash. This increases
+	// chances of collision but we can tolerate it as our input space is tiny.
+	pr := strconv.Itoa(prNumber)
+	maxLen := 253 - len("rehearse---") - len(purpose) - len(pr)
+	if len(source) > maxLen {
+		source = source[:maxLen]
+	}
+	return fmt.Sprintf("rehearse-%s-%s-%s", pr, purpose, source)
+}
```
petr-muller, line 81: @bbguimaraes I wanted to consult the naming of the temporary CMs with you here. I remember we initially put PR numbers into the name, and then you changed it to contain SHAs instead. I could not keep the SHAs because in my model I disconnect the CM name from its content (because I now work purely with CM names). 

Was there any problem with PR number-based names? If so I would consider reintroducing the effect of the content on the name (would need to sum multiple hashes though...), or even using the `BUILD_ID` or somehting like that.

petr-muller:
I still want to fix this:

```
INFO[0012] creating rehearsal configMap                  buildCluster=api.ci cm-name=rehearse-10084-cluster-profile-cluster-profile-aws-centos-40 org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.gitignore key=.gitignore org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.type key=.type org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/OWNERS key=OWNERS org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/vars.yaml key=vars.yaml org=openshift repo=release
INFO[0012] creating rehearsal configMap                  buildCluster=api.ci cm-name=rehearse-10084-cluster-profile-cluster-profile-aws-centos-40 org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.gitignore key=.gitignore org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.type key=.type org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/OWNERS key=OWNERS org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/vars.yaml key=vars.yaml org=openshift repo=release
INFO[0012] creating rehearsal configMap                  buildCluster=api.ci cm-name=rehearse-10084-cluster-profile-cluster-profile-aws-centos-40 org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.gitignore key=.gitignore org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.type key=.type org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/OWNERS key=OWNERS org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/vars.yaml key=vars.yaml org=openshift repo=release
INFO[0012] creating rehearsal configMap                  buildCluster=api.ci cm-name=rehearse-10084-cluster-profile-cluster-profile-aws-centos-40 org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.gitignore key=.gitignore org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/.type key=.type org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/OWNERS key=OWNERS org=openshift repo=release
DEBU[0012] Populating key.                               buildCluster=api.ci filename=cluster/test-deploy/aws-4.0/vars.yaml key=vars.yaml org=openshift repo=release
```

petr-muller:
OK that's gonna be a followup. In short: we create a `CMManager` for each involved cluster (here, one), which then uses `updateconfig` code to update the CMs, but that in turn evaluates its config and finds out that given change needs to be propagated to X clusters (here, four), but the callsite ignores the cluster field and ends up updating a single CM on a single cluster X times.

LOL

petr-muller:
Another thing is that we do not sample for cluster profiles, so touching a CP affects a lot of jobs and we're likely to hit the threshold.

petr-muller:
Oopsie: this needs better tests: https://github.com/openshift/ci-tools/compare/0c363343cbc6f53ca67ca6b18d4cad7684da56a3..3455ccff240e31c87d9b70657c260d241264e1e4

pkg/rehearse/configmaps.go
bbguimaraes: Here's a blast from the past:

The only change I made was to use the hash already present in Git instead of recalculating it by other means: https://github.com/openshift/ci-operator-prowgen/pull/167.

Re PR vs. hash: https://github.com/openshift/ci-operator-prowgen/pull/100#pullrequestreview-210698890.

pkg/config/release.go
```diff
@@ -225,30 +217,24 @@ func GetChangedRegistrySteps(path, baseRev string, graph registry.NodeByName) ([
...
 // `root`.  Paths are relative to `root`.
-func getRevChanges(root, path, base string, rec bool) ([]ConfigMapSource, error) {
+func getRevChanges(root, path, base string) ([]string, error) {
 	// Sample output (with abbreviated hashes) from git-diff-tree(1):
 	// :100644 100644 bcd1234 0123456 M file0
-	cmd := []string{"diff-tree", "--diff-filter=ABCMRTUX", base + ":" + path, "HEAD:" + path}
-	if rec {
-		cmd = append(cmd, "-r")
-	}
+	cmd := []string{"diff-tree", "-r", "--diff-filter=ABCMRTUX", base + ":" + path, "HEAD:" + path}
```
bbguimaraes, line 230: A year later and we still can't use the much nicer `--diff-filter=d` syntax =(

bbguimaraes:
I believe the discussion linked above means we will still need to keep the hash in CM names.

My other comments are contingent on how we intend to address that.

petr-muller:
> I believe the discussion linked above means we will still need to keep the hash in CM names.
> 
> My other comments are contingent on how we intend to address that.

Hash is hard b/c multiple files (with different hashes) can go into a single CM, and the current code depends on tracking the full paths of changed files to be able to match them against the config-updater patterns. I'm inclined towards just stuffing a `$BUILD_ID` (basically do a CM for each separate pj-rehearse run). WDYT?

bbguimaraes:
I may be missing something, but the current code already does all this.  Why not enforce explicitly the previous assumption that files go to a ConfigMap with the same name?  Specially since we are on the way to sunseting both templates and cluster profiles.  What benefit is there in supporting other cases?

The code cleanup looks good, that section has needed one for a long time.  I'm just not convinced that the rest of the churn is worth the effort — although I'm also not strictly opposed to it.

petr-muller:
> I may be missing something, but the current code already does all this. Why not enforce explicitly the previous assumption that files go to a ConfigMap with the same name?

The missing part is that the current code does not reliably work. The reason is the fragile guesswork over the ConfigMap names. The CM name is inferred wrong (specifically, for this cluster profile: https://github.com/openshift/release/tree/master/cluster/test-deploy/aws-4.0), and that then triggers whole plethora of bad paths - all of which are fixed here. The fix could be an *additional* piece of guesswork ("if it is a cluster profile, do not attempt to trim an extension from the filename, because it's probably a directory"), but I thought I'll bite the bullet and remove the layers of assumptions and workarounds.

It's not just a theoretical fix, it was hit by our users - working with cluster profiles is rare, but we should either have the feature on and maintain it, or disable it.

petr-muller:
It was not a new use case, it's a modification of an existing cluster profile where a user expected a rehearsal.

bbguimaraes:
I don't see any guesswork.  The code was built on an assumption.  I see one template and two profiles that do not follow the rule.  *That* is a fault of the code, it should have validated that the implicit assumption always held.  The extension problem could also be trivially fixed by only trimming `.yaml`, not all "extensions".  Those are what should be fixed, IMO.

Again, I don't see how removing those requirements and supporting arbitrary entries in `updateconfig` gives us any benefit.

All that said, I also don't want to waste the effort put into this PR: as I said, I think the cleanup is valid and good.  `BUILD_ID` sounds reasonable, I don't think we gain much by deduplicating based on content anyway.

petr-muller:
How would you validate that the assumption holds? You could compare it against the config-updater config, but at that point you know the names and no longer need the assumptions at all. What I called "guesswork" is recursive `git diff-tree` for cluster profiles and not for templates, extension trimming, and then inferring the names from file/dir names while the real name may be entirely different.

The benefits: previously we had complicated code for processing inputs following some unwritten naming conventions, with bugs stemming from these conventions not being followed. What I attempted to do is to replace it with simpler code (witnessed by simpler signatures and fewer methods without trying to be smart) that can process inputs without assumptions, removing opportunities for bugs to happen.

petr-muller:
/hold cancel

Made `$BUILD_ID` embedded in the CM names and added tests - all in new commits. I do not intend to make any more changes. 

@alvaroaleman @bbguimaraes PTAL and sorry for the size

petr-muller:
@bbguimaraes thx. I'll merge it tomorrow and keep an eye on it.

petr-muller:
Looks like configresolver tests are still flaking.

petr-muller:
_grumble grumble grumble_

/test e2e

PR #997, opened 2020-07-10
Registry replacer pruning: Consider copy --from and allow alias replacement

alvaroaleman:


cmd/registry-replacer/main.go
```diff
@@ -143,7 +143,11 @@ func replacer(
 		}
 
 		getter := githubFileGetterFactory(info.Org, info.Repo, info.Branch)
-		allSourceImages := sets.String{}
+		allReplacementCandidates := sets.String{}
+
+		// We have to skip pruning if we only get empty dockerfiles because it might mean
+		// that we do not have the appropriate permissions.
+		var hasNonEmptyDockerfile bool
```
stevekuznetsov, line 150: Let's make this discoverable since it will mean that we still have non-conformant configs

alvaroaleman: Done

alvaroaleman:
> So I vaguely remember Clayton saying that the image replacement does not actually work when you want to replace the as <name> and target the name -- did we confirm that users which configured it to do so, actually got what they wanted?

It does work if what is aliased in the `as` is not getting replaced already. The code this is using now is copied from what their replacer does

PR #1002, opened 2020-07-13
CI-Operator: Dump step graph to artifacts

alvaroaleman:
The usecase is to be able to determine what pods belong to a given test by getting all step names for the given test. The dependencies slice is something I added in case we need it in the future.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -31,6 +31,7 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/sets"
 	authclientset "k8s.io/client-go/kubernetes/typed/authorization/v1"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	rbacclientset "k8s.io/client-go/kubernetes/typed/rbac/v1"
@@ -576,6 +577,10 @@ func (o *options) Run() []error {
 
 	if err := printExecutionOrder(nodes); err != nil {
 		return []error{fmt.Errorf("could not print execution order: %w", err)}
+	}
+
+	if err := dumpGraph(o.artifactDir, nodes); err != nil {
+		return []error{fmt.Errorf("failed to dump graph to artifacts dir: %w", err)}
 	}
 
 	// initialize the namespace if necessary and create any resources that must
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -1380,6 +1385,64 @@ func printExecutionOrder(nodes []*api.StepNode) error {
...
+	}
+	return false
+}
+
+func iterateAllEdges(nodes []*api.StepNode, alreadyIterated sets.String, f func(*api.StepNode)) {
+	for _, node := range nodes {
+		iterateAllEdges(node.Children, alreadyIterated, f)
+		if alreadyIterated.Has(node.Step.Name()) {
+			continue
+		}
```
petr-muller, line 1440: Is there a reason why the `alreadyIterated` guard is done only after processing the children? I would think we could expect all children to be already seen too, no?

cmd/ci-operator/main.go
```diff
@@ -1380,6 +1385,64 @@ func printExecutionOrder(nodes []*api.StepNode) error {
...
+		result = append(result, r)
+	})
+
+	serialized, err := json.Marshal(result)
+	if err != nil {
+		return fmt.Errorf("failed to marshal: %w", err)
+	}
+
+	dest := filepath.Join(artifactsEnv, "graph.json")
```
petr-muller, line 1418: Can we get a better name? Something that's more descriptive when you eyeball artifacts in the viewer? Or maybe we should claim an artifact subdir? `ci-operator/graph.json` would work for me.

cmd/ci-operator/main.go
```diff
@@ -1380,6 +1385,64 @@ func printExecutionOrder(nodes []*api.StepNode) error {
 	return nil
 }
 
+type stepWithDependencies struct {
+	StepName     string
+	Dependencies []string
+}
+
+func dumpGraph(nodes []*api.StepNode) error {
+	artifactsEnv := os.Getenv(artifactsEnv)
```
bbguimaraes, line 1394: We have a `--artifacts` argument (which defaults to this), so we should use that.

cmd/ci-operator/main.go
alvaroaleman: Its done after processing the children because along the children chain there could be the current element, hence we avoid processing it twice. 

It can not be done before processing the children, because if we have an element `e` that is both a direct child of the initial element `i` and a grandchild via `i -> other-child -> e`, that latter `e` won't have its own children processed. No idea how likely this case is, but I don't really know much about the tree we are building here so I tried to avoid making assumptions.

alvaroaleman: Wait you are right, its a graph not a tree and the elements are unique per name, so if we reached ourselves via a child before, we can assume that our children are processed already since we process children first.

cmd/ci-operator/main.go
alvaroaleman: Renamed it to `ci-operator-step-graph.json`

PR #1005, opened 2020-07-15
[DPTP-1119] Allow ReporterConfig for Slack notifications to be added …

mgencur:
…manually

But only for periodic jobs where it makes sense.

https://issues.redhat.com/browse/DPTP-1119


mgencur:
Thanks!

PR #1006, opened 2020-07-16
Add maketarget for deploying the bugzilla backporter from a PR

GeoBK:


hack/pr-deploy.yaml
```diff
@@ -199,3 +199,158 @@ objects:
         - name: config
           configMap:
             name: config
+- apiVersion: build.openshift.io/v1
+kind: BuildConfig
+metadata:
```
petr-muller, line 204: This is badly indented - looks like pretty much everything from L203 below should be shifted right.

GeoBK: I fixed the indentation issue

stevekuznetsov:
I would expect this to be a separate target, to allow a dev to choose what it is that they are deploying.

hack/pr-deploy-backporter.yaml
```diff
@@ -0,0 +1,203 @@
...
+      kind: Service
+      name: bp-server
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: bugzilla-credentials-openshift-bugzilla-robot
+    namespace: ci-tools-${PULL_REQUEST}
+  type: Opaque 
+  data:
+    api: o
```
GeoBK, line 156: I need an empty file at "/etc/bugzilla/api". Will having a data element named "api" create that file there - since I am using the secret at the "/etc/bugzilla" volume?

stevekuznetsov: Will the server even work without a valid token? You could instead add to the Makefile the `oc` commands necessary to copy the credential from the `ci` namespace into the `ci-tools-${PULL_REQUEST}` one, as we do for config maps.

stevekuznetsov:
I'd also expect you to be able to run this on your laptop to test it out (as you'll be running it to deploy it for your PRs, too), so you should be able to test-run it and validate before merging.

hack/pr-deploy-backporter.yaml
GeoBK: The server should show the view even without the keys - its only for the create clones(ie. write) functionality that the token is needed

GeoBK:
@stevekuznetsov , I cant run it on my local since I dont have access to the online cluster and so cant copy the plugins configmap from the ci namespace

Makefile
```diff
@@ -165,12 +165,19 @@ update-integration:
...
 	for cm in ci-operator-master-configs step-registry config; do oc  --context app.ci --as system:admin get --export configmap $${cm} -n ci -o json | oc  --context app.ci --as system:admin create -f - -n ci-tools-$(PULL_REQUEST); done
 	echo "server is at https://$$( oc  --context app.ci --as system:admin get route server -n ci-tools-$(PULL_REQUEST) -o jsonpath={.spec.host} )"
 .PHONY: pr-deploy
 
+pr-deploy-backporter:
+	oc --context app.ci --as system:admin process -p USER=$(USER) -p BRANCH=$(BRANCH) -p PULL_REQUEST=$(PULL_REQUEST) -f hack/pr-deploy-backporter.yaml | oc  --context app.ci --as system:admin apply -f -
+	oc  --context app.ci --as system:admin get --export configmap plugins -n ci -o json | oc  --context app.ci --as system:admin create -f - -n ci-tools-$(PULL_REQUEST)
+	oc  --context app.ci --as system:admin get --export secret bugzilla-credentials-openshift-bugzilla-robot -n ci -o json | oc  --context app.ci --as system:admin create -f - -n ci-tools-$(PULL_REQUEST)
+	echo "server is at https://$$( oc  --context app.ci --as system:admin get route server -n ci-tools-$(PULL_REQUEST) -o jsonpath={.spec.host} )"
```
alvaroaleman, line 178: The route is named `bp-server` and not `server`, other than that it works like a charm

GeoBK: cool. I changed server to bp-server in the oc get route

PR #1015, opened 2020-07-22
Add a caching transport for the Bugzilla client

GeoBK:
Inject the HTTPRoundTripper transport to the bugzilla client when caching is necessary

### DIFF
--- a/pkg/backporter/httpcache.go
+++ b/pkg/backporter/httpcache.go
@@ -3,14 +3,22 @@ package backporter
 import (
 	"bufio"
 	"bytes"
+	"context"
 	"fmt"
 	"net/http"
 	"net/http/httputil"
 	"sync"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
+)
+
+const (
+	xFromCache       = "X-From-Cache"
+	cacheRefreshFreq = 10 * time.Minute
 )
 
 type bugzillaCache struct {
@@ -18,24 +26,19 @@ type bugzillaCache struct {
 	cache map[string][]byte
 }
 
-func (bc *bugzillaCache) Get(key string) ([]byte, bool) {
+func (bc *bugzillaCache) get(key string) ([]byte, bool) {
 	bc.lock.Lock()
 	defer bc.lock.Unlock()
 	cachedVal, ok := bc.cache[key]
 	return cachedVal, ok
 }
 
-func (bc *bugzillaCache) Set(key string, respBytes []byte) {
+func (bc *bugzillaCache) set(key string, respBytes []byte) {
 	bc.lock.Lock()
 	defer bc.lock.Unlock()
 	bc.cache[key] = respBytes
 }
 
-func (bc *bugzillaCache) Delete(key string) {
-	bc.lock.Lock()
-	defer bc.lock.Unlock()
-	delete(bc.cache, key)
-}
 func newBugzillaCache() *bugzillaCache {
 	return &bugzillaCache{cache: map[string][]byte{}}
 }
@@ -59,25 +62,27 @@ func (t *cachingTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 	g.Go(func() error {
 		resp, err = t.transport.RoundTrip(req)
 		if err != nil {
-			t.cache.Delete(req.URL.String())
 			return err
 		}
-		body, err := httputil.DumpResponse(resp, true)
-		if err != nil {
-			t.cache.Delete(req.URL.String())
-			return fmt.Errorf("err while prepping response to cache: %w", err)
-		}
 		if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
-			t.cache.Set(req.URL.String(), body)
-		} else {
-			t.cache.Delete(req.URL.String())
+			body, err := httputil.DumpResponse(resp, true)
+			if err != nil {
+				return fmt.Errorf("err while serializing response to cache: %w", err)
+			}
+			t.cache.set(req.URL.String(), body)
 		}
+
 		return nil
 	})
 
-	if cachedVal, isCached := t.cache.Get(req.URL.String()); isCached {
+	if cachedVal, isCached := t.cache.get(req.URL.String()); isCached {
 		b := bytes.NewBuffer(cachedVal)
-		return http.ReadResponse(bufio.NewReader(b), req)
+		cachedResp, err := http.ReadResponse(bufio.NewReader(b), req)
+		if err != nil {
+			return nil, err
+		}
+		cachedResp.Header.Set(xFromCache, "1")
+		return cachedResp, nil
 	}
 	if err := g.Wait(); err != nil {
 		return nil, err
@@ -85,29 +90,31 @@ func (t *cachingTransport) RoundTrip(req *http.Request) (*http.Response, error) 
 	return resp, nil
 }
 
-func refreshCache(bc *bugzillaCache) {
-	concurrencyLimiter := make(chan string, 10)
+func refreshCache(bc *bugzillaCache, m prometheus.Gauge) {
+	var sem = semaphore.NewWeighted(int64(10))
+	ctx := context.Background()
+	logrus.WithField("cache_entries", len(bc.cache)).Info("Refreshing cache")
+	m.Set(float64(len(bc.cache)))
 	for url := range bc.cache {
-		concurrencyLimiter <- url
+		if err := sem.Acquire(ctx, 1); err != nil {
+			logrus.WithError(fmt.Errorf("failed to acquire semaphore for key %s: %w", url, err))
+		}
+		url := url
 		go func() {
-			url := <-concurrencyLimiter
+			defer sem.Release(1)
 			resp, err := http.Get(url)
 			if err != nil {
-				bc.Delete(url)
 				logrus.WithError(fmt.Errorf("cache refresh error - failed to fetch %s: %w", url, err))
 				return
 			}
 			defer resp.Body.Close()
-			body, err := httputil.DumpResponse(resp, true)
-			if err != nil {
-				bc.Delete(url)
-				logrus.WithError(fmt.Errorf("cache refresh error - DumpResponse failed %s: %w", url, err))
-				return
-			}
 			if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
-				bc.Set(url, body)
-			} else {
-				bc.Delete(url)
+				body, err := httputil.DumpResponse(resp, true)
+				if err != nil {
+					logrus.WithError(fmt.Errorf("cache refresh error - DumpResponse failed %s: %w", url, err))
+					return
+				}
+				bc.set(url, body)
 			}
 		}()
 	}
@@ -124,12 +131,19 @@ func NewCachingTransport() http.RoundTripper {
 		cache:     newBugzillaCache(),
 		transport: http.DefaultTransport,
 	}
-	ticker := time.NewTicker(10 * time.Minute)
+	cacheRefreshMetrics := prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Name: "bugzilla_backporter_cached_bugs",
+			Help: "bugs in cache to be refreshed",
+		},
+	)
+	prometheus.MustRegister(cacheRefreshMetrics)
+	ticker := time.NewTicker(cacheRefreshFreq)
 	go func() {
 		defer ticker.Stop()
 		for {
 			<-ticker.C
-			refreshCache(t.cache)
+			refreshCache(t.cache, cacheRefreshMetrics)
 		}
 	}()
 	return &t
--- a/pkg/backporter/httpcache.go
+++ b/pkg/backporter/httpcache.go
@@ -1,0 +1,89 @@
+package backporter
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"net/http"
+	"net/http/httputil"
+	"sync"
+
+	"golang.org/x/sync/errgroup"
+)
+
+type bugzillaCache struct {
+	lock  sync.Mutex
+	cache map[string][]byte
+}
+
+func (bc *bugzillaCache) Get(key string) ([]byte, bool) {
+	bc.lock.Lock()
+	defer bc.lock.Unlock()
+	cachedVal, ok := bc.cache[key]
+	return cachedVal, ok
+}
+
+func (bc *bugzillaCache) Set(key string, respBytes []byte) {
+	bc.lock.Lock()
+	defer bc.lock.Unlock()
+	bc.cache[key] = respBytes
+}
+func newBugzillaCache() *bugzillaCache {
+	return &bugzillaCache{cache: map[string][]byte{}}
+}
+
+// CachedTransport is an implementation http.RoundTripper
+// which first checks for cached values
+type CachedTransport struct {
+	mirror    *bugzillaCache
+	Transport http.RoundTripper
+}
+
+// RoundTrip will first check if there are any cached responses and return that
+// if not it will make an HTTP call using the DefaultTransport
+func (t *CachedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Method != "GET" {
+		return http.DefaultTransport.RoundTrip(req)
+	}
+	var err error
+	var resp *http.Response
+	g := new(errgroup.Group)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	g.Go(func() error {
+		defer wg.Done()
+		resp, err = http.DefaultTransport.RoundTrip(req)
+		if err != nil {
+			return err
+		}
+		body, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			return fmt.Errorf("err while prepping response to cache: %v", err)
+		}
+		t.mirror.Set(req.URL.String(), body)
+		return nil
+	})
+
+	if cachedVal, isCached := t.mirror.Get(req.URL.String()); isCached {
+		b := bytes.NewBuffer(cachedVal)
+		return http.ReadResponse(bufio.NewReader(b), req)
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+// Client returns an HTTP Client with the CachedTransport as the Transport
+func (t *CachedTransport) Client() *http.Client {
+	return &http.Client{
+		Transport: t,
+	}
+}
+
+// NewCachedTransport is a constructor for CachedTransport
+func NewCachedTransport() *CachedTransport {
+	return &CachedTransport{
+		mirror: newBugzillaCache(),
+	}
+}
--- a/pkg/backporter/httpcache.go
+++ b/pkg/backporter/httpcache.go
@@ -7,7 +7,9 @@ import (
 	"net/http"
 	"net/http/httputil"
 	"sync"
+	"time"
 
+	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/errgroup"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+	quit := make(chan struct{})
+	go func() {
+		for {
+			select {
+			case <-ticker.C:
+				refreshCache(t.mirror)
+			case <-quit:
+				ticker.Stop()
+				return
+			}
```
GeoBK, line 127: @alvaroaleman  Not sure if this is the best way to refresh the cache periodically. Any suggestions?

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+	delete(bc.cache, key)
+}
+func newBugzillaCache() *bugzillaCache {
+	return &bugzillaCache{cache: map[string][]byte{}}
+}
+
+// CachedTransport is an implementation http.RoundTripper
+// which first checks for cached values
+type CachedTransport struct {
+	mirror    *bugzillaCache
```
alvaroaleman, line 45: please rename this to `cache`

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+	g := new(errgroup.Group)
+	var wg sync.WaitGroup
+	wg.Add(1)
+	g.Go(func() error {
+		defer wg.Done()
+		resp, err = http.DefaultTransport.RoundTrip(req)
+		if err != nil {
+			return err
+		}
+		body, err := httputil.DumpResponse(resp, true)
```
alvaroaleman, line 66: We need to check for the status code first, there is no point in caching non-200 statuses

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+// RoundTrip will first check if there are any cached responses and return that
+// if not it will make an HTTP call using the DefaultTransport
+func (t *CachedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Method != "GET" {
+		return http.DefaultTransport.RoundTrip(req)
+	}
+	var err error
+	var resp *http.Response
+	g := new(errgroup.Group)
+	var wg sync.WaitGroup
```
alvaroaleman, line 58: Did you intend this to be an `errorgroup`?

pkg/backporter/httpcache.go
alvaroaleman: `quit` will never get closed so having it is useless. You can just do a `defer ticker.Stop()` instead. Other than that this looks fine.

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+type CachedTransport struct {
+	mirror    *bugzillaCache
+	Transport http.RoundTripper
+}
+
+// RoundTrip will first check if there are any cached responses and return that
+// if not it will make an HTTP call using the DefaultTransport
+func (t *CachedTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Method != "GET" {
+		return http.DefaultTransport.RoundTrip(req)
```
alvaroaleman, line 53: In order to make this testable it makes sense to make the http roundtripper a property of `CachedTransport` rather than hardcoding it to the `http.DefaultTransport` (and default it to `http.DefaultTransport` in the constructor)

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func refreshCache(bc *bugzillaCache) {
+	for url := range bc.cache {
+		url := url
+		go func() {
+			resp, err := http.Get(url)
+			if err != nil {
+				bc.Delete(url)
```
alvaroaleman, line 97: Why delete the old entry just because a http request failed?

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+func (bc *bugzillaCache) Delete(key string) {
+	bc.lock.Lock()
+	defer bc.lock.Unlock()
+	delete(bc.cache, key)
+}
+func newBugzillaCache() *bugzillaCache {
+	return &bugzillaCache{cache: map[string][]byte{}}
+}
+
+// CachedTransport is an implementation http.RoundTripper
```
alvaroaleman, line 42: Rename to `CachingTransport`? The thing itself doesn't get cached, it does do the caching

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,131 @@
...
+				// log these errors as well
+			}
+			bc.Set(url, body)
+		}()
+	}
+
+}
+
+// NewCachedTransport is a constructor for CachedTransport
+func NewCachedTransport() *CachedTransport {
```
alvaroaleman, line 113: I would just make this return a `http.RoundTripper` and unexport the `CachedTransport`. Also, this absolutely needs extended go docs on what it caches how and when

pkg/backporter/httpcache.go
GeoBK: yeah.. that was leftover from my synchronization implementation without errorgroups

pkg/backporter/httpcache.go
GeoBK: The reasoning behind that was basically if the Get request is returning an error, I probably want to invalidate that entry since there is probably some issue there, but I guess checking the status code and deleting should be more appropriate, right?

alvaroaleman: Yeah we can probably also store a 404

pkg/backporter/httpcache.go
GeoBK: And delete it when its a non 200 response?

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		body, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			t.cache.Delete(req.URL.String())
+			return fmt.Errorf("err while prepping response to cache: %w", err)
+		}
+		if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
+			t.cache.Set(req.URL.String(), body)
+		} else {
+			t.cache.Delete(req.URL.String())
```
alvaroaleman, line 73: Same here, it shouldn't delete just because a request failed

test/integration/pj-rehearse/expected.yaml
```diff
@@ -1760,7 +1760,7 @@
       prow.k8s.io/refs.pull: "1234"
       prow.k8s.io/refs.repo: release
       prow.k8s.io/type: presubmit
-    name: test-prowjob
+    name: 52b7ac34-ce00-11ea-b9f1-3e22fb21a9dc
```
alvaroaleman, line 1763: These fixture changes are wrong

pkg/backporter/httpcache.go
GeoBK: But wouldnt we be returning the wrong values if we return the old cached 200 values, even though the server is giving an error?

test/integration/pj-rehearse/expected.yaml
GeoBK: That happened when i ran the make update-integration.. I'll try running it again

GeoBK: i got rid of the update integration run changes.. and the integration seems to pass without that

alvaroaleman:
/retitle Add a caching transport for the Bugzilla client

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		body, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			t.cache.delete(req.URL.String())
+			return fmt.Errorf("err while prepping response to cache: %w", err)
+		}
+		if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
+			t.cache.set(req.URL.String(), body)
+		} else {
+			t.cache.delete(req.URL.String())
```
alvaroaleman, line 78: Deleting this here is wrong. Just keep whats in the cache, requests can fail for all kind of random reasons, there is no reason to delete a cache entry for every unanticipated error we get. If the entry is actually gone, we cache that

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+	if req.Method != "GET" {
+		return t.transport.RoundTrip(req)
+	}
+	var err error
+	var resp *http.Response
+	g := new(errgroup.Group)
+	g.Go(func() error {
+		resp, err = t.transport.RoundTrip(req)
+		if err != nil {
+			t.cache.delete(req.URL.String())
```
alvaroaleman, line 67: Don't delete here, networks and services may have hiccups, no reason to delete cache entries

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+			return err
+		}
+		body, err := httputil.DumpResponse(resp, true)
+		if err != nil {
+			t.cache.delete(req.URL.String())
+			return fmt.Errorf("err while prepping response to cache: %w", err)
```
alvaroaleman, line 73: s/prepping response to cache/serializing response

Also flow wise it makes more sense to first check the status code if we actually want to store this response in the cache before we serialize it rather than serializing unconditionally

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+
+func refreshCache(bc *bugzillaCache) {
+	concurrencyLimiter := make(chan string, 10)
+	for url := range bc.cache {
+		concurrencyLimiter <- url
+		go func() {
+			url := <-concurrencyLimiter
+			resp, err := http.Get(url)
+			if err != nil {
+				bc.delete(url)
```
alvaroaleman, line 106: Don't delete

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+				logrus.WithError(fmt.Errorf("cache refresh error - failed to fetch %s: %w", url, err))
+				return
+			}
+			defer resp.Body.Close()
+			body, err := httputil.DumpResponse(resp, true)
+			if err != nil {
+				bc.delete(url)
```
alvaroaleman, line 113: Don't delete

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+				logrus.WithError(fmt.Errorf("cache refresh error - DumpResponse failed %s: %w", url, err))
+				return
+			}
+			if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
+				bc.set(url, body)
+			} else {
+				bc.delete(url)
```
alvaroaleman, line 120: Don't delete

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+				return
+			}
+			defer resp.Body.Close()
+			body, err := httputil.DumpResponse(resp, true)
+			if err != nil {
+				bc.delete(url)
+				logrus.WithError(fmt.Errorf("cache refresh error - DumpResponse failed %s: %w", url, err))
+				return
+			}
+			if resp.StatusCode == http.StatusOK || resp.StatusCode == http.StatusNotFound {
```
alvaroaleman, line 117: Do this check before doing the `DumpResponse`

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+		return cachedResp, nil
+	}
+	if err := g.Wait(); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
+func refreshCache(bc *bugzillaCache) {
+	concurrencyLimiter := make(chan string, 10)
```
alvaroaleman, line 99: Would be nice to log here that we are doing our periodic refresh and how many bugs we have in the cache. The latter might also might sense as a metric

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+func TestBugzillaCache(t *testing.T) {
+	testcases := []struct {
+		name      string
+		cachedKey string
+		value     []byte
+		absentKey string
+	}{
+		{
+			name:      "key 1",
+			cachedKey: "key1",
```
alvaroaleman, line 22: Rather than having a `cachedKey`, `value` and `absentProperty` in the testcase it would be easier to just make a `cache` part of the testcase, that imediately shows whats in there and what not.

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+			value:     []byte("Test string"),
+			absentKey: "keyx",
+		},
+	}
```
alvaroaleman, line 26: There should be a testcase that checks that getting uncached responses works properly

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+			_, isAbsentKeyPresent := bc.get(tc.absentKey)
+			if isAbsentKeyPresent {
+				t.Errorf("invalid key marked ok - %s", tc.absentKey)
+			}
+
+		})
+	}
+}
+
+type transportMock struct {
```
alvaroaleman, line 52: usually, mocks are called `fakeTHING` in golang

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+	}{
+		{
+			name: "Succesful request to mock server",
+			tmock: transportMock{
+				response: &http.Response{
+					Status:     http.StatusText(http.StatusOK),
+					StatusCode: http.StatusOK,
+					Body:       ioutil.NopCloser(bytes.NewBuffer([]byte("some data"))),
+				},
+				err: nil,
```
alvaroaleman, line 74: You don't need to explicitly set fields to `nil`, they are by default (if they are pointers)

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,109 @@
...
+				transport: tc.tmock,
+			}
+			resp, err := tp.RoundTrip(r)
+			if err != tc.tmock.err {
+				t.Errorf("wrong error - expected %v, got %v", tc.tmock.err, err)
+			}
+			if resp != tc.tmock.response {
+				t.Errorf("incorrect response - expected %v, got %v", tc.tmock.response, resp)
+			}
+			resp, err = tp.RoundTrip(r)
```
alvaroaleman, line 99: Why do another `RoundTrip` call here? we already get the `resp` on line 92?

pkg/backporter/httpcache_test.go
GeoBK: The test does that, since the first stage requests data which is uncached. 
The next stage checks whether the data is cached and if the cached data is returned the second time that data is requested

pkg/backporter/httpcache.go
GeoBK: OK. At what point would I delete something from the cache if its no longer relevant?

pkg/backporter/httpcache_test.go
GeoBK: The first stage requests data which is uncached.
The next stage checks whether the data is cached and if the cached data is returned the second time that data is requested

pkg/backporter/httpcache.go
alvaroaleman: Never

pkg/backporter/httpcache_test.go
alvaroaleman: So two tests in one testcase?

GeoBK: I can move that out if thats what is needed

pkg/backporter/httpcache.go
GeoBK: I have added logging and metrics here

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,180 @@
...
+			postSetCache: map[string][]byte{"key": []byte("New data")},
+		},
+	}
+	for _, tc := range testcases {
+		t.Run(tc.name, func(t *testing.T) {
+			bc := newBugzillaCache()
+			bc.cache = tc.cache
+			bc.set(tc.key, tc.value)
+			if !reflect.DeepEqual(tc.postSetCache, bc.cache) {
+				t.Errorf("cached entry mismatch")
```
alvaroaleman, line 43: We usually do this kind of check by producing a diff and verifying that the diff is empty, the current approach makes it impossible to debug without a debugger: https://github.com/openshift/ci-tools/blob/196398e6cfff28d99285ff22fab979b243ff6e60/pkg/defaults/defaults_test.go#L423

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,180 @@
...
+			}
+
+		})
+	}
+
+}
+func TestBugzillaCacheGet(t *testing.T) {
```
alvaroaleman, line 50: Not sure if this test is needed, everything we test here is also tested in `TestRoundTrip` already I think?

pkg/backporter/httpcache.go
alvaroaleman: Would you mind to use a field for this:
`logrus.WithField("cache_entries", len(bc.cache)).Info("Refreshing cache")`

This has the advantage that we can easely filter for this in the logs later on, because we can filter for the `"Refreshing cache"` message, right now we would need a regex

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,146 @@
...
+// while also generating an async HTTP call to the bugzilla server to get the latest value
+// which is stored in the cache.
+// Therefore this cache does *NOT* reduce the HTTP traffic, and is only used to speed up the response.
+func NewCachingTransport() http.RoundTripper {
+	t := cachingTransport{
+		cache:     newBugzillaCache(),
+		transport: http.DefaultTransport,
+	}
+	cacheRefreshMetrics := prometheus.NewHistogram(
+		prometheus.HistogramOpts{
```
alvaroaleman, line 130: This is not a histogram, its a `counter`: https://prometheus.io/docs/concepts/metric_types/

pkg/backporter/httpcache_test.go
GeoBK: It's not exactly the same test, and as you mentioned in a previous comment having a separate test for the Get function keeps with the spirit of unit testing I guess. I can remove them if you strongly feel that its unnecessary

pkg/backporter/httpcache.go
GeoBK: I do not understand why, could you please explain that?
Since I was visualizing that the metrics would keep track of how many bugs were refreshed during each refresh and not just being incremented everytime a call was made.

alvaroaleman: Check out the link above, it has some nice explanations for the different types

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,180 @@
...
+			tp := cachingTransport{
+				cache:     bc,
+				transport: tc.fake,
+			}
+			resp, err := tp.RoundTrip(r)
+			if err != tc.fake.err {
+				t.Errorf("wrong error - expected %v, got %v", tc.fake.err, err)
+			}
+			if tc.isCached {
+				respFromCache := resp.Header.Get(xFromCache)
```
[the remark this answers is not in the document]
alvaroaleman: Actually I am realizing we need to do this check regardless of the test having `isCached` set. If its not set, the headers value must be an empty string

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,179 @@
...
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				t.Fatalf("failed to parse response body: %v", err)
+			}
+			if string(body) != bodyText {
+				t.Errorf("incorrect body - expected %v, got %v", bodyText, string(body))
+			}
+			if tc.isCached {
+				respFromCache := resp.Header.Get(xFromCache)
+				if respFromCache != "1" {
```
alvaroaleman, line 172: @GeoBK I can not unresolve the conversation above but did you see the second coment? That one is still open

GeoBK: Yeah, I saw that 👍

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,145 @@
...
+}
+
+func refreshCache(bc *bugzillaCache, m prometheus.Gauge) {
+	concurrencyLimiter := make(chan string, 10)
+	logrus.WithField("cache_entries", len(bc.cache)).Info("Refreshing cache")
+	m.Set(float64(len(bc.cache)))
+	for url := range bc.cache {
+		concurrencyLimiter <- url
+		go func() {
+			url := <-concurrencyLimiter
```
stevekuznetsov, line 99: Why this over a semaphore? This is implicit, a semaphore explicit in intent

pkg/backporter/httpcache.go
```diff
@@ -0,0 +1,145 @@
...
+// If an entry is present in the cache, it is immediately returned
+// while also generating an async HTTP call to the bugzilla server to get the latest value
+// which is stored in the cache.
+// Therefore this cache does *NOT* reduce the HTTP traffic, and is only used to speed up the response.
+func NewCachingTransport() http.RoundTripper {
+	t := cachingTransport{
+		cache:     newBugzillaCache(),
+		transport: http.DefaultTransport,
+	}
+	cacheRefreshMetrics := prometheus.NewGauge(
```
stevekuznetsov, line 129: I would expect to also expose Bugzilla API HTTP metrics like the client does

pkg/backporter/httpcache.go
GeoBK: Yeah - a semaphore is probably appropriate here. I'll change that

pkg/backporter/httpcache.go
GeoBK: I didn't understand this comment. Did you mean keeping track of the individual calls made to the Bugzilla API by the refresh function?

pkg/backporter/httpcache_test.go
```diff
@@ -0,0 +1,184 @@
...
+			}
+			if resp.StatusCode != tc.fake.response.StatusCode {
+				t.Errorf("incorrect status code - expected %v, got %v", tc.fake.response.StatusCode, resp.StatusCode)
+			}
+			defer resp.Body.Close()
+			body, err := ioutil.ReadAll(resp.Body)
+			if err != nil {
+				t.Fatalf("failed to parse response body: %v", err)
+			}
+			if string(body) != bodyText {
```
GeoBK, line 167: @alvaroaleman  does this have the potential to be a flaky test. This has failed once and I'm not sure why. ie the way in which I'm extracting "body" from response, does that look ok?

alvaroaleman: I don't see how this could flake

pkg/backporter/httpcache_test.go
alvaroaleman: I think this can be better expressed as:
```
if isRespFromCache := resp.Header.Get(xFromCache) == "1"; isResponseFromCache != tc.isCached {
  t.Errorf("Expected resp to be from cache: %t, response was from cache: %t", tc.isCached, isResponseFromCache)
}
```

pkg/backporter/httpcache_test.go
alvaroaleman: Do you have a link to the flake?

GeoBK: I dont have a link to it but basically all the other runs (out of the 24?) passed except one which failed with "body" being empty.

GeoBK: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/1015/pull-ci-openshift-ci-tools-master-unit/1288173736800817152

PR #1019, opened 2020-07-22
determinize-step-registry: create new tool

AlexNPavel:
This PR adds a `determinize-step-registry` tool that performs a similar
task to the other determinize tools in this repo. It also adds a new
metadata field to the registry components that will be used by the web
UI to generate github links and get the OWNERS files for registry
components.

This PR also makes the `XSuffix` consts in the load package public and
replaces all previous hardcoded strings with those consts.

### DIFF
--- a/cmd/determinize-step-registry/main.go
+++ b/cmd/determinize-step-registry/main.go
@@ -1,0 +1,107 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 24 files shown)


petr-muller:
@bbguimaraes the context is mentioned in the description. We want to generate links to GH from `steps.ci` pages. For that we need the path into the repo, so we need to add it to the metadata, because it's not otherwise available (`steps.ci` only has the flattened CM). For that you need to machine-write the metadata into the YAML, which reorders the fields inside...

I wonder if we could build a separate index for the step->path mapping? That way we would not need to machine-write to the step YAMLs...

AlexNPavel:
Will reimplement as an registry index file in different PR.

/close

PR #1023, opened 2020-07-24
pkg/load: partially validate registry contents

bbguimaraes:
Full validation can only be done in the presence of test configurations,
but many useful validations can be performed when the registry is initially
loaded.

There is probably a more efficient algorithm, but this way we maintain a
single execution path for all validation phases.  Validating the entire
step registry in `openshift/release` still stays below 1s on my machine and
I don't see any change in total execution time.

The original bug report has been added as an integration tests.  I've also
verified that the new code correctly identifies the error:

```
$ cd openshift/release/
$ git checkout --quiet 67cb826934
$ ci-operator-configresolver --validate-only …
INFO[0000] Configs reloaded                             duration=683.772282ms
FATA[0000] Failed to get registry agent: Failed to load registry: failed to load ci-operator registry ([openshift-e2e-aws-loki: duplicate name: manifests, openshift-e2e-gcp-loki: duplicate name: manifests])
```

***

/hold

Opening for review, but I want to fix the bug that is causing integration tests
to fail intermittently first.  Some of these commits may be moved to separate
PRs in the process.

### DIFF
--- a/test/integration/ci-operator-configresolver.sh
+++ b/test/integration/ci-operator-configresolver.sh
@@ -1,9 +1,7 @@
 #!/bin/bash
 source "$(dirname "${BASH_SOURCE}")/../../hack/lib/init.sh"
 
-rsync_tmp_dir=$(mktemp -d)
 function cleanup() {
-    rm -rf $rsync_tmp_dir
     os::test::junit::reconcile_output
     os::integration::configresolver::check_log
     os::cleanup::processes
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -200,8 +200,6 @@ func main() {
 	}
 	level, _ := logrus.ParseLevel(o.logLevel)
 	logrus.SetLevel(level)
-	health := pjutil.NewHealth()
-	metrics.ExposeMetrics("ci-operator-configresolver", prowConfig.PushGateway{}, flagutil.DefaultMetricsPort)
 
 	configAgent, err := agents.NewConfigAgent(o.configPath, agents.WithConfigMetrics(configresolverMetrics.ErrorRate))
 	if err != nil {
@@ -216,7 +214,8 @@ func main() {
 	if o.validateOnly {
 		os.Exit(0)
 	}
-
+	health := pjutil.NewHealth()
+	metrics.ExposeMetrics("ci-operator-configresolver", prowConfig.PushGateway{}, flagutil.DefaultMetricsPort)
 	simplifier := simplifypath.NewSimplifier(l("", // shadow element mimicing the root
 		l("config"),
 		l("resolve"),
--- a/pkg/load/load.go
+++ b/pkg/load/load.go
@@ -329,7 +329,10 @@ func Registry(root string, flat bool) (registry.ReferenceByName, registry.ChainB
 		logrus.WithError(err).Warn("failed to parse metadata file")
 	}
 	// create graph to verify that there are no cycles
-	_, err = registry.NewGraph(references, chains, workflows)
+	if _, err = registry.NewGraph(references, chains, workflows); err != nil {
+		return nil, nil, nil, nil, nil, err
+	}
+	err = registry.Validate(references, chains, workflows)
 	return references, chains, workflows, documentation, metadata, err
 }
 
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -17,6 +17,29 @@ type ReferenceByName map[string]api.LiteralTestStep
 type ChainByName map[string]api.RegistryChain
 type WorkflowByName map[string]api.MultiStageTestConfiguration
 
+// Validate verifies the internal consistency of steps, chains, and workflows.
+// A superset of this validation is performed later when actual test
+// configurations are resolved.
+func Validate(stepsByName ReferenceByName, chainsByName ChainByName, workflowsByName WorkflowByName) error {
+	reg := registry{stepsByName, chainsByName, workflowsByName}
+	var ret []error
+	for k := range chainsByName {
+		if _, err := reg.process([]api.TestStep{{Chain: &k}}, sets.NewString(), stackForChain()); err != nil {
+			ret = append(ret, err...)
+		}
+	}
+	for k, v := range workflowsByName {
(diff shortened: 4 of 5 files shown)


bbguimaraes:
We need to improve this error message too:

```
Running hack/lib/integration.sh:79: executing 'curl http://127.0.0.1:8081/healthz/ready' expecting any result and text 'OK'; re-trying every 0.2s until completion or 60.000s...
FAILURE after 59.465s: hack/lib/integration.sh:79: executing 'curl http://127.0.0.1:8081/healthz/ready' expecting any result and text 'OK'; re-trying every 0.2s until completion or 60.000s: the command timed out
Standard error from the command:
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed

  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (7) Failed connect to 127.0.0.1:8081; Connection refused
... repeated 143 times
[ERROR] PID 41: hack/lib/cmd.sh:617: `return "${return_code}"` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: hack/lib/cmd.sh:617: `return "${return_code}"`
[INFO] 		  2: hack/lib/integration.sh:79: os::cmd::try_until_text
[INFO] 		  3: hack/lib/integration.sh:61: os::integration::configresolver::wait_for_ready
[INFO] 		  4: test/e2e/multi-stage.sh:16: os::integration::configresolver::start
[INFO]   Exiting with code 1. 
```

test/integration/ci-operator-configresolver.sh
```diff
@@ -1,65 +1,88 @@
...
+    # Duplicate names at resolve time generate an error.
+    os::cmd::expect_code_and_text \
+        "$(echo \
+            ${cmd[*]} --registry ${suite_dir}/validation/duplicate/ \
+            '|&' tee ${out}/duplicate.log)" \
+        1 'openshift-e2e-aws-loki: duplicate name: manifests'
+}
+test_validation
+
+mkdir -p "${BASETMPDIR}/runtime/"
```
stevekuznetsov, line 39: Can we move the other way, so we don't need to change anything below?

test/integration/ci-operator-configresolver.sh
```diff
@@ -1,65 +1,88 @@
...
-actual="${BASETMPDIR}/actual"
-mkdir -p "${actual}"
-expected="${suite_dir}/expected"
+unused_dir=${suite_dir}/unused
 
 os::test::junit::declare_suite_start "integration/ci-operator-configresolver"
-# This test validates the ci-operator-configresolver tool
 
-os::integration::configresolver::start "${BASETMPDIR}/configs" "${BASETMPDIR}/multistage-registry/registry" "${BASETMPDIR}/config.yaml"
+test_validation() {
```
stevekuznetsov, line 16: Why is this a function?

test/integration/ci-operator-configresolver.sh
```diff
@@ -1,65 +1,88 @@
...
+    local cmd=(
+        ci-operator-configresolver
+        --config "${unused_dir}/"
+        --jobs "${unused_dir}/"
+        --prow-config "${unused_dir}/unused.md"
+        --validate-only)
+    # Validate empty registry directory.
+    os::cmd::expect_success "${cmd[*]} --registry ${unused_dir}/"
+    # Unresolved parameters are valid at this stage.
+    os::cmd::expect_success \
```
stevekuznetsov, line 27: We don't care about line width

test/integration/ci-operator-configresolver.sh
stevekuznetsov: maybe

```
os::test::junit::declare_suite_start "integration/ci-operator-configresolver/validation"
```

test/integration/ci-operator-configresolver.sh
```diff
@@ -1,65 +1,88 @@
...
+        "${cmd[*]} --registry ${suite_dir}/validation/unresolved/"
+    mkdir -p "${out}"
+    # Duplicate names at resolve time generate an error.
+    os::cmd::expect_code_and_text \
```
stevekuznetsov, line 31: This is crazy -- IMHO we do not gain much from this test and we should just omit the integration stuff entirely (it is checked at the unit level totally fine)

test/integration/ci-operator-configresolver.sh
bbguimaraes: I made it a function just to isolate the local variables from the rest of the script.

The two types of tests are different enough that I considered making a separate script, but didn't know exactly what file/directory structure to use.  WDYT?

test/integration/ci-operator-configresolver.sh
bbguimaraes: I do, but more importantly, if the comment applies to all commands broken over multiple lines in this file: I find it much harder to read long lines with several flags/arguments.

(but I guess I'll have to live with it)

test/integration/ci-operator-configresolver.sh
bbguimaraes: I don't understand what is "crazy" about it.  It adds an actual bug report as an integration test.

We have found more than one bug exclusively via registry integration tests, and potentially have one right now (I haven't been able to track it yet).

stevekuznetsov: The crazy part is how complicated the test is

test/integration/ci-operator-configresolver.sh
stevekuznetsov: Can we expense you a 4k monitor? :)

test/integration/ci-operator-configresolver.sh
stevekuznetsov: Do something to make it look as similar to the rest as you can, if that means a new script, sure, it's light-weight

test/integration/ci-operator-configresolver.sh
stevekuznetsov: Like IMHO testing the content of a specific error mesage is *so* much lower level than a full-blown run-the-binary integration test. What, specifically, does this test add that well-formed unit tests in Go could not?

bbguimaraes:
Updated.

PR #1029, opened 2020-07-27
generate-registry-metadata: create new tool

AlexNPavel:
This PR adds a generate-registry-metadata tool that create a
`metadata.json` file containing a mapping of filenames to full paths
that will be used by the configresolver's web UI to provide github links
to components and retrieve OWNERS information for the components.


cmd/generate-registry-metadata/main.go
```diff
@@ -0,0 +1,79 @@
...
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.registry, "registry", "", "Path to the step registry directory.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatal("could not parse input")
+	}
+	return o
+}
+
+func updateMetadata(registryPath string) (api.RegistryMetadata, error) {
```
stevekuznetsov, line 38: nit: this generates, but does not update, the metadata

cmd/generate-registry-metadata/main.go
```diff
@@ -0,0 +1,79 @@
...
+	if err := filepath.Walk(registryPath, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info != nil && !info.IsDir() && filepath.Ext(info.Name()) == ".yaml" {
+			relpath, err := filepath.Rel(registryPath, path)
+			if err != nil {
+				return fmt.Errorf("failed to determine relative path for %s: %w", path, err)
+			}
+			metadata.ComponentPath[info.Name()] = filepath.Dir(relpath)
```
stevekuznetsov, line 50: If this is all we store, we will need to do GitHub (?) calls to get the OWNERS? Or are you expecting to clone the repo? Why not store a 

```go
type info struct {
    Path string
    Owners owners.FullConfig
}

var metadata map[string]info

Comment (AlexNPavel) [reply cmd/generate-registry-metadata/main.go:50]: I was originally planning to just get the `OWNERS` files via the `raw.githubusercontent` pages, but that was mainly because I didn't want to clutter the component files too much since it could be confusing for users. Since this is now being handled as a separate metadata file, it would make sense to store them in the metadata.

**Comment by AlexNPavel** [comment]
@stevekuznetsov PR updated to include `OWNERS` information

File: test/multistage-registry/registry/ipi/OWNERS [hunk 3]
```diff
@@ -0,0 +1,7 @@
+approvers:
```
alvaroaleman, line 1: Could we not re-use these folders? The fact that they are used within multiple tests makes any changes here very hard and IMHO everything needed for any test should only be used by that test (Copying everything to a new dir is totally fine)

cmd/generate-registry-metadata/main.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return nil
+}
+
+func gatherOptions() options {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.registry, "registry", "", "Path to the step registry directory.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatal("could not parse input")
```
alvaroaleman, line 35: Nit: Don't exit the application outside of `function main`

cmd/generate-registry-metadata/main.go
```diff
@@ -0,0 +1,96 @@
+package main
```
alvaroaleman, line 1: Could we get a brief readme? :)

AlexNPavel:
@alvaroaleman updated

PR #1030, opened 2020-07-27
pkg/steps/artifacts: Wait for running artifacts container

alvaroaleman:
Supersedes and thereby closes https://github.com/openshift/ci-tools/pull/958

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -23,6 +23,7 @@ import (
 	"k8s.io/apimachinery/pkg/fields"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"k8s.io/client-go/kubernetes/scheme"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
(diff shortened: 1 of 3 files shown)


pkg/steps/multi_stage_test.go
```diff
@@ -547,6 +548,9 @@ func TestRun(t *testing.T) {
 }
 
 func TestArtifacts(t *testing.T) {
+	timeSecond = time.Nanosecond
+	defer func() { timeSecond = time.Second }()
```
bbguimaraes, line 552: Shouldn't this result in an error?

alvaroaleman: It does but we only log it: https://github.com/openshift/ci-tools/pull/1030/files#diff-892295443de5592354f856e247809216R452

alvaroaleman: github is so great, doesn't expand the are the link is in, this one should work: https://github.com/openshift/ci-tools/blob/77d4e0ac850628341ffc1075527bee60a4d9dbc5/pkg/steps/artifacts.go#L452

pkg/steps/artifacts.go
```diff
@@ -216,6 +217,36 @@ type PodClient interface {
...
+var timeSecond = time.Second
+
+func waitForContainer(podClient PodClient, ns, name, containerName string) error {
+	logrus.WithFields(logrus.Fields{
+		"namespace": ns,
+		"name":      name,
+		"container": containerName,
+	}).Trace("Waiting for container to be running.")
+
+	return wait.PollImmediate(time.Second, 30*timeSecond, func() (bool, error) {
```
stevekuznetsov, line 230: Could e.g. volume attach take longer than 30s? Do we care?

alvaroaleman: Probably, the 30 seconds was a random guess. Failures here are non-terminal though, we just log them. 30 seconds is definitely better than not waiting at all. We can use ci-search to figure out if this is too little later on and if yes finetune

PR #1032, opened 2020-07-27
pkg/webreg: Document openshift-priv management via private-prow-configs-mirror

wking:
@alvaroaleman suggested this location for documentation, even though this information is not multi-step specific.  openshift/release#10479 should at least make it more discoverable.

/assign @alvaroaleman

[1]: https://github.com/openshift/release/pull/10479


alvaroaleman:
@wking I am not sure if having a `auto generated page` is the best approach, what about adding a `private` repos page that mentions that everything there is autogenerated?

wking:
> ... what about adding a private repos page that mentions that everything there is autogenerated?

I think this will make it too difficult for openshift/release contributors to answer:

    I am thinking of touching $PART of $PATH. Is that auto-generated?

But I've pushed 1e2db2ef7 -> aaa655d05 pivoting to this approach.  We can always consolidate later if we get too many tiny pages.

alvaroaleman:
Thanks @wking !

PR #1034, opened 2020-07-28
rehearse: verify the test exists when inlining

bbguimaraes:
Should not happen with generated jobs, but it is an open interface so it
was just a matter of time:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/10191/pull-ci-openshift-release-master-pj-rehearse/1288013504346853376


bbguimaraes:
/hold

apparently `[images]` and other targets go through the same path, this will need changes.

bbguimaraes:
/close

This would require too much surgery to support template tests, which do not have a correspondent entry in the configuration file.

PR #1036, opened 2020-07-28
Add ocp-build-data enforcer

alvaroaleman:
This PR adds the ocp-build data enforcer. It does not yet:
* Properly dereference
* Create PRs
* Work on multiple branches

It does:
* Update Dockerfiles without messing them up
* Have the basic types needed to parse the data on ocp-build-data
* Read the data
* Do major/minor string replacing in the yaml files before unmarshaling them


cmd/ocp-build-data-enforcer/main.go
```diff
@@ -0,0 +1,304 @@
...
+		if err := dereferenceConfig(&cfg, streamMap, groupYAML); err != nil {
+			errs = append(errs, fmt.Errorf("failed dereferencing config for %s: %w", cfg.SourceFileName, err))
+			continue
+		}
+		configs = append(configs, cfg)
+	}
+
+	errGroup := &errgroup.Group{}
+	for idx := range configs {
+		idx := idx
```
stevekuznetsov, line 72: This is super opaque -- why not 

```go
for i := range configs {
    errGroup.Go(func(config ocpImageConfig) error {
        return processDockerfile(config)
    }(configs[i]))
}
```

cmd/ocp-build-data-enforcer/main.go
```diff
@@ -0,0 +1,304 @@
...
+	for blder := range config.From.Builder {
+		if replacement, hasReplacement := streamMap[config.From.Builder[blder].Stream]; hasReplacement {
+			config.From.Builder[blder].Stream = replacement.UpstreamImage
+		}
+	}
+	if config.Content.Source.Alias != "" {
+		if _, hasReplacement := groupYAML.Sources[config.Content.Source.Alias]; !hasReplacement {
+			return fmt.Errorf("groups.yaml has no replacement for alias %s", config.Content.Source.Alias)
+		}
+		config.Content.Source.Git = &ocpImageConfigSourceGit{}
```
stevekuznetsov, line 105: Why this line and the following? I don't understand this code

cmd/ocp-build-data-enforcer/main.go
```diff
@@ -0,0 +1,304 @@
...
+
+func processDockerfile(config ocpImageConfig) {
+	orgRepo := config.orgRepo()
+	log := logrus.WithField("file", config.SourceFileName).WithField("org/repo", orgRepo)
+	split := strings.Split(orgRepo, "/")
+	if n := len(split); n != 2 {
+		log.Errorf("splitting orgRepo didn't yield 2 but %d results", n)
+		return
+	}
+	if split[0] == "openshift-priv" {
```
stevekuznetsov, line 120: So it's not that we need to handle updating Dockerfiles in private forks for public repos, but for e.g. `openshift/origin` I think there is not and will not be an entry in `ocp-build-data`. In that case we need to find the actual upstream we need to update by dereferencing (with the publicize config?). We should check that we understand this workflow with Justin.

cmd/ocp-build-data-enforcer/main.go
```diff
@@ -0,0 +1,304 @@
...
+	diffStr, err := difflib.GetUnifiedDiffString(diff)
+	if err != nil {
+		log.WithError(err).Error("Failed to construct diff")
+	}
+	log.Infof("Diff:\n---\n%s\n---\n", diffStr)
+}
+
+func readStreamMap(ocpBuildDataDir string, majorMinor majorMinor) (streamMap, error) {
+	streamMap := &streamMap{}
+	return *streamMap, readYAML(filepath.Join(ocpBuildDataDir, "streams.yml"), streamMap, majorMinor)
```
stevekuznetsov, line 162: Maybe this is just a personal preference, but I think I prefer to do the pointers like this:

```go
	streamMap := streamMap{}
	return streamMap, readYAML(filepath.Join(ocpBuildDataDir, "streams.yml"), &streamMap, majorMinor)
```

IMHO this is conceptually cleaner (and similar to the "borrowing" concept in Rust and how pointers are often thought about) -- we pass a reference to the function that needs to borrow the value and edit it, but we work with the value itself when we are creating and returning.

cmd/ocp-build-data-enforcer/main.go
```diff
@@ -0,0 +1,304 @@
...
+}
+
+type majorMinor struct{ major, minor string }
+
+func readYAML(path string, unmarshalTarget interface{}, majorMinor majorMinor) error {
+	data, err := ioutil.ReadFile(path)
+	if err != nil {
+		return fmt.Errorf("failed to read %s: %w", path, err)
+	}
+	data = bytes.ReplaceAll(data, []byte("{MAJOR}"), []byte(majorMinor.major))
```
stevekuznetsov, line 177: This is fairly bizarre - does the ART-managed `ocp-build-data` loader do this, too?

stevekuznetsov: Or, like, why do these fields embed other fields that need to be replace at load-time, instead of just being partial fields and having formatting/creation/etc done as necessary before use?

cmd/ocp-build-data-enforcer/main_test.go
```diff
@@ -0,0 +1,144 @@
...
+)
+
+func TestUpdateDockerfile(t *testing.T) {
+	testCases := []struct {
+		name              string
+		dockerfile        []byte
+		config            ocpImageConfig
+		expectedErrMsg    string
+		expectUpdate      bool
+		expectdDockerfile []byte
```
stevekuznetsov, line 16: `s/expectd/expected/g`

cmd/ocp-build-data-enforcer/main_test.go
```diff
@@ -0,0 +1,144 @@
...
+COPY --from=rhel7 /go/src/github.com/dougbtv/whereabouts/bin/whereabouts /usr/src/whereabouts/rhel7/bin
+COPY --from=rhel7 /go/src/github.com/dougbtv/whereabouts/bin/whereabouts /usr/src/whereabouts/bin
+COPY --from=rhel8 /go/src/github.com/dougbtv/whereabouts/bin/whereabouts /usr/src/whereabouts/rhel8/bin
+
+LABEL io.k8s.display-name="Whereabouts CNI" \
+      io.k8s.description="This is a component of OpenShift Container Platform and provides a cluster-wide IPAM CNI plugin." \
+      io.openshift.tags="openshift" \
+      maintainer="CTO Networking <nfvpe-container@redhat.com>"`),
+			config: ocpImageConfig{From: ocpImageConfigFrom{
+				Builder: []ocpImageConfigFromStream{{"rhel-8-golang"}, {"golang"}},
```
stevekuznetsov, line 52: Shouldn't a linter be upset with struct literals that don't name their fields?

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: because errGroup.Go takes a `func() error` as first argument: https://godoc.org/golang.org/x/sync/errgroup#Group.Go

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: The field is a pointer but I don't want to create a pointer to `groupYAML.Sources[config.Content.Source.Alias]` so I first set it to an empty object, then set the value of that empty object to `groupYAML.Sources[config.Content.Source.Alias]`

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: They have a mapping for that which I will be using in the other PR: https://github.com/openshift/ocp-build-data/blob/e8c46fac4007e65ed13338ec0c29633119a5d417/group.yml#L74

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: Yeah they do that, no idea why it was chosen to go down this route.

cmd/ocp-build-data-enforcer/main_test.go
alvaroaleman: No, why?

alvaroaleman:
@stevekuznetsov updated, PTAL

cmd/ocp-build-data-enforcer/main.go
stevekuznetsov: ew. Maybe a double closure? Or some comment -- the code as written is non-obvious

cmd/ocp-build-data-enforcer/main.go
stevekuznetsov: Why is the field a pointer?

cmd/ocp-build-data-enforcer/main_test.go
stevekuznetsov: I swear that is part of `go vet`...

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: Sorry, what is non-obvious about this?

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: To be able to tell if it was set in the yaml

cmd/ocp-build-data-enforcer/main.go
alvaroaleman: I find the whole passing it in as argument to copy much harder to read than this

PR #1037, opened 2020-07-29
Enable allow_skip_on_success also for non-literal MultiStageTestConfi…

mgencur:
…guration

* also make this argument optional for literal
MultiStageTestConfiguration

The test on https://github.com/openshift/ci-tools/blob/master/test/e2e/multi-stage/config.yaml#L34 was passing because the optional step was NOT skipped but it failed to create the file via `touch ${ARTIFACT_DIR}/file`. I've changed that to use `echo` and $SHARED_DIR and separated the test to check both conditions.

### DIFF
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -59,9 +59,13 @@ func (r *registry) Resolve(name string, config api.MultiStageTestConfiguration) 
 				config.Environment[k] = v
 			}
 		}
+		if config.AllowSkipOnSuccess == nil {
+			config.AllowSkipOnSuccess = workflow.AllowSkipOnSuccess
+		}
 	}
 	expandedFlow := api.MultiStageTestConfigurationLiteral{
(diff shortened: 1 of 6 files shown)


pkg/registry/resolver.go
```diff
@@ -59,9 +59,13 @@ func (r *registry) Resolve(name string, config api.MultiStageTestConfiguration)
 				config.Environment[k] = v
 			}
 		}
+		if config.AllowSkipOnSuccess == nil {
```
stevekuznetsov, line 62: Why only when is this not `nil`?

mgencur: Because if you define it at this level you want to override what was set in the "workflow" that you're reusing. At least that's my understanding. You can be re-using a workflow that has some settings but you want to be able to override them. At least in this case.

mgencur: If you don't define it directly in the multi-step test you'll get what is in the workflow.

PR #1052, opened 2020-07-30
Parallelize config loading

alvaroaleman:
This round about halves the time needed to load the config which is great to speed up inner dev cycle


bbguimaraes:
/lgtm
/retest

It would be good to merge this and `pkg/config/load`, my first test did not work as expected because I used a tool that loads these files via a separate code path =/

bbguimaraes:
https://issues.redhat.com/browse/DPTP-1396

PR #1056, opened 2020-07-31
Fix pod timeout check

alvaroaleman:


bbguimaraes:
Could we close the channel when we see a `Running` state so there aren't requests every minute unnecessarily?

PR #1060, opened 2020-08-03
Secret generator tool

GeoBK:
Tool to ease reduce the overhead while generating secrets and 
uploading to bitwarden.
This will also act as a form of documentation of the process for 
generating various secrets.


cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,232 @@
...
+			}
+		}
+
+		if bwItem.Field.Name != "" {
+			out, err := executeCommand(bwItem.Field.Cmd)
+			if err != nil {
+				return fmt.Errorf("failed to execute command: %w", err)
+			}
+			if dryRun {
+				if _, err := fmt.Fprintln(tmpFile, "\tField: "+string(out)); err != nil {
```
bbguimaraes, line 158: Use `Fprintf` instead of concatenating strings (same for the other cases below).

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,232 @@
...
+				return err
+			}
+		}
+		if dryRun {
+			if _, err := fmt.Fprintln(tmpFile, "Item: "+bwItem.ItemName); err != nil {
+				return err
+			}
+		}
+
+		if bwItem.Field.Name != "" {
```
bbguimaraes, line 152: We could DRY the contents of all these branches into something like:

```go
if bwItem.Field.Name != "" {
    processItem(tmpFile, bwItem.Field.Cmd, "field", dryRun, func() {
        if err = bwClient.SetFieldOnItem(bwItem.ItemName, bwItem.Field.Name, out); err != nil {
            return fmt.Errorf("failed to set field item: %s, field: %s - %w", bwItem.ItemName, bwItem.Field.Name, err)
        }
        return nil
    })
}
```

pkg/bitwarden/cli.go
```diff
@@ -142,3 +144,170 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
 func (c *cliClient) Logout() ([]byte, error) {
 	return c.run("logout")
 }
+
+func createItem(itemTemplate string, targetItem *Item, c *cliClient) error {
+	// the bitwarden cli expects the item to be base64 encoded
+	encItem := b64.StdEncoding.EncodeToString([]byte(itemTemplate))
+	out, err := c.run("--session", c.session, "create", "item", encItem)
```
bbguimaraes, line 151: It looks like we would benefit from a `runWithSession` method on `cliClient`.

pkg/bitwarden/cli.go
```diff
@@ -142,3 +144,170 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(out, targetItem)
+	if err != nil {
+		return err
+	}
+	return nil
```
bbguimaraes, line 159:
```suggestion
	return json.Unmarshal(out, targetItem)
```

pkg/bitwarden/cli.go
```diff
@@ -142,3 +144,170 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+}
+
+func createAttachment(fileContents []byte, fileName string, itemID string, newAttachment *Attachment, c *cliClient) error {
+	filePath := "/tmp/" + fileName
```
bbguimaraes, line 163:
```suggestion
	filePath := filepath.Join(os.TempDir(), fileName)
```

bbguimaraes: Can we have name clashes since we have a concurrency option?  We can use a random name if the file name is not important for the `create` command.  We also need to delete the temporary file.

pkg/bitwarden/cli.go
```diff
@@ -142,3 +144,170 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+
+func createEmptyItem(itemName string, targetItem *Item, c *cliClient) error {
+	itemTemplate := fmt.Sprintf(`{"type":1,"name":"%s","login":{}}`, itemName)
+	return createItem(itemTemplate, targetItem, c)
+}
+
+func createItemWithPassword(itemName string, password []byte, targetItem *Item, c *cliClient) error {
+	itemTemplate := fmt.Sprintf(`{"type":1,"name":"%s","login":{"password":"%s"}}`, itemName, string(password))
+	return createItem(itemTemplate, targetItem, c)
+}
```
bbguimaraes, line 186: Quotes are usually valid characters in passwords, we should go through the `json` library instead of using string formatting to be safe.

pkg/bitwarden/cli.go
```diff
@@ -142,3 +144,170 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
 func (c *cliClient) Logout() ([]byte, error) {
 	return c.run("logout")
 }
+
+func createItem(itemTemplate string, targetItem *Item, c *cliClient) error {
```
bbguimaraes, line 148: These can all be `cliClient` methods instead of taking a pointer argument.

pkg/bitwarden/cli.go
```diff
@@ -142,3 +144,170 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+			itemFound = true
+			targetItem = item
+			targetItemIndex = index
+			for fieldIndex, field := range item.Fields {
+				if field.Name == fieldName {
+					fieldFound = true
+					if field.Value != string(fieldValue) {
+						if err := editItem(targetItem, c); err != nil {
+							return fmt.Errorf("failed to set field, itemName: %s, fieldName: %s - %w", itemName, fieldName, err)
+						}
```
bbguimaraes, line 224: This can be handled outside the loop so we get a clear separation between the "search" and the "modify" phases.  We're also not `break`ing, so subsequent items are processed unnecessarily.

We can also use pointers instead of `found`/`target` pairs.

Similarly for the other functions below and the ones in `fake.go`.

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,232 @@
...
+	if o.bwPasswordPath == "" {
+		return fmt.Errorf("--bw-password-path is empty")
+	}
+	if o.configPath == "" {
+		return fmt.Errorf("--config is empty")
+	}
+	return nil
+}
+
+func (o *options) completeOptions(secrets *sets.String) error {
```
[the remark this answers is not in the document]
GeoBK: Yeah, makes sense

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,232 @@
...
+	}))
+	if err := o.validateOptions(); err != nil {
+		logrus.WithError(err).Fatal("invalid arguments.")
+	}
+	if err := o.completeOptions(&secrets); err != nil {
+		logrus.WithError(err).Fatal("failed to complete options.")
+	}
+	bwClient, err := bitwarden.NewClient(o.bwUser, o.bwPassword, func(s string) {
+		secrets.Insert(s)
+	})
```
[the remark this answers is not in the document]
GeoBK: The insert function's signature is slightly different from the func(string), since it returns a set and takes in variadic as input. Making this change would force me to change the function signature where its being defined even though the return type is unnecessary. Does it still make sense to proceed with this change?

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,232 @@
...
+			if err != nil {
+				logrus.WithError(err).Fatal("failed to create tempfile")
+			}
+			defer tmpFile.Close()
+			logrus.Infof("Dry-Run enabled, writing secrets to %s", tmpFile.Name())
+			if _, err := fmt.Fprintln(tmpFile, "---"); err != nil {
+				return err
+			}
+		}
+		if dryRun {
```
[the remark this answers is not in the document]
GeoBK: I am not sure I understand what you meant here

cmd/ci-secret-generator/main.go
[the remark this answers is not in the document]
bbguimaraes: My bad, I didn't know it was variadic.  Never mind, then.

cmd/ci-secret-generator/main.go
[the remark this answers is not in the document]
bbguimaraes: We're retesting the same condition as line 134, this can be a single block (i.e. these two lines can be removed, the `diff` is not the best at showing this =).

pkg/bitwarden/cli.go
GeoBK: I think the file name is important when creating the attachment.. I can add a /tmp/itemID/fileName to avoid concurrency issues.
But I guess itemID will have illegal characters for a folder(I'll pass in item name as an argument to handle that)
Yes,  I'll close the file as well.

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,232 @@
...
+	cmd := strings.Fields(command)
+	out, err := exec.Command(cmd[0], cmd[1:]...).CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("bw cmd failed: %v", string(out))
+	}
+	return out, nil
+}
+
+func updateSecrets(bwItems []bitWardenItem, bwClient bitwarden.Client, dryRun bool) error {
+	var tmpFile *os.File
```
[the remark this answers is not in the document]
GeoBK: I guess we could write to stderr, but the intention here has been to just write the output of the different commands to a file to check what the output would be in bitwarden, would it still be appropriate to output it to stderr?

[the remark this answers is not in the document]
bbguimaraes: We can always redirect output to get the same effect (`ci-secret-generator … > tmp.txt`, I guess `stdout` makes more sense in dry-run mode).  Having it create a temporary file and write the file name seems like an unnecessary step to me.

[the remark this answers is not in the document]
bbguimaraes: But more importantly, as I wrote in the review summary, I think we should have a safe way to verify the actions taken by the tool as much as possible.  As implemented, we can use the dry-run mode to verify that the correct commands are issued to the client in `pkg/bitwarden`, which is a good start, but things could still go wrong inside that package.

Unfortunately, it looks like the commands act immediately:

https://bitwarden.com/help/article/cli/#sync
> It’s important to note that the sync command is only a pull operation. It does not push data to the server since that is already done for you each time you make a change to your vault (create, edit, delete).

so any tracing would need to be done in `pkg/biwarden`.  Maybe a combination of the fake client and the [export command](https://bitwarden.com/help/article/cli/#export) would work?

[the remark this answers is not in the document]
GeoBK: I can change the output to stdout.. I have also added a few tests for the pkg/bitwarden. 
I am not using the export package right now though. I'll need some time to think through how I could incorporate the export command in there....

pkg/bitwarden/cli.go
GeoBK: Yup I changed this as well

GeoBK:
cc @alvaroaleman

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,226 @@
...
+func parseOptions() options {
+	var o options
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.BoolVar(&o.dryRun, "dry-run", false, "Whether to actually create the secrets with oc command")
+	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
+	fs.StringVar(&o.bwUser, "bw-user", "", "Username to access BitWarden.")
+	fs.StringVar(&o.bwPasswordPath, "bw-password-path", "", "Path to a password file to access BitWarden.")
+	fs.StringVar(&o.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
+	fs.IntVar(&o.maxConcurrency, "concurrency", 1, "Maximum number of concurrent in-flight goroutines to BitWarden.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
```
alvaroaleman, line 57: There is btw `flag.Commandline` for this

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+	}
+	return nil
+}
+
+func executeCommand(command string) ([]byte, error) {
+	cmd := strings.Fields(command)
+	out, err := exec.Command(cmd[0], cmd[1:]...).CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("bw cmd failed: %v", string(out))
```
alvaroaleman, line 128: Showing the command here might be nice

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,226 @@
...
+	out, err := executeCommand(cmd)
+	if err != nil {
+		return fmt.Errorf("failed to execute command: %w", err)
+	}
+	if dryRun {
+		if _, err := fmt.Fprintf(w, "\t%s: %s", entryType, string(out)); err != nil {
+			return err
+		}
+	} else {
+		return procesor(out)
```
alvaroaleman, line 143: I think this could be factored more nicely by not passing in an `io.Writer` and `dryRun` but instead just passing in a different `processor` for dry run mode

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	return nil
+}
+
+func updateSecrets(bwItems []bitWardenItem, bwClient bitwarden.Client, dryRun bool) error {
+	if dryRun {
+		logrus.Infof("Dry-Run enabled, writing secrets to stdout")
```
alvaroaleman, line 150: Yeah I know you copied this elsewhere, but we can never write secrets to stdout, they must be written to a file. If ppl _really_ want that because they run this locally or sth, they can `cat` the file afterwards. Printing it introduces the possibility that someone accidentally does that in a job run

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,226 @@
...
+					return fmt.Errorf("failed to set field item: %s, field: %s - %w", bwItem.ItemName, bwItem.Field.Name, err)
+				}
+				return nil
+			}); err != nil {
+				return err
+			}
+		}
+		if bwItem.Attachment.Name != "" {
+			if err := processItem(os.Stdout, bwItem.Attachment.Cmd, "Attachment", dryRun, func(out []byte) error {
+				if err := bwClient.SetAttachmentOnItem(bwItem.ItemName, bwItem.Attachment.Name, out); err != nil {
```
alvaroaleman, line 173: Rather than passing in a func that wraps the `bwClient`, why not directly pass in the `bwClient` as an interface and use a mock-implementation for dry run and tests?

pkg/bitwarden/cli.go
```diff
@@ -1,11 +1,14 @@
 package bitwarden
 
 import (
+	"bytes"
+	b64 "encoding/base64"
```
alvaroaleman, line 5: Nit: I think I would prefer it if you didn't alias this because to me that looks like it was referring to a non-stdlib base64 later on in the code

pkg/bitwarden/cli.go
```diff
@@ -142,3 +151,206 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+		return fmt.Errorf("failed to delete attachment, attachmentID: %s, itemID: %s: %w", attachmentID, itemID, err)
+	}
+	return nil
+}
+
+func (c *cliClient) SetFieldOnItem(itemName, fieldName string, fieldValue []byte) error {
+	var targetItem *Item
+	var targetField *Field
+	for index, item := range c.savedItems {
+		if itemName == item.Name {
```
alvaroaleman, line 239: Please avoid the deep nesting by expressing this and the other `if` below as

```
if itemName != item.Name { continue }
/// handle matching Item here
```

pkg/bitwarden/fake.go
```diff
@@ -50,6 +53,103 @@ func (c fakeClient) GetPassword(itemName string) ([]byte, error) {
...
+	return fmt.Sprintf("%x-%x-%x-%x-%x",
+		b[0:4], b[4:6], b[6:8], b[8:10], b[10:]), nil
+
+}
+
+func (c fakeClient) SetFieldOnItem(itemName, fieldName string, fieldValue []byte) error {
+	var targetItem *Item
+	var targetField *Field
+	for index, item := range c.items {
+		if itemName == item.Name {
```
alvaroaleman, line 71: Same comment as above, please avoid the nesting by doing
```
if itemName != item.Name { continue }
```

pkg/bitwarden/fake.go
```diff
@@ -50,6 +53,103 @@ func (c fakeClient) GetPassword(itemName string) ([]byte, error) {
...
+	}
+	targetField.Value = string(fieldValue)
+	return nil
+}
+
+func (c fakeClient) SetAttachmentOnItem(itemName, attachmentName string, fileContents []byte) error {
+	var targetItem *Item
+	var targetAttachment *Attachment
+	for index, item := range c.items {
+		if itemName == item.Name {
```
alvaroaleman, line 102: Avoid nesting :)

cmd/ci-secret-generator/main.go
GeoBK: I understand how I could probably get rid of writer, but to get rid of dryRun I would have to check that before calling the processItems function which I would have to do everytime I call processItems(three times). Wouldnt that be duplicated code?

alvaroaleman: As noted later on, the best approach for this is presumably to just use a fake `bitwardenClient` that prints rather than ever having the `dryRun` passed around or using the `processor` abstraction.

The usage of `dryRun` bool in production code outside of `construct different implementation of thing` was a big mistake we are trying to avoid in the future :)

cmd/ci-secret-generator/main.go
GeoBK: bwClient is already an interface, and I could probably change the Run function to make it print the commands to be executed to a file for dry run. But that would be the 64bit encoded data (the output of the command wouldn't be human readable then but would be exactly the command executed )
Or as you mentioned write a separate mock implementation to just spit out the data into a file.
Which would be better?

cmd/ci-secret-generator/main.go
GeoBK: I did not understand this...i didnt find any commandline in the flag package

GeoBK:
@alvaroaleman  I also have not tested the attachments functionality using the bitwarden cli, since I don't have access to that feature from my personal account, I'll have to test that once I can use an organization account

cmd/ci-secret-generator/main.go
alvaroaleman: A seperate mock implementation

pkg/bitwarden/cli.go
GeoBK: I have now made these tempDir - so I think we wouldn't have concurrency issues if we were to make it parallel

cmd/ci-secret-generator/main.go
alvaroaleman: Its a variable: https://godoc.org/flag#pkg-variables

Its basically just a shortcut to what you do in line 49

GeoBK: ok. understood

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,221 @@
...
+func (o *options) validateCompletedOptions() error {
+	if o.bwPassword == "" {
+		return fmt.Errorf("--bw-password-file was empty")
+	}
+
+	for i, bwItem := range o.config {
+		if bwItem.ItemName == "" {
+			return fmt.Errorf("config[%d].itemName: empty key is not allowed", i)
+		}
+		switch bwItem.Attribute.Name {
```
alvaroaleman, line 108: thats a lot easier expressed as `if bwItem.Attribute.Name != attributeTypePassword { error out }`

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,221 @@
...
+			(bwItem.Attachment.Name != "" && bwItem.Attachment.Cmd == "") {
+			return fmt.Errorf("config[%d]: empty field not allowed for cmd if name is specified for any of attribute, field or attachment", i)
+		}
+	}
+	return nil
+}
+
+func executeCommand(command string) ([]byte, error) {
+	cmd := strings.Fields(command)
+	out, err := exec.Command(cmd[0], cmd[1:]...).CombinedOutput()
```
alvaroaleman, line 124: I think this will panic if `cmd` has a `len` of 1

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,221 @@
...
+		return fmt.Errorf("failed to execute command: %w", err)
+	}
+	return procesor(out)
+}
+
+func updateSecrets(bwItems []bitWardenItem, bwClient bitwarden.Client) error {
+
+	for _, bwItem := range bwItems {
+		if bwItem.Field.Name != "" {
+			if err := processItem(bwItem.Field.Cmd, func(out []byte) error {
```
alvaroaleman, line 143: This would be a lot easier to read if `procesItem` just returns the `[]byte` output and we call the processing function afterwards rather than passing it in as a closure, that would allow us to avoid all this nesting

cmd/ci-secret-generator/main.go
GeoBK: its attributeTypePassword and the "" that I am checking for there. I can change it to 
`if bwItem.Attribute.Name != attributeTypePassword and bwItem.Attribute.Name!= ""` if that's more stylistically correct

cmd/ci-secret-generator/main.go
GeoBK: I dont think so - https://play.golang.org/p/XVkmgjMrGEj
lmk if i missed something/this was not what you meant
It would panic for length 0 though - but I think that would be caught in the command empty or not check

cmd/ci-secret-generator/main.go
GeoBK: In that case I guess I wont need the processItem function at all, I can just call executeCommand directly
I kept this after removing the dryrun logic just to keep avoid making that call each time. But I guess with that dry run logic removed DRY is no longer worth it

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,207 @@
...
+		}
+	}
+	return nil
+}
+
+func executeCommand(command string) ([]byte, error) {
+	cmd := strings.Fields(command)
+	out, err := exec.Command(cmd[0], cmd[1:]...).CombinedOutput()
+	if err != nil {
+		return nil, fmt.Errorf("bw cmd '%s' failed: %v", command, string(out))
```
alvaroaleman, line 124: This also needs to include the error, certain errors are not in the commands output, like for example if a binary is missing

pkg/bitwarden/cli.go
```diff
@@ -142,3 +151,246 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+		if err := c.createEmptyItem(itemName, newItem); err != nil {
+			return fmt.Errorf("failed to create new	 bw entry: %w", err)
+		}
+		c.savedItems = append(c.savedItems, *newItem)
+		targetItem = &c.savedItems[len(c.savedItems)-1]
+	}
+	attachmentChanged := true
+	if targetAttachment != nil {
+		// read the attachment file
+		tempDir, err := ioutil.TempDir("", "attachment")
```
alvaroaleman, line 303: Why create a tempdir per attachment rather than one per client lifecycle?

pkg/bitwarden/cli.go
```diff
@@ -142,3 +151,246 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+		if err != nil {
+			return fmt.Errorf("failed to create temporary file for getting: %w", err)
+		}
+		defer os.RemoveAll(tempDir)
```
alvaroaleman, line 307: Please err check and print the error, if any

pkg/bitwarden/cli.go
```diff
@@ -142,3 +151,246 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+		filePath := filepath.Join(tempDir, attachmentName)
+		existingFileContents, err := c.getAttachmentOnItemToFile(itemName, attachmentName, filePath)
+		if err != nil {
+			return fmt.Errorf("error reading attachment: %w", err)
+		}
+		if bytes.Equal(fileContents, existingFileContents) {
+			attachmentChanged = false
```
alvaroaleman, line 314: Just `return nil` here to save yourself the variable, the else tree and the `if attachmentChanged`

pkg/bitwarden/cli.go
```diff
@@ -142,3 +151,246 @@ func (c *cliClient) GetPassword(itemName string) ([]byte, error) {
...
+			targetItem = &c.savedItems[index]
+			break
+		}
+	}
+	if targetItem == nil {
+		newItem := &Item{}
+		if err := c.createItemWithPassword(itemName, password, newItem); err != nil {
+			return fmt.Errorf("failed to create new	 bw entry: %w", err)
+		}
+		c.savedItems = append(c.savedItems, *newItem)
```
alvaroaleman, line 348: `return nil` to save yourself the `else` tree

pkg/bitwarden/fake.go
```diff
@@ -50,6 +53,106 @@ func (c fakeClient) GetPassword(itemName string) ([]byte, error) {
 	return nil, fmt.Errorf("failed to find password in item %s", itemName)
 }
 
+func getNewUUID() (string, error) {
```
alvaroaleman, line 56: probably irrelevant here, but cryptographically secure entropy tends to be limited and isn't needed here so something like `time.Now().Nanoseconds()` should be enough

pkg/bitwarden/cli.go
GeoBK: Thats for if we end up making this concurrent, then we would need unique locations for each of the attachment files and avoid conflicts with similar attachment names

pkg/bitwarden/fake.go
GeoBK: If we were to use nanoseconds and have tests running in parallel, we "might" have both of them executing at the same time resulting in the same ID and flaky tests?
Also what do you mean by limited here?

cmd/ci-secret-generator/main.go
```diff
@@ -0,0 +1,207 @@
...
+	bwPasswordPath string
+	maxConcurrency int
+
+	config     []bitWardenItem
+	bwPassword string
+}
+
+type bitWardenItem struct {
+	ItemName   string         `json:"item_name"`
+	Field      fieldGenerator `json:"field,omitempty"`
```
alvaroaleman, line 33: The `omitempty` doesn't do anything on literal structs, only on pointers: https://stackoverflow.com/questions/18088294/how-to-not-marshal-an-empty-struct-into-json-with-go

pkg/bitwarden/cli.go
alvaroaleman: But right now its not concurrent so its a premature optimization. I would except one base tempdir with a subdir per item and a file per field/property/attachment in which case this wouldn't be needed

pkg/bitwarden/fake.go
alvaroaleman: Chances of collissions are so slim that this is part of the UUID1 standard, whose sole purpose is to create unique (but not cryptographically secure) uids: https://stackoverflow.com/questions/40411451/why-will-timeuuid-not-have-any-collisions.

See https://en.wikipedia.org/wiki/Entropy_(computing)#Practical_implications for an explanation of why it can be limited

GeoBK: So by limiting, you meant that there is a chance that the system would get stuck while waiting for random characters to appear

cmd/ci-secret-generator/main.go
GeoBK: Ok. I am removing the omitempty. Since I have checks for stringempty on the names fields before using them

pkg/bitwarden/fake.go
alvaroaleman: Yes, either for this application or some other, as the entropy pool is per host. There is no reason to use cryptographically secure entropy when the only thing you need is something that is collision-free.

pkg/bitwarden/cli.go
GeoBK: I would need either a temp directory or temp file for holding the attachment file to pass as a parameter to bw create. If the user wants to see the output of the commands the user is expected to run it in dry run mode which would not create a temp-directory. The indentation in the output file would give the user an idea of the itemname-field/attachment /attribute organization in bitwarden

PR #1061, opened 2020-08-03
Add option to enter source bug

GeoBK:
Add a dropdown listing the possible bugs from which the new clone
can be created to ease the process of creating clones by being
able to easily specify the dependency of the new clone.

### DIFF
--- a/pkg/backporter/backporter.go
+++ b/pkg/backporter/backporter.go
@@ -180,8 +180,15 @@ const clonesTemplateConstructor = `
 		{{ end }}
 		</tbody>
 	</table>
-	<form class="form-inline my-2 my-lg-0" role="search" action="/clones/create" method="post">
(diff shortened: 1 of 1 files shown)


GeoBK:
cc @alvaroaleman

alvaroaleman:
Wouldn't it make more sense to add a `clone` button to the table?

GeoBK:
Do you mean one clone button for every entry in the table?

alvaroaleman:
> Do you mean one clone button for every entry in the table?

Yes

GeoBK:
I dont have any strong opinions either way, an additional dropdown was what was requested in the feedback document, but I can do it the other way as well. lmk if you feel this is the way to go.

stevekuznetsov:
For any given target release, isn't there just one possible bug that can be the source of the clone? If so, we should just make the current clone drop-down do the right thing every time.

PR #1062, opened 2020-08-03
Simplify usage of registry and config agent

alvaroaleman:
/assign @bbguimaraes 

Rationale: The registry and config agent are useful in a variety of cases, not just for the configresolver. Most (all?) other components don't really care if they encounter any error in the registries, they have higher-level errors ("I want to do this with the registry and that failed").

Also, I find it generally questionable to put arbittrary errors in a metric label but that might be a distinct discussion.


alvaroaleman:
/retest
images presubmit failing
@bbguimaraes can you have a look, please?

bbguimaraes:
Sorry I wasn't able to get to #1059 in time.  These changes look OK, but I don't understand what the overall plan is.

Removing the coalescer was a good move, the code cannot work with it.  But we still have a race* and we removed the safety feature that eventually corrected it (periodic reloads).

***

\* Execute in a loop:

<details>

```diff
diff --git a/pkg/load/agents/registryAgent.go b/pkg/load/agents/registryAgent.go
index 9f6bb06c..92b78677 100644
--- a/pkg/load/agents/registryAgent.go
+++ b/pkg/load/agents/registryAgent.go
@@ -82,7 +82,9 @@ func (a *registryAgent) GetRegistryComponents() (registry.ReferenceByName, regis
 	return a.references, a.chains, a.workflows, a.documentation, a.metadata
 }
 
+var sleeping bool
 func (a *registryAgent) loadRegistry() error {
+	g := a.generation
 	logrus.Debug("Reloading registry")
 	startTime := time.Now()
 	references, chains, workflows, documentation, metadata, err := load.Registry(a.registryPath, a.flatRegistry)
@@ -90,6 +92,10 @@ func (a *registryAgent) loadRegistry() error {
 		a.recordError("failed to load ci-operator registry")
 		return fmt.Errorf("failed to load ci-operator registry (%w)", err)
 	}
+	if !sleeping && g == 3 {
+		sleeping = true
+		time.Sleep(2 * time.Second)
+	}
 	a.lock.Lock()
 	a.references = references
 	a.chains = chains
diff --git a/test/integration/ci-operator-configresolver.sh b/test/integration/ci-operator-configresolver.sh
index 4013e28b..c75407e0 100755
--- a/test/integration/ci-operator-configresolver.sh
+++ b/test/integration/ci-operator-configresolver.sh
@@ -38,7 +39,9 @@ os::integration::configresolver::check_log
 generation="$( os::integration::configresolver::generation::registry )"
 rm "${BASETMPDIR}/multistage-registry/registry/ipi/install/empty-parameter/ipi-install-empty-parameter-chain.yaml"
 mv "${BASETMPDIR}/multistage-registry/registry2/ipi/install/install/ipi-install-install-commands.sh"  "${BASETMPDIR}/multistage-registry/registry/ipi/install/install/ipi-install-install-commands.sh"
+os::integration::configresolver::wait_for_registry_update "$(($generation+2))"
 mv "${BASETMPDIR}/multistage-registry/registry2/ipi/install/install/ipi-install-install-ref.yaml"  "${BASETMPDIR}/multistage-registry/registry/ipi/install/install/ipi-install-install-ref.yaml"
+sleep 1
 mv "${BASETMPDIR}/multistage-registry/registry2/ipi/install/rbac/ipi-install-rbac-commands.sh"  "${BASETMPDIR}/multistage-registry/registry/ipi/install/rbac/ipi-install-rbac-commands.sh"
 rm "${BASETMPDIR}/multistage-registry/registry/ipi/install/with-parameter/ipi-install-with-parameter-chain.yaml"
 os::integration::configresolver::wait_for_registry_update "$(($generation+5))"
```

</details>

alvaroaleman:
@bbguimaraes can you elaborate on the race? My understanding is that we require a lock for the reloading and all events will trigger a reload. So if we get an event while we are reloading, we will just defer the second reload until the first one is complete since we wait for the lock.

alvaroaleman:
> These changes look OK, but I don't understand what the overall plan is.

Not requiring args that are not needed for callers. The thing was apparently build for the configresolver, with the config options the configresolver wanted in mind (load error rate, reload interval).

bbguimaraes:
That's correct, but loading the files from the file system happens outside the mutual exclusion section.  The race is essentially:

- F0 is updated
- T0 loads the files
- F1 is updated
- T1 loads the files
- --- race ---
- T1 locks and updates
- T0 locks and updates, but with stale data

The dumb way to solve it would be to move the calls to `pkg/load` into the section.  Our update times at pretty low, so that might be sufficient: https://grafana-prow-monitoring.apps.ci.l2s4.p1.openshiftapps.com/d/703f0ccf02cc4339a374b52eb10f653b/ci-operator-configresolver-dashboard?orgId=1&fullscreen&panelId=4&from=now-30d&to=now.  I'd have to test the recursive watcherr.

My initial idea was to set a `bool` flag whenever a new event came from `inotify` and have a single thread do serial loads while that flag is set.  We would most certainly load more than once per update, but I believe it would be dramatically less than one load per number of files updated.

bbguimaraes:
\* overall plan for the agents in general, that wasn't very clear in my initial message

alvaroaleman:
@bbguimaraes okay but that simply means we have to use the lock for the loading itself as well rather than just for updating the var?

alvaroaleman:
>  overall plan for the agents in general, that wasn't very clear in my initial message

Did you find the answer sufficient? If not, what is still unclear?

bbguimaraes:
> @bbguimaraes okay but that simply means we have to use the lock for the loading itself as well rather than just for updating the var?

Yeah, that's what I meant by my first suggestion.  I think we can do that and monitor to see if updates don't get too expensive.

In any case, I just wanted to make sure we did not loose track of the issues that are still present in the code.  This PR

/lgtm

PR #1063, opened 2020-08-03
[autopublicizeconfig] allow validation to change the options

droslean:
The tool was ignoring the whitelist since the map was always empty.

### DIFF
--- a/cmd/autopublicizeconfig/main.go
+++ b/cmd/autopublicizeconfig/main.go
@@ -71,7 +71,7 @@ func parseOptions() options {
 	return o
 }
 
-func validateOptions(o options) error {
+func defaultAndValidateOptions(o *options) error {
 	if o.githubLogin == "" {
 		return errors.New("--github-login is not specified")
 	}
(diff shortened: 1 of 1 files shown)


cmd/autopublicizeconfig/main.go
```diff
@@ -96,7 +96,7 @@ type Config struct {
 
 func main() {
 	o := parseOptions()
-	if err := validateOptions(o); err != nil {
+	if err := validateOptions(&o); err != nil {
```
alvaroaleman, line 99: I don't understand this change, the `Validation` doesn't seem to mutate anything?

Also if it does, you should rename it to `defaultAndValidate`

droslean: It mutates the `whitelistOptions`

droslean: +1 for the name but I should change this in the `whitelist` stuff. I will do this in a follow-up, when I will rename this to `allowlist` as well.

alvaroaleman: Why not change it now? Its not only the `whitelist` stuff its also `validateOptions` because its actually `defaultAndValidateOptions`.

PR #1069, opened 2020-08-05
fix link in successful publicize message

droslean:

### DIFF
--- a/cmd/publicize/server.go
+++ b/cmd/publicize/server.go
@@ -102,7 +102,7 @@ func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/publicize/server.go
```diff
@@ -102,7 +102,7 @@ func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 		return
 	}
 
-	destOrgRepoLink := fmt.Sprintf("https:///%s/tree/%s", destOrgRepo, baseBranch)
```
alvaroaleman, line 105: What about using tests to avoid this kind of issue?

droslean: I will add more tests.

PR #1071, opened 2020-08-05
Fix race in configagents

alvaroaleman:
We need to hold the lock during loading, otherwise:
1. Routine A loads
2. A new event comes in, routine B loads as well
3. Routine b finishes, acquires lock, sets config, releases lock
4. Routine a finishes, acquires lock, sets config, releases lock

We now have an outdated config.

/assign @bbguimaraes


bbguimaraes:
The irony of our PRs racing…  hahah

We'll miss the `Unlock`s in case of errors, my [PR](https://github.com/openshift/ci-tools/pull/1072) takes care of that.

PR #1072, opened 2020-08-05
pkg/load/agents: expand critical section

bbguimaraes:
See discussion in https://github.com/openshift/ci-tools/pull/1062.

### DIFF
--- a/pkg/load/agents/configAgent.go
+++ b/pkg/load/agents/configAgent.go
@@ -62,10 +62,9 @@ func NewFakeConfigAgent(configs load.ByOrgRepo) ConfigAgent {
 		errorMetrics: prometheus.NewCounterVec(prometheus.CounterOpts{}, []string{"error"}),
 	}
 	a.reloadConfig = func() error {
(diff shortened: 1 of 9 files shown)


pkg/load/agents/configAgent.go
```diff
@@ -62,10 +62,9 @@ func NewFakeConfigAgent(configs load.ByOrgRepo) ConfigAgent {
 		errorMetrics: prometheus.NewCounterVec(prometheus.CounterOpts{}, []string{"error"}),
 	}
 	a.reloadConfig = func() error {
-		indexes := a.buildIndexes(a.configs)
 		a.lock.Lock()
-		a.indexes = indexes
```
alvaroaleman, line 67: This confused me, why do we have both `loadFilenameToConfig` and `reloadConfig`?

bbguimaraes: It looks like it is used to implement the fake version of the agent (see `NewFakeConfigAgent` vs. `NewConfigAgent`).

PR #1074, opened 2020-08-05
Keep ssh bastion jobs on api.ci

hongkailiu:
After merge (this should not be a breaking change), I will contact authors of each bastion in https://github.com/openshift/release/tree/master/core-services/sshd-bastion/ to ask them add the label `"dptp.openshift.io/ssh-bastion": "true"` to the jobs which uses the bastions.

/cc @stevekuznetsov @alvaroaleman @bbguimaraes


stevekuznetsov:
Err - I meant jobs that use `core-services/sshd/` bastions. I don't think there's any reason that these disruptive jobs can't move from api.ci?

jaypoulz:
CC @r4f4 @hamzy @mkumatag @clnperez

PR #1078, opened 2020-08-05
ci-operator: ensure a one-to-one mapping from Dockerfile to image

stevekuznetsov:
We cannot allow two unique image names to be promoted after
being built from the same Dockerfile, as that would make it
impossible to take the build configuration in ocp-build-data
and determine the image build in CI that corresponds with it.
Technically, the validation we do here is more stringent --
in the most specific case it could be possible to build two
images from one Dockerfile if both were not promoted. However,
there is no real utility to that complex analysis and it's
easy and simple to enforce the stronger restriction that no
two images are built from the same Dockerfile at all.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -3,6 +3,7 @@ package api
 import (
 	"errors"
 	"fmt"
+	"path"
 	"path/filepath"
 	"regexp"
 	"strings"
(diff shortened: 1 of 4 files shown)


stevekuznetsov:
@alvaroaleman the breaking-changes failure was from a coding bug on my end. I don't expect any configs to actually do this today and if they don't we can land it without a gate

alvaroaleman:
@stevekuznetsov Seems you were wrong about that :upside_down_face:

PR #1084, opened 2020-08-06
ci-operator: add support for building operator index images

AlexNPavel:
This PR adds support for building operator index images by adding an
`operator_index` field for images. The process is handled in a similar
manner to the bundle images, which require 2 steps, with the first step
involving building a new image to be used as a source image.

First, an image build is made that runs the `opm` command that generates
the `index.db` file and a dockerfile to build the corresponding index
image.

Second, a build is made with the previous build's image as its source,
using the dockerfile generated by `opm`.

This PR also adds validation for the `images` section of the config as
both `operator_manifests` and `operator_index` add potential for user
error.

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -75,6 +75,10 @@ func (config *ReleaseBuildConfiguration) validate(org, repo string, resolved boo
 	// parts of the configuration, so it's written as a standalone method
 	validationErrors = append(validationErrors, config.validateTestStepDependencies()...)
 
+	if config.Images != nil {
+		validationErrors = append(validationErrors, validateImages("images", config.Images)...)
+	}
+
 	if config.InputConfiguration.BaseImages != nil {
 		validationErrors = append(validationErrors, validateImageStreamTagReferenceMap("base_images", config.InputConfiguration.BaseImages)...)
 	}
@@ -265,6 +269,31 @@ func validateBuildRootImageConfiguration(fieldRoot string, input *BuildRootImage
 		validationErrors = append(validationErrors, fmt.Errorf("%s: both image_stream_tag and project_image cannot be set", fieldRoot))
 	} else if input.ProjectImageBuild == nil && input.ImageStreamTagReference == nil {
 		validationErrors = append(validationErrors, fmt.Errorf("%s: you have to specify either image_stream_tag or project_image", fieldRoot))
+	}
+	return validationErrors
+}
+
+func validateImages(fieldRoot string, input []ProjectDirectoryImageBuildStepConfiguration) []error {
+	var validationErrors []error
+	for num, image := range input {
+		fieldRootN := fmt.Sprintf("%s[%d]", fieldRoot, num)
+		if image.To == "" {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `to` must be set", fieldRootN))
+		}
+		if len(image.Substitute) > 0 {
+			if image.OperatorManifests == "" {
+				validationErrors = append(validationErrors, fmt.Errorf("%s.substitute: %s.operator_manifests must also be set", fieldRootN, fieldRootN))
+			}
+			for num, sub := range image.Substitute {
+				fieldRootN2 := fmt.Sprintf("%s.substitute[%d]", fieldRootN, num)
+				if sub.PullSpec == "" {
+					validationErrors = append(validationErrors, fmt.Errorf("%s.pullspec: must be set", fieldRootN2))
+				}
+				if sub.With == "" {
+					validationErrors = append(validationErrors, fmt.Errorf("%s.with: must be set", fieldRootN2))
+				}
+			}
+		}
 	}
 	return validationErrors
 }
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -79,6 +79,10 @@ func (config *ReleaseBuildConfiguration) validate(org, repo string, resolved boo
 		validationErrors = append(validationErrors, validateImages("images", config.Images)...)
 	}
 
+	if config.Operator != nil {
+		validationErrors = append(validationErrors, validateOperator("operator", config.Operator)...)
+	}
+
 	if config.InputConfiguration.BaseImages != nil {
 		validationErrors = append(validationErrors, validateImageStreamTagReferenceMap("base_images", config.InputConfiguration.BaseImages)...)
 	}
@@ -280,18 +284,35 @@ func validateImages(fieldRoot string, input []ProjectDirectoryImageBuildStepConf
 		if image.To == "" {
 			validationErrors = append(validationErrors, fmt.Errorf("%s: `to` must be set", fieldRootN))
 		}
-		if len(image.Substitute) > 0 {
-			if image.OperatorManifests == "" {
-				validationErrors = append(validationErrors, fmt.Errorf("%s.substitute: %s.operator_manifests must also be set", fieldRootN, fieldRootN))
+		if image.To == BundleSourceName {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `to` cannot be %s", fieldRootN, BundleSourceName))
+		}
+		if strings.HasPrefix(string(image.To), BundlePrefix) {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `to` cannot begin with `%s`", fieldRootN, BundlePrefix))
+		}
+		if image.To == IndexImageGeneratorName {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `to` cannot be %s", fieldRootN, IndexImageGeneratorName))
+		}
+		if image.To == IndexImageName {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `to` cannot be %s", fieldRootN, IndexImageName))
+		}
+	}
+	return validationErrors
+}
+
+func validateOperator(fieldRoot string, input *OperatorStepConfiguration) []error {
+	var validationErrors []error
+	if len(input.Substitutions) > 0 {
+		if len(input.Manifests) == 0 {
+			validationErrors = append(validationErrors, fmt.Errorf("%s.substitute: %s.operator_manifests must also be set", fieldRoot, fieldRoot))
+		}
+		for num, sub := range input.Substitutions {
+			fieldRootN := fmt.Sprintf("%s.substitute[%d]", fieldRoot, num)
+			if sub.PullSpec == "" {
+				validationErrors = append(validationErrors, fmt.Errorf("%s.pullspec: must be set", fieldRootN))
 			}
-			for num, sub := range image.Substitute {
-				fieldRootN2 := fmt.Sprintf("%s.substitute[%d]", fieldRootN, num)
-				if sub.PullSpec == "" {
-					validationErrors = append(validationErrors, fmt.Errorf("%s.pullspec: must be set", fieldRootN2))
-				}
-				if sub.With == "" {
-					validationErrors = append(validationErrors, fmt.Errorf("%s.with: must be set", fieldRootN2))
-				}
+			if sub.With == "" {
+				validationErrors = append(validationErrors, fmt.Errorf("%s.with: must be set", fieldRootN))
 			}
 		}
 	}
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -302,18 +302,13 @@ func validateImages(fieldRoot string, input []ProjectDirectoryImageBuildStepConf
 
 func validateOperator(fieldRoot string, input *OperatorStepConfiguration) []error {
 	var validationErrors []error
-	if len(input.Substitutions) > 0 {
-		if len(input.Manifests) == 0 {
-			validationErrors = append(validationErrors, fmt.Errorf("%s.substitute: %s.operator_manifests must also be set", fieldRoot, fieldRoot))
+	for num, sub := range input.Substitutions {
+		fieldRootN := fmt.Sprintf("%s.substitute[%d]", fieldRoot, num)
+		if sub.PullSpec == "" {
+			validationErrors = append(validationErrors, fmt.Errorf("%s.pullspec: must be set", fieldRootN))
 		}
-		for num, sub := range input.Substitutions {
-			fieldRootN := fmt.Sprintf("%s.substitute[%d]", fieldRoot, num)
-			if sub.PullSpec == "" {
-				validationErrors = append(validationErrors, fmt.Errorf("%s.pullspec: must be set", fieldRootN))
-			}
-			if sub.With == "" {
-				validationErrors = append(validationErrors, fmt.Errorf("%s.with: must be set", fieldRootN))
-			}
+		if sub.With == "" {
+			validationErrors = append(validationErrors, fmt.Errorf("%s.with: must be set", fieldRootN))
 		}
 	}
 	return validationErrors
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -394,6 +394,7 @@ type StepConfiguration struct {
 	PipelineImageCacheStepConfiguration         *PipelineImageCacheStepConfiguration         `json:"pipeline_image_cache_step,omitempty"`
 	SourceStepConfiguration                     *SourceStepConfiguration                     `json:"source_step,omitempty"`
 	BundleSourceStepConfiguration               *BundleSourceStepConfiguration               `json:"bundle_source_step,omitempty"`
+	IndexGeneratorStepConfiguration             *IndexGeneratorStepConfiguration             `json:"index_generator_step,omitempty"`
 	ProjectDirectoryImageBuildStepConfiguration *ProjectDirectoryImageBuildStepConfiguration `json:"project_directory_image_build_step,omitempty"`
 	RPMImageInjectionStepConfiguration          *RPMImageInjectionStepConfiguration          `json:"rpm_image_injection_step,omitempty"`
 	RPMServeStepConfiguration                   *RPMServeStepConfiguration                   `json:"rpm_serve_step,omitempty"`
@@ -1002,6 +1003,17 @@ type SourceStepConfiguration struct {
 	// ClonerefsPath is the path in the above image where the
 	// clonerefs tool is placed
 	ClonerefsPath string `json:"clonerefs_path"`
+}
+
+// IndexGeneratorStepConfiguration describes a step that creates an index database and
+// Dockerfile to build an operator index that uses the generated database based on
+// bundle names provided in OperatorIndex
+type IndexGeneratorStepConfiguration struct {
+	To PipelineImageStreamTagReference `json:"to,omitempty"`
+
+	// OperatorIndex is a list of the names of the bundle images that the
+	// index will contain in its database.
+	OperatorIndex []string `json:"operator_index,omitempty"`
 }
 
 // BundleSourceStepConfiguration describes a step that performs a set of
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1012,15 +1012,17 @@ type SourceStepConfiguration struct {
 // bundle build dockerfiles, and images the operator(s) depends on that must
 // be substituted to run in a CI test cluster
 type OperatorStepConfiguration struct {
-	// DockerfilePath describes the path(s) to the dockerfile(s) that builds the operator bundle
-	DockerfilePath []string `json:"dockerfile_path,omitempty"`
-
-	// Manifests describes the path(s) of the operator manifests that the bundle(s) requires
-	Manifests []string `json:"manifests,omitempty"`
+	// Bundles define a dockerfile and build context to build a bundle
+	Bundles []Bundle `json:"bundles,omitempty"`
 
 	// Substitutions describes the pullspecs in the operator manifests that must be subsituted
 	// with the pull specs of the images in the CI registry
 	Substitutions []PullSpecSubstitution `json:"substitutions,omitempty"`
+}
+
+type Bundle struct {
+	DockerfilePath string `json:"dockerfile_path,omitempty"`
+	ContextDir     string `json:"context_dir,omitempty"`
 }
 
 // IndexGeneratorStepConfiguration describes a step that creates an index database and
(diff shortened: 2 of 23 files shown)


AlexNPavel:
The code is ready for review, but I'm marking it as `WIP` because I want to make sure it actually works when run against the cluster.

pkg/api/types.go
```diff
@@ -983,6 +984,17 @@ type SourceStepConfiguration struct {
...
 
+// IndexGeneratorStepConfiguration describes a step that creates an index database and
+// Dockerfile to build an operator index that uses the generated database based on
+// bundle names provided in OperatorIndex
+type IndexGeneratorStepConfiguration struct {
+	To PipelineImageStreamTagReference `json:"to,omitempty"`
+
+	// OperatorIndex is a list of the names of the bundle images that the
+	// index will contain in its database.
+	OperatorIndex []string `json:"operator_index,omitempty"`
```
stevekuznetsov, line 995: Why are we asking the user to provide this? Why would someone ever not want all of the bundles they build to be in their index? Why would we expect them to ask us to build a bundle but not an index?

pkg/steps/bundle_source.go
```diff
@@ -89,8 +89,8 @@ func (s *bundleSourceStep) bundleSourceDockerfile() (string, error) {
 	return strings.Join(dockerCommands, "\n"), nil
 }
 
-func (s *bundleSourceStep) getFullPullSpec(tag string) (string, error) {
-	is, err := s.imageClient.ImageStreams(s.jobSpec.Namespace()).Get(context.TODO(), api.StableImageStream, meta.GetOptions{})
+func getFullPullSpec(tag, namespace string, ic imageclientset.ImageStreamsGetter) (string, error) {
```
stevekuznetsov, line 92: Yuck. We have like 30 of these littered in the code. Can we please use `utils.ImageDigestFor`

pkg/steps/index_generator.go
```diff
@@ -0,0 +1,128 @@
...
+	dockerCommands = append(dockerCommands, fmt.Sprintf("WORKDIR %s", IndexOutputDir))
+	var bundles []string
+	for _, bundleName := range s.config.OperatorIndex {
+		fullSpec, err := getFullPullSpec(bundleName, s.jobSpec.Namespace(), s.imageClient)
+		if err != nil {
+			return "", fmt.Errorf("failed to get replacement imagestream for image tag `%s`", bundleName)
+		}
+		bundles = append(bundles, fullSpec)
+	}
+	dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["opm", "index", "add", "--bundles", "%s", "-d", "%s", "--generate"]`, strings.Join(bundles, ","), IndexDockerfileName))
```
stevekuznetsov, line 85: nit: is it possible to use a long-form flag, not `-d`? It will be clearer to read and understand for people not intimate with `opm`

pkg/steps/index_generator.go
```diff
@@ -0,0 +1,128 @@
...
+			},
+		},
+		"",
+		s.resources,
+		s.pullSecret,
+	)
+	return handleBuild(ctx, s.buildClient, build, s.artifactDir)
+}
+
+func (s *indexGeneratorStep) indexGenDockerfile() (string, error) {
```
stevekuznetsov, line 72: IIRC if this is a Build, we need to inject the pull credentials for the bundle into it. If w run it as a Pod and get the output Dockerfile as an artifact, we can mount the credentials as a secret.

AlexNPavel: I found another way to do it that I believe is reasonable. Secrets can be mounted as build sources in openshift builds. We can use a multistage docker build to where the builder stage copies the pull secret to the correct location and runs the `opm` command and then use a scratch image as the output stage that just copies the database directory and dockerfile.

This should work with our registry and give us the data + dockerfile we need to build the actual index image.

AlexNPavel:
Current status of work: the `index-gen` image is being built correctly with the `Dockerfile` and `database` dir in the correct location in the image. Currently trying to fix an issue with getting the openshift build to correctly find the dockerfile for the actual index build.

Other issues that need to be worked on:
1. The `e2e` tests don't work in the `ci-tools` CI. The existing credentials for the e2e test don't provide access to the local registries, which is necessary for the test. For local testing, I am using the `regcred` credentials that all normal `ci-operator` runs get in CI. We may need to use those credentials for the e2e tests here as well.
2. This currently depends on a @petr-muller's metering-operator for for manifests. I should be able to copy the manifests directory from there into the `test` directory in this repo and point the e2e test to the correct ref using the CI env vars so we don't depend on an external repo.

stevekuznetsov:
@AlexNPavel sounds good -- do we still explicitly ask the user to configure the index build or implicitly create it when bundles are configured? I agree that updating the secret available to the e2e and moving static definitions (as minimal as possible) into this repo are good next steps

AlexNPavel:
E2E tests for the bundle now use the manifests from the PR they are being run against.

AlexNPavel:
This PR has been updated, rebased, and cleaned up. It should be ready once we update the pull credentials for the `e2e` tests.

AlexNPavel:
> @AlexNPavel sounds good -- do we still explicitly ask the user to configure the index build or implicitly create it when bundles are configured?

We still require it to be explicitly specified. I'm looking into how to handle it implicitly.

petr-muller:
/assign

Will look! For the record, my intention was to first make the specifications explicit and after the functionality is validated I wanted to followup with a "infer indexes from bundle presence" story to have smaller iterations.

pkg/api/config.go
```diff
@@ -125,6 +129,40 @@ func validateBuildRootImageConfiguration(fieldRoot string, input *BuildRootImage
...
+				fieldRootN2 := fmt.Sprintf("%s.substitute[%d]", fieldRootN, num)
+				if sub.PullSpec == "" {
+					validationErrors = append(validationErrors, fmt.Errorf("%s.pullspec: must be set", fieldRootN2))
+				}
+				if sub.With == "" {
+					validationErrors = append(validationErrors, fmt.Errorf("%s.with: must be set", fieldRootN2))
+				}
+			}
+		}
+		if len(image.OperatorIndex) > 0 && image.OperatorManifests != "" {
```
stevekuznetsov, line 153: IDK this looks like more code and more work. Inferring is simple. Let's do it. I don't want to onboard teams to this.

pkg/steps/index_generator.go
```diff
@@ -0,0 +1,143 @@
...
+	dockerCommands = append(dockerCommands, "RUN mkdir $HOME/.docker && mv .dockerconfigjson $HOME/.docker/config.json")
+	var bundles []string
+	is, err := s.imageClient.ImageStreams(s.jobSpec.Namespace()).Get(context.TODO(), api.StableImageStream, meta.GetOptions{})
+	if err != nil {
+		return "", fmt.Errorf("failed to get stable imagestream: %w", err)
+	}
+	for _, bundleName := range s.config.OperatorIndex {
+		fullSpec, exists := util.ResolvePullSpec(is, bundleName, false)
+		if !exists {
+			return "", fmt.Errorf("failed to get fulle pull spec for bundle `%s`", bundleName)
```
stevekuznetsov, line 92: `s/fulle/full/`

pkg/steps/index_generator.go
```diff
@@ -0,0 +1,143 @@
...
+
+func (s *indexGeneratorStep) Inputs() (api.InputDefinition, error) {
+	return nil, nil
+}
+
+func (s *indexGeneratorStep) Run(ctx context.Context) error {
+	return results.ForReason("building_index_generator").ForError(s.run(ctx))
+}
+
+func (s *indexGeneratorStep) run(ctx context.Context) error {
```
stevekuznetsov, line 39: we might want to identify specific `results.ForReason()` with the errors coming out of this so it's not opaque -- which of these errors are infra-caused? All of them? Is it possible for any to be user-based?

test/e2e/simple.sh
```diff
@@ -47,5 +47,18 @@ RELEASE_IMAGE_LATEST="$( curl -s -H "Accept: application/json"  "https://api.ope
 export RELEASE_IMAGE_LATEST
 os::cmd::expect_success "ci-operator --secret-dir ${PULL_SECRET_DIR} --target [release:latest] --config ${suite_dir}/dynamic-releases.yaml"
 unset RELEASE_IMAGE_LATEST
+os::test::junit::declare_suite_end
 
+os::test::junit::declare_suite_start "e2e/simple/optional-operator"
+if [[ -z "${PULL_BASE_SHA:-}" ]]; then
```
stevekuznetsov, line 53: Confused about these, and why `$JOB_SPEC` doesn't cover it?

test/e2e/simple.sh
```diff
@@ -47,5 +47,18 @@ RELEASE_IMAGE_LATEST="$( curl -s -H "Accept: application/json"  "https://api.ope
...
+  os::log::fatal "\$PULL_BASE_SHA must be set for this test"
+fi
+if [[ -z "${PULL_NUMBER:-}" ]]; then
+  os::log::fatal "\$PULL_NUMBER must be set for this test"
+fi
+if [[ -z "${PULL_PULL_SHA:-}" ]]; then
+  os::log::fatal "\$PULL_PULL_SHA must be set for this test"
+fi
+export JOB_SPEC='{"type":"postsubmit","job":"branch-ci-openshift-ci-tools-master-ci-operator-e2e","buildid":"0","prowjobid":"uuid","refs":{"org":"openshift","repo":"ci-tools","base_ref":"master","base_sha":"'"${PULL_BASE_SHA}"'","pulls":[{"number":'"${PULL_NUMBER}"',"sha":"'"${PULL_PULL_SHA}"'","author":"AlexNPavel"}]}}'
+os::cmd::expect_success "ci-operator --image-import-pull-secret ${PULL_SECRET_DIR}/.dockerconfigjson --target [images] --config ${suite_dir}/optional-operators.yaml"
```
stevekuznetsov, line 63: Could we somehow write a simple, lightweight `tests` step in this config that either runs in the index or uses my PR to depend on the index image and does something trivial to validate that the index is workable?

pkg/api/config.go
petr-muller: Yeah, I don't insist on allowing explicit. I never wanted people to actually use the explicit config.

AlexNPavel: Added implicit generation of index image steps.

pkg/steps/index_generator.go
AlexNPavel: There is 1 situation where it could be user-based, which would be the bundle not containing certain info (like missing a CRD) or containing incorrect info. I'll add a regex to try and identify if that is the issue based on logs.

test/e2e/simple.sh
AlexNPavel: It should. Changed to use that instead.

pkg/steps/index_generator.go
AlexNPavel: Added check

pkg/api/types.go
```diff
@@ -1004,26 +1008,58 @@ type SourceStepConfiguration struct {
 	ClonerefsPath string `json:"clonerefs_path"`
 }
 
+// OperatorStepConfiguration describes the locations of operator bundle information,
+// bundle build dockerfiles, and images the operator(s) depends on that must
+// be substituted to run in a CI test cluster
+type OperatorStepConfiguration struct {
```
stevekuznetsov, line 1014: I think I'd prefer a type

```go
type OperatorStepConfiguration struct {
	Bundles []Bundle `json:"bundles,omitempty"`
	Substitutions []PullSpecSubstitution `json:"substitutions,omitempty"`
}

type Bundle struct {
	DockerfilePath string `json:"dockerfile_path,omitempty"`
	Manifests string `json:"manifests,omitempty"`
}
```

Since we have a 1:1 relationship between the dockerifle path and manifests dir

pkg/api/types.go
```diff
@@ -1004,26 +1008,58 @@ type SourceStepConfiguration struct {
...
+	Substitutions []PullSpecSubstitution `json:"substitutions,omitempty"`
+}
+
+// IndexGeneratorStepConfiguration describes a step that creates an index database and
+// Dockerfile to build an operator index that uses the generated database based on
+// bundle names provided in OperatorIndex
+type IndexGeneratorStepConfiguration struct {
+	To PipelineImageStreamTagReference `json:"to,omitempty"`
+
+	// OperatorIndex is a list of the names of the bundle images that the
```
stevekuznetsov, line 1032: Defaults to all? None?

pkg/api/types.go
```diff
@@ -1004,26 +1008,58 @@ type SourceStepConfiguration struct {
...
-
-	// ContextDir is the directory in the project
-	// from which this build should be run.
-	ContextDir string `json:"context_dir,omitempty"`
-
-	// OperatorManifests is the subdir of context_dir where optional
+	// Manifests is the subdir of context_dir where optional
 	// operator manifests are stored for operator bundle images.
-	OperatorManifests string `json:"operator_manifests,omitempty"`
+	Manifests []string `json:"operator_manifests,omitempty"`
```
stevekuznetsov, line 1050: Why is this a slice?

stevekuznetsov: Also don't we still need `to` and `context_dir` on this config? What if someone adds one via a `raw_step`?

pkg/defaults/defaults.go
```diff
@@ -451,35 +454,56 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
-		if image.OperatorManifests != "" {
-			buildSteps = append(buildSteps, api.StepConfiguration{BundleSourceStepConfiguration: &api.BundleSourceStepConfiguration{
-				To:                steps.BundleSourceName(image.To),
-				ContextDir:        image.ContextDir,
-				OperatorManifests: image.OperatorManifests,
-				Substitute:        image.Substitute,
-			}})
-		}
 		buildSteps = append(buildSteps, api.StepConfiguration{ProjectDirectoryImageBuildStepConfiguration: image})
+		var outputImageName string
```
stevekuznetsov, line 458: This is an image stream name, not an image name?

pkg/defaults/defaults.go
```diff
@@ -451,35 +454,56 @@ func stepConfigsForBuild(config *api.ReleaseBuildConfiguration, jobSpec *api.Job
...
+			To: api.ImageStreamTagReference{
+				Name: outputImageName,
+				Tag:  string(image.To),
+			},
+			Optional: image.Optional,
+		}})
+	}
+
+	if config.Operator != nil {
+		// Build a bundle source image that substitutes all values in `substitutions` in all `manifests` directories
```
stevekuznetsov, line 475: I don't understand this - why is it different than it used to be?

pkg/steps/bundle_source.go
```diff
@@ -69,58 +69,49 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
 	return handleBuild(ctx, s.buildClient, build, s.artifactDir)
 }
 
-func replaceCommand(manifestDir, pullSpec, with string) string {
-	return fmt.Sprintf("find %s -type f -exec sed -i 's?%s?%s?g' {} +", manifestDir, pullSpec, with)
+func replaceCommand(pullSpec, with string) string {
+	return fmt.Sprintf("find . -type f -name \\*.yaml -exec sed -i 's?%s?%s?g' {} +", pullSpec, with)
```
stevekuznetsov, line 73: Must these be YAML files?

pkg/api/types.go
```diff
@@ -1004,26 +1008,56 @@ type SourceStepConfiguration struct {
...
+type OperatorStepConfiguration struct {
+	// Bundles define a dockerfile and build context to build a bundle
+	Bundles []Bundle `json:"bundles,omitempty"`
+
+	// Substitutions describes the pullspecs in the operator manifests that must be subsituted
+	// with the pull specs of the images in the CI registry
+	Substitutions []PullSpecSubstitution `json:"substitutions,omitempty"`
+}
+
+type Bundle struct {
```
stevekuznetsov, line 1023: Needs GoDoc

pkg/steps/bundle_source.go
```diff
@@ -69,58 +69,49 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
-	}
-	if len(is.Status.DockerImageRepository) > 0 {
-		return is.Status.DockerImageRepository + ":" + tag, nil
-	}
-	return "", fmt.Errorf("no pull spec available for image stream %s", api.StableImageStream)
-}
-
 func (s *bundleSourceStep) Requires() []api.StepLink {
 	return []api.StepLink{
 		api.InternalImageLink(api.PipelineImageStreamTagReferenceSource),
```
stevekuznetsov, line 99: This is problematic - we need to parse the substitutions to ensure we correctly expose images that we require to be present. If a user wants to substitute with a pull-spec from `pipeline` it must be built before this runs.

pkg/steps/bundle_source.go
```diff
@@ -69,58 +69,49 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
+	if err != nil {
+		return "", fmt.Errorf("failed to get stable imagestream: %w", err)
+	}
+	for _, sub := range s.config.Substitutions {
+		replaceSpec, exists := util.ResolvePullSpec(is, sub.With, false)
+		if !exists {
 			return "", fmt.Errorf("failed to get replacement imagestream for image tag `%s`", sub.With)
 		}
-		dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["bash", "-c", "%s"]`, replaceCommand(manifestDir, sub.PullSpec, replaceSpec)))
+		// The \ character has to be escaped in the dockerfile to run correctly
```
stevekuznetsov, line 89: Why doesn't `replaceCommand()` handle this?

pkg/steps/bundle_source.go
```diff
@@ -69,58 +69,49 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
+		replace := strings.ReplaceAll(replaceCommand(sub.PullSpec, replaceSpec), `\`, `\\`)
+		dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["bash", "-c", "%s"]`, replace))
```
stevekuznetsov, line 91: Why do we need bash for this? Can't we just run `find` and `sed`?

pkg/steps/bundle_source_test.go
```diff
@@ -49,11 +50,21 @@ func TestReplaceCommand(t *testing.T) {
 	if err := exec.Command("cp", "-a", "testdata/4.6", temp).Run(); err != nil {
 		t.Fatalf("Failed to copy testdata to tempdir: %v", err)
 	}
+	originalDir, err := os.Getwd()
```
stevekuznetsov, line 53: Prefer to pass the working dir to `exec.Command()` instead of changing global state

pkg/steps/index_generator.go
```diff
@@ -0,0 +1,142 @@
...
+			return "", fmt.Errorf("failed to get full pull spec for bundle `%s`", bundleName)
+		}
+		bundles = append(bundles, fullSpec)
+	}
+	dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["opm", "index", "add", "--bundles", "%s", "--out-dockerfile", "%s", "--generate"]`, strings.Join(bundles, ","), IndexDockerfileName))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("FROM %s:%s", api.PipelineImageStream, api.PipelineImageStreamTagReferenceSource))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("WORKDIR %s", IndexDataDirectory))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("COPY --from=builder %s %s", IndexDockerfileName, IndexDockerfileName))
+	dockerCommands = append(dockerCommands, ("COPY --from=builder /database/ database"))
+	dockerCommands = append(dockerCommands, "")
```
stevekuznetsov, line 105: ?

pkg/steps/index_generator.go
```diff
@@ -0,0 +1,142 @@
...
+		if !exists {
+			return "", fmt.Errorf("failed to get full pull spec for bundle `%s`", bundleName)
+		}
+		bundles = append(bundles, fullSpec)
+	}
+	dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["opm", "index", "add", "--bundles", "%s", "--out-dockerfile", "%s", "--generate"]`, strings.Join(bundles, ","), IndexDockerfileName))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("FROM %s:%s", api.PipelineImageStream, api.PipelineImageStreamTagReferenceSource))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("WORKDIR %s", IndexDataDirectory))
+	dockerCommands = append(dockerCommands, fmt.Sprintf("COPY --from=builder %s %s", IndexDockerfileName, IndexDockerfileName))
+	dockerCommands = append(dockerCommands, ("COPY --from=builder /database/ database"))
```
stevekuznetsov, line 104: Why is the string literal in parentheses here?

pkg/steps/index_generator_test.go
```diff
@@ -0,0 +1,77 @@
...
+			},
+			Status: apiimagev1.ImageStreamStatus{
+				PublicDockerImageRepository: "some-reg/target-namespace/stable",
+			},
+		}),
+		t: t,
+	}
+
+	var expectedDockerfileSingleBundle = `
+FROM quay.io/operator-framework/upstream-opm-builder AS builder
```
stevekuznetsov, line 27: Technically would be good to import these as `base_images` to the namespace so that they're considered when detecting the version of the job and so that we're not reliant on external registries/doing cluster-local pulls

pkg/steps/project_image.go
```diff
@@ -76,18 +72,44 @@ func (s *projectDirectoryImageBuildStep) run(ctx context.Context) error {
 
 	images := buildInputsFromStep(s.config.Inputs)
 	// If image being built is an operator bundle, use the bundle source instead of original source
-	if s.config.OperatorManifests != "" {
+	if strings.HasPrefix(string(s.config.To), api.BundlePrefix) {
```
stevekuznetsov, line 75: Can we write a function for this logic - to make it clearly labeled and obvious. Raw string manipulations scattered around made a lot of our `ci-operator` code rot over time in e.g. env vars

stevekuznetsov: And presumably a function to create the name from an index and prefix, and a test that `detect(wrap()) == true` and `unwrap(wrap(input)) == input` if there is an `unwrap`

test/e2e/simple.sh
stevekuznetsov: Did you?

pkg/steps/bundle_source.go
petr-muller: AFAIK yes, they are always manifest files. Technically they could be called differently, I guess. `.yml` or something like that.

pkg/steps/bundle_source.go
petr-muller: I don't think this is that problematic - the images corresponding to the pullspecs only need to be available once you actually use the bundle - after you deploy the CatalogSource and create a `Subscription`, etc, but you don't need it at the bundle build time. So as long as we validate that the image names in `with` stanzas `will` resolve to somehting, we should be fine.

PR #1085, opened 2020-08-07
Auto detect source bug for create clones

GeoBK:
The create clones functionality will automatically chose
the immediately higher target release as the source bug while
creating a clone.

### DIFF
--- a/pkg/backporter/backporter.go
+++ b/pkg/backporter/backporter.go
@@ -132,10 +132,15 @@ const htmlPageEnd = `
 `
 
 const clonesTemplateConstructor = `
-{{if .NewCloneID }}
+{{if .NewCloneIDs }}
 	<div class="alert alert-success alert-dismissible" id="success-banner">
 	<a href="#" class="close" data-dismiss="alert" aria-label="close">&times;</a>
-	<strong>Success!</strong> Clone created - <a href="/clones?ID={{ .NewCloneID }}" >Bug#{{ .NewCloneID }}</a>.
+	<strong>Success!</strong> Clone created - 
+	{{range $index, $bug := .NewCloneIDs }}
+		{{ if $index}}, {{end}}
+		<a href="/clones?ID={{ $bug }}" >Bug#{{ $bug }}</a>
+	{{end}}
+	.
 	</div>
 {{ end }}
 {{if .MissingReleases }}
@@ -169,7 +174,15 @@ const clonesTemplateConstructor = `
 		<tbody>
 		{{ if .Clones }}
 			{{ range $clone := .Clones }}
-				<tr {{if eq $clone.ID $.Bug.ID }}class="table-active"{{ end }}>
+				<tr class=" {{if gt (len $clone.TargetRelease) 0 }} 
+								{{if eq (index $clone.TargetRelease 0) "---"}}
+									table-danger 
+								{{else}}
+									{{if eq $clone.ID $.Bug.ID }}table-active{{ end }}
+								{{end}}
+							{{else}}
+								{{if eq $clone.ID $.Bug.ID }}table-active{{ end }}
+							{{ end }}">
 					<td style="vertical-align: middle;">
 					{{ if $clone.TargetRelease }} 
 						{{ index $clone.TargetRelease 0 }}
@@ -281,7 +294,7 @@ type ClonesTemplateData struct {
 	Parent          *bugzilla.Bug          // Root bug if it is a a bug, otherwise holds itself
 	PRs             []bugzilla.ExternalBug // Details of linked PR
 	CloneTargets    []string
-	NewCloneID      int
+	NewCloneIDs     []string
 	MissingReleases []string
 }
 
@@ -299,6 +312,19 @@ func writePage(w http.ResponseWriter, title string, body *template.Template, dat
 		return err
 	}
 	return nil
+}
+
+func sortByTargetRelease(clones []*bugzilla.Bug) {
+	sort.SliceStable(clones, func(i, j int) bool {
+		if len(clones[i].TargetRelease) == 0 && len(clones[j].TargetRelease) == 0 {
+			return false
+		} else if len(clones[i].TargetRelease) == 0 {
+			return true
+		} else if len(clones[j].TargetRelease) == 0 {
+			return false
+		}
+		return clones[i].TargetRelease[0] < clones[j].TargetRelease[0]
+	})
 }
 
 // GetLandingHandler will return a simple bug search page
@@ -363,7 +389,11 @@ func GetBugHandler(client bugzilla.Client, m *metrics.Metrics) http.HandlerFunc 
 	}
 }
 
-func getMajorRelease(release string) (string, error) {
+func isTargetReleaseSet(bug *bugzilla.Bug) bool {
+	return len(bug.TargetRelease) > 0 && bug.TargetRelease[0] != "---"
+}
+
+func getMajorMinorRelease(release string) (string, error) {
 	periodIndex := strings.LastIndex(release, ".")
 	if periodIndex == -1 {
 		return "", fmt.Errorf("invalid release - must be of the form x.y.z")
@@ -371,7 +401,7 @@ func getMajorRelease(release string) (string, error) {
 	return release[:periodIndex], nil
 }
 
-func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions sets.String) (*ClonesTemplateData, int, error) {
+func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions []string) (*ClonesTemplateData, int, error) {
 	bug, err := client.GetBug(bugID)
 	if err != nil {
 		return nil, http.StatusNotFound, fmt.Errorf("Bug#%d not found: %w", bugID, err)
@@ -385,7 +415,7 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions 
 	}
 	root := clones[0]
 	// Target versions would be used to populate the CreateClone dropdown
-	targetVersions := sets.NewString(allTargetVersions.List()...)
+	targetVersions := sets.NewString(allTargetVersions...)
 	// Remove target versions of the original bug
 	targetVersions.Delete(bug.TargetRelease...)
 	g := new(errgroup.Group)
@@ -398,15 +428,16 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions 
 	clonedReleases := sets.NewString()
 	for _, clone := range clones {
 		clone := clone
-		if len(clone.TargetRelease) > 0 {
-			majorRelease, err := getMajorRelease(clone.TargetRelease[0])
+		if isTargetReleaseSet(clone) {
+			majorMinorRelease, err := getMajorMinorRelease(clone.TargetRelease[0])
 			if err != nil {
 				return nil, http.StatusInternalServerError, err
 			}
-			clonedReleases.Insert(majorRelease)
+			clonedReleases.Insert(majorMinorRelease)
+			// Remove target releases which already have clones
+			targetVersions.Delete(majorMinorRelease + ".z").Delete(majorMinorRelease + ".0")
 		}
-		// Remove target releases which already have clones
-		targetVersions.Delete(clone.TargetRelease...)
+
 		g.Go(func() error {
 			clonePRs, err := client.GetExternalBugPRsOnBug(clone.ID)
 			if err != nil {
@@ -419,35 +450,23 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions 
 	if err := g.Wait(); err != nil {
 		return nil, http.StatusInternalServerError, err
 	}
-	sort.SliceStable(clones, func(i, j int) bool {
-		if clones[i].TargetRelease == nil && clones[j].TargetRelease == nil {
-			return false
-		} else if clones[i].TargetRelease == nil {
-			return true
-		} else if clones[j].TargetRelease == nil {
-			return false
-		}
-		return clones[i].TargetRelease[0] < clones[j].TargetRelease[0]
-	})
-	sortedTargets := allTargetVersions.List()
-	sort.SliceStable(sortedTargets, func(i, j int) bool {
-		return sortedTargets[i] < sortedTargets[j]
-	})
+	sortByTargetRelease(clones)
+
 	firstClone := ""
 	lastClone := ""
 	// find the major release of the clone targeting the first release
 	for _, clone := range clones {
 		var err error
-		if len(clone.TargetRelease) > 0 {
-			firstClone, err = getMajorRelease(clone.TargetRelease[0])
+		if isTargetReleaseSet(clone) {
+			firstClone, err = getMajorMinorRelease(clone.TargetRelease[0])
 			if err != nil {
 				return nil, http.StatusInternalServerError, err
 			}
 			break
 		}
 	}
-	if len(clones[len(clones)-1].TargetRelease) > 0 {
-		lastClone, err = getMajorRelease(clones[len(clones)-1].TargetRelease[0])
+	if isTargetReleaseSet(clones[len(clones)-1]) {
+		lastClone, err = getMajorMinorRelease(clones[len(clones)-1].TargetRelease[0])
 		if err != nil {
 			return nil, http.StatusInternalServerError, err
 		}
@@ -458,28 +477,34 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions 
 	lastCloneIndex := -1
 	firstCloneNotFound := true
 	lastCloneNotFound := true
-	for i, release := range sortedTargets {
-		majRel, err := getMajorRelease(release)
+	for i, release := range allTargetVersions {
+		majorMinorRelease, err := getMajorMinorRelease(release)
 		if err != nil {
 			return nil, http.StatusInternalServerError, err
 		}
-		if majRel == firstClone && firstCloneNotFound {
+		if majorMinorRelease == firstClone && firstCloneNotFound {
 			firstCloneIndex = i
 			firstCloneNotFound = false
 		}
-		if majRel == lastClone && lastCloneNotFound {
+		if majorMinorRelease == lastClone && lastCloneNotFound {
 			lastCloneIndex = i
 			lastCloneNotFound = false
 		}
 	}
+
 	missingReleases := []string{}
 	for i := firstCloneIndex; i < lastCloneIndex; i++ {
-		majorRel, err := getMajorRelease(sortedTargets[i])
+		majorMinorRelease, err := getMajorMinorRelease(allTargetVersions[i])
 		if err != nil {
 			return nil, http.StatusInternalServerError, err
 		}
-		if !clonedReleases.Has(majorRel) {
-			missingReleases = append(missingReleases, sortedTargets[i])
+		if !clonedReleases.Has(majorMinorRelease) {
+			missingReleases = append(missingReleases, allTargetVersions[i])
+		}
+	}
+	if lastCloneIndex != -1 {
+		for i := lastCloneIndex; i < len(allTargetVersions); i++ {
+			targetVersions.Delete(allTargetVersions[i])
 		}
 	}
 	wrpr := ClonesTemplateData{
@@ -494,7 +519,7 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions 
 }
 
 // GetClonesHandler returns an HTML page with detais about the bug and its clones
-func GetClonesHandler(client bugzilla.Client, allTargetVersions sets.String, m *metrics.Metrics) http.HandlerFunc {
+func GetClonesHandler(client bugzilla.Client, allTargetVersions []string, m *metrics.Metrics) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		if req.Method != "GET" {
 			handleError(w, fmt.Errorf("invalid request method, expected GET got %s", req.Method), "invalid request method", http.StatusBadRequest, req.URL.Path, 0, m)
@@ -524,8 +549,12 @@ func GetClonesHandler(client bugzilla.Client, allTargetVersions sets.String, m *
 	}
 }
 
+func releaseInvalidErrorMsg(release string) string {
+	return fmt.Sprintf("%s cannot be parsed as release, must be of the form X.X.X", release)
+}
+
 // CreateCloneHandler will create a clone of the specified ID and return success/error
-func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m *metrics.Metrics) http.HandlerFunc {
+func CreateCloneHandler(client bugzilla.Client, sortedTargetReleases []string, m *metrics.Metrics) http.HandlerFunc {
 	return func(w http.ResponseWriter, req *http.Request) {
 		endpoint := req.URL.Path
 		if req.Method != "POST" {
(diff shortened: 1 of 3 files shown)


pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+		}
+		var sourceBug *bugzilla.Bug
+		var sourceBugMajRel string
+		for _, clone := range clones {
+			if clone.TargetRelease == nil || clone.TargetRelease[0] == "---" {
+				continue
+			}
+			cloneMajRel, err := getMajorRelease(clone.TargetRelease[0])
+			if err != nil {
+				handleError(w, err, fmt.Sprintf("invalid argument 'release' - %s", req.FormValue("release")), http.StatusBadRequest, endpoint, bugID, m)
```
alvaroaleman, line 595: Please make a proper sentence explaining the problem: `release %s can not be parsed as a release because $reason`.

pkg/backporter/backporter.go
```diff
@@ -398,7 +403,7 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions
 	clonedReleases := sets.NewString()
 	for _, clone := range clones {
 		clone := clone
-		if len(clone.TargetRelease) > 0 {
+		if clone.TargetRelease != nil && clone.TargetRelease[0] != "---" {
 			majorRelease, err := getMajorRelease(clone.TargetRelease[0])
```
alvaroaleman, line 407: This function is btw incorrectly named, sorry for not catching this earlier. Its `getMajorMinorRelease`

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			} else if clones[i].TargetRelease == nil {
+				return true
+			} else if clones[j].TargetRelease == nil {
+				return false
+			}
+			return clones[i].TargetRelease[0] < clones[j].TargetRelease[0]
+		})
+		toCloneMajRelease, err := getMajorRelease(req.FormValue("release"))
+		if err != nil {
+			handleError(w, err, fmt.Sprintf("invalid argument 'release' - %s", req.FormValue("release")), http.StatusBadRequest, endpoint, bugID, m)
```
alvaroaleman, line 584: Its completely unclear from this error how it was wrong. Didn't exist? Can not be parsed as version? Not a known version?

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			if clone.TargetRelease == nil || clone.TargetRelease[0] == "---" {
+				continue
+			}
+			cloneMajRel, err := getMajorRelease(clone.TargetRelease[0])
+			if err != nil {
+				handleError(w, err, fmt.Sprintf("invalid argument 'release' - %s", req.FormValue("release")), http.StatusBadRequest, endpoint, bugID, m)
+				return
+			}
+			if toCloneMajRelease == cloneMajRel {
+				handleError(w, err, fmt.Sprintf("clone with the same major release exists - %s", clone.TargetRelease[0]), http.StatusBadRequest, endpoint, bugID, m)
```
alvaroaleman, line 599: Can we make the error messages a bit clearer, what about `clone for major release %s already exists`? This approach of trying to always put the `%s` at the end makes for very weird to read errors

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+				return
+			}
+			if cloneMajRel > toCloneMajRelease {
```
alvaroaleman, line 602: I don't think this is correct. If we create a bug targeting 4.2 it must be from 4.3, not from anything > 4.2 like 4.4, 4.5, 4.6...

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			if clones[i].TargetRelease == nil && clones[j].TargetRelease == nil {
+				return false
+			} else if clones[i].TargetRelease == nil {
+				return true
+			} else if clones[j].TargetRelease == nil {
+				return false
+			}
+			return clones[i].TargetRelease[0] < clones[j].TargetRelease[0]
+		})
+		toCloneMajRelease, err := getMajorRelease(req.FormValue("release"))
```
alvaroaleman, line 582: There is no penality for longer variable names so please make sure they are clear, here `toCloneMajorRelease` (although as commented elsewhere that is wrong anyways, because its a majorMinor release, majorRelease would just be the `4`)

pkg/backporter/backporter.go
```diff
@@ -398,7 +403,7 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions
 	clonedReleases := sets.NewString()
 	for _, clone := range clones {
 		clone := clone
-		if len(clone.TargetRelease) > 0 {
+		if clone.TargetRelease != nil && clone.TargetRelease[0] != "---" {
```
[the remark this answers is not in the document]
GeoBK: The reason I replaced the len check with this is that the targetRelease field is always populated ie. its either [---] or the actual target release. The nil check was only added for my testcases where I might not populate the targetRelease

pkg/backporter/backporter.go
GeoBK: I believe that would be handled when we go through the loop. ie if its 4.4 then we clone a 4.3 and 4.2

pkg/backporter/backporter.go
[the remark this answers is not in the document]
alvaroaleman: the `len` check implies the nil check. If your assessment is wrong for any bug ever, the whole app will crash. Not sure if thats great.

pkg/backporter/backporter_test.go
```diff
@@ -240,25 +242,29 @@ func TestCreateCloneHandler(t *testing.T) {
...
 			},
-			statusCode: http.StatusNotFound,
-			data:       "unable to fetch bug details- Bug#1000",
-			tmplt:      errorTemplate,
-			pageTitle:  "Not Found",
+			statusCode:   http.StatusNotFound,
+			data:         "unable to fetch bug details- Bug#1000",
+			tmplt:        errorTemplate,
+			pageTitle:    "Not Found",
+			existingBugs: map[int]bugzilla.Bug{},
```
alvaroaleman, line 267: You can leave this out, read access on a nil map is fine

pkg/backporter/backporter_test.go
```diff
@@ -219,17 +219,19 @@ func TestCreateCloneHandler(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error getting bug details from Fake.")
 	}
+	toBeCloned.TargetRelease = []string{"4.5.0"}
 	expectedCloneID := toBeClonedID + 1
 	expectedClone := *toBeCloned
 	expectedClone.ID = expectedCloneID
 	expectedClone.TargetRelease = []string{clonedRelease}
 	testcases := []struct {
```
alvaroaleman, line 227: We should somehow verify the new functionality?

GeoBK: This test should verify the new functionality, since the child relation would be updated in the Blocks and DependsOn fields. But I guess I could add a testcase to verify that multiple clones are created as well when necessary

GeoBK:
@stevekuznetsov I have added the logic for auto-determining the source bug - But this would still not work properly if there are both 4.6.0 and 4.6.z in the list. (Current logic would just take 4.6.0 as the parent)
Another possible scenario is that the clones list contains 4.5.0 and 4.6.0.  The user now wants a TargetRelease of 4.4.z, in this case should we create the additional 4.5.z or should the 4.5.0 be automatically determined as the parent?

pkg/backporter/backporter.go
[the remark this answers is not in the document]
alvaroaleman: This is still unresolved

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
 			handleError(w, err, fmt.Sprintf("unable to fetch bug details- Bug#%d", bugID), http.StatusNotFound, endpoint, bugID, m)
 			return
 		}
-		// Create a clone of the bug
-		cloneID, err := client.CloneBug(bug)
+		if !allTargetVersions.Has(req.FormValue("release")) {
+			handleError(w, fmt.Errorf("invalid argument - %s is not a valid TargetRelease", req.FormValue("release")), fmt.Sprintf("invalid argument - %s is not a valid TargetRelease", req.FormValue("release")), http.StatusBadRequest, endpoint, bugID, m)
```
alvaroaleman, line 562: Please also include which target releases are valid :)

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			sortedMajorRelease = append(sortedMajorRelease, mRel)
+		}
+		// Sort the releases in descending order
+		sort.SliceStable(sortedMajorRelease, func(i, j int) bool {
+			return sortedMajorRelease[i] > sortedMajorRelease[j]
+		})
+		var targetRelease int
+		for ind, rel := range sortedMajorRelease {
+			if rel == sourceBugMajorMinorRel {
+				targetRelease = ind + 1
```
alvaroaleman, line 628: please name the first variable `idx` or just `i`, `ind` is uncommon so makes the code hearder to read

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+				break
+			}
+		}
+		newClones := []string{}
```
alvaroaleman, line 632: just use `var newClones []string`, no need to initialize the var, `append` will do that when needed

pkg/backporter/backporter.go
GeoBK: sorry i missed that, i fixed that now

pkg/backporter/backporter.go
```diff
@@ -398,12 +403,12 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions
 	clonedReleases := sets.NewString()
 	for _, clone := range clones {
 		clone := clone
-		if len(clone.TargetRelease) > 0 {
-			majorRelease, err := getMajorRelease(clone.TargetRelease[0])
+		if len(clone.TargetRelease) > 0 && clone.TargetRelease[0] != "---" {
```
stevekuznetsov, line 406: We might want to expose this to the user -- you can't clone because your bug is not targeting anything.

stevekuznetsov: Can we highlight the cell of the table in red or something?

stevekuznetsov: Mouseover text to tell you that you messed up

pkg/backporter/backporter.go
```diff
@@ -438,16 +443,16 @@ func getClonesTemplateData(bugID int, client bugzilla.Client, allTargetVersions
 	// find the major release of the clone targeting the first release
 	for _, clone := range clones {
 		var err error
-		if len(clone.TargetRelease) > 0 {
-			firstClone, err = getMajorRelease(clone.TargetRelease[0])
+		if len(clone.TargetRelease) > 0 && clone.TargetRelease[0] != "---" {
```
stevekuznetsov, line 446: The fact that we're checking that the release is not empty or set to `"---"` over N different places is an opportunity to DRY.

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
 			handleError(w, err, fmt.Sprintf("unable to fetch bug details- Bug#%d", bugID), http.StatusNotFound, endpoint, bugID, m)
 			return
 		}
-		// Create a clone of the bug
-		cloneID, err := client.CloneBug(bug)
+		if !allTargetVersions.Has(req.FormValue("release")) {
+			handleError(w, fmt.Errorf("invalid argument - %s is not a valid TargetRelease, must be one of %v", req.FormValue("release"), allTargetVersions.List()), fmt.Sprintf("invalid argument - %s is not a valid TargetRelease, must be one of %v", req.FormValue("release"), allTargetVersions.List()), http.StatusBadRequest, endpoint, bugID, m)
```
stevekuznetsov, line 562: The fact that you have two format strings here is a sign you can break that out to DRY

```go
msg :=  fmt.Sprintf("invalid argument - %s is not a valid TargetRelease, must be one of %v", req.FormValue("release"), allTargetVersions.List())
handleError(w, errors.New(msg), msg, http.StatusBadRequest, endpoint, bugID, m)
```
```

File: pkg/backporter/backporter.go [hunk 16]
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
 		if err != nil {
-			handleError(w, err, "clone creation failed", http.StatusInternalServerError, endpoint, bugID, m)
+			handleError(w, err, fmt.Sprintf("unable to convert \"ID\" parameter from string to int: %s", req.FormValue("ID")), http.StatusBadRequest, req.URL.Path, 0, m)
 			return
 		}
-		targetRelease := bugzilla.BugUpdate{
-			TargetRelease: []string{
-				req.FormValue("release"),
-			},
+		sort.SliceStable(clones, func(i, j int) bool {
```
stevekuznetsov, line 572: We do this identical sort more than once, right? We should create a named sorting function and re-use it

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			if len(clones[i].TargetRelease) == 0 && len(clones[j].TargetRelease) == 0 {
+				return false
+			} else if len(clones[i].TargetRelease) == 0 {
+				return true
+			} else if len(clones[j].TargetRelease) == 0 {
+				return false
```
stevekuznetsov, line 578: Are you use that the normal `<` for strings doesn't have this behavior?

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			if len(clone.TargetRelease) == 0 || clone.TargetRelease[0] == "---" {
+				continue
+			}
+			cloneMajorMinorRel, err := getMajorMinorRelease(clone.TargetRelease[0])
+			if err != nil {
+				handleError(w, err, fmt.Sprintf("%s cannot be parsed as release, must be of the form X.X.X", clone.TargetRelease[0]), http.StatusBadRequest, endpoint, bugID, m)
+				return
+			}
+			if toCloneMajorMinorRelease == cloneMajorMinorRel {
+				handleError(w, err, fmt.Sprintf("clone for major release %s already exists", clone.TargetRelease[0]), http.StatusBadRequest, endpoint, bugID, m)
```
stevekuznetsov, line 599: Why can the user ask to clone to a target release for which a bug already exists?

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+		}
+		var sourceBug *bugzilla.Bug
+		var sourceBugMajorMinorRel string
+		for _, clone := range clones {
+			if len(clone.TargetRelease) == 0 || clone.TargetRelease[0] == "---" {
+				continue
+			}
+			cloneMajorMinorRel, err := getMajorMinorRelease(clone.TargetRelease[0])
+			if err != nil {
+				handleError(w, err, fmt.Sprintf("%s cannot be parsed as release, must be of the form X.X.X", clone.TargetRelease[0]), http.StatusBadRequest, endpoint, bugID, m)
```
stevekuznetsov, line 595: DRY out this error- it's identical with the above for the parent

pkg/backporter/backporter.go
```diff
@@ -553,30 +558,118 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+				sourceBug = clone
+				sourceBugMajorMinorRel = cloneMajorMinorRel
+				break
+			}
 		}
-		// Updating the cloned bug with the right target version
-		if err := client.UpdateBug(cloneID, targetRelease); err != nil {
-			handleError(w, err, fmt.Sprintf("clone created - Bug#%d, but failed to specify version for the cloned bug.", cloneID), http.StatusInternalServerError, endpoint, bugID, m)
+		if sourceBug == nil {
+			handleError(w, err, fmt.Sprintf("one bug with greater release needs to be present to clone from (highest release present is %s)", clones[len(clones)-1].TargetRelease[0]), http.StatusBadRequest, endpoint, bugID, m)
```
stevekuznetsov, line 609: Shouldn't the behavior be to backfill clones for the missing ones? So if I have one for 4.7 and want to backport to 4.3 the tool will create 4.6, 4.5, 4.4 ?

pkg/backporter/backporter_test.go
```diff
@@ -20,7 +20,7 @@ var (
 	landingPage = fmt.Sprintf(htmlPageStart, "Home") + helpTemplateConstructor + htmlPageEnd
 )
 
-var allTargetVersions = sets.NewString("4.0.0", "4.1.0", "4.4.z")
+var allTargetVersions = sets.NewString("3.11.z", "4.0.z", "4.1.z", "4.2.z", "4.3.z", "4.4.z", "4.5.z", "4.6.0")
```
stevekuznetsov, line 23: Why is this a global?

pkg/backporter/backporter.go
GeoBK: I think this check is necessary since TargetRelease is []string not string

pkg/backporter/backporter.go
GeoBK: The user would ideally not be allowed to do it from the UI, since it would not be available in the dropdown.
This error would only be thrown if the user somehow calls the endpoint directly

pkg/backporter/backporter.go
GeoBK: Yes, that would be the expected behavior.
This error is expected when the user tries to create a clone with 4.6.0 as the TargetRelease even though the latest targetRelease that already exists is 4.5.0

pkg/backporter/backporter_test.go
GeoBK: I thought global would be appropriate since its used in multiple tests and if it were to change at some point it should ideally be changed for all the tests. Do you think its appropriate to create separate copies of it for each test?

pkg/backporter/backporter.go
GeoBK: I can add some JS for that, make it a UI validation instead of keeping it as an error message from the backend. 
Also this is being exposed to the end user in the form of an error message right now.
Should I proceed with adding the JS

pkg/backporter/backporter.go
GeoBK: I am guessing you mean to move `len(clone.TargetRelease) > 0 && clone.TargetRelease[0] != "---"` this into a isTargetReleaseValid function. lmk if that's not what you meant

GeoBK:
The target releases are being populated [here](https://github.com/openshift/ci-tools/blob/master/cmd/bugzilla-backporter/main.go#L56)

pkg/backporter/backporter.go
alvaroaleman: Its not possible to highlight based on content without javascript? It also needs to stay a validation on the backend, input from the frontend is not trust worthy

pkg/backporter/backporter.go
alvaroaleman: yeah and ideally return an error if that happens rather than just ignoring the bug

pkg/backporter/backporter.go
alvaroaleman: This can be simplified further by making `sortByTargetRelease` take the whole `clones` slice rather than only making it the `less` func

pkg/backporter/backporter_test.go
alvaroaleman: Yeah and the tests don't necesarily need it filled with all available versions. They only need something that is relevant for what they are testing

pkg/backporter/backporter.go
stevekuznetsov: Statically highlighting the cell would be enough IMO

pkg/backporter/backporter.go
stevekuznetsov: Not only to not ignore it, but also in your code maybe

```go
func releaseUnset(bug bugzilla.Bug) bool {
    return bug.TargetRelease == "" || bug.TargetRelease == "---"
}

// ...
if releaseUnset(clone) {
    return fmt.Errorf("could not ... for bug %d, target release field was unset", clone.Id)
}
```

pkg/backporter/backporter_test.go
```diff
@@ -211,43 +212,55 @@ func TestCreateCloneHandler(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error creating bug: %v", err)
 	}
-	clonedRelease := "4.4.z"
-	prunedReleaseSet := sets.NewString(allTargetVersions.List()...)
-	prunedReleaseSet.Delete(clonedRelease)
+	clonedRelease := "4.3.z"
+	intermediateRelease := "4.4.z"
+	prunedReleaseSetForSingleClone := sets.NewString(allTargetVersions.List()...)
```
alvaroaleman, line 217: All these variables should be defined in their individual testcase and not globally because otherwise understanding the testcase requires scrolling back and forth between the testcase and this global state. Additionally, other testcases could modify this global state.

pkg/backporter/backporter_test.go
```diff
@@ -211,43 +212,55 @@ func TestCreateCloneHandler(t *testing.T) {
...
+	prunedReleaseSetForSingleClone.Delete(intermediateRelease)
+	prunedReleaseSetForMultipleClones := sets.NewString(allTargetVersions.List()...)
+	prunedReleaseSetForMultipleClones.Delete(clonedRelease)
+	prunedReleaseSetForMultipleClones.Delete(intermediateRelease)
```
alvaroaleman, line 221: You can simplify all of this by doing `sets.NewString(slice...).Delete(clonedRelease).Delete(intermetdiateRelease()`

pkg/backporter/backporter.go
```diff
@@ -553,30 +574,111 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
 			return
 		}
+		var sortedMajorRelease []string
+		for release := range allTargetVersions {
```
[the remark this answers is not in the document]
GeoBK: allTargetVersions is a set and using the set makes sense for when I want to lookup stuff/ remove stuff from it

pkg/backporter/backporter.go
```diff
@@ -553,30 +574,111 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
 			handleError(w, err, fmt.Sprintf("unable to fetch bug details- Bug#%d", bugID), http.StatusNotFound, endpoint, bugID, m)
 			return
 		}
-		// Create a clone of the bug
-		cloneID, err := client.CloneBug(bug)
+		absentReleaseErrMsg := fmt.Sprintf("invalid argument - %s is not a valid TargetRelease, must be one of %v", req.FormValue("release"), allTargetVersions.List())
```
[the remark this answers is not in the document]
GeoBK: Thats a mistake, moving it into the if block

pkg/backporter/backporter.go
```diff
@@ -553,30 +574,111 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+		}
+
+		// Get clones and sort them
+		clones, err := client.GetAllClones(bug)
+		if err != nil {
+			handleError(w, err, fmt.Sprintf("unable to convert \"ID\" parameter from string to int: %s", req.FormValue("ID")), http.StatusBadRequest, req.URL.Path, 0, m)
+			return
+		}
+		sortByTargetRelease(clones)
+		releaseInvalidErrorMsg := "%s cannot be parsed as release, must be of the form X.X.X"
```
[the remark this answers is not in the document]
GeoBK: Why do it like this though?

pkg/backporter/backporter.go
```diff
@@ -553,30 +574,111 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+			if rel == sourceBugMajorMinorRel {
+				targetRelease = i + 1
+				break
+			}
+		}
+		var newClones []string
+		finalTargetFound := false
+		// Find source bug and keep iterating till we hit the target release
+		for !finalTargetFound {
+			if sortedMajorRelease[targetRelease] == toCloneMajorMinorRelease {
```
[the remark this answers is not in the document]
GeoBK: There should technically never be that scenario as the sourceBugMajorMinorRel must be greater than at least one release. But adding it would probable be a good precaution

pkg/backporter/backporter.go
```diff
@@ -553,30 +574,111 @@ func CreateCloneHandler(client bugzilla.Client, allTargetVersions sets.String, m
...
+				finalTargetFound = true
+			}
+			var err error
```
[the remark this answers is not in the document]
GeoBK: yup.. removed it

pkg/backporter/backporter.go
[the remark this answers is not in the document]
alvaroaleman: Yeah I agree but its a static set so constructing the sorted list once rather than during each request would make more sense or not?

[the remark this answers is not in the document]
GeoBK: There are two things being done here
1. Its used to populate the dropdown where we would need a set
2. Its used for determining the things which would need to be cloned for which a static sorted list would be fine
I could  pass in two things(a sorted list and a set) for that, or I could construct the set from the sorted list I pass in.
I'll do option 2 unless you say otherwise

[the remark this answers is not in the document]
alvaroaleman: Its a tiny amount of data and its completely static during runtime, just store both the set and the sorted list, no reason to construct one of them in the handler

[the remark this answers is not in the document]
GeoBK: I would need a separate copy of the set anyway since I am modifying it. So I would have to construct a set anyway - the only difference would be whether its from a list or a set. Thats why passing in the sorted list as you suggested would be enough

alvaroaleman:
/retitle Auto detect source bug for create clones

PR #1086, opened 2020-08-07
[publicize] extend unittest coverage to cover also the git operations

droslean:


droslean:
@jupierce Can you check if we need to add any other special test case?

PR #1089, opened 2020-08-08
Only enq changed imagestreamtags on imagestream update event

alvaroaleman:
Currently we have two controllers that care about imagestreamtags but
watch imagestreams because imagestreamtags do not support watching. One
implication of this is that whenever any tag in a given imagestream
changes, all tags in that stream get reconciled, which is very
inefficient.

This PR adds an imagestreamtagmapper which maps imagestreams to images,
giving update events a special treatment by only enqeueing the tags that
have changed.

This mapper is then used in the promotionreconciler and the
test-images-distributor. The usage of this mapper allows the
promotionreconciler to drop request coalescing.

/assign @stevekuznetsov

### DIFF
--- a/pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
+++ b/pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
@@ -1,0 +1,93 @@
+package imagestreamtagmapper
+
+import (
+	"fmt"
+	"reflect"
+
+	imagev1 "github.com/openshift/api/image/v1"
+	"github.com/sirupsen/logrus"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+// New returns a new ImageStreamTagMapper. Its purpose is to extract all ImageStreamTag events
+// from an ImageStream watch. It ignores unchanged tags on Update events.
+// If no additional filtering/mapping is required, upstream should just return its input.
+func New(upstream func(reconcile.Request) []reconcile.Request) handler.EventHandler {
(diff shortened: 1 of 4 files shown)


pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
```diff
@@ -0,0 +1,93 @@
...
+			},
+		}) {
+			q.Add(request)
+		}
+	}
+}
+
+func namedTagEventListHasElement(slice []imagev1.NamedTagEventList, element imagev1.NamedTagEventList) bool {
+	for _, item := range slice {
+		if reflect.DeepEqual(item, element) {
```
stevekuznetsov, line 61: Do we technically need to use `equality.Semantic.DeepEqual()` in case e.g. slice ordering in the object is not important?

pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
```diff
@@ -0,0 +1,93 @@
...
+	newStream, newOK := e.ObjectNew.(*imagev1.ImageStream)
+	if !oldOK || !newOK {
+		logrus.WithFields(logrus.Fields{
+			"old_type": fmt.Sprintf("%T", e.ObjectOld),
+			"new_type": fmt.Sprintf("%T", e.ObjectNew),
+		}).Error("Got object that was not an *imagev1.ImageStream")
+		return
+	}
+
+	for _, newTag := range newStream.Status.Tags {
```
stevekuznetsov, line 44: Why do we not care about changes to `imageStream.Spec.Tags`?

pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
alvaroaleman: AFAIK the only thing `equality.Semantic.DeepEqual` does is ignoring differences that are semantically irrelevant, like an empty slice/map is equal to a nil slice/map. I don't think it ignores slices ordering, because two differently ordered slice are not semantically equal.

pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
alvaroaleman: Good point, I thought imagestreamtags are constructed only based on whats in the `imagestream.Status` and `Image`, but they do indeed care about a couple of fields in the imagestreams spec.

PR #1093, opened 2020-08-10
Registry replacer: Update dockerfiles from ocp-build-data

alvaroaleman:
The diff this PR produces is here: https://github.com/openshift/release/pull/10799

First commit is pure moving code around, second commit is the actual change.

### DIFF
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -26,27 +26,37 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/api/ocpbuilddata"
 	"github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/github"
+	"github.com/openshift/ci-tools/pkg/steps/release"
 )
 
 type options struct {
-	configDir                   string
-	createPR                    bool
-	githubUserName              string
-	selfApprove                 bool
-	pruneUnusedReplacements     bool
-	pruneOCPBuilderReplacements bool
+	configDir                                    string
+	createPR                                     bool
+	githubUserName                               string
+	selfApprove                                  bool
+	ensureCorrectPromotionDockerfile             bool
+	ocpBuildDataRepoDir                          string
+	currentRelease                               ocpbuilddata.MajorMinor
+	pruneUnusedReplacements                      bool
+	pruneOCPBuilderReplacements                  bool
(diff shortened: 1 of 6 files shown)


cmd/registry-replacer/main.go
```diff
@@ -509,3 +552,65 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+}
+
+func updateDockerfilesToMatchOCPBuildData(
+	config *api.ReleaseBuildConfiguration,
+	promotionTargetToDockerfileMapping map[string]string,
+	majorMinorVersion string,
+	ignoredRepos sets.String,
+) {
+
+	// The tool only works for the current release
```
stevekuznetsov, line 575: Technically we just need it to not be a branch that's being fast forwarded _to_ right? So like `release-4.2` would be fine

cmd/registry-replacer/main.go
```diff
@@ -509,3 +552,65 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+
+	for idx, image := range config.Images {
+		promotionTarget, ok := promotedTags[string(image.To)]
+		if !ok {
+			continue
+		}
+		stringifiedPromotionTarget := fmt.Sprintf("registry.svc.ci.openshift.org/%s/%s:%s", promotionTarget.Namespace, promotionTarget.Name, image.To)
+		dockerfilePath, ok := promotionTargetToDockerfileMapping[stringifiedPromotionTarget]
+		if !ok {
+			// TODO: Is "ocp build data doesn't know this" something we should handle?
```
stevekuznetsov, line 603: If you promote e.g. a test image that is not productized that's fine I think

cmd/registry-replacer/main.go
```diff
@@ -509,3 +552,65 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+			continue
+		}
+		actualDockerFilePath := "Dockerfile"
+		if image.DockerfilePath != "" {
+			actualDockerFilePath = image.DockerfilePath
+		}
+		actualDockerFilePath = filepath.Join(image.ContextDir, actualDockerFilePath)
+		if dockerfilePath != actualDockerFilePath {
+			config.Images[idx].ContextDir = ""
```
stevekuznetsov, line 612: This sets the working directory for the build, so if their Dockerfile uses paths relative to this, and we change it, that could be breaking

cmd/registry-replacer/main.go
alvaroaleman: Right. Currently the assumption is that we will only use this going forward. We should probably make this configurable but we won't need it for older releases.

PR #1095, opened 2020-08-11
[publicize] include a link to the direct merge commit in the success message

droslean:
also part of https://issues.redhat.com/browse/DPTP-1425

/cc @openshift/openshift-team-developer-productivity-test-platform @jupierce 

### DIFF
--- a/cmd/publicize/server.go
+++ b/cmd/publicize/server.go
@@ -95,13 +95,14 @@ func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 	}, func() (login string, err error) { return s.githubLogin, nil }, s.githubTokenGenerator)
 
 	logger.Infof("Trying to merge the PR to destination: %s/%s@%s", destOrg, destRepo, baseBranch)
-	if err := s.mergeAndPushToRemote(org, repo, destOrg, destRepo, sourceRemoteResolver, baseBranch, s.dry); err != nil {
+	headCommitRef, err := s.mergeAndPushToRemote(org, repo, destOrg, destRepo, sourceRemoteResolver, baseBranch, s.dry)
+	if err != nil {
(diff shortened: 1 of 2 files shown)


cmd/publicize/server.go
```diff
@@ -150,37 +151,43 @@ func (s *server) mergeAndPushToRemote(sourceOrg, sourceRepo, destOrg, destRepo s
...
 
 	if !dry {
 		if err := repoClient.PushToCentral(branch, false); err != nil {
-			return fmt.Errorf("couldn't push to upstream repository: %w", err)
+			return "", fmt.Errorf("couldn't push to upstream repository: %w", err)
 		}
 	}
 
-	return nil
+	refs, err := repoClient.ShowRef(branch)
```
petr-muller, line 184: This is unnecessarily broad and that makes the guess below (get first SHA) not as reliable as possible. You can ask  for `refs/head/$branch` ("head of a local branch $branch", filtering out remotes and tags)  and then you can (should) check you got *exactly* a single line, so you know you have a reliable SHA.

cmd/publicize/server_test.go
```diff
@@ -433,7 +433,7 @@ and the repository exists.
 			t.Fatalf("test id: %s: %v", tc.id, err)
 		}
 
-		err := s.mergeAndPushToRemote(privateOrg, privateRepo, publicOrg, publicRepo, tc.remoteResolver, tc.branch, false)
+		_, err := s.mergeAndPushToRemote(privateOrg, privateRepo, publicOrg, publicRepo, tc.remoteResolver, tc.branch, false)
```
petr-muller, line 436: We should not ignore returned values of a method in tests for that method. We cannot predict the SHA but we can test it's a 40-character string when there is no error.

PR #1097, opened 2020-08-11
[private-org-sync] don't fail on merge conflict errors, instead print a warning

droslean:
The request is described in https://issues.redhat.com/browse/DPTP-1426

/cc @openshift/openshift-team-developer-productivity-test-platform @petr-muller @jupierce 

### DIFF
--- a/cmd/private-org-sync/main.go
+++ b/cmd/private-org-sync/main.go
@@ -457,7 +457,8 @@ func mergeRemotesAndPush(logger *logrus.Entry, git gitFunc, repoDir, srcRemote, 
 			mergeErrs = append(mergeErrs, fmt.Errorf("failed to perform merge --abort: %w", err))
 		}
 
-		return utilerrors.NewAggregate(mergeErrs)
+		logger.WithError(utilerrors.NewAggregate(mergeErrs)).Warn("error occured while fetching remote and merge")
+		return nil
 	}
 
 	cmd := []string{"push", "--tags"}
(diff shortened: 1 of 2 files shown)


cmd/private-org-sync/main.go
```diff
@@ -365,9 +365,14 @@ func (g gitSyncer) mirror(repoDir string, src, dst location) error {
 
 		if depth == unshallow {
 			logger.Info("Trying to fetch source and destination full history and perform a merge")
-			if err := mergeRemotesAndPush(logger, g.git, repoDir, srcRemote, dst.branch, destUrl.String(), g.confirm, g.gitName, g.gitEmail); err != nil {
+			err, canBeSkippedErr := mergeRemotesAndPush(logger, g.git, repoDir, srcRemote, dst.branch, destUrl.String(), g.confirm, g.gitName, g.gitEmail)
```
petr-muller, line 368: I think this is totally unnecessary. The `canBeSkippedErr` case can be handled internally in `mergeRemotesAndPush` and the callsite does not need to care, so the change can be much simpler and we don't need to make the interfaces more complicated by having multiple errors with different meanings.

cmd/private-org-sync/main.go
```diff
@@ -457,7 +469,7 @@ func mergeRemotesAndPush(logger *logrus.Entry, git gitFunc, repoDir, srcRemote,
 			mergeErrs = append(mergeErrs, fmt.Errorf("failed to perform merge --abort: %w", err))
 		}
 
-		return utilerrors.NewAggregate(mergeErrs)
+		return nil, utilerrors.NewAggregate(mergeErrs)
```
petr-muller, line 472: So at this place, we can just log the warning and return error==nil. We don't need the second return value.

petr-muller:
/hold

Until we fix e2e

PR #1098, opened 2020-08-11
ci-operator: expose ephemeral cluster versions based on parents

stevekuznetsov:
When we import a release from a stream that's configured to have a name
and be released in some way, we can piggy-back off of that name to
create a name for our release that clearly labels us as a derivative
while also uniquely identifying the test configuration that is running.
All of this is best-effort and will no-op if no release config is
present, allowing us to still mark the cluster version as one udner test
but not identifying the parent release.

/assign @petr-muller @smarterclayton

### DIFF
--- a/pkg/steps/release/create_release.go
+++ b/pkg/steps/release/create_release.go
@@ -2,6 +2,7 @@ package release
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"github.com/openshift/ci-tools/pkg/results"
 	"github.com/openshift/ci-tools/pkg/steps/utils"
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Weird registry flake?

```
error: some steps failed:
  * could not run steps: step [release:custom] failed: release "release-images-custom" failed: the pod ci-op-lhj964qb/release-images-custom failed after 8s (failed containers: release): ContainerFailed one or more containers exited

Container release exited with code 1, reason Error
---
info: Using registry public hostname default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com
Saved credentials for default-route-openshift-image-registry.apps.build01.ci.devcluster.openshift.com
error: unable to read image registry.svc.ci.openshift.org/ocp/release@sha256:48680a39e6cbd4f9532c773df2f7aaa565c8632ba26f1fcc9dd37d0dad161323: unauthorized: authentication required
---
```

pkg/steps/release/create_release.go
```diff
@@ -175,6 +176,21 @@ func (s *assembleReleaseStep) run(ctx context.Context) error {
 		return results.ForReason("missing_release").WithError(err).Errorf("could not resolve imagestream %s: %v", streamName, err)
 	}
 
+	// we want to expose the release payload as a CI version that looks just like
+	// the release versions for nightlies and CI release candidates
+	prefix := "0.0.1-0"
```
JAORMX, line 181: So, now the prefix would be 0.0.1-0 instead of 4.<something> ?

petr-muller: Only for the cases where we cannot determine the version, it's a fallback.

petr-muller:
/hold

e2e are broken now, holding to prevent the bot to go crazy with retests

pkg/steps/release/create_release.go
JAORMX: oh, somehow  I missed that. Thanks!

PR #1104, opened 2020-08-12
ci-operator: document operator-sdk testing

stevekuznetsov:

/assign @petr-muller @AlexNPavel 

https://server-ci-tools-1104.apps.ci.l2s4.p1.openshiftapps.com/help/operators

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -401,6 +401,116 @@ const templateDefinitions = `
 {{ end }}
 `
 
+const optionalOperatorOverviewPage = `<h2 id="title"><a href="#title">Testing Operators Built With The Operator SDK and Deployed Through OLM</a></h2>
+
+<p>
+<code>ci-operator</code> has dedicated support for building, deploying and testing
+operator bundles, whether the operator repository uses the Operator SDK or not.
+This document outlines how to configure <code>ci-operator</code> to build a bundle
+and index image, and how to use those in end-to-end tests.
+</p>
+
+<p>
+Consult the <code>ci-operator</code> <a href="/help/ci-operator">overview</a> and
+the step environment <a href="/help">reference</a> for detailed descriptions of the
+broader test infrastructure that an operator test is defined in.
+</p>
+
+<h3 id="bundles"><a href="#bundles">Building Operator Bundles</a></h3>
+
+<p>
+Configuring <code>ci-operator</code> to build operator bundles from a repository
+is as simple as adding a new <code>operator</code> stanza, specifying the bundles
+that can be built from the repository and what sorts of container image pull
+specification substitutions are necessary during bundle build time. Substitutions
+allow for the operator manifests to refer to images that were built from the
+repository during the test or imported from other sources. The following example
+builds an operator, a bundle and replaces the operator's pull specification in
+manifests with the version built during the test:
+</p>
+
+<code>ci-operator</code> configuration:
+{{ yamlSyntax (index . "optionalOperatorBundleConfig") }}
+
+<p>
+When configuring a bundle build, two options are available:
+</p>
+
+<ul>
+  <li><code>dockerfile_path</code>: path to the Dockerfile that builds the bundle image, defaulting to <code>bundle.Dockerfile</code></li>
+  <li><code>context_dir</code>: base directory for the bundle image build, defaulting to the root of the source tree</li>
+</ul>
+
+<h3 id="index"><a href="#index">Building an Index</a></h3>
+
+<p>
+If <code>ci-operator</code> is configured to build operator bundles from a
+repository, an index image will automatically be built to package those bundles.
+</p>
+
+<h3 id="tests"><a href="#tests">Running Tests</a></h3>
+
+<p>
+Once <code>ci-operator</code> builds the operator bundle and index, they are
+available to be used as a <code>CatalogSource</code> by OLM for deploying and
+testing the operator. Today, the only style of testing that's supported is an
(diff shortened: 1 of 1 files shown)


pkg/webreg/webreg.go
```diff
@@ -401,6 +401,112 @@ const templateDefinitions = `
...
+    tag: "8"
+  operand:           # imports the latest operand image
+    namespace: "ocp"
+    name: "operand"
+    tag: "latest"
+images:
+- from: ubi
+  to: operator                         # the pipeline:operator image, built on top of ubi:8
+  dockerfile_path: Dockerfile
+- from: ubi
```
stevekuznetsov, line 484: imho we should not allow configuring the base image for a bundle?

pkg/webreg/webreg.go
```diff
@@ -401,6 +401,112 @@ const templateDefinitions = `
...
+  to: operator-bundle                  # the pipeline:bundle image, build on top of ubi:8
+  dockerfile_path: bundle.Dockerfile
```
stevekuznetsov, line 486: imho we should default this if unset to the operator sdk default (`bundle.Dockerfile`)

pkg/webreg/webreg.go
```diff
@@ -401,6 +401,112 @@ const templateDefinitions = `
...
+  operator_manifests: bundle/manifests # where operator manifests reside in the repository
```
stevekuznetsov, line 487: imho we should default this if unset to the operator sdk default (`bundle/manifests`)

pkg/webreg/webreg.go
petr-muller: that's probably correct

pkg/webreg/webreg.go
```diff
@@ -401,6 +401,112 @@ const templateDefinitions = `
...
+  operand:           # imports the latest operand image
+    namespace: "ocp"
+    name: "operand"
+    tag: "latest"
+images:
+- from: ubi
+  to: operator                         # the pipeline:operator image, built on top of ubi:8
+  dockerfile_path: Dockerfile
+- from: ubi
+  to: operator-bundle                  # the pipeline:bundle image, build on top of ubi:8
```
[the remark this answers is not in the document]
petr-muller: Unless you have a repo with multiple operators, each of which will need a bundle. I was told that this is possible.

pkg/webreg/webreg.go
petr-muller: yeah, why not

pkg/webreg/webreg.go
petr-muller: this looks like a pitfall to me though, I don't think I've ever seen a real operator repo that would have a simple path like this

pkg/webreg/webreg.go
```diff
@@ -401,6 +401,112 @@ const templateDefinitions = `
...
+repository, an index image will automatically be built to package those bundles.
+</p>
+
+<h3 id="tests"><a href="#tests">Running Tests</a></h3>
+
+<p>
+Once <code>ci-operator</code> builds the operator bundle and index, they are
+available to be used as a <code>CatalogSource</code> by OLM for deploying and
+testing the operator. Today, the only style of testing that's supported is an
+end-to-end test, implemented with the <a href="/workflow/optional-operators-cvp-common-aws"><code>optional-operators-cvp-common-aws</code></a>
```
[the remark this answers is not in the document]
petr-muller: No, and the end to end workflow is still basically a nop, AFAIK. Plus, currently the built artifacts will not be exposed in the variables that the subscribe step consumes - that wiring is still missing.

stevekuznetsov:
@AlexNPavel @petr-muller updated, redeployed the doc -- we need to hash out the workflow to use but this should otherwise be ready

pkg/webreg/webreg.go
```diff
@@ -401,6 +401,116 @@ const templateDefinitions = `
...
+    tag: "latest"
+operator:
+  bundles: # entries create bundle images from Dockerfiles and an index containing all bundles
+  - dockerfile_path: path/to/Dockerfile # defaults to bundle.Dockerfile
+    context_dir: path/                  # defaults to .
+  substitutions:
+  - pullspec: "quay.io/openshift/operand:1.3"  # this will replace references to the operand with the imported version
+    with: "stable:operand"
+  - pullspec: "quay.io/openshift/operator:1.3" # this will replace references to the operator with the built version
+    with: "pipeline:ci-bundle-0"
```
AlexNPavel, line 496: `ci-bundle-0` is the bundle image, not the operator image. This should instead point to the operator image which would be in the `images` section in the config.

pkg/webreg/webreg.go
```diff
@@ -401,6 +401,116 @@ const templateDefinitions = `
...
+`
+
+const optionalOperatorTestConfig = `tests:
+- as: "operator-e2e"
+  steps:
+    workflow: "optional-operators-cvp-common-aws"
```
AlexNPavel, line 502: There are currently 3 env vars in the workflow that are required according to the bash for the `optional-operators-subscribe` step: https://ci-operator-configresolver-ui-ci.apps.ci.l2s4.p1.openshiftapps.com/reference/optional-operators-subscribe#line7. I think this was created before env var support in step registry, so we'll need to update the `optional-operators-subscribe` step to specifically require them.

We should probably add the required env vars to the example (`OO_INDEX`, `OO_PACKAGE`, and `OO_CHANNEL`; we can probably update the workflow to not require the index though and just set that to `ci-index`).

PR #1109, opened 2020-08-13
upgrade: annotate prowjob with version info

stevekuznetsov:

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -18,6 +18,8 @@ parameters:
   required: true
 - name: IMAGE_TESTS
   required: true
+- name: IMAGE_CLI
+  required: true
 - name: CLUSTER_TYPE
   required: true
 - name: TEST_COMMAND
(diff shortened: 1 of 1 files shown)


pkg/steps/clusterinstall/template.go
```diff
@@ -113,6 +115,33 @@ objects:
...
+          path: sa.release-bot.app.ci.config
+        secretName: appci
+
+    initContainers:
+    # annotate the information about this job so that the release controller adds it to the upgrade graph
+    - name: annotate	
+      image: ${IMAGE_CLI}
+      env:
+      - name: RELEASE_IMAGE_INITIAL
+      - name: RELEASE_IMAGE_LATEST
```
wking, line 131: for what its worth, we also have chained update jobs like `release-openshift-origin-installer-e2e-aws-upgrade-4.1-to-4.2-to-4.3-to-4.4-nightly`, which will include several versions, not just two.  Not that those need to be added to the release controller's graph, but we will presumably want them managed with similar multi-step tooling at some point.

PR #1120, opened 2020-08-17
Add the job-release annotation

GeoBK:
Add the job-release annotation based on the releases[latest].Candidate.Version
https://issues.redhat.com/browse/DPTP-1309

### DIFF
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -122,7 +122,10 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	presubmits := map[string][]prowconfig.Presubmit{}
 	postsubmits := map[string][]prowconfig.Postsubmit{}
 	var periodics []prowconfig.Periodic
-
+	var jobRelease string
+	if release, found := configSpec.Releases[cioperatorapi.LatestReleaseName]; found && release.Candidate != nil {
+		jobRelease = release.Candidate.Version
+	}
 	for _, element := range configSpec.Tests {
 		var podSpec *corev1.PodSpec
 		if element.Secret != nil {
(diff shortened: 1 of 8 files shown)


GeoBK:
cc @alvaroaleman

pkg/prowgen/prowgen.go
```diff
@@ -122,7 +122,12 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	presubmits := map[string][]prowconfig.Presubmit{}
 	postsubmits := map[string][]prowconfig.Postsubmit{}
 	var periodics []prowconfig.Periodic
-
+	var jobRelease string
+	if configSpec.Releases != nil {
+		if release, found := configSpec.Releases["latest"]; found && release.Candidate != nil {
```
alvaroaleman, line 127: You don't need the nilcheck, readaccess to a nil map is fine, it will just give `false` for the found bool.

Generally the PR looks fine but I do not know if inferring the release from here is correct, I would have thought `latest` is the `latest` release, not the one the job is targeting.

GeoBK: As per discussion with @stevekuznetsov - "We need to read the value of releases.latest.version - if present - and use it to populate the job-release field when generating jobs"

pkg/prowgen/prowgen.go
```diff
@@ -122,7 +122,10 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	presubmits := map[string][]prowconfig.Presubmit{}
 	postsubmits := map[string][]prowconfig.Postsubmit{}
 	var periodics []prowconfig.Periodic
-
+	var jobRelease string
+	if release, found := configSpec.Releases["latest"]; found && release.Candidate != nil {
```
stevekuznetsov, line 126: Please use `api.LatestReleaseName`

pkg/prowgen/prowgen_test.go
```diff
@@ -242,10 +257,19 @@ func TestGeneratePostSubmitForTest(t *testing.T) {
...
 			}},
 		},
+		{
+			name: "fourth",
+			repoInfo: &ProwgenInfo{Metadata: ciop.Metadata{
+				Org:    "organization",
+				Repo:   "repository",
+				Branch: "branch",
+			}},
+			jobRelease: "4.6",
```
stevekuznetsov, line 267: Do we have a higher level test that would validate that we're reading the correct field in the ci-op config to populate this?

pkg/prowgen/prowgen.go
stevekuznetsov: Yeah the version of the latest release we're importing is the definition of what this lablel should be

pkg/prowgen/prowgen_test.go
GeoBK: I don't think we have anything already. I would expect something like that to be in config/load_test.go
I couldn't find any existing tests for the rest of the fields, am I looking in the wrong place or did you mean that I should write the test for all the fields?

stevekuznetsov: We have integration tests for this binary -- please add a test to that one

GeoBK: Alright, I added an integration test to test for the job-release label, I have done it only for the periodic though.

GeoBK:
PR to update breaking changes https://github.com/openshift/release/pull/11141

PR #1126, opened 2020-08-19
pkg/steps/lease: add support for multiple leases

bbguimaraes:
An internal change that should not have any externally observable behavior
change (other than a few log messages).  Required to support multiple
leases per test in the future.

/hold

Reviewable, but I'll work on adding tests for the existing code tomorrow.

### DIFF
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -18,30 +18,37 @@ import (
 	"github.com/openshift/ci-tools/pkg/results"
 )
 
-const leaseEnv = "LEASED_RESOURCE"
+const DefaultLeaseEnv = "LEASED_RESOURCE"
 
 var NoLeaseClientErr = errors.New("step needs a lease but no lease client provided")
 
-// leaseStep wraps another step and acquires/releases a lease.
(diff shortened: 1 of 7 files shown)


pkg/steps/lease.go
```diff
@@ -73,25 +82,25 @@ func (s *leaseStep) Run(ctx context.Context) error {
...
+		}
+		return err
 	}
 	heartbeatCancel()
-	log.Printf("Acquired lease %q for %q", lease, s.leaseType)
-	s.leasedResource = lease
 	wrappedErr := results.ForReason("executing_test").ForError(s.wrapped.Run(ctx))
-	log.Printf("Releasing lease for %q", s.leaseType)
-	releaseErr := results.ForReason("releasing_lease").ForError(client.Release(lease))
+	log.Printf("Releasing lease for %q", s.leases[0].ResourceType)
```
stevekuznetsov, line 102: Why only print the first one?

bbguimaraes: Residual intermediary code, good catch.

petr-muller (approved):

PR #1129, opened 2020-08-19
image distributor: use local image reference policy

stevekuznetsov:
When users query for an image on a build cluster, we want them to pull
from the internal registry of that cluster. This will force the internal
registry to use pull-through and cache the image layers, which was the
point of distributing the images in the first place.

/assign @alvaroaleman

### DIFF
--- a/pkg/controller/test-images-distributor/test_images_distributor.go
+++ b/pkg/controller/test-images-distributor/test_images_distributor.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 
 	imagev1 "github.com/openshift/api/image/v1"
-	"github.com/openshift/ci-tools/pkg/util/imagestreamtagwrapper"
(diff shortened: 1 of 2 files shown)


pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -24,6 +23,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	"github.com/openshift/ci-tools/pkg/util/imagestreamtagwrapper"
```
alvaroaleman, line 26: You are impressively creative in finding ways of doing this wrong

stevekuznetsov: It's just JetBrains doing it automatically lol -- the entire imports section is hidden in my editor

PR #1130, opened 2020-08-20
E2E test: Print ci-operator output

alvaroaleman:
Currently this is redirected to a file, which makes it impossible to
debug these tests on timeout.


stevekuznetsov:
Meant to request changes instead of approving

/hold

alvaroaleman:
yeah it doesn't actually work, the problem is we never reach an error state when we timeout so we never get the output. We need to have the output unconditionally

PR #1136, opened 2020-08-21
pkg/steps: some cleanups for index+bundle image building

AlexNPavel:
This PR addresses a variety of comments from #1084 that were unaddressed when the PR merged.
These are the changes:

- Make functions to create bundle image name and detect if image is bundle name
- Add brief Godoc for api.Bundle
- Use `find` command directly in bundle substitution instead of executing through bash
- Use exact pullspecs from ResolvePullSpec, which also verifies that the image tag exists
- Add image substitutions that get built in the pipeline to the required images for bundle-sub
- Don't change global state (changing directory) in unit tests
- Handle both yaml and yml file extensions for manifests
- Handle escaping in dockerfile more cleanly
- Remove extra, empty "" lines from beginning and end of generated dockerfiles
- Expand replaceCommand unit tests by adding a .yml and .txt file

/cc @stevekuznetsov


pkg/steps/bundle_source.go
```diff
@@ -69,35 +70,56 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
-	dockerCommands = append(dockerCommands, "")
 	dockerCommands = append(dockerCommands, fmt.Sprintf("FROM %s:%s", api.PipelineImageStream, api.PipelineImageStreamTagReferenceSource))
-	is, err := s.imageClient.ImageStreams(s.jobSpec.Namespace()).Get(context.TODO(), api.StableImageStream, meta.GetOptions{})
-	if err != nil {
-		return "", fmt.Errorf("failed to get stable imagestream: %w", err)
-	}
+	imagestreams := make(map[string]*imageapi.ImageStream)
 	for _, sub := range s.config.Substitutions {
-		replaceSpec, exists := util.ResolvePullSpec(is, sub.With, false)
+		imageSubSplit := strings.Split(sub.With, ":")
```
stevekuznetsov, line 83: We should use `DependencyParts` in order to provide users the same UX here as we do everywhere else they're asked to provide an image name.

pkg/steps/bundle_source.go
```diff
@@ -69,35 +70,56 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
+		if len(imageSubSplit) != 2 {
+			return "", fmt.Errorf("image to be substituted into manifests is invalid (does not contain `:`")
+		}
+		streamName := imageSubSplit[0]
+		tagName := imageSubSplit[1]
+		var is *imageapi.ImageStream
+		if stream, ok := imagestreams[streamName]; ok {
+			is = stream
+		} else {
+			stream, err := s.imageClient.ImageStreams(s.jobSpec.Namespace()).Get(context.TODO(), streamName, meta.GetOptions{})
```
stevekuznetsov, line 93: You get a context in `run()` for the step - use that

pkg/steps/bundle_source.go
```diff
@@ -69,35 +70,56 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
+		replaceSpec, exists := util.ResolvePullSpec(is, tagName, true)
 		if !exists {
-			return "", fmt.Errorf("failed to get replacement imagestream for image tag `%s`", sub.With)
+			return "", fmt.Errorf("failed to get replacement imagestream and tag for image `%s`", sub.With)
 		}
-		// The \ character has to be escaped in the dockerfile to run correctly
-		replace := strings.ReplaceAll(replaceCommand(sub.PullSpec, replaceSpec), `\`, `\\`)
-		dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["bash", "-c", "%s"]`, replace))
+		// format command for dockerfile
+		dockerFormattedCommand := `"` + strings.Join(replaceCommand(sub.PullSpec, replaceSpec), `" "`) + `"`
```
stevekuznetsov, line 105: Hrm - shouldn't this be something like 

```go
var formattedParts []string
for _, part := range replaceCommand(sub.PullSpec, replaceSpec) {
    formattedParts = append(formattedParts. fmt.Sprintf("%q", part))
}
dockerFormattedCommand := strings.Join(formattedParts, " ")
```

?

stevekuznetsov: Maybe write a function and unit test?

pkg/steps/bundle_source.go
```diff
@@ -69,35 +70,56 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
+		dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN %s`, dockerFormattedCommand))
 	}
-	dockerCommands = append(dockerCommands, "")
 	return strings.Join(dockerCommands, "\n"), nil
 }
 
-func (s *bundleSourceStep) Requires() []api.StepLink {
-	return []api.StepLink{
-		api.InternalImageLink(api.PipelineImageStreamTagReferenceSource),
+func getPipelineSubs(subs []api.PullSpecSubstitution) []api.StepLink {
```
stevekuznetsov, line 111: Why only pipeline? We have `api.LinkForImage` to be used with `config.DependencyParts` - please see the multi-stage or dependency code

pkg/steps/bundle_source.go
```diff
@@ -69,35 +70,56 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
...
+			is = stream
+		} else {
+			stream, err := s.imageClient.ImageStreams(s.jobSpec.Namespace()).Get(context.TODO(), streamName, meta.GetOptions{})
+			if err != nil {
+				return "", fmt.Errorf("failed to get %s imagestream: %w", streamName, err)
+			}
+			imagestreams[streamName] = stream
+			is = stream
+		}
+		replaceSpec, exists := util.ResolvePullSpec(is, tagName, true)
```
stevekuznetsov, line 100: You can also use the higher level wrapper `ImageDigestFor`

pkg/steps/bundle_source.go
AlexNPavel: I've changed the Dockerfile `RUN` commands to just take the command directly, without quoting every argument. This way, it is run exactly as it would be in the `/bin/sh` of the container. This is the easiest to understand way to generate the dockerfile IMO.

petr-muller:
/hold

This looks legit:
```
could not run steps: step ci-index-gen failed: failed to get image digest for bundle `ci-bundle0`: image stream pipeline has no tag ci-bundle0 in spec or status
```

AlexNPavel:
Everything should be functional now

PR #1137, opened 2020-08-21
ci-secret-generator: quality-of-life fixes

stevekuznetsov:

/assign @GeoBK @alvaroaleman

### DIFF
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -10,10 +10,11 @@ import (
 
 	"sigs.k8s.io/yaml"
 
-	"github.com/openshift/ci-tools/pkg/bitwarden"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/logrusutil"
+
+	"github.com/openshift/ci-tools/pkg/bitwarden"
 )
 
 type options struct {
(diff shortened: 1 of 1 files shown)


pkg/bitwarden/cli.go
```diff
@@ -82,7 +82,7 @@ func (c *cliClient) loginAndListItems() error {
 			}
 			err = json.Unmarshal(out, &items)
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to parse bw item list output %s: %v", out, err)
```
GeoBK, line 85: my compiler usually throws a warning if I use %v instead of %w for errors. but I guess it doesn't matter since it doesn't look like we use Is or As in our code anyway.

cmd/ci-secret-generator/main.go
```diff
@@ -121,38 +122,60 @@ func executeCommand(command string) ([]byte, error) {
...
 			out, err := executeCommand(bwItem.Attachment.Cmd)
 			if err != nil {
-				return fmt.Errorf("failed to set attachment, item: %s, attachment: %s - %w", bwItem.ItemName, bwItem.Attachment.Name, err)
+				logrus.WithError(err).Error("failed to generate attachment")
+				return err
 			}
 			if err := bwClient.SetAttachmentOnItem(bwItem.ItemName, bwItem.Attachment.Name, out); err != nil {
-				return fmt.Errorf("failed to set attachment, item: %s, attachment: %s - %w", bwItem.ItemName, bwItem.Attachment.Name, err)
+				logrus.WithError(err).Error("failed to upload attachment")
+				return err
```
GeoBK, line 162: Why not return the wrapped error here?

pkg/bitwarden/cli.go
```diff
@@ -28,9 +28,9 @@ func newCliClient(username, password string, addSecret func(s string)) (Client,
 	return newCliClientWithRun(username, password, addSecret, func(args ...string) ([]byte, error) {
 		// bw-password is protected, session in args is not
 		logrus.WithField("args", args).Debug("running bw command ...")
-		out, err := exec.Command("bw", args...).CombinedOutput()
+		out, err := exec.Command("bw", args...).Output()
```
GeoBK, line 31: Did it fail to stderr on mac because of this?

pkg/bitwarden/cli.go
stevekuznetsov: Fixed in the other PR

cmd/ci-secret-generator/main.go
stevekuznetsov: It doesn't really matter since what we want is the nice logging

pkg/bitwarden/cli.go
stevekuznetsov: No, this is to stop it from reading stderr

petr-muller:
/hold

e2e is likely b0rked

PR #1132, opened 2020-08-20
e2e: use import secrets always

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

PR #1141, opened 2020-08-21
e2e: use import secrets always

petr-muller:
This reintroduces the actual e2e test fix commit from https://github.com/openshift/ci-tools/pull/1132 that got reverted in https://github.com/openshift/ci-tools/pull/1140. I think that the *other* commit (in `template.go`) that was the reason for the revert was not needed for the actual e2e test fix, so I think we can get the e2e fix back in separately while we figure out how to fix the `tempalte.go` change.


petr-muller:
/reopen

This *should* fix E2E

petr-muller:
Added a missing part of the e2e fix originally present in https://github.com/openshift/ci-tools/pull/1132/commits/3bc5886778970845a70c7430694f47336276bec7

petr-muller:
/hold

The new commit is mostly just a test -- even if it passes I'd probably like to understand why do we even need that.

PR #1155, opened 2020-08-24
PR creator: Only push when local branch differs from remote

alvaroaleman:
The main drawback of this implementation is that its pretty slow, because git is extremely slow compared to http requests. Additionally, there is no clientside retrying.

For doing it via the api we would need to copy big parts of the bumper downstream in order to not repeat certain http requests and waste tokens.

This can be seen in action here: https://github.com/openshift/release/pull/11197

### DIFF
--- a/pkg/github/prcreation/prcreation.go
+++ b/pkg/github/prcreation/prcreation.go
@@ -1,8 +1,11 @@
 package prcreation
 
 import (
+	"bytes"
+	"errors"
 	"flag"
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
Why doesn't upstream GitCommitAndPush want this?

alvaroaleman:
Replaced by https://github.com/kubernetes/test-infra/pull/18980

PR #1159, opened 2020-08-25
ci-operator: allow overriding dependency definitions

stevekuznetsov:
This allows e.g. the normal install step to define the install target to
be release:latest and the upgrade workflow to override it to be
release:initial.


stevekuznetsov:
https://server-ci-tools-1159.apps.ci.l2s4.p1.openshiftapps.com/help/ci-operator#dependency-overrides

PR #1164, opened 2020-08-25
ci-operator: create imagestreams with local reference policy

stevekuznetsov:
We want users of these ImageStreams to refer to their content with the
local registry's domain, not wherever their content might be sourced
during pull-through.


stevekuznetsov:
The hope is that with this change when I create a `Pod` with `spec.containers.image = imagestream:tag` the OpenShift automation will place the local registry's pull specification into the `image` field instead of the remote registry from where the layer is eventually pulled.

PR #1169, opened 2020-08-25
Prompotionreconciler: Only start once

alvaroaleman:
Currently, we add the promotionreconciler twice to the manager,
resulting in it getting started twice, resulting in a datarace as we
replace an in-use workqueue and double the configured workers.


alvaroaleman:
This is where we add it the first time: https://github.com/openshift/ci-tools/blob/4f5709712eedbf38d46372737892adb71e332d41/pkg/controller/promotionreconciler/prowjobreconciler/prowjobreconciler.go#L54

PR #1173, opened 2020-08-26
Config sanitizer: Force vsphere jobs onto the vsphere cluster

alvaroaleman:
/hold
until all vsphere jobs are actualy migrated


pkg/dispatcher/config_test.go
```diff
@@ -258,43 +258,43 @@ func TestDetermineClusterForJob(t *testing.T) {
 	}{
 		{
 			name:     "some job",
-			config:   &configWithBuildFarmWithJobs,
```
petr-muller, line 261: Looks like we can remove it from the struct too?

PR #1174, opened 2020-08-26
CI secret boostrap: Allow targeting a named group of clusters

alvaroaleman:
This PR allows defining and targeting a named group of clusters via ci-secret-bootstrap.

The resolving is implemented as part of a custom unmarshaler, which makes this completely transparent for consumers.

This PR breaks the config format, which means after we merge it, we need to adjust the config in o/release to have the secret configs below the `secret_configs` as seen in the testcase changes.

Its recommended to review this PR with whitespace changes disabled.


hongkailiu:
The changes on core code look good to me (have not checked on the tests).
Let me prepare the PR for the config change to verify if my understanding on the code is correct.

PR #1175, opened 2020-08-27
Secret Generator: Improve input mechanism

GeoBK:
1. Allow user to enter notes for an item
2. Allow user to enter multiple attachments and fields for a single
item
3. Allow user to pass parameters in the config file

### DIFF
--- a/cmd/ci-secret-generator/README.md
+++ b/cmd/ci-secret-generator/README.md
@@ -12,14 +12,14 @@ The output of the command is stored into bitwarden as the contents of the field/
 
 ```yaml
 - item_name: first_item
-  field:
-    name: field1
-    cmd: echo -n secret
+  fields:
+    - name: field1
+      cmd: echo -n secret
   attribute:
     name: password
     cmd: echo -n new_password
 - item_name: second_item
-  field:
+  fields:
     name: field2
     cmd: echo -n field2_contents
 ```
@@ -31,6 +31,22 @@ create two bitwarden entries - 'first_item' and 'second_item'
 
 * `field2` of `second_item`, would be `field2_contents` in Bitwarden with item-name `second_item`
 
+Parameters can be passed in to decrease repetition in the config file by adding the params dictionary in the config file.
+Eg.
+```yaml
+- item_name: item$(cluster)$(env)
+  fields:
+    - name: field1
+      cmd: echo "$(cluster) $(env)"
+  params:
+    cluster:
+      - build01
+      - build02
+    env:
+      - prod
+      - stagin
+```
+This would create four items with itemnames - itembuild01prod, itembuild02prod,itembuild01env, itembuild02env and the corresponding field1 which would contain the output of the corresponding echo, where the $(paramname) would be replaced with the values of the corresponding paramname.
 ## Run
 
 ```bash
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -10,11 +10,12 @@ import (
 
 	"sigs.k8s.io/yaml"
 
+	"github.com/getlantern/deepcopy"
+	"github.com/openshift/ci-tools/pkg/bitwarden"
 	"github.com/sirupsen/logrus"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/logrusutil"
-
-	"github.com/openshift/ci-tools/pkg/bitwarden"
 )
 
 type options struct {
@@ -30,20 +31,18 @@ type options struct {
 }
 
 type bitWardenItem struct {
-	ItemName   string         `json:"item_name"`
-	Field      fieldGenerator `json:"field"`
-	Attachment fieldGenerator `json:"attachment"`
-	Attribute  fieldGenerator `json:"attribute"`
+	ItemName    string              `json:"item_name"`
+	Fields      []fieldGenerator    `json:"fields,omitempty"`
+	Attachments []fieldGenerator    `json:"attachments,omitempty"`
+	Password    string              `json:"password,omitempty"`
+	Notes       string              `json:"notes"`
+	Params      map[string][]string `json:"params,omitempty"`
 }
 
 type fieldGenerator struct {
 	Name string `json:"name,omitempty"`
 	Cmd  string `json:"cmd,omitempty"`
 }
-
-const (
-	attributeTypePassword string = "password"
-)
 
 func parseOptions() options {
 	var o options
@@ -97,6 +96,10 @@ func (o *options) completeOptions(secrets sets.String) error {
 	return o.validateCompletedOptions()
 }
 
+func cmdEmptyErr(itemIndex, entryIndex int, entry string) error {
+	return fmt.Errorf("config[%d].%s[%d]: empty field not allowed for cmd if name is specified", itemIndex, entry, entryIndex)
+}
+
 func (o *options) validateCompletedOptions() error {
 	if o.bwPassword == "" {
 		return fmt.Errorf("--bw-password-file was empty")
@@ -106,13 +109,21 @@ func (o *options) validateCompletedOptions() error {
 		if bwItem.ItemName == "" {
 			return fmt.Errorf("config[%d].itemName: empty key is not allowed", i)
 		}
-		if bwItem.Attribute.Name != attributeTypePassword && bwItem.Attribute.Name != "" {
-			return fmt.Errorf("config[%d].attribute: only the '%s' is supported, not %s", i, attributeTypePassword, bwItem.Attribute.Name)
+
+		for fieldIndex, field := range bwItem.Fields {
+			if field.Name != "" && field.Cmd == "" {
+				return cmdEmptyErr(i, fieldIndex, "fields")
+			}
 		}
-		if (bwItem.Field.Name != "" && bwItem.Field.Cmd == "") ||
-			(bwItem.Attribute.Name != "" && bwItem.Attribute.Cmd == "") ||
-			(bwItem.Attachment.Name != "" && bwItem.Attachment.Cmd == "") {
-			return fmt.Errorf("config[%d]: empty field not allowed for cmd if name is specified for any of attribute, field or attachment", i)
+		for attachmentIndex, attachment := range bwItem.Fields {
+			if attachment.Name != "" && attachment.Cmd == "" {
+				return cmdEmptyErr(i, attachmentIndex, "attachments")
+			}
+		}
+		for paramName, params := range bwItem.Params {
+			if len(params) == 0 {
+				return fmt.Errorf("at least one argument required for param: %s, itemName: %s", paramName, bwItem.ItemName)
+			}
 		}
 	}
 	return nil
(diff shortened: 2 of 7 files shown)


cmd/ci-secret-generator/main.go
```diff
@@ -96,7 +99,9 @@ func (o *options) completeOptions(secrets sets.String) error {
 	}
 	return o.validateCompletedOptions()
 }
-
+func cmdEmptyErr(itemIndex, entryIndex int, entry string) error {
+	return fmt.Errorf("config[%d].%s[%d]: empty field not allowed for cmd if name is specified", itemIndex, entry, entryIndex)
+}
```
alvaroaleman, line 104: Please have an empty line before and after function declarations

cmd/ci-secret-generator/main.go
```diff
@@ -127,55 +153,104 @@ func executeCommand(command string) ([]byte, error) {
...
+	for _, bwItemWithParams := range bwItems {
+		bwItemsQ := []bitWardenItem{bwItemWithParams}
+		for paramName, params := range bwItemWithParams.Params {
+			bwItemsProcessed := []bitWardenItem{}
+			for _, qItem := range bwItemsQ {
+				for _, param := range params {
+					argItem := bitWardenItem{}
+					err := deepcopy.Copy(&argItem, &qItem)
+					if err != nil {
+						logrus.Fatal(err)
```
alvaroaleman, line 170: Don't use fatals outside of `main()`

cmd/ci-secret-generator/main.go
```diff
@@ -127,55 +153,104 @@ func executeCommand(command string) ([]byte, error) {
...
-			logger.Info("processing field")
-			out, err := executeCommand(bwItem.Field.Cmd)
-			if err != nil {
-				logrus.WithError(err).Error("failed to generate field")
-				return err
-			}
-			if err := bwClient.SetFieldOnItem(bwItem.ItemName, bwItem.Field.Name, out); err != nil {
-				logrus.WithError(err).Error("failed to upload field")
-				return err
+	for _, bwItemWithParams := range bwItems {
```
alvaroaleman, line 161: Please have a dedicated function that does the resolving and do it before any command is executed, to not mix config resolving steps with program execution steps

cmd/ci-secret-generator/main.go
```diff
@@ -127,55 +153,104 @@ func executeCommand(command string) ([]byte, error) {
...
+			for _, field := range bwItem.Fields {
+				if field.Name != "" {
+					logger = logger.WithFields(logrus.Fields{
+						"field":   field.Name,
+						"command": field.Cmd,
+					})
+					logger.Info("processing field")
+					out, err := executeCommand(field.Cmd)
+					if err != nil {
+						logrus.WithError(err).Error("failed to generate field")
```
alvaroaleman, line 201: IN many cases, the error will be very hard to reason about without the `out`

cmd/ci-secret-generator/main.go
```diff
@@ -127,55 +153,104 @@ func executeCommand(command string) ([]byte, error) {
 	return out, nil
 }
 
+func replaceParameter(paramName, param, template string) string {
+	return strings.ReplaceAll(template, fmt.Sprintf("$(%s)", paramName), param)
+}
+
 func updateSecrets(bwItems []bitWardenItem, bwClient bitwarden.Client) error {
```
alvaroaleman, line 160: For the future: If you ever end up writing a function that contains 9 loops in four nesting levels, that is a very clear sign that you are not factoring your code well

cmd/ci-secret-generator/main.go
alvaroaleman: Furthermore, we want to keep our ability to process everything we can before we exit the program with an error. The way this is built, we will stop processing on the first error here, rather then process the rest

cmd/ci-secret-generator/main.go
```diff
@@ -127,55 +153,104 @@ func executeCommand(command string) ([]byte, error) {
...
-				return err
+			if bwItem.Notes != "" {
+				logger = logger.WithFields(logrus.Fields{
+					"notes": bwItem.Notes,
+				})
+				logger.Info("adding notes")
+				if err := bwClient.UpdateNotesOnItem(bwItem.ItemName, bwItem.Notes); err != nil {
+					logrus.WithError(err).Error("failed to update notes")
+					return err
+				}
```
alvaroaleman, line 253: The new logic has no tests at all?

cmd/ci-secret-generator/main.go
GeoBK: i am assuming you mean the cmd here.
Since the out would be irrelevant if the err != nil, right?

cmd/ci-secret-generator/main.go
GeoBK: I couldn't think of a better way of doing this, could you please let me know how you would approach it?

cmd/ci-secret-generator/main.go
GeoBK: added tests for the new logic

cmd/ci-secret-generator/README.md
```diff
@@ -31,6 +31,22 @@ create two bitwarden entries - 'first_item' and 'second_item'
...
+      cmd: echo "$(cluster) $(env)"
+  params:
+    cluster:
+      - build01
+      - build02
+    env:
+      - prod
+      - stagin
+```
+This would create four items with itemnames - itembuild01prod, itembuild02prod,itembuild01env, itembuild02env and the corresponding field1 which would contain the output of the corresponding echo, where the $(paramname) would be replaced with the values of the corresponding paramname.
```
stevekuznetsov, line 49: This approach is a little fraught - why do we assume that the combinatorics should be a full cross product between the input parameter lists? IMHO a more resilient approach is:

```yaml
params:
- cluster: build01
  env: prod
- cluster: build02
  env: prod
...
```

I'm also not super sure that this is the most straightforward, transparent approach - doesn't YAML have templating/anchors built in that can do this for us?

GeoBK: We can do it in that manner as well, but taking the cross-product is what I and @alvaroaleman  had discussed and decided on to reduce the stuff which would need to be explicitly typed out by the user(which was the overall intention of this change). 
Regarding YAML templating I am not aware of anything and my initial search only gave me some third party tools and default go templating which I'm not sure would serve our purpose directly. Is there something specific you had in mind?

cmd/ci-secret-generator/main.go
```diff
@@ -170,16 +249,27 @@ func updateSecrets(bwItems []bitWardenItem, bwClient bitwarden.Client) error {
 			logger.Info("processing attribute")
 			out, err := executeCommand(bwItem.Attribute.Cmd)
 			if err != nil {
-				logrus.WithError(err).Error("failed to generate attribute")
-				return err
+				errs = append(errs, fmt.Errorf("bwItem.ItemName: %s, bwItem.Password:, %s failed: %w", bwItem.ItemName, bwItem.Attribute.Cmd, err))
```
stevekuznetsov, line 252: Why are we getting back to this error message? It's not useful, the logged and structured one is

cmd/ci-secret-generator/main.go
```diff
@@ -30,10 +31,12 @@ type options struct {
...
 
 type bitWardenItem struct {
-	ItemName   string         `json:"item_name"`
-	Field      fieldGenerator `json:"field"`
-	Attachment fieldGenerator `json:"attachment"`
-	Attribute  fieldGenerator `json:"attribute"`
+	ItemName    string              `json:"item_name"`
+	Fields      []fieldGenerator    `json:"fields,omitempty"`
+	Attachments []fieldGenerator    `json:"attachments,omitempty"`
+	Attribute   fieldGenerator      `json:"attribute,omitempty"`
```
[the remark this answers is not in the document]
stevekuznetsov: Hm - if we only want to support updating the password, we should just make this a password field.

cmd/ci-secret-generator/main_test.go
```diff
@@ -0,0 +1,391 @@
...
+							Name: "Attachment1prod",
+							Cmd:  "echo -n Attachment1prod",
+						},
+					},
+					Attribute: fieldGenerator{
+						Name: "password1prod",
+						Cmd:  "echo -n password1prod",
+					},
+					Notes: "Note1prod",
+					Params: map[string][]string{
```
stevekuznetsov, line 291: Seems weird to have these on the output type since they've been consumed in the resolution

cmd/ci-secret-generator/main_test.go
```diff
@@ -0,0 +1,391 @@
...
+			sort.Slice(outputBwItems, func(p, q int) bool {
+				return outputBwItems[p].ItemName < outputBwItems[q].ItemName
+			})
+			equalError(t, tc.expectedError, err)
+			equal(t, tc.expected, outputBwItems)
+		})
+	}
+}
+
+func equalError(t *testing.T, expected, actual error) {
```
stevekuznetsov, line 378: Why isn't this `reflect.DeepEqual()`?

cmd/ci-secret-generator/main_test.go
```diff
@@ -0,0 +1,391 @@
...
+	if expected != nil && actual == nil || expected == nil && actual != nil {
+		t.Errorf("expecting error %v, got %v", expected, actual)
+	}
+	if expected != nil && actual != nil && expected.Error() != actual.Error() {
+		t.Errorf("expecting error msg %q, got %q", expected.Error(), actual.Error())
+	}
+}
+
+func equal(t *testing.T, expected, actual interface{}) {
```
stevekuznetsov, line 387: What utility does this serve?

GeoBK: line 333 - to compare the output of the function with the expected output

stevekuznetsov: Right, but why do you feel like you need this method? Just use the actual code underneath.

cmd/ci-secret-generator/main_test.go
GeoBK: I believe thats to explicitly print out a prettier difference message since we know the types here. This looked like a common pattern in some of the other tools I saw. I can get rid of the equalError and just use the "equal" function below instead.

cmd/ci-secret-generator/main_test.go
GeoBK: True, but to remove this alone I would need to start copying the terms explicitly instead of just using a deep copy function. I didn't think that removing the params field is worth moving stuff over by hand.

cmd/ci-secret-generator/main_test.go
stevekuznetsov: IIRC `cmp.Diff()` will do what you need regardless of type - and the type underneath isn't checked here, either, you just check non-nil `error` interfaces. Unless there is a special, significant reason, this should be 100% of test comparisons:

```go
if diff := cmp.Diff(a, b); diff != "" {
    t.Errorf("%s: got the incorrect thingy: %v", testCase.name, diff)
}
```

Some specific types may require a novel `Comparer` but it's rare.

cmd/ci-secret-generator/main.go
```diff
@@ -109,10 +116,27 @@ func (o *options) validateCompletedOptions() error {
...
+		}
+		for attachmentIndex, attachment := range bwItem.Fields {
+			if attachment.Name != "" && attachment.Cmd == "" {
+				return cmdEmptyErr(i, attachmentIndex, "attachments")
+			}
+		}
+		if bwItem.Attribute.Name != "" && bwItem.Attribute.Cmd == "" {
+			return cmdEmptyErr(i, 0, "attribute")
+		}
+		// Adding the notes not empty check here since we dont want to overwrite any notes that might already be present
```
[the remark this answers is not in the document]
GeoBK: I missed the comments while refactoring, I have moved them to the right place now

cmd/ci-secret-generator/main.go
GeoBK: I am also outputting the earlier structure of the error message on the line below, but I am also returning the wrapped error message

cmd/ci-secret-generator/main_test.go
GeoBK: right now I would be using it in only 2 places, but as we add more tests I would keep writing the extra "t.Errorf("actual differs from expected:\n%s", cmp.Diff(expected, actual))" instead of just calling the function.

cmd/ci-secret-generator/main.go
```diff
@@ -127,39 +151,94 @@ func executeCommand(command string) ([]byte, error) {
...
+func replaceParameter(paramName, param, template string) string {
+	return strings.ReplaceAll(template, fmt.Sprintf("$(%s)", paramName), param)
+}
+
+func processBwParameters(bwItems []bitWardenItem) ([]bitWardenItem, error) {
+	var errs []error
+	processedBwItems := []bitWardenItem{}
+	for _, bwItemWithParams := range bwItems {
+		hasErrors := false
+		bwItemsQ := []bitWardenItem{bwItemWithParams}
```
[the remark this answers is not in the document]
GeoBK: bitWardenItems queue is what its supposed to mean - it started as a q but now is a q ish structure. I'll change it to something else

stevekuznetsov (approved):

PR #1176, opened 2020-08-27
registry docs: cleanups

petr-muller:
Fix step parameter example. Before, the example suggested the `env` is a
member of the `TestStepConfiguration` struct, but it is actually a
member of the `MultiStageTestConfiguration` one (under `steps:`).

Also:
- fix anchors
- provide captions for the examples to separate them


petr-muller:
https://server-ci-tools-1176.apps.ci.l2s4.p1.openshiftapps.com/help

PR #1178, opened 2020-08-27
api/config: add support for using ci-index as step dependency

AlexNPavel:
This allows users to depend on the `ci-index` image in multistage test steps, which is required for optional-operator tests.

/cc @petr-muller

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -175,6 +175,10 @@ func (config *ReleaseBuildConfiguration) validateTestStepDependencies() []error 
 						if config.RpmBuildCommands == "" {
 							errs = append(errs, validationError("this dependency requires built RPMs, which are not configured"))
 						}
+					case string(PipelineImageStreamTagReferenceIndexImage):
+						if config.Operator == nil {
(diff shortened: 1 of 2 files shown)


petr-muller (commented):
This looks it only adds the validation that if you attempt to use such step, you need to have `operator:` stanza, but not a fix for https://github.com/openshift/release/pull/11339#issuecomment-682081163

Or am I missing something and this is enough?

AlexNPavel:
> This looks it only adds the validation that if you attempt to use such step, you need to have `operator:` stanza, but not a fix for [openshift/release#11339 (comment)](https://github.com/openshift/release/pull/11339#issuecomment-682081163)
> 
> Or am I missing something and this is enough?

The issue in that PR is that it is failing this validation. I believe that this is enough, because all the other code handling the resolution of step dependencies is in place. This validation is the only thing failing, because it sees that the `ci-index` image is a `pipeline` image, but the validation thinks that the required section (either in `images` or as a base image) hasn't been set up in the config.

petr-muller:
:facepalm: of course

I'm apparently not in the greatest mental shape today :)

petr-muller (approved):

PR #1179, opened 2020-08-28
webreg: document build_root images

petr-muller:
https://server-ci-tools-1179.apps.ci.l2s4.p1.openshiftapps.com/help/ci-operator#buildroot

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -498,6 +498,28 @@ either:
   </li>
 </ul>
 
+<h4 id="buildroot"><a href="#buildroot">Build Root Image</a></h4>
+<p>
+The build root image must contain all dependencies for building executables and
+non-image artifacts. Additionally, <code>ci-operator</code> requires this image
+to include a <code>git</code> executable in <code>$PATH</code>. Most repositories
+will want to use an image already present in the cluster, using the <code>image_stream_tag</code>
+stanza like described in <a href="#inputs">Configuring Inputs</a>.
+</p>
+
+<p>
+Alternatively, a project can be configured to build a build root image using
+a <code>Dockerfile</code> in the repository:
+</p>
+
+{{ yamlSyntax (index . "ciOperatorProjectImageBuildroot") }}
+
+<p>
+In this case, the <code>Dockerfile</code> will <b>always</b> be obtained from
+current <code>HEAD</code> of the given branch, even if ci-operator runs in the
+context of a PR that updates that <code>Dockerfile</code>.
+</p>
+
 <h4 id="artifacts"><a href="#artifacts">Building Artifacts</a></h4>
 
 <p>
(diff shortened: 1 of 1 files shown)


petr-muller:
ping @hongkailiu @stevekuznetsov

pkg/webreg/webreg.go
```diff
@@ -498,6 +498,28 @@ either:
   </li>
 </ul>
 
+<h4 id="buildroot"><a href="#buildroot">Buildroot Image</a></h4>
```
stevekuznetsov, line 501: nit: could we use "Build Root" or `build_root`?

pkg/webreg/webreg.go
```diff
@@ -498,6 +498,28 @@ either:
...
+<p>
+The buildroot image must contain all dependencies for building executables and
```
stevekuznetsov, line 503: Can we move this section below the `"artifacts"` one? We introduce `build_root` there and it makes sense to into more depth there

pkg/webreg/webreg.go
petr-muller: "Build Root" it is

pkg/webreg/webreg.go
petr-muller: That's what I did initially, and I believe the current state is better. Build root is something between an input and a built artifact - it is special in a way that it can be both built and imported. Half of the section describes what the build root needs to contain, which is unrelated to building artifacts.

Having it separate also allows it to have an header with an anchor so it's linkable (custom buildroot is not that rare #forum topic) and not mixed with other sections. I think that the two sections together would be unnecessarily long.

So if you insist then yes, but I ask you to consider ^^^ ;)

pkg/webreg/webreg.go
```diff
@@ -498,6 +498,28 @@ either:
   </li>
 </ul>
 
+<h4 id="buildroot"><a href="#buildroot">Build Root Image</a></h4>
+<p>
+The build root image must contain all dependencies for building executables and
```
stevekuznetsov, line 503: So originally I was going to ask you to explain what the build root was in the first place, since it's not really introduced here - then I realized that the sentences that explain it and introduce the concept already exist, a paragraph below, so it seemed best to first introduce it, then explain the options. I think what I want is that this explanation comes after an introduction so someone reading top-down understands what we're talking about. At this point in the doc we have not yet even talked about the artifacts we build or introduced the build root as a concept so it is confusing. If you want to keep it up here and just give it the introduction it needs that's fine, or move it down (and keep the header) after the current introduction.

petr-muller: I'm not sure I follow. The build root is already mentioned in the paragraph above ("Configuring inputs") - I used the definition there verbatim and just elaborated on it by describing the other (first one is the import) way how it can get into a ci-operator run.

What exact paragraph you mean by "the introduction" of the build root? The _Starting FROM the image described as the build_root..._ one? Because to me that paragraph does not deal with build root at all, just mentions it. Would adding a sentence or two why the build root image is necessary to my new section achieve what you want?

stevekuznetsov: I'm dumb, did not scroll up high enough

stevekuznetsov (approved):

PR #1184, opened 2020-08-31
OCP Build Data: Don't replace with entrys in stream.yaml that have mirror: false

alvaroaleman:


pkg/api/ocpbuilddata/types.go
```diff
@@ -278,6 +279,9 @@ func replaceStream(streamName string, streamMap StreamMap) (string, error) {
 	if replacement.UpstreamImage == "" {
 		return "", fmt.Errorf("stream.yml.%s.upstream_image is an empty string", streamName)
 	}
+	if replacement.Mirror != nil && !*replacement.Mirror {
```
jupierce, line 282: Please let nil be equivalent to false.

alvaroaleman: @jupierce updated

PR #1186, opened 2020-08-31
Imagebuild: connection reset by peer is a transient failure

alvaroaleman:
Sample: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/1185/pull-ci-openshift-ci-tools-master-images/1300548570507644928


bbguimaraes:
I'm wary of adding such a generic `strerror` message (even though it would not be the first on the list).  How about `error creating buildah builder`, which appears earlier in the message?

droslean (approved):

alvaroaleman:
> I'm wary of adding such a generic strerror message (even though it would not be the first on the list). How about error creating buildah builder, which appears earlier in the message?

I do realize that there are also other parts that could emit this error message but I would argue that a connection reset is guaranteed to be transient

PR #1189, opened 2020-09-01
update test-infra vendors

droslean:


stevekuznetsov:
@droslean is that change expected on load/dump?

droslean:
> is that change expected on load/dump?

@stevekuznetsov Yes it should be.

PR #1191, opened 2020-09-01
Add allow-list as input to testgrid

GeoBK:
This change allows the release-type to be specified in the allowed
list thereby deprecating the need to add the release-type annotation.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -28,6 +28,7 @@ type options struct {
 	releaseConfigDir  string
 	testGridConfigDir string
 	prowJobConfigDir  string
+	jobsAllowListFile string
 }
 
 func (o *options) Validate() error {
@@ -40,6 +41,7 @@ func (o *options) Validate() error {
 	if o.testGridConfigDir == "" {
 		return errors.New("--testgrid-config is required")
 	}
+
 	return nil
 }
 
@@ -49,6 +51,7 @@ func gatherOptions() options {
 	fs.StringVar(&o.prowJobConfigDir, "prow-jobs-dir", "", "Path to a root of directory structure with Prow job config files (ci-operator/jobs in openshift/release)")
 	fs.StringVar(&o.releaseConfigDir, "release-config", "", "Path to Release Controller configuration directory.")
 	fs.StringVar(&o.testGridConfigDir, "testgrid-config", "", "Path to TestGrid configuration directory.")
+	fs.StringVar(&o.jobsAllowListFile, "allow-list", "", "Path to file containing jobs to be overridden to informing jobs")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("could not parse input")
 	}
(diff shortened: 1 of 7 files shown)


cmd/testgrid-config-generator/main.go
```diff
@@ -223,12 +225,23 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatalf("Failed to load Prow jobs %s", o.prowJobConfigDir)
 	}
+
+	// read the list of jobs from the allow list along with its release-type
```
stevekuznetsov, line 229: Blocking jobs must be in the release controller configuration, since it's inclusion in that config that gets them triggered in the first place, so we should validate that none of these are set to blocking here.

cmd/testgrid-config-generator/main.go
```diff
@@ -223,12 +225,23 @@ func main() {
...
+	if err := yaml.Unmarshal(data, &allowList); err != nil {
+		logrus.WithError(err).Fatalf("could not unmarshal allow-list at %s", o.jobsAllowListFile)
+	}
+
 	for _, p := range jobConfig.Periodics {
 		name := p.Name
 		calculateDays := len(p.Cron) > 0 || len(p.Interval) > 0
-
 		var dashboardType string
-		switch label := p.Labels["ci.openshift.io/release-type"]; label {
```
stevekuznetsov, line 231: We should support both cases while we migrate to the allow-list and remove the annotations.

GeoBK: I think I have added all the cases where there is the release-type annotation to the _allow-list.yaml file. Unless you are expecting users to continue adding these annotations to the jobs configs directly, we should be ok I think. Is there anything I missed?

cmd/testgrid-config-generator/main.go
GeoBK: I can add that check there. 
Also some of the jobs seem to be set to broken, is that something that needs to be validated?

cmd/testgrid-config-generator/main.go
GeoBK: This would also force users to start using the _allow-list.yaml file instead of continuing with modifying the jobs config

cmd/testgrid-config-generator/main.go
stevekuznetsov: No, we shouldn't bother making any statement other than this file not containing blocking jobs

test/integration/testgrid-config-generator/config/jobs/org/repo/org-repo-master-periodics.yaml
```diff
@@ -3,7 +3,6 @@ periodics:
   decorate: true
   interval: 12h
   labels:
-    ci.openshift.io/release-type: informing
```
stevekuznetsov, line 6: This test case should not change, until we remove support for the annotation.

test/integration/testgrid-config-generator/config/_allow-list.yaml
```diff
@@ -0,0 +1,3 @@
+release-openshift-origin-job: informing
```
stevekuznetsov, line 1: I was expecting a new set of jobs in the allow-list to cause us to have new output from the tool. This test as written doesn't prove that your code works :)

cmd/testgrid-config-generator/main.go
```diff
@@ -223,12 +226,37 @@ func main() {
...
+			logrus.WithError(err).Fatalf("could not read allow-list at %s", o.jobsAllowListFile)
+		}
+		if err := yaml.Unmarshal(data, &allowList); err != nil {
+			logrus.WithError(err).Fatalf("could not unmarshal allow-list at %s", o.jobsAllowListFile)
+		}
+		// Validate that there is no entry in the allow-list marked as blocking
+		// since blocking jobs must be in the release controller configuration
+		for jobName, releaseType := range allowList {
+			if releaseType == "blocking" {
+				logrus.Fatalf("release_type \"blocking\" not permitted in the allow-list for %s, blocking jobs must be in the release controller configuration", jobName)
```
stevekuznetsov, line 244: We should unit-test validation logic

PR #1192, opened 2020-09-01
ci-operator: check if init containers are running

bbguimaraes:
> A Pod that is initializing is in the Pending state but should have a
> condition Initialized set to true.

https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -553,8 +553,8 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
 				log.Printf("warning: failed to get pod %s: %v", name, err)
 				continue
 			}
-			if pod.Status.Phase != coreapi.PodRunning && time.Since(pod.CreationTimestamp.Time) > 30*time.Minute {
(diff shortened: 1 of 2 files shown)


pkg/steps/template.go
```diff
@@ -553,7 +553,7 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
 				log.Printf("warning: failed to get pod %s: %v", name, err)
 				continue
 			}
-			if pod.Status.Phase != coreapi.PodRunning && time.Since(pod.CreationTimestamp.Time) > 30*time.Minute {
+			if !podHasStarted(pod) && time.Since(pod.CreationTimestamp.Time) > 30*time.Minute {
```
alvaroaleman, line 556: Can we reduce the time? We picked that big timeframe explicitly because it includes `initContainers` but now we would assume a couple of mins (10 tops?) to be sufficient, right? The worst case is basically `We have to create a new node and download our image there`

bbguimaraes: Updated.  I also changed the implementation slightly after I realized the check would become useless for multi-stage tests, which always have one init container.

PR #1199, opened 2020-09-02
Promotionreconciler: Filter out ISTs to which we don't promote

alvaroaleman:


pkg/promotion/promotion.go
```diff
@@ -24,6 +25,32 @@ func PromotesImagesInto(configSpec *cioperatorapi.ReleaseBuildConfiguration, pro
...
+	result := sets.String{}
+
+	if isDisabled(configSpec) {
+		return result
+	}
+
+	namespace := extractPromotionNamespace(configSpec)
+	name := extractPromotionName(configSpec)
+
+	if namespace == "" || name == "" {
```
droslean, line 39: It is good to have this check here with its unittest cases included, but I don't think that we will ever hit this. There is already a validation in the ci-operator config and I am pretty sure that a configuration that misses those fields in the promotion will never exist in the release repo because the presubmits won't pass (finger crossed, maybe its a good idea to double check this.)

alvaroaleman: Can you approve the PR if its fine?

droslean (approved):

PR #1203, opened 2020-09-04
pkg/steps/lease: print resource metrics on failure

bbguimaraes:
The upstream client and server errors are confusing because both
non-existent and exhausted resources result in an HTTP 404 error.

We often receive support requests because the error message (`failed to
acquire lease: resources not found`) can give the false impression that
the test infrastructure is at fault.

Sample output:

```
2020/09/04 18:08:42 Acquiring lease for "aws-quota-slice"
2020/09/04 18:08:47 warning: Failed to acquire resource, current capacity: 0 free, 1 leased
2020/09/04 18:08:47 Ran for 2m17s
error: some steps failed:
  * could not run steps: step test failed: failed to acquire lease: resources not found
```

### DIFF
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -81,17 +81,20 @@ func (s *leaseStep) run(ctx context.Context) error {
 	ctx, cancel := context.WithCancel(ctx)
 	heartbeatCtx, heartbeatCancel := context.WithCancel(ctx)
 	go heartbeatNamespace(s.namespace, s.namespaceClient, heartbeatCtx)
-	lease, err := client.Acquire(s.leaseType, ctx, cancel)
+	resource, err := client.Acquire(s.leaseType, ctx, cancel)
 	if err != nil {
 		heartbeatCancel()
+		if err == lease.ErrNotFound {
+			printResourceMetrics(*s.client, s.leaseType)
+		}
 		return results.ForReason(results.Reason("acquiring_lease:"+s.leaseType)).WithError(err).Errorf("failed to acquire lease: %v", err)
 	}
 	heartbeatCancel()
-	log.Printf("Acquired lease %q for %q", lease, s.leaseType)
-	s.leasedResource = lease
+	log.Printf("Acquired lease %q for %q", resource, s.leaseType)
+	s.leasedResource = resource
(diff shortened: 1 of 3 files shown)


pkg/steps/lease.go
```diff
@@ -136,3 +139,12 @@ func heartbeatNamespace(namespace func() string, client coreclientset.NamespaceI
...
 	}
 }
+
+func printResourceMetrics(client lease.Client, rtype string) {
+	m, err := client.Metrics(rtype)
+	if err != nil {
+		log.Printf("warning: Could not get resource metrics: %v", err)
+		return
+	}
+	log.Printf("warning: Failed to acquire resource, current capacity: %d free, %d leased", m.Free, m.Leased)
```
petr-muller, line 149: IMO this is not a "warning" anymore, just "Failed to ..." makes sense on its own. Also, can we say "Failed to acquire resource in Xm time window." or something similar to indicate how long we attempted to get a lease (to avoid implying we checked just once).

I'm a little worried that we'll get pinged about races where we first give up, then something will return a lease, then we get the metric, say something like `1 free, 149 lease`, but other than that I generally like showing the metrics.

bbguimaraes: Changed to `error: …` to follow the format of other `log` messages.

bbguimaraes: > Also, can we say "Failed to acquire resource in Xm time window." or something similar to indicate how long we attempted to get a lease (to avoid implying we checked just once).

That is usually deducible from the time stamps.  In the original sample output, I used a 5s timeout for testing:

```
2020/09/04 18:08:42 Acquiring lease for "aws-quota-slice"
2020/09/04 18:08:47 warning: Failed to acquire resource, current capacity: 0 free, 1 leased
```

Do you think that is enough?

> I'm a little worried that we'll get pinged about races where we first give up, then something will return a lease, then we get the metric, say something like `1 free, 149 lease`, but other than that I generally like showing the metrics.

I considered that, but it should be exceedingly rare (although, with a code base such as `boksos`', you never know).  I think we can adjust later if it manifests frequently.

petr-muller (approved):

bbguimaraes:

PR #1204, opened 2020-09-04
pkg/steps/lease: fail faster if client not present

bbguimaraes:
```
2020/09/10 14:07:47 Running [input:root], [release-inputs], src, [images], bin, [release:latest], e2e
2020/09/10 14:07:47 Ran for 1s
error: some steps failed:
  * step "e2e" failed validation: step needs a lease but no lease client provided
  * a lease client was required but none was provided, add the --lease-... arguments
```

### DIFF
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -46,6 +46,8 @@ func (s *leaseStep) Inputs() (api.InputDefinition, error) {
 	return s.wrapped.Inputs()
 }
 
+func (s *leaseStep) Validate() error { return nil }
+
 func (s *leaseStep) Name() string             { return s.wrapped.Name() }
 func (s *leaseStep) Description() string      { return s.wrapped.Description() }
 func (s *leaseStep) Requires() []api.StepLink { return s.wrapped.Requires() }
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -20,6 +20,8 @@ import (
 
 const leaseEnv = "LEASED_RESOURCE"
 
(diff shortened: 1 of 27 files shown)


pkg/steps/lease.go
```diff
@@ -20,6 +20,8 @@ import (
 
 const leaseEnv = "LEASED_RESOURCE"
 
+var noLeaseClientErr = errors.New("invalid configuration, step needs a lease but no lease client provided")
```
petr-muller, line 23: can we try to suggest where would the the lease client be provided? In ci-op config with `X` stanza?

stevekuznetsov: This will only affect people who hand-craft a job that uses a template and does not pass the flags we need, right?

bbguimaraes: I had this sitting around in the comment text box yesterday:

The only two possibilities for this error to occur are:
- `prowgen` doesn't generate the job correctly
- someone writes a job manually

The first has never happened, and it would be on us to fix it anyway.

The second happens when a job is missing the `--lease-…` arguments.  It seemed fragile and a layering violation to add that error message here, so yesterday I worked on changing the way this is handled and validated so it can be detected in a better way, faster, and with better diagnostics.  I'll update this PR later today.

bbguimaraes:
Updated.  A `Validate` method was added to the step interface to allow early validation before execution.

From a cursory review, this could also be used to improve some of the existing steps.

stevekuznetsov (approved):

bbguimaraes:

PR #1205, opened 2020-09-04
Fix mountPath vs mount_path in README

johnbieren:
It seems that it is mountPath in volumeMounts: but mount_path in credentials:

Signed-off-by: Johnny Bieren <jbieren@redhat.com>

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -1455,7 +1455,7 @@ const credentialExample = `ref:
   credentials:
   - namespace: test-credentials # this entry injects the custom credential
     name: my-data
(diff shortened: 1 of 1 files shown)


pkg/webreg/webreg.go
```diff
@@ -1455,7 +1455,7 @@ const credentialExample = `ref:
   credentials:
   - namespace: test-credentials # this entry injects the custom credential
     name: my-data
-    mountPath: /var/run/my-data
+    mount_path: /var/run/my-data
```
adduarte, line 1458: Just a vote of support. This is correct. make validate-step-registr throws invalid error on the mountPath field, but accepts mount_path

PR #1207, opened 2020-09-08
ci-operator image: install python2

petr-muller:
Some handcrafted jobs (some of them important upgrade jobs) are doing
`python -c ...` calls. Recently we bumped tools to centos8 which does
not include python by default.

/cc @droslean @deads2k

### DIFF
--- a/images/ci-operator/Dockerfile
+++ b/images/ci-operator/Dockerfile
@@ -1,6 +1,7 @@
 FROM centos:8
 LABEL maintainer="skuznets@redhat.com"
 
-RUN yum install -y git
+RUN yum install -y git python2
+RUN alternatives --set python /usr/bin/python2
 ADD ci-operator /usr/bin/ci-operator
 ENTRYPOINT ["/usr/bin/ci-operator"]


droslean (approved):

images/ci-operator/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM centos:8
 LABEL maintainer="skuznets@redhat.com"
 
-RUN yum install -y git
+RUN yum install -y git python2
+RUN alternatives --set python /usr/bin/python2
```
eparis, line 5: can we try
```
RUN yum install -y git python3
￼RUN alternatives --set python /usr/bin/python3
```

petr-muller: Unfortunately the [jobs are actually Python 2](https://github.com/openshift/release/blob/1a0316177b37855c5dafab592ecead3c7571a1da/ci-operator/jobs/openshift/release/openshift-release-release-4.6-periodics.yaml#L6005-L6008), using `print` statement rather than function:

```
from=$( python -c 'import json,sys; print json.load(open(sys.argv[1], "r"))["name"]' "${initial}" )
```

PR #1209, opened 2020-09-08
pj-rehearse: stop assuming to run on api.ci

petr-muller:
We should just provide correct kubeconfigs for all clusters. Also, change the default prowjob cluster to app.ci.


petr-muller:
/hold

Needs https://github.com/openshift/release/pull/11692 to merge first.

stevekuznetsov (approved):

PR #1210, opened 2020-09-08
Improve failed pod output

stevekuznetsov:
ci-operator: expose message only when there is one

---

ci-operator: spew events for pods that fail to start

In almost all cases, the PodStatus does not contain enough information
and until we have a better way to show off the output of a job, we
should put the event data into the output.

---

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -68,6 +68,7 @@ func FromConfig(
 	var rbacClient rbacclientset.RbacV1Interface
 	var saGetter coreclientset.ServiceAccountsGetter
 	var namespaceClient coreclientset.NamespaceInterface
+	var eventClient coreclientset.EventsGetter
 
 	if clusterConfig != nil {
 		buildGetter, err := buildclientset.NewForConfig(clusterConfig)
@@ -116,6 +117,7 @@ func FromConfig(
 		configMapGetter = coreGetter
 		secretGetter = coreGetter
 		namespaceClient = coreGetter.Namespaces()
+		eventClient = coreGetter
 
 		podClient = steps.NewPodClient(coreGetter, clusterConfig, coreGetter.RESTClient())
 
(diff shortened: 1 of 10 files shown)


pkg/steps/source.go
```diff
@@ -690,7 +693,32 @@ func getReasonsForUnreadyContainers(p *coreapi.Pod) string {
...
-		_, _ = builder.WriteString(fmt.Sprintf("\n* Container %s is not ready with reason %s and message %s", c.Name, reason, message))
+		if message != "" {
+			message = fmt.Sprintf(" and message %s", message)
+		}
+		_, _ = builder.WriteString(fmt.Sprintf("\n* Container %s is not ready with reason %s%s", c.Name, reason, message))
+	}
+	return builder.String()
+}
+
+func getEventsForPod(pod *coreapi.Pod, client coreclientset.EventInterface, ctx context.Context) string {
```
petr-muller, line 704: I don't see how this is related to the `source` step -- it's only called from `template.go` so I think it should be in that file.

pkg/defaults/defaults.go
```diff
@@ -68,6 +68,7 @@ func FromConfig(
 	var rbacClient rbacclientset.RbacV1Interface
 	var saGetter coreclientset.ServiceAccountsGetter
 	var namespaceClient coreclientset.NamespaceInterface
+	var eventClient coreclientset.EventsGetter
```
alvaroaleman, line 71: instead of adding a 13th typed client, can we add the controller-runtime client and try to replace over time the existing typed clients with it? That would greatly simplify the code and testability

stevekuznetsov: Orthogonal to this effort, please.

pkg/steps/source.go
stevekuznetsov: That's where all of our helpers are for this stuff, if we want to clean it up we can do that separately?

pkg/defaults/defaults.go
alvaroaleman: Replacing existing client, yes. Not adding a new typed client vs using the dynamic one, no.

stevekuznetsov: Just another PR please.

alvaroaleman:
> Just another PR please.

I disagree. We are adding something that we shouldn't be adding. We shouldn't be doing a follow-up to replace this client, we should start off by not adding it.

petr-muller:
Personally I think that when we have 12 clients there, adding a 13th there is not that much of an issue (readability is not that much worse). I agree that they should be replaced with what Alvaro is saying, but I would find the intermediate state quite hard to maintain, especially for people seeing that code for the fitrst time. In this case I'd quite happily let it merge with another typed client, and made a card for getting rid of them as a focused effort.

stevekuznetsov:
@petr-muller @alvaroaleman if you can come to a consensus on merging this to make debugging easier, I've rebased it. We now have a card to move to the new clients that can be tackled out of band.

petr-muller (approved):
Let's merge it. It does not make the current situation that much worse, we have a cleanup task.

PR #1211, opened 2020-09-09
Tool to periodically generate testgrids in test-infra

GeoBK:
This tool will generate a PR similar to https://github.com/GeoBK/test-infra/pull/3

### DIFF
--- a/cmd/autotestgridgenerator/main.go
+++ b/cmd/autotestgridgenerator/main.go
@@ -1,0 +1,184 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/test-infra/experiment/autobumper/bumper"
+	"k8s.io/test-infra/prow/config/secret"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/labels"
+)
+
+const (
+	githubOrg    = "kubernetes"
+	githubRepo   = "test-infra"
+	githubLogin  = "openshift-bot"
+	githubTeam   = "openshift/openshift-team-developer-productivity-test-platform"
+	matchTitle   = "Automate testgrid generator"
+	remoteBranch = "auto-testgrid-generator"
+)
+
+type options struct {
+	selfApprove bool
+
+	githubLogin       string
+	gitName           string
+	gitEmail          string
+	targetDir         string
+	assign            string
+	testgridConfigDir string
+	releaseConfigDir  string
+	prowJobsDir       string
+	allowList         string
+
+	flagutil.GitHubOptions
+}
+
+func parseOptions() options {
+	var o options
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
+	fs.StringVar(&o.gitName, "git-name", "", "The name to use on the git commit. Requires --git-email. If not specified, uses the system default.")
+	fs.StringVar(&o.gitEmail, "git-email", "", "The email to use on the git commit. Requires --git-name. If not specified, uses the system default.")
+	fs.StringVar(&o.targetDir, "target-dir", "", "The directory containing the target repo.")
+	fs.StringVar(&o.assign, "assign", githubTeam, "The github username or group name to assign the created pull request to.")
+	fs.StringVar(&o.testgridConfigDir, "testgrid-config", "../../../../kubernetes/test-infra/config/testgrids/openshift", "The directory where the testgrid output will be stored")
+	fs.StringVar(&o.releaseConfigDir, "release-config", "../../../release/core-services/release-controller/_releases", "The directory of release config files")
+	fs.StringVar(&o.prowJobsDir, "prow-jobs-dir", "../../../release/ci-operator/jobs", "The directory where prow-job configs are stored")
+	fs.StringVar(&o.allowList, "allow-list", "../../../release/core-services/testgrid-config-generator/_allow-list.yaml", "File containing release-type information to override the defaults")
+
+	fs.BoolVar(&o.selfApprove, "self-approve", false, "Self-approve the PR by adding the `approved` and `lgtm` labels. Requires write permissions on the repo.")
+	o.AddFlags(fs)
+	o.AllowAnonymous = true
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Errorf("cannot parse args: '%s'", os.Args[1:])
+	}
+	return o
+}
+
+func validateOptions(o options) error {
+	if o.githubLogin == "" {
+		return fmt.Errorf("--github-login cannot be empty string")
+	}
+	if (o.gitEmail == "") != (o.gitName == "") {
+		return fmt.Errorf("--git-name and --git-email must be specified together")
+	}
+	if o.targetDir == "" {
+		return fmt.Errorf("--target-dir is mandatory")
+	}
+	if o.assign == "" {
+		return fmt.Errorf("--assign is mandatory")
+	}
+	return o.GitHubOptions.Validate(false)
+}
+
+func runAndCommitIfNeeded(stdout, stderr io.Writer, author, cmd string, args []string) (bool, error) {
+	fullCommand := fmt.Sprintf("%s %s", filepath.Base(cmd), strings.Join(args, " "))
+
+	logrus.Infof("Running: %s", fullCommand)
+	if err := bumper.Call(stdout, stderr, cmd, args...); err != nil {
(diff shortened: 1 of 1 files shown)


cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+	fs.StringVar(&o.gitName, "git-name", "", "The name to use on the git commit. Requires --git-email. If not specified, uses the system default.")
+	fs.StringVar(&o.gitEmail, "git-email", "", "The email to use on the git commit. Requires --git-name. If not specified, uses the system default.")
+	fs.StringVar(&o.targetDir, "target-dir", "", "The directory containing the target repo.")
+	fs.StringVar(&o.assign, "assign", githubTeam, "The github username or group name to assign the created pull request to.")
+	fs.StringVar(&o.testgridConfigDir, "testgrid-config", "../../../../kubernetes/test-infra/config/testgrids/openshift", "The directory where the testgrid output will be stored")
+	fs.StringVar(&o.releaseConfigDir, "release-config", "../../../release/core-services/release-controller/_releases", "The directory of release config files")
+	fs.StringVar(&o.prowJobsDir, "prow-jobs-dir", "../../../release/ci-operator/jobs", "The directory where prow-job configs are stored")
+	fs.StringVar(&o.allowList, "allow-list", "../../../release/core-services/testgrid-config-generator/_allow-list.yaml", "File containing release-type information to override the defaults")
```
droslean, line 56: Those default values seem very specific. I would prefer not to have them at all or put a full path.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+		logrus.WithError(err).Errorf("cannot parse args: '%s'", os.Args[1:])
+	}
+	return o
+}
+
+func validateOptions(o options) error {
+	if o.githubLogin == "" {
+		return fmt.Errorf("--github-login cannot be empty string")
+	}
+	if (o.gitEmail == "") != (o.gitName == "") {
```
droslean, line 71: Reading this, I realize that we reach a point that we have a lot of `auto*` tools and this code is duplicated everywhere. Maybe it's better to generalize those options and its validation and import it in this tool and update the rest of the tools in a follow-up PR.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+	if err := sa.Start([]string{o.GitHubOptions.TokenPath}); err != nil {
+		logrus.WithError(err).Fatal("Failed to start secrets agent")
+	}
+
+	gc, err := o.GitHubOptions.GitHubClient(sa, false)
+	if err != nil {
+		logrus.WithError(err).Fatal("error getting GitHub client")
+	}
+
+	steps := []struct {
```
droslean, line 135: since there is only one step, there is no reason for this struct to be a slice.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+		command   string
+		arguments []string
+	}{
+		{
+			command: "/usr/bin/testgrid-config-generator",
+			arguments: []string{
+				"-testgrid-config", o.testgridConfigDir, // "../../../../k8s.io/test-infra/config/testgrids/openshift",
```
droslean, line 142: there is no need to add those comments.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/labels"
+)
+
+const (
+	githubOrg    = "kubernetes"
+	githubRepo   = "test-infra"
+	githubLogin  = "openshift-bot"
+	githubTeam   = "openshift/openshift-team-developer-productivity-test-platform"
+	matchTitle   = "Automate testgrid generator"
```
petr-muller, line 25: Heh, this is going to be the PR title, correct? The fact that our current tools say this in their PRs triggers me. The PR title should say what the PR does in the repo, not "Automate $tool". This should say something like "Update OpenShift testgrid definitions" or something meaningful like that.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+	remoteBranch = "auto-testgrid-generator"
+)
+
+type options struct {
+	selfApprove bool
```
petr-muller, line 30: Do we plan to convince someone to give write rights to `k/test-infra`? (I guess that's more a question for @stevekuznetsov). Because if not, then we do not need this code.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+
+	githubLogin       string
+	gitName           string
+	gitEmail          string
```
petr-muller, line 34: The embedded `flagutil.GithubOptions` already has these.

cmd/autotestgridgenerator/main.go
petr-muller: Yeah, I think these should just be required rather than trying to guess.

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+		return fmt.Errorf("--target-dir is mandatory")
+	}
+	if o.assign == "" {
+		return fmt.Errorf("--assign is mandatory")
+	}
+	return o.GitHubOptions.Validate(false)
+}
+
+func runAndCommitIfNeeded(stdout, stderr io.Writer, author, cmd string, args []string) (bool, error) {
+	fullCommand := fmt.Sprintf("%s %s", filepath.Base(cmd), strings.Join(args, " "))
```
petr-muller, line 84: I don't think we should clean up what we actually log with `filepath.Base`. While debugging I'll want to see what the tool *actually* does, including paths etc,

cmd/autotestgridgenerator/main.go
```diff
@@ -0,0 +1,184 @@
...
+}
+
+func main() {
+	o := parseOptions()
+	if err := validateOptions(o); err != nil {
+		logrus.WithError(err).Fatal("Invalid arguments.")
+	}
+	// targetDir needs to be the test-infra directory
+	logrus.Infof("Changing working directory to '%s' ...", o.targetDir)
+	if err := os.Chdir(o.targetDir); err != nil {
```
petr-muller, line 121: IMO you cannot simply change the working dir like this. The users expect to provide inputs (parameters) relative to where they are, and when you `chdir`, the paths they supplied (token path, allow list, everything) are now invalid when they are relative.

cmd/autotestgridgenerator/main.go
petr-muller: ...unless we expect to add more steps in the immediate development. Do we? If no then Nikos is right.

cmd/autotestgridgenerator/main.go
GeoBK: I cant seem to find them, I can see a "Host" field in there. Is that what you meant?

cmd/autotestgridgenerator/main.go
GeoBK: Currently we dont plan on that, but I left it in there in case we wanted to in the future. I'll remove it for now though

cmd/autotestgridgenerator/main.go
GeoBK: I am changing the working directory so that the git operations would work properly ie the git add of the changed files happen in the necessary repository. I can instead add something like --work-tree and --add-dir.
Is that what you meant?

cmd/autotestgridgenerator/main.go
GeoBK: Right now we dont, but I still expect to have the steps stuff after refactor (I'll move the runandcommit out as a common function). Will that work?

cmd/autotestgridgenerator/main.go
petr-muller: Sorry, I misread. My mistake.

cmd/autotestgridgenerator/main.go
petr-muller: I think you can use just `-C`, but yes. I think that's preferable to remembering to make all user input paths absolute first so you can `chdir`.

GeoBK: Ok.. I'll switch to -C. Is this just a convenient way to change the directory or is there some additional benefit as well?

cmd/autotestgridgenerator/main.go
stevekuznetsov: That was my point at the very start about having these as separate tools that call `os.Command` but you and @alvaroaleman disagreed :trollface:

stevekuznetsov (approved):
Love it!

cmd/autotestgridgenerator/main.go
alvaroaleman: This exists now in `pkg/github/prcreation`

PR #1212, opened 2020-09-10
ci-operator: validate operator.substitute.with better

petr-muller:
`operator.substitute.with` must be resolvable to a image involved in the
ci-operation execution, otherwise we start using returning `nil`s when
computing step dependencies, leading to `nil` dereference.

Addressed in two places: on validation, and on usage (defensively -- log a
warning. this should not happen with the validation but this way it's more
resilient to changes).

Also took the opportunity and fixed few code style issues.

Fixes: DPTP-1484


petr-muller:
ping @hongkailiu @stevekuznetsov @AlexNPavel

stevekuznetsov (approved):

PR #1216, opened 2020-09-11
Delete attachment from local bw mirror after running bw cmd

GeoBK:
Since the bw delete command was being called after the item was removed from the list, we were actually deleting the element which occurs after intended element - since the pointer points to the item at that index.
This change would move the delete from the list after the call to bw delete


GeoBK:
cc @stevekuznetsov

stevekuznetsov (approved):

PR #1221, opened 2020-09-14
Testgrid: Remove support for release-type annotation

GeoBK:


cmd/testgrid-config-generator/main.go
```diff
@@ -258,11 +258,7 @@ func main() {
...
 		calculateDays := len(p.Cron) > 0 || len(p.Interval) > 0
 		var dashboardType string
-		var label string
-		var exists bool
-		if label, exists = allowList[name]; !exists {
-			label = p.Labels["ci.openshift.io/release-type"]
-		}
+		label := allowList[name]
 		switch label {
 		case "informing", "blocking", "broken", "generic-informing":
```
stevekuznetsov, line 263: Don't we validate that entries in the allow-list are one of these already?

Can we switch up this logic to be a little more clear:

```go
var dasboardType string
allowListLabel, inAllowList := allowList[name]
configLabel, inConfig := configuredJobs[name]

if !inAllowList && !inConfig {
    continue
} else if inAllowList && inConfig && allowListLabel == "informing" && configLabel == "blocking" {
    dasboardType = configLabel
    calculateDays = false
} else if inConfig {
    dasboardType = configLabel
    calculateDays = false
} else if inAllowList {
    dasboardType = allowListLabel
}
```

IDK ... does that read better to you? It's weird to me that the most common and useful case - when it's in the config - is a default case and not explicitly listed labeled as such, and comes *after* a bunch of this other logic?

GeoBK: > Don't we validate that entries in the allow-list are one of these already?
No we don't. We are just checking if that value is not blocking. I can add a check to ensure that the value is in "informing", "blocking", "generic-informing". I haven't seen "generic-informing" in the allow-list yet though, is that a valid value for the allow_list?

stevekuznetsov (approved):
Better to just leave it ...

PR #1226, opened 2020-09-15
Allow rpm server to run in python2 or 3

jupierce:
RHEL7 base images contain only python2 while RHEL8 contains only python3.


stevekuznetsov (approved):

petr-muller:
This is beautiful :heart_eyes:

PR #1229, opened 2020-09-16
Add a test image with boskos

bbguimaraes:


petr-muller (approved):

alvaroaleman:
Out of curiosity, why not just use the `gcr.io/k8s-staging-boskos/boskos` image?

bbguimaraes:
> Out of curiosity, why not just use the `gcr.io/k8s-staging-boskos/boskos` image?

See https://github.com/openshift/release/pull/11878 for the purpose.

stevekuznetsov:
Part of DPTP-1365

PR #1232, opened 2020-09-16
e2e: add an initial version of lease tests

bbguimaraes:
Provisional version for reference.

### DIFF
--- a/test/e2e/lease/config.yaml
+++ b/test/e2e/lease/config.yaml
@@ -1,0 +1,43 @@
+base_images:
+  os:
+    name: centos
+    namespace: openshift
+    tag: '7'
+releases:
+  latest:
+    release:
+      channel: stable
+      version: "4.4"
+resources:
+  '*':
+    limits:
+      cpu: 500m
+    requests:
+      cpu: 10m
+tests:
+- as: success
+  literal_steps:
+    cluster_profile: aws
+    test:
+    - as: test
+      commands: exit 0
+      from: os
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+- as: invalid-lease
+  literal_steps:
+    cluster_profile: gcp
+    test:
+    - as: test
+      commands: exit 0
+      from: os
+      resources:
+        requests:
+          cpu: 100m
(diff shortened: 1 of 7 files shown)


petr-muller (approved):

test/e2e/lease/config.yaml
```diff
@@ -0,0 +1,43 @@
...
+      resources:
+        requests:
+          cpu: 100m
+          memory: 200Mi
+- as: invalid-lease
+  literal_steps:
+    cluster_profile: gcp
+    test:
+    - as: test
+      commands: exit 0
```
stevekuznetsov, line 34: Can you do something with `$LEASED_RESOURCE` as well?

```sh
test -n "${LEASED_RESOURCE}"
```

stevekuznetsov:
/lgtm cancel
This seems to be perma-failing:

```
   * could not run steps: step success failed: failed to acquire lease: status 403 Forbidden, status code 403 
```

test/e2e/lease/config.yaml
bbguimaraes: Fun fact: this variable is fundamentally broken.

I though your suggestion was entirely reasonable and implemented it…  Only to go on a bug-hunting session and discover that `LEASED_RESOURCE` does not work if there is more than one test that requires a lease in the configuration.

We create steps for each test and add the values from their `Provides` method before pruning the graph (the same issue that caused the need for https://github.com/openshift/ci-tools/pull/1204#issuecomment-690323902), so each step overwrites the last and `LEASED_RESOURCE` ends up being supplied by the last in evaluation order.

This magically works today because:
- no multi-stage test uses the value yet
- template steps are extraneous and are added after the steps from the configuration are processed

Given all this and the fact that we are in the process of deprecating templates and adding configurable leases, I propose we leave the test as is.   Future E2E tests for the configurable leases should verify the value of the variable.

bbguimaraes:
Given https://github.com/openshift/ci-tools/pull/1232#discussion_r492717389, I believe this is ready to be reviewed/merged.

stevekuznetsov:
Part of DPTP-1365

PR #1233, opened 2020-09-16
prow-job-dispatcher: save and load job volumes

hongkailiu:
https://issues.redhat.com/browse/DPTP-1520

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -33,6 +33,7 @@ type options struct {
 	prometheusUsername     string
 	prometheusPasswordPath string
 	jobVolumesPath         string
+	refreshJobVolumes      bool
 	maxConcurrency         int
 
 	prometheusPassword string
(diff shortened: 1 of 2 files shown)


alvaroaleman:
@hongkailiu https://issues.redhat.com/browse/DPTP-1520 is titled "Make prow-jop-dispatcher Idempotent", what does that have to do with saving and loading?

hongkailiu:
> @hongkailiu https://issues.redhat.com/browse/DPTP-1520 is titled "Make prow-jop-dispatcher Idempotent", what does that have to do with saving and loading?

It is in the acceptance criterion.
"given an input set of values from Prometheus, ..."
If we query prometheus each time when run the tool, it is highly likely that we got different answers from Prometheus. And then we do dispatch based on the new answers. Dispatching results would be different.

alvaroaleman:
@hongkailiu yes, but there is no "Make sure the input values don't change". There is just "Make sure the output values don't change assuming the same inputs".

alvaroaleman:
@hongkailiu if we end up needing to make sure the prometheus data doesn't change, we can just hardcode an absolute time window to query (i.E. sep 1 - sep 7). But that is an entirely separate discussion from the issue described in the card.

hongkailiu:
/hold

I will come back to this later (after the manual migration).

hongkailiu:
/close
will do anther PR for time frame.

PR #1236, opened 2020-09-16
Registry replacer: Add support to replace copy --from

alvaroaleman:


cmd/registry-replacer/main.go
```diff
@@ -282,7 +282,7 @@ func (ort orgRepoTag) String() string {
 func ensureReplacement(image *api.ProjectDirectoryImageBuildStepConfiguration, dockerfile []byte) ([]orgRepoTag, error) {
 	var toReplace []string
 	for _, line := range bytes.Split(dockerfile, []byte("\n")) {
-		if !bytes.Contains(line, []byte("FROM")) {
+		if !bytes.Contains(line, []byte("FROM")) && !bytes.Contains(line, []byte("COPY")) && !bytes.Contains(line, []byte("copy")) {
```
petr-muller, line 285: /hold

Curious, is there a good reason for these checks being `Contains` and not `HasPrefix`?
Holding for answer :)

petr-muller (approved):

cmd/registry-replacer/main.go
alvaroaleman: Its valid to have leading spaces

PR #1237, opened 2020-09-17
Autotestgridgen: Create fork if not present

GeoBK:


GeoBK:
cc @stevekuznetsov

cmd/autotestgridgenerator/main.go
```diff
@@ -98,9 +99,26 @@ func main() {
 		logrus.Info("no new commits, exiting ...")
 		os.Exit(0)
 	}
+	newRepo := githubRepo
+	gc.SetMax404Retries(0)
+	if _, err := gc.GetRepo(o.githubLogin, githubRepo); err != nil {
+		// Somehow github.IsNotFound doesn't recognize this?
```
petr-muller, line 105: This looks suspicious, have you looked at why it doesn't? We should fix that instead of relying on `strings.Contains` here.

cmd/autotestgridgenerator/main.go
```diff
@@ -98,9 +99,26 @@ func main() {
...
+		if !strings.Contains(err.Error(), "status code 404") {
+			logrus.WithError(err).Fatalf("unexpected error when getting repo %s/%s", o.GitName, githubRepo)
+		}
+		logrus.Info("Creating fork")
```
petr-muller, line 109: nit: logging *why* (it does not exist) and logging more (creating fork of x/y) coyuld be helpful

cmd/autotestgridgenerator/main.go
```diff
@@ -98,9 +99,26 @@ func main() {
 		logrus.Info("no new commits, exiting ...")
 		os.Exit(0)
 	}
+	newRepo := githubRepo
```
droslean, line 102: This is not a `newRepo`. I can understand that this value can change in the below code and become a new repo but now it is not.

pkg/github/prcreation/prcreation.go
```diff
@@ -84,7 +84,7 @@ func (o *PRCreationOptions) UpsertPR(localSourceDir, org, repo, branch, prTitle,
 			return fmt.Errorf("failed to fork %s/%s: %w", org, repo, err)
 		}
 		repo = forkName
-		if err := waitForRepo(username, repo, o.GithubClient); err != nil {
+		if err := WaitForRepo(username, repo, o.GithubClient); err != nil {
```
droslean, line 87: why this is now exposed?

GeoBK: I am trying to resue that function (here)[https://github.com/openshift/ci-tools/pull/1237/files#diff-4440cf6f38e6026b7a8028aa83c9f664R115]

cmd/autotestgridgenerator/main.go
GeoBK: I'll change the name to forkedRepoName. Please correct me if thats not what you meant

cmd/autotestgridgenerator/main.go
```diff
@@ -98,9 +99,26 @@ func main() {
...
+	gc.SetMax404Retries(0)
```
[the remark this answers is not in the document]
GeoBK: i think its 2. [link](https://github.com/kubernetes/test-infra/blob/master/prow/github/client.go#L498)

cmd/autotestgridgenerator/main.go
```diff
@@ -98,9 +99,26 @@ func main() {
 		logrus.Info("no new commits, exiting ...")
 		os.Exit(0)
 	}
```
alvaroaleman, line 101: Whats the reason to copy all of this code instead of just using the `UpsertPR` functionality from the prcreation package?

GeoBK: Thats mostly because most of the functionality was already checked in before I was aware of this module and adding the fork creation logic in here seemed like the faster approach, but you are right I should reuse the UpsertPR function.

stevekuznetsov (commented):
Let's see a revision using UpsertPR

cmd/autotestgridgenerator/main.go
```diff
@@ -67,45 +68,33 @@ func main() {
...
 
-	if !committed {
-		logrus.Info("no new commits, exiting ...")
-		os.Exit(0)
-	}
+	fullCommand := fmt.Sprintf("%s %s", command, strings.Join(arguments, " "))
+	logrus.Infof("Running: %s", fullCommand)
+	logrus.WithField("cmd", command).
+		// The default formatting uses a space as separator, which is hard to read if an arg contains a space
+		WithField("args", fmt.Sprintf("['%s']", strings.Join(arguments, "', '"))).
```
stevekuznetsov, line 87: Ideally we'd just not have spaces in args - which one does? Or just add the list as the field, the default print will be fine

pkg/github/prcreation/prcreation.go
```diff
@@ -71,7 +71,7 @@ func (o *PRCreationOptions) UpsertPR(localSourceDir, org, repo, branch, prTitle,
 	stdout := bumper.HideSecretsWriter{Delegate: os.Stdout, Censor: o.secretAgent}
 	stderr := bumper.HideSecretsWriter{Delegate: os.Stderr, Censor: o.secretAgent}
 
-	sourceBranchName := strings.ReplaceAll(strings.ToLower(prTitle), " ", "-")
+	sourceBranchName := strings.ReplaceAll(strings.ToLower(remoteBranch), " ", "-")
```
stevekuznetsov, line 74: Why isn't this a breaking change?

GeoBK: This is a breaking change but UpsertPr is used in only one other place where I am passing in the prTitle as the remoteBranch.

pkg/github/prcreation/prcreation.go
```diff
@@ -121,7 +121,7 @@ func (o *PRCreationOptions) UpsertPR(localSourceDir, org, repo, branch, prTitle,
 		repo,
 		prTitle,
 		prBody,
-		prTitle,
+		remoteBranch,
```
alvaroaleman, line 124: What is this `remoteBranch` and why is it used as value for `matchTitle`?

cmd/ocp-build-data-enforcer/main.go
```diff
@@ -309,6 +309,7 @@ func (dp *diffProcessor) process() error {
 					"[1]: https://github.com/openshift/ci-tools/tree/master/cmd/ocp-build-data-enforcer",
 					"[2]: https://github.com/openshift/ocp-build-data/tree/openshift-4.6/images",
 				}, "\n"),
+				fmt.Sprintf("Updating %s baseimages to mach ocp-build-data config", d.path),
```
alvaroaleman, line 312: This doesn't sound like a valid branch name to me

cmd/autotestgridgenerator/main.go
GeoBK: Ok, I'll just add it as a list

cmd/ocp-build-data-enforcer/main.go
GeoBK: In UpsertPr we used to use the prTitle as the [sourceBranchName](https://github.com/openshift/ci-tools/pull/1237/files#diff-f7d366a8393f769e69b516187d845078L74), so I don't believe I have changed how this tool would behave with this change

pkg/github/prcreation/prcreation.go
GeoBK: I added the remoteBranch as I couldn't reuse the prTitle as a branch which was being done previously.(I think it was because the prTitle was too long)
In my case the remoteBranch is the prefix for my prTitle and would match the title without the timestamp. But I see how this would be incorrect and I'll change the name from remoteBranch to matchTitle

alvaroaleman: okay, would you then mind to
* Rename the variable to something that reflects its purpose (`matchTitle`)
* Use the [functional opts pattern](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md) to make this an optional option

stevekuznetsov (approved):

PR #1245, opened 2020-09-21
TestgrigGen: Add option to validate the allow-list

GeoBK:


cmd/testgrid-config-generator/main.go
```diff
@@ -188,6 +194,23 @@ var reVersion = regexp.MustCompile(`-(\d+\.\d+)(-|$)`)
 // upgraded from informing to blocking if the job is set to informing.
 func main() {
 	o := gatherOptions()
+
+	// read the list of jobs from the allow list along with its release-type
+	var allowList map[string]string
+	if o.jobsAllowListFile != "" {
```
droslean, line 200: lets put all this in the `Validate` method and rename it to `DefaultAndValidate`. Also we can have the `allowList` var under the options struct as well.

stevekuznetsov: Why? The option is the flag value, not the list itself

stevekuznetsov (approved):

PR #1238, opened 2020-09-17
Revert "validateLiteralTestStepCommon: allow using imagestreamtag references"

petr-muller:
Reverts openshift/ci-tools#1230

Just removing the validation does not seem to be sufficient, IST references need to be processed differently later. Right now ci-operator spawns invalid pods, and then image-registry operator chokes on them:

```
The following objects have invalid references:

  Pod[ci-op-2cq11w99/e2e-azure-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-5hr1jn9h/e2e-gcp-image-ecosystem-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-7gyk1rg4/e2e-ovn-hybrid-step-registry-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-9rkqc2x8/e2e-operator-with-custom-vxlan-port-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-b6ddjmcv/e2e-ovn-step-registry-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-sv6g320c/e2e-windows-hybrid-network-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-sxltbxjz/e2e-aws-sdn-multi-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
  Pod[ci-op-ypvwxw58/e2e-aws-sdn-single-ipi-install-install]: invalid container image reference "stable:stable-initial:installer": invalid reference format
```

FYI @vrutkovs

PR #1246, opened 2020-09-22
multistage: handle imagestream references in From

vrutkovs:
Resubmit of https://github.com/openshift/ci-tools/pull/1230 (reverted by https://github.com/openshift/ci-tools/pull/1238).

If multistage "from" contains imagestreamtag it would be used in the pod. This is useful for upgrade jobs, which require installer to be used from "stable-initial:installer" imagstreamtag.

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -353,11 +353,9 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 		if link, ok := step.FromImageTag(); ok {
 			image = fmt.Sprintf("%s:%s", api.PipelineImageStream, link)
 		} else {
-			parts := strings.Split(image, ":")
-			if len(parts) == 1 {
-				stream, _ := s.config.ImageStreamFor(image)
-				image = fmt.Sprintf("%s:%s", stream, image)
-			}
+			dep := api.StepDependency{Name: image}
+			stream, tag, _ := s.config.DependencyParts(dep)
+			image = fmt.Sprintf("%s:%s", stream, tag)
 		}
 		resources, err := resourcesFor(step.Resources)
 		if err != nil {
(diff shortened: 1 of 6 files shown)


bbguimaraes:
The reason the validation existed in the first place was that the origin of an image could be determined from the configuration — any missing name was considered to be imported into `stable`, which we cannot validate before importing the tags.  The origin is required not only to determine the value for the Pod's `image`, but also for dependency resolution.

Since the customization is only required when multiple releases are imported, can we limit it to `stable-<name>:<image>`?  That way we keep the ability to correctly identify the source and also support multiple releases.

@stevekuznetsov

/hold

vrutkovs:
>Since the customization is only required when multiple releases are imported, can we limit it to stable-<name>:<image>? That way we keep the ability to correctly identify the source and also support multiple releases.

Added validation of allowed imagestreams: `pipeline`/`stable`/`stable-initial`/`release`

pkg/steps/multi_stage.go
```diff
@@ -353,8 +353,11 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 		if link, ok := step.FromImageTag(); ok {
 			image = fmt.Sprintf("%s:%s", api.PipelineImageStream, link)
 		} else {
-			stream, _ := s.config.ImageStreamFor(image)
-			image = fmt.Sprintf("%s:%s", stream, image)
+			parts := strings.Split(image, ":")
```
stevekuznetsov, line 356: You want `DependencyParts` here - see other usage in the repo

pkg/steps/multi_stage.go
```diff
@@ -353,8 +353,10 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 		if link, ok := step.FromImageTag(); ok {
 			image = fmt.Sprintf("%s:%s", api.PipelineImageStream, link)
 		} else {
-			stream, _ := s.config.ImageStreamFor(image)
-			image = fmt.Sprintf("%s:%s", stream, image)
+			dep := api.StepDependency{Name: image}
```
stevekuznetsov, line 356: Sorry, implicit in that statement was that you also need to expose those dependencies to the graph builder - you'll need to do this resolution and use `LinkForImage()`, expose using the step's `Requires()` method

stevekuznetsov (approved):

PR #1247, opened 2020-09-22
docs: more details about built oo indices

petr-muller:
Be more informational about how the indices are built.

~~WIP: I'd like to add a link to the docs that describe `opm index add` modes but I need to find one.~~ turns out there is no good doc page (asked on #forum-operator-fw) so I tried to elaborate.

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -471,6 +471,15 @@ image. The index image is named <code>ci-index</code> and can be exposed to test
 steps via the <a href="/help/ci-operator#literal-references"><code>dependencies</code></a> feature.
 </p>
 
+<p>
+The ephemeral index is built from scratch and only the bundles built in the
+current <code>ci-operator</code> run will be added to it, nothing else. The
+bundles are added to the index using the <code>semver</code> mode, which means
+that the <code>spec.version</code> stanza in the CSV must be a valid semantic
+version. Also, if the CSV has a <code>spec.replaces</code> stanza, it is
+ignored, because the index will not contain a bundle with the replaced version.
+</p>
+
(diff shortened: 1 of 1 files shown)


pkg/webreg/webreg.go
```diff
@@ -471,6 +471,15 @@ image. The index image is named <code>ci-index</code> and can be exposed to test
 steps via the <a href="/help/ci-operator#literal-references"><code>dependencies</code></a> feature.
 </p>
 
+<p>
+Thi ephemeral index is built from scratch and only the bundles built in the
```
stevekuznetsov, line 475: `s/Thi/This/`

pkg/webreg/webreg.go
```diff
@@ -471,6 +471,15 @@ image. The index image is named <code>ci-index</code> and can be exposed to test
...
+current <code>ci-operator</code> run will be added to it, nothing else. The
+bundles are added to the index using the <code>semver</code> mode, which means
+that the CSV version needs to be a valid semver. Also, if the CSV has a <code>spec.replaces</code>
```
stevekuznetsov, line 478: nit: CSV stands for Cluster Service Version so "CSV Version" is like PIN Number. Maybe specifically call out `csv.spec.version`?

stevekuznetsov (changes requested):

pkg/webreg/webreg.go
```diff
@@ -471,6 +471,15 @@ image. The index image is named <code>ci-index</code> and can be exposed to test
...
+stanza, it is ignored, because the index will not contain a bundle with the
+replaced version.
```
mgencur, line 480: @petr-muller thanks for adding these details! I'll need to try it soon:)

petr-muller:
Thanks, addressed!

PR #1250, opened 2020-09-22
empty agent jobs are running on k8s

hongkailiu:
According to https://github.com/kubernetes/test-infra/blob/6d83fa84beb472a6b3e916a89f371e251b321e84/prow/config/jobs.go#L93-L94

The value of `agent` is "kubernetes".

### DIFF
--- a/pkg/dispatcher/testdata/TestLoadConfig/good_config_with_build_farm_with_jobs.yaml
+++ b/pkg/dispatcher/testdata/TestLoadConfig/good_config_with_build_farm_with_jobs.yaml
@@ -1,5 +1,4 @@
 default: api.ci
-nonKubernetes: app.ci
 groups:
   "api.ci":
     paths:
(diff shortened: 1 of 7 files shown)


pkg/dispatcher/testdata/TestLoadConfig/good_config_with_build_farm_with_jobs.yaml
```diff
@@ -1,5 +1,5 @@
 default: api.ci
-nonKubernetes: app.ci
+nonKubernetes: ""
```
alvaroaleman, line 2: I think you can just remove this field, it doesn't serve any purpose

hongkailiu: The purpose is that we have this field (not its value) `nonKubernetes` defined/implemented?

alvaroaleman: This field must be somewhere in the go type, but is completely irrelevant what cluster non-Kubernetes type jobs are on (could probably be hardcoded to the empty string)

hongkailiu: Removed.

alvaroaleman (approved):

PR #1251, opened 2020-09-22
Allow specifying build_root_image in the repository

alvaroaleman:
This PR allows to specify the `build_root_image` inside the repository by setting `build_root.from_repository: true`. If this is set, we will:
* Enable cloning on all prowjobs that are based off of that ci-operator config (We might optimize this later on to only do so when we actually have that image in our graph, but that requires some bigger refacotoring of the ci-operator and prowgen)
* Read the imagestream from a file named `.build_root_image` in the repo

/assign @stevekuznetsov 
/cc @marun 

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -273,7 +273,7 @@ func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependen
 func validateBuildRootImageConfiguration(fieldRoot string, input *BuildRootImageConfiguration, hasImages bool) error {
 	if input == nil {
 		if hasImages {
-			return errors.New("when 'images' are specified 'build_root' is required and must have image_stream_tag or project_image")
+			return errors.New("when 'images' are specified 'build_root' is required and must have image_stream_tag, project_image or from_repository set")
 		}
 		return nil
 	}
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -68,7 +68,7 @@ func (config *ReleaseBuildConfiguration) validate(org, repo string, resolved boo
 	var validationErrors []error
 
 	validationErrors = append(validationErrors, validateReleaseBuildConfiguration(config, org, repo)...)
-	validationErrors = append(validationErrors, validateBuildRootImageConfiguration("build_root", config.InputConfiguration.BuildRootImage, len(config.Images) > 0)...)
+	validationErrors = append(validationErrors, validateBuildRootImageConfiguration("build_root", config.InputConfiguration.BuildRootImage, len(config.Images) > 0))
 	validationErrors = append(validationErrors, validateTestStepConfiguration("tests", config.Tests, config.ReleaseTagConfiguration, resolved)...)
 
 	// this validation brings together a large amount of data from separate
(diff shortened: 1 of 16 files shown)


pkg/api/config.go
```diff
@@ -270,21 +270,27 @@ func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependen
 	}
 }
 
-func validateBuildRootImageConfiguration(fieldRoot string, input *BuildRootImageConfiguration, hasImages bool) []error {
+func validateBuildRootImageConfiguration(fieldRoot string, input *BuildRootImageConfiguration, hasImages bool) error {
 	if input == nil {
 		if hasImages {
-			return []error{errors.New("when 'images' are specified 'build_root' is required and must have image_stream_tag or project_image")}
+			return errors.New("when 'images' are specified 'build_root' is required and must have image_stream_tag or project_image")
```
stevekuznetsov, line 276: this is missing your new option

pkg/api/types.go
```diff
@@ -305,13 +305,19 @@ const (
 	ReleaseChannelStable    ReleaseChannel = "stable"
 	ReleaseChannelFast      ReleaseChannel = "fast"
 	ReleaseChannelCandidate ReleaseChannel = "candidate"
+
+	// BuildRootImageFileName is the name of the file that contains the build root images
+	// pullspec.
+	BuildRootImageFileName = ".build_root_image"
```
stevekuznetsov, line 311: Could we go with a `.ci-operator.yaml` or something to future-proof this?

pkg/defaults/defaults.go
```diff
@@ -591,3 +605,27 @@ func defaultImageFromReleaseTag(base api.ImageStreamTagReference, release *api.R
 	base.Namespace = release.Namespace
 	return base
 }
+
+func buildRootImageStreamFromRepository(readFile readFile) (*api.ImageStreamTagReference, error) {
```
stevekuznetsov, line 609: Why can't we expect structured data here in normal name/namespace/tag format like everywhere else in config?

pkg/prowgen/prowgen.go
```diff
@@ -125,6 +126,11 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	if release, found := configSpec.Releases[cioperatorapi.LatestReleaseName]; found && release.Candidate != nil {
 		jobRelease = release.Candidate.Version
 	}
+
+	skipCloning := true
+	if configSpec.BuildRootImage != nil && configSpec.BuildRootImage.FromRepository {
```
stevekuznetsov, line 131: Is it worth it to try to make this shallow, too, to save on time?

stevekuznetsov (commented):
Minor nits

pkg/api/types.go
alvaroaleman: Sure, so make it a struct (which for now will only have this one option)?

petr-muller (commented):
LGTM modulo what Steve said.

pkg/prowgen/prowgen.go
alvaroaleman: I don't see an option to enable shallow cloning in the decoration utils

pkg/defaults/defaults.go
alvaroaleman: done

pkg/api/types.go
alvaroaleman: done

sttts:
/hold

Why do we need another source other than Dockerfile?

sttts:
And before unhold, this MUST go into an enhancement. Document what CI does.

tnozicka:
I support the `.build_root_image` file. Dockefile FROM and build _root images are different.

This should also update the doc here https://github.com/openshift/ci-tools/blob/master/CONFIGURATION.md#build_root

sttts:
Questions:
- How many non-standard build_root_images are used in which repos?
- What is the standard build_root_image? Where is its source code?
- How is the standard build_root_image derived from the Dockerfile FROM image? Are they linked? They must IMO.
- Which things-to-version are in the build images? protoc? Other generators?
- How is the standard build_root_image linked to the Kube version used by the control plane?

alvaroaleman:
This PR doesn't change anything about what build_root you can or can not use (I believe there are no restrictions on this), just where to configure it, in order to allow for an atomic change that changes both code and image.

The build_root is used to as the image to clone the code into. The result of that in turn is used as the image to execute the `binary_build_commands`, `test_binary_build_commands` and `rpm_build_commands`.

The standard images used for the build_root are the `openshift/release:golang-$VERSION` ones, their dockerfiles are here: https://github.com/openshift/release/tree/b2770007d21ea33f7ac1ea2ad780904e127448b9/projects/origin-release

stevekuznetsov (approved):

stevekuznetsov:
/hold

Let's get the documentation page updated with this feature and I think the implementation here is merge-able.

alvaroaleman:
Docs are at https://server-ci-tools-1251.apps.ci.l2s4.p1.openshiftapps.com/help/ci-operator

stevekuznetsov:
/lgtm
/hold cancel

LGTM!

sttts:
This PR is definitely the right direction. I have not seen an answer for how we guarantee that we won't test using different base image than what is used in the Dockerfile. We will probably build something into build-machinery as CI is not willing to build or talk about a machanism that guarantees that from the start.

PR #1253, opened 2020-09-22
Make prBody and MatchTitle optional parameters

GeoBK:


GeoBK:
@alvaroaleman  I broke out the changes for UpsertPr and have added the optional parameters here. I am not sure if you meant that I should make all parameters optional, but for now I have only made prBody, prTitle and matchTitle optional

pkg/github/prcreation/prcreation.go
```diff
@@ -44,9 +44,47 @@ func (o *PRCreationOptions) Finalize() error {
...
+	prTitle    string
+	prBody     string
+	matchTitle string
+}
+
+// PrOption is the type for Optional Parameters
+type PrOption func(*PrOptions)
+
+// PrTitle is the wrapper to pass in prTitle as a parameter
+func PrTitle(prTitle string) PrOption {
```
alvaroaleman, line 58: I think this is mandatory, a PR must have a title

alvaroaleman (approved):
Thanks!

PR #1255, opened 2020-09-23
Verify duplications in the keys from secret-bootstrap's config

hongkailiu:
Follow up https://github.com/openshift/release/pull/12043#issuecomment-696971732

### DIFF
--- a/pkg/api/secretbootstrap/secretboostrap_test.go
+++ b/pkg/api/secretbootstrap/secretboostrap_test.go
@@ -1,6 +1,8 @@
 package secretbootstrap
 
 import (
+	"fmt"
+	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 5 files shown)


alvaroaleman (approved):
A more generic solution for this issue would be to use https://github.com/adrienverge/yamllint which can detect this and other issues

alvaroaleman:
/lgtm cancel
as lint doesn't pass

pkg/api/secretbootstrap/secretboostrap_test.go
```diff
@@ -86,3 +88,69 @@ func TestResolving(t *testing.T) {
...
+			actualError := LoadConfigFromFile(filepath.Join("testdata", fmt.Sprintf("%s.yaml", t.Name())), &actual)
+			equalError(t, tc.expectedError, actualError)
+			if diff := cmp.Diff(tc.expected, actual); diff != "" {
+				t.Errorf("expected config differs from actual config: %s", diff)
+			}
+		})
+	}
+}
+
+func equalError(t *testing.T, expected, actual error) {
```
stevekuznetsov, line 149: Can we stop writing a version of this function in every test file? Just inline the thing to the test method

hongkailiu: Inlined.

hongkailiu:
> A more generic solution for this issue would be to use https://github.com/adrienverge/yamllint which can detect this and other issues

Just found this upstream.
Maybe we can do something similar?

PR #1258, opened 2020-09-24
e2e: add a template test with a lease

bbguimaraes:


test/e2e/lease/config.yaml
```diff
@@ -3,17 +3,15 @@ base_images:
     name: centos
     namespace: openshift
     tag: '7'
-releases:
```
stevekuznetsov, line 6: Ideally we'd have a case with both?

bbguimaraes:
Updated to use https://github.com/openshift/ci-tools/pull/1267.

stevekuznetsov:
/test e2e

```
   * could not initialize namespace: could not create role binding for: rolebindings.rbac.authorization.k8s.io is forbidden: User "system:serviceaccount:ci:ci-operator" cannot create resource "rolebindings" in API group "rbac.authorization.k8s.io" in the namespace "ci-op-426jmksd" 
```

stevekuznetsov:
Part of DPTP-1365

PR #1260, opened 2020-09-28
Reporter: Add error message for missing report-password file

GeoBK:


pkg/results/report.go
```diff
@@ -53,6 +53,9 @@ func (o *Options) Reporter(spec *api.JobSpec, consoleHost string) (Reporter, err
 	if o.address == "" {
 		return &noopReporter{}, nil
 	}
+	if o.password == "" {
+		return nil, fmt.Errorf("missing 'report-password-file'")
```
alvaroaleman, line 57: lets make it clearer that this is a flag: `"mandatory flag -report-password-file is unset"`.

Also, please don't use `fmt.Errorf` for errors without a format directive, use `errors.New` from the stdlibs `error` package instead

alvaroaleman (approved):

PR #1261, opened 2020-09-29
Use kubeconfig to promote images

hongkailiu:
https://issues.redhat.com/browse/DPTP-1564

This will allow us to promote images to a registry running on any cluster.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -298,6 +298,8 @@ type options struct {
 	cloneAuthConfig *steps.CloneAuthConfig
 
 	resultsOptions results.Options
+
+	promoteKubeConfigPath string
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
@@ -359,6 +361,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.StringVar(&opt.variant, "variant", "", "Variant of the project's ci-operator config (used by configresolver)")
 
 	flag.String("kubeconfig", "", "Legecay flag kept for compatibility reasons. Doesn't do anything.")
+	flag.StringVar(&opt.promoteKubeConfigPath, "promote-kubeconfig-path", "", "The Path to a kubeconfig file to promote images. InClusterConfig will be used if not set.")
 
 	flag.StringVar(&opt.pullSecretPath, "image-import-pull-secret", "", "A set of dockercfg credentials used to import images for the tag_specification.")
 
@@ -521,6 +524,7 @@ func (o *options) Complete() error {
 			return fmt.Errorf("could not get pull secret from path %s: %w", o.pullSecretPath, err)
 		}
 	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


cmd/ci-operator/main.go
```diff
@@ -359,6 +361,9 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.StringVar(&opt.variant, "variant", "", "Variant of the project's ci-operator config (used by configresolver)")
 
 	flag.String("kubeconfig", "", "Legecay flag kept for compatibility reasons. Doesn't do anything.")
+	// SA ci-operator on app.ci will be used to generate the kubeconfig file
```
alvaroaleman, line 364: For now it will be on api.ci or not?

pkg/defaults/defaults.go
```diff
@@ -307,7 +309,26 @@ func FromConfig(
...
 			return nil, nil, fmt.Errorf("could not determine promotion defaults: %w", err)
 		}
-		postSteps = append(postSteps, release.PromotionStep(*cfg, config.Images, requiredNames, imageClient, imageClient, jobSpec))
+
+		dstImageClient := imageClient
+
+		if promoteKubeConfigPath != "" {
+			kubeConfigs, _, err := util.LoadKubeConfigs(promoteKubeConfigPath)
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to load kubeconfigs: %w", err)
```
alvaroaleman, line 318: Please clarify which kubeconfig this was and from where we attempted to load it

pkg/defaults/defaults.go
```diff
@@ -307,7 +309,26 @@ func FromConfig(
...
+			}
+			// Will be changed to app.ci after migration
+			promoteClusterContext := "api.ci"
```
alvaroaleman, line 321: I wouldn't expect there to be more than one context, we can just swap the kubeconfig when we want to change this

cmd/ci-operator/main.go
hongkailiu: For now, this flag is not be used.
This PR wont have any impact for any promotion job.

My next step:
Move one promotion job to build01 (by providing such kubeconfig) to see if it works (push image from build01 to app.ci). 
If it goes through, promotion jobs are ready for migration.

(We still need to mirror the whole existing images from api.ci to app.ci. But that is not in the scope of the current PR.)

pkg/defaults/defaults.go
hongkailiu: Updated.

pkg/defaults/defaults.go
hongkailiu: Done.

cmd/ci-operator/main.go
alvaroaleman: My point is, that comment is wrong. Initially, this will be a SA on api.ci, later one on app.ci. I don't think we should reference either as comment

alvaroaleman (approved):

PR #1262, opened 2020-09-29
Steps: Make errors a bit more helpful

alvaroaleman:
For https://issues.redhat.com/browse/DPTP-1493

This PR makes errors related to dependency issues in the steps a bit easier to understand by:
* De-Duplicating them
* Avoiding to print steps whose depend on steps that have unfulfilled dependencies by first creating a list of all available dependencies
* Allowing StepLinks to contain an explanation as to why this is needed, print that instead of the raw StepLink and fill it in for the `Inputs` of the `ProjectImage` step

A sample output for a config similiar to whats in the card looks like this:
```
step <*steps.projectDirectoryImageBuildStep> is missing dependencies: "cli" is neither an imported nor a built image
```

This doesn't fully cover the AC, because to do that, we would need to also allow Steps to contain the info where exactly in the config they come from (implying extending the interface) and would need to fill that in for the `projectDirectoryImageBuilderStep`. Since that is a lot of work I have punted on it for now, but if we think its worth the time, I can create a follow-up.

/assign @petr-muller

### DIFF
--- a/cmd/ci-operator/main_test.go
+++ b/cmd/ci-operator/main_test.go
@@ -391,7 +391,7 @@ func TestBuildPartialGraph(t *testing.T) {
 					fakeimageclientset.NewSimpleClientset(&imagev1.ImageStreamTag{ObjectMeta: metav1.ObjectMeta{Name: ":"}}).ImageV1(),
 					nil,
 				),
-				steps.SourceStep(api.SourceStepConfiguration{}, api.ResourceConfiguration{}, nil, nil, "", &api.JobSpec{}, nil, nil),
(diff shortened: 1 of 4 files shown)


cmd/ci-operator/main_test.go
```diff
@@ -371,3 +375,56 @@ func TestErrWroteJUnit(t *testing.T) {
 		t.Errorf(`expected full reason for error to be "something", but got %q"`, results.FullReason(defaulted))
 	}
 }
+
+func TestBuildPartialGraph(t *testing.T) {
```
alvaroaleman, line 379: This testcase doesn't really provide a lot of value as the returned error is always the same but I needed it in order to reproduce the issue

PR #1257, opened 2020-09-24
pkg/steps: Configurable activeDeadlineSeconds and terminationGracePeriodSeconds

wking:
Sometimes test steps like updates [take a long time][1].  When that happens, something wrapping the workflow may be timed out instead, and the whole thing gets torn down without any post-test gather/teardown steps running.  With this commit, we expose two timeout-related [properties from `PodSpec`][2] to step maintainers to use to control this behavior.  Step maintainers can now set per-step timeouts that are comfortably less than the wrapper timeout, ensuring that slow/hung steps get killed, and the remaining steps have some time to gather details that will explain the delay and perform other teardown.

[1]: https://github.com/openshift/cluster-network-operator/pull/785#issuecomment-698017421
[2]: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.18/#podspec-v1-core

PR #1265, opened 2020-10-02
activeDeadlineSeconds support for workflows

vrutkovs:
Alternative to https://github.com/openshift/ci-tools/pull/1257. It doesn't use additional channels, but notifies the pod has completed if it failed and activeDeadlineSeconds set (not checking for container statuses due to k8s bug)


vrutkovs:
No longer needed

PR #1271, opened 2020-10-05
CI Operator: Bump the namespace active annotation every 10 minutes

alvaroaleman:
To prevent races where the ci-ns-ttl controllers cleans up the NS because the ci-operator deleted all pods but plans on creating more

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -155,6 +155,9 @@ const (
 	annotationIdleCleanupDurationTTL = "ci.openshift.io/ttl.soft"
 	// annotationCleanupDurationTTL is the annotation for requesting namespace cleanup after the namespace has been active
 	annotationCleanupDurationTTL = "ci.openshift.io/ttl.hard"
+	// annotationNamespaceLastActive contains time.RFC3339 timestamp at which the namespace was last in active use. We
+	// update this every ten minutes.
+	annotationNamespaceLastActive = "ci.openshift.io/active"
 	// leaseServerUsername is the default lease server username in api.ci
 	leaseServerUsername = "ci"
 	leaseAcquireTimeout = 120 * time.Minute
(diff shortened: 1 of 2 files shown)


stevekuznetsov (commented):
Can we remove the logic from `lease.go`?

alvaroaleman:
> Can we remove the logic from lease.go?

Done. How could this bug happen in the first place if we already had this logic? Because not all jobs have a lease?

petr-muller:
> > Can we remove the logic from lease.go?
> 
> Done. How could this bug happen in the first place if we already had this logic? Because not all jobs have a lease?

Reading the bits you removed from `lease.go`, I think we actually did heartbeat the temporary namespace only while waiting for the lease (the heartbeat was cancelled just before the "lease acquired" message.

petr-muller (approved):

bbguimaraes:
> Reading the bits you removed from `lease.go`, I think we actually did heartbeat the temporary namespace only while waiting for the lease (the heartbeat was cancelled just before the "lease acquired" message.

Correct, that was specifically for the case where a lease is unavailable for a period longer than the soft TTL, because we only schedule test pods after it is acquired.  See https://github.com/openshift/ci-tools/pull/807.

The problem we are solving with this is for the short period between a test being cancelled (and its pod deleted) and its `post` steps running, where there is no active pod in the namespace.  There is no active pod in the namespace during that window, so the controller will decide to delete the namespace if it sees that state.

cmd/ci-operator/main.go
```diff
@@ -815,7 +818,7 @@ func (o *options) initializeNamespace() error {
 
 	// This label makes sure that the namespace is active, and the value will be updated
 	// if the namespace will be reused.
-	updates["ci.openshift.io/active"] = time.Now().Format(time.RFC3339)
+	updates[annotationNamespaceLastActive] = time.Now().Format(time.RFC3339)
 
 	if len(updates) > 0 {
```
bbguimaraes, line 823: This is a bit silly =)

(unrelated to the change)

PR #1274, opened 2020-10-06
template-deprecator: add new tool

petr-muller:
~This code makes `sanitize-prow-jobs` _optionally_ read config-updater config, get a list of all existing templates, and compile an allowlist of all jobs that currently use these templates. This allowlist is merged with a previously saved one~.

I made the second iteration a separate binary, because the functionality did not really fit into `sanitize-prow-jobs` in the end (it would become a totally modal tool, wiith separate options etc). Also, we may want to make interaction with the allowllist easier for users via the tool, and an explicit tool is better for that.

This is how the allowlist looks like:

```yaml
templates:
  prow-job-cluster-launch-e2e:
    blockers:
      JIRA-1234:
        description: "Things do not thing the way they should thing"
        jobs:
          periodic-ci-openshift-origin-release-3.11-e2e-gcp:
            generated: false
            kind: periodic
          pull-ci-image-registry-e2e-3.10:
            generated: false
            kind: presubmit
    template_name: prow-job-cluster-launch-e2e
    unknown_blocker:
      description: unknown
      jobs:
        pull-ci-openshift-image-registry-release-3.11-e2e-gcp:
          generated: true
          kind: presubmit
```

By default, new jobs are always added to the `unknown_blocker` list. From there, humans can (currenlty only manually) move the jobs to the `blockers` list, keyed by Jira numbers. If a job is in a known blocker list for a template, it will not be added to the unknown blocker list anymore.

### DIFF
--- a/cmd/template-deprecator/main.go
+++ b/cmd/template-deprecator/main.go
@@ -1,0 +1,86 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/sirupsen/logrus"
+
+	prowconfig "k8s.io/test-infra/prow/config"
+	prowplugins "k8s.io/test-infra/prow/plugins"
+
+	"github.com/openshift/ci-tools/pkg/deprecatetemplates"
+)
+
+type options struct {
+	prowJobConfigDir     string
+	prowConfigPath       string
+	prowPluginConfigPath string
+	allowlistPath        string
+
+	help bool
+}
+
+func bindOptions(fs *flag.FlagSet) *options {
+	opt := &options{}
+
+	fs.StringVar(&opt.prowJobConfigDir, "prow-jobs-dir", "", "Path to a root of directory structure with Prow job config files (ci-operator/jobs in openshift/release)")
+	fs.StringVar(&opt.prowConfigPath, "prow-config-path", "", "Path to the Prow configuration file")
+	fs.StringVar(&opt.prowPluginConfigPath, "prow-plugin-config-path", "", "Path to the Prow plugin configuration file")
+	fs.StringVar(&opt.allowlistPath, "allowlist-path", "", "Path to template deprecation allowlist")
+
+	return opt
+}
+
+func (o *options) validate() error {
+	for param, value := range map[string]string{
+		"--prow-jobs-dir":           o.prowJobConfigDir,
+		"--prow-config-path":        o.prowConfigPath,
+		"--prow-plugin-config-path": o.prowPluginConfigPath,
+		"--allowlist-path":          o.allowlistPath,
+	} {
+		if value == "" {
+			return fmt.Errorf("mandatory argument %s was not set", param)
+		}
+	}
+
+	return nil
+}
+
+func main() {
+	opt := bindOptions(flag.CommandLine)
+	flag.Parse()
+
+	if opt.help {
+		flag.Usage()
+		os.Exit(0)
+	}
+
+	if err := opt.validate(); err != nil {
+		logrus.WithError(err).Fatal("Invalid parameters")
+	}
+
+	agent := prowplugins.ConfigAgent{}
+	if err := agent.Load(opt.prowPluginConfigPath, true); err != nil {
+		logrus.WithError(err).Fatal("Failed to read Prow plugin configuration")
+	}
+	pluginCfg := agent.Config().ConfigUpdater
+
+	prowCfg, err := prowconfig.Load(opt.prowConfigPath, opt.prowJobConfigDir)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load Prow configuration")
+	}
+
+	enforcer, err := deprecatetemplates.NewEnforcer(opt.allowlistPath)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to initialize template deprecator")
+	}
+
+	enforcer.LoadTemplates(pluginCfg)
+	enforcer.ProcessJobs(prowCfg)
+
+	if err := enforcer.SaveAllowlist(opt.allowlistPath); err != nil {
+		logrus.WithError(err).Fatal("Failed to save template deprecation allowlist")
+	}
+}
--- a/pkg/deprecatetemplates/allowlist.go
+++ b/pkg/deprecatetemplates/allowlist.go
@@ -1,0 +1,148 @@
+package deprecatetemplates
+
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/config"
+
+	jc "github.com/openshift/ci-tools/pkg/jobconfig"
+)
+
+const (
+	release = "release"
+	unknown = "unknown"
+)
+
+func isGenerated(job config.JobBase) bool {
+	_, generated := job.Labels[jc.ProwJobLabelGenerated]
+	return generated
+}
+
+func getKind(job config.JobBase) string {
+	if strings.HasPrefix(job.Name, "pull-ci") {
+		return string(prowv1.PresubmitJob)
+	} else if strings.HasPrefix(job.Name, "branch-ci-") {
+		return string(prowv1.PostsubmitJob)
+	} else if strings.HasPrefix(job.Name, "release-") {
+		return release
+	} else if strings.HasPrefix(job.Name, "periodic-") {
+		return string(prowv1.PeriodicJob)
+	}
+
+	// this is fine, it is best effort
+	// the allowlist does not need to be 100% precise category-wise
+	return unknown
+}
+
+type blockedJob struct {
+	Generated bool   `json:"generated"`
+	Kind      string `json:"kind"`
+}
+
+type blockedJobs map[string]blockedJob
+
+func (b blockedJobs) Has(job config.JobBase) bool {
+	_, has := b[job.Name]
+	return has
+}
+
+func (b blockedJobs) Insert(job config.JobBase) {
+	b[job.Name] = blockedJob{
+		Generated: isGenerated(job),
+		Kind:      getKind(job),
+	}
+}
+
+func (b blockedJobs) Union(other blockedJobs) blockedJobs {
+	union := blockedJobs{}
+	for k, v := range b {
+		union[k] = v
+	}
+	for k, v := range other {
+		union[k] = v
+	}
+	return union
+}
+
+type deprecatedTemplateBlocker struct {
+	Description string      `json:"description"`
+	Jobs        blockedJobs `json:"jobs"`
+}
+
+type deprecatedTemplate struct {
+	Name           string                               `json:"template_name"`
+	UnknownBlocker deprecatedTemplateBlocker            `json:"unknown_blocker"`
+	Blockers       map[string]deprecatedTemplateBlocker `json:"blockers,omitempty"`
+}
+
+func (d deprecatedTemplate) insert(job config.JobBase) {
+	for _, blocker := range d.Blockers {
+		if blocker.Jobs.Has(job) {
+			return
+		}
+	}
+	d.UnknownBlocker.Jobs.Insert(job)
+}
+
+type Allowlist interface {
+	Insert(job config.JobBase, template string)
+	Save(path string) error
+}
+
+type allowlist struct {
+	Templates map[string]deprecatedTemplate `json:"templates"`
+}
+
+func (a *allowlist) Insert(job config.JobBase, template string) {
+	if a.Templates == nil {
+		a.Templates = map[string]deprecatedTemplate{}
+	}
+
+	if _, ok := a.Templates[template]; !ok {
+		a.Templates[template] = deprecatedTemplate{
+			Name: template,
+			UnknownBlocker: deprecatedTemplateBlocker{
+				Description: "unknown",
+				Jobs:        blockedJobs{},
+			},
+		}
+	}
+
+	a.Templates[template].insert(job)
+}
+
+func loadAllowlist(allowlistPath string) (Allowlist, error) {
+	var allowlist allowlist
+
+	raw, err := ioutil.ReadFile(allowlistPath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	if err == nil {
+		err := yaml.Unmarshal(raw, &allowlist)
+		return &allowlist, err
+	}
+
+	logrus.Warn("template deprecation allowlist does not exist, will populate a new one")
+	return &allowlist, nil
+}
+
+func (a allowlist) Save(path string) error {
+	raw, err := yaml.Marshal(a)
+	if err != nil {
+		return err
+	}
+
+	if err := ioutil.WriteFile(path, raw, 0644); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/deprecatetemplates/allowlist.go
+++ b/pkg/deprecatetemplates/allowlist.go
@@ -11,18 +11,13 @@ import (
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config"
 
-	jc "github.com/openshift/ci-tools/pkg/jobconfig"
+	"github.com/openshift/ci-tools/pkg/prowgen"
 )
 
 const (
 	release = "release"
 	unknown = "unknown"
 )
-
-func isGenerated(job config.JobBase) bool {
-	_, generated := job.Labels[jc.ProwJobLabelGenerated]
-	return generated
-}
 
 func getKind(job config.JobBase) string {
 	if strings.HasPrefix(job.Name, "pull-ci") {
@@ -54,7 +49,7 @@ func (b blockedJobs) Has(job config.JobBase) bool {
 
 func (b blockedJobs) Insert(job config.JobBase) {
 	b[job.Name] = blockedJob{
-		Generated: isGenerated(job),
+		Generated: prowgen.IsGenerated(job),
 		Kind:      getKind(job),
 	}
 }
--- a/pkg/deprecatetemplates/allowlist.go
+++ b/pkg/deprecatetemplates/allowlist.go
@@ -72,7 +72,7 @@ func (b blockedJobs) Union(other blockedJobs) blockedJobs {
 
 type deprecatedTemplateBlocker struct {
 	Description string      `json:"description"`
-	Jobs        blockedJobs `json:"jobs"`
+	Jobs        blockedJobs `json:"jobs,omitempty"`
 }
 
 type deprecatedTemplate struct {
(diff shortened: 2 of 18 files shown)


cmd/sanitize-prow-jobs/main.go
```diff
@@ -75,6 +93,10 @@ func determinizeJobs(prowJobConfigDir string, config *dispatcher.Config) error {
 
 			defaultJobConfig(jobConfig, path, config)
 
+			if dtEnforcer.Enabled {
```
alvaroaleman, line 96: please spell this out to improve readability

cmd/sanitize-prow-jobs/main.go
```diff
@@ -127,18 +149,34 @@ func main() {
...
 
-	if len(opt.prowJobConfigDir) == 0 {
-		logrus.Fatal("mandatory argument --prow-jobs-dir wasn't set")
+	if err := opt.validate(); err != nil {
+		logrus.WithError(err).Fatal("Invalid options.")
 	}
-	if len(opt.configPath) == 0 {
-		logrus.Fatal("mandatory argument --config-path wasn't set")
+
+	dtEnforcer, err := opt.deprecateTemplates.GetEnforcer()
```
alvaroaleman, line 156: Spell out

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type deprecatedTemplateBlocker struct {
+	Jira        string      `json:"jira"`
+	Description string      `json:"description"`
+	Jobs        sets.String `json:"jobs"`
+}
+
+type deprecatedTemplate struct {
+	Name           string                               `json:"template_name"`
```
stevekuznetsov, line 20: If this is stored in a map keyed by name, why do we need this field?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -127,18 +149,34 @@ func main() {
...
 	}
+
+	if dtEnforcer.Enabled {
+		if err := dtEnforcer.SaveAllowlist(opt.deprecateTemplates.AllowlistPath); err != nil {
+			logrus.WithError(err).Fatal("Failed to save template deprecation allowlist")
+		}
+
+		if errs := dtEnforcer.Check(); len(errs) > 0 {
+			for _, err := range errs {
+				logrus.WithError(err).Errorf("Failed template deprecation enforcement check")
```
alvaroaleman, line 177: s/Errorf/Error (Why does the linter not fail over this?)

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+type deprecatedTemplateBlocker struct {
+	Jira        string      `json:"jira"`
+	Description string      `json:"description"`
+	Jobs        sets.String `json:"jobs"`
```
stevekuznetsov, line 16: Can this be a slice so the serialized form does not look so silly?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -127,18 +149,34 @@ func main() {
...
 		logrus.WithError(err).Fatalf("Failed to load config from %q", opt.configPath)
 	}
-	if err := determinizeJobs(opt.prowJobConfigDir, config); err != nil {
+
+	if err := determinizeJobs(opt.prowJobConfigDir, config, dtEnforcer); err != nil {
 		logrus.WithError(err).Fatal("Failed to determinize")
 	}
+
+	if dtEnforcer.Enabled {
+		if err := dtEnforcer.SaveAllowlist(opt.deprecateTemplates.AllowlistPath); err != nil {
```
alvaroaleman, line 171: Shouldn't this be behind another flag? I.E the allowlist will be only manually updated whereas the check if all template jobs are in the allowlist will eventually run in a presubmit

stevekuznetsov (commented):
Small comments on the shape of the file, since humans will be reading it and updating it. Maybe we want a (tiny) CLI to add job(s) and their blocking Jira if the manual edit will be annoying?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -127,18 +149,34 @@ func main() {
...
 	}
 
 	config, err := dispatcher.LoadConfig(opt.configPath)
 	if err != nil {
 		logrus.WithError(err).Fatalf("Failed to load config from %q", opt.configPath)
 	}
-	if err := determinizeJobs(opt.prowJobConfigDir, config); err != nil {
+
+	if err := determinizeJobs(opt.prowJobConfigDir, config, dtEnforcer); err != nil {
 		logrus.WithError(err).Fatal("Failed to determinize")
```
alvaroaleman, line 167: It would be better if we combined the errors here and from the new check if enabled rather than fatal out directly, otherwise ppl might need multiple pushes to see all mistakes they made, which is annoying

pkg/deprecatetemplates/options.go
```diff
@@ -0,0 +1,68 @@
+package deprecatetemplates
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+
+	_ "k8s.io/test-infra/prow/hook/plugin-imports"
```
alvaroaleman, line 8: Is this really needed rather than just importing the config-updater plugin?

pkg/deprecatetemplates/enforcer.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+
+	return unused
+}
+
+func (e *Enforcer) Check() []error {
+	var allErrors []error
+
+	unusedTemplates := e.unusedTemplates()
```
alvaroaleman, line 75: This is not unused, its unallowed or am I misreading this?

cmd/sanitize-prow-jobs/main.go
```diff
@@ -75,6 +93,10 @@ func determinizeJobs(prowJobConfigDir string, config *dispatcher.Config) error {
 
 			defaultJobConfig(jobConfig, path, config)
 
+			if dtEnforcer.Enabled {
+				dtEnforcer.IngestAll(jobConfig)
```
alvaroaleman, line 97: Also, I would have expected the `IngestAll` call to be made part of the be executed in the `GetEnforcer` call, its not obvious from the public api that this must be called first and IMHO it doesn't really make sense to force consumers of the Enforcer to know this

pkg/deprecatetemplates/enforcer.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+	}
+
+	e.existingTemplates = templates
+}
+
+func (e *Enforcer) ingest(job prowconfig.JobBase) {
+	for template := range e.existingTemplates {
+		if rehearse.UsesConfigMap(job, template) {
+			e.allowlist.Insert(job.Name, template)
```
alvaroaleman, line 36: Shouldn't the `allowList.Insert` be gated by an `if update` check? Right now the verification can never fail

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,90 @@
...
+func loadAllowlist(allowlistPath string) (*Allowlist, error) {
+	raw, err := ioutil.ReadFile(allowlistPath)
+	if err != nil && !os.IsNotExist(err) {
+		return nil, err
+	}
+
+	var allowlist Allowlist
+	if err == nil {
+		err := yaml.Unmarshal(raw, &allowlist)
+		return &allowlist, err
```
alvaroaleman, line 72: nit: `return &allowlist, yaml.Unmarshal(raw, &allowlist)`

pkg/deprecatetemplates/allowlist.go
petr-muller: I wanted the structure to be independent, carrying all relevant informaiton. Carry the pointer to the JIRA so it is usable even without passing the key separately. 

But I guees I could do that programatically on load without polluting the file, and only add that when it's useful. :+1:

pkg/deprecatetemplates/allowlist.go
petr-muller: Not really. I want to add some per-job information to the allowlist (mostly if the job is handcrafted or generated), so it is obvious from the file.

petr-muller:
This is supposed to be supplemented with the doc I mentioned in the scrum, explaining my idea of how this should evolve. I don't have that doc yet, but I will have it today. It should explain some of the design choices. Let me write it, let's discuss it, and some of the review comments should be addressed by that.

pkg/deprecatetemplates/allowlist.go
stevekuznetsov: Err either this is a list of jobs (which can be deduped on load) or it's a mapping of job to something, and not a set. I am just saying don't use the set type since the serialization looks bad.

alvaroaleman:
This is for dptp-1515

petr-muller:
@stevekuznetsov @alvaroaleman just pushed a new version of this here, happy to get reviews. Some of the previous comments were obsoleted by the move to a separate binary, the rest were addressed.

I've got a little bogged down by the tests for the stats and the checks, so for now I've taken that out of here and the thing only maintains the allowlist currently (I will finish the rest shortly, but thought this part can as well go in) so it's possible to set up that presubmit.

cmd/template-deprecator/main.go
```diff
@@ -0,0 +1,88 @@
...
+	fs.StringVar(&opt.prowJobConfigDir, "prow-jobs-dir", "", "Path to a root of directory structure with Prow job config files (ci-operator/jobs in openshift/release)")
+	fs.StringVar(&opt.prowConfigPath, "prow-config-path", "", "Path to the Prow configuration file")
+	fs.StringVar(&opt.prowPluginConfigPath, "prow-plugin-config-path", "", "Path to the Prow plugin configuration file")
+	fs.StringVar(&opt.allowlistPath, "allowlist-path", "", "Path to template deprecation allowlist")
+
+	return opt
+}
+
+func (o *options) validate() error {
+	for param, value := range map[string]string{
```
stevekuznetsov, line 36: golly do i like anonymous maps

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+	"k8s.io/test-infra/prow/config"
+
+	jc "github.com/openshift/ci-tools/pkg/jobconfig"
+)
+
+const (
+	presubmit  = "presubmit"
```
stevekuznetsov, line 16: These have upstream constants in test-infra somewhere if you care

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+	postsubmit = "postsubmit"
+	periodic   = "periodic"
+	release    = "release"
+	unknown    = "unknown"
+)
+
+func isGenerated(job config.JobBase) bool {
```
stevekuznetsov, line 23: we have private copies of this function in a number of places now in this repo - can we DRY it out?

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+
+	dt := a.Templates[template]
+	dt.insert(job)
+}
+
+func loadAllowlist(allowlistPath string) (Allowlist, error) {
+	var allowlist allowlist
+
+	raw, err := ioutil.ReadFile(allowlistPath)
+	if err != nil && !os.IsNotExist(err) {
```
stevekuznetsov, line 126: if we require the flag why do we allow a non-existent file?

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+		return nil, err
+	}
+
+	if err == nil {
+		err := yaml.Unmarshal(raw, &allowlist)
+		return &allowlist, err
+	}
+
+	logrus.Warn("template deprecation allowlist does not exist, will populate a new one")
```
stevekuznetsov, line 135: ah that's why

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+	return &allowlist, nil
+}
+
+func (a *allowlist) Save(path string) error {
+	raw, err := yaml.Marshal(*a)
```
stevekuznetsov, line 140: this could NPE

stevekuznetsov (commented):
mostly lgtm

cmd/template-deprecator/main.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+	}
+
+	return nil
+}
+
+func main() {
+	flagSet := flag.NewFlagSet("", flag.ExitOnError)
+	opt := bindOptions(flagSet)
+	if err := flagSet.Parse(os.Args[1:]); err != nil {
```
alvaroaleman, line 53: Nit: These three lines are the same as:
```
opt := bindOptions(flag.CommandLine)
flag.Parse()
```

And its impossible to go into the error condition below because `flag.ExitOnError` is set

cmd/template-deprecator/main.go
```diff
@@ -0,0 +1,88 @@
...
+
+	enforcer, err := deprecatetemplates.NewEnforcer(opt.allowlistPath)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to initialize template deprecator")
+	}
+
+	enforcer.LoadTemplates(pluginCfg)
+	enforcer.ProcessJobs(prowCfg)
+
+	if err := enforcer.SaveAllowlist(opt.allowlistPath); err != nil {
```
alvaroaleman, line 85: I still don't understand this. Shouldn't this be gated by a flag so we can run the tool in a presubmit to prevent ppl from adding new jobs that are not in the allowlist?

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+type blockedJob struct {
+	Generated bool   `json:"generated"`
+	Kind      string `json:"kind"`
+}
+
+type blockedJobs map[string]blockedJob
+
+func (b blockedJobs) Has(job config.JobBase) bool {
```
alvaroaleman, line 51: I don't see us ever caring about the values in this map, can this be a simple `string.Set` or did I miss it?

cmd/template-deprecator/main.go
```diff
@@ -0,0 +1,88 @@
+package main
```
alvaroaleman, line 1: No readme?

pkg/deprecatetemplates/allowlist.go
```diff
@@ -0,0 +1,150 @@
...
+		a.Templates[template] = deprecatedTemplate{
+			Name: template,
+			UnknownBlocker: deprecatedTemplateBlocker{
+				Description: "unknown",
+				Jobs:        blockedJobs{},
+			},
+		}
+	}
+
+	dt := a.Templates[template]
```
alvaroaleman, line 118: Nit: Why not just `a.Templates[template].insert(job)`?

pkg/deprecatetemplates/enforcer_test.go
```diff
@@ -0,0 +1,158 @@
+package deprecatetemplates
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	v1 "k8s.io/api/core/v1"
```
alvaroaleman, line 7: can we please alias this as `corev1`?

pkg/deprecatetemplates/enforcer_test.go
```diff
@@ -0,0 +1,158 @@
...
+	presubmits  []config.Presubmit
+	postsubmits []config.Postsubmit
+	periodics   []config.Periodic
+}
+
+func (m *mockJobConfig) AllStaticPostsubmits(_ []string) []config.Postsubmit {
+	return append([]config.Postsubmit{}, m.postsubmits...)
+}
+func (m *mockJobConfig) AllStaticPresubmits(_ []string) []config.Presubmit {
+	return append([]config.Presubmit{}, m.presubmits...)
```
alvaroaleman, line 75: Why the append? To get a new slice?

cmd/template-deprecator/main.go
petr-muller: Not sure if this is irony or not, but I think it's fine here :)

cmd/template-deprecator/main.go
petr-muller: I don't think so. Adding jobs to the allowlist is still legit thing to do for some time. The presubmit will 

1. Run the tool.
2. If fails, it means "you should do something else". There's no such check in the tool ATM, it's a followup. We would be adding more checks like this as we tighten the template world. And you are right, in that case the allowlist should not be updated on disk.
2. If is succeeds, compare the allowlist (with git). If the allowlist changed, the presubmit will also fail, but ask to include the allowlist change in the PR.

pkg/deprecatetemplates/allowlist.go
petr-muller: They are not programatically used right now, correct. I had two reasons to have these:

1. I want to be able to add a check like "no generated jobs should be in the allowlist for a template X"
2. I want to be able to eyeball the allowlist and understand what kinds of jobs are involved (that's why they are serialized)

pkg/deprecatetemplates/enforcer_test.go
petr-muller: Yeah, I wanted the mock to vend out copies, to prevent any possible interference.

pkg/deprecatetemplates/allowlist.go
petr-muller: Hm, the reason was 
```
pkg/deprecatetemplates/allowlist.go:111:23: cannot call pointer method on a.Templates[template]
pkg/deprecatetemplates/allowlist.go:111:23: cannot take the address of a.Templates[template]
```

but it looks like I can just just make the `insert` have a non-pointer receiver.

cmd/template-deprecator/main.go
petr-muller: Good point, I totally forgot

petr-muller:
Thanks for feedback, PR updated!

cmd/template-deprecator/main.go
alvaroaleman: Ah I see, I guess I expected the check to be in the tool

pkg/deprecatetemplates/allowlist.go
alvaroaleman: So remove them now and re-add them once you need them? Right now its premature optimization (and because the methods are exported, the deadcode checker won't complain)

petr-muller: I'd rather not. It's not premature optimization b/c of reason 2. It's also not some hypothetical futureproofing, I *know* I will want to use the fields as soon as next week when we migrate all generated jobs for a template and I will want to add a "no generated jobs for this template" check.

alvaroaleman (approved):

PR #1275, opened 2020-10-07
Support generating testgrid dashboards for e2e presubmit jobs

marun:
Because of the volume of presubmit jobs across all repos, jobs for a given branch are sharded across multiple dashboards to minimize the potential for hitting the per-dashboard timeout testgrid imposes on data collection. 

Example output: https://github.com/kubernetes/test-infra/pull/19473

/cc @stevekuznetsov

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -19,6 +19,7 @@ import (
 	"github.com/sirupsen/logrus"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	prowconfig "k8s.io/test-infra/prow/config"
 	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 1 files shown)


cmd/testgrid-config-generator/main.go
```diff
@@ -182,6 +183,94 @@ type release struct {
...
+	for _, p := range jobConfig.AllStaticPresubmits(nil) {
+		// Only track mandatory jobs
+		if !p.AlwaysRun || p.Optional {
+			continue
+		}
+		// Only track presubmits for repos in the openshift org. This is intended
+		// to minimize the cost of testgrid collection and can be relaxed if
+		// necessary. openshift-priv should not be collected as it rarely sees
+		// presubmit activity.
+		if !strings.HasPrefix(p.Name, "pull-ci-openshift-") ||
```
stevekuznetsov, line 199: The ignored key in the mapping on line 190 is the org/repo, use that instead of stringified names for filtering. You can also ask for jobs for a repo.

cmd/testgrid-config-generator/main.go
```diff
@@ -182,6 +183,94 @@ type release struct {
 	} `json:"verify"`
 }
 
+// e2ePresubmitJobsByBranch collects statically-defined e2e presubmit jobs (e2e,
+// serial and upgrade) grouped by the targeted branch.
+func e2ePresubmitJobsByBranch(presubmitBranches []string, jobConfig *prowconfig.JobConfig) map[string][]string {
+	presubmitJobs := map[string][]string{}
+	for _, p := range jobConfig.AllStaticPresubmits(nil) {
+		// Only track mandatory jobs
```
stevekuznetsov, line 191: This is not technically a Prow-ism, so if you were hoping for this to catch all that is required for merge, we need something else

cmd/testgrid-config-generator/main.go
```diff
@@ -182,6 +183,94 @@ type release struct {
...
+		// to minimize the cost of testgrid collection and can be relaxed if
+		// necessary. openshift-priv should not be collected as it rarely sees
+		// presubmit activity.
+		if !strings.HasPrefix(p.Name, "pull-ci-openshift-") ||
+			strings.HasPrefix(p.Name, "pull-ci-openshift-priv") {
+
+			continue
+		}
+		// Group jobs by branch
+		for _, branch := range presubmitBranches {
```
stevekuznetsov, line 205: The job's configuration has the branches it can be triggered on, let's use that

stevekuznetsov (commented):
We shouldn't use job name as a stable API. How sure are we that adding N dashboards is kosher upstream? This seems to allow for a proliferation of dashboards and I was under the understanding that they did not want us to abuse this.

cmd/testgrid-config-generator/main.go
marun: I care less about what is required to merge as limiting the scope of collection. There are a ton of optional on-demand jobs that wouldn't add much signal but would potentially slow testgrid collection down.

marun:
> We shouldn't use job name as a stable API. How sure are we that adding N dashboards is kosher upstream? This seems to allow for a proliferation of dashboards and I was under the understanding that they did not want us to abuse this.

Regarding 'job name as stable API' what are you referring to? Renaming jobs?

I asked upstream in a thread I'll link you to in slack, and so long as we weren't hitting timeouts they seemed open to the idea of sharding to ensure we stay within the limitations imposed by testgrid's configured timeouts.

cmd/testgrid-config-generator/main.go
marun: I get that I can ask for jobs for a repo. How do I get repos for an org?

cmd/testgrid-config-generator/main.go
marun: As per the comment below this line, I care about more than the branch. I want to limit skip random repo-local e2e jobs that are typically characterized by not following the naming convention of <branch>-e2e. Really I want to target only jobs that are derived from workflows, but that implementation would be considerably more involved afaict.

marun:
Also, thanks for reviewing!

cmd/testgrid-config-generator/main.go
marun: Updated the comment to reflect my intent, does that make more sense?

PR #1058, opened 2020-07-31
Update test-infra and k8s.io/* dependencies

alvaroaleman:
Downstream follow-up to https://github.com/kubernetes/test-infra/pull/17869
Its a big changeset because the signature of the generated clients was changed

/label tide/merge-method-squash

PR #1276, opened 2020-10-07
pkg/steps/template: Replace context.TODO in waitForPodCompletionOrTimeout

wking:
The TODO are from 0dcdf1090b (#1058).  Possibly they were just `sed`ed in, because the function has taken a Context argument since c242af1190 (#510).

CC @alvaroaleman


alvaroaleman (approved):
Yeah, these were added by [clientgofix](https://github.com/kubernetes-sigs/clientgofix)

wking:
Mostly reverted in #1281.

PR #1277, opened 2020-10-07
pkg/steps: Add a finalizer for pod collection

wking:
And, when we see that finalizer in the pod and are no longer interested in watching it, remove the finalizer so the deletion can go through.

This should avoid cases where the parallel Delete (e.g. the one in `podStep.run()`) removes the pod, the change somehow slips through the watch, and we [lose track of the pod][1]:

    2020/10/06 23:04:25 error: could not wait for pod 'timeout-timeout': it is no longer present on the cluster (usually a result of a race or resource pressure. re-running the job should help)

Spun out of #1257.

/assign @alvaroaleman

[1]: https://github.com/openshift/ci-tools/pull/1257#issuecomment-704613892

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -39,6 +39,7 @@ const (
 	RefsOrgLabel    = "ci.openshift.io/refs.org"
 	RefsRepoLabel   = "ci.openshift.io/refs.repo"
 	RefsBranchLabel = "ci.openshift.io/refs.branch"
+	finalizer       = "ci.openshift.io/read-final-status"
 
 	TestContainerName = "test"
 )
@@ -514,7 +515,7 @@ func waitForPodCompletion(ctx context.Context, podClient coreclientset.PodInterf
 	return pod, nil
 }
 
-func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.PodInterface, eventClient coreclientset.EventInterface, name string, completed map[string]time.Time, notifier ContainerNotifier, skipLogs bool) (*coreapi.Pod, bool, error) {
+func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.PodInterface, eventClient coreclientset.EventInterface, name string, completed map[string]time.Time, notifier ContainerNotifier, skipLogs bool) (pod *coreapi.Pod, retry bool, err error) {
 	watcher, err := podClient.Watch(ctx, meta.ListOptions{
 		FieldSelector: fields.Set{"metadata.name": name}.AsSelector().String(),
 		Watch:         true,
@@ -534,7 +535,13 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
 			" (usually a result of a race or resource pressure. re-running the job should help)", name)
 		return nil, false, fmt.Errorf("pod was deleted while ci-operator step was waiting for it")
 	}
-	pod := &list.Items[0]
+	pod = &list.Items[0]
+	defer func() {
+		if !retry || err != nil {
+			removeFinalizer(ctx, podClient, pod, finalizer)
+		}
+	}()
+
 	if pod.Spec.RestartPolicy == coreapi.RestartPolicyAlways {
 		return pod, false, nil
 	}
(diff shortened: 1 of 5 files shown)


pkg/steps/template.go
```diff
@@ -591,6 +598,38 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
...
+	}
+
+	nextFinalizers := make([]string, 0, len(pod.ObjectMeta.Finalizers))
+	for _, entry := range pod.ObjectMeta.Finalizers {
+		if entry != finalizer {
+			nextFinalizers = append(nextFinalizers, entry)
+		}
+	}
+	if len(nextFinalizers) != len(pod.ObjectMeta.Finalizers) {
+		data, err := json.Marshal(map[string]interface{}{
```
alvaroaleman, line 616: lol

pkg/steps/template.go
```diff
@@ -591,6 +598,38 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
...
+func removeFinalizer(ctx context.Context, podClient coreclientset.PodInterface, pod *coreapi.Pod, finalizer string) {
+	if ctx.Err() != nil {
+		// give ourselves a bit of a grace period to remove the finalizer
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), 5*time.Minute)
+		defer cancel()
+	}
+
+	nextFinalizers := make([]string, 0, len(pod.ObjectMeta.Finalizers))
+	for _, entry := range pod.ObjectMeta.Finalizers {
```
alvaroaleman, line 610: Can be simpliefied as `sets.NewString(pod.FInalizers).Delete(finalizer).List()`

pkg/steps/template.go
```diff
@@ -591,6 +598,38 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
...
+			"value": nextFinalizers,
+		})
+		if err != nil {
+			log.Printf("Failed to marshal patch for removing finalizer %s from pod %s: %v", finalizer, pod.Name, err)
+			return
+		}
+		if _, err = podClient.Patch(ctx, pod.Name, types.JSONPatchType, data, meta.PatchOptions{}); err == nil {
+			log.Printf("Finished watching pod %s; removed finalizer %s", pod.Name, finalizer)
+		} else {
+			log.Printf("Finished watching pod %s; failed to remove finalizer %s: %v", pod.Name, finalizer, err)
```
alvaroaleman, line 628: We need something else that removes this finalizers if ci-operator fails to do so for whatever reasons, otherwise we will have stuck nodes

pkg/steps/template.go
wking: >    sets.NewString(pod.FInalizers).Delete(finalizer).List()

Does that preserve order?  Because I want to delete our entry without shuffling the remaining entries, if any have been added out of band.

alvaroaleman: A very explicit mention about finalizers is that order must not matter (because if it did, everyone would want to be last)

alvaroaleman: Ref: https://github.com/kubernetes/kubernetes/pull/83110/files#diff-0bc759fffb6a495573c6b70a9afbf741R666

pkg/steps/template.go
wking: What something else would that be?

And narrowly for #1257, I'm comfortable with bf10feb3420588b5345cbc6f75986ab5e29a562a, instead of the parallel `Delete` workaround, so feel free to reject this PR if you don't want the guard for the existing, parallel `Delete`.

alvaroaleman: I don't know, but until we have something, we can not merge this change. Maybe we could add it to the namespace reaper controller?

pkg/steps/template.go
wking: I'm not saying it has semantic meaning, I'm just saying I would like to avoid unneccessary churn.  Maybe someone is diffing YAML or some such to track changes, etc.

pkg/steps/template.go
wking: Can you link me to that controller?

alvaroaleman: https://github.com/openshift/ci-ns-ttl-controller

Its currently based on the assumption that the existence of pods indicates that a namespace is in use, so that would need to be changed so that deletionTImeStamp != nil is considered as non-existent pod as well

pkg/steps/template.go
alvaroaleman: Ppl are not supposed to rely on implementation details like this and this is the only finalizer that could be there

wking: It's possible that admission hooks or other out-of-band actors add their own finalizers.  And yeah, while ideally folks keep track of which properties are semantically ordered and which are unordered, it's still cheap to avoid churning even the unordered properties.

wking:
Rebased around #1276 and bumped the unit tests with 7c7e7c17e -> ceefda704.

pkg/steps/template.go
wking: Makes sense, even if we don't land this finalizer PR: openshift/ci-ns-ttl-controller#7

pkg/steps/template.go
alvaroaleman: There is no ideally. Finalizer order must not matter. Its explicitly in the upstream docs. Please change this to simplify the code instead of writing complicated code to account for behavior that is explicitly documented as wrong.

wking:
Rebased around #1285 with ceefda704 -> c81e902f2.

alvaroaleman:
Not needed anymore, the pods themselves now upload their artifacts via decoration utils.
/close

PR #1282, opened 2020-10-08
blocking-issue-creator: avoid duplicate branch tokens

petr-muller:
When running with `--current-release=X --future-release=X` (rare state
when we do not have X+1 config prepared yet), we may end up with
duplicate tokens in the blocker issues:

```
Future Release Branches Frozen For Merging | branch:release-4.7 branch:release-4.7
```

### DIFF
--- a/cmd/blocking-issue-creator/main.go
+++ b/cmd/blocking-issue-creator/main.go
@@ -9,11 +9,12 @@ import (
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


petr-muller:
Example: https://github.com/openshift/console/issues/1433#event-3835825073

stevekuznetsov (approved):

PR #1296, opened 2020-10-13
ci-operator: allow users to inject the oc cli into test steps

stevekuznetsov:

This implements [DPTP-1242](https://issues.redhat.com/browse/DPTP-1242)

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -34,6 +34,12 @@ const (
 	SecretMountEnv = "SHARED_DIR"
 	// ClusterProfileMountEnv is the env we use to expose the cluster profile dir
 	ClusterProfileMountEnv = "CLUSTER_PROFILE_DIR"
+	// CliMountPath is where we mount the cli in a pod
+	CliMountPath = "/cli"
+	// CliEnv if the env we use to expose the path to the cli
+	CliEnv = "CLI_DIR"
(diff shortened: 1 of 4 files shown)


pkg/steps/multi_stage.go
```diff
@@ -34,6 +34,12 @@ const (
 	SecretMountEnv = "SHARED_DIR"
 	// ClusterProfileMountEnv is the env we use to expose the cluster profile dir
 	ClusterProfileMountEnv = "CLUSTER_PROFILE_DIR"
+	// CliMountPath is where we mount the cli in a pod
+	CliMountPath = "/cli"
+	// OcEnv if the env we use to expose the path to the cli
+	OcEnv = "OC"
+	// OcEnvValue is the value of OcEnv
+	OcEnvValue = CliMountPath + "/oc"
```
petr-muller, line 42: I wonder if we should try harder to place the `oc` to a path with a chance of being in `$PATH`? Exposing the `oc` only via the envvar would block us from seamlessly migrate existing `openshift_installer_src` tests to multi-stage (which I hoped would be possible), because the current workloads are unlikely to use `$OC`

stevekuznetsov:
```
 /bin/bash: line 2: oc: command not found 
```

:(

stevekuznetsov:
This should now be ready.

alvaroaleman (approved):

PR #1308, opened 2020-10-15
slack-bot: expose interaction flows in messages

stevekuznetsov:
Nobody wants to go to the "lightning bolt menu" and find the interaction
shortcut buttons for the bot, so now we can just mention the bot in a
conversation and it will respond with a button that starts a flow.

As always feel free to test it out on the robot-testing workspace.
Screenshot from 2020-10-14 17-50-20

### DIFF
--- a/pkg/slack/events/handler.go
+++ b/pkg/slack/events/handler.go
@@ -1,0 +1,93 @@
+package events
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/slack-go/slack/slackevents"
+)
+
+// Handler knows how to handle an event callback.
+type Handler interface {
+	Handle(callback *slackevents.EventsAPIEvent, logger *logrus.Entry) error
+	Identifier() string
+}
+
+type handler struct {
+	handle     func(callback *slackevents.EventsAPIEvent, logger *logrus.Entry) error
+	identifier string
+}
+
+func (h *handler) Handle(callback *slackevents.EventsAPIEvent, logger *logrus.Entry) error {
+	return h.handle(callback, logger)
+}
+func (h *handler) Identifier() string {
+	return h.identifier
+}
+
+// HandlerFunc returns a Handler for a handling func
+func HandlerFunc(identifier string, handle func(callback *slackevents.EventsAPIEvent, logger *logrus.Entry) error) Handler {
+	return &handler{
+		handle:     handle,
+		identifier: identifier,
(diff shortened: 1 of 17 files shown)


stevekuznetsov:
Needed to sort some test data ...

alvaroaleman:
@stevekuznetsov is the CD to the testing slack when we create PRs here?

pkg/slack/events/mention/mention.go
```diff
@@ -0,0 +1,137 @@
...
+
+type messagePoster interface {
+	PostMessage(channelID string, options ...slack.MsgOption) (string, string, error)
+}
+
+// Handler returns a handler that knows how to respond to
+// new messages that mention the robot by showing users
+// which interactive workflows they might be interested in,
+// based on the phrasing that they used to mention the bot.
+func Handler(client messagePoster) events.PartialHandler {
```
stevekuznetsov, line 29: I tried to test this method but two things made it impossible:

1. the `*slackevents.EventsAPIEvent` type is _incredibly_ dynamic and does not implement custom JSON marshalling, so there's no way to round-trip it with a fixture (it does not serialize)
2. the `slack.Client.PostMessage` function - the only thing we really want to test - takes all of it's important arguments in a functional options pattern, which uses anonymous functions operating over unexported types to achieve what we need, and validating that the right options were passed is basically impossible without incredibly ugly reflection and other nonsense

pkg/slack/events/handler.go
```diff
@@ -0,0 +1,93 @@
+package events
+
+import (
+	"github.com/sirupsen/logrus"
+	"github.com/slack-go/slack/slackevents"
+)
+
+// Handler knows how to handle an event callback.
+type Handler interface {
```
alvaroaleman, line 9: Whats the point of defining an interface instead of just exporting the concrete types if we don't have tests that implement their version of the interface?

stevekuznetsov: The caller should not infer anything from the individual handlers, at the routing level they should be identical and indistinguishable

alvaroaleman: But there is only one implementation or did I miss it?

stevekuznetsov: For now, yes. There will be many more as the bot learns how to respond to more events

alvaroaleman (approved):

PR #1259, opened 2020-09-24
pkg/steps/lease: add support for multiple leases

bbguimaraes:
Reintroduces https://github.com/openshift/ci-tools/pull/1126.

/hold

Includes/requires https://github.com/openshift/ci-tools/pull/1258.

PR #1311, opened 2020-10-15
ci-operator: add option for read-only shared dir

bbguimaraes:
Add a configuration option in multi-stage steps to not add the
`secret-wrapper` to the test process.  Results in noticeably faster
tests when a read-only version of the directory is sufficient.


stevekuznetsov (commented):
Seems fine to me

bbguimaraes:
Added a test for pod generation.  Execution will be tested in https://github.com/openshift/ci-tools/pull/1259.

Documented here: https://github.com/openshift/ci-docs/pull/16.

/hold cancel

bbguimaraes:

PR #1316, opened 2020-10-21
Migrate openshift_installer_src template

alvaroaleman:


cmd/determinize-ci-operator/main.go
```diff
@@ -34,6 +35,35 @@ func main() {
...
+				Test: []api.TestStep{{LiteralTestStep: &api.LiteralTestStep{
+					As:       "test",
+					From:     string(api.PipelineImageStreamTagReferenceSource),
+					Commands: test.Commands,
+					Cli:      api.LatestReleaseName,
+					Resources: api.ResourceRequirements{
+						Requests: api.ResourceList{"cpu": "100m"},
+					},
+				}}},
+				Workflow: &[]string{"ipi-gcp"}[0],
```
stevekuznetsov, line 59: wow

alvaroaleman: I learned that one from Sen

cmd/determinize-ci-operator/main.go
```diff
@@ -53,3 +73,44 @@ func main() {
...
+			Test: []api.TestStep{{LiteralTestStep: &api.LiteralTestStep{
+				As:       "test",
+				From:     string(api.PipelineImageStreamTagReferenceSource),
+				Commands: test.Commands,
+				Cli:      api.LatestReleaseName,
+				Resources: api.ResourceRequirements{
+					Requests: api.ResourceList{"cpu": "100m"},
+				},
+			}}},
+			Workflow: utilpointer.StringPtr(fmt.Sprintf("ipi-%s", clusterProfile)),
```
petr-muller, line 106: There's this stupid thing where cluster profile is called `azure4`  and the corresponding workflow is `ipi-azure`, so we may need a table for this :(

alvaroaleman: Fixed

cmd/determinize-ci-operator/main.go
```diff
@@ -2,21 +2,36 @@ package main
...
+	templateMigrationCeiling                int
+	templateMigrationAllowedBranches        flagutil.Strings
+	templateMigrationAllowedOrgs            flagutil.Strings
+	templateMigrationAllowedClusterProfiles flagutil.Strings
+}
+
+func gatherOptions() options {
+	o := options{}
+	o.Bind(flag.CommandLine)
+	flag.BoolVar(&o.migrateOpenshiftInstallerSRCTemplates, "migrate-openshift-installer-src-templates", false, "If tests based on the openshift_installer_src template should be migrated to multi-step tests. What tests exactly to migrate can be limited further via the template-migration-allowed-{branch,org,cluster-profile} flags")
```
petr-muller, line 28: Can we future-proof this a *tiny bit* and make it `--migrate-templates=openshift-installer-src,...` so this is easy to extend when we do the remaining ones?

petr-muller (approved):
/hold

LGTM as is, with a suggestion that I'm not insisting on

cmd/determinize-ci-operator/main.go
alvaroaleman: fixed

cmd/determinize-ci-operator/main.go
```diff
@@ -34,6 +69,10 @@ func main() {
 			return nil
 		}
 
+		if sets.NewString(o.enabledTemplateMigrations.Strings()...).Has(openshiftInstallerSRCTemplateName) && migratedCount <= o.templateMigrationCeiling {
+			migratedCount += migrateOpenshiftInstallerSRCTemplates(&output, sets.NewString(o.templateMigrationAllowedBranches.Strings()...), sets.NewString(o.templateMigrationAllowedOrgs.Strings()...), sets.NewString(o.templateMigrationAllowedClusterProfiles.Strings()...))
```
stevekuznetsov, line 73: might make sense to upstream a `.Set()`

cmd/determinize-ci-operator/main.go
```diff
@@ -53,3 +92,52 @@ func main() {
...
+	allowedOrgs sets.String,
+	allowedCloudproviders sets.String,
+) (migratedCount int) {
+	if (len(allowedBranches) != 0 && !allowedBranches.Has(configuration.Info.Branch)) || (len(allowedOrgs) != 0 && !allowedOrgs.Has(configuration.Info.Org)) {
+		return 0
+	}
+
+	for idx, test := range configuration.Configuration.Tests {
+		if test.OpenshiftInstallerSrcClusterTestConfiguration == nil ||
+			(len(allowedCloudproviders) != 0 && !allowedCloudproviders.Has(string(test.OpenshiftInstallerSrcClusterTestConfiguration.ClusterProfile))) {
```
stevekuznetsov, line 108: are we going to run into an issue when we allow N templates but the full cross product with M cloud providers is not supported and we need allowed providers per template?

cmd/determinize-ci-operator/main.go
```diff
@@ -53,3 +92,52 @@ func main() {
...
+		test.OpenshiftInstallerSrcClusterTestConfiguration = nil
+		test.MultiStageTestConfiguration = &api.MultiStageTestConfiguration{
+			ClusterProfile: clusterProfile,
+			Test: []api.TestStep{{LiteralTestStep: &api.LiteralTestStep{
+				As:       "test",
+				From:     string(api.PipelineImageStreamTagReferenceSource),
+				Commands: test.Commands,
+				Cli:      api.LatestReleaseName,
+				Resources: api.ResourceRequirements{
+					Requests: api.ResourceList{"cpu": "100m"},
```
stevekuznetsov, line 122: Is this what the template sets right now?

cmd/determinize-ci-operator/main.go
alvaroaleman: yeah, already did that last week, updated the PR

cmd/determinize-ci-operator/main.go
alvaroaleman: We do have this issue already, because for example there is no `ipi-vsphere`. The purpose of this PR is to provide an automatic migration and flags to limit it so we won't be ending up doing `git add -p`, but actually doing the migration is a follow-up

cmd/determinize-ci-operator/main.go
alvaroaleman: It doesn't set anything at all, so it was a pretty random guess: ci-operator/templates/openshift/installer/cluster-launch-installer-src.yaml

stevekuznetsov (approved):

PR #1318, opened 2020-10-21
Secretwrapper: Create a writeable kubeconfig

alvaroaleman:
To avoid issues like [this](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_release/12990/rehearse-12990-pull-ci-openshift-hive-master-e2e-gcp/1318994575771045888#1:build-log.txt%3A70) where ppl jobs fail because they try to modify a read-only kubeconfig

### DIFF
--- a/cmd/secret-wrapper/main.go
+++ b/cmd/secret-wrapper/main.go
@@ -5,7 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -14,6 +14,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	coreapi "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -98,6 +99,9 @@ func (o *options) run() error {
 		return fmt.Errorf("failed to copy secret mount: %w", err)
 	}
 	var errs []error
+	if err := createWriteableKubeconfig(); err != nil {
+		errs = append(errs, fmt.Errorf("failed to create writeable kubeconfig: %w", err))
+	}
 	if err := execCmd(o.cmd); err != nil {
 		errs = append(errs, fmt.Errorf("failed to execute wrapped command: %w", err))
 	}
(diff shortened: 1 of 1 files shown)


bbguimaraes:
Can we add a [test](https://github.com/openshift/ci-tools/blob/master/test/secret-wrapper-integration.sh)?

petr-muller:
So until now I was completely oblivious to how `secretwrapper` works... Will modifications to $KUBECONFIG survive to between steps? We should also document this in https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables

bbguimaraes:
> So until now I was completely oblivious to how secretwrapper works... 

It's very simple:

- copy the contents of the secret mount to a writable location
- update `SHARED_DIR` to point there
- execute the test command
- update the secret with the new contents of `SHARED_DIR`

> Will modifications to $KUBECONFIG survive between steps?

No, because the file is placed outside of `$SHARED_DIR`.

> We should also document this in https://docs.ci.openshift.org/docs/architecture/step-registry/#available-environment-variables

I think it already describes the interface well enough:

`${KUBECONFIG} Path to system:admin credentials for the ephemeral OpenShift cluster under test.`

(also: https://docs.ci.openshift.org/docs/architecture/step-registry/#a-note-on-kubeconfig)

What is missing?

petr-muller:
> It's very simple:

I guess what always confused me is how this is presented in the logs where `secret-wrapper` is said to execute first and finish before the actual test command.

```
2020/10/21 20:08:45 Container cp-secret-wrapper in pod e2e-gcp-ipi-deprovision-deprovision completed successfully
2020/10/21 20:13:56 Container test in pod e2e-gcp-ipi-deprovision-deprovision completed successfully 
```

> No, because the file is placed outside of $SHARED_DIR.

Yes, that's what I thought. Is this desirable? It's what the users currently have, so if they want the changes to persist, they need to put the kubeconfig to `$SHARED_DIR`, but they know they *need* to do this because they also create the writable one, so there's the expected symmetry. I think that providing writable config without persisting it ourselves will be confusing for users unless we really point that behaviour out.

> What is missing? 


I'm sorry for that unbound "this"  - what I meant by "this" is the new feature brought by this PR.

What is missing are the expectations for what can be done with the kubeconfig. Right now, it is missing the information that the file is read only. After this PR merges, it will miss the information that the file is writable, but modifications will be lost between steps.

alvaroaleman:
> I guess what always confused me is how this is presented in the logs where secret-wrapper is said to execute first and finish before the actual test command.

Its `cp-secret-wrapper`, not `secret-wrapper` that finishes first :P

>  I think that providing writable config without persisting it ourselves will be confusing for users unless we really point that behaviour out.

So from what I saw the usecase for the writeable kubeconfig is to configure a namespace on it. I would assume its okay if we don't persist that. One issue with persisting it that it would allow users to break the kubeconfig.

petr-muller:
> Its cp-secret-wrapper, not secret-wrapper that finishes first :P

Ooooooooooooooo now I finally get it :facepalm: 

> So from what I saw the usecase for the writeable kubeconfig is to configure a namespace on it. I would assume its okay if we don't persist that. One issue with persisting it that it would allow users to break the kubeconfig.

Yeah, that's totally reasonable: `$KUBECONFIG` will be writable inside step, changes will not persist between steps, so next steps are guaranteed to receive working credentials, if you need persistence, DIY. Works for me, but IMO needs a sentence or two in docs.

/lgtm
/hold
(hold for Bruno and the convo in Slack)

bbguimaraes:
LGTM, but I still think it would be good to have a test.

alvaroaleman:
Superseded by https://github.com/openshift/ci-tools/pull/1323

PR #1319, opened 2020-10-22
[ci-secret-bootstrap] add the ability to validate the usage of the bitwarden items

droslean:
This commit includes:

- A way how to compare the bitwarden vs config items
- flag `--validate-bitwarden-items-usage` to enable the validation
- flag `--bw-allow-unused` to allow bitwarden items to existing without being used
- unittests

[DPTP-1547](https://issues.redhat.com/browse/DPTP-1547)

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -19,15 +19,19 @@ import (
 	"golang.org/x/sync/semaphore"
 
 	coreapi "k8s.io/api/core/v1"
+
 	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubejson "k8s.io/apimachinery/pkg/runtime/serializer/json"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+
 	"k8s.io/client-go/kubernetes/scheme"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+
+	"k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/logrusutil"
 
 	"github.com/openshift/ci-tools/pkg/api/secretbootstrap"
@@ -36,27 +40,32 @@ import (
 )
 
 type options struct {
-	dryRun         bool
-	kubeConfigPath string
-	configPath     string
-	bwUser         string
-	bwPasswordPath string
-	cluster        string
-	force          bool
-	logLevel       string
+	dryRun               bool
+	force                bool
+	validateBWItemsUsage bool
 
+	kubeConfigPath  string
+	configPath      string
+	bwUser          string
+	bwPasswordPath  string
+	cluster         string
+	logLevel        string
 	impersonateUser string
-
-	bwPassword     string
-	secretsGetters map[string]coreclientset.SecretsGetter
-	config         secretbootstrap.Config
+	bwPassword      string
 
 	maxConcurrency int
+
+	secretsGetters map[string]coreclientset.SecretsGetter
+	config         secretbootstrap.Config
+	bwAllowUnused  flagutil.Strings
 }
 
 func parseOptions() options {
 	var o options
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	o.bwAllowUnused = flagutil.NewStrings()
+	fs.Var(&o.bwAllowUnused, "bw-allow-unused", "One or more bitwarden items that will be ignored when the --validate-bitwarden-items-usage is specified")
+	fs.BoolVar(&o.validateBWItemsUsage, "validate-bitwarden-items-usage", false, "If set, the tool only validates if all attachments and custom fields that exist in BitWarden are being used in the given config.")
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
 	fs.StringVar(&o.kubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
 	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
@@ -87,6 +96,9 @@ func (o *options) validateOptions() error {
 	}
 	if o.configPath == "" {
 		return fmt.Errorf("--config is empty")
+	}
+	if len(o.bwAllowUnused.Strings()) > 0 && !o.validateBWItemsUsage {
+		return fmt.Errorf("--bw-allow-unused must be specified with --validate-bitwarden-items-usage")
 	}
 	return nil
 }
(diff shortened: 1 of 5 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -457,6 +469,142 @@ func writeSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
...
+	return bwComparableItemsByName
+}
+
+func costructConfigItemsByName(config secretbootstrap.Config) map[string]*comparable {
+	cfgComparableItemsByName := make(map[string]*comparable)
+
+	for _, cfg := range config.Secrets {
+		for _, bwContext := range cfg.From {
+			if bwContext.BWItem != "" {
+				item := &comparable{
```
hongkailiu, line 523: NIT: init `item` only when needed.

```suggestion
				item, ok := cfgComparableItemsByName[bwContext.BWItem]
				if !ok {
					item = &comparable{
						fields:      sets.NewString(),
						attachments: sets.NewString(),
					}
					cfgComparableItemsByName[bwContext.BWItem] = item
				}
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -457,6 +469,142 @@ func writeSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
...
+			comparableItem.hasPassword = true
+		}
+
+		bwComparableItemsByName[name] = comparableItem
+	}
+
+	return bwComparableItemsByName
+}
+
+func costructConfigItemsByName(config secretbootstrap.Config) map[string]*comparable {
```
hongkailiu, line 517:
```suggestion
func constructConfigItemsByName(config secretbootstrap.Config) map[string]*comparable {
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -457,6 +469,142 @@ func writeSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
...
+
+				if bwContext.Attribute == secretbootstrap.AttributeTypePassword {
+					item.hasPassword = true
+				}
+				cfgComparableItemsByName[bwContext.BWItem] = item
+			}
+
+			if len(bwContext.DockerConfigJSONData) > 0 {
+				for _, context := range bwContext.DockerConfigJSONData {
+					item := &comparable{
```
hongkailiu, line 543: Same here.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -480,30 +628,54 @@ func main() {
 			logrus.WithError(err).Fatal("Failed to logout.")
 		}
 	})
-	defer logrus.Exit(0)
```
hongkailiu, line 483: Is it moved somewhere?
I believe we still need it.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -480,30 +628,54 @@ func main() {
...
-		if err != nil {
-			logrus.WithError(err).Fatal("failed to create tempfile")
-		}
-		defer tmpFile.Close()
-		logrus.Infof("Dry-Run enabled, writing secrets to %s", tmpFile.Name())
-		if err := writeSecrets(secretsMap, tmpFile); err != nil {
-			errs = append(errs, fmt.Errorf("failed to write secrets on dry run: %w", err))
+
+	// if validate-bitwarden-items-usage is specified we only want to check if all the
+	// bitwarden items are being used in the given configuration.
```
hongkailiu, line 641: Should we make a new binary for this function since there is no logic overlapping with the existing tool?

cmd/ci-secret-bootstrap/main.go
droslean: Well, this change here is not related with my PR. However, this line can make the tool to exit 0 when the code panics and also the panic message is not even shown. This is why I removed it. Why we needed this from the first place?

cmd/ci-secret-bootstrap/main.go
droslean: Well this tool does already what we need here. Also the config file belongs here so it makes more sense to make the tool validate its own config

pkg/bitwarden/bitwarden.go
```diff
@@ -35,6 +37,7 @@ type Item struct {
 // Client is used to communicate with BitWarden
 type Client interface {
 	GetFieldOnItem(itemName, fieldName string) ([]byte, error)
+	GetSavedItems() []Item
```
alvaroaleman, line 40: There are no `unsaved` items, so maybe just `GetAllItems`?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -457,6 +469,132 @@ func writeSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
...
+		if len(item.Attachments) > 0 {
+			attachments := sets.NewString()
+			for _, attachment := range item.Attachments {
+				attachments.Insert(attachment.FileName)
+			}
+
+			comparableItem.attachments = attachments
+		}
+
+		if item.Login != nil {
```
alvaroaleman, line 501: I think the check needs to be `item.Login != nil && item.Login.Password != ""`

cmd/ci-secret-bootstrap/main.go
```diff
@@ -457,6 +469,132 @@ func writeSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
...
+			if len(bwContext.DockerConfigJSONData) > 0 {
+				for _, context := range bwContext.DockerConfigJSONData {
+					item, ok := cfgComparableItemsByName[context.BWItem]
+					if !ok {
+						item = &comparable{
+							fields:      sets.NewString(),
+							attachments: sets.NewString(),
+						}
+					}
+					item.fields.Insert(context.RegistryURLBitwardenField)
```
alvaroaleman, line 542: I think we need to check if this is an empty string first

cmd/ci-secret-bootstrap/main.go
```diff
@@ -480,30 +618,54 @@ func main() {
...
+			defer tmpFile.Close()
+			logrus.Infof("Dry-Run enabled, writing secrets to %s", tmpFile.Name())
+			if err := writeSecrets(secretsMap, tmpFile); err != nil {
+				errs = append(errs, fmt.Errorf("failed to write secrets on dry run: %w", err))
+			}
+		} else {
+			if err := updateSecrets(o.secretsGetters, secretsMap, o.force); err != nil {
+				errs = append(errs, fmt.Errorf("failed to update secrets: %w", err))
+			}
+			logrus.Info("Updated secrets.")
```
alvaroaleman, line 665: @droslean can you elaborate a bit about what you imagine when and by whom this will be used?

IMHO it would make sense to both update and check, so we can do this in the same CI job. And the tool definitely needs to error out if it finds unused items

cmd/ci-secret-bootstrap/main.go
droslean: I think that sets library won't insert anything if the string is empty. I will double check though.

cmd/ci-secret-bootstrap/main.go
hongkailiu: Ack. You can mark this as resolved.

cmd/ci-secret-bootstrap/main.go
droslean: Well, since we already have a periodic job, there is no need to add a new one and use the same tool with different args. So what you are suggesting makes total sense to me. I will update the tool accordingly.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -36,27 +40,32 @@ import (
...
-
-	maxConcurrency int
+	bwAllowItems   flagutil.Strings
 }
 
 func parseOptions() options {
 	var o options
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	o.bwAllowItems = flagutil.NewStrings()
+	fs.Var(&o.bwAllowItems, "bw-allow-items", "The items that will be ignored when the --validate-bitwarden-items-usage is specified")
```
alvaroaleman, line 67: Nit: Rename to `bw-allow-unused`? Also, please mention in the docs that it can be specified multiple times

cmd/ci-secret-bootstrap/main.go
```diff
@@ -457,6 +469,139 @@ func writeSecrets(secretsMap map[string][]*coreapi.Secret, w io.Writer) error {
...
+
+		if item.hasPassword && !cfgComparableItemsByName[bwName].hasPassword {
+			if _, ok := unused[bwName]; !ok {
+				unused[bwName] = &comparable{}
+			}
+			itemLogger = itemLogger.WithField("login", "password")
+			unused[bwName].hasPassword = true
+		}
+
+		itemLogger.Info("Unused item")
```
alvaroaleman, line 599: That needs to be `Error`

cmd/ci-secret-bootstrap/main.go
```diff
@@ -488,6 +633,14 @@ func main() {
 	if err != nil {
 		errs = append(errs, err)
 	}
+
+	if o.validateBWItemsUsage {
+		unusedItems := getUnusedBWItems(o.config, bwClient, sets.NewString(o.bwAllowItems.Strings()...))
+		if len(unusedItems) > 0 {
+			logrus.Fatal("unused Bitwarden items found.")
```
alvaroaleman, line 640: Please append to `errs` instead, so we will only fail the job after syncing everything

alvaroaleman (approved):

stevekuznetsov (approved):

PR #1323, opened 2020-10-23
Steps: Provide a writable kubeconfig location

alvaroaleman:
Supersedes https://github.com/openshift/ci-tools/pull/1318

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -39,7 +39,10 @@ const (
 	// CliEnv if the env we use to expose the path to the cli
 	CliEnv = "CLI_DIR"
 	// CommandPrefix is the prefix we add to a user's commands
-	CommandPrefix = "#!/bin/bash\nset -eu\n"
+	CommandPrefix = "#!/bin/bash\n" +
+		"set -eu\n" +
+		// provide a writeable kubeconfig so ppl can for example set a namespace, but do not pass it on to subsequent steps to limit the amount of possible footshooting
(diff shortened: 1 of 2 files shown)


pkg/steps/multi_stage.go
```diff
@@ -39,7 +39,10 @@ const (
 	// CliEnv if the env we use to expose the path to the cli
 	CliEnv = "CLI_DIR"
 	// CommandPrefix is the prefix we add to a user's commands
-	CommandPrefix = "#!/bin/bash\nset -eu\n"
+	CommandPrefix = "#!/bin/bash\n" +
+		"set -eu\n" +
+		// provide a writeable kubeconfig so ppl can for example set a namespace, but do not pass it on to subsequent steps to limit the amount of possible footshooting
+		"if [[ -e ${KUBECONFIG:-} ]]; then WRITEABLE_KUBECONFIG_LOCATION=$(mktemp) && cp $KUBECONFIG $WRITEABLE_KUBECONFIG_LOCATION && export KUBECONFIG=$WRITEABLE_KUBECONFIG_LOCATION; fi\n"
```
bbguimaraes, line 45: This could use a sub-shell to avoid setting the variable globally:

```sh
if [[ -e "${KUBECONFIG:-}" ]]; then
    KUBECONFIG=$(f=$(mktemp) && cp "$KUBECONFIG" "$f" && echo "$f")
fi
```

alvaroaleman: @bbguimaraes is a mostly unused env var enough of a problem to go through the overhead of creating a subshell to avoid it? We could also `unset` but I am not sure we should bother

bbguimaraes: I don't think the performance of either is significant.  It is more about making this transparent to the test script.  If we can not touch the environment, I think that is preferable.

alvaroaleman: updated

PR #1331, opened 2020-10-26
slack-bot: add context to job links

stevekuznetsov:
When a user pastes a link to a job or artifact, we often want to see the
full Spyglass link and the ci-operator configuration that gave rise to
the job in the first place. Instead of asking them for this information
we can just calculate it and post a reply.

### DIFF
--- a/pkg/slack/events/joblink/link.go
+++ b/pkg/slack/events/joblink/link.go
@@ -1,0 +1,387 @@
+package joblink
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"net/url"
+	"strconv"
+	"strings"
+
+	"cloud.google.com/go/storage"
+	"github.com/GoogleCloudPlatform/testgrid/util/gcs"
+	"github.com/sirupsen/logrus"
+	"github.com/slack-go/slack"
+	"github.com/slack-go/slack/slackevents"
+	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/gcsupload"
+	"k8s.io/test-infra/prow/pjutil"
+	"k8s.io/test-infra/prow/pod-utils/downwardapi"
+	gcsutil "k8s.io/test-infra/prow/pod-utils/gcs"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/jobconfig"
+	"github.com/openshift/ci-tools/pkg/rehearse"
+	"github.com/openshift/ci-tools/pkg/slack/events"
+)
+
+// job abstracts the types of Prow jobs. Only one field can be non-nil
+type job struct {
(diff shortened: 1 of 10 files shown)


pkg/slack/events/joblink/link.go
```diff
@@ -0,0 +1,389 @@
...
+
+// NewJobGetter exposes a simplified interface from the Prow configuration
+func NewJobGetter(config config.Getter) JobGetter {
+	return &getter{
+		config: config,
+	}
+}
+
+type getter struct {
+	config config.Getter
```
alvaroaleman, line 53: Nit: You can alias the config.Getter type and define methods on the alias

pkg/slack/events/joblink/link.go
```diff
@@ -0,0 +1,389 @@
...
+		Repo:    parts[1],
+		Branch:  rehearse.BranchFromRegexes(brancher.Branches),
+		Variant: rehearse.VariantFromLabels(job.Labels),
+	}
+}
+
+func (g *getter) JobForName(name string) *job {
+	var found job
+
+	for orgRepo, presubmits := range g.config().PresubmitsStatic {
```
alvaroaleman, line 72: I don't think the name is enough, I think we have the same name on different branches

pkg/slack/events/joblink/link.go
```diff
@@ -0,0 +1,389 @@
...
+	case "prow.ci.openshift.org":
+		switch strings.Split(url.Path, "/")[1] {
+		case "job-history":
+			return infoForJobHistory(url)
+		case "log":
+			return infoForJobLog(url)
+		case "view":
+			return infoForJobView(url)
+		}
+	case "gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com":
```
alvaroaleman, line 319: can we add this url to the api package so we don't break the tool when we give it a nicer one?

pkg/slack/events/joblink/link.go
```diff
@@ -0,0 +1,389 @@
...
+				} `json:"elements"`
+			} `json:"elements"`
+		} `json:"blocks"`
+	}
+	var data blocks
+	fmt.Println(string(*raw.InnerEvent))
+	if err := json.Unmarshal(*raw.InnerEvent, &data); err != nil {
+		return nil, errors.New("could not get blocks from event data")
+	}
+	fmt.Println(data)
```
bbguimaraes, line 289: Debug statements.

pkg/slack/events/joblink/link.go
stevekuznetsov: Not for generated jobs. I'm pretty sure name is unique in our config.

pkg/slack/events/joblink/link.go
stevekuznetsov: :exploding_head:

stevekuznetsov: `Cannot define methods on non-local type config.Getter`

stevekuznetsov:
@alvaroaleman @bbguimaraes updated

alvaroaleman (approved):

PR #1342, opened 2020-10-28
secret-wrapper: upload $KUBECONFIG when possible

stevekuznetsov:
Run a best-effort attempt to upload the $KUBECONFIG when we notice that
it exists. Re-try upload failures forever, expecting that the process
will exit when needed and we should attempt to do the upload until then.

/assign @bbguimaraes

### DIFF
--- a/cmd/secret-wrapper/main.go
+++ b/cmd/secret-wrapper/main.go
@@ -6,11 +6,15 @@ import (
 	"fmt"
 	"io"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"log"
 	"os"
 	"os/exec"
 	"os/signal"
+	"path"
 	"path/filepath"
 	"syscall"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	coreapi "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
Part of DPTP-1668

cmd/secret-wrapper/main.go
```diff
@@ -202,3 +207,23 @@ func createSecret(client coreclientset.SecretInterface, name, dir string, dry bo
...
 }
+
+// uploadKubeconfig will do a best-effort attempt at uploading a kubeconfig
+// file if one does not exist at the time we start running but one does get
+// created while executing the command
+func uploadKubeconfig(client coreclientset.SecretInterface, name, dir string, dry bool) {
+	if _, err := os.Stat(path.Join(dir, "kubeconfig")); err == nil {
+		// kubeconfig already exists, no need to do anything
+		return
+	}
```
bbguimaraes, line 218: This could checked before launching the go routine.

cmd/secret-wrapper/main.go
```diff
@@ -202,3 +207,23 @@ func createSecret(client coreclientset.SecretInterface, name, dir string, dry bo
...
+	if err := wait.PollInfinite(time.Second, func() (done bool, err error) {
+		if _, err := os.Stat(path.Join(dir, "kubeconfig")); err != nil {
+			return false, nil
+		}
+		// kubeconfig exists, we can upload it
+		uploadErr := createSecret(client, name, dir, dry)
```
bbguimaraes, line 224: Can this race with the call in `run`?

bbguimaraes: Also: is `kubeconfig` written atomically or can it happen that an intermediary state is used?

cmd/secret-wrapper/main.go
stevekuznetsov: Yes, but as written it keeps things self-contained in this method - any downside to starting a Goroutine that exits quickly?

cmd/secret-wrapper/main.go
stevekuznetsov: The race would be:

 - process A reads secret from APIServer
 - process B reads data from APIServer
 - process B updates data on APIServer
 - process A fails to update data on APIServer (resource version too old)

The only case in which this could potentially be bad is if process B read the data from disk before process A did, and had a stale view of that data but managed to update the APIServer. I can add a mutex to serialize updates but I am struggling to think of a way the race would end up with a bad state in the secret.

re: atomic ... I have no idea, but I assume yes.

stevekuznetsov: Checked, it is atomic

stevekuznetsov: fixed the race

PR #1343, opened 2020-10-28
WIP: cmd/ci-operator-configresolver: Expose step API

wking:
For simple forms of [the header][1], so folks can use:

```console
$ curl -H Accept:application/json https://steps.ci.openshift.org/reference/openshift-e2e-test
```

and similar to get a JSON response with the rendered metadata.  This makes the structures we pass into the templates an official API, which is probably a tighter binding between the HTML and JSON renderings than we really want.  But in practice I expect the structures are stable enough to be an internal API, and it avoids the overhead of trying to design more formal APIs for the different pages.

Unit tests would be nice, but there aren't any in place today for the HTML templates.  We'd need to stub in a step config or something, and I haven't bothered with that yet.

[1]: https://tools.ietf.org/html/rfc7231#section-5.3.2

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -8,6 +8,8 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"path"
+	"strings"
 	"time"
 
 	"github.com/openshift/ci-tools/pkg/api"
@@ -20,6 +22,7 @@ import (
 	"k8s.io/test-infra/prow/pjutil"
 	"k8s.io/test-infra/prow/simplifypath"
 
+	"github.com/openshift/ci-tools/pkg/load"
 	"github.com/openshift/ci-tools/pkg/load/agents"
 	"github.com/openshift/ci-tools/pkg/webreg"
 )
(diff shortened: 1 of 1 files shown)


wking:
@stevekuznetsov says that config.ci is for the API and steps.ci is just for HTML docs and will not grow support for Accept or other response media types.  Within that restriction, I've pushed 1f94620c3 -> f532531a2 to shift this handling out of `pkg/webreg` and into `cmd/ci-operator-configresolver`.  I've copy/paste/edited to bring over a bunch of path handling and other scaffolding to maintain as much similarity between the two path structures as possible, so folks can easily find the API for a particular reference, chain, workflow, etc. without having to learn a new path structure.  WIP because a bunch of scaffolding (e.g. queries, chain/job/etc. listing, ...) is commented out while we kick around the general approach.

cmd/ci-operator-configresolver/main.go
```diff
@@ -158,20 +161,101 @@ func resolveAndRespond(registryAgent agents.RegistryAgent, config api.ReleaseBui
...
+}
+
+func referenceHandler(agent agents.RegistryAgent, w http.ResponseWriter, req *http.Request) {
+	start := time.Now()
+	defer func() { logrus.Infof("rendered in %s", time.Since(start)) }()
+	name := path.Base(req.URL.Path)
+
+	refs, _, _, docs, metadata := agent.GetRegistryComponents()
+	if _, ok := refs[name]; !ok {
+		writeErrorPage(w, fmt.Errorf("Could not find reference `%s`. If you reached this page via a link provided in the logs of a failed test, the failed step may be a literal defined step, which does not exist in the step registry. Please look at the job info page for the failed test instead.", name), http.StatusNotFound)
```
stevekuznetsov, line 234: This doesn't port because nothing links the config

cmd/ci-operator-configresolver/main.go
```diff
@@ -244,6 +328,7 @@ func main() {
 	http.HandleFunc("/", handler(http.HandlerFunc(http.NotFound)).ServeHTTP)
 	http.HandleFunc("/config", handler(resolveConfig(configAgent, registryAgent)).ServeHTTP)
 	http.HandleFunc("/resolve", handler(resolveLiteralConfig(registryAgent)).ServeHTTP)
+	http.HandleFunc("/step-registry", handler(resolveStepRegistry(registryAgent, configAgent)).ServeHTTP)
```
stevekuznetsov, line 331: could just be

```
/step
/chain
/workflow
```
?

cmd/ci-operator-configresolver/main.go
```diff
@@ -158,20 +161,101 @@ func resolveAndRespond(registryAgent agents.RegistryAgent, config api.ReleaseBui
...
+
+func resolveStepRegistry(regAgent agents.RegistryAgent, confAgent agents.ConfigAgent) http.HandlerFunc {
+	return func(w http.ResponseWriter, req *http.Request) {
+		trimmedPath := strings.TrimPrefix(req.URL.Path, req.URL.Host)
+		// remove leading path prefix
+		trimmedPath = strings.TrimPrefix(trimmedPath, "/step-registry/")
+		// remove trailing slash
+		trimmedPath = strings.TrimSuffix(trimmedPath, "/")
+		splitURI := strings.Split(trimmedPath, "/")
+		if len(splitURI) == 1 {
```
stevekuznetsov, line 197: if we did e.g. `GET /step/{name}` and the like for `chain` this could be simpler

stevekuznetsov (commented):
mostly lgtm

cmd/ci-operator-configresolver/main.go
wking: could be, but then, so could pkg/webreg, right?  Keeping this nested allows us to copy/paste the prefix handling out of pkg/webreg if/when that evolves so we keep the path structure consistent between the two hosts for these resources.

cmd/ci-operator-configresolver/main.go
wking: this is just copy/pasted out of pkg/webreg.  If the implementation there improves in some future PR, I can copy/paste the improved logic in here to pick it up.

cmd/ci-operator-configresolver/main.go
wking: Dropped `If you reached...` with f532531a2 -> e80cd5572.

cmd/ci-operator-configresolver/main.go
stevekuznetsov: but these are separate APIs doing separate things. there will never be a search page here ...

cmd/ci-operator-configresolver/main.go
stevekuznetsov: and the job config is already not consistent. i don't see any utilty in this

cmd/ci-operator-configresolver/main.go
wking: Folks won't want to search for jobs and such?

cmd/ci-operator-configresolver/main.go
wking: job config?  I'm just trying to minimize divergence from the existing path structure that webreg uses to identify config subsections.  Are you concerned about `jobHandler` being wired up to a different place in this config.ci host (other than the `/step-registry` namespacing)?

PR #1351, opened 2020-10-29
Step registry ui: Add handler for ci-operator config reference

alvaroaleman:
The content for the handler is pre-generated because the generation needs the source code which we do not have anymore in the final image.

### DIFF
--- a/pkg/webreg/zz_generated.ci_operator_reference.go
+++ b/pkg/webreg/zz_generated.ci_operator_reference.go
@@ -1,0 +1,1210 @@
+package webreg
+
+const ciOperatorReferenceYaml = "# The list of base images describe" +
+	"# which images are going to be necessary outside" +
+	"# of the pipeline. The key will be the alias that other" +
+	"# steps use to refer to this image." +
+	"base_images:" +
+	"    \"\":" +
+	"        # As is an optional string to use as the intermediate name for this reference." +
+	"        as: ' '" +
+	"        name: ' '" +
+	"        namespace: ' '" +
+	"        tag: ' '" +
+	"# BaseRPMImages is a list of the images and their aliases that will" +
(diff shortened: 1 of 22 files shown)


pkg/webreg/zz_generated.ci_operator_reference.go
```diff
@@ -0,0 +1,1210 @@
...
+	"                namespace: ' '\n" +
+	"                tag: ' '\n" +
+	"            optional_on_success: false\n" +
+	"            # Reference is the name of a step reference.\n" +
+	"            ref: \"\"\n" +
+	"            # Resources defines the resource requirements for the step.\n" +
+	"            resources:\n" +
+	"                # LiteralTestStep is a full test step definition.\n" +
+	"                limits:\n" +
+	"                    # LiteralTestStep is a full test step definition.\n" +
```
stevekuznetsov, line 1196: This don't look quite right

alvaroaleman: How so?

stevekuznetsov: This string "LiteralTestStep..." is repeated like 400 times on unrelated fields in here

stevekuznetsov:
mechanisms for generating and serving this LGTM but the actual content has bugs

/lgtm

pkg/webreg/zz_generated.ci_operator_reference.go
alvaroaleman: Hm, that doc comes from a promoted field: https://github.com/openshift/ci-tools/blob/581ca445940d60c43bce33451236e3ce8d134cab/pkg/api/types.go#L650

Seems like a bug in genyaml. Do you mind merging this already withotu advertising it and I will look into the genyaml stuff later?

PR #1356, opened 2020-10-30
Don't automount ServiceAccount token in pods

alvaroaleman:
This is not only not needed, it may create issues when it interacts with
the OC command by for example setting a namespace that doesn't exist in
the target cluster.


alvaroaleman:
Closing for now, ppl might use it even though that was never intended to happen, so this would need some cleanup efforts afterwards

PR #1357, opened 2020-10-30
ci-operator: allow configuring observers, serve their config

stevekuznetsov:

Part of [DPTP-1668](https://issues.redhat.com/browse/DPTP-1668)

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -555,6 +555,20 @@ type RegistryWorkflow struct {
 	Documentation string `json:"documentation,omitempty"`
 }
 
+// RegistryObserverConfig is the struct that observer configs are unmarshalled into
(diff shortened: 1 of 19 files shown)


pkg/api/types.go
```diff
@@ -602,6 +638,8 @@ type LiteralTestStep struct {
 	// Cli is the (optional) name of the release from which the `oc` binary
 	// will be injected into this step.
 	Cli string `json:"cli,omitempty"`
+	// Observers are the observers that should be running
```
alvaroaleman, line 641: Nit: It would be nice to write one sentence about what an oberver is here

stevekuznetsov: I expect to add an architecture page to the docs site for this

PR #1358, opened 2020-10-30
Fix verify-gen to correctly fail in CI

alvaroaleman:


pkg/steps/run.go
```diff
@@ -3,12 +3,11 @@ package steps
 import (
 	"context"
 	"errors"
-	"sync"
-	"time"
-
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/junit"
 	"github.com/openshift/ci-tools/pkg/results"
+	"sync"
```
stevekuznetsov, line 9: that dont look so good

Makefile
```diff
@@ -193,7 +192,8 @@ generate:
 
 .PHONY: verify-gen
 verify-gen: generate
-	@if !(git diff --quiet HEAD); then \
+	@# Don't add --quiet here, it disables --exit code in the git 1.7 we have in CI, making this unusuable
+	if  ! git diff --exit-code &>/dev/null; then \
 		git diff; \
 		echo "generated files are out of date, run make generate"; exit 1; \
 	fi
```
bbguimaraes, line 199: Why not just

```sh
	if  ! git diff --exit-code; then \
		echo "generated files are out of date, run make generate"; exit 1; \
	fi
```

?

alvaroaleman: done

droslean (approved):

stevekuznetsov (approved):

PR #1359, opened 2020-10-30
pkg/testhelper: mark as helper, include filename

bbguimaraes:
- mark `CompareWithFixture` as a test helper so failure messages display
  the true origin
- include the fixture file name in the diff to make it a real patch


stevekuznetsov (approved):

stevekuznetsov:
Part of DPTP-1365

PR #1362, opened 2020-10-30
pkg/defaults: give each test its own parameters

bbguimaraes:


bbguimaraes:
Hmm, for some reason we add `Provides` values twice, so https://github.com/openshift/ci-tools/pull/1362/commits/ddf9bc36d70fba49c411813058d3819bf6cbe05c generates unnecessary warnings.

stevekuznetsov (approved):

bbguimaraes:
OK, this seems to be it.

- The last commit changes `pkg/defaults/defaults.go` to process each step's `Provides` values once.  We ended up processing them twice after a refactoring.
- I initially changed `DeferredParameters::Has` to also check `values` (it only tested `fns`, which seemed wrong), but I found after some painful bisecting that it doesn't work.  I may look into this later.
- The rest of the PR remains mostly the same.
- It has and will continue to conflict badly with the refactoring that is ongoing changing the clients in `pkg/defaults/defaults.go`.
- Now that I'm writing this, I realized it will be better to separate the actual content of this PR (the now penultimate commit) from the rest.

E2E tests pass, but I want to merge this when I have time to monitor the result.
so on Monday I'll:

- merge only the fixes that duplicate the processing of `Provides`
- merge this
- finally unwind the stack back to the original PR for configurable leases

`integration`/`breaking-changes` failures seem unrelated.

/retest

bbguimaraes:
I am working on adding much needed tests for `FromConfig` and will be updating the first commit, as it makes the subsequent changes safer.  Feel free to approve at any point.

stevekuznetsov (approved):

stevekuznetsov:
Part of DPTP-1365

PR #1365, opened 2020-11-02
ci-operator: let artifacts worker cancel itself

stevekuznetsov:
In all cases that the worker had it's work cancelled, this was done as a
result of a context being done. In every case, it is improper to not
cancel the worker if the test context is cancelled, so there is no use
in forcing callers that construct an artifact worker remember that they
need to cancel it appropriately. We can instead just require the context
to be passed to the constructor and always do the right thing.


petr-muller (commented):
LGTM but lint hates you

petr-muller (approved):

bbguimaraes:
Nice.

PR #1366, opened 2020-11-02
Write extended step graph to artifacts

alvaroaleman:


stevekuznetsov (commented):
How are you expecting to add the manifests to the `Message`?

alvaroaleman:
> How are you expecting to add the manifests to the Message?

The steps will need to return that info

alvaroaleman:

pkg/steps/run.go
```diff
@@ -11,16 +11,17 @@ import (
 	"github.com/openshift/ci-tools/pkg/results"
 )
 
-type message struct {
+type Message struct {
```
bbguimaraes, line 14: Left by accident?  The type does not seem to be used outside this package.

pkg/api/graph.go
```diff
@@ -418,9 +420,68 @@ func HasAllLinks(needles, haystack []StepLink) bool {
 
 type CIOperatorStepGraph []CIOperatorStepWithDependencies
 
+// MergeFrom merges to CIOperatorStepGraphs together using StepNames as merge keys.
```
bbguimaraes, line 423:
```suggestion
// MergeFrom merges two CIOperatorStepGraphs together using StepNames as merge keys.
```

pkg/api/graph.go
```diff
@@ -418,9 +420,68 @@ func HasAllLinks(needles, haystack []StepLink) bool {
...
+func (graph *CIOperatorStepGraph) MergeFrom(from ...CIOperatorStepWithDependencies) {
+	for _, step := range from {
+		var found bool
+		for idx, existing := range *graph {
+			if step.StepName != existing.StepName {
+				continue
+			}
+			found = true
+			(*graph)[idx] = mergeSteps(existing, step)
+		}
```
bbguimaraes, line 435: Why is this check necessary?

pkg/steps/run.go
```diff
@@ -105,18 +108,28 @@ type subtestReporter interface {
...
+	out <- Message{
 		node:            node,
 		duration:        duration,
 		err:             err,
 		additionalTests: additionalTests,
+		stepDetails: api.CIOperatorStepWithDependencies{
+			StepName:    node.Step.Name(),
+			Description: node.Step.Description(),
+			StartedAt:   &start,
+			FinishedAt:  func() *time.Time { start.Add(duration); return &start }(),
```
bbguimaraes, line 130: What is the purpose of the lambda?

pkg/steps/run.go
```diff
@@ -105,18 +108,28 @@ type subtestReporter interface {
...
 }
 
-func runStep(ctx context.Context, node *api.StepNode, out chan<- message) {
+func runStep(ctx context.Context, node *api.StepNode, out chan<- Message) {
 	start := time.Now()
 	err := node.Step.Run(ctx)
 	var additionalTests []*junit.TestCase
 	if reporter, ok := node.Step.(subtestReporter); ok {
 		additionalTests = reporter.SubTests()
 	}
```
bbguimaraes, line 117: This has no impact on the execution time of the step, so it could be moved outside the timing code to allow us to reuse this function in `cmd/ci-operator/main.go`.

pkg/steps/run.go
alvaroaleman: Getting a pointer for the result of `start.Add(duration)` inline

pkg/steps/run.go
alvaroaleman: Which function do you want to use in `cmd/ci-operator/main.go`?

alvaroaleman:
@bbguimaraes PTAL

alvaroaleman:
@bbguimaraes can you have another look, please?

bbguimaraes:
Sorry, I missed the other notification.

/lgtm

PR #1369, opened 2020-11-03
De-Duplicate tide queries

alvaroaleman:
De-Duplicate our tide queries in order to reduce the amount of config we have.

Produces this, which shaves off ~200 lines of the config file: https://github.com/openshift/release/pull/13346

### DIFF
--- a/cmd/determinize-prow-config/main.go
+++ b/cmd/determinize-prow-config/main.go
@@ -3,15 +3,18 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
+	"sort"
 
 	"github.com/sirupsen/logrus"
 
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/plugins"
 	"sigs.k8s.io/yaml"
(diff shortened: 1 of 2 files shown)


alvaroaleman:
/hold
until openshift/release#13346 is merged

cmd/determinize-prow-config/main.go
```diff
@@ -61,14 +64,110 @@ func updateProwConfig(configDir string) error {
...
+type tideQueryTarget struct {
+	Orgs          []string
+	Repos         []string
+	ExcludedRepos []string
+}
+
+// tideQueryMap is a map[tideQueryConfig]*tideQueryTarget. Because slices are not comparable, they
+// or structs containing them are not allowed as map keys. We sidestep this by using a json serialization
+// of the object as key instead. This is horribly inefficient, but will never be able to beat the
+// inefficiency of our Python validation scripts.
```
petr-muller, line 102: :trollface:

cmd/determinize-prow-config/main.go
```diff
@@ -61,14 +64,110 @@ func updateProwConfig(configDir string) error {
...
+			Labels:                 queryConfig.Labels,
+			MissingLabels:          queryConfig.MissingLabels,
+			Milestone:              queryConfig.Milestone,
+			ReviewApprovedRequired: queryConfig.ReviewApprovedRequired,
+		})
+
+	}
+	var errs []error
+	sort.SliceStable(result, func(i, j int) bool {
+		iSeralized, err := json.Marshal(result[i])
```
petr-muller, line 128: nit: Seralized -> Serialized

petr-muller (commented):
The code is fine.

/hold

I'm all for simplifying the config, but I'm wondering how much worse the UX would be for the groups that had their own small queries that held for their repos, for example the ones that collected all repos from individual non-openshift orgs. But I guess the repos like these will still be together so the use case of adding a new repo is as simple as it was before. The UX of "change merge criteria for some repos" use case gets a bit harder, but that does not seem to happen that often, so I'm inclined to like this.

alvaroaleman:
> I'm all for simplifying the config, but I'm wondering how much worse the UX would be for the groups that had their own small queries that held for their repos, for example the ones that collected all repos from individual non-openshift orgs. But I guess the repos like these will still be together so the use case of adding a new repo is as simple as it was before. The UX of "change merge criteria for some repos" use case gets a bit harder, but that does not seem to happen that often, so I'm inclined to like this.

Yeah, so first I generally have the impression that queries do not get changed a lot, this mostly happens for openshift itself during freeze etc. And second, ppl can still write a completely new query for themselves and the deduplication logic will then add it to an existing query, if present.

stevekuznetsov:
This will negatively impact `status` controller updates until we shard behind the scenes, right?

alvaroaleman:
> This will negatively impact status controller updates until we shard behind the scenes, right?

Hard to say. The problem is that there is some threshold for query size at which we time out. This change does probably increase the average query size, but it doesn't necessarily push any query over the threshold. Also, I have the hope that the change that limits each page to 37 results will fix that issue regardless.

PR #1370, opened 2020-11-03
Add missing inline annotations to fix config reference generation

alvaroaleman:


pkg/api/types.go
```diff
@@ -620,7 +620,7 @@ type LiteralTestStep struct {
 	// when the command finishes. Defaults to "/tmp/artifacts"
 	ArtifactDir string `json:"artifact_dir,omitempty"`
 	// Resources defines the resource requirements for the step.
-	Resources ResourceRequirements `json:"resources,omitempty"`
+	Resources ResourceRequirements `json:"resources"`
```
alvaroaleman, line 623: `ResourceRequirements` is a literal struct, so the omitempy doesn't do anything, it will always get serialized. The only way to omit structs from serialization is to make them pointers.

PR #1375, opened 2020-11-03
CI operator graph: Add manifests from multi-stage step and image_import steps into graph

alvaroaleman:
Based on top of https://github.com/openshift/ci-tools/pull/1366

This PR makes the image_import step and the multi_stage step report their manifests back so that they will end up in the graphfile.


alvaroaleman:

PR #1376, opened 2020-11-04
dptp-cm: add registrysyncer

hongkailiu:
https://issues.redhat.com/browse/DPTP-1656

### DIFF
--- a/cmd/dptp-controller-manager/main.go
+++ b/cmd/dptp-controller-manager/main.go
@@ -22,12 +22,14 @@ import (
 	"k8s.io/test-infra/prow/pjutil"
 	controllerruntime "sigs.k8s.io/controller-runtime"
 	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/yaml"
 
 	imagev1 "github.com/openshift/api/image/v1"
 
 	"github.com/openshift/ci-tools/pkg/api/secretbootstrap"
 	"github.com/openshift/ci-tools/pkg/controller/promotionreconciler"
+	"github.com/openshift/ci-tools/pkg/controller/registrysyncer"
 	"github.com/openshift/ci-tools/pkg/controller/secretsyncer"
 	secretsyncerconfig "github.com/openshift/ci-tools/pkg/controller/secretsyncer/config"
 	testimagesdistributor "github.com/openshift/ci-tools/pkg/controller/test-images-distributor"
@@ -44,6 +46,7 @@ var allControllers = sets.NewString(
 	promotionreconciler.ControllerName,
 	testimagesdistributor.ControllerName,
 	secretsyncer.ControllerName,
+	registrysyncer.ControllerName,
 )
 
 type options struct {
@@ -60,6 +63,7 @@ type options struct {
 	blockProfileRate             time.Duration
 	testImagesDistributorOptions testImagesDistributorOptions
 	secretSyncerConfigOptions    secretSyncerConfigOptions
+	registrySyncerOptions        registrySyncerOptions
 	*flagutil.GitHubOptions
 }
 
@@ -77,6 +81,10 @@ type testImagesDistributorOptions struct {
 	additionalImageStreamNamespaces    sets.String
 	forbiddenRegistriesRaw             flagutil.Strings
 	forbiddenRegistries                sets.String
+}
+
+type registrySyncerOptions struct {
+	imagePullSecretPath string
 }
 
 type secretSyncerConfigOptions struct {
@@ -110,6 +118,7 @@ func newOpts() (*options, error) {
 	flag.Var(&opts.testImagesDistributorOptions.additionalImageStreamsRaw, "testImagesDistributorOptions.additional-image-stream", "An imagestream that will be distributed even if no test explicitly references it. It must be in namespace/name format (e.G `ci/clonerefs`). Can be passed multiple times.")
 	flag.Var(&opts.testImagesDistributorOptions.additionalImageStreamNamespacesRaw, "testImagesDistributorOptions.additional-image-stream-namespace", "A namespace in which imagestreams will be distributed even if no test explicitly references them (e.G `ci`). Can be passed multiple times.")
 	flag.Var(&opts.testImagesDistributorOptions.forbiddenRegistriesRaw, "testImagesDistributorOptions.forbidden-registry", "The hostname of an image registry from which there is no synchronization of its images. Can be passed multiple times.")
+	flag.StringVar(&opts.registrySyncerOptions.imagePullSecretPath, "registrySyncerOptions.imagePullSecretPath", "", "A file to use for reading an ImagePullSecret that will be bound to all `default` ServiceAccounts in all namespaces that have a test ImageStream on all build clusters")
 	flag.StringVar(&opts.secretSyncerConfigOptions.configFile, "secretSyncerConfigOptions.config", "", "The config file for the secret syncer controller")
 	flag.StringVar(&opts.secretSyncerConfigOptions.secretBoostrapConfigFile, "secretSyncerConfigOptions.secretBoostrapConfigFile", "", "The config file for ci-secret-boostrap")
 	flag.DurationVar(&opts.blockProfileRate, "block-profile-rate", time.Duration(0), "The block profile rate. Set to non-zero to enable.")
@@ -279,6 +288,9 @@ func main() {
 	if opts.testImagesDistributorOptions.imagePullSecretPath != "" {
 		secretPaths = append(secretPaths, opts.testImagesDistributorOptions.imagePullSecretPath)
 	}
+	if opts.registrySyncerOptions.imagePullSecretPath != "" {
+		secretPaths = append(secretPaths, opts.registrySyncerOptions.imagePullSecretPath)
+	}
 	secretAgent := &secret.Agent{}
 	if err := secretAgent.Start(secretPaths); err != nil {
 		logrus.WithError(err).Fatal("Failed to start secret agent")
@@ -388,6 +400,19 @@ func main() {
 		}
 		if err := secretsyncer.AddToManager(mgr, allManagers[apiCIContextName], targetClusters, secretSyncerConfigAgent.Config, secretBootstrapConfig); err != nil {
 			logrus.WithError(err).Fatal("failed to add secret syncer controller")
+		}
+	}
+
+	if opts.enabledControllersSet.Has(registrysyncer.ControllerName) {
+		if opts.registrySyncerOptions.imagePullSecretPath == "" {
+			logrus.Fatal("The registrysyncer requires the --registrysyncer.imagePullSecretPath flag to be set ")
+		}
+		if err := registrysyncer.AddToManager(
+			mgr,
+			map[string]manager.Manager{apiCIContextName: allManagers[apiCIContextName], appCIContextName: allManagers[appCIContextName]},
+			secretAgent.GetTokenGenerator(opts.registrySyncerOptions.imagePullSecretPath),
+		); err != nil {
+			logrus.WithError(err).Fatal("failed to add registrysyncer")
 		}
 	}
 
--- a/pkg/api/domain.go
+++ b/pkg/api/domain.go
@@ -10,6 +10,8 @@ const (
 	ServiceDomainCI    = "ci.openshift.org"
 	ServiceDomainAPICI = "svc.ci.openshift.org"
 	ServiceDomainAPPCI = "apps.ci.l2s4.p1.openshiftapps.com"
+
+	ServiceDomainAPPCIRegistry = "registry.ci.openshift.org"
 )
 
 type Service string
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -1,0 +1,340 @@
+package registrysyncer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	crcontrollerutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	imagev1 "github.com/openshift/api/image/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
+	"github.com/openshift/ci-tools/pkg/util/imagestreamtagmapper"
+	"github.com/openshift/ci-tools/pkg/util/imagestreamtagwrapper"
+)
+
+const ControllerName = "registry_syncer"
+
+func AddToManager(mgr manager.Manager,
+	managers map[string]manager.Manager,
+	pullSecretGetter func() []byte,
+) error {
+	log := logrus.WithField("controller", ControllerName)
+	r := &reconciler{
+		ctx:              context.Background(),
+		log:              log,
+		registryClients:  map[string]ctrlruntimeclient.Client{},
+		pullSecretGetter: pullSecretGetter,
+	}
+	for clusterName, m := range managers {
+		r.registryClients[clusterName] = imagestreamtagwrapper.MustNew(m.GetClient(), m.GetCache())
+	}
+	c, err := controller.New(ControllerName, mgr, controller.Options{
+		Reconciler: r,
+		// We conflict on ImageStream level which means multiple request for imagestreamtags
+		// of the same imagestream will conflict so stay at one worker in order to reduce the
+		// number of errors we see. If we hit performance issues, we will probably need cluster
+		// and/or imagestream level locking.
+		MaxConcurrentReconciles: 1,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	for _, m := range managers {
+		if err := c.Watch(
+			source.NewKindWithCache(&imagev1.ImageStream{}, m.GetCache()),
+			handlerFactory(testInputImageStreamTagFilterFactory()),
+		); err != nil {
+			return fmt.Errorf("failed to create watch for ImageStreams: %w", err)
+		}
+	}
+
+	r.log.Info("Successfully added reconciler to manager")
+	return nil
+}
+
+type objectFilter func(types.NamespacedName) bool
+
+// handlerFactory produces a handler that:
+// * Watches ImageStreams because ImageStreamTags do not support the watch verb
+// * Extracts all ImageStramTags out of the Image
+// * Filters out the ones that are not in use
+// Note: We can not use a predicate because that is directly applied on the source and the source yields ImageStreams, not ImageStreamTags
+// * Creates a reconcile.Request per cluster and ImageStreamTag
+func handlerFactory(filter objectFilter) handler.EventHandler {
+	return imagestreamtagmapper.New(func(in reconcile.Request) []reconcile.Request {
+		if !filter(in.NamespacedName) {
+			return nil
+		}
+		return []reconcile.Request{in}
+	})
+}
+
+type reconciler struct {
+	ctx              context.Context
+	log              *logrus.Entry
+	registryClients  map[string]ctrlruntimeclient.Client
+	pullSecretGetter func() []byte
+}
+
+func (r *reconciler) Reconcile(req reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithField("request", req.String())
+	log.Info("Starting reconciliation")
+	err := r.reconcile(req, log)
+	if err != nil {
+		log.WithError(err).Error("Reconciliation failed")
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+const (
+	annotationDPTPRequester = "dptp.openshift.io/requester"
+)
+
+func (r *reconciler) reconcile(req reconcile.Request, log *logrus.Entry) error {
+	isTags := map[string]*imagev1.ImageStreamTag{}
+	for clusterName, client := range r.registryClients {
+		*log = *log.WithField("clusterName", clusterName)
+		imageStreamTag := &imagev1.ImageStreamTag{}
+		if err := client.Get(r.ctx, req.NamespacedName, imageStreamTag); err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Debug("Source imageStreamTag not found")
+				continue
+			}
+			return fmt.Errorf("failed to get imageStreamTag %s from registry cluster: %w", req.NamespacedName.String(), err)
+		}
+		isTags[clusterName] = imageStreamTag
+	}
+
+	srcClusterName := findNewest(isTags)
+	if srcClusterName == "" {
+		// the isTag does NOT exist on both clusters
+		// This case is not an error but expected to happen occasionally
+		return nil
+	}
+	sourceImageStreamTag := isTags[srcClusterName]
+
+	imageStreamNameAndTag := strings.Split(req.Name, ":")
+	if n := len(imageStreamNameAndTag); n != 2 {
+		return fmt.Errorf("when splitting imagestreamtagname %s by : expected two results, got %d", req.Name, n)
+	}
+	imageStreamName, imageTag := imageStreamNameAndTag[0], imageStreamNameAndTag[1]
+	isName := types.NamespacedName{Namespace: req.Namespace, Name: imageStreamName}
+	sourceImageStream := &imagev1.ImageStream{}
+	if err := r.registryClients[srcClusterName].Get(r.ctx, isName, sourceImageStream); err != nil {
+		// received a request on the isTag, but the 'is' is no longer there
+		return fmt.Errorf("failed to get imageStream %s from registry cluster: %w", isName.String(), err)
+	}
+
+	*log = *log.WithField("docker_image_reference", sourceImageStreamTag.Image.DockerImageReference)
+
+	for clusterName, client := range r.registryClients {
+		if clusterName == srcClusterName {
+			continue
+		}
+		if err := client.Get(r.ctx, types.NamespacedName{Name: req.Namespace}, &corev1.Namespace{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to check if namespace %s exists: %w", req.Namespace, err)
+			}
+			if err := client.Create(r.ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{
+				Name:        req.Namespace,
+				Annotations: map[string]string{annotationDPTPRequester: ControllerName},
+			}}); err != nil && !apierrors.IsAlreadyExists(err) {
+				return fmt.Errorf("failed to create namespace %s: %w", req.Namespace, err)
+			}
+		}
+
+		if err := r.ensureImageStream(sourceImageStream, client, log); err != nil {
+			return fmt.Errorf("failed to ensure imagestream: %w", err)
+		}
+
+		isTag, found := isTags[clusterName]
+		if found && isTag.Image.Name == sourceImageStreamTag.Image.Name {
+			log.Debug("ImageStreamTag is current")
+			return nil
+		}
+
+		if err := r.ensureImagePullSecret(req.Namespace, client, log); err != nil {
+			return fmt.Errorf("failed to ensure imagePullSecret: %w", err)
+		}
+		imageName, err := publicDomainForImage(srcClusterName, sourceImageStreamTag.Image.DockerImageReference)
+		if err != nil {
+			return fmt.Errorf("failed to get the public domain: %w", err)
+		}
+
+		imageStreamImport := &imagev1.ImageStreamImport{
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace: req.Namespace,
+				Name:      imageStreamName,
+			},
+			Spec: imagev1.ImageStreamImportSpec{
+				Import: true,
+				Images: []imagev1.ImageImportSpec{{
+					From: corev1.ObjectReference{
+						Kind: "DockerImage",
+						Name: imageName,
+					},
+					To: &corev1.LocalObjectReference{Name: imageTag},
+					ReferencePolicy: imagev1.TagReferencePolicy{
+						Type: imagev1.LocalTagReferencePolicy,
+					},
+				}},
+			},
+		}
+
+		*log = *log.WithField("imageStreamImport.Namespace", imageStreamImport.Namespace).WithField("imageStreamImport.Name", imageStreamImport.Name)
+		log.Debug("creating imageStreamImport")
+		// TODO (hongkliu): Add metrics for recording the results
+		if err := createAndCheckStatus(r.ctx, client, imageStreamImport); err != nil {
+			return fmt.Errorf("failed to create and check the status for imageStreamImport %v : %w", imageStreamImport, err)
+		}
+		log.Debug("Imported successfully")
+	}
+	return nil
+}
+
+func createAndCheckStatus(ctx context.Context, client ctrlruntimeclient.Client, imageStreamImport *imagev1.ImageStreamImport) error {
+	// ImageStreamImport is not an ordinary api but a virtual one that does the import synchronously
+	if err := client.Create(ctx, imageStreamImport); err != nil {
+		return fmt.Errorf("failed to import Image: %w", err)
+	}
+
+	// This should never be needed, but we shouldn't panic if the server screws up
+	if imageStreamImport.Status.Images == nil {
+		imageStreamImport.Status.Images = []imagev1.ImageImportStatus{{}}
+	}
+	if imageStreamImport.Status.Images[0].Image == nil {
+		return fmt.Errorf("imageStreamImport did not succeed: reason: %s, message: %s", imageStreamImport.Status.Images[0].Status.Reason, imageStreamImport.Status.Images[0].Status.Message)
+	}
+	return nil
+}
+
+func findNewest(isTags map[string]*imagev1.ImageStreamTag) string {
+	result := ""
+	var t metav1.Time
+	for clusterName, isTag := range isTags {
+		if t.Before(&isTag.Image.CreationTimestamp) {
+			t = isTag.Image.CreationTimestamp
+			result = clusterName
+		}
+	}
+	return result
+}
+
+const pullSecretName = "registry-cluster-pull-secret"
+
+func (r *reconciler) ensureImagePullSecret(namespace string, client ctrlruntimeclient.Client, log *logrus.Entry) error {
+	secret, mutateFn := r.pullSecret(namespace)
+	return upsertObject(r.ctx, client, secret, mutateFn, log)
+}
+
+// https://issues.redhat.com/browse/DPTP-1656?focusedCommentId=15345756&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15345756
+// E.g., ci-operator uses the release controller configuration to determine
+// the version of OpenShift we create from the ImageStream, so we need
+// to copy the annotation if it exists
+const releaseConfigAnnotationPrefix = "release.openshift.io"
+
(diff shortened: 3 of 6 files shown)


cmd/dptp-controller-manager/main.go
```diff
@@ -79,6 +81,17 @@ type testImagesDistributorOptions struct {
 	forbiddenRegistries                sets.String
 }
 
+type registrySyncerOptions struct {
+	excludeImageStreamTagsRaw       flagutil.Strings
+	excludeImageStreamTags          sets.String
+	excludeImageStreamsRaw          flagutil.Strings
+	excludeImageStreams             sets.String
+	excludeImageStreamNamespacesRaw flagutil.Strings
```
alvaroaleman, line 89: I don't think excludedNamespaces or forbiddenRegistries make a lot of sense for this controller

cmd/dptp-controller-manager/main.go
```diff
@@ -174,6 +191,44 @@ func newOpts() (*options, error) {
 		opts.testImagesDistributorOptions.forbiddenRegistries.Insert(vals...)
 	}
 
+	opts.registrySyncerOptions.excludeImageStreamTags = sets.String{}
+	if vals := opts.registrySyncerOptions.excludeImageStreamTagsRaw.Strings(); len(vals) > 0 {
```
alvaroaleman, line 195: We should probably factor the namespace/name and imagestreamtag flag parsing logic out

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,461 @@
...
+func AddToManager(mgr manager.Manager,
+	apiCIManager manager.Manager,
+	appCIManager manager.Manager,
+	pullSecretGetter func() []byte,
+	excludeImageStreamTags sets.String,
+	excludeImageStreams sets.String,
+	excludeImageStreamNamespaces sets.String,
+	forbiddenRegistries sets.String,
+) error {
+	apiCIControllerName := fmt.Sprintf("%s_api_ci", ControllerName)
```
alvaroaleman, line 42: Why do we have two controllers? We should just have one controller that:
* Watches in both clusters
* Will during its reconcile get the imagestreamtag from both clusters
* If they are not equal, will create an import on the cluster where its older

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,461 @@
...
+	apiCILog := logrus.WithField("controller", apiCIControllerName)
+	apiCIReconsiler := &reconciler{
```
alvaroaleman, line 44: s/Reconsiler/Reconciler/g

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,461 @@
...
+		// of the same imagestream will conflict so stay at one worker in order to reduce the
+		// number of errors we see. If we hit performance issues, we will probably need cluster
+		// and/or imagestream level locking.
+		MaxConcurrentReconciles: 1,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	apiCIObjectFilter, err := testInputImageStreamTagFilterFactory(apiCILog, excludeImageStreamTags, excludeImageStreams, excludeImageStreamNamespaces)
```
alvaroaleman, line 64: As mentioned in Slack, I think it would be good to only have a no-op filter here in the beginning and create a follow-up PR to implement the filtering logic.

hack/run-dptp-controller-manager.sh
```diff
@@ -27,15 +27,8 @@ kubectl --context build01 get secret -n ci regcred  -o json  \
 go build  -v -o /tmp/dptp-cm ./cmd/dptp-controller-manager
 /tmp/dptp-cm \
   --leader-election-namespace=ci \
-  --ci-operator-config-path="$(go env GOPATH)/src/github.com/openshift/release/ci-operator/config" \
-  --config-path="$(go env GOPATH)/src/github.com/openshift/release/core-services/prow/02_config/_config.yaml" \
-  --job-config-path="$(go env GOPATH)/src/github.com/openshift/release/ci-operator/jobs" \
```
alvaroaleman, line 32: Can you leave the config path variables? They are not needed now, but make it easier when someone wants to test one of the controllers that do need them

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+		// number of errors we see. If we hit performance issues, we will probably need cluster
+		// and/or imagestream level locking.
+		MaxConcurrentReconciles: 1,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	allContexts := sets.NewString()
+	for ctx := range managers {
```
alvaroaleman, line 63: Nit: You can move this in the loop below, no reason to loop twice

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+
+type objectFilter func(types.NamespacedName) bool
+
+// registryClusterHandlerFactory produces a handler that:
+// * Watches ImageStreams because ImageStreamTags do not support the watch verb
+// * Extracts all ImageStramTags out of the Image
+// * Filters out the ones that are not in use
+// Note: We can not use a predicate because that is directly applied on the source and the source yields ImageStreams, not ImageStreamTags
+// * Creates a reconcile.Request per cluster and ImageStreamTag
+func registryClusterHandlerFactory(buildClusters sets.String, filter objectFilter) handler.EventHandler {
```
alvaroaleman, line 89: please rename to just `handlerFactory`

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+	return imagestreamtagmapper.New(func(in reconcile.Request) []reconcile.Request {
+		if !filter(in.NamespacedName) {
+			return nil
+		}
+
+		var requests []reconcile.Request
```
alvaroaleman, line 95: We don't need any of this, just `return in` here. One `reconcile.Request` will reconcile both clusters

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+				Namespace: buildCluster + clusterAndNamespaceDelimiter + in.Namespace,
+				Name:      in.Name,
+			}
+			requests = append(requests, reconcile.Request{NamespacedName: name})
+		}
+		return requests
+	})
+}
+
+const clusterAndNamespaceDelimiter = "_"
```
alvaroaleman, line 110: Not needed

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+
+func decodeRequest(req reconcile.Request) (string, types.NamespacedName, error) {
```
alvaroaleman, line 112: Not needed

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+	return clusterAndNamespace[0], types.NamespacedName{Namespace: clusterAndNamespace[1], Name: req.Name}, nil
+}
+
+type reconciler struct {
+	ctx                 context.Context
+	log                 *logrus.Entry
+	registryClients     map[string]ctrlruntimeclient.Client
+	pullSecretGetter    func() []byte
+	forbiddenRegistries sets.String
```
alvaroaleman, line 125: Not needed, we should just filter in the `Handler`

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		log.WithError(err).Error("Reconciliation failed")
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+func (r *reconciler) reconcile(req reconcile.Request, log *logrus.Entry) error {
+	cluster, decoded, err := decodeRequest(req)
```
alvaroaleman, line 142: Not needed

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to decode request %s: %w", req, err)
+	}
+
+	// Propagate the cluster field back up
+	*log = *log.WithField("cluster", cluster)
```
alvaroaleman, line 148: Not needed

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+
+	isTags := map[string]*imagev1.ImageStreamTag{}
+	for ctx, client := range r.registryClients {
+		imageStreamTag := &imagev1.ImageStreamTag{}
+		if err := client.Get(r.ctx, decoded, imageStreamTag); err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Debug("Source imageStreamTag not found")
```
alvaroaleman, line 155: We don't know what cluster the request came from, so we can say which is the `source`. We should:
* Do nothing if it doesn't exist on both clusters
* Do nothing if its the same on both clusters
* Import if its newer on one cluster or doesn't exist on the other

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+}
+
+func findNewest(isTags map[string]*imagev1.ImageStreamTag) (string, error) {
+	resultContext := ""
+	t := metav1.Now()
+	for ctx, isTag := range isTags {
+		if isTag == nil {
+			continue
+		}
+		if isTag.Image.CreationTimestamp.Before(&t) {
```
alvaroaleman, line 270: If `creationTimestamp` is `before now()` its the `sourceContext`? This condition is always true. You need to compare the creation timestamps of the two, not compare against `now()`

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+	}
+	imageStreamName, imageTag := imageStreamNameAndTag[0], imageStreamNameAndTag[1]
+	isName := types.NamespacedName{Namespace: decoded.Namespace, Name: imageStreamName}
+	sourceImageStream := &imagev1.ImageStream{}
+	if err := r.registryClients[srcContext].Get(r.ctx, isName, sourceImageStream); err != nil {
+		return fmt.Errorf("failed to get imageStream %s from registry cluster: %w", isName.String(), err)
+	}
+
+	*log = *log.WithField("docker_image_reference", sourceImageStreamTag.Image.DockerImageReference)
+	if isImportForbidden(sourceImageStreamTag.Image.DockerImageReference, r.forbiddenRegistries) {
```
alvaroaleman, line 182: We don't need this check, hence we also don't need to get the `sourceImageStream`

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+		if err := client.Get(r.ctx, types.NamespacedName{Name: decoded.Namespace}, &corev1.Namespace{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to check if namespace %s exists: %w", decoded.Namespace, err)
+			}
+			if err := client.Create(r.ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: decoded.Namespace}}); err != nil && !apierrors.IsAlreadyExists(err) {
+				return fmt.Errorf("failed to create namespace %s: %w", decoded.Namespace, err)
+			}
+		}
+
+		imageStreamImport := &imagev1.ImageStreamImport{
```
alvaroaleman, line 201: We need to upsert the ImagePullSeret first

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+					},
+				}},
+			},
+		}
+
+		r.log.Debugf("creating import %v on cluster %s", imageStreamImport, ctx)
+		if r.dryRun {
+			continue
+		}
+
```
alvaroaleman, line 225: Please remove the rolebinding and role creation code

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,437 @@
...
+			return fmt.Errorf("failed to ensure rolebinding: %w", err)
+		}
+		if err := r.ensureCIOperatorRole(decoded.Namespace, client, log); err != nil {
+			return fmt.Errorf("failed to ensure role: %w", err)
+		}
+		if err := r.ensureImageStream(sourceImageStream, client, log); err != nil {
+			return fmt.Errorf("failed to ensure imagestream: %w", err)
+		}
+
+		if isTag != nil && isTag.Image.Name == sourceImageStreamTag.Image.Name {
```
alvaroaleman, line 236: The ImageStreamTagImport should not be defined before this check and we should not log that we might create an import when we probably won't. The client itself has dryRun mode enabled btw, it doesn't work on api.ci because its too old (it will just error) but it works fine on 4.X clusters. So there is no need to have a `dryRun` config option

hack/run-dptp-controller-manager.sh
hongkailiu: This wont be merged for real: Just want you to check the script that I plan to run.
I will recovery this before merge.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Need to calculate `allContexts ` first and use it in the 2nd loop.

alvaroaleman: Right, except that you don't need it there because you don't need to create cluster-scoped `reconcile.Requests`

pkg/controller/registrysyncer/registrysyncer.go
Will add unit tests for this function.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Removed `client.get`.
I think we still need `ensureImageStream(sourceImageStream, client, log)`: Can we create `imageStreamImport` without the `IS`?

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: I moved it up here because I wanted to log it at dry run without creating new objects.
Moved it back now because dryRun is recognized by the controller as pointed out https://github.com/openshift/ci-tools/pull/1376#discussion_r517650793

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Why do we need them for the other controller?
What is the diff?

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
alvaroaleman: For the images distributed by the other controller we know that they are used in tests so we need to give ci-operator access to allow importing it. Here, we don't know and we can safely assume that the other controller will already do that if needed.

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+			}
+			return fmt.Errorf("failed to get imageStreamTag %s from registry cluster: %w", req.NamespacedName.String(), err)
+		}
+		isTags[ctx] = imageStreamTag
+	}
+
+	srcContext, err := findNewest(isTags)
+	if err != nil {
+		// if it doesn't exist on both clusters; only log the error
+		log.Warnf("failed to find the new isTag from %v: %v", isTags, err)
```
alvaroaleman, line 127: If it doesn't exist on both clusters its not an error at all, so no reason to log anything. Its also not really an error

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+	sourceImageStream := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      imageStreamName,
+			Namespace: req.Namespace,
+		},
+	}
+
+	*log = *log.WithField("docker_image_reference", sourceImageStreamTag.Image.DockerImageReference)
+
+	for ctx, isTag := range isTags {
```
alvaroaleman, line 146: can we globally `s/ctx/clusterName/g` for readability, please?

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+					},
+					To: &corev1.LocalObjectReference{Name: imageTag},
+					ReferencePolicy: imagev1.TagReferencePolicy{
+						Type: imagev1.LocalTagReferencePolicy,
+					},
+				}},
+			},
+		}
+
+		r.log.Debugf("creating import %v on cluster %s", imageStreamImport, ctx)
```
alvaroaleman, line 197: We will absolutely need metrics for this btw( but can be a follow-up)

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+
+		// ImageStreamImport is not an ordinary api but a virtual one that does the import synchronously
+		if err := client.Create(r.ctx, imageStreamImport); err != nil {
```
alvaroaleman, line 200: It might make sense to factor this until the status check out, because we want a success or error counter metric

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+		log.Debug("Imported successfully")
+	}
+	return nil
+}
+
+func findNewest(isTags map[string]*imagev1.ImageStreamTag) (string, error) {
+	resultContext := ""
+	t := metav1.NewTime(time.Time{})
+	for ctx, isTag := range isTags {
+		if isTag == nil {
```
alvaroaleman, line 221: How can it be nil?

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		if t.Before(&isTag.Image.CreationTimestamp) {
+			t = isTag.Image.CreationTimestamp
+			resultContext = ctx
+		}
+	}
+	if resultContext != "" {
+		return resultContext, nil
+	}
+	return "", fmt.Errorf("cannot find the newest isTag from %v", isTags)
```
alvaroaleman, line 232: Thats not an error but expected to happen occasionally

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,316 @@
...
+
+func imagestream(imageStream *imagev1.ImageStream) (*imagev1.ImageStream, crcontrollerutil.MutateFn) {
+	stream := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: imageStream.Namespace,
+			Name:      imageStream.Name,
+		},
+	}
+	return stream, func() error {
+		if config, set := imageStream.Annotations[releaseConfigAnnotation]; set {
```
alvaroaleman, line 255: Can we just copy everything, modulo readOnly fields?

pkg/api/domain.go
```diff
@@ -41,3 +43,25 @@ func DomainForService(service Service) string {
 	}
 	return fmt.Sprintf("%s.%s", service, serviceDomain)
 }
+
+func DomainForContextAndService(context string, service Service) (string, error) {
```
alvaroaleman, line 47: I don't think we should change anything in here, the only consume of this will remain that controller

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -0,0 +1,178 @@
...
+			isTags:        map[string]*imagev1.ImageStreamTag{},
+			expectedError: fmt.Errorf("cannot find the newest isTag from map[]"),
+		},
+		{
+			name: "both nil",
+			isTags: map[string]*imagev1.ImageStreamTag{
+				"ctx1": nil,
+				"ctx2": nil,
+			},
+			expectedError: fmt.Errorf("cannot find the newest isTag from map[ctx1:nil ctx2:nil]"),
```
alvaroaleman, line 92: I would just make the underlying func return a `found bool` instead of an error

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Removed the log and replaced it with comments.

pkg/controller/registrysyncer/registrysyncer.go

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Added a TODO.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go

pkg/controller/registrysyncer/registrysyncer.go

pkg/controller/registrysyncer/registrysyncer.go

Thought to do it in another PR.

pkg/api/domain.go
hongkailiu: Thought we make a central place for domains, clusterNames, etc.
But I can do it later.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Update.

cmd/dptp-controller-manager/main.go
```diff
@@ -110,6 +117,7 @@ func newOpts() (*options, error) {
 	flag.Var(&opts.testImagesDistributorOptions.additionalImageStreamsRaw, "testImagesDistributorOptions.additional-image-stream", "An imagestream that will be distributed even if no test explicitly references it. It must be in namespace/name format (e.G `ci/clonerefs`). Can be passed multiple times.")
 	flag.Var(&opts.testImagesDistributorOptions.additionalImageStreamNamespacesRaw, "testImagesDistributorOptions.additional-image-stream-namespace", "A namespace in which imagestreams will be distributed even if no test explicitly references them (e.G `ci`). Can be passed multiple times.")
 	flag.Var(&opts.testImagesDistributorOptions.forbiddenRegistriesRaw, "testImagesDistributorOptions.forbidden-registry", "The hostname of an image registry from which there is no synchronization of its images. Can be passed multiple times.")
+	flag.StringVar(&opts.registrySyncerOptions.imagePullSecretPath, "registrySyncerOptions.imagePullSecretPath", "", "A file to use for reading an ImagePullSecret that will be bound to all `default` ServiceAccounts in all namespaces that have a test ImageStream on all build clusters")
```
alvaroaleman, line 120: Can't we re-use the one from the other controller?

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,336 @@
...
+}
+
+func (r *reconciler) reconcile(req reconcile.Request, log *logrus.Entry) error {
+
+	isTags := map[string]*imagev1.ImageStreamTag{}
+	for clusterName, client := range r.registryClients {
+		imageStreamTag := &imagev1.ImageStreamTag{}
+		if err := client.Get(r.ctx, req.NamespacedName, imageStreamTag); err != nil {
+			if apierrors.IsNotFound(err) {
+				log.Debugf("Source imageStreamTag %s not found on cluster %s", req.NamespacedName.String(), clusterName)
```
alvaroaleman, line 115: Please never use `log.Anythingf` and use proper fields instead

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	svcDomainAndPort := "image-registry.openshift-image-registry.svc:5000"
+	if clusterName == "api.ci" {
+		svcDomainAndPort = "docker-registry.default.svc:5000"
+	}
+
+	return strings.ReplaceAll(potentiallyPrivate, svcDomainAndPort, d), nil
+}
+
+func domainForContextAndService(context string, service api.Service) (string, error) {
```
alvaroaleman, line 312: Please rename this to `clusterName`. Using the word `context` here makes this highly confusing with the `context` package

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,336 @@
...
+
+func imagestream(imageStream *imagev1.ImageStream) (*imagev1.ImageStream, crcontrollerutil.MutateFn) {
+	stream := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: imageStream.Namespace,
+			Name:      imageStream.Name,
+		},
+	}
+	return stream, func() error {
+		for k, v := range imageStream.Annotations {
```
alvaroaleman, line 255: We never get the `imageStram`, the object we pass in here is guaranteed to be be empty and not have any annotations or anything set in the spec field

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,336 @@
...
+		return nil
+	}
+	sourceImageStreamTag := isTags[srcClusterName]
+
+	imageStreamNameAndTag := strings.Split(req.Name, ":")
+	if n := len(imageStreamNameAndTag); n != 2 {
+		return fmt.Errorf("when splitting imagestreamtagname %s by : expected two results, got %d", req.Name, n)
+	}
+	imageStreamName, imageTag := imageStreamNameAndTag[0], imageStreamNameAndTag[1]
+	sourceImageStream := &imagev1.ImageStream{
```
alvaroaleman, line 136: See the other comment, we have to GET this to make sure we correctly copy its settings over

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,336 @@
...
+	}
+	if imageStreamImport.Status.Images[0].Image == nil {
+		return fmt.Errorf("imageStreamImport did not succeed: reason: %s, message: %s", imageStreamImport.Status.Images[0].Status.Reason, imageStreamImport.Status.Images[0].Status.Message)
+	}
+	return nil
+}
+
+func findNewest(isTags map[string]*imagev1.ImageStreamTag) string {
+	resultContext := ""
+	t := metav1.NewTime(time.Time{})
```
alvaroaleman, line 224: You can just do a `var t metav1.Time`, no reason to go through the constructor for an empty object

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,336 @@
...
+		imageStreamImport.Status.Images = []imagev1.ImageImportStatus{{}}
+	}
+	if imageStreamImport.Status.Images[0].Image == nil {
+		return fmt.Errorf("imageStreamImport did not succeed: reason: %s, message: %s", imageStreamImport.Status.Images[0].Status.Reason, imageStreamImport.Status.Images[0].Status.Message)
+	}
+	return nil
+}
+
+func findNewest(isTags map[string]*imagev1.ImageStreamTag) string {
+	resultContext := ""
```
alvaroaleman, line 223: s/Context/ClusterName/g

pkg/controller/registrysyncer/registrysyncer.go
alvaroaleman, line 312: The second arg to this func is useless, it can only be `api.ServiceRegistry` which then means you can get rid of the inner switch statement below

cmd/dptp-controller-manager/main.go
hongkailiu: That one is called `opts.testImagesDistributorOptions.imagePullSecretPath`.
Feel strange to use it in the current one.

pkg/controller/registrysyncer/registrysyncer.go
Not sure if i did it right.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Copied the code back. :)

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

hack/run-dptp-controller-manager.sh
hongkailiu: Have to recover them because they are mandatory flags.

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,332 @@
...
+
+	for clusterName, client := range r.registryClients {
+		if clusterName == srcClusterName {
+			continue
+		}
+		if err := client.Get(r.ctx, types.NamespacedName{Name: req.Namespace}, &corev1.Namespace{}); err != nil {
+			if !apierrors.IsNotFound(err) {
+				return fmt.Errorf("failed to check if namespace %s exists: %w", req.Namespace, err)
+			}
+			if err := client.Create(r.ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: req.Namespace}}); err != nil && !apierrors.IsAlreadyExists(err) {
```
alvaroaleman, line 152: Nit: Can we add a label if we create the NS to identify it as created by this controller?

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,332 @@
...
+func imagestream(imageStream *imagev1.ImageStream) (*imagev1.ImageStream, crcontrollerutil.MutateFn) {
+	stream := &imagev1.ImageStream{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: imageStream.Namespace,
+			Name:      imageStream.Name,
+		},
+	}
+	return stream, func() error {
+		for k, v := range imageStream.Annotations {
+			if stream.Annotations == nil {
```
alvaroaleman, line 255: We should create the map outside the loop. And IMHO it makes more sense to copy all labels or not?

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,332 @@
...
+				stream.Annotations = map[string]string{}
+			}
+			if strings.HasPrefix(k, releaseConfigAnnotationPrefix) {
+				stream.Annotations[k] = v
+			}
+		}
+		stream.Spec.LookupPolicy.Local = true
```
alvaroaleman, line 262: we should copy this field from whatever it is in the source imagestream

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,332 @@
...
+		for i := range stream.Spec.Tags {
+			stream.Spec.Tags[i].ReferencePolicy.Type = imagev1.LocalTagReferencePolicy
```
alvaroaleman, line 264: same here, lets copy and not hardcode

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,332 @@
...
+					},
+					To: &corev1.LocalObjectReference{Name: imageTag},
+					ReferencePolicy: imagev1.TagReferencePolicy{
+						Type: imagev1.LocalTagReferencePolicy,
+					},
+				}},
+			},
+		}
+
+		r.log.Debugf("creating imageStreamImport %v on cluster %s", imageStreamImport, clusterName)
```
alvaroaleman, line 195: No debugf, use fields

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -0,0 +1,332 @@
...
+		// TODO (hongkliu): Add metrics for recording the results
+		if err := createAndCheckStatus(r.ctx, client, imageStreamImport); err != nil {
```
alvaroaleman, line 197: Can you wrap and log this error?

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -0,0 +1,454 @@
...
+		} else {
+			asserted.Status.Images[0].Image = &imagev1.Image{}
+		}
+	}
+	return client.Client.Create(ctx, obj, opts...)
+}
+
+var (
+	//https://github.com/google/go-cmp/issues/24#issuecomment-317635190
+	equateErrorMessage = cmp.Comparer(func(x, y error) bool {
```
alvaroaleman, line 448: Can you move this into `pkg/testhelper`? I know for sure that I want to use it in the future as well :)

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -0,0 +1,454 @@
...
+				return nil
+			},
+		},
+		{
+			name: "app.ci is newer",
+			request: types.NamespacedName{
+				Name:      "applyconfig:latest",
+				Namespace: "ci",
+			},
+			apiCIClient: bcc(fakeclient.NewFakeClientWithScheme(scheme.Scheme, []runtime.Object{applyconfigISTag.DeepCopy()}...)),
```
alvaroaleman, line 317: ftr, `NewFakeClientWithScheme(scheme.Scheme)` produces the exact same result as `NewFakeClient`. The latter got undeprecated btw. So maybe change it for brevity?

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -0,0 +1,454 @@
...
+				Namespace: "ci",
+			},
+			apiCIClient: fakeclient.NewFakeClientWithScheme(scheme.Scheme, []runtime.Object{applyconfigISTag.DeepCopy()}...),
+			appCIClient: fakeclient.NewFakeClientWithScheme(scheme.Scheme, []runtime.Object{applyconfigISTagNewerSameName.DeepCopy(), applyconfigIS.DeepCopy()}...),
+
+			verify: func(apiCIClient ctrlruntimeclient.Client, appCIClient ctrlruntimeclient.Client) error {
+				for clusterName, client := range map[string]ctrlruntimeclient.Client{apiCI: apiCIClient, appCI: appCIClient} {
+					actualImageStreamImport := &imagev1.ImageStreamImport{}
+					err := client.Get(ctx, types.NamespacedName{Name: "applyconfig", Namespace: "ci"}, actualImageStreamImport)
+					if !errors.IsNotFound(err) {
```
alvaroaleman, line 375: Nit: You could `list` here and verify len(result)  == 0, that would make the test correctly fail even if the import had a different name

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Unless you have already seen the issues coming, I would just stick to the jira card for now.
https://issues.redhat.com/browse/DPTP-1656?focusedCommentId=15345756&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15345756

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: I moved it into further inside.
It might have done some unneeded checks but init the map only when necessary.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Updated.

`undeprecated `?: TIL.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Left a comment to address why we cannot use `List` here.

alvaroaleman (approved):

PR #1377, opened 2020-11-04
prowgen: move code to pkg/ and simplify it

petr-muller:
As a followup to DRYing `IsGenerated()` in #1274, this change moves more
logic from `cmd/ci-operator-prowgen` to `pkg/prowgen`. Some code was
made an internal implementation detail, resulting in simpler signatures. Some
prowgen-related code was also moved from `jobconfig` to `prowgen`.

The labels that prowgen uses to mark (newly) generated jobs are now
internal and no longer configurable, but that was not used anyway.
`GenerateJobs` now always generates jobs with a `newly-generated` label
value, and `Prune` then switches the label to `true`.
Unfortunately many tests for `[gG]Generate*` methods passed different
labels so a lot of fixtures needed to be changed, but the actual
behavior of `ci-operator-prowgen` does not change, witnessed by
integration tests.

/hold 
Needs https://github.com/openshift/ci-tools/pull/1274 first

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -94,7 +94,7 @@ func readProwgenConfig(path string) (*config.Prowgen, error) {
 // appropriate location, and either stored a pointer to the parsed config if if was
 // successfully read, or stored `nil` when the prowgen config could not be read (usually
 // because the drop-in is not there).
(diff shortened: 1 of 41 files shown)


stevekuznetsov (approved):

stevekuznetsov:
/retest
Flaked on busybox

PR #1381, opened 2020-11-04
Test Images Distributor: Use ResolveConfig instead of re-implementing it

alvaroaleman:
What is confusing though:  Why does the registry not resolve `RawSteps` but only `Tests`? https://github.com/openshift/ci-tools/blob/7b99e122c7ec94a8f7b0d9f5b6b33da4570a4574/pkg/registry/resolver.go#L327


bbguimaraes:
> What is confusing though: Why does the registry not resolve `RawSteps` but only `Tests`?

`RawSteps` is used to ~hack~ add step configurations that are not available via the configuration format.  Multi-stage steps are fully exposed, I cannot think of a case that would require a raw step.

PR #1382, opened 2020-11-05
Skip a linux-only test

hongkailiu:
This test prevents me from running `make test` successfully on a non-Linux PC.

### DIFF
--- a/pkg/steps/bundle_source_test.go
+++ b/pkg/steps/bundle_source_test.go
@@ -5,6 +5,7 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


bbguimaraes:
It looks pretty portable, which part of it does not work on a Mac?

hongkailiu:
> It looks pretty portable, which part of it does not work on a Mac?

I did not debug into each step.
My guess would be

https://github.com/openshift/ci-tools/blob/4abd357c14eb0a0d40110cca7904018a7d5bf2e3/pkg/steps/bundle_source.go#L77

stevekuznetsov (approved):

PR #1393, opened 2020-11-06
controllers: add retries for failed images jobs

stevekuznetsov:
The current image promotion reconciler ensures that we have correct
output in the form of ImageStreamTags, but misses the most common reason
that the content may be incorrect: a post-submit job building images
failed. This new controller re-triggers failed post-submit images jobs
when they are for the latest commit in the branch they're meant to
build.

/assign @alvaroaleman 
implements [DPTP-1647](https://issues.redhat.com/browse/DPTP-1647)

### DIFF
--- a/pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
+++ b/pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
@@ -8,21 +8,20 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/workqueue"
 	prowjobsv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/github"
 	"sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
@@ -45,7 +44,6 @@ const (
 func AddToManager(mgr controllerruntime.Manager, opts Options) error {
 	log := logrus.WithField("controller", ControllerName)
 	r := &reconciler{
-		ctx:          context.Background(),
 		log:          log,
 		client:       mgr.GetClient(),
 		gitHubClient: opts.GitHubClient,
@@ -65,13 +63,13 @@ func AddToManager(mgr controllerruntime.Manager, opts Options) error {
 			UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
 				q.Add(reconcile.Request{
 					NamespacedName: types.NamespacedName{
-						Namespace: e.MetaNew.GetNamespace(),
-						Name:      e.MetaNew.GetName(),
+						Namespace: e.ObjectNew.GetNamespace(),
+						Name:      e.ObjectNew.GetName(),
 					},
 				})
 			},
 		},
-		predicate.NewPredicateFuncs(func(meta metav1.Object, object runtime.Object) bool {
+		predicate.NewPredicateFuncs(func(object client.Object) bool {
 			prowJob, ok := object.(*prowjobsv1.ProwJob)
 			if !ok {
 				return false
@@ -97,20 +95,19 @@ type RefGetter interface {
 }
 
 type reconciler struct {
-	ctx          context.Context
 	log          *logrus.Entry
 	client       ctrlruntimeclient.Client
 	gitHubClient RefGetter
 	enqueueJob   prowjobreconciler.Enqueuer
 }
 
-func (r *reconciler) Reconcile(req controllerruntime.Request) (controllerruntime.Result, error) {
+func (r *reconciler) Reconcile(ctx context.Context, req controllerruntime.Request) (controllerruntime.Result, error) {
 	log := r.log.WithField("name", req.Name).WithField("namespace", req.Namespace)
 	log.Trace("Starting reconciliation")
 	startTime := time.Now()
 	defer func() { log.WithField("duration", time.Since(startTime)).Trace("Finished reconciliation") }()
 
-	err := r.reconcile(req, log)
+	err := r.reconcile(ctx, req, log)
 	if err != nil {
 		log.WithError(err).Error("Reconciliation failed")
 	}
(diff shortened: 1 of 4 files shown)


pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
```diff
@@ -0,0 +1,185 @@
...
+	CIOperatorConfigAgent agents.ConfigAgent
+	ConfigGetter          config.Getter
+	GitHubClient          github.Client
+}
+
+const (
+	ControllerName = "promotion_job_retrigger"
+)
+
+func AddToManager(mgr controllerruntime.Manager, opts Options) error {
```
alvaroaleman, line 45: This could be achieved much simpler by simply watching prowjobs in the promotionreconciler and enquing the imagestreamtag(s) they promote to: https://github.com/openshift/ci-tools/blob/2e70752110c84876f70ae7899d491992b527f09c/pkg/controller/promotionreconciler/reconciler.go#L85

pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
```diff
@@ -0,0 +1,185 @@
...
+	log := logrus.WithField("controller", ControllerName)
+	r := &reconciler{
+		ctx:          context.Background(),
+		log:          log,
+		client:       mgr.GetClient(),
+		gitHubClient: opts.GitHubClient,
+		enqueueJob:   prowJobEnqueuer,
+	}
+
+	c, err := controller.New(ControllerName, mgr, controller.Options{
```
alvaroaleman, line 60: This could be achieved much simpler by simply watching prowjobs in the promotionreconciler and enquing the imagestreamtag(s) they promote to: https://github.com/openshift/ci-tools/blob/2e70752110c84876f70ae7899d491992b527f09c/pkg/controller/promotionreconciler/reconciler.go#L85

pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	if err := c.Watch(
+		&source.Kind{Type: &prowjobsv1.ProwJob{}},
+		handler.Funcs{
+			UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+				q.Add(e)
+			},
+		},
+		predicate.NewPredicateFuncs(func(meta metav1.Object, object runtime.Object) bool {
+			prowJob, ok := object.(*prowjobsv1.ProwJob)
```
alvaroaleman, line 75: Should also test for type postsubmit

pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
```diff
@@ -0,0 +1,185 @@
...
+			log.Debug("Reconciliation failed")
+		} else {
+			log.Error("Reconciliation failed")
+		}
+	}
+
+	return controllerruntime.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+func (r *reconciler) reconcile(req controllerruntime.Request, log *logrus.Entry) error {
```
alvaroaleman, line 125: We need something ton ratelimit this in case of jobs that are broken

stevekuznetsov: I'm not sure we do - we don't enqueue new jobs if one is running and images postsubmits have `max_concurrency` set to 1 so we're effectively rate-limited at serial retry. Not sure how exactly we could achieve e.g. exponential backoff for this since it's not one specific key in the queue that fails, the queue keys get processed correctly and it's the ensuing job that fails. I will add alerts for this controller doing too many re-runs and we *should* be OK with the serial behavior?

stevekuznetsov (commented):
Updated this - with sharing the prowjob queue, this is now ~15LoC boilerplate on top of the actual logic. I don't want to force this separate flow through the other controller, it will be more LoC and mental overhead to this approach, which is simple and straightforward.

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -32,47 +32,36 @@ import (
...
 
 const ControllerName = "promotionreconciler"
 
-func AddToManager(mgr controllerruntime.Manager, opts Options) error {
+func AddToManager(mgr controllerruntime.Manager, opts Options, enqueuer prowjobreconciler.Enqueuer) error {
 	// Pre-Allocate the Image informer rather than letting it allocate on demand, because
 	// starting the watch takes very long (~2 minutes) and having that delay added to our
 	// first (# worker) reconciles skews the workqueue duration metric bigtimes.
-	if _, err := opts.RegistryManager.GetCache().GetInformer(context.TODO(), &imagev1.Image{}); err != nil {
+	if _, err := mgr.GetCache().GetInformer(context.TODO(), &imagev1.Image{}); err != nil {
```
alvaroaleman, line 45: This is wrong I think, this needs to be the api.ci manager and its the app.ci one now
This part can be just deleted though, it happens in `imagestreamwarpper.MustNew`

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -32,47 +32,36 @@ import (
...
-	if err != nil {
-		return fmt.Errorf("failed to construct prowjobreconciler: %w", err)
-	}
-
 	log := logrus.WithField("controller", ControllerName)
 	r := &reconciler{
 		ctx:    context.Background(),
 		log:    log,
-		client: imagestreamtagwrapper.MustNew(opts.RegistryManager.GetClient(), opts.RegistryManager.GetCache()),
+		client: imagestreamtagwrapper.MustNew(mgr.GetClient(), mgr.GetCache()),
```
alvaroaleman, line 57: Same here, wrong `manager`

stevekuznetsov: No, I changed the one that we pass in, since we no longer needed to have both around

pkg/controller/promotionreconciler/reconciler.go
stevekuznetsov: removed

pkg/controller/promotionreconciler/reconciler.go
alvaroaleman: Yeah but this has a set of issues:
* We are not explicit anymore about the fact that this must be the registry cluster, making it likely that we screw this up when we change the registry
* It messes with the start order, now we start main manager -> registry manager -> controller, as opposed to main manager -> controller

pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
```diff
@@ -0,0 +1,173 @@
...
+type Options struct {
+	CIOperatorConfigAgent agents.ConfigAgent
+	GitHubClient          github.Client
+}
+
+const (
+	ControllerName = "promotion_job_retrigger"
+)
+
+func AddToManager(mgr controllerruntime.Manager, opts Options, enqueuer prowjobreconciler.Enqueuer) error {
```
alvaroaleman, line 42: Why do we have an options struct and then don't use it for all options?

pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
```diff
@@ -0,0 +1,173 @@
...
+		MaxConcurrentReconciles: 10,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+	if err := c.Watch(
+		&source.Kind{Type: &prowjobsv1.ProwJob{}},
+		handler.Funcs{
+			UpdateFunc: func(e event.UpdateEvent, q workqueue.RateLimitingInterface) {
+				q.Add(e)
```
alvaroaleman, line 63: While this compiles ts not going to work at all. You must put `reconcile.Request` in the q and not events, otherwise your controller will never do anything except log errors. I recommend using `hack/run-dptp-controller-manager.sh`  btw to test the basic functionality

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -32,47 +32,29 @@ import (
...
 
-func AddToManager(mgr controllerruntime.Manager, opts Options) error {
-	// Pre-Allocate the Image informer rather than letting it allocate on demand, because
-	// starting the watch takes very long (~2 minutes) and having that delay added to our
-	// first (# worker) reconciles skews the workqueue duration metric bigtimes.
-	if _, err := opts.RegistryManager.GetCache().GetInformer(context.TODO(), &imagev1.Image{}); err != nil {
-		return fmt.Errorf("failed to get informer for image: %w", err)
-	}
-
+func AddToManager(mgr controllerruntime.Manager, opts Options, enqueuer prowjobreconciler.Enqueuer) error {
```
alvaroaleman, line 41: Same as above, why have an options struct and then not put all options in it?

pkg/controller/promotionreconciler/prowjobretrigger/reconciler.go
stevekuznetsov: I like to rely on types to be my crutch for not making dumb mistakes :)

stevekuznetsov:
@alvaroaleman updated to make options contain the ProwJob queue, made registry manager explicit in initialization, updated queue to use the correct namespaced name request obejct

alvaroaleman (approved):

PR #1394, opened 2020-11-06
Ci-Operator: Replace some more clients

alvaroaleman:


pkg/steps/multi_stage.go
```diff
@@ -144,8 +127,8 @@ func (s *multiStageTestStep) Run(ctx context.Context) error {
 func (s *multiStageTestStep) run(ctx context.Context) error {
 	var env []coreapi.EnvVar
 	if s.profile != "" {
-		secret := s.profileSecretName()
-		if _, err := s.secretClient.Secrets(s.jobSpec.Namespace()).Get(context.TODO(), secret, meta.GetOptions{}); err != nil {
+		secret := &coreapi.Secret{}
+		if err := s.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: s.jobSpec.Namespace(), Name: s.profileSecretName()}, secret); err != nil {
 			return fmt.Errorf("could not find secret %q: %w", secret, err)
```
bbguimaraes, line 132:
```suggestion
		secret := s.jobSpec.Namespace()
		if err := s.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: secret, Name: s.profileSecretName()}, &coreapi.Secret{}); err != nil {
			return fmt.Errorf("could not find secret %q: %w", secret, err)
```

alvaroaleman: updated

cmd/ci-operator/main.go
```diff
@@ -704,10 +708,11 @@ func (o *options) resolveInputs(steps []api.Step) error {
 	o.jobSpec.SetNamespace(o.namespace)
 
 	//If we can resolve the field, use it. If not, don't.
-	if routeGetter, err := routeclientset.NewForConfig(o.clusterConfig); err != nil {
-		log.Printf("could not get route client for cluster config")
+	if client, err := ctrlruntimeclient.New(o.clusterConfig, ctrlruntimeclient.Options{}); err != nil {
```
droslean, line 713: Why not just add all the required schemes in Options here instead of adding schemes to the runtime in `addSchemes` etc.. ?

alvaroaleman: Then I would need to create a new scheme everytime I construct a client, using the global scheme allows me to do it once and be done

droslean: Also, another concern is why we need to add those schemes in runtime?

alvaroaleman: Because otherwise we don't have it. The typed clients do this in an `init()`, e.G. https://github.com/openshift/ci-tools/blob/76e84167a77c09851dba6de95b46f95ecaeadc60/vendor/github.com/openshift/client-go/template/clientset/versioned/scheme/register.go#L39

bbguimaraes:
Changes LGTM.  Out of curiosity, why were the watchers replaced with polling?

alvaroaleman:
> Changes LGTM. Out of curiosity, why were the watchers replaced with polling?

Because the client doesn't expose watch, as its pretty low-level. There is an implementation that is cache-backed though, if we care.

bbguimaraes:
/lgtm
/hold

if there are still outstanding questions.

PR #1312, opened 2020-10-15
ci-operator: add option to configure leases

bbguimaraes:
Documented in https://github.com/openshift/ci-docs/pull/13.

PR #1403, opened 2020-11-10
pkg/defaults: only process `Provides` once

bbguimaraes:
Part of https://github.com/openshift/ci-tools/pull/1362 (which in turn is part
of https://github.com/openshift/ci-tools/pull/1312).  Opened separately to limit
the number of changes in a single PR.

See https://github.com/openshift/ci-tools/pull/1362#issuecomment-723284065 for
more details.  Due to refactorings in `pkg/defaults`, we eventually started
processing `Provides` more than once for some types of build steps.  This does
not appear to have a negative effect, but is certainly wrong and misleading.

### DIFF
--- a/pkg/api/parameters.go
+++ b/pkg/api/parameters.go
@@ -2,6 +2,7 @@ package api
 
 import (
 	"fmt"
+	"log"
 	"os"
 	"sync"
 )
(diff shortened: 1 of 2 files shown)


alvaroaleman (approved):

stevekuznetsov:
Part of DPTP-1365

PR #1405, opened 2020-11-10
Bump controller-runtime and kube dependencies

alvaroaleman:

Main differences:
* Client now takes a client.Object or client.ObjectList, both are aggregate interfaces of metav1.Object(,List) and runtime.Object
* Some more methods exposed on the client
* Same for handlers
* Reconciler interface was changed to take a `context.Context` as first arg, getting rid of the requirement to have one stored in the Reconciler

~The main motivation for this is the [`NamespaceWrapper`](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/client#NewNamespacedClient), because in contrast to what I had built downstream it can cope with both namespaced and non-namespaced objects. This allows for example to configure it in the beginning in the ci-operator and not require each step to potentially add the wrapper, depending on if a certain funcs expects a namespaced client or not~

Doesn't work at all, as we don't even know the namespace in pkg/defaults and because we also interact with other namespaces.


alvaroaleman:
Needs https://github.com/openshift/release/pull/13537

hack/lib/integration.sh
```diff
@@ -114,8 +114,8 @@ readonly -f os::integration::configresolver::wait_for_registry_update
 
 # os::integration::configresolver::check_log searches for errors in the log.
 function os::integration::configresolver::check_log() {
-    if grep -qE "level=(error|fatal)" "${LOG_DIR}/configresolver.log"; then
-        grep -E "level=(error|fatal)" "${LOG_DIR}/configresolver.log"
+    if cat ${LOG_DIR}/configresolver.log|grep -v 'Failed to initialize commentMap'|grep -qE "level=(error|fatal)"; then
```
stevekuznetsov, line 117: Why `cat`? Just pass the file to `grep` like it used to

PR #1410, opened 2020-11-11
Allow custom releases in ci-operator validations

GeoBK:


pkg/api/config.go
```diff
@@ -684,7 +684,18 @@ func validateLiteralTestStepCommon(fieldRoot string, step LiteralTestStep, seen
 				switch obj {
 				case PipelineImageStream, ReleaseStreamFor(LatestReleaseName), ReleaseStreamFor(InitialReleaseName), ReleaseImageStream:
 				default:
-					ret = append(ret, fmt.Errorf("%s.from: unknown imagestream '%s'", fieldRoot, imageParts[0]))
+					releaseParts := strings.Split(obj, "-")
```
stevekuznetsov, line 687: You should just use `ReleaseNameFrom`

pkg/api/config.go
```diff
@@ -684,7 +684,18 @@ func validateLiteralTestStepCommon(fieldRoot string, step LiteralTestStep, seen
...
+					if len(releaseParts) != 2 || (len(releaseParts) == 2 && releaseParts[0] != StableImageStream) {
+						ret = append(ret, fmt.Errorf("%s.from: '%s' is not a valid imagestream reference", fieldRoot, step.From))
+					} else {
+						names := sets.NewString()
```
stevekuznetsov, line 691: If all we need is the set, why pass the full releases map through the call chain?

GeoBK: I would expect that performance-wise there wouldn't be much difference since its probably just a pointer being passed around?
If so I would prefer to create the set near where its actually being used (unless it is being reused somewhere else)

stevekuznetsov:
Implements DPTP-1671

pkg/api/config.go
stevekuznetsov: Performance-wise, creating new data structures (the set) on every validation is certainly worse.

The stronger argument, though, is that you should not pass data that is not needed. The callers need to know a list of valid releases (or, even, a closure that could validate an ImageStream) and not anything else. Exposing more data than necessary to other parts of a program leads to bad encapsulation and creep.

stevekuznetsov (approved):

PR #1411, opened 2020-11-12
Ensure ImageStreamSpec while sync images

hongkailiu:
```
   dockerImageRepository	<string>
     dockerImageRepository is optional, if specified this stream is backed by a
     container repository on this server Deprecated: This field is deprecated as
     of v3.7 and will be removed in a future release. Specify the source for the
     tags to be imported in each tag via the spec.tags.from reference instead.

   lookupPolicy	<Object>
     lookupPolicy controls how other resources reference images within this
     namespace.

   tags	<[]Object>
     tags map arbitrary string values to specific image locators
```

We copied `lookupPolicy` already and `dockerImageRepository` is deprecated.
Maybe keep the whole `tags` as the sourceImageStream which (has the latest isTag)?

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -279,9 +279,6 @@ func imagestream(imageStream *imagev1.ImageStream) (*imagev1.ImageStream, crcont
 			}
 		}
 		stream.Spec.LookupPolicy.Local = imageStream.Spec.LookupPolicy.Local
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):

stevekuznetsov (commented):
Might be safer to blindly copy everything instead of trying to pull out specific fields in spec and annotations etc but this seems ok if we're doing it piece by piece

PR #1413, opened 2020-11-12
CI-Operator: Use runtimeclient for pods

alvaroaleman:

ref https://issues.redhat.com/browse/DPTP-1523


pkg/steps/artifacts.go
```diff
@@ -661,14 +637,13 @@ func hasMountsArtifactsVolume(pod *coreapi.Pod) bool {
 
 func gatherContainerLogsOutput(podClient PodClient, artifactDir, namespace, podName string) error {
 	var validationErrors []error
-	list, err := podClient.Pods(namespace).List(context.TODO(), meta.ListOptions{FieldSelector: fields.Set{"metadata.name": podName}.AsSelector().String()})
-	if err != nil {
+	pod := &coreapi.Pod{}
+	if err := podClient.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: namespace, Name: podName}, pod); err != nil {
```
stevekuznetsov, line 641: This is not a list. Why?

pkg/steps/multi_stage.go
```diff
@@ -72,20 +69,18 @@ func MultiStageTestStep(
...
+	return newMultiStageTestStep(testConfig, config, params, client, artifactDir, jobSpec)
 }
 
 func newMultiStageTestStep(
 	testConfig api.TestStepConfiguration,
 	config *api.ReleaseBuildConfiguration,
 	params api.Parameters,
-	podClient PodClient,
-	client ctrlruntimeclient.Client,
+	client PodClient,
```
stevekuznetsov, line 83: I don't understand this change - was the `ctrlruntimeclient` unused before?

pkg/steps/multi_stage.go
```diff
@@ -323,7 +317,7 @@ func (s *multiStageTestStep) runSteps(
 	select {
 	case <-ctx.Done():
 		log.Printf("cleanup: Deleting pods with label %s=%s", MultiStageTestLabel, s.name)
-		if err := deletePods(s.podClient.Pods(s.jobSpec.Namespace()), s.name); err != nil {
+		if err := deletePods(ctrlruntimeclient.NewNamespacedClient(s.client, s.jobSpec.Namespace()), s.name); err != nil {
```
stevekuznetsov, line 320: Can we just remove the `deletePods` func entirely?

pkg/steps/template.go
```diff
@@ -490,7 +489,7 @@ func waitForPodCompletion(ctx context.Context, podClient coreclientset.PodInterf
 		case <-ctxDone:
 		default:
 			skipLogs = true
-			if !retry || err == nil {
```
stevekuznetsov, line 493: Why are we changing this?

pkg/steps/template.go
```diff
@@ -521,84 +518,65 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
...
 		select {
 		case <-done:
-			return pod, false, ctx.Err()
-		// Check minutely if we ran into the pod timeout
+			return pod, ctx.Err()
 		case <-podCheckTicker.C:
-			newPod, err := podClient.Get(context.Background(), name, meta.GetOptions{})
-			if err != nil {
+			if err := podClient.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: namespace, Name: name}, pod); err != nil {
+				if kerrors.IsNotFound(err) {
```
stevekuznetsov, line 559: This looks like a new check - why?

pkg/steps/template.go
```diff
@@ -521,84 +518,65 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient coreclientset.
 	//   the state of the pod is correctly reported, artifacts are gathered,
 	//   and termination happens deterministically for both success and failure
 	//   scenarios.
-	watcher, err := podClient.Watch(context.Background(), meta.ListOptions{
```
stevekuznetsov, line 524: Why did we lose this watch?

stevekuznetsov (commented):
The `template.go` changes here are hard to review and seem wrong? We shouldn't change function in these PRs, just types...

pkg/steps/artifacts.go
alvaroaleman: Because a list with a `metadata.Name` selector followed by a `len(items) == 0` check is just a complicated way of doing a `Get`

pkg/steps/multi_stage.go
alvaroaleman: the `PodClient` is a superinterface of the `ctrlruntimeclient` and the pod extension interface now, so we can just use it where we used the `ctrlruntimeclient` before

pkg/steps/template.go
alvaroaleman: Because there is no retry without a watch

pkg/steps/template.go
alvaroaleman: Because its not supported by the client to open a raw watch. We could use a cache-backed client but I am not sure its worth the effort. This has the nice sideffect that writing a warpper is possible and writing tests is easier

alvaroaleman (commented):
> The template.go changes here are hard to review and seem wrong? We shouldn't change function in these PRs, just types...

Can you be more precise about what seems wrong?

alvaroaleman:
/retest
pods didn't start running :upside_down_face:

pkg/steps/template.go
alvaroaleman: it isn't, check above

pkg/steps/artifacts.go
stevekuznetsov: Maybe, maybe there was some reason to do it this way - I want this PR to be a client change, not a functionality change please

pkg/steps/template.go
stevekuznetsov: Wait so what is the replacement for the watch? I am confused

pkg/steps/artifacts.go
alvaroaleman: It is a different way to do the exact same thing

pkg/steps/template.go
alvaroaleman: There is no replacement for the watch on the client level because you are not supposed to hand-manage watches. You can make the client itself cache-backed in which case reading requests will get served from a cache that is in turn backed by a standard SharedIndexInformer

PR #1420, opened 2020-11-16
testgrid-config-generator: disallow ignored allow-list entries

stevekuznetsov:
When a job is listed as required in the `release-controller`
configuration, it will be added to a blocking dashboard and it is not
possible to overwrite that behavior with an entry in the allow-list. It
is not therefore useful to allow users to put any such jobs into the
allow-list in the first place.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -33,26 +33,21 @@ type options struct {
 
 	validationOnlyRun bool
 	jobsAllowListFile string
-	withoutAllowList  bool
 }
 
 func (o *options) Validate() error {
-	if o.prowJobConfigDir == "" {
+	if o.prowJobConfigDir == "" && !o.validationOnlyRun {
 		return errors.New("--prow-jobs-dir is required")
 	}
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
This is part of DPTP-1739

cmd/testgrid-config-generator/main.go
```diff
@@ -270,6 +254,31 @@ func main() {
...
+	var allowList map[string]string
+	if o.jobsAllowListFile != "" {
+		data, err := ioutil.ReadFile(o.jobsAllowListFile)
+		if err != nil {
+			logrus.WithError(err).Fatalf("could not read allow-list at %s", o.jobsAllowListFile)
+		}
+		allowList, err = getAllowList(data)
+		if err != nil {
+			logrus.WithError(err).Fatalf("failed to build allow-list dictionary")
+		}
```
GeoBK, line 267: I think the initial reasoning behind keeping this bit before the validation function was so that we could get away with not passing the some parameters (the testgrid directory I think) if we were running it in the validationOnly mode. Why would we not need that any longer?

PR #1422, opened 2020-11-16
repo-init: allow users basic custom release choice

stevekuznetsov:
When a user is not promoting with the OpenShift release we don't put
them into the CI stream with `tag_specification`, so we can ask them
some basic questions about the release they want to run e2es with. We
don't expose every possible option as that's par for the course for this
simple bootstrapping tool.

fixes [DPTP-1722](https://issues.redhat.com/browse/DPTP-1722)


stevekuznetsov:
cc @petr-muller

petr-muller (approved):

PR #1426, opened 2020-11-18
ci-secret-bootstrap: validate config

hongkailiu:
The added validation step will fail the run as early as possible.
Otherwise, it will be rejected by k8s server eventually.

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -54,12 +54,15 @@ type options struct {
 	secretsGetters map[string]coreclientset.SecretsGetter
 	config         secretbootstrap.Config
 	bwAllowUnused  flagutil.Strings
+
+	validateOnly bool
 }
 
 func parseOptions() options {
 	var o options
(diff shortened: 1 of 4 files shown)


hongkailiu:
/assign @alvaroaleman 

I still think https://github.com/openshift/release/pull/13668 is a simpler implementation if we only want to have a presubmit to validate the config with this case.

Maybe you want to include something else?

Other concern:
We run secret-job periodically and show the logs in internal deck to avoid leaking expectedly.
It might be odd to see a link in the PR of release repo (which is public) and the log page is not accessible.
We could make another tool with GO to validate the yaml config. Is it worth the effort?
WDYT?

stevekuznetsov (commented):
Keying off of `o.validateOnly` inside of `completeOptions()` to create a set of options that will validate correctly when empty seems fragile  and will require more work in the future if this tool grows. Can we, at the `main()` level, just load and validate the config (and only that) when the flag is set?

stevekuznetsov (approved):

PR #1429, opened 2020-11-19
pj-rehearse: no longer detect references to ci-op config CMs

petr-muller:
Previously, pj-rehearse had code that detected when Prow jobs mounted
ci-operator config CMs (`ci-operator-*-configs`) directly, and handled
resolving and inlining this case. OpenShift CI Prowjobs are no longer
supposed to mount ci-op configs like this (and no longer do so [1]), so
all this code is no longer necessary.

[1]
```console
$ git grep 'ci-operator-.*-configs' ci-operator/jobs/
$
```

This code cleanup is part of [DPTP-1685](https://issues.redhat.com/browse/DPTP-1685)


alvaroaleman (approved):
Code removal, I like

PR #1430, opened 2020-11-19
pj-rehearse: support handcrafted jobs with UNRESOLVED_CONFIG

petr-muller:
/hold 
Builds on https://github.com/openshift/ci-tools/pull/1429

This is a thing of beauty! The tests needed to be completely rewritten (there were no real tests for the inlining functionality not working over the CM references -- removed in #1249 ), and the method was refactored - it is probably easier to review as new code than looking at the diff.

There is an annoying gotcha in that our code avoids resolving whole ci-op configs, but it resolves only the test relevant for the job, which is impossible to detect for handcrafted jobs in other way than looking at the `--target=XXX` parameter. Other option was to implement a "if no matching test was found, resolve *all present*, but that changes the behavior more significantly, and breaks multiple integration tests (which are a pinnacle of engineering themselves :trollface: ) so I decided to go for the more targeted approach of inspecting `--target`.

Fixes: [DPTP-1685](https://issues.redhat.com/browse/DPTP-1685)

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -221,13 +221,9 @@ func hasRehearsableLabel(labels map[string]string) bool {
 }
 
 // getResolverConfigForTest returns a resolved ci-operator based on the provided filename and only includes the specified test in the
-// `tests` section of the config.
+// `tests` section of the config. If `testname` is empty, the resolved config will contain all items from the original `tests`.
 // The ImageStreamTagMap contains all imagestreamtags used within this config and is used to ensure they exist on all target clusters.
-func getResolvedConfigForTest(ciopConfigs config.DataByFilename, resolver registry.Resolver, filename, testname string) (string, apihelper.ImageStreamTagMap, error) {
-	ciopConfig, ok := ciopConfigs[filename]
-	if !ok {
-		return "", nil, fmt.Errorf("ci-operator config file %s was not found", filename)
-	}
+func getResolvedConfigForTest(ciopConfig config.DataWithInfo, resolver registry.Resolver, testname string) (string, apihelper.ImageStreamTagMap, error) {
 	// make copy so we don't change in-memory config
 	ciopCopy := config.DataWithInfo{
 		Configuration: ciopConfig.Configuration,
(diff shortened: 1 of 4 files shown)


pkg/rehearse/jobs.go
```diff
@@ -266,37 +263,78 @@ func getResolvedConfigForTest(ciopConfigs config.DataByFilename, resolver regist
...
+		case env.Name == "UNRESOLVED_CONFIG" && env.Value != "":
+			if err := yaml.Unmarshal([]byte(env.Value), &ciopConfig.Configuration); err != nil {
+				return nil, fmt.Errorf("failed to unmarshal UNRESOLVED_CONFIG: %w", err)
+			}
+			// Annoying hack: UNRESOLVED_CONFIG means this is a handcrafted job, which means
+			// `testname` cannot be relied on (it is derived from job name, which is arbitrary
+			// in handcrafted jobs). We need the test name to know which `tests` field to
+			// resolve, so we try to detect it from `--target` arg, if present.
+			//
+			// The worst case is that we do not find the matching name. In such case,
```
stevekuznetsov, line 295: Wouldn't `ci-operator` run all tests when no `--target` is present?

pkg/rehearse/jobs.go
```diff
@@ -266,37 +263,78 @@ func getResolvedConfigForTest(ciopConfigs config.DataByFilename, resolver regist
...
-		envs = append(envs, env)
-		container.Env = envs
+	ciOpConfigContent, imageStreamTags, err := getResolvedConfigForTest(ciopConfig, resolver, testname)
+	if err != nil {
+		loggers.Job.WithError(err).Error("Failed to get resolved config for test")
+		return nil, err
 	}
+	apihelper.MergeImageStreamTagMaps(allImageStreamTags, imageStreamTags)
+	container.Env = append(envs, v1.EnvVar{Name: "CONFIG_SPEC", Value: ciOpConfigContent})
 	return allImageStreamTags, nil
```
stevekuznetsov, line 338: Could we update the godoc to explain why a method that is supposed to in-line config returns a map of imagestreamtags?

petr-muller: heh, I'll need to figure out that information myself, but yeah

pkg/rehearse/jobs.go
petr-muller: Hmm I thought it would just build all artifacts and run no tests. How much would we care about that? I'm a bit worried about hitting some content limit with full resolution, I suppose someone added that optimization for a reason.

AlexNPavel:
We have a lot of jobs that use environment variables in the target and/or the `UNRESOLVED_CONFIG` itself. Could we run a quick substitute over the `target` and config using the env vars set for the container? Here is an example of a job that does this: https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-release-4.7-periodics.yaml#L5640:L5687

Both the `BRANCH` and `CLUSTER_TYPE` are defined as env vars in the container spec, so it should be as simple as running a `for` loop over those env vars and substituting all `$(var_name)` with the corresponding value.

petr-muller:
> Both the BRANCH and CLUSTER_TYPE are defined as env vars in the container spec, so it should be as simple as running a for loop over those env vars and substituting all $(var_name) with the corresponding value.

@AlexNPavel I was thinking about how to handle this best and I'd like to avoid getting into some weird middle state where we would substitute variables "somewhat like kube but not quite". Maybe there is code we could reuse that does this for pods in kube? Not sure. But I'd like to tackle this in a dedicated followup.

petr-muller:
@stevekuznetsov addressed, PTAL

stevekuznetsov (approved):

PR #1432, opened 2020-11-19
registrysyncer: support denied imagestreams

hongkailiu:
Follow up https://coreos.slack.com/archives/CBN38N3MW/p1605814866021000?thread_ts=1605798608.421500&cid=CBN38N3MW

`origin/release` and `ocp/release` are special.
We do not touch them for now.
We will have to handle those specially when we do registry migration to app.ci.

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -37,6 +37,7 @@ func AddToManager(mgr manager.Manager,
 	imageStreams sets.String,
 	imageStreamPrefixes sets.String,
 	imageStreamNamespaces sets.String,
+	deniedImageStreams sets.String,
 ) error {
 	log := logrus.WithField("controller", ControllerName)
 	r := &reconciler{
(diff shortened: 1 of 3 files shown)


alvaroaleman (commented):
why a new config option rather than not adding them to the config?

hongkailiu:
/close

Waiting for the way out of the mess.

hongkailiu:
/reopen

Because we still on to sync everything in ocp and origin except release.

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -312,20 +315,24 @@ func testInputImageStreamTagFilterFactory(
 	imageStreams sets.String,
 	imageStreamPrefixes sets.String,
 	imageStreamNamespaces sets.String,
+	deniedImageStreams sets.String,
 ) objectFilter {
 	l = logrus.WithField("subcomponent", "test-input-image-stream-tag-filter")
 	return func(nn types.NamespacedName) bool {
+		imageStreamName, err := imageStreamNameFromImageStreamTagName(nn)
```
stevekuznetsov, line 322: this returns `namespace/name` ?

hongkailiu: It returns `types.NamespacedName`. Then we compare with `imageStreamName.String()` which is form of `namespace/name`.

PR #1434, opened 2020-11-20
Repo brancher: Retry git commands

alvaroaleman:
This PR factors out the git command execution into a func that does
retrying. It also moves all the command-related logging into that func.

ref https://issues.redhat.com/browse/DPTP-1625

/assign @droslean

### DIFF
--- a/cmd/repo-brancher/main.go
+++ b/cmd/repo-brancher/main.go
@@ -12,6 +12,7 @@ import (
 	"path"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
@@ -117,6 +118,8 @@ func main() {
 			return nil
 		}
 
+		executeGitCmd := executeGitCMDFactory(repoDir)
+
 		remote, err := url.Parse(fmt.Sprintf("https://github.com/%s/%s", repoInfo.Org, repoInfo.Repo))
 		if err != nil {
 			logger.WithError(err).Fatal("Could not construct remote URL.")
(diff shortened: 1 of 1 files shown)


cmd/repo-brancher/main.go
```diff
@@ -235,3 +204,33 @@ func main() {
...
+			}
+			break
+		}
+		l = l.WithField("output", string(b))
+		if err != nil {
+			l.Error("Failed to execute command.")
+		} else {
+			l.Debug("Executed command.")
+		}
+		return err
```
droslean, line 234: Can we just do this instead ?
```suggestion
		if err != nil {
			l.Error("Failed to execute command.")
			return err
		} 
		l.Debug("Executed command.")
		return nil
```

alvaroaleman: good point, updated

droslean (approved):

cmd/repo-brancher/main.go
```diff
@@ -151,66 +147,39 @@ func main() {
...
-			cmd.Dir = repoDir
-			cmdLogger.Debug("Running command.")
-			if out, err := cmd.CombinedOutput(); err != nil {
-				cmdLogger.WithError(err).WithFields(logrus.Fields{"output": string(out)}).Error("Failed to execute command.")
+			if err := executeGitCmd(logger, command...); err != nil {
 				failed = true
 				continue
-			} else {
-				cmdLogger.WithFields(logrus.Fields{"output": string(out)}).Debug("Executed command.")
-				if string(out) == "" && !o.fastForward {
```
wking, line 166: This drops the only `fastForward` consumer that I can find.  Should we deprecate `--fast-forward`, drop it from openshift/release, and then drop the command-line option here?

PR #1437, opened 2020-11-20
registrysyncer: use isTag instead of isImport

hongkailiu:
* replace ISImport with ISTag: the import will be created by openshift. Unlike imports, the creation of isTag will be async. This will help to digest the queue. The disadvantage is that it is harder to check the import status (See the TODO).
* Copy the annotation over on the isTags: This is why we do this PR. The release controller replies on the annotation on isTags to do things.
* if found soft-delete annotation on any IS, delete the IS on all clusters. https://github.com/openshift/release-controller/pull/233

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -150,7 +150,7 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 	if n := len(imageStreamNameAndTag); n != 2 {
 		return fmt.Errorf("when splitting imagestreamtagname %s by : expected two results, got %d", req.Name, n)
 	}
-	imageStreamName, imageTag := imageStreamNameAndTag[0], imageStreamNameAndTag[1]
+	imageStreamName := imageStreamNameAndTag[0]
 	isName := types.NamespacedName{Namespace: req.Namespace, Name: imageStreamName}
 	sourceImageStream := &imagev1.ImageStream{}
 	if err := r.registryClients[srcClusterName].Get(ctx, isName, sourceImageStream); err != nil {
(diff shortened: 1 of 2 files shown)


hongkailiu:
/hold

It is now late afternoon on Friday.

stevekuznetsov (commented):
I don't see the annotations in a test anywhere?

hongkailiu:
> I don't see the annotations in a test anywhere?

https://github.com/openshift/ci-tools/pull/1437/files#diff-8064718ecb2cbf89e0f7be9cd189ec664b54a997fee0b5fbbb820a546ccf1c72R352

hongkailiu:
If we merge this, it should help to resolve the queue.
But we need a way to watch it because imports will be implicitly done by openshift (but they could have more workers than us).

https://prometheus-prow-monitoring.apps.ci.l2s4.p1.openshiftapps.com/graph?g0.range_input=3d&g0.expr=workqueue_depth%7Bjob%3D%22dptp-controller-manager%22%2Cname%3D%22registry_syncer%22%2Cnamespace%3D%22ci%22%7D&g0.tab=0

Screen Shot 2020-11-23 at 8 59 08 AM

stevekuznetsov (commented):
The OpenShift ImageStream controller just creates ImageStreamImport objects from the ImageStreamTag spec fields so I'm not sure we really gain anything by trying to track import. This otherwise LGTM. @bparees would you mind approving that the structure of the tag will be appropriate?

bparees:
what you're doing here doesn't look like an import at all, since you're not setting an istag.From field?

and i would not expect you to set istag.image, that should be set by the import process once it happens.

stevekuznetsov:
@bparees  right -- this is the PR to change us from creating an `ImageStreamImport` to instead create and `ImageStreamSpec` value in `Tags` and allow the `ImageStream` controller to do the work - so we're looking for the correct translation from the `ImageStreamImport` object to the spec field

bparees:
yeah so you should just be setting istag.spec.From and .Name  (and annotations if you want)

stevekuznetsov (approved):

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -151,14 +151,36 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
-	imageStreamName, imageTag := imageStreamNameAndTag[0], imageStreamNameAndTag[1]
+	imageStreamName := imageStreamNameAndTag[0]
 	isName := types.NamespacedName{Namespace: req.Namespace, Name: imageStreamName}
 	sourceImageStream := &imagev1.ImageStream{}
 	if err := r.registryClients[srcClusterName].Get(ctx, isName, sourceImageStream); err != nil {
 		// received a request on the isTag, but the 'is' is no longer there
 		return fmt.Errorf("failed to get imageStream %s from cluster %s: %w", isName.String(), srcClusterName, err)
 	}
 
+	// if found soft-delete annotation on any IS, delete the IS on all clusters
```
hongkailiu, line 162: If the delete failed for any reason: sync (reconcile), annotation (by release-controller), and then delete (reconcile).

stevekuznetsov:
Can we separate the soft-delete logic? If we get the ISTag sync in, we can also have the option of just ignoring deletion. This means that the `release-controller` won't be able to delete things while it's being migrated, but if we make the migration period short, it should not matter.

stevekuznetsov:
(if soft-delete is harder to acheive than simple ISTag)

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -326,20 +306,29 @@ func dockerImageImportedFromTargetingCluster(cluster string, tag *imagev1.ImageS
 	return strings.HasPrefix(from, api.ServiceDomainAPICIRegistry) && cluster == "api.ci" || strings.HasPrefix(from, api.ServiceDomainAPPCIRegistry) && cluster == "app.ci"
 }
 
-func createAndCheckStatus(ctx context.Context, client ctrlruntimeclient.Client, imageStreamImport *imagev1.ImageStreamImport) error {
-	// ImageStreamImport is not an ordinary api but a virtual one that does the import synchronously
-	if err := client.Create(ctx, imageStreamImport); err != nil {
-		return fmt.Errorf("failed to import Image: %w", err)
-	}
+func (r *reconciler) ensureImageStreamTag(ctx context.Context, imageStreamTag *imagev1.ImageStreamTag, dockerImageReference string, client ctrlruntimeclient.Client, log *logrus.Entry) error {
+	secret, mutateFn := imagestreamtag(imageStreamTag, dockerImageReference)
```
bparees, line 310:
```suggestion
	istag, mutateFn := imagestreamtag(imageStreamTag, dockerImageReference)
```

?

hongkailiu: Corrected.

stevekuznetsov (approved):

PR #1439, opened 2020-11-23
registrySyncer: #workers=100

hongkailiu:
The queue got longer over timve.
https://user-images.githubusercontent.com/4013349/99980877-b9b57580-2d76-11eb-8c50-e3f0327c20e0.png

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -51,17 +52,19 @@ func AddToManager(mgr manager.Manager,
 		imageStreamPrefixes:   imageStreamPrefixes,
 		imageStreamNamespaces: imageStreamNamespaces,
 		deniedImageStreams:    deniedImageStreams,
+		imageStreamLocks: &shardedLock{
+			mapLock: &sync.Mutex{},
+			locks:   map[simpleImageStream]*sync.Mutex{},
+			log:     log,
+		},
 	}
+	r.imageStreamLocks.runCleanup()
 	for clusterName, m := range managers {
 		r.registryClients[clusterName] = imagestreamtagwrapper.MustNew(m.GetClient(), m.GetCache())
 	}
 	c, err := controller.New(ControllerName, mgr, controller.Options{
-		Reconciler: r,
-		// We conflict on ImageStream level which means multiple request for imagestreamtags
-		// of the same imagestream will conflict so stay at one worker in order to reduce the
-		// number of errors we see. If we hit performance issues, we will probably need cluster
-		// and/or imagestream level locking.
-		MaxConcurrentReconciles: 1,
+		Reconciler:              r,
+		MaxConcurrentReconciles: 100,
 	})
 	if err != nil {
 		return fmt.Errorf("failed to construct controller: %w", err)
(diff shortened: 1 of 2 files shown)


pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -54,12 +54,8 @@ func AddToManager(mgr manager.Manager,
...
 	}
 	c, err := controller.New(ControllerName, mgr, controller.Options{
-		Reconciler: r,
-		// We conflict on ImageStream level which means multiple request for imagestreamtags
-		// of the same imagestream will conflict so stay at one worker in order to reduce the
-		// number of errors we see. If we hit performance issues, we will probably need cluster
-		// and/or imagestream level locking.
-		MaxConcurrentReconciles: 1,
+		Reconciler:              r,
+		MaxConcurrentReconciles: 100,
```
alvaroaleman, line 58: You need to synchronize this on imagestream level to prevent races

alvaroaleman: i.E. at any time, at most one worker can run per imagestream

alvaroaleman: See the comment you removed for an explanation

alvaroaleman: /hold

hongkailiu: Stole the code from https://github.com/kubernetes/test-infra/blob/dbc92a6558ab6ae3fc8ef5d10a2bb629f8f0d26a/prow/crier/reporters/github/reporter.go#L141-L149

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -78,6 +81,55 @@ func AddToManager(mgr manager.Manager,
...
+		delete(s.locks, key)
+		lock.Unlock()
+	}
+}
+
+// runCleanup asynchronously runs the cleanup once per hour.
+func (s *shardedLock) runCleanup() {
+	go func() {
+		for range time.Tick(time.Hour) {
+			logrus.Debug("Starting to clean up presubmit locks")
```
alvaroaleman, line 125: Please use a logger that has the controller name and update the logs here and the godoc comment above to not refer to "presubmits" anymore, that will be wildly confusing later on :)

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -216,17 +269,20 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
 		}
 		controllerutil.CountImportResult(ControllerName, clusterName, req.Namespace, true)
 		log.Debug("Imported successfully")
 	}
 	return nil
 }
 
-func createAndCheckStatus(ctx context.Context, client ctrlruntimeclient.Client, imageStreamImport *imagev1.ImageStreamImport) error {
+func (r *reconciler) createAndCheckStatus(ctx context.Context, key simpleImageStream, client ctrlruntimeclient.Client, imageStreamImport *imagev1.ImageStreamImport) error {
+	lock := r.imageStreamLocks.getLock(key)
```
alvaroaleman, line 283: Please remove the `cluster` part from the lock and acquire/release it and the beginning/end of the reconciliation:
* It is much easier to understand
* It makes sure that we will either get to run the whole reconciliation or nothing, instead of possibly getting stuck waiting for the lock halfway through


A no-op reconciliation is extremely fast as its entirely based on in-memory data, so I don't think this will be a performance issue

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -164,6 +165,7 @@ func init() {
 }
 
 func TestReconcile(t *testing.T) {
+	t.Parallel()
```
alvaroaleman, line 168: if you enable t.Parallel(), you need to copy the `tc` variable in the loop below:
```
tc := tc
```

Because go will reuse the memory of the variable for each loop interation, which means that if the loop body is now in its own goroutine, we modify the `tc` data without waiting for the testcase to finish

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Tried this way while coding. Did not think this through. Cool.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Done.

alvaroaleman:
Can we defer merging this until we see the effect of not re-importing?

While this is a nice performance optimization, it shouldn't be really needed and might hide bugs, I would expect that even without this we will (slowly) reduce our import backlog

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -172,6 +227,11 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		return fmt.Errorf("failed to ensure finalizer to %s from cluster %s: %w", isName.String(), srcClusterName, err)
 	}
 
+	key := simpleImageStream{NamespacedName: types.NamespacedName{Namespace: req.Namespace, Name: imageStreamName}}
```
alvaroaleman, line 230: This should be at the very beginning of the reconciliation

alvaroaleman (approved):

PR #1446, opened 2020-11-24
ci-operator: don't cancel notifiers on interrupt

stevekuznetsov:
Today, we cancel the notifier in order to stop a very specific style of
deadlock:

When we run a test workload that hangs forever upon receipt of SIGINT,
the artifacts will not be collected for that Pod as we wait for all
containers to finish before starting the artifact retrieval. If the
`ci-operator` process cannot wait for all artifacts to be retrieved,
nothing is uploaded to GCS.

If a test process is well-formed and exits early before the grace period
is up but after an interrupt is sent, however, the artifact retrieval
process should complete correctly and no explicit cancellation will be
necessary.

However, the side effect of cancelling the notifier explicitly as soon
as an interrupt is received is that we will never retrieve artifacts for
a Pod even if it is well formed and exits before the grace period.

By not cancelling the notifier, we make a trade-off: for poorly-formed
test processes, no artifacts whatsoever will be present in the output of
the job as `ci-operator` will hang, waiting on the artifacts container
(which is waiting on the test process). However, in cases where the
process *is* well-formed, we will get artifacts on an interrupt or
timeout. This latter case is by far the most common case in our jobs
today and the one we should be optimizing for.

/assign @bbguimaraes

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -401,7 +401,7 @@ type ArtifactWorker struct {
 	hasArtifacts sets.String
 }
 
-func NewArtifactWorker(ctx context.Context, podClient PodClient, artifactDir, namespace string) *ArtifactWorker {
+func NewArtifactWorker(podClient PodClient, artifactDir, namespace string) *ArtifactWorker {
 	// stream artifacts in the background
 	w := &ArtifactWorker{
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Needed for DPTP-1668

PR #1447, opened 2020-11-24
registrySyncer: No importing back

hongkailiu:

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -166,6 +166,10 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		if clusterName == srcClusterName {
 			continue
 		}
+		if dockerImageImportedFromTargetingCluster(clusterName, sourceImageStreamTag) {
+			log.Debug("dockerImage imported from targeting cluster")
+			continue
+		}
 		if err := client.Get(ctx, types.NamespacedName{Name: req.Namespace}, &corev1.Namespace{}); err != nil {
 			if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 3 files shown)


pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -235,7 +235,8 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		log.Debug("creating imageStreamImport")
 		if err := createAndCheckStatus(ctx, client, imageStreamImport); err != nil {
 			controllerutil.CountImportResult(ControllerName, clusterName, req.Namespace, imageStreamName, false)
-			return fmt.Errorf("failed to create and check the status for imageStreamImport on cluster %s %v : %w", clusterName, imageStreamImport, err)
+			//no return of the error to avoid staying in the queue forever
+			log.WithError(err).Errorf("failed to create and check the status for imageStreamImport for tag %s of %s in namespace %s on cluster %s", imageTag, imageStreamImport.Name, imageStreamImport.Namespace, clusterName)
```
alvaroaleman, line 239: This kind of defeats the purpose of using a controller in the first place, as we will never retry on transient errors. Also, metrics are wrong

alvaroaleman: I think what we should be doing is if the src import had any errors and skip the ones where it had errors

hongkailiu: Implemented as discussed in slack: no importing back to where the image came from.

hongkailiu: >metrics are wrong

Can you elaborate?

alvaroaleman (approved):

PR #1448, opened 2020-11-24
Steps: Track and expose all objects accessed

alvaroaleman:
Ref: https://issues.redhat.com/browse/DPTP-1703

This pr extends the steps to track all objects they use. This is achieved by:
* Adding a loggingclient that will track all objects requested through it
* Use that loggingclient throughout the steps
* Extend the steps interface with an `Objects() []ctrlruntimeclient.Object` method that gets populated via that logging client


pkg/steps/loggingclient/loggingclient.go
```diff
@@ -0,0 +1,121 @@
...
+	}
+	lc.logObject(obj)
+	return nil
+}
+
+func (lc *loggingClient) Delete(ctx context.Context, obj ctrlruntimeclient.Object, opts ...ctrlruntimeclient.DeleteOption) error {
+	if err := lc.upstream.Delete(ctx, obj, opts...); err != nil {
+		return err
+	}
+	lc.logObject(obj)
```
stevekuznetsov, line 59: Does this log the object that we had in memory which we used to issue the Delete call or does Delete respond with something? The glaring issue with our current artifacts is that when we Delete we get no object, and this code will at least give us "some" version of that object, but it would be amazing to get the *latest* version of it so we can debug with status, etc.

alvaroaleman: err good point, that does indeed not work. We could just get it first here, wdyt?

alvaroaleman: Updated the client to get -> log -> delete inside its delete

petr-muller (approved):

PR #1449, opened 2020-11-24
Promotionstep: Remove unused src/destClient

alvaroaleman:
The step has a src, dst and PodClient but all of them are backed by the same client, making the whole thing confusing and useless.

### DIFF
--- a/pkg/steps/release/promote.go
+++ b/pkg/steps/release/promote.go
@@ -31,10 +31,8 @@ type promotionStep struct {
 	config         api.PromotionConfiguration
 	images         []api.ProjectDirectoryImageBuildStepConfiguration
 	requiredImages sets.String
-	srcClient      ctrlruntimeclient.Client
-	dstClient      ctrlruntimeclient.Client
 	jobSpec        *api.JobSpec
-	podClient      steps.PodClient
+	client         steps.PodClient
 	pushSecret     *coreapi.Secret
 }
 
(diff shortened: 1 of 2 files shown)


pkg/steps/release/promote.go
```diff
@@ -94,11 +92,11 @@ func (s *promotionStep) run(ctx context.Context) error {
 	if len(s.config.Name) > 0 {
 		return retry.RetryOnConflict(promotionRetry, func() error {
 			is := &imagev1.ImageStream{}
-			err := s.dstClient.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: s.config.Namespace, Name: s.config.Name}, is)
+			err := s.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: s.config.Namespace, Name: s.config.Name}, is)
 			if errors.IsNotFound(err) {
 				is.Namespace = s.config.Namespace
 				is.Name = s.config.Name
-				if err := s.dstClient.Create(ctx, is); err != nil {
+				if err := s.client.Create(ctx, is); err != nil {
```
droslean, line 99: This is a podclient right? This line will create an imagestream.

alvaroaleman: @droslean a pod client looks like this: https://github.com/openshift/ci-tools/blob/e51f853f419026ba115871cfac4c224f7d4b1180/pkg/steps/artifacts.go#L178-L182

It can do CRUD for anything and additionally has methods for the `Exec/GetLogs` pod subresources

stevekuznetsov (approved):
In the past the source and destination clients could be on different clusters, but now that we're pushing via a Pod and `oc` it does not matter.

PR #1455, opened 2020-11-25
Move validation code to its own package

bbguimaraes:
The single file is getting very dense.  Moving to separate files in a separate
package, with a small simplification on top.


bbguimaraes:

cmd/ci-operator/main.go
```diff
@@ -422,7 +423,7 @@ func (o *options) Complete() error {
 		o.jobSpec.Refs.PathAlias = *config.CanonicalGoRepository
 	}
 	o.configSpec = config
-	if err := o.configSpec.ValidateResolved(); err != nil {
+	if err := validation.ValidateResolvedConfiguration(o.configSpec); err != nil {
```
petr-muller, line 426: if we called the package `validate` then methods would not need to contain `Validate` and this would not stutter.

bbguimaraes: Changed public functions to use the same pattern as `k8s.io/apimachinery/pkg/util/validation`, which is where the package name came from.

stevekuznetsov (approved):

PR #1457, opened 2020-11-26
registrySyncer: handle deleted ImageStreams

hongkailiu:
When a src IS is deleted, it will be deleted on other clusters.

We still need to clean up the current mess after merge: delete each IS on app.ci as src IS (or equivalently, has our own finalizer). Then our queue should be back to normal.

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -159,6 +159,18 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		return fmt.Errorf("failed to get imageStream %s from cluster %s: %w", isName.String(), srcClusterName, err)
 	}
 
+	if sourceImageStream.DeletionTimestamp != nil {
+		if err := finalizeIfNeeded(ctx, sourceImageStream, srcClusterName, r.registryClients); err != nil {
+			return fmt.Errorf("failed to finalize %s from cluster %s: %w", isName.String(), srcClusterName, err)
+		}
+		// no sync if the srcIS is deleted
+		return nil
+	}
+
+	if err := ensureFinalizer(ctx, sourceImageStream, r.registryClients[srcClusterName]); err != nil {
+		return fmt.Errorf("failed to ensure finalizer to %s from cluster %s: %w", isName.String(), srcClusterName, err)
+	}
+
 	*log = *log.WithField("docker_image_reference", sourceImageStreamTag.Image.DockerImageReference)
 
 	for clusterName, client := range r.registryClients {
@@ -245,6 +257,57 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		log.Debug("Imported successfully")
 	}
 	return nil
+}
+
+func finalizeIfNeeded(ctx context.Context, stream *imagev1.ImageStream, srcClusterName string, clients map[string]ctrlruntimeclient.Client) error {
+	finalizerSet := sets.NewString(stream.Finalizers...)
+	if !finalizerSet.Has(ControllerName) {
+		return nil
+	}
+	for clusterName, client := range clients {
+		if clusterName == srcClusterName {
+			continue
+		}
+
+		isToDelete := &imagev1.ImageStream{}
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: stream.Name, Namespace: stream.Namespace}, isToDelete); err != nil {
+			if apierrors.IsNotFound(err) {
+				continue
+			}
+			return err
+		}
+
+		if err := ensureRemoveFinalizer(ctx, isToDelete, client); err != nil {
+			return err
+		}
+
+		// populate deleting to all clusters except srcClusterName
+		if err := client.Delete(ctx, isToDelete); err != nil && !apierrors.IsNotFound(err) {
+			return err
+		}
+	}
+
+	// remove our finalizer from the list and update it.
+	return ensureRemoveFinalizer(ctx, stream, clients[srcClusterName])
+}
+
+func ensureRemoveFinalizer(ctx context.Context, stream *imagev1.ImageStream, client ctrlruntimeclient.Client) error {
+	finalizerSet := sets.NewString(stream.Finalizers...)
+	if !finalizerSet.Has(ControllerName) {
+		return nil
+	}
+	originalStream := stream.DeepCopy()
+	stream.Finalizers = finalizerSet.Delete(ControllerName).List()
+	// Use Patch instead of Update to avoid conflicting
+	return client.Patch(ctx, stream, ctrlruntimeclient.MergeFrom(originalStream))
+}
+
+func ensureFinalizer(ctx context.Context, stream *imagev1.ImageStream, client ctrlruntimeclient.Client) error {
+	if sets.NewString(stream.Finalizers...).Has(ControllerName) {
+		return nil
+	}
+	stream.Finalizers = append(stream.Finalizers, ControllerName)
+	return client.Update(ctx, stream)
 }
 
 func dockerImageImportedFromTargetingCluster(cluster string, tag *imagev1.ImageStreamTag) bool {
--- a/pkg/controller/registrysyncer/registrysyncer_test.go
+++ b/pkg/controller/registrysyncer/registrysyncer_test.go
@@ -10,7 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
@@ -80,7 +80,7 @@ func TestPublicDomainForImage(t *testing.T) {
 }
 
 func TestFindNewest(t *testing.T) {
-	now := time.Now()
+	now := metav1.Now()
 	testCases := []struct {
 		name     string
 		isTags   map[string]*imagev1.ImageStreamTag
@@ -99,7 +99,7 @@ func TestFindNewest(t *testing.T) {
 				"cluster1": {
 					Image: imagev1.Image{
 						ObjectMeta: metav1.ObjectMeta{
-							CreationTimestamp: metav1.NewTime(now),
+							CreationTimestamp: now,
 						},
 					},
 				},
@@ -112,7 +112,7 @@ func TestFindNewest(t *testing.T) {
 				"cluster1": {
 					Image: imagev1.Image{
 						ObjectMeta: metav1.ObjectMeta{
-							CreationTimestamp: metav1.NewTime(now),
+							CreationTimestamp: now,
 						},
 					},
 				},
@@ -132,7 +132,7 @@ func TestFindNewest(t *testing.T) {
 				"cluster1": {
 					Image: imagev1.Image{
 						ObjectMeta: metav1.ObjectMeta{
-							CreationTimestamp: metav1.NewTime(now),
+							CreationTimestamp: now,
 						},
 					},
 				},
@@ -181,6 +181,9 @@ func TestReconcile(t *testing.T) {
 		return []byte("some-secret")
 	}
 
+	now := metav1.Now()
+	threeMinLater := metav1.NewTime(now.Add(3 * time.Minute))
+
 	applyconfigISTag := &imagev1.ImageStreamTag{
 		ObjectMeta: metav1.ObjectMeta{
 			Namespace: "ci",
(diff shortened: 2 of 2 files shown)


pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -247,6 +259,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 	return nil
 }
 
+func finalizeIfNeeded(ctx context.Context, stream *imagev1.ImageStream, srcClusterName string, clients map[string]ctrlruntimeclient.Client) error {
+	finalizerSet := sets.NewString(stream.ObjectMeta.Finalizers...)
```
alvaroaleman, line 263: Nit: There is no need to do the extra hoop via ObjectMeta, just write `stream.Finalizers` (multiple places)

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -247,6 +259,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
+					Name:      stream.Name,
+					Namespace: stream.Namespace,
+				},
+			}); !apierrors.IsNotFound(err) {
+				return err
+			}
+		}
+
+		// remove our finalizer from the list and update it.
+		stream.ObjectMeta.Finalizers = finalizerSet.Difference(sets.NewString(ControllerName)).List()
```
alvaroaleman, line 281: This needs to be `finalizerSet.Delete(ControllerName).List()`

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -159,6 +159,18 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		return fmt.Errorf("failed to get imageStream %s from cluster %s: %w", isName.String(), srcClusterName, err)
 	}
 
+	if sourceImageStream.ObjectMeta.DeletionTimestamp.IsZero() {
+		if err := ensureFinalizer(ctx, sourceImageStream, r.registryClients[srcClusterName]); err != nil {
+			return fmt.Errorf("failed to ensure finalizer to %s from cluster %s: %w", isName.String(), srcClusterName, err)
+		}
+	} else {
+		if err := finalizeIfNeeded(ctx, sourceImageStream, srcClusterName, r.registryClients); err != nil {
```
alvaroaleman, line 167: This code can be more left-aligned and avoid the else by doing:

```
if sourceImageStream.DeletionTimestamp != nil {
  if err := finalizeIfNeeded(); err != nil {
    return fmt.Errorf("cleanup failed") 
  }
  return nil
}

if err := ensureFinalizer(); err != nil {
  return fmt.Errorf("failed to ensure finalizer")
}


/// Rest  of reconciler
```

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -247,6 +259,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
+			if clusterName == srcClusterName {
+				continue
+			}
+			// populate deleting to all clusters except srcClusterName
+			if err := client.Delete(ctx, &imagev1.ImageStream{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      stream.Name,
+					Namespace: stream.Namespace,
+				},
+			}); !apierrors.IsNotFound(err) {
```
alvaroaleman, line 275: This is wrong, because if the error is nil we will return here without removing the finalizer, needs to be `err != nil && !apierrors.IsNotFound(err)`

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -305,6 +321,30 @@ func TestReconcile(t *testing.T) {
 					return fmt.Errorf("actual does not match expected, diff: %s", diff)
 				}
 
+				actualImageStreamFin := &imagev1.ImageStream{}
+				if err := apiCIClient.Get(ctx, types.NamespacedName{Name: "applyconfig", Namespace: "ci"}, actualImageStreamFin); err != nil {
+					return err
+				}
+				expectedImageStreamFin := &imagev1.ImageStream{
```
alvaroaleman, line 328: Please avoid abbreviations like `Fin` and spell out what you mean (`WithFinalizer`?), you will have forgotten what they mean in two weeks

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -434,6 +474,53 @@ func TestReconcile(t *testing.T) {
...
+			request: types.NamespacedName{
+				Name:      "applyconfig:latest",
+				Namespace: "ci",
+			},
+			apiCIClient: fakeclient.NewFakeClient(applyconfigISTagNewer.DeepCopy(), applyconfigISDeleted.DeepCopy()),
+			appCIClient: fakeclient.NewFakeClient(applyconfigISTag.DeepCopy(), applyconfigIS.DeepCopy()),
+
+			verify: func(apiCIClient ctrlruntimeclient.Client, appCIClient ctrlruntimeclient.Client) error {
+				actualImageStreamImport := &imagev1.ImageStreamImport{}
+				if err := appCIClient.Get(ctx, types.NamespacedName{Name: "applyconfig", Namespace: "ci"}, actualImageStreamImport); !errors.IsNotFound(err) {
```
alvaroaleman, line 488: please always alias the kubernetes errors package as `apierrors` or `kerrors`, the `errors` name is reserved for the stdlib package and code becomes confusing to read if the same import name refers to different packages in different files

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -434,6 +474,53 @@ func TestReconcile(t *testing.T) {
...
+					},
+				}
+				//ignoring DeletionTimestamp
+				expectedImageStreamFin.DeletionTimestamp = actualImageStreamFin.DeletionTimestamp
+				if diff := cmp.Diff(expectedImageStreamFin, actualImageStreamFin); diff != "" {
+					return fmt.Errorf("actual does not match expected, diff: %s", diff)
+				}
+				actualImageStream := &imagev1.ImageStream{}
+				if err := appCIClient.Get(ctx, types.NamespacedName{Name: "applyconfig", Namespace: "ci"}, actualImageStream); !errors.IsNotFound(err) {
+					t.Errorf("expected NotFound eoror did not occur")
```
alvaroaleman, line 519: s/eoror/error, also do yourself a favor and include the actual error in the `t.Errorf` :)

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -434,6 +474,53 @@ func TestReconcile(t *testing.T) {
...
+						APIVersion: "image.openshift.io/v1",
+					},
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: "ci",
+						Name:      "applyconfig",
+						Annotations: map[string]string{
+							"release.openshift.io-something": "copied",
+							"something":                      "not-copied",
+						},
+						Finalizers:        []string{"registry_syncer"},
```
alvaroaleman, line 508: I am confused by this, shouldn't the reconciliation remove the finalizer?

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -434,6 +474,53 @@ func TestReconcile(t *testing.T) {
...
+						DeletionTimestamp: &now,
+					},
+				}
+				//ignoring DeletionTimestamp
+				expectedImageStreamFin.DeletionTimestamp = actualImageStreamFin.DeletionTimestamp
```
alvaroaleman, line 513: Why ignore it? It should be the same as before or not?

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -745,3 +832,38 @@ func TestDockerImageImportedFromTargetingCluster(t *testing.T) {
 		})
 	}
 }
+
+func TestWeird(t *testing.T) {
```
alvaroaleman, line 836: ?

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -247,6 +259,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
+		stream.ObjectMeta.Finalizers = finalizerSet.Difference(sets.NewString(ControllerName)).List()
+		if err := clients[srcClusterName].Update(ctx, stream); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func ensureFinalizer(ctx context.Context, stream *imagev1.ImageStream, client ctrlruntimeclient.Client) error {
+	if !sets.NewString(stream.ObjectMeta.Finalizers...).Has(ControllerName) {
```
alvaroaleman, line 290: stylistic nit: Prefer left-aligned code, so here that would be:
```
if sets.NewString(stream.Finalizers).Has(ControllerName) {
  return nil
}
stream.Finalizers = append(stream.Finalizers, ControllerName)
return client.Update(ctx, stream)
```
That both saves you some lines and makes the code quicker to read.

See https://blog.codinghorror.com/flattening-arrow-code/ for some more examples of this

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: I did not know what I was thinking.
I would have caught `err != nil && !apierrors.IsNotFound(err)` otherwise.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: GoLand did not show me `Delete` in the auto-completion.
It was too late for my laptop too.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: Updated.

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: This would become a dead loop again.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Removed.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Updated.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Updated.

pkg/controller/registrysyncer/registrysyncer.go

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Because the time is slightly different from what we defined.
Added a nil check to ensure non-nil value.

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: This is to show that fakeClient does not return the expected timestamp.

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -247,6 +259,41 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
 
+func finalizeIfNeeded(ctx context.Context, stream *imagev1.ImageStream, srcClusterName string, clients map[string]ctrlruntimeclient.Client) error {
+	finalizerSet := sets.NewString(stream.Finalizers...)
+	if finalizerSet.Has(ControllerName) {
+		for clusterName, client := range clients {
+			if clusterName == srcClusterName {
+				continue
+			}
+			// populate deleting to all clusters except srcClusterName
+			if err := client.Delete(ctx, &imagev1.ImageStream{
```
alvaroaleman, line 270: This will need to delete the finalizer if present which can happen if we do some back and forth syncing

pkg/controller/registrysyncer/registrysyncer_test.go
alvaroaleman: okay but we should probably not merge that :)

alvaroaleman (approved):

pkg/controller/registrysyncer/registrysyncer.go
I am curious of what would happen otherwise.

`finalize` on A -> `delete` B -> `finalize` on B -> `delete` A again (if A is still on the system)?

pkg/controller/registrysyncer/registrysyncer_test.go
hongkailiu: Removed.

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -247,6 +259,53 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
+				if apierrors.IsNotFound(err) {
+					continue
+				}
+				return err
+			}
+
+			finalizerSetOfISToDelete := sets.NewString(isToDelete.Finalizers...)
+			if finalizerSetOfISToDelete.Has(ControllerName) {
+				isToDelete.Finalizers = finalizerSetOfISToDelete.Delete(ControllerName).List()
+				if err := client.Update(ctx, isToDelete); err != nil {
```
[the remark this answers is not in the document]

alvaroaleman (approved):

PR #1458, opened 2020-11-26
Multi-stage tests: Track subtests in graph

alvaroaleman:
Exposes the steps in multi-stage tests as distinct test in the graph

### DIFF
--- a/pkg/steps/multi_stage_test.go
+++ b/pkg/steps/multi_stage_test.go
@@ -26,6 +26,9 @@ import (
 	"github.com/openshift/ci-tools/pkg/testhelper"
 )
 
+// the multiStageTestStep implements the subStepReporter interface
+var _ SubStepReporter = &multiStageTestStep{}
+
 func TestRequires(t *testing.T) {
 	for _, tc := range []struct {
 		name   string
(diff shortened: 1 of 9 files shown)


pkg/steps/multi_stage_test.go
```diff
@@ -26,6 +26,9 @@ import (
 	"github.com/openshift/ci-tools/pkg/testhelper"
 )
 
+// the multiStageTestStep implements the subStepReporter interface
+var _ SubStepReporter = &multiStageTestStep{}
+
```
petr-muller, line 31: Huh why do we need an unused var here?

petr-muller (commented):
LGTM, needs rebase

pkg/steps/multi_stage_test.go
alvaroaleman: This verifies that the *multistageTestStep implements the `SubStepReporter`, otherwise the only indication that we don't do that anymore would be the data not appearing

alvaroaleman:

stevekuznetsov:
Merging 1464. 

/close

PR #1464, opened 2020-11-27
Allow unmarshaling ci-operator step graph

alvaroaleman:
Includes https://github.com/openshift/ci-tools/pull/1458 to avoid needing another rebase

The json unmarshaler refuses to unmarshal into a non-empty interface as it can't know the underlying type. So this adds a custom unmarshaler, that unmarshals the manifests into `*unstructured.Unstructured`. This in turn requires TypeMeta which is not set by default, so the second commit adds that.

### DIFF
--- a/pkg/api/graph.go
+++ b/pkg/api/graph.go
@@ -2,12 +2,14 @@ package api
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
 
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/util/sets"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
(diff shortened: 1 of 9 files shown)


pkg/api/graph.go
```diff
@@ -470,22 +472,65 @@ func mergeSteps(into, from CIOperatorStepWithDependencies) CIOperatorStepWithDep
...
 	StartedAt    *time.Time                 `json:"started_at"`
 	FinishedAt   *time.Time                 `json:"finished_at"`
 	Duration     *time.Duration             `json:"duration,omitempty"`
-	Manifests    []ctrlruntimeclient.Object `json:"manifests"`
+	Manifests    []ctrlruntimeclient.Object `json:"manifests,omitempty"`
 	LogURL       string                     `json:"log_url,omitempty"`
 	Failed       *bool                      `json:"failed,omitempty"`
 }
 
+func (c *CIOperatorStepDetailInfo) UnmarshalJSON(data []byte) error {
```
alvaroaleman, line 499: If someone can come up with a less inefficient way to do this, that would be great

alvaroaleman:

stevekuznetsov (approved):
Confusing and inefficient but we do it once per execution so I think we're OK...

PR #1466, opened 2020-11-27
pkg/api: check `values` in `Params::has`

bbguimaraes:
I noticed while working on 5b88367b55758f6b7310bc4bcd46a4033019522d that this
discrepancy existed, but did not have time to correct it.  I do not see a reason
for it to exist.

### DIFF
--- a/pkg/api/parameters.go
+++ b/pkg/api/parameters.go
@@ -133,7 +133,11 @@ func (p *DeferredParameters) Has(name string) bool {
 func (p *DeferredParameters) has(name string) bool {
 	p.lock.Lock()
 	defer p.lock.Unlock()
-	_, ok := p.fns[name]
+	_, ok := p.values[name]
+	if ok {
+		return true
+	}
+	_, ok = p.fns[name]
 	return ok
 }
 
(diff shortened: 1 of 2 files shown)


bbguimaraes:

stevekuznetsov (approved):

bbguimaraes:
There seems to be some obscure bug here.

/hold

bbguimaraes:
With this change, E2E template tests fail to pull the `cli` image (referenced in the template as `${IMAGE_CLI}`).  Prior to this change, the resulting pod uses this pull spec:

```
registry.build01.ci.openshift.org/bbguimaraes/stable:cli
```

When `DeferredParameters::Has` is corrected (?) to check the cached values in the `values` map, it is changed to:

```
registry.build01.ci.openshift.org/ocp/4.5@sha256:d08df67d3b0031f29baf6f41ac896b383fe70b04141de48c4e4646d6066f8d1a
```

This is caused by this section:

https://github.com/openshift/ci-tools/blob/7cb311dcb781a92b5ab12e9cb382e5c8b57246e6/pkg/steps/template.go#L77-L94

which checks `params.Has` and uses the paramter value without change if it exists.  The following section (which is skipped if the previous check succeeds) sets the value of stable images according to `$IMAGE_FORMAT`.

`Has` is a misleading name for the method.  Because it only checks `fns` (which apparently is a requirement), it ultimately determines if there exists a step in the graph that provides the specified parameter.  This PR is, consequently, incorrect.

This API is insane and/or works by accident.  We could rename/refactor the code, but I am more inclined to leave this brain dump here and revisit it after we deprecate and remove templates.  Multi-stage tests do not have this magical treatment.

/close

PR #1472, opened 2020-11-30
isTag handler: reconcile when DeletionTimestamp is defined

hongkailiu:
Changes on `DeletionTimestamp` generate UpdateEvent.
The DeleteEvent occurs only when the the object is deleted.

When we handles the annotations of isTag, I believe we need to modify this function too.

### DIFF
--- a/pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
+++ b/pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
@@ -42,8 +42,9 @@ func (m *imagestreamtagmapper) Update(e event.UpdateEvent, q workqueue.RateLimit
 		return
 	}
 
+	isDeleted := newStream.DeletionTimestamp != nil
 	for _, newTag := range newStream.Status.Tags {
-		if namedTagEventListHasElement(oldStream.Status.Tags, newTag) {
+		if !isDeleted && namedTagEventListHasElement(oldStream.Status.Tags, newTag) {
 			continue
 		}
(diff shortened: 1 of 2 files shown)


pkg/util/imagestreamtagmapper/imagestreamtagmapper.go
```diff
@@ -43,7 +43,9 @@ func (m *imagestreamtagmapper) Update(e event.UpdateEvent, q workqueue.RateLimit
 	}
 
 	for _, newTag := range newStream.Status.Tags {
-		if namedTagEventListHasElement(oldStream.Status.Tags, newTag) {
+		if namedTagEventListHasElement(oldStream.Status.Tags, newTag) &&
+			//generate requests for the isTags when DeletionTimestamp is defined
+			!(oldStream.DeletionTimestamp == nil && newStream.DeletionTimestamp != nil) {
```
alvaroaleman, line 48: Can you change this to:
```
isDeleted := oldStream.DeletionTimestamp == nil && newStream.DeletionTimestamp != nil
for _, newTag := range newStream.Status.Tags {
  if !isDeleted && namedTagEventListHasElement(oldStream.Status.Tags, newTag) {
    continue
  }
}
```

The current multiline if header is kinda annoying to read

alvaroaleman (approved):
/retitle isTag handler: reconcile when DeletionTimestamp is defined

alvaroaleman (approved):

alvaroaleman (approved):

PR #1474, opened 2020-12-01
Revert "ipi-deprovisioner: tighen timeout to 10m per cluster"

wking:
Reverts openshift/ci-tools#1463

@stevekuznetsov got timestamped logs from [this run][1], and they included:

```
2020-12-01T17:42:50.801274924Z level=debug msg=OpenShift Installer unreleased-master-3986-g36cf196e4bb097967ed6d0355f448eaf3c9763af-dirty
2020-12-01T17:42:50.801300795Z level=debug msg=Built from commit 36cf196e4bb097967ed6d0355f448eaf3c9763af
...
2020-12-01T17:42:51.549404286Z level=debug msg=search for IAM roles
2020-12-01T17:44:58.393568493Z level=debug msg=search for IAM users
2020-12-01T17:51:39.370784031Z level=debug msg=search for IAM instance profiles
...
2020-12-01T17:52:50.71621937Z SIGQUIT: quit
```

So the slow IAM user listing is only leaving us ~70s to perform the actual deletion.  We should raise the cap back up to give things more time.  And also probably prune whatever stale users are slowing things down.

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ipi-deprovision/1333798399111073792

### DIFF
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -54,7 +54,7 @@ done
 
 failed=""
 for workdir in $( find /tmp/deprovision -mindepth 1 -type d | shuf ); do
-  if ! timeout --signal=SIGQUIT 10m openshift-install --dir "${workdir}" --log-level debug destroy cluster; then
+  if ! timeout --signal=SIGQUIT 30m openshift-install --dir "${workdir}" --log-level debug destroy cluster; then
     failed+=",$( basename "${workdir}" )"
   fi
 done


petr-muller:
I observed something entirely different, which I described in https://github.com/openshift/ci-tools/pull/1463:

> I checked the logs for the duration of successful cleanups, and in 99% cases, we either manage to cleanup under 10m, or we hit the 30m timeout. From the ~370 successful cleanups, just 7 were over 10 minutes, which I think is acceptable for now when we need to get that job under the timeout so that we can continue to analyze which clusters are stuck long-term.

petr-muller (approved):

petr-muller:
But yeah, it should not hurt to revert back, provided we can address the underlying issue with SGs

PR #1476, opened 2020-12-02
registrySyncer: delete IS when it is deleted on any cluster

hongkailiu:
- cover the case: the older IS is deleted. This is actually what happens in our production. IS is synced from api.ci to app.ci and then the one on api.ci is deleted. When reconcile, the one on app.ci becomes the sourceIS because it is newer. But we cannot enter the deleting logic because the one on app.ci is not with deleteTimestamp. The fix here is: if IS on any cluster has a deleteTimestamp, we remove the finalizer (if existing) for IS on all clusters and delete the ones with deleteTimestamp.

- adding logs for deleting IS and removing finalizer

/cc @alvaroaleman

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -158,16 +158,17 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 		return fmt.Errorf("failed to get imageStream %s from cluster %s: %w", isName.String(), srcClusterName, err)
 	}
 
-	if sourceImageStream.DeletionTimestamp != nil {
-		if err := finalizeIfNeeded(ctx, sourceImageStream, srcClusterName, r.registryClients); err != nil {
-			return fmt.Errorf("failed to finalize %s from cluster %s: %w", isName.String(), srcClusterName, err)
-		}
-		// no sync if the srcIS is deleted
-		return nil
-	}
-
 	if err := ensureFinalizer(ctx, sourceImageStream, r.registryClients[srcClusterName]); err != nil {
 		return fmt.Errorf("failed to ensure finalizer to %s from cluster %s: %w", isName.String(), srcClusterName, err)
+	}
+
+	deleted, err := finalizeIfNeeded(ctx, sourceImageStream, r.registryClients, log)
+	if err != nil {
+		return fmt.Errorf("failed to finalize %s from cluster %s: %w", isName.String(), srcClusterName, err)
+	}
+	if deleted {
+		// no sync if the srcIS is deleted
+		return nil
 	}
 
 	*log = *log.WithField("docker_image_reference", sourceImageStreamTag.Image.DockerImageReference)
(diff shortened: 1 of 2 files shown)


pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -258,44 +259,70 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 	return nil
 }
 
-func finalizeIfNeeded(ctx context.Context, stream *imagev1.ImageStream, srcClusterName string, clients map[string]ctrlruntimeclient.Client) error {
-	finalizerSet := sets.NewString(stream.Finalizers...)
-	if !finalizerSet.Has(finalizerName) {
-		return nil
+func finalizeIfNeeded(ctx context.Context, sourceImageStream *imagev1.ImageStream, srcClusterName string, clients map[string]ctrlruntimeclient.Client, log *logrus.Entry) (bool, error) {
+	deleteTimeStampClusters := sets.NewString()
```
alvaroaleman, line 263: This seems overly complicated. Please just:
* Get the other imagestream
* Return if both have a nil DeletionTimestamp, without putting them in a different data structure first
* Iterate over them and
    * Remove finalizer if present
     * Delete

pkg/controller/registrysyncer/registrysyncer_test.go
```diff
@@ -525,8 +537,133 @@ func TestReconcile(t *testing.T) {
...
+			appCIClient: fakeclient.NewFakeClient(applyconfigISTagNewer.DeepCopy(), applyconfigIS.DeepCopy()),
+
+			verify: func(apiCIClient ctrlruntimeclient.Client, appCIClient ctrlruntimeclient.Client) error {
+				actualImageStreamImport := &imagev1.ImageStreamImport{}
+				if err := apiCIClient.Get(ctx, types.NamespacedName{Name: "applyconfig", Namespace: "ci"}, actualImageStreamImport); !apierrors.IsNotFound(err) {
+					t.Errorf("expected NotFound error did not occur and the actual error is: %v", err)
+				}
+
+				actualImageStream := &imagev1.ImageStream{}
+				if err := apiCIClient.Get(ctx, types.NamespacedName{Name: "applyconfig", Namespace: "ci"}, actualImageStream); err != nil {
```
alvaroaleman, line 556: I don't get it. If we delete it, why does it still exist?

hongkailiu: It is actually not deleted. It has a deleteTimestamp.
We need to remove our finalizer and kubelet will delete it when there is no finalizer.

alvaroaleman: Ah, it already has the deletion timestamp set when we create it and we only verify that the finalizer is abset. Could make that more explicit, as we don't care about the rest

pkg/controller/registrysyncer/registrysyncer.go
hongkailiu: This roughly does that. I have tried to simplify this but failed.
I guess this makes the function complicated:

1. handle differently with the one with DeleteTimestamp and without. For the former, we just remove the finalizer. For the latter, we remove the finalizer and delete. Maybe doing delete too for the former will make the function easy but probably we will get NotFound. Our concern yesterday was it was not fast enough. Maybe do only what we have to do. WDYT?
2. We have not assumed that we have only 2 clusters to sync. I agree that maybe we never need sync more than 2 clusters. But I do not want to make this function special if that is OK.

alvaroaleman: > handle differently with the one with DeleteTimestamp and without. For the former, we just remove the finalizer. For the latter, we remove the finalizer and delete. Maybe doing delete too for the former will make the function easy but probably we will get NotFound. Our concern yesterday was it was not fast enough. Maybe do only what we have to do. WDYT?

That still doesn't require all the additional data structures:
```
stream := &imagev1.ImageStream{}
if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: sourceImageStream.Name, Namespace: sourceImageStream.Namespace}, stream); err != nil {
  if apierrors.IsNotFound{
     stream = nil
  }
  return false, err
}

if stream != nil && stream.DeletionTimestamp == nil && sourceImageStream.DeletionTimeStamp == nil {
  return false, nil
}
// finalize if needed stream
// finalize if needed sourceStream
// Delete if deletionTimestamp is unset for Stream, tolerate IsNotFound
// Delete if deletinTimestamp is unset for SourceStream, tolreate IsNotFound
```

> We have not assumed that we have only 2 clusters to sync. I agree that maybe we never need sync more than 2 clusters. But I do not want to make this function special if that is OK.

That is premature optimization, do not solve problems you don't have, especially not if they complicate your solution.

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -258,44 +259,54 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
 
-func finalizeIfNeeded(ctx context.Context, stream *imagev1.ImageStream, srcClusterName string, clients map[string]ctrlruntimeclient.Client) error {
-	finalizerSet := sets.NewString(stream.Finalizers...)
-	if !finalizerSet.Has(finalizerName) {
-		return nil
-	}
-	for clusterName, client := range clients {
-		if clusterName == srcClusterName {
-			continue
+func finalizeIfNeeded(ctx context.Context, sourceImageStream *imagev1.ImageStream, clients map[string]ctrlruntimeclient.Client, log *logrus.Entry) (bool, error) {
```
alvaroaleman, line 262: much, much nicer, thanks for updating :)

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -258,44 +259,54 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
+		if err := ensureRemoveFinalizer(ctx, stream, client, log); err != nil {
+			return false, err
 		}
-
-		// populate deleting to all clusters except srcClusterName
-		if err := client.Delete(ctx, isToDelete); err != nil && !apierrors.IsNotFound(err) {
-			return err
+		if stream.DeletionTimestamp == nil {
+			log.Debug("deleting imagestream")
+			// populate deleting to all clusters except srcClusterName
```
alvaroaleman: That comment isn't correct anymore I think?

alvaroaleman (approved):

pkg/controller/registrysyncer/registrysyncer.go

PR #1477, opened 2020-12-02
[DPTP-1731] add the ability to create rbacs and wait until the image pull secret has been created

droslean:
When a service account is being created, the [DockercfgController](https://github.com/openshift/openshift-controller-manager/blob/master/pkg/serviceaccounts/controllers/create_dockercfg_secrets.go#L315) creates the pull secret and then update the service account. Therefore if the service account will be used before the creation of that secret, the running pod will pull images as anonymous instead. 

This PR fixes this race condition by creating a new function that will be responsible for the RBAC creation and waits until the dockercfg controller will create the secret and sync the service account.

Jobs with the related error: [ci-search - Container .* is not ready with reason ImagePullBackOff](https://search.ci.openshift.org/?search=Container+.*+is+not+ready+with+reason+ImagePullBackOff&maxAge=24h&context=0&type=build-log&name=.*&maxMatches=1&maxBytes=&groupBy=job)

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/junit"
 	"github.com/openshift/ci-tools/pkg/results"
 	"github.com/openshift/ci-tools/pkg/steps/utils"
+	"github.com/openshift/ci-tools/pkg/util"
 )
 
 const (
@@ -136,7 +137,7 @@ func (s *multiStageTestStep) run(ctx context.Context) error {
 	if err := s.createCredentials(); err != nil {
 		return fmt.Errorf("failed to create credentials: %w", err)
 	}
-	if err := s.setupRBAC(); err != nil {
+	if err := s.setupRBAC(ctx); err != nil {
 		return fmt.Errorf("failed to create RBAC objects: %w", err)
 	}
 	var errs []error
@@ -213,7 +214,7 @@ func (s *multiStageTestStep) Provides() api.ParameterMap {
 }
 func (s *multiStageTestStep) SubTests() []*junit.TestCase { return s.subTests }
 
-func (s *multiStageTestStep) setupRBAC() error {
+func (s *multiStageTestStep) setupRBAC(ctx context.Context) error {
 	labels := map[string]string{MultiStageTestLabel: s.name}
 	m := meta.ObjectMeta{Namespace: s.jobSpec.Namespace(), Name: s.name, Labels: labels}
 	sa := &coreapi.ServiceAccount{ObjectMeta: m}
@@ -240,16 +241,7 @@ func (s *multiStageTestStep) setupRBAC() error {
 		RoleRef:    rbacapi.RoleRef{Kind: "Role", Name: s.name},
 		Subjects:   subj,
 	}
-	check := func(err error) bool {
-		return err == nil || kerrors.IsAlreadyExists(err)
-	}
-	if err := s.client.Create(context.TODO(), sa); !check(err) {
-		return err
-	}
-	if err := s.client.Create(context.TODO(), role); !check(err) {
-		return err
-	}
-	if err := s.client.Create(context.TODO(), binding); !check(err) {
+	if err := util.CreateRBACs(ctx, sa, role, binding, s.client, 1*time.Minute, 5*time.Minute); err != nil {
 		return err
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


alvaroaleman:
> This PR fixes this race condition by creating a new function that will be responsible for the RBAC creation and waits until the dockercfg controller will create the secret and sync the service accoun

Why not simply poll until the Secret is created? The approach taken here seems overly complicated.

droslean:
> Why not simply poll until the Secret is created? The approach taken here seems overly complicated.

Well, we don't know the secret name to start with.  Also, this approach isn't complicated. We just wait until our service account has a value in its `imagePullSecrets`

pkg/util/rbacs.go
```diff
@@ -0,0 +1,57 @@
...
+		return results.ForReason("creating_service_account").WithError(err).Errorf("could not create service account '%s'", sa.Name)
+	}
+	if err := client.Create(ctx, role); err != nil && !kerrors.IsAlreadyExists(err) {
+		return results.ForReason("creating_roles").WithError(err).Errorf("could not create role '%s'", role.Name)
+	}
+	if err := client.Create(ctx, roleBinding); err != nil && !kerrors.IsAlreadyExists(err) {
+		return results.ForReason("binding_roles").WithError(err).Errorf("could not create role binding '%s'", roleBinding.Name)
+	}
+
+	if err := retry.OnError(
```
alvaroaleman, line 29: This is a lot simpler with `k8s.io/apimachinery/pkg/util/wait.Poll`

alvaroaleman:
> Well, we don't know the secret name to start with. Also, this approach isn't complicated. We just wait until our service account has a value in its imagePullSecrets

Ah sorry, I was confused by the PRs title and didn't understand why we create RBAC now, but its just code that got moved around

pkg/steps/multi_stage.go
```diff
@@ -240,18 +241,11 @@ func (s *multiStageTestStep) setupRBAC() error {
...
-	}
-	if err := s.client.Create(context.TODO(), sa); !check(err) {
-		return err
-	}
-	if err := s.client.Create(context.TODO(), role); !check(err) {
-		return err
-	}
-	if err := s.client.Create(context.TODO(), binding); !check(err) {
+
+	if err := util.CreateRBACs(context.TODO(), sa, role, binding, s.client, 1*time.Minute); err != nil {
```
alvaroaleman, line 245: Why did you move that to a different package when its only used here? And its kind of confusing to have a `CreateRBAC` func that requires a ServiceAccount as argument. Also, doesn't the ServiceAccount poll code belong into `InitializeNamespace`?

pkg/util/rbacs.go
droslean: Thanks. Let me check that.

pkg/steps/multi_stage.go
alvaroaleman: Ah, its also used in CreateRelease. Still, I think the part that needs factoring out is the SA creation + Polling, combining it with the RBAC just makes it less re-usable and more confusing

droslean: There are currently two places that we are creating rbacs. Multi-stage and create_release. I moved it to `util` package because it makes sense to have this kind of functionality in a neutral place. In `initializeNamespace` for some reason, we don't create any related rbacs.

droslean: yea. Also, they are creating different serviceaccounts. The whole concept here needs some refactoring for sure.

pkg/util/rbacs.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	if err := client.Create(ctx, role); err != nil && !kerrors.IsAlreadyExists(err) {
+		return results.ForReason("creating_roles").WithError(err).Errorf("could not create role '%s'", role.Name)
+	}
+	if err := client.Create(ctx, roleBinding); err != nil && !kerrors.IsAlreadyExists(err) {
+		return results.ForReason("binding_roles").WithError(err).Errorf("could not create role binding '%s'", roleBinding.Name)
+	}
+
+	if !skipPoll {
```
alvaroaleman, line 39:
```suggestion
	if skipPoll {
	   return nil
	}
```

See https://blog.codinghorror.com/flattening-arrow-code/ for some samples on how to avoid this kind of arrow code

droslean: addressed

stevekuznetsov (approved):

PR #1484, opened 2020-12-03
Increase timeout when trying to create rbacs.

droslean:

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -241,7 +241,7 @@ func (s *multiStageTestStep) setupRBAC(ctx context.Context) error {
 		RoleRef:    rbacapi.RoleRef{Kind: "Role", Name: s.name},
 		Subjects:   subj,
 	}
-	if err := util.CreateRBACs(ctx, sa, role, binding, s.client, 1*time.Minute, 5*time.Minute); err != nil {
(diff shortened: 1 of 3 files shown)


alvaroaleman:
@droslean this takes more than five mintes? Can you create a BZ? That is definitely not supposed to take this long.

droslean:
@alvaroaleman The main problem is that the dockercfgcontroller doesn't leave any annotation with information of when it actually synced the service account. I will open a BZ to see if there is any workaround to get this info or implement this. In the meanwhile, I would like to increase the timeout and monitor the jobs to see if we still hit the error.

alvaroaleman:
> The main problem is that the dockercfgcontroller doesn't leave any annotation with information of when it actually synced the service account

Why would we need an annotation? Our success check is `if len(actualSA.ImagePullSecrets) > 0`. A controller that needs more than five minutes for syncing some data in the same cluster is absolutely not expected and if it indeed takes so long, that is a bug in that controller (probably too little workers).

droslean:
@alvaroaleman We don't need annotation here. I need the annotation to figure out if the problem if there is a problem with the controller at all. Now we are blind.

alvaroaleman:
@droslean can you please just create a BZ instead and link to that BZ from the code, together with sample jobs where the 5 min timeout wasn't enough? If this happens its a product bug and the responsible team needs to debug and fix it.

PR #1487, opened 2020-12-03
cmd/ipi-deprovision: Parallel cluster teardown

wking:
Cluster teardown is not particularly CPU-intensive, and there is a lot of waiting around for cloud-provider queries to come back or cloud-provider state to become eventually consistent.  Tearing down clusters in parallel will allow us to grind through a backlog of leaked clusters more quickly, and will also give us some ability to continue reaping in the face of small numbers of stuck-in-teardown clusters (which currently gum up the works for their full 30m quota before being SIGQUITed).

The `queue` function is lifted from `pkg/steps/clusterinstall/template.go` with minor changes (no need to redirect stdout, since the installer is mostly writing to stderr).  I've dropped the parallel-job to 10, which will still be 10x faster than the outgoing, serial logic, and seems like plenty until we have more experience with this approach.

I'm moving the working directory under `ARTIFACT_DIR` so we have the installer's `.openshift_install.log` (with timestamps!) streaming into a capture location, even if the overall container times out and is killed.  That means I can tune down --log-level (which only affects stderr) to error, because we don't need a lot of noise there, especially now that all the standard streams are getting interleaved.

With the backgrounded teardown, counting failed deletions is a bit more awkward.  We could track in-flight PIDs, and `wait "${PID}"` each of them to collect exit codes, but I'm taking the easy way out and just touching `success`/`failure` files in the artifact directory from the backgrounded process.

### DIFF
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -4,13 +4,30 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+trap 'CHILDREN=$(jobs -p); if test -n "${CHILDREN}"; then kill ${CHILDREN} && wait; fi' TERM
+
+function queue() {
+  local LIVE="$(jobs | wc -l)"
+  while [[ "${LIVE}" -ge 10 ]]; do
+    sleep 1
+    LIVE="$(jobs | wc -l)"
+  done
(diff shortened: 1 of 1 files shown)


stevekuznetsov (approved):

wking:
I flubbed some env. vars; fixups landed in #1489 and #1490.

PR #1491, opened 2020-12-04
Make CI Registry configurable

hongkailiu:
https://issues.redhat.com/browse/DPTP-1657

The target of this PR is that promotion and DPTP-CM are ready for CI registry migration from api.ci to app.ci:
Modify the two TODOs in this PR.

### DIFF
--- a/cmd/dptp-controller-manager/main.go
+++ b/cmd/dptp-controller-manager/main.go
@@ -61,6 +61,7 @@ type options struct {
 	leaderElectionSuffix         string
 	enabledControllers           flagutil.Strings
 	enabledControllersSet        sets.String
+	registryClusterName          string
 	dryRun                       bool
 	blockProfileRate             time.Duration
 	testImagesDistributorOptions testImagesDistributorOptions
@@ -139,6 +140,7 @@ func newOpts() (*options, error) {
 	flag.StringVar(&opts.secretSyncerConfigOptions.configFile, "secretSyncerConfigOptions.config", "", "The config file for the secret syncer controller")
 	flag.StringVar(&opts.secretSyncerConfigOptions.secretBoostrapConfigFile, "secretSyncerConfigOptions.secretBoostrapConfigFile", "", "The config file for ci-secret-boostrap")
 	flag.DurationVar(&opts.blockProfileRate, "block-profile-rate", time.Duration(0), "The block profile rate. Set to non-zero to enable.")
+	flag.StringVar(&opts.registryClusterName, "registry-cluster-name", "api.ci", "the cluster name on which the CI central registry is running")
 	flag.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
 	flag.Parse()
 
@@ -273,8 +275,8 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatal("failed to load kubeconfigs")
 	}
-	if _, hasApiCI := kubeconfigs[apiCIContextName]; !hasApiCI {
-		logrus.Fatalf("--kubeconfig must include a context named `%s`", apiCIContextName)
+	if _, hasRegistryCluster := kubeconfigs[opts.registryClusterName]; !hasRegistryCluster {
+		logrus.Fatalf("--kubeconfig must include a context named `%s`", opts.registryClusterName)
 	}
 	if _, hasAppCi := kubeconfigs[appCIContextName]; !hasAppCi {
 		kubeconfigs[appCIContextName], err = rest.InClusterConfig()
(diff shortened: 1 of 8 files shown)


stevekuznetsov (commented):
This doesn't make it configurable, it makes it compiled into the binary. I don't think we want to be juggling deployment rollout versions instead of merging a config flag change...

cmd/dptp-controller-manager/main.go
```diff
@@ -317,27 +326,29 @@ func main() {
...
-		DryRunClient:       opts.dryRun,
-		Logger:             ctrlruntimelog.NullLogger{},
-	})
-	if err != nil {
-		logrus.WithError(err).Fatal("Failed to construct manager for registry")
-	}
-	if err := mgr.Add(registryMgr); err != nil {
-		logrus.WithError(err).Fatal("Failed to add registry manager to main manager.")
+	registryMgr := mgr
+	if !registryMigrated {
```
alvaroaleman, line 330: Please don't use opaque and current-state-of-the-world based variables like `registryMigrated`. Iterate over all configs, construct a manager for each, if name == registryClusterName then you add the `SyncPeriod` to the options and set the `registryManager` variable

hongkailiu: Updated:
Had to loop twice:
* create the main manager for app.ci in the first loop
* add the others to the main manager in the 2nd loop

cmd/dptp-controller-manager/main.go
```diff
@@ -273,8 +275,8 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatal("failed to load kubeconfigs")
 	}
-	if _, hasApiCI := kubeconfigs[apiCIContextName]; !hasApiCI {
-		logrus.Fatalf("--kubeconfig must include a context named `%s`", apiCIContextName)
+	if _, hasApiCI := kubeconfigs[opts.registryClusterName]; !hasApiCI {
```
alvaroaleman, line 278: nit: `s/hasApiCI/hasRegistryCluster/g`

cmd/dptp-controller-manager/main.go
```diff
@@ -317,27 +347,18 @@ func main() {
...
-		// The normal manager already serves these metrics and we must disable it here to not
-		// get an error when attempting to create the second listener on the same address.
-		MetricsBindAddress: "0",
-		SyncPeriod:         &resyncInterval,
-		DryRunClient:       opts.dryRun,
-		Logger:             ctrlruntimelog.NullLogger{},
-	})
-	if err != nil {
-		logrus.WithError(err).Fatal("Failed to construct manager for registry")
+	errs = []error{}
```
alvaroaleman, line 350: Nit: `errs = nil`, no need to allocate an array if we might not need it

cmd/dptp-controller-manager/main.go
```diff
@@ -377,39 +398,12 @@ func main() {
...
-
-	allClustersExceptAPICI := map[string]controllerruntime.Manager{}
+	allClustersExceptRegistryCluster := map[string]controllerruntime.Manager{}
 	for cluster, manager := range allManagers {
-		if cluster == apiCIContextName {
+		if cluster == opts.registryClusterName {
 			continue
 		}
-		allClustersExceptAPICI[cluster] = manager
+		allClustersExceptRegistryCluster[cluster] = manager
```
alvaroaleman, line 406: You can do this in the first loop where you create the managers

cmd/dptp-controller-manager/main.go
```diff
@@ -474,6 +472,9 @@ func main() {
 		if opts.registrySyncerOptions.imagePullSecretPath == "" {
 			logrus.Fatal("The registrysyncer requires the --registrySyncerOptions.imagePullSecretPath flag to be set ")
 		}
+		if _, hasApiCI := kubeconfigs[apiCIContextName]; !hasApiCI {
```
alvaroaleman, line 475: why do we use `apiCIContextName` and not `opts.registryClusterName` here?

cmd/dptp-controller-manager/main.go
alvaroaleman: actually not even with that, if errors is not nil we will `Fatal` before we reach this statement so no need to do anything about the `errs` variable

cmd/dptp-controller-manager/main.go

cmd/dptp-controller-manager/main.go

cmd/dptp-controller-manager/main.go

cmd/dptp-controller-manager/main.go
hongkailiu: Because `opts.registryClusterName` could be a cluster other than api.ci in the future and we do registry sync for app.ci and api.ci for the moment which still holds after migrating the registry.

alvaroaleman (approved):

PR #1492, opened 2020-12-04
Bump test-infra

petr-muller:
Upstream removed `requiresig` so now checkconfig complains about its
presence in our config, but our determinizer (`make prow-config`) is
still running old code so it is putting the line back to config for
people when they remove it.


petr-muller:
`ci/prow/breaking-changes` job failure is expected

PR #1499, opened 2020-12-08
Add feature to automatically migrate openshift_installer jobs

petr-muller:
Moves forward [DPTP-1775](https://issues.redhat.com/browse/DPTP-1775)

Right now we have these two types in the repo:
1. upgrade tests (are migrated to the matching workflow)
```diff
 - as: e2e-aws-upgrade
-  commands: TEST_SUITE=all run-upgrade-tests
-  openshift_installer:
+  steps:
     cluster_profile: aws
-    upgrade: true
+    workflow: openshift-upgrade-aws-loki
```

2. disruptive tests (cannot be simply migrated right now, so they are
  kept)

I added code to migrate plain `openshift_installer` jobs too b/c it was
simple enough. Not handling that case would be suprising to an
unsuspecting reader of the code.

### DIFF
--- a/cmd/determinize-ci-operator/main.go
+++ b/cmd/determinize-ci-operator/main.go
@@ -20,9 +20,10 @@ const (
 	openshiftInstallerSRCTemplateName             = "openshift_installer_src"
 	openshiftInstallerCustomTestImageTemplateName = "openshift_installer_custom_test_image"
 	OpenshiftInstallerUPITemplateName             = "openshift_installer_upi"
+	OpenShiftInstallerTemplateName                = "openshift_installer"
 )
 
-var validTemplateMigrations = sets.NewString(openshiftInstallerSRCTemplateName, openshiftInstallerCustomTestImageTemplateName, OpenshiftInstallerUPITemplateName)
+var validTemplateMigrations = sets.NewString(openshiftInstallerSRCTemplateName, openshiftInstallerCustomTestImageTemplateName, OpenshiftInstallerUPITemplateName, OpenShiftInstallerTemplateName)
 
 type options struct {
 	config.ConfirmableOptions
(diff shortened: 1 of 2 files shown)


cmd/determinize-ci-operator/main.go
```diff
@@ -103,6 +110,56 @@ func main() {
...
+		clusterProfile := test.OpenshiftInstallerClusterTestConfiguration.ClusterProfile
+		switch {
+		case test.OpenshiftInstallerClusterTestConfiguration.Upgrade:
+			test.OpenshiftInstallerClusterTestConfiguration = nil
+			test.MultiStageTestConfiguration = &api.MultiStageTestConfiguration{
+				ClusterProfile: clusterProfile,
+				Workflow:       utilpointer.StringPtr(upgradeWorkflowForClusterProfile(clusterProfile)),
+			}
+		case test.Commands == "setup_ssh_bastion; TEST_SUITE=openshift/disruptive run-tests; TEST_SUITE=openshift/conformance/parallel run-tests":
+			// TODO(muller): Unfortunately there is no easy way to express this ("run same step twice")
```
stevekuznetsov, line 146: Technically what's missing is passing separate parameters to different invocations of a single ref, right? You can run the same step as many times as you need. Any way to just get `openshift-tests` to understand two test suites via one invocation?

stevekuznetsov (approved):

cmd/determinize-ci-operator/main.go
petr-muller: Yes, passing separate parameters to multiple invocations of a single ref is exactly what we'd need. I'm not sure if same step can be ran multiple times (wouldn't the resulting `Pods` have same names, leading to some kind of false cache ("this step was already run == this pod already exists") hit?

I considered adding a parameter to the step that would be a list of suites, and the step would iterate over them. I'm worried about the step timeout that we have on the step though.

PR #1501, opened 2020-12-08
ipi-deprovision: fix failure report

bbguimaraes:
Prevents failures such as:

```
ls: cannot access '/logs/artifacts/deprovision/*/failure': No such file or directory
```

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ipi-deprovision/1336292322861649920

### DIFF
--- a/cmd/ipi-deprovision/ipi-deprovision.sh
+++ b/cmd/ipi-deprovision/ipi-deprovision.sh
@@ -69,7 +69,8 @@ EOF
   echo "will deprovision GCE cluster ${infraID} in region ${region}"
 done
 
-for workdir in $( find "${ARTIFACTS}/deprovision" -mindepth 1 -type d | shuf ); do
+clusters=$( find "${ARTIFACTS}/deprovision" -mindepth 1 -type d )
+for workdir in $(shuf <<< ${clusters}); do
   queue deprovision "${workdir}"
 done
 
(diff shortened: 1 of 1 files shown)


cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -89,11 +90,11 @@ if [[ "${#buckets[@]}" -gt 0 ]]; then
   timeout 30m gsutil -m rm -r "${buckets[@]}"
 fi
 
-FAILED="$(ls "${ARTIFACTS}"/deprovision/*/failure)"
+FAILED="$(find ${clusters} -name failure -printf '%H\n')"
```
alvaroaleman, line 93: What does the `%H` do? When I tried this out it always printed a literal dot newline

cmd/ipi-deprovision/ipi-deprovision.sh
```diff
@@ -89,11 +90,11 @@ if [[ "${#buckets[@]}" -gt 0 ]]; then
...
 if [[ -n "${FAILED}" ]]; then
   echo "Deprovision failed on the following clusters:"
-  echo "${FAILED}"
+  sed 's,.*/,,' <<< $FAILED
```
alvaroaleman, line 96: This isn't incorrect, but using the `basename` CLI makes this much easier to undestand ;)

cmd/ipi-deprovision/ipi-deprovision.sh
bbguimaraes:
```
%H     Starting-point under which file was found.
```

cmd/ipi-deprovision/ipi-deprovision.sh
bbguimaraes: Changed to `xargs --max-args 1 basename`.

cmd/ipi-deprovision/ipi-deprovision.sh
alvaroaleman: Doesn't that mean this will always print the value of the `clusters` variable?

bbguimaraes: `$clusters` is a list of directories (see line 72).  `find` will print the directory that originated the search that found each file named `failure`:

```
$ tree
.
└── deprovision
    ├── a
    │   └── failure
    ├── b
    ├── c
    │   └── failure
    └── d

5 directories, 2 files
$ find deprovision/* -name failure -printf '%H\n'
deprovision/a
deprovision/c
$ ARTIFACTS=$PWD ipi-deprovision/ipi-deprovision.sh
Deprovision failed on the following clusters:
a
c
```

PR #1508, opened 2020-12-09
e2e: verify that the boskos server started

bbguimaraes:
There is no status API endpoint, so the name of a valid resource which
will be acquired and released is required.

Output with the current E2E failure:

```
[INFO] Waiting for boskos to be ready...
FAILURE after 0.456s: hack/lib/integration.sh:136: executing 'curl --request POST --silent --show-error --fail 'http://localhost:8080/acquire?type=aws-quota-slice&state=free&dest=leased&owner=test'' expecting success; re-trying every 0.2s until completion or 0.030s: the command timed out
Standard error from the command:
curl: (7) Failed to connect to localhost port 8080: Connection refused
... repeated 2 times
[INFO] Stopping boskos...
/home/bbguimaraes/src/ci-tools/hack/lib/integration.sh: line 153: kill: (298912) - No such process
[ERROR] boskos exited early, logs:
{"component":"unset","error":"failed to construct rest config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined","file":"/home/bbguimaraes/src/boskos/cmd/boskos/boskos.go:113","func":"main.main","level":"fatal","msg":"Failed to get kubeconfig","severity":"fatal","time":"2020-12-09T14:43:15Z"}
[ERROR] PID 298896: hack/lib/integration.sh:137: `return 1` exited with status 1.
[INFO]          Stack Trace:
[INFO]            1: hack/lib/integration.sh:137: `return 1`
[INFO]            2: test/e2e/lease.sh:28: os::integration::boskos::start
[INFO]   Exiting with code 1.
[INFO] [CLEANUP] Killing child processes
```

### DIFF
--- a/hack/lib/integration.sh
+++ b/hack/lib/integration.sh
@@ -124,10 +124,25 @@ readonly -f os::integration::configresolver::check_log
 __os_integration_boskos_pid=
 
 # os::integration::boskos::start starts an in-memory boskos server.
+# The name of a valid resource is required to verify that the server has
+# successfully started.
 function os::integration::boskos::start() {
-    local config=$1
+    local config=$1 test_lease=$2
+    local url=http://localhost:8080
+    local curl='curl --request POST --silent --show-error --fail'
     boskos --in_memory --config "$1" &> "${LOG_DIR}/boskos.log" &
     __os_integration_boskos_pid=$!
+    os::log::info "Waiting for boskos to be ready..."
+    if ! os::cmd::try_until_success "${curl} '${url}/acquire?type=${test_lease}&state=free&dest=leased&owner=test'" 30; then
+        os::integration::boskos::stop
+        cat "${LOG_DIR}/boskos.log"
+        os::log::fatal 'failed to acquire the test lease'
+    fi
+    if ! os::cmd::expect_success "${curl} '${url}/reset?type=${test_lease}&state=leased&dest=free&expire=0'"; then
+        os::integration::boskos::stop
+        cat "${LOG_DIR}/boskos.log"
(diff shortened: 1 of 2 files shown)


hack/lib/integration.sh
```diff
@@ -124,10 +124,25 @@ readonly -f os::integration::configresolver::check_log
...
 function os::integration::boskos::start() {
-    local config=$1
+    local config=$1 test_lease=$2
+    local url=http://localhost:8080
+    local curl='curl --request POST --silent --show-error --fail'
     boskos --in_memory --config "$1" &> "${LOG_DIR}/boskos.log" &
     __os_integration_boskos_pid=$!
+    os::log::info "Waiting for boskos to be ready..."
+    if ! os::cmd::try_until_success "${curl} '${url}/acquire?type=${test_lease}&state=free&dest=leased&owner=test'" 30; then
+        os::integration::boskos::stop
```
alvaroaleman, line 137: if this fails, will we still see the log?

bbguimaraes: Yes.  The `bash` is a bit inscrutable, but I've tested both error scenarios.

That's the source of the error message in the description:

https://github.com/openshift/ci-tools/blob/a4e5d83ca6233425f558ccf6658da259a2bbc198/hack/lib/integration.sh#L139-L141

If the `stop` function succeeds (we failed to acquire but the server is up), then we get:

```bash
cat "${LOG_DIR}/boskos.log"
os::log::fatal 'failed to acquire the test lease'
```

stevekuznetsov (approved):

PR #1512, opened 2020-12-09
ci-operator: allow marking post steps as best-effort

stevekuznetsov:

/assign @alvaroaleman 
Implements https://issues.redhat.com/browse/DPTP-1799

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -637,6 +637,11 @@ type LiteralTestStep struct {
 	// flag is set to true in MultiStageTestConfiguration. This option is
 	// applicable to `post` steps.
 	OptionalOnSuccess *bool `json:"optional_on_success,omitempty"`
+	// BestEffort defines if this step should cause the job to fail when the
+	// step fails. This only applies when AllowBestEffortPostSteps flag is set
+	// to true in MultiStageTestConfiguration. This option is applicable to
+	//`post` steps.
+	BestEffort *bool `json:"best_effort,omitempty"`
 	// ReadonlySharedDir reduces the run time of steps that do not update the
 	// shared directory.
 	ReadonlySharedDir bool `json:"readonly_shared_dir,omitempty"`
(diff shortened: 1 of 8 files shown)


stevekuznetsov:
Made this opt-in as I imagine existing post steps like the installer will remain relevant

pkg/steps/multi_stage.go
```diff
@@ -352,7 +355,15 @@ func (s *multiStageTestStep) runSteps(
 const multiStageTestStepContainerName = "test"
 
 func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []coreapi.EnvVar,
-	hasPrevErrs bool) ([]coreapi.Pod, error) {
+	hasPrevErrs bool) ([]coreapi.Pod, func(string) bool, error) {
+	bestEffort := sets.NewString()
+	isBestEffort := func(podName string) bool {
+		if s.allowBestEffortPostSteps == nil || (s.allowBestEffortPostSteps != nil && !*s.allowBestEffortPostSteps) {
```
bbguimaraes, line 361: Why not just return an empty set from the outer function when this condition is true?

bbguimaraes:
```suggestion
		if s.allowBestEffortPostSteps == nil || !*s.allowBestEffortPostSteps {
```

pkg/api/types.go
```diff
@@ -735,6 +740,10 @@ type MultiStageTestConfiguration struct {
 	// all previous `pre` and `test` steps were successful. The given step must explicitly
 	// ask for being skipped by setting the OptionalOnSuccess flag to true.
 	AllowSkipOnSuccess *bool `json:"allow_skip_on_success,omitempty"`
+	// AllowBestEffortPostSteps defines if any `post` steps can be ignored when
+	// they fail. The given step must explicitly ask for being ignored by setting
+	// the OptionalOnSuccess flag to true.
+	AllowBestEffortPostSteps *bool `json:"allow_best_effort_post_steps,omitempty"`
```
bbguimaraes, line 746: Is this flag necessary given that the default for `BestEffort` is `false`?

bbguimaraes (commented):
For the record, while I am not opposed to this change, it was part of the original design that the steps would be responsible for managing their exit status.

pkg/api/types.go
stevekuznetsov: Without it, a step becomes best-effort on all jobs when it's configuration is changed. Is that what we want?

stevekuznetsov:
@bbguimaraes the complication is that we now have per-step timeouts and having a process exit 0 when terminated - especially when the process is in a script, might be hard. This seems simpler.

pkg/api/types.go
bbguimaraes: Ah, make sense, that's what was trying to understand.

PR #1522, opened 2020-12-11
webreg: surface `from_image` in UI 

petr-muller:
Fixes: [DPTP-1653](https://issues.redhat.com/browse/DPTP-1653)

~Preview is supposed to be on https://server-ci-tools-1522.apps.ci.l2s4.p1.openshiftapps.com but :thisisfine: is happening on app.ci.~

- [example step with `from`](https://server-ci-tools-1522.apps.ci.l2s4.p1.openshiftapps.com/reference/ipi-install-rbac) ([current state](https://steps.ci.openshift.org/reference/ipi-install-rbac))
- [example test with `from_image`](https://server-ci-tools-1522.apps.ci.l2s4.p1.openshiftapps.com/reference/ipi-conf) ([current_state](https://steps.ci.openshift.org/reference/ipi-conf))

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -164,7 +164,8 @@ const mainPage = `
 const referencePage = `
 <h2 id="title"><a href="#title">Step:</a> <nobr style="font-family:monospace">{{ .Reference.As }}</nobr></h2>
 <p id="documentation">{{ .Reference.Documentation }}</p>
-<h3 id="image"><a href="#image">Container image used for this step:</a> <span style="font-family:monospace">{{ .Reference.From }}</span></h3>
+<h3 id="image"><a href="#image">Container image used for this step:</a> <span style="font-family:monospace">{{ fromImage .Reference.From .Reference.FromImage }}</span></h3>
+<p id="image">{{ fromImageDescription .Reference.From .Reference.FromImage }}</p>
 <h3 id="source"><a href="#source">Source Code</a></h3>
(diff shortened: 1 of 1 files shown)


petr-muller:
/test e2e

release-controller flakes:
```
failed to resolve release custom: failed to request latest release: server responded with 504: <html><body><h1>504 Gateway Time-out</h1>
```

alvaroaleman:
@petr-muller the `Github link` seems to be a little too close to the code snipped, making it look broken

pkg/webreg/webreg.go
```diff
@@ -838,6 +839,26 @@ func syntaxBash(source string) (string, error) {
 	return syntax(source, lexers.Get("bash"))
 }
 
+func fromImage(name string, reference *api.ImageStreamTagReference) string {
```
alvaroaleman, line 842: That request is unrelated to your change, but could you move `getBaseTemplate` into an `init` that populates a `var` so that:
* We don't parse the template over and over
* We notice when its broken at unit test time (by making the `init` panic)
* We won't crash the binary when the template is broken and someone visits a page that uses it because in that case we currently just log an error and return a nilpointer: https://github.com/openshift/ci-tools/blob/64c968a7395db4f3dce9706d90ba23cf902fdb41/pkg/webreg/webreg.go#L540-L544

petr-muller: SGTM

PR #1523, opened 2020-12-11
Check app.ci's kubeconfig after loading InClusterConfig

hongkailiu:
Follow up https://github.com/openshift/release/pull/14283
Did not notice that we use the inClusterConfig for app.ci.

/cc @alvaroaleman

### DIFF
--- a/cmd/dptp-controller-manager/main.go
+++ b/cmd/dptp-controller-manager/main.go
@@ -256,15 +256,16 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatal("failed to load kubeconfigs")
 	}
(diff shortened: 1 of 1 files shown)


cmd/dptp-controller-manager/main.go
```diff
@@ -267,6 +264,10 @@ func main() {
 		logrus.Infof("Loaded %q context from in-cluster config", appCIContextName)
 	}
 
+	if _, hasRegistryCluster := kubeconfigs[opts.registryClusterName]; !hasRegistryCluster {
```
alvaroaleman, line 267: We will `Fatalf`  in line 262  if that kubeconfig doesn't exist so I don't think this check adds any value

alvaroaleman: nvm, I misread.
/lgtm

PR #1526, opened 2020-12-11
Fix test_images_distributor when app.ci is registry cluster

hongkailiu:
`buildClusterManagers`  does not contain registry cluster.
When `app.ci` is the registry cluster, the client isn't in the map `r.buildClusterClients`.

Tested locally with or without `--registry-cluster-name=app.ci`.

### DIFF
--- a/pkg/controller/test-images-distributor/test_images_distributor.go
+++ b/pkg/controller/test-images-distributor/test_images_distributor.go
@@ -72,18 +72,26 @@ func AddToManager(mgr manager.Manager,
 	for buildClusterName, buildClusterManager := range buildClusterManagers {
 		buildClusters.Insert(buildClusterName)
 		r.buildClusterClients[buildClusterName] = imagestreamtagwrapper.MustNew(buildClusterManager.GetClient(), buildClusterManager.GetCache())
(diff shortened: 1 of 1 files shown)


pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -72,18 +72,26 @@ func AddToManager(mgr manager.Manager,
...
-				testImageStreamTagImportHandler(),
-			); err != nil {
-				return fmt.Errorf("failed to create watch for testimagestreamtagimports: %w", err)
-			}
-		}
 	}
 
-	objectFilter, err := testInputImageStreamTagFilterFactory(log, configAgent, r.buildClusterClients["app.ci"], resolver, additionalImageStreamTags, additionalImageStreams, additionalImageStreamNamespaces)
+	// TODO: Watch buildCluster ImageStreams as well. For now we assume no one will tamper with them.
+	if err := c.Watch(
```
alvaroaleman, line 78: IMHO both the `c.Watch` and the `testInputImageStreamTagFilterFactory` should simply be using `registryManager.GetCache()`/`registryManager.GetClient()`

alvaroaleman (approved):

PR #1528, opened 2020-12-12
registrySyncer: Add a flag to ignore RC imagestreams

hongkailiu:
Steve wanted this feature during outage.
https://coreos.slack.com/archives/GB7NB0CUC/p1607724296134700

I would like to leave them alone since they could lead to DPTP-CM's infinite looping of DPTP-create and RC-delete.
(although in production DPTP-CM breaks the loop after several hours.)

Need to remove the finalizers manually afterwards.


hongkailiu:
/hold

might not need this at all.

PR #1534, opened 2020-12-15
Specify registry-config for oc adm release new command

hongkailiu:
Follow up https://coreos.slack.com/archives/CB95J6R4N/p1608041576301800?thread_ts=1608025161.293000&cid=CB95J6R4N
and
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/promote-release-openshift-machine-os-content-e2e-aws-4.7/1338794407553404928#1:build-log.txt%3A71

I think the best is to move the job to app.ci.

Otherwise, our source images are on app.ci now. We need to be able to pull from it.
https://github.com/openshift/ci-tools/pull/1102/files

We had a similar case: https://github.com/openshift/ci-tools/pull/1102/files
We cannot use directly in commands because we need to push to the local registry (ci-op-XXXX namespace) as well.
It needs to record the login token to the dockerconfig file.
So we copy it to the temp folder first.

/cc @jupierce @stevekuznetsov

### DIFF
--- a/pkg/steps/release/create_release.go
+++ b/pkg/steps/release/create_release.go
@@ -51,6 +51,7 @@ type assembleReleaseStep struct {
 	client      steps.PodClient
 	artifactDir string
 	jobSpec     *api.JobSpec
+	pullSecret  *coreapi.Secret
 }
 
 func (s *assembleReleaseStep) Inputs() (api.InputDefinition, error) {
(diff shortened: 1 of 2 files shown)


pkg/steps/release/create_release.go
```diff
@@ -177,6 +178,26 @@ func (s *assembleReleaseStep) run(ctx context.Context) error {
 
 	destination := fmt.Sprintf("%s:%s", releaseImageStreamRepo, s.name)
 	log.Printf("Create release image %s", destination)
+
+	var registryConfig, copyCmd string
+	var secrets []*api.Secret
+	if s.pullSecret != nil {
```
hongkailiu, line 184: We will need to mount the secret as well in the periodic jobs.

PR #1536, opened 2020-12-15
Add controller to rotate old serviceaccount secrets

alvaroaleman:
Needed for https://issues.redhat.com/browse/DPTP-1763

### DIFF
--- a/pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
+++ b/pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
@@ -1,0 +1,195 @@
+package serviceaccountsecretrefresher
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+const ControllerName = "serviceaccount_secret_refresher"
+
+func AddToManager(clusterName string, mgr manager.Manager, enabledNamespaces sets.String, removeOldSecrets bool) error {
+	r := &reconciler{
+		client:           mgr.GetClient(),
+		filter:           func(r reconcile.Request) bool { return enabledNamespaces.Has(r.Namespace) },
+		log:              logrus.WithField("controller", ControllerName).WithField("cluster", clusterName),
+		second:           time.Second,
+		removeOldSecrets: removeOldSecrets,
+	}
+	c, err := controller.New(fmt.Sprintf("%s_%s", ControllerName, clusterName), mgr, controller.Options{
+		Reconciler:              r,
+		MaxConcurrentReconciles: 20,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	if err := c.Watch(&source.Kind{Type: &corev1.ServiceAccount{}}, &handler.EnqueueRequestForObject{}); err != nil {
+		return fmt.Errorf("failed to construct watch for ServiceAccounts: %w", err)
+	}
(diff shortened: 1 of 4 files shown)


pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
```diff
@@ -0,0 +1,181 @@
...
+	if !r.removeOldSecrets {
+		return nil
+	}
+
+	var allSecrets corev1.SecretList
+	if err := r.client.List(ctx, &allSecrets, ctrlruntimeclient.InNamespace(req.Namespace)); err != nil {
+		return fmt.Errorf("failed to list secrets in %s namespace: %w", req.Namespace, err)
+	}
+
+	for _, secret := range allSecrets.Items {
```
stevekuznetsov, line 151: What's the trigger for this reconciliation if no SA/Secrets change?

alvaroaleman: the fact that we restart the controller at least daily

alvaroaleman: I'll add a reconcile after

alvaroaleman: fixed

stevekuznetsov (approved):

pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
```diff
@@ -0,0 +1,190 @@
...
+
+	sa := &corev1.ServiceAccount{}
+	if err := r.client.Get(ctx, req.NamespacedName, sa); err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil, nil
+		}
+		return nil, fmt.Errorf("failed to get serviaccount %s: %w", req.String(), err)
+	}
+
+	var imagePullSecetsToKeep []corev1.LocalObjectReference
```
petr-muller, line 98: nit:
```suggestion
	var imagePullSecretsToKeep []corev1.LocalObjectReference
```

pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
```diff
@@ -0,0 +1,190 @@
...
+	var requeueAfter time.Duration
+	for _, pullSecretRef := range sa.ImagePullSecrets {
+		creationTimestamp, err := r.creationTimestampSecret(ctx, req.Namespace, pullSecretRef.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check docker secrets creationTimestamp: %w", err)
```
petr-muller, line 103: nit: "image pull secrets"

pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
```diff
@@ -0,0 +1,190 @@
...
+		l.WithField("imagepullsecretname", pullSecretRef.Name).Info("Not keeping image pull secret")
+	}
+
+	var tokenSecretsToKeep []corev1.ObjectReference
+	for _, tokenSecretRef := range sa.Secrets {
+		creationTimestamp, err := r.creationTimestampSecret(ctx, req.Namespace, tokenSecretRef.Name)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if secret should be kept: %w", err)
+		}
+		if isObjectLessThan30DaysOld(*creationTimestamp) {
```
petr-muller, line 121: semi-nit: should 30 be a param or a const to avoid hardcoding? What if we nee

pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
```diff
@@ -0,0 +1,190 @@
...
+			tokenSecretsToKeep = append(tokenSecretsToKeep, tokenSecretRef)
+			if newReqeueAfter := -time.Since(creationTimestamp.Time.Add(30 * 24 * time.Hour)); requeueAfter == 0 || newReqeueAfter < requeueAfter {
```
petr-muller, line 123: ...especialyy when we prbably want the 30 to match the number above

petr-muller: I'd prefer 30 to simply be a const

alvaroaleman:
@petr-muller fixed. Also added some synchronization in the `serviceaccountSecretRecreatingClient` to avoid the unit test race (races that do not qualify as data race because they are "race to acquire mutex" are the worst)

pkg/controller/serviceaccount_secret_refresher/serviceaccount_secret_refresher.go
petr-muller: @alvaroaleman Let's also rename the method so it does not hardcode the number? So that if we change the number we don't need the method name. Perhaps `isCreatedRecently` or smt

alvaroaleman: @petr-muller updated

petr-muller (approved):

PR #1538, opened 2020-12-16
e2e: implement tests in go

stevekuznetsov:

/hold


test/e2e/framework/framework.go
```diff
@@ -0,0 +1,210 @@
...
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func init() {
+	rand.Seed(time.Now().Unix())
+}
+
+// CompareWithFixture will compare output files with a test fixture and allows to automatically update them
+// by setting the UPDATE env var. The output and golden paths are relative to the test's directory.
+func CompareWithFixture(t *testing.T, golden, output string) {
```
bbguimaraes, line 28:
```suggestion
func CompareWithFixture(t *testing.T, golden, output string) {
	t.Helper()
```

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,241 @@
...
+			t.Parallel()
+			cmd := framework.CiOperatorCommand(t)
+			cmd.Args = append(cmd.Args, append(testCase.args, "--config=config.yaml")...)
+			cmd.Env = append(cmd.Env, `JOB_SPEC={"type":"postsubmit","job":"branch-ci-openshift-ci-tools-master-ci-operator-e2e","buildid":"0","prowjobid":"uuid","refs":{"org":"openshift","repo":"ci-tools","base_ref":"master","base_sha":"6d231cc37652e85e0f0e25c21088b73d644d89ad","pulls":[]}}`)
+			output, err := cmd.CombinedOutput()
+			if os.Getenv("VERBOSE") == "true" {
+				t.Logf("%s: ci-operator output:\n%v", testCase.name, string(output))
+			}
+			if testCase.success != (err == nil) {
+				t.Fatalf("%s: didn't expect an error from ci-operator: %v; output:\n%v", testCase.name, err, string(output))
```
bbguimaraes, line 55: The condition matches both unexpected success and failure.

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,241 @@
...
+		}
+		relPath, err := filepath.Rel("artifacts/template", path)
+		if err != nil {
+			// this should not happen
+			t.Errorf("bug: could not compute relative path in artifact dir: %v", err)
+		}
+		framework.CompareWithFixture(t, path, filepath.Join(artifactDir, "template", relPath))
+		return nil
+	}); err != nil {
+		t.Errorf("failed to walk aritfacts tree for comparison: %v", err)
```
bbguimaraes, line 104:
```suggestion
		t.Errorf("failed to walk artifacts tree for comparison: %v", err)
```

test/e2e/framework/framework.go
stevekuznetsov: We don't print any file info ;)

hack/test-go.sh
```diff
@@ -7,4 +7,4 @@ if [[ -n ${ARTIFACT_DIR:-} ]]; then
   JUNIT_ARG="--junitfile=$ARTIFACT_DIR/junit.xml"
 fi
 
-gotestsum $JUNIT_ARG -- ./... -race ${TESTFLAGS:-}
+gotestsum $JUNIT_ARG -- ${PACKAGES:-"./..."} -race ${TESTFLAGS:-}
```
alvaroaleman, line 10: The common approach to not run tests by default is to use build tags: https://golang.org/pkg/go/build/#hdr-Build_Constraints

This has the advantage that things like this are not needed and ppl can still run `go test ./...`

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,268 @@
...
+
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
+func init() {
+	rand.Seed(time.Now().Unix())
+}
+
+// CompareWithFixtureDir will compare all files in a directory with a corresponding test fixture directory.
+func CompareWithFixtureDir(t *testing.T, golden, output string) {
```
alvaroaleman, line 29: These are not specific to e2e tests, so can we keep them where they are?

Mainly asking because we want the actual e2e tests to be gated by a buildtag but we do not want these helpers to be gated by anything, as they are used in unittests

stevekuznetsov: These are unrelated to what is used in unit

hack/test-go.sh
stevekuznetsov: If someone's running `go test ./...` I want them to have the full suite. Build tags are pretty heavily linked with OS/ARCH and that seems like a misuse to me IMO

alvaroaleman: Arch-specifics and tests are the main use case for build tags, just Google it, you will find plenty of examples for this. No one wants tests that require special setup and won't work by default to run when they run a simple go test. /..... As a sideffrct, this also avoids having to adjust all existing test commands

stevekuznetsov:
I'm testing in prod since the timeout behavior is hard to repro locally

stevekuznetsov:
I think it's ready soon - found a big error in how we pass around secrets

stevekuznetsov:
WE DID IT

```
$ oc --context build01 --namespace ci-op-19mylnkf logs -c test -f e2e-e2e
PACKAGES="./test/e2e/..." TESTFLAGS=" -tags e2e -timeout 120m" hack/test-go.sh
+ gotestsum --junitfile=/tmp/artifacts/junit.xml -- ./test/e2e/... -race -tags e2e -timeout 120m
∅  test/e2e/framework
✓  test/e2e/simple (22m45.327s)

DONE 12 tests in 1367.137s
```

@alvaroaleman ready for review

petr-muller:
The nested ci-op JUnits get mixed up with the top-level one which can be confusing. Can we somehow avoid that? See the failures in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/1538/pull-ci-openshift-ci-tools-master-e2e/1339375577009754112 (which is itself a passing job).

alvaroaleman:
> @alvaroaleman ready for review

Please split e2e tests out from the normal unit tests as asked in  https://github.com/openshift/ci-tools/pull/1538#discussion_r544438125. The standard approach for that is build tags and not packages, see for example https://stackoverflow.com/questions/25965584/separating-unit-tests-and-integration-tests-in-go

stevekuznetsov:
@alvaroaleman it's already using build tags..

stevekuznetsov:
@petr-muller looks like we require a `junit` prefix for those files, added a cleanup helper task to rename them to hopefully avoid that.

hack/test-go.sh
```diff
@@ -7,4 +7,5 @@ if [[ -n ${ARTIFACT_DIR:-} ]]; then
   JUNIT_ARG="--junitfile=$ARTIFACT_DIR/junit.xml"
 fi
 
-gotestsum $JUNIT_ARG -- ./... -race ${TESTFLAGS:-}
+set -o xtrace
+gotestsum $JUNIT_ARG -- ${PACKAGES:-"./..."} -race ${TESTFLAGS:-}
```
alvaroaleman, line 11: We don't need the packages anymore now that w use buildflags, right?

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,327 @@
...
+	)
+	cmd.Env = append(cmd.Env, KubernetesClientEnv(t)...)
+	return cmd, artifactDir
+}
+
+func Execute(t *testing.T, cmd *exec.Cmd, testDone <-chan struct{}, cleanupDone chan<- struct{}) ([]byte, error) {
+	t.Helper()
+	verbose := os.Getenv("VERBOSE") == "true"
+	if verbose {
+		t.Logf("running: %v", cmd.Args)
```
alvaroaleman, line 135: go itself will hide everything you log via `t.Log` unless the test fails or you run the tests with `-v` so IMHO we don't need another verbosity configuration layer

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,327 @@
...
+	if deadline, ok := t.Deadline(); ok {
+		go func() {
+			defer func() {
+				cleanupDone <- struct{}{}
+			}()
+			select {
+			case <-testDone:
+				// nothing to do
+				return
+			case <-time.After(time.Until(deadline.Add(-1 * time.Minute))):
```
alvaroaleman, line 152: You can get rid of this by using `exec.CommandContext`, that will kill the command once the context is cancelled

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,327 @@
...
+}
+
+// BoskosOptions are options for running the boskos server
+type BoskosOptions struct {
+	ConfigPath string
+}
+
+// InstallBoskos begins the boskos server and makes sure it is ready to serve
+// before returning the port it is serving on.
+func InstallBoskos(t *testing.T, opt BoskosOptions) string {
```
alvaroaleman, line 221: This is more of a `RunBoskos` than `InstallBoskos`, isn't it?

test/e2e/framework/framework.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+	return out
+}
+
+func flag(flag, value string) string {
+	return fmt.Sprintf("--%s=%s", flag, value)
+}
+
+// getFreePort asks the kernel for a free open port that is ready to use.
+func getFreePort(t *testing.T) string {
```
alvaroaleman, line 310: I think this needs a lock for threadsafety

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,248 @@
...
+			defer func() {
+				close(testDone)
+				<-cleanupDone
+			}()
+			cmd, _ := framework.CiOperatorCommand(t)
+			cmd.Args = append(cmd.Args, append(testCase.args, "--config=config.yaml")...)
+			cmd.Env = append(cmd.Env, `JOB_SPEC={"type":"postsubmit","job":"branch-ci-openshift-ci-tools-master-ci-operator-e2e","buildid":"0","prowjobid":"uuid","refs":{"org":"openshift","repo":"ci-tools","base_ref":"master","base_sha":"6d231cc37652e85e0f0e25c21088b73d644d89ad","pulls":[]}}`)
+			output, err := framework.Execute(t, cmd, testDone, cleanupDone)
+			if testCase.success != (err == nil) {
+				t.Fatalf("%s: didn't expect an error from ci-operator: %v; output:\n%v", testCase.name, err, string(output))
```
alvaroaleman, line 60: I think we should test for the precise failure we are expecting and not any failure to prevent these tests from rotting

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,248 @@
...
+			for _, line := range testCase.output {
+				if !bytes.Contains(output, []byte(line)) {
+					t.Errorf("%s: could not find line %q in output; output:\n%v", testCase.name, line, string(output))
+				}
+			}
+		})
+	}
+}
+
+func TestTemplate(t *testing.T) {
```
alvaroaleman, line 71: Maybe parallelize this?

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		t.Fatalf("ci-operator failed: %v; output:\n%v", err, string(output))
+	}
+	framework.CompareWithFixtureDir(t, "artifacts/template", filepath.Join(artifactDir, "template"))
+	outputjUnit := filepath.Join(artifactDir, "junit_operator.xml")
+	raw, err := ioutil.ReadFile(outputjUnit)
+	if err != nil {
+		t.Fatalf("could not read jUnit artifact: %v", err)
+	}
+	matcher := regexp.MustCompile(`time=".*"`)
```
alvaroaleman, line 106: compile regexes to a var please so this fails during startup if there is an issue with it

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,248 @@
...
+			for _, line := range []string{`Resolved release ` + testCase.release + ` to`, `Imported release.*to tag release:` + testCase.release} {
+				if !regexp.MustCompile(line).Match(output) {
+					t.Errorf("%s: could not find line %q in output; output:\n%v", testCase.name, line, string(output))
+				}
+			}
+		})
+	}
+}
+
+func TestLiteralDynamicRelease(t *testing.T) {
```
alvaroaleman, line 167: paralelize?

test/e2e/simple/e2e_test.go
```diff
@@ -0,0 +1,248 @@
...
+	}
+	for _, line := range []string{`Using explicitly provided pull-spec for release latest`, `Imported release.*to tag release:latest`} {
+		if !regexp.MustCompile(line).Match(output) {
+			t.Errorf("explicit var: could not find line %q in output; output:\n%v", line, string(output))
+		}
+	}
+}
+
+func TestOptionalOperators(t *testing.T) {
+	testDone, cleanupDone := make(chan struct{}), make(chan struct{})
```
alvaroaleman, line 225: paralelize?

hack/test-go.sh
stevekuznetsov: We do, because `go test ./... -tags e2e` will run the e2e and the rest, too

test/e2e/framework/framework.go
stevekuznetsov: I already use it and no, you can't. We need SIGINT, not SIGKILL, as the comment one line under this explains -.-

test/e2e/framework/framework.go
stevekuznetsov: Why? The `ResolveTPCAddr` should lock implicitly through the kernel, no?

test/e2e/simple/e2e_test.go
stevekuznetsov: We do, literally two lines down we check for precise messages for that reason.

test/e2e/simple/e2e_test.go
stevekuznetsov: I understood that to be for sub-tests, not top-level tests? But I guess both? The docs are confusing.

hack/test-go.sh
stevekuznetsov: (I think we could unset GOARCH or something to get around that? Unclear where the logical AND for the tags is coming from)

test/e2e/framework/framework.go
alvaroaleman: sure but there is a timegap between finding a port and the component that wants to use it stating to use it. In that time, we can re-allocate it to something else.

I think we might have to track the ports we handed out?

stevekuznetsov: That's a good point, we will need to track them and ask for another if we get a repeat

stevekuznetsov: Nothing actually uses this yet (decided to scope this PR to only simple.sh) so I will add this in the next one

stevekuznetsov:
@alvaroaleman addressed comments

alvaroaleman (approved):

stevekuznetsov:
Integration failed to pull an image for 15 minutes ...

/test integration

PR #1547, opened 2020-12-17
test: e2e: run multi-stage in Go

stevekuznetsov:


cmd/ci-operator-configresolver/main.go
```diff
@@ -49,12 +53,15 @@ func gatherOptions() (options, error) {
 	fs.StringVar(&o.prowPath, "prow-config", "", "Path to prow config")
 	fs.StringVar(&o.jobPath, "jobs", "", "Path to job config dir")
 	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
-	fs.StringVar(&o.address, "address", ":8080", "Address to run server on")
-	fs.StringVar(&o.uiAddress, "ui-address", ":8082", "Address to run the registry UI on")
+	fs.StringVar(&o.address, "address", ":8080", "DEPRECATED: Address to run server on")
```
stevekuznetsov, line 56: I didn't find any users of this :|

test/e2e/framework/framework.go
```diff
@@ -206,16 +117,68 @@ func KubernetesClientEnv(t *testing.T) []string {
...
 }
 
+type Accessory interface {
+	Ready()
+	Bind(cmd CiOperatorCommand)
+}
+
+var _ Accessory = &accessory{}
+
+type accessory struct {
```
alvaroaleman, line 127: Hiding the implementation behind an interface like this doesn't really provide any value when there is only one implementation and makes navigating the code much harder

alvaroaleman: (You do this in a variety of places in this PR. Generally it is recommended to construct concrete implementations and define the interface at the consumer side which doesn't really make much sense here so should be omitted)

test/e2e/framework/framework.go
```diff
@@ -206,16 +117,68 @@ func KubernetesClientEnv(t *testing.T) []string {
...
+		if getErr != nil {
+			a.t.Logf("`%s` readiness probe error: %v:", a.command, getErr)
+		}
+		return (resp != nil && resp.StatusCode == http.StatusOK) && getErr == nil, nil
+	}); waitErr != nil {
+		a.t.Fatalf("could not wait for `%s` to be healthy: %v", a.command, waitErr)
+	}
+}
+
+func (a *accessory) Bind(cmd CiOperatorCommand) {
```
alvaroaleman, line 160: I find the `accessory` abstraction extremely confusing, why is this modeled in a way where we do the running in a different level than the readyness checking? It would be much more logical to start the binary in a `Run` method of the `accessory` rather than in an accessory constructor

test/e2e/framework/framework.go
```diff
@@ -206,16 +117,68 @@ func KubernetesClientEnv(t *testing.T) []string {
...
+}
+
+var _ Accessory = &accessory{}
+
+type accessory struct {
+	command    string
+	port       string
+	healthPort string
+
+	binder func(port string, cmd CiOperatorCommand)
```
alvaroaleman, line 132: This would be easier to reason about if the `accessory` just had a `CIOperatorFlags` method rather than getting a `CiOperatorCommand` and then manipulating that. The CIOperatorCommand depends on the accessory, not the other way round like the current implementation suggests

test/e2e/framework/framework.go
```diff
@@ -244,49 +208,54 @@ func RunConfigResolver(t *testing.T, opt ConfigResolverOptions) string {
 	})...)
 }
 
-func runAccessory(t *testing.T, command string, args ...string) string {
-	t.Helper()
-	if _, err := exec.LookPath(command); err != nil {
-		t.Fatalf("no `%s` binary was found: %v", command, err)
-	}
+func runAccessory(binder func(port string, cmd CiOperatorCommand), command string, args ...string) AccessoryFunc {
```
alvaroaleman, line 211: if `command` was a `func(port, healthport int) string` you wouldn't need to require all commands to have identically named flags for port/health-port

test/e2e/multi-stage/e2e_test.go
```diff
@@ -0,0 +1,118 @@
...
+			name:    "step without best effort post-step failure",
+			args:    []string{"--unresolved-config=best-effort.yaml", "--target=best-effort-failure"},
+			env:     []string{defaultJobSpec},
+			success: false,
+			output:  []string{`could not run steps: step best-effort-failure failed`},
+		},
+	}
+
+	for _, tc := range testCases {
+		testCase := tc
```
alvaroaleman, line 98: nit: Copying this into a different variable is extremely prone to make ppl in the future accidentally use the wrong one, it would be better to shadow it instead to not make that possible

test/e2e/framework/test.go
```diff
@@ -0,0 +1,146 @@
...
+			<-cleanupDone
+		}()
+		cmd.testDone = testDone
+		cmd.cleanupDone = cleanupDone
+
+		var a []Accessory
+		for _, accessory := range accessories {
+			// binding the accessory to ctx ensures it's lifetime is only
+			// as long as the test we are running in this specific case
+			a = append(a, accessory(&bottom, ctx))
```
alvaroaleman, line 132: Nit: I would try to paralelize the starting of the supplemental processes

test/e2e/multi-stage/e2e_test.go
stevekuznetsov: There was *something* goofy with not copying it ... I can't remember

test/e2e/framework/test.go
stevekuznetsov: eek, another layer of synchronization would be hell. these either pass in <2s or fail so IDK if it's worth it

test/e2e/framework/framework.go
stevekuznetsov: I guess I'm a proponent of not letting people do the wrong thing unless they really mean it. I don't want to expose this since you can get into really hairy, irritating deadlocks and timeouts if you mess with things ...

test/e2e/framework/framework.go
stevekuznetsov: I guess I'm a proponent of not letting people do the wrong thing unless they really mean it. I don't want to expose this since you can get into really hairy, irritating deadlocks and timeouts if you mess with things ...

alvaroaleman: That makes sense, but you already get that by having put these types into a distinct package and only exporting a subset of their fields and methods

test/e2e/framework/framework.go
```diff
@@ -206,20 +117,106 @@ func KubernetesClientEnv(t *testing.T) []string {
...
+	ctx, cancel := context.WithCancel(parentCtx)
+	t.Cleanup(func() {
+		t.Logf("cleanup: killing `%s` process", a.command)
+		cancel()
+	})
+	cmd := exec.CommandContext(ctx, a.command, append(a.args, flags(map[string]string{
+		"port":        a.port,
+		"health-port": a.healthPort,
+	})...)...)
+	t.Logf("running: %v", cmd.Args)
```
alvaroaleman, line 141: Nit: Would make sense to also include the `command` here or not?

alvaroaleman (approved):

alvaroaleman:
/lgtm cancel
Failures are genuine

test/e2e/framework/framework.go
stevekuznetsov: It ends up as `Args[0]` actually, lol

stevekuznetsov:
/hold cancel

Unclear why that one config failed to unmarshal... other tests use it as well and have no issues?

alvaroaleman (approved):

PR #1560, opened 2020-12-30
ci-operator: resolve imported release image pullspec better

stevekuznetsov:
The image reference field in the image import API is documented as being
the correct place from which to pull the imported image, but from
experience we see that it's just the image reference in the stream,
pointing to the image that was imported. As the mechanisms for
authenticating the import are different from the mechanisms used for
using the image directly, it's possible (and likely) that a client that
may import the image may not be able to use it directly. Furthermore,
there's no utility to using the image reference, without a direct pull,
as that is the entire point in doing the import in the first place. By
using out image reference resolution code we can make sure that the
fully resolved pull-spec that we end up using is a local reference and
the kubelet will be able to pull it correctly.

/assign @smarterclayton

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -160,9 +160,11 @@ TMPDIR ?= /tmp
 #   make e2e
 #   make e2e SUITE=multi-stage
 e2e:
-	PACKAGES="./test/e2e/..." TESTFLAGS="$(TESTFLAGS) -tags e2e -timeout 120m -parallel 100" hack/test-go.sh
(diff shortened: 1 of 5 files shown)


smarterclayton:
I will look at the import api and see whether fixing this is tractable.  It *may* be the same as the source so that you can locate the outcome of an import in status (if you are importing multiple), but we can probably add another field.

stevekuznetsov:
We now resolve to the local registry but still grab the external name, which is not useful. Let me see what we do for the others. ..

PR #1566, opened 2021-01-04
steps: Use the pull secret passed to import release from the command

smarterclayton:
The release extract command needs access to all of the locations for
the images in a release image, not just the release image itself.
The command needs to use the pull secret mounted into the pod by
copying it into the default location and then using login to add
the local cluster's auth.

### DIFF
--- a/pkg/steps/release/import_release.go
+++ b/pkg/steps/release/import_release.go
@@ -207,10 +207,8 @@ func (s *importReleaseStep) run(ctx context.Context) error {
 		return fmt.Errorf("unable to tag the 'cli' image into the stable stream: %w", err)
 	}
 
-	var registryConfig string
 	var secrets []*api.Secret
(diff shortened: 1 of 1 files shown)


pkg/steps/release/import_release.go
```diff
@@ -219,6 +219,10 @@ func (s *importReleaseStep) run(ctx context.Context) error {
 	commands := fmt.Sprintf(`
 set -euo pipefail
 export HOME=/tmp
+mkdir -p $HOME/.docker
+if [[ -f /pull ]]; then
+	cp /pull $HOME/.docker/config.json
```
alvaroaleman, line 224: must be `[[ -d` and the source location is wrong, I opened https://github.com/openshift/ci-tools/pull/1567/files and think its correct there
/hold

PR #1570, opened 2021-01-04
Import: Use the same writeable registry secret everywhere

alvaroaleman:


stevekuznetsov:
/hold

I think this is the wrong approach. The `oc registry login` will give rights to the local registry, but in all cases (except those on app.ci if we ever have any) the release we're pulling in will *not* be on the local registry. IMHO we should remove the HOME crap and the registry login entirely.

alvaroaleman:
The issue is that the secret is simply not mounted for a large number of jobs, https://github.com/openshift/release/pull/14591 fixes that.

smarterclayton:
That is not correct.  The pull secret should be for "all source locations".  Releases can come from the local cluster.  The local cluster isn't requried to be one of the source locations.

stevekuznetsov:
The secret that's mounted in provides all of that already, so yes it's sufficient.

stevekuznetsov:
```
$ oc --context app.ci --namespace ci extract secret/registry-pull-credentials --to=- | jq '.auths | keys'
# .dockerconfigjson
[
  "image-registry.openshift-image-registry.svc.cluster.local:5000",
  "image-registry.openshift-image-registry.svc:5000",
  "registry.ci.openshift.org",
  "registry.svc.ci.openshift.org"
]
```

This secret - on every cluster - always contains the credentials required to pull from the source cluster, the source cluster's public hostname, and the central clusters. We did this because SA pull token minting was broken for the source cluster's public hostname.

PR #1571, opened 2021-01-04
Test image distributor: Name pull secret after original

alvaroaleman:
This PR changes the test-image-distributor to use the same pull secret name when creating the secret as it has at the source: https://github.com/openshift/release/blob/5ee2cd373314273f0be04dec82fa842c2c36c178/clusters/app.ci/assets/dptp-controller-manager.yaml#L256

This avoids confusion, especially in app.ci where we end up having the secret twice under different names.

We can not rename the source secret, because it is referenced at various places throughout our jobs.

I will manually delete the old secret everywhere after this was deployed, to avoid issues the next time we rotate

### DIFF
--- a/pkg/controller/registrysyncer/registrysyncer.go
+++ b/pkg/controller/registrysyncer/registrysyncer.go
@@ -26,6 +26,7 @@ import (
 	imagev1 "github.com/openshift/api/image/v1"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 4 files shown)


stevekuznetsov (approved):

hongkailiu (approved):

hongkailiu:
/hold

remove the syncing of the secret from the registry-syncer?

PR #1578, opened 2021-01-06
[DPTP-1789] Ignore context cancelled errors from steps when the job is interrupted

droslean:
In this example [job](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/eclipse_che-operator/553/pull-ci-eclipse-che-operator-master-v4-che-operator-update/1332264893667086336#1:build-log.txt%3A34) the ci-operator reports the cancelled context error from all of the interrupted steps. 

This PR changes this logic and allows only the `interrupted` error to be reported in the results server.

### DIFF
--- a/pkg/steps/run.go
+++ b/pkg/steps/run.go
@@ -56,7 +56,9 @@ func Run(ctx context.Context, graph []*api.StepNode) (*junit.TestSuites, []api.C
 			stepDetails = append(stepDetails, out.stepDetails)
 			if out.err != nil {
 				testCase.FailureOutput = &junit.FailureOutput{Output: out.err.Error()}
-				executionErrors = append(executionErrors, results.ForReason("step_failed").WithError(out.err).Errorf("step %s failed: %v", out.node.Step.Name(), out.err))
+				if out.err != context.Canceled {
+					executionErrors = append(executionErrors, results.ForReason("step_failed").WithError(out.err).Errorf("step %s failed: %v", out.node.Step.Name(), out.err))
+				}
 			} else {
 				seen = append(seen, out.node.Step.Creates()...)
 				if !interrupted {
--- a/pkg/steps/run_test.go
+++ b/pkg/steps/run_test.go
@@ -3,12 +3,18 @@ package steps
 import (
 	"context"
 	"errors"
+	"fmt"
+	"reflect"
+	"sort"
 	"sync"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/results"
 )
 
 type fakeStep struct {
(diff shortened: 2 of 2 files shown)


pkg/steps/run.go
```diff
@@ -56,7 +56,9 @@ func Run(ctx context.Context, graph []*api.StepNode) (*junit.TestSuites, []api.C
 			stepDetails = append(stepDetails, out.stepDetails)
 			if out.err != nil {
 				testCase.FailureOutput = &junit.FailureOutput{Output: out.err.Error()}
-				executionErrors = append(executionErrors, results.ForReason("step_failed").WithError(out.err).Errorf("step %s failed: %v", out.node.Step.Name(), out.err))
+				if out.err != context.Canceled {
```
alvaroaleman, line 59: Can you add a testcase for this so we don't regress?

droslean: Well, it seems that we don't even have tests for this. Sounds like a good opportunity to write some.

petr-muller: Should this be `&& interrupted ` or am I missing something? I think there are other causes for cancellations, and I thnk we're interested in them.

droslean: We check this in the above case.

stevekuznetsov: How are we checking for `interrupted`?

droslean: we are passing already a context.WithCancel, so this will be checked in the `context.Done()` case

stevekuznetsov: Hmm I guess so the issue here is that we cannot tell *why* we were interrupted - was it a timeout or kubelet killing the pod or whatever. I don't think there's any way around that. SGTM

pkg/steps/run_test.go
```diff
@@ -40,135 +46,289 @@ func (*fakeStep) Objects() []ctrlruntimeclient.Object { return nil }
...
 
-	suites, _, err := Run(context.Background(), api.BuildGraph([]api.Step{root, other, src, bin, testBin, rpm, unrelated, final}))
-	if err == nil {
-		t.Error("got no error but expected one")
-	}
-	if suites.Suites != nil && len(suites.Suites) != 1 || len(suites.Suites[0].TestCases) != 6 || suites.Suites[0].NumTests != 6 || suites.Suites[0].NumFailed != 1 {
-		t.Errorf("unexpected junit output: %#v", suites.Suites[0])
-	}
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
```
stevekuznetsov, line 289: Won't this double-cancel for cases that don't need a cancel? Is that not an error?

pkg/steps/run_test.go
```diff
@@ -40,135 +46,289 @@ func (*fakeStep) Objects() []ctrlruntimeclient.Object { return nil }
...
+			if tc.cancelled {
+				cancel()
+			}
+			suites, _, errs := Run(ctx, api.BuildGraph(steps))
+			if errs == nil && len(tc.errExpected) > 0 {
+				t.Error("got no error but expected one")
+			}
+
+			if errs != nil && len(tc.errExpected) > 0 {
+				var expectedErrorsString []string
```
stevekuznetsov, line 299: Why all of this complex error unwrapping logic instead of a `reflect.DeepEqual` ?

droslean: It panics because of unexported fields. I made it this way to make it more readable if a test fails. I will try to check this out though.

pkg/steps/run_test.go
droslean: This will run after we fail or succeed. In other words, we don't need it, but linter complains :)

pkg/steps/run_test.go
petr-muller: Maybe we can use `cmp.Diff` and its ability to forciby compare (or ignore) unexported fields? https://godoc.org/github.com/google/go-cmp/cmp#AllowUnexported

droslean: Thanks @petr-muller! I will check this out

droslean: Well, it seems that `AllowUnexported` works only for structs. In our case, we have slices of structs. I wasn't able to find any solution. I guess we can stick with this approach.

petr-muller: It definitely works over slices: https://play.golang.org/p/5ze2Rw6gwBT

pkg/results/error.go
```diff
@@ -120,3 +120,20 @@ func DefaultReason(err error) error {
 
 	return ForReason(ReasonUnknown).ForError(err)
 }
+
+func ConvertToErrorList(errs []error) []*Error {
+	var ret []*Error
+	for _, e := range errs {
+		ret = append(ret, e.(*Error))
```
alvaroaleman, line 127: Please do not put code that can just panic in non-test files

pkg/steps/run_test.go
```diff
@@ -40,135 +44,282 @@ func (*fakeStep) Objects() []ctrlruntimeclient.Object { return nil }
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			if tc.cancelled {
+				cancel()
+			}
+			suites, _, errs := Run(ctx, api.BuildGraph(steps))
+			if errs == nil && len(tc.errExpected) > 0 {
+				t.Error("got no error but expected one")
+			}
+			actualErrors := results.MapMessagesByReason(results.ConvertToErrorList(errs))
```
alvaroaleman, line 298: Can't we simply do this via a string instead if two funcs that need to be used in conjunction and require some weird data structure as input? Something like `func DeterministicErrorMsg(errs []error) (string, error)`

pkg/steps/run_test.go
```diff
@@ -40,135 +44,282 @@ func (*fakeStep) Objects() []ctrlruntimeclient.Object { return nil }
...
-		creates:   []api.StepLink{api.InternalImageLink("unrelated")},
-	}
-	final := &fakeStep{
-		name:      "final",
-		shouldRun: true,
-		requires:  []api.StepLink{api.InternalImageLink("unrelated")},
-		creates:   []api.StepLink{api.InternalImageLink("final")},
+func TestStepsRun(t *testing.T) {
+	testCases := []struct {
+		id          string
```
alvaroaleman, line 49: the fact that the name attribute of your testcases is called `id` and not `name` is btw a good way to universally identify testcases originally written by you :)

pkg/steps/run_test.go
droslean: I don't want to compare only the error messages, but the reason too.

pkg/steps/run_test.go
droslean: That's my signature :P

pkg/steps/run_test.go
alvaroaleman: And the reason is not part of the error string or what?

droslean: We can have different messages for the same reason. That's why I am using a `map[string][]string`

pkg/results/error.go
droslean: addressed

pkg/steps/run_test.go
alvaroaleman: I still don't follow, sorry, So the error message only includes one of `Reason` and `Other error details`? Otherwise you could simply string sort

droslean: the error message doesn't include any `Reason`. This is another field of the struct.

alvaroaleman: Ah, can you then just iterate through all untyped errors, type assert, if match store a a string of `reason: message` in a string slice, if not just the message, sort the slice and join with `\n`? That would make using this significantly easier

droslean: This is exactly what I did in the first version of this PR :P

droslean: nevermind. I reverted it back. Can you check?

alvaroaleman (approved):

PR #1580, opened 2021-01-06
ServiceAccountRefresher: Allow excluding ServiceAccounts

alvaroaleman:
We can not just rotate all secrets in a namespace, some need to be
excluded.


alvaroaleman:
Actually, nevermind. This is what the grace period is for

PR #1582, opened 2021-01-07
determinize-ci-operator: Use TEST_TYPE instead of TEST_COMMAND

smarterclayton:
The openshift-e2e-test step is changing to focus on test types
(upgrade, upgrade then test, multiple upgrades) and TEST_COMMAND
is no longer the appropriate mapping for an upgrade test. Update
the migrator to take that into account.


alvaroaleman:
@smarterclayton will someone update all the existing jobs that already use TEST_COMMAND?
/lgtm
/approve

PR #1584, opened 2021-01-07
applyconfig: optionally allow server-side dry-runs with validation

petr-muller:
Make applyconfig run `oc version` to detect the version of OpenShift we
run against, and select the set of parameters used for dry run based on
the version (it can be overriden by the `--dry-run-method` parameter).

The tool prefers stricter, server-side validation so it is intentionally
selected on various detection errors.

Initially, the detection is opt-in and client dry-run is selected by
default, because our manifests currently do not pass stricter
validation. I want to fix our manifests by running rehearsal with the
auto mode enabled, then we can switch the auto mode to be the default,
without the need to change the actual jobs.

### DIFF
--- a/cmd/applyconfig/applyconfig.go
+++ b/cmd/applyconfig/applyconfig.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"bytes"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
@@ -11,6 +12,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/hashicorp/go-version"
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -22,21 +24,30 @@ import (
 )
 
 type command string
+type dryRunMethod string
 
 type options struct {
-	confirm     bool
 	user        *nullableStringFlag
 	directories flagutil.Strings
 	context     string
 	kubeConfig  string
+	dryRun      dryRunMethod
 }
 
 const (
 	ocApply   command = "apply"
 	ocProcess command = "process"
+	ocVersion command = "version"
+
+	dryNone   dryRunMethod = ""
+	dryAuto   dryRunMethod = "auto"
+	dryServer dryRunMethod = "server"
+	dryClient dryRunMethod = "client"
 )
 
 const defaultAdminUser = "system:admin"
+
+var validDryRunMethods = []string{string(dryAuto), string(dryClient), string(dryServer)}
 
 type nullableStringFlag struct {
 	val     string
@@ -54,12 +65,20 @@ func (n *nullableStringFlag) Set(val string) error {
 }
 
 func gatherOptions() *options {
-	opt := &options{user: &nullableStringFlag{}}
-	flag.BoolVar(&opt.confirm, "confirm", false, "Set to true to make applyconfig commit the config to the cluster")
+	// nonempy dryRun is a safe default (empty means not a dry run)
+	// TODO(muller): client dry run is default until we pass server one.
+	opt := &options{user: &nullableStringFlag{}, dryRun: dryClient}
+
+	var confirm bool
+	flag.BoolVar(&confirm, "confirm", false, "Set to true to make applyconfig commit the config to the cluster")
 	flag.Var(opt.user, "as", "Username to impersonate while applying the config")
 	flag.Var(&opt.directories, "config-dir", "Directory with config to apply. Can be repeated multiple times.")
 	flag.StringVar(&opt.context, "context", "", "Context name to use while applying the config")
 	flag.StringVar(&opt.kubeConfig, "kubeconfig", "", "Path to the kubeconfig file to apply the config")
+
+	var dryMethod string
+	dryRunMethods := strings.Join(validDryRunMethods, ",")
+	flag.StringVar(&dryMethod, "dry-run-method", string(opt.dryRun), fmt.Sprintf("Method to use when running when --confirm is not set to true (valid values: %s)", dryRunMethods))
 	flag.Parse()
 
 	if len(opt.directories.Strings()) < 1 || opt.directories.Strings()[0] == "" {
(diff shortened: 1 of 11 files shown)


cmd/applyconfig/applyconfig.go
```diff
@@ -54,23 +65,46 @@ func (n *nullableStringFlag) Set(val string) error {
...
+	var confirm bool
+	flag.BoolVar(&confirm, "confirm", false, "Set to true to make applyconfig commit the config to the cluster")
 	flag.Var(opt.user, "as", "Username to impersonate while applying the config")
 	flag.Var(&opt.directories, "config-dir", "Directory with config to apply. Can be repeated multiple times.")
 	flag.StringVar(&opt.context, "context", "", "Context name to use while applying the config")
 	flag.StringVar(&opt.kubeConfig, "kubeconfig", "", "Path to the kubeconfig file to apply the config")
+
+	var dryMethod string
+	dryRunMethods := strings.Join(validDryRunMethods, ",")
+	flag.StringVar(&dryMethod, "dry-run-method", string(opt.dryRun), fmt.Sprintf("Method to use when running when --confirm is not set to true (valid values: %s)", dryRunMethods))
```
alvaroaleman, line 81: This seems overly complicated to me. Why not simply add a boolean for one of the two dry runs, then use dry-run when confirm=calse and switch the method based on the bool?

cmd/applyconfig/applyconfig.go
```diff
@@ -119,13 +153,23 @@ type configApplier struct {
...
+func makeOcApply(kubeConfig, context, path, user string, dry dryRunMethod) *exec.Cmd {
 	cmd := makeOcCommand(ocApply, kubeConfig, context, path, user)
-	if dry {
-		cmd.Args = append(cmd.Args, "--dry-run")
+	switch dry {
+	case dryAuto:
+		logrus.Warn("BUG: Automated dryrun detection should be performed earlier; Using server-side validation")
+		fallthrough
+	case dryServer:
+		cmd.Args = append(cmd.Args, "--dry-run=server", "--validate=true")
```
alvaroaleman, line 166: Some openshift resources still do not have proper openapi schema, we can not use validate=true (I can tell from experience, because I usually use `kubectl` and it reguarly needs a `--validate=false` (it defaults to true)`

Most prominent example are the various openshift api groups that can be accessed by using no api group rather than the real one

cmd/applyconfig/applyconfig.go
```diff
@@ -22,22 +24,31 @@ import (
...
+	dryRun      dryRunMethod
 }
 
 const (
 	ocApply   command = "apply"
 	ocProcess command = "process"
+	ocVersion command = "version"
+
+	dryNone   dryRunMethod = ""
+	dryAuto   dryRunMethod = "auto"
```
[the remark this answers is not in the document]
alvaroaleman: Ah I did indeed miss your detection code. But then follow-up: Why would we ever not want to use `auto`?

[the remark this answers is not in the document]
petr-muller: I think we'll want `auto` most of the time. My plan was to make it default *after* I fix the validation problems in our manifests (using a `--dry-run-method=auto` rehearsal - https://github.com/openshift/release/pull/14693) -- I assumed they can be fixed, not sure about that now seeing your other comment.

I generally like to have flags of this kind, they allow flexibility when needed (dunno, `auto` can guess wrong and we may need to keep PRs going, that kind of thing). I don't think it's that much of a burden, `auto` has 99% of the complexity anyway.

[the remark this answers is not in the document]
alvaroaleman: They can be fixed, I think all of these issues is ppl using `v1` as the apiGroup instead of the real apigroup for resources served by the openshift apiserver. For compatibility reasons `oc` translates that, but it breaks the openapi validation.

cmd/applyconfig/applyconfig.go
petr-muller: Good insight. I only really tested it on our app.ci manifests, and on a simple eyeballing the results did not look that bad, but I haven't looked at the details too much.

Could you point me to one of our manifests where this would be currently unfixable?

cmd/applyconfig/applyconfig.go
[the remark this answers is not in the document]
alvaroaleman: Which is deprecated, but oc users do not see that so no one has an incentive to stop cargoculting it from fold manifests: https://coreos.slack.com/archives/CB48XQ4KZ/p1610042971090100

cmd/applyconfig/applyconfig.go
alvaroaleman: It is fixable, it's just quite a bit of a chore :)

alvaroaleman: (I think, I am not entirely sure tbh. Try it out :) Historically, it was just broken and until today, `oc` defaults the `--validate` flag to `false`. )

petr-muller:
@alvaroaleman after the conversation, is there something that you'd prefer to be changed in this PR. My plan is the following:

1. Merge this PR. `client` is default in this PR, so the jobs will not change behavior (modulo bugs that I made :D )
2. Use https://github.com/openshift/release/pull/14693 to make our manifests validation-confomant
3. Switch the method default to auto, which will start using server+validation dry run, without needing to change the jobs.

I think it's not that big effort to fix the manifests. I tried today and I'm mostly done, I think.

cmd/applyconfig/applyconfig.go
```diff
@@ -298,14 +342,68 @@ func init() {
...
 	}
 
+	if o.dryRun == dryAuto {
+		logrus.Info("Detecting dry run capabilities")
+		executor := commandExecutor{}
+		// Generally we want to always prefer stronger (server) validation unless
+		// we *know* we need to use the weak, client one -> so use server on errors
+		cmd := makeOcCommand(ocVersion, o.kubeConfig, o.context, "", o.user.val, "--output=json")
+		out, err := executor.runAndCheck(cmd, "detect server version")
+		if err == nil {
```
alvaroaleman, line 388: This is really hard to read, requires two levels of conditional logic and two places where we do defaulting. Please just factor all the detection logic into a new func that returns the dry run method or an error and make the caller log that error and default to server if it occurs

petr-muller: Addressed. I refactored it slightly differently. Extracting to a func allows using guard clause which reads quite nicely, so I did that.

alvaroaleman (approved):

PR #1600, opened 2021-01-13
webreg: collect and expose workflow/chain dependencies in UI

petr-muller:
Resolves: [DPTP-1652](https://issues.redhat.com/browse/DPTP-1652)

Expose dependencies of the steps involved in the workflows and chains together with the envvars that expose them to tests. The workflow view also shows whether the value is set by the workflow or the step. The data structure for the presentation is a bit complicated because the same dependency can be exposed under different names in multiple steps.

Examples:
- Workflow: https://server-ci-tools-1600.apps.ci.l2s4.p1.openshiftapps.com/workflow/openshift-upgrade-gcp#dependencies
- Chain: https://server-ci-tools-1600.apps.ci.l2s4.p1.openshiftapps.com/chain/ipi-install-stableinitial#dependencies

The code needs a bit more polishing to be mergeable but I am interested in the feedback for how the data is presented before I do that.
/hold

### DIFF
--- a/pkg/webreg/webreg.go
+++ b/pkg/webreg/webreg.go
@@ -16,6 +16,7 @@ import (
 	"github.com/alecthomas/chroma/styles"
 	"github.com/sirupsen/logrus"
 
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/repoowners"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):

petr-muller:
@stevekuznetsov PTAL, I added tests and made some code more robust

stevekuznetsov (approved):
/hold cancel

PR #1606, opened 2021-01-15
repo-init: copycat CVO merge criteria instead of origin

petr-muller:
o/origin is now a snowflake (its newer branches do not participate in
the OCP process) so we should no longer use it as an anchor for adding
new repos.


stevekuznetsov (approved):

petr-muller:
@stevekuznetsov I had to update, can you look again? Do you mind about us copying merge criteria for CVO, but e.g. plugin cfg and few other elements still from `origin`? I started to base everything on CVO but then the change grew a lot, mostly useless because origin is only special in Tide config, nowhere else.

stevekuznetsov (approved):
Yeah, that *should* be fine, our plugin configurations are pretty similar across the board.

PR #1611, opened 2021-01-15
Add azure-arc cluster profile

fabianofranz:
Add an additional Azure profile for clusters with the Arc-enabled
Kubernetes feature.

For CI purposes, Arc-enabled clusters have absolutely no difference
when compared to other `azure4` clusters and are created the same way,
however they [must run in specific Azure regions](https://docs.microsoft.com/en-us/azure/azure-arc/kubernetes/connect-cluster#supported-regions) where that feature is
already provided.

In this PR we create a new cluster profile to allow us to have a
separate Boskos quota slice only with the supported regions, but
everything else (cluster installation, etc) must work like a regular 
`azure4` cluster.

Related to: https://github.com/openshift/release/pull/14798

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -828,6 +828,7 @@ const (
 	ClusterProfileAWSGluster         ClusterProfile = "aws-gluster"
 	ClusterProfileAzure              ClusterProfile = "azure"
 	ClusterProfileAzure4             ClusterProfile = "azure4"
+	ClusterProfileAzureArc           ClusterProfile = "azure-arc"
 	ClusterProfileGCP                ClusterProfile = "gcp"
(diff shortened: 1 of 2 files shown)


fabianofranz:
@bbguimaraes mind taking a look to make sure this does what I'm expecting it to do?

TL;DR: I need my job to run in Azure but for [this](https://docs.microsoft.com/en-us/azure/azure-arc/kubernetes/connect-cluster#supported-regions) reason it can only use one specific region which is `eastus`.

So my proposal is to have a new entry in `generate-boskos.py` that looks like the one below. The `cluster-profile` setting in my test would be `azure-arc`, but from that point the cluster leasing and everything else happens like it was `azure4`.

```
azure-arc': {
        'eastus': 2
    },
```

Makes sense?

PR #1613, opened 2021-01-18
[DPTP-1836] ci-secret-bootstrap: allow recreating secrets on `--force` when an immutable field changes

droslean:
When `--force` is used and an immutable field needs to be changed, we allow recreating the secret

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -416,39 +416,58 @@ func constructSecrets(ctx context.Context, config secretbootstrap.Config, bwClie
 
 func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secretsMap map[string][]*coreapi.Secret, force bool) error {
 	var errs []error
+	shouldCreate := false
 	for cluster, secrets := range secretsMap {
 		for _, secret := range secrets {
-			logrus.Debugf("handling secret: %s:%s/%s", cluster, secret.Namespace, secret.Name)
-			secretsGetter := secretsGetters[cluster]
-			if existingSecret, err := secretsGetter.Secrets(secret.Namespace).Get(context.TODO(), secret.Name, meta.GetOptions{}); err == nil {
+			logger := logrus.WithFields(logrus.Fields{"cluster": cluster, "namespace": secret.Namespace, "name": secret.Name, "type": secret.Type})
+			logrus.Debugf("handling secret")
+
+			secretClient := secretsGetters[cluster].Secrets(secret.Namespace)
+
+			existingSecret, err := secretClient.Get(context.TODO(), secret.Name, meta.GetOptions{})
+			if err != nil && !kerrors.IsNotFound(err) {
+				errs = append(errs, fmt.Errorf("error reading secret %s:%s/%s: %w", cluster, secret.Namespace, secret.Name, err))
+			}
+
+			if err == nil {
 				if secret.Type != existingSecret.Type {
-					errs = append(errs, fmt.Errorf("cannot change secret type from %q to %q (immutable field): %s:%s/%s", existingSecret.Type, secret.Type, cluster, secret.Namespace, secret.Name))
-					continue
-				}
-				for k, v := range existingSecret.Data {
-					if _, exists := secret.Data[k]; exists {
+					if !force {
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -419,10 +419,6 @@ func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secret
 			logrus.Debugf("handling secret: %s:%s/%s", cluster, secret.Namespace, secret.Name)
 			secretsGetter := secretsGetters[cluster]
 			if existingSecret, err := secretsGetter.Secrets(secret.Namespace).Get(context.TODO(), secret.Name, meta.GetOptions{}); err == nil {
-				if secret.Type != existingSecret.Type {
```
alvaroaleman, line 422: This field is immutable, that is why the check exists. You need to recognize this situation and recreate the secret, rather than updating it

droslean: Yes I checked that too. I will update this PR to recreate the secret instead. Patching will make sense here.

pkg/util/secret.go
```diff
@@ -62,3 +66,35 @@ func UpdateSecret(ctx context.Context, client ctrlruntimeclient.Client, secret *
...
+
+type SecretOperationCase string
+
+const (
+	SecretCreate   SecretOperationCase = "create"
+	SecretUpdate   SecretOperationCase = "update"
+	SecretRecreate SecretOperationCase = "recreate"
+)
+
+func OperateOnSecret(secretClient coreclientset.SecretInterface, secret *coreapi.Secret, operationCase SecretOperationCase) error {
```
stevekuznetsov, line 78: What is the point of this misdirection?

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -1667,131 +1667,6 @@ func TestUpdateSecrets(t *testing.T) {
 				},
 			},
 		},
-		{
```
stevekuznetsov, line 1670: Why did we only delete test cases?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -414,40 +414,42 @@ func constructSecrets(ctx context.Context, config secretbootstrap.Config, bwClie
...
+			logger := logrus.WithFields(logrus.Fields{"cluster": cluster, "namespace": secret.Namespace, "name": secret.Name, "type": secret.Type})
+			logger.Debug("handling secret")
+
+			secretClient := secretsGetters[cluster].Secrets(secret.Namespace)
+			existingSecret, err := secretClient.Get(context.TODO(), secret.Name, meta.GetOptions{})
+			if err == nil {
 				if secret.Type != existingSecret.Type {
-					errs = append(errs, fmt.Errorf("cannot change secret type from %q to %q (immutable field): %s:%s/%s", existingSecret.Type, secret.Type, cluster, secret.Namespace, secret.Name))
-					continue
+					logrus.WithField("existing-type", existingSecret.Type).Info("immutable field needs to be updated. Will re-create the secret")
```
alvaroaleman, line 428: I think what we want to do is much easier expressed by issuing a Delete here and ignoring IsNotFound for that and otherwise keeping everything as-is

cmd/ci-secret-bootstrap/main_test.go
droslean: Those test cases expecting an error for the immutable field. I will add a test case that makes sure that the re-creating is happening correctly

pkg/util/secret.go
droslean: Create/Update/ Delete-Create checks are complicated so I thought it is better to have the logic here.

cmd/ci-secret-bootstrap/main.go
droslean: The whole thing got more complicated so I thought its better to refactor it

pkg/util/secret.go
droslean: updated

cmd/ci-secret-bootstrap/main_test.go
droslean: updated.

cmd/ci-secret-bootstrap/main.go
droslean: updated this way.

alvaroaleman (approved):

PR #1614, opened 2021-01-18
Multiple secrets: causes spec.volumes[].name duplication

jupierce:
Example:
could not run steps: step launch failed: failed to create or restart test pod: unable to create pod: Pod "launch" is invalid: spec.volumes[1].name: Duplicate value: "test-secret"

### DIFF
--- a/pkg/steps/pod.go
+++ b/pkg/steps/pod.go
@@ -20,9 +20,9 @@ import (
 )
 
 const (
-	testSecretName        = "test-secret"
-	testSecretDefaultPath = "/usr/test-secrets"
-	homeVolumeName        = "home"
+	testSecretVolumePrefix = "test-secret"
+	testSecretDefaultPath  = "/usr/test-secrets"
+	homeVolumeName         = "home"
 
 	openshiftCIEnv = "OPENSHIFT_CI"
 )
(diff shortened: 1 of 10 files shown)


alvaroaleman:
@jupierce mind adding a test for this?

pkg/steps/pod.go
```diff
@@ -240,7 +240,7 @@ func (s *podStep) generatePodForStep(image string, containerResources coreapi.Re
 	pod.Spec.ServiceAccountName = s.config.ServiceAccountName
 	container := &pod.Spec.Containers[0]
 	for _, secret := range s.config.Secrets {
```
alvaroaleman, line 242: Is it safe to assume that the secret name here will be a valid kube secret name? Otherwise this might break a lot of things

jupierce: ack. Will add testcase. Naming works for our dataset, but I'll tighten the logic to work with any valid k8s.
/hold

alvaroaleman:
Going to uncc, as this creates a lot of notifications, feel free to ping me once its ready for review
/uncc

jupierce:
@alvaroaleman - cleaned up now.
- Removed reliance on secret name being a valid mount name.
- Out of an abundance of caution, the existing mount name and mount location are preserved for the first secret in a config. Only additional secrets will use the new convention.
- Tests have been added.

alvaroaleman (approved):

PR #1620, opened 2021-01-20
ci-operator: releases: use artifact dir env

stevekuznetsov:
When we create artifacts in these steps, we should honor the environment
variable that tells us where to put them instead of guessing as to where
the literal artifact dir may end up being.

/assign @alvaroaleman @smarterclayton


stevekuznetsov:
Part of DPTP-1819

alvaroaleman (approved):

stevekuznetsov:
e2e looks broken :(

PR #1621, opened 2021-01-20
ci-operator: release: use a configmap for data transfer

stevekuznetsov:
In order to not rely on the artifact gathering code, we can take
advantage of the fact that we are running as the `ci-operator` service
account in the release `Pod` and we have access to the `oc` CLI by putting
the data we need into a `ConfigMap` and reading it out of that.

Part of [DPTP-1819](https://issues.redhat.com/browse/DPTP-1819)
Depends on https://github.com/openshift/ci-tools/pull/1620


petr-muller (approved):

stevekuznetsov:
```
         Error from server (Forbidden): configmaps is forbidden: User "system:serviceaccount:ci-op-31c3jqjq:ci-operator" cannot create resource "configmaps" in API group "" in the namespace "ci-op-31c3jqjq" 
```

wat

PR #1601, opened 2021-01-13
dptp-cm: add controller image_pusher

hongkailiu:
https://issues.redhat.com/browse/DPTP-1856

PR #1622, opened 2021-01-20
Use the same label for CRs created by ci-tools

hongkailiu:
Follow up https://github.com/openshift/ci-tools/pull/1601#discussion_r557520722


petr-muller (approved):
Quick `git grep` shows no obvious consumers so this looks fine.

PR #1632, opened 2021-01-21
validation: Support devices.kubevirt.io/kvm

cgwalters:
We are going to use this for CoreOS builds/testing.
xref https://github.com/openshift/os/issues/417


cgwalters:
Got some feedback over chat:

> err at the very end this needs to translate into a pod spec change
> Allowing it in the validation and then not using it isn't really enough to solve the problem

So I need to figure out where the pod spec is being built (but it seems to me it should really be pulling in the resources as is?)

alvaroaleman (approved):

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-1697

PR #1641, opened 2021-01-26
prowgen: explicitly opt out of pod-utils with artifact dir

stevekuznetsov:
There's one major difference between the current artifact gathering code
and using the Prow pod utilities: the former allows for users to
configure their artifact directory, whereas the latter expects to be
declarative itself and for the user to simply consume the directory as
an environment variable. As we work to make the latter approach default,
we first need to opt-out if we notice a user is setting their directory
to something non-standard.

/assign @alvaroaleman

### DIFF
--- a/pkg/api/config.go
+++ b/pkg/api/config.go
@@ -39,13 +39,13 @@ func (config *ReleaseBuildConfiguration) Default() {
 			}
 		}
 	}
-	for _, step := range config.RawSteps {
-		if test := step.TestStepConfiguration; test != nil {
+	for i := range config.RawSteps {
+		if test := config.RawSteps[i].TestStepConfiguration; test != nil {
 			defTest(test)
 		}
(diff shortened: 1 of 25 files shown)


stevekuznetsov:
/hold

wtf

alvaroaleman (approved):

stevekuznetsov:
https://github.com/openshift/release/compare/master...stevekuznetsov:skuznets/opt-out-pod-utils?expand=1

stevekuznetsov:
https://github.com/openshift/release/pull/15274

stevekuznetsov:
ugh, shit - fixing the defaulting code means we need to now write it everywhere. sigh

alvaroaleman (approved):

stevekuznetsov:
part of DPTP-1819

PR #1643, opened 2021-01-26
ci-operator: gather artifacts only when needed

stevekuznetsov:
When the pod utils are being used to push artifacts up, we should not
bother grabbing them manually, especially since we do not know from
where to do so.

/assign @alvaroaleman @bbguimaraes


alvaroaleman (approved):

stevekuznetsov:
part of DPTP-1819

PR #1647, opened 2021-01-27
config/load: do something with the VERSION file

stevekuznetsov:

/assign @bbguimaraes

### DIFF
--- a/pkg/load/load.go
+++ b/pkg/load/load.go
@@ -17,6 +17,7 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/test-infra/prow/config"
 	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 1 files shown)


petr-muller (approved):

pkg/load/load.go
```diff
@@ -363,6 +364,10 @@ func Registry(root string, flat bool) (registry.ReferenceByName, registry.ChainB
 				observers[observer.Observer.Name] = observer.Observer.Observer
 			} else if strings.HasSuffix(path, CommandsSuffix) {
 				// ignore
+			} else if filepath.Base(path) == config.ConfigVersionFileName {
+				if version, err := ioutil.ReadFile(path); err != nil {
+					logrus.WithField("version", string(version)).Info("Resolved configuration version")
```
bbguimaraes, line 369: I think you're missing one of the branches?  This should be the `else` branch.

bbguimaraes (changes requested):

pkg/load/load.go
stevekuznetsov: :trollface:

alvaroaleman (approved):

PR #1648, opened 2021-01-27
backport-verifier: a new Prow plugin for forks

stevekuznetsov:

/assign @alvaroaleman 
implements https://issues.redhat.com/browse/DPTP-1842
/cc @sttts

### DIFF
--- a/cmd/backport-verifier/main.go
+++ b/cmd/backport-verifier/main.go
@@ -1,0 +1,186 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/ghodss/yaml"
+	"github.com/sirupsen/logrus"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	prowconfig "k8s.io/test-infra/prow/config"
+	"k8s.io/test-infra/prow/config/secret"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/githubeventserver"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/logrusutil"
+	"k8s.io/test-infra/prow/pjutil"
+)
+
+// Config maps upstreams to downstreams for verification
+type Config struct {
+	// Repositories is a mapping of downstream org/repo to upstream org/repo
+	Repositories map[string]string `json:"repositories,omitempty"`
+}
+
+func (c *Config) validate() error {
+	var errs []error
+	for downstreamRepo, upstreamRepo := range c.Repositories {
+		if len(strings.Split(downstreamRepo, "/")) != 2 {
+			return fmt.Errorf("%s should be in org/repo format", downstreamRepo)
+		}
+
+		if len(strings.Split(upstreamRepo, "/")) != 2 {
+			return fmt.Errorf("%s should be in org/repo format", upstreamRepo)
+		}
+	}
+
+	return utilerrors.NewAggregate(errs)
+}
+
+type options struct {
+	mut *sync.RWMutex
+
+	configPath        string
+	webhookSecretFile string
+
+	config *Config
+
+	githubEventServerOptions githubeventserver.Options
+	github                   prowflagutil.GitHubOptions
+
+	dryRun bool
+}
+
+func gatherOptions() options {
+	o := options{
+		mut: &sync.RWMutex{},
+	}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configPath, "config-path", "", "Path to backport verifier configuration.")
+	fs.StringVar(&o.webhookSecretFile, "hmac-secret-file", "", "Path to the file containing the GitHub HMAC secret.")
+
+	o.github.AddFlags(fs)
+	o.githubEventServerOptions.Bind(fs)
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatalf("cannot parse args: '%s'", os.Args[1:])
+	}
+	return o
+}
+
+func (o *options) Validate() error {
+	if err := o.github.Validate(o.dryRun); err != nil {
+		return err
+	}
+
+	bytes, err := ioutil.ReadFile(o.configPath)
+	if err != nil {
+		return fmt.Errorf("couldn't read configuration file: %v", o.configPath)
+	}
+
+	var config Config
+	if err := yaml.Unmarshal(bytes, &config); err != nil {
+		return fmt.Errorf("couldn't unmarshal configuration: %w", err)
+	}
+	o.config = &config
+
+	if err := o.config.validate(); err != nil {
+		return err
+	}
+
+	if err := o.githubEventServerOptions.DefaultAndValidate(); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (o *options) getConfigWatchAndUpdate() (func(ctx context.Context), error) {
+	errFunc := func(err error, msg string) {
+		logrus.WithError(err).Error(msg)
+	}
+
+	eventFunc := func() error {
+		bytes, err := ioutil.ReadFile(o.configPath)
+		if err != nil {
+			return fmt.Errorf("couldn't read configuration file %s: %w", o.configPath, err)
+		}
+
+		var c Config
+		if err := yaml.Unmarshal(bytes, &c); err != nil {
+			return fmt.Errorf("couldn't unmarshal configuration: %w", err)
+		}
+
+		if err := c.validate(); err != nil {
+			return err
+		}
+
+		o.mut.Lock()
+		defer o.mut.Unlock()
+		o.config = &c
+		logrus.Info("Configuration updated")
+
+		return nil
+	}
+	watcher, err := prowconfig.GetCMMountWatcher(eventFunc, errFunc, filepath.Dir(o.configPath))
+	if err != nil {
+		return nil, fmt.Errorf("couldn't get the file watcher: %w", err)
+	}
+
+	return watcher, nil
+}
+
+func main() {
+	logrusutil.ComponentInit()
+	logger := logrus.WithField("plugin", "backport-verifier")
+
+	o := gatherOptions()
+	if err := o.Validate(); err != nil {
+		logger.Fatalf("Invalid options: %v", err)
+	}
+
+	configWatchAndUpdate, err := o.getConfigWatchAndUpdate()
+	if err != nil {
+		logger.WithError(err).Fatal("couldn't get config file watch and update function")
+	}
+	interrupts.Run(configWatchAndUpdate)
+
+	secretAgent := &secret.Agent{}
+	if err := secretAgent.Start([]string{o.github.TokenPath, o.webhookSecretFile}); err != nil {
+		logger.WithError(err).Fatal("Error starting secrets agent.")
+	}
+
+	githubClient, err := o.github.GitHubClient(secretAgent, o.dryRun)
+	if err != nil {
+		logger.WithError(err).Fatal("Error getting GitHub client.")
+	}
+
+	serv := &server{
+		config: func() *Config {
+			o.mut.Lock()
+			defer o.mut.Unlock()
+			return o.config
+		},
+		ghc: githubClient,
+	}
+
+	eventServer := githubeventserver.New(o.githubEventServerOptions, secretAgent.GetTokenGenerator(o.webhookSecretFile), logger)
+	eventServer.RegisterHandleIssueCommentEvent(serv.handleIssueComment)
+	eventServer.RegisterHandlePullRequestEvent(serv.handlePullRequestEvent)
+	eventServer.RegisterHelpProvider(helpProvider, logger)
+
+	health := pjutil.NewHealth()
+	health.ServeReady()
+
+	interrupts.ListenAndServe(eventServer, time.Second*30)
+	interrupts.WaitForGracefulShutdown()
+}
--- a/cmd/backport-verifier/server.go
+++ b/cmd/backport-verifier/server.go
@@ -1,0 +1,188 @@
+package main
+
+import (
+	"fmt"
+	"regexp"
+	"sort"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	prowconfig "k8s.io/test-infra/prow/config"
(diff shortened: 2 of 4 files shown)


alvaroaleman:
@stevekuznetsov can you please provide a bit of description?

cmd/backport-verifier/main.go
```diff
@@ -0,0 +1,185 @@
...
+	dryRun bool
+}
+
+func gatherOptions() options {
+	o := options{
+		mut: &sync.RWMutex{},
+	}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.StringVar(&o.configPath, "config-path", "", "Path to publicize configuration.")
```
alvaroaleman, line 68: `path to backport-verifier config`

cmd/backport-verifier/main.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func (o *options) Validate() error {
+	if err := o.github.Validate(o.dryRun); err != nil {
+		return err
+	}
+
+	bytes, err := ioutil.ReadFile(o.configPath)
+	if err != nil {
+		return fmt.Errorf("couldn't read publicize configuration file: %v", o.configPath)
```
alvaroaleman, line 87: ditto

cmd/backport-verifier/main.go
```diff
@@ -0,0 +1,185 @@
...
+		logrus.WithError(err).Fatal("couldn't get config file watch and update function")
+	}
+	interrupts.Run(configWatchAndUpdate)
+
+	secretAgent := &secret.Agent{}
+	if err := secretAgent.Start([]string{o.github.TokenPath, o.webhookSecretFile}); err != nil {
+		logrus.WithError(err).Fatal("Error starting secrets agent.")
+	}
+
+	logger := logrus.NewEntry(logrus.New())
```
alvaroaleman, line 159: I believe this short-circuits the initialization in `ComponentInit` as that only applies to the `StandardLogger` or not?

cmd/backport-verifier/main.go
```diff
@@ -0,0 +1,185 @@
...
+	serv := &server{
+		config: func() *Config {
+			o.mut.Lock()
+			defer o.mut.Unlock()
+			return o.config
+		},
+		ghc: githubClient,
+	}
+
+	eventServer := githubeventserver.New(o.githubEventServerOptions, secretAgent.GetTokenGenerator(o.webhookSecretFile), logger)
```
alvaroaleman, line 175: Do we really need to build a server per plugin, can't we have one server for all plugins to reduce the boilerplate?

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+	RemoveLabel(org, repo string, number int, label string) error
+}
+
+const (
+	validBackportsLabel   = "backports/valid-commits"
+	invalidBackportsLabel = "backports/invalid-commits"
+)
+
+var (
+	publicizeRe    = regexp.MustCompile(`(?mi)^/validate-backports\s*$`)
```
alvaroaleman, line 33: not punlicize

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+	if err != nil {
+		if commentErr := s.ghc.CreateComment(org, repo, num, fmt.Sprintf(`@%s: could not list commits in this pull request. Please try again with /validate-backports.
+
+<details>
+
+%s
+
+</details>`, user, err)); commentErr != nil {
+			logger.WithError(commentErr).Warn("couldn't list commits")
+		}
```
alvaroaleman, line 101: I think we should `return  err` here

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+		if parts := upstreamPullRe.FindStringSubmatch(commit.Commit.Message); len(parts) > 0 {
+			pr, err := strconv.Atoi(parts[1])
+			if err != nil {
+				// based on the regex this should not happen ...
+				logger.WithError(err).Warn("Failed to parse PR as integer")
+				errorsByCommit[commit.Commit.SHA] = fmt.Sprintf("failed to parse PR identifier: %s", err.Error())
+			} else {
+				upstreamPullsByCommit[commit.Commit.SHA] = pr
+			}
+		} else {
```
alvaroaleman, line 118: Can we follow the standard golang pattern and handle the error case first, not second and use `continue` to not require the nested if-else branchching? Ref https://blog.codinghorror.com/flattening-arrow-code/

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+	upstreamPullsByCommit := map[string]int{}
+	errorsByCommit := map[string]string{}
+	for _, commit := range commits {
+		if parts := upstreamPullRe.FindStringSubmatch(commit.Commit.Message); len(parts) > 0 {
+			pr, err := strconv.Atoi(parts[1])
+			if err != nil {
+				// based on the regex this should not happen ...
+				logger.WithError(err).Warn("Failed to parse PR as integer")
+				errorsByCommit[commit.Commit.SHA] = fmt.Sprintf("failed to parse PR identifier: %s", err.Error())
+			} else {
```
alvaroaleman, line 115: Please use `continue` in the `if` branch to avoid the `else`

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+		} else {
+			invalidCommits[commit.Commit.SHA] = "does not specify an upstream backport in the commit message"
+		}
+	}
+
+	for commit, pull := range upstreamPullsByCommit {
+		pr, err := s.ghc.GetPullRequest(upstreamOrg, upstreamRepo, pull)
+		if err != nil {
+			logger.WithError(err).Warn("Failed to get upstream PR")
+			errorsByCommit[commit] = fmt.Sprintf("failed to fetch upstream PR: %s", err.Error())
```
alvaroaleman, line 127: `continue`, no `else`

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+				upstreamPullsByCommit[commit.Commit.SHA] = pr
+			}
+		} else {
+			invalidCommits[commit.Commit.SHA] = "does not specify an upstream backport in the commit message"
+		}
+	}
+
+	for commit, pull := range upstreamPullsByCommit {
+		pr, err := s.ghc.GetPullRequest(upstreamOrg, upstreamRepo, pull)
+		if err != nil {
```
alvaroaleman, line 125: I would add a dedicated handling for IsNotFound, as that is `InvalidCommit` not `ErrorCommit`

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,182 @@
...
+	validCommits := map[string]string{}
+	upstreamPullsByCommit := map[string]int{}
+	errorsByCommit := map[string]string{}
+	for _, commit := range commits {
+		if parts := upstreamPullRe.FindStringSubmatch(commit.Commit.Message); len(parts) > 0 {
+			pr, err := strconv.Atoi(parts[1])
+			if err != nil {
+				// based on the regex this should not happen ...
+				logger.WithError(err).Warn("Failed to parse PR as integer")
+				errorsByCommit[commit.Commit.SHA] = fmt.Sprintf("failed to parse PR identifier: %s", err.Error())
```
alvaroaleman, line 114: IMHO this is `invalidCommits`, not `errorsByCommit`

cmd/backport-verifier/main.go
stevekuznetsov: It should be global - but what's a better way?

cmd/backport-verifier/main.go
stevekuznetsov: Yeah, true, we could have one

stevekuznetsov: The `githubeventserver` does not support this. Whatever.

cmd/backport-verifier/server.go
stevekuznetsov: The handlers for this thing don't return errors

cmd/backport-verifier/server.go
stevekuznetsov: I honestly generally disagree with this - it's not arrow code, we just handle both cases in the if/else. I'll change it, but when you explicitly handle both cases, I find that using a `continue` or `break` and handling the other case implicitly (without `else`) is less readable.

cmd/backport-verifier/server.go
alvaroaleman: thats not the point, we need to `return`, with or without error

cmd/backport-verifier/server.go
alvaroaleman: `continue` doesn't make sense if you have another case. What you built here is not `handlig another case`, it is arrowcode for "do a bunch of checks and and abort on precondition check failure". It is the definition of what is in that blog and expressing it with two nested if else is making a very simple thing hard to understand for no good reason.

cmd/backport-verifier/server.go
stevekuznetsov: It can't happen and would be a bug in my code, an error in handling the content of the repo IMO. In any case I think it's impossible given the regex ... unless I guess they put in something larger than int64? lol

stevekuznetsov:
@alvaroaleman updated

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,188 @@
...
+	commits, err := s.ghc.ListPRCommits(org, repo, num)
+	if err != nil {
+		if commentErr := s.ghc.CreateComment(org, repo, num, fmt.Sprintf(`@%s: could not list commits in this pull request. Please try again with /validate-backports.
+
+<details>
+
+%s
+
+</details>`, user, err)); commentErr != nil {
+			logger.WithError(commentErr).Warn("couldn't list commits")
```
alvaroaleman, line 100: The point I made below and which you resolved so you don't see my response anymore still applies: We need to return if `err != nil`, not only when `commentErr != nil`, your return is a branch too high

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,188 @@
...
+		parts := upstreamPullRe.FindStringSubmatch(commit.Commit.Message)
+		if len(parts) != 2 {
+			invalidCommits[commit.Commit.SHA] = "does not specify an upstream backport in the commit message"
+			continue
+		}
+		pr, err := strconv.Atoi(parts[1])
+		if err != nil {
+			// based on the regex this should not happen ...
+			logger.WithError(err).Warn("Failed to parse PR as integer")
+			errorsByCommit[commit.Commit.SHA] = fmt.Sprintf("failed to parse PR identifier: %s", err.Error())
```
alvaroaleman, line 119: `continue` and remove the `else`

cmd/backport-verifier/server.go
```diff
@@ -0,0 +1,188 @@
...
+	var unwanted string
+	if desired == validBackportsLabel {
+		unwanted = invalidBackportsLabel
+	} else {
+		unwanted = validBackportsLabel
+	}
+	if err := client.AddLabel(org, repo, num, desired); err != nil {
+		l.WithError(err).Warn("could not add label", err)
+	}
+	if err := client.RemoveLabel(org, repo, num, unwanted); err != nil {
```
alvaroaleman, line 185: Is it safe to assume that this already exists? Shouldn't we tolerate a 404 here?

stevekuznetsov: I thought the client did internally, let me check

cmd/backport-verifier/server.go
stevekuznetsov: seriously how is that better

cmd/backport-verifier/main.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	eventFunc := func() error {
+		bytes, err := ioutil.ReadFile(o.configPath)
+		if err != nil {
+			return fmt.Errorf("couldn't read configuration file %s: %w", o.configPath, err)
+		}
+
+		var c *Config
+		if err := yaml.Unmarshal(bytes, &c); err != nil {
```
alvaroaleman, line 117: You are attempting to write into a nullpointer here

stevekuznetsov: we have this a bunch in our code (this was copy-pasta). I don't understand why it works, but it does ...

cmd/backport-verifier/server.go
stevekuznetsov: yeah it does

cmd/backport-verifier/main.go
stevekuznetsov: updated

alvaroaleman (approved):

cmd/backport-verifier/main.go
```diff
@@ -0,0 +1,184 @@
...
+	}
+
+	eventFunc := func() error {
+		bytes, err := ioutil.ReadFile(o.configPath)
+		if err != nil {
+			return fmt.Errorf("couldn't read configuration file %s: %w", o.configPath, err)
+		}
+
+		var c *Config
+		if err := yaml.Unmarshal(bytes, c); err != nil {
```
alvaroaleman, line 117: /hold

This definitely doesn't work and if we have code elsewhere that does the same then its similarly broken. It makes total sense as the `yaml.Unmrshal` can not manipulate the pointer address we store in line 116, as it gets a copy of that address.

```
$ cat main.go 
package main

import (
	"fmt"
	"github.com/ghodss/yaml"
	"log"
)

type Config struct {
	// Repositories is a mapping of downstream org/repo to upstream org/repo
	Repositories map[string]string `json:"repositories,omitempty"`
}

func main() {
	var c *Config
	fmt.Printf("%p\n", c)
	raw := []byte(`
repositories:
  hello: world`,
	)
	if err := yaml.Unmarshal(raw, c); err != nil {
		log.Fatalf("unmarshal failed: %v", err)
	}
	re, err := yaml.Marshal(c)
	if err != nil {
		log.Fatalf("Remarshal failed: %v", err)
	}
	println(string(re))
}
```

gives
```
0x0
2021/01/28 16:37:41 unmarshal failed: error unmarshaling JSON: json: Unmarshal(nil *main.Config)
exit status 1
```

alvaroaleman (approved):

alvaroaleman (approved):

alvaroaleman (approved):

sttts:
Went briefly through the logic and it looks good :+1:

Am I right that this needs a deployment follow-up somewhere it can be used? Where would I put the config for openshift/kubernetes? Is that also a prow plugin to be configured similarly to https://github.com/openshift/release/pull/15213 ?

PR #1649, opened 2021-01-28
deprovisioner: do not fail when there is nothing to cleanup

petr-muller:
`$ARTIFACTS/deprovision` did not exist when there was no cluster to clean up


petr-muller:
/hold cancel
(checked whether the rest of the script can cope with empty "$clusters")

alvaroaleman (approved):

PR #1651, opened 2021-01-28
prow-job-dispatcher: add a flag to create a PR

hongkailiu:
https://issues.redhat.com/browse/DPTP-1765

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
@@ -19,24 +20,38 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/experiment/autobumper/bumper"
 	prowconfig "k8s.io/test-infra/prow/config"
-	"k8s.io/test-infra/prow/logrusutil"
+	"k8s.io/test-infra/prow/config/secret"
 	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/ci-tools/pkg/dispatcher"
+	"github.com/openshift/ci-tools/pkg/github/prcreation"
+)
+
+const (
+	githubOrg    = "openshift"
+	githubRepo   = "release"
+	githubLogin  = "openshift-bot"
+	matchTitle   = "Automate prow job dispatcher"
+	remoteBranch = "auto-prow-job-dispatcher"
 )
 
 type options struct {
 	prowJobConfigDir string
 	configPath       string
 
-	prometheusURL          string
-	prometheusUsername     string
-	prometheusPasswordPath string
-	prometheusDaysBefore   int
-	maxConcurrency         int
+	maxConcurrency       int
+	prometheusDaysBefore int
 
-	prometheusPassword string
+	createPR    bool
+	githubLogin string
+	targetDir   string
+	assign      string
+
+	bumper.GitAuthorOptions
+	dispatcher.PrometheusOptions
+	prcreation.PRCreationOptions
 }
 
 func gatherOptions() options {
@@ -45,13 +60,19 @@ func gatherOptions() options {
 
 	fs.StringVar(&o.prowJobConfigDir, "prow-jobs-dir", "", "Path to a root of directory structure with Prow job config files (ci-operator/jobs in openshift/release)")
 	fs.StringVar(&o.configPath, "config-path", "", "Path to the config file (core-services/sanitize-prow-jobs/_config.yaml in openshift/release)")
-
-	fs.StringVar(&o.prometheusURL, "prometheus-url", "https://prometheus-prow-monitoring.apps.ci.l2s4.p1.openshiftapps.com", "The prometheus URL")
-	fs.StringVar(&o.prometheusUsername, "prometheus-username", "", "The Prometheus username.")
-	fs.StringVar(&o.prometheusPasswordPath, "prometheus-password-path", "", "The path to a file containing the Prometheus password")
 	fs.IntVar(&o.prometheusDaysBefore, "prometheus-days-before", 1, "Number [1,15] of days before. Time 00-00-00 of that day will be used as time to query Prometheus. E.g., 1 means 00-00-00 of yesterday.")
 	fs.IntVar(&o.maxConcurrency, "concurrency", 0, "Maximum number of concurrent in-flight goroutines to handle files.")
 
+	fs.BoolVar(&o.createPR, "create-pr", false, "Create a pull request to the change made with this tool.")
+	fs.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
+	fs.StringVar(&o.targetDir, "target-dir", "", "The directory containing the target repo.")
+	fs.StringVar(&o.assign, "assign", "ghost", "The github username or group name to assign the created pull request to.")
+
+	o.GitAuthorOptions.AddFlags(fs)
+	o.PrometheusOptions.AddFlags(fs)
+	o.PRCreationOptions.AddFlags(fs)
+
+	o.AllowAnonymous = true
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("could not parse input")
 	}
@@ -59,6 +80,9 @@ func gatherOptions() options {
 }
 
 func (o *options) validate() error {
+	if o.maxConcurrency == 0 {
+		o.maxConcurrency = runtime.GOMAXPROCS(0)
+	}
 	if o.prowJobConfigDir == "" {
 		return fmt.Errorf("mandatory argument --prow-jobs-dir wasn't set")
 	}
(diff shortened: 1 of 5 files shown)


cmd/autoprowjobdispatcher/main.go
```diff
@@ -0,0 +1,179 @@
...
+
+	prometheusPasswordPath string
+
+	flagutil.GitHubOptions
+}
+
+func parseOptions() options {
+	var o options
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
```
alvaroaleman, line 46: Instead of having more and more periodic jobs where the result of one might impact the result of others, can we just make this part of the `autoconfigbrancher` and remove all the pr-creation functionality?

hongkailiu: 2 excuses that I did not add this into `auto-config-brancher` process:

- In the card, we plan to run this tool weekly. However, the brancher job is hourly. I do not think it is a big deal (even it is hourly) since the current implementation respects the previous dispatching results. So the next is the real reason.
-  I would start with `self-approve=false` to build up my confidence that the process does the right thing. (I have run the tool locally several times but still I want to be careful).

alvaroaleman: Ok, the different cadence is a good point. I think you can canary out the thing locally, no need to use `self-approve=false`

cmd/autoprowjobdispatcher/main.go
```diff
@@ -0,0 +1,179 @@
...
+			commitsCounter++
+		}
+	}
+	if commitsCounter == 0 {
+		logrus.Info("no new commits, existing ...")
+		os.Exit(0)
+	}
+
+	title := fmt.Sprintf("%s by auto-prow-job-dispatcher job at %s", matchTitle, time.Now().Format(time.RFC1123))
+	if err := bumper.GitPush(fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", o.githubLogin, string(sa.GetTokenGenerator(o.GitHubOptions.TokenPath)()), o.githubLogin, githubRepo), remoteBranch, stdout, stderr, ""); err != nil {
```
alvaroaleman, line 167: We have `pkg/github/prcreation/` for the whole pr creation logic, any reason not to use that?

cmd/autoprowjobdispatcher/main.go
```diff
@@ -0,0 +1,179 @@
...
+		logrus.WithError(err).Fatal("Failed to push changes.")
+	}
+
+	var labelsToAdd []string
+	if o.selfApprove {
```
alvaroaleman, line 172: We automatically merge everything that comes from the ci-robot, so adding the labels or not doesn't make a difference anymore, we will automerge everything

cmd/autoprowjobdispatcher/main.go
```diff
@@ -0,0 +1,179 @@
...
+	"k8s.io/test-infra/prow/labels"
+)
+
+const (
+	githubOrg    = "openshift"
+	githubRepo   = "release"
+	githubLogin  = "openshift-bot"
+	githubTeam   = "openshift/openshift-team-developer-productivity-test-platform"
+	matchTitle   = "Automate config brancher"
+	remoteBranch = "auto-config-brancher"
```
alvaroaleman, line 26: We should probably use our own branch

cmd/autoprowjobdispatcher/main.go
```diff
@@ -0,0 +1,179 @@
...
+		logrus.WithError(err).Fatal("Failed to change to root dir")
+	}
+
+	steps := []struct {
+		command   string
+		arguments []string
+	}{
+		{
+			command:   "/usr/bin/prow-job-dispatcher",
+			arguments: []string{"--prow-jobs-dir", "./ci-operator/jobs", "--prometheus-password-path", o.prometheusPasswordPath, "--prometheus-username", "ci", "--config-path", "core-services/sanitize-prow-jobs/_config.yaml"},
```
alvaroaleman, line 138: The prometheus address is hardcoded into the tool?

cmd/autoprowjobdispatcher/main.go
hongkailiu: Updated.

cmd/autoprowjobdispatcher/main.go
hongkailiu: I will add hold label as discussed over slack.

cmd/autoprowjobdispatcher/main.go
hongkailiu: Updated.

cmd/autoprowjobdispatcher/main.go
hongkailiu: Make `dispatcher.PrometheusOptions` to share between those 2 tools.

cmd/prow-job-dispatcher/main.go
```diff
@@ -370,4 +424,61 @@ func main() {
 	if err := dispatcher.SaveConfig(config, o.configPath); err != nil {
 		logrus.WithError(err).Fatalf("Failed to save config file to %s", o.configPath)
 	}
+
+	if !o.createPR {
+		logrus.Info("Finished dispatching and create no PR, exiting ...")
+		os.Exit(0)
+	}
+
+	logrus.Infof("Changing working directory to '%s' ...", o.targetDir)
```
alvaroaleman, line 433: Please do not `printf` logging, use fields

cmd/prow-job-dispatcher/main.go
```diff
@@ -370,4 +424,61 @@ func main() {
...
+		logrus.WithError(err).Fatal("Failed to change to root dir")
+	}
+
+	steps := []struct {
+		gitOnly   bool
+		command   string
+		arguments []string
+	}{
+		{
+			gitOnly:   true,
```
alvaroaleman, line 444: So `gitOnly` means you define a command and arguments and then never call it? That is confusing at best. I'd rather have it you make a `commandWithArgs ...string` variadic arg to `runAndCommitIfNeeded` and do not call anything (aka what is called "git only" today) `if len(commandWithArgs == 0)`

cmd/prow-job-dispatcher/main.go
```diff
@@ -370,4 +424,61 @@ func main() {
...
+		os.Exit(0)
+	}
+
+	if err := bumper.GitPush(fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", o.githubLogin, string(sa.GetTokenGenerator(o.PRCreationOptions.GitHubOptions.TokenPath)()), o.githubLogin, githubRepo), remoteBranch, stdout, stderr, ""); err != nil {
+		logrus.WithError(err).Fatal("Failed to push changes.")
+	}
+
+	title := fmt.Sprintf("%s at %s", matchTitle, time.Now().Format(time.RFC1123))
+	// TODO (hongkliu): Remove this if it passes the manual test
+	labelsToAdd := []string{labels.Hold}
```
alvaroaleman, line 480: Please do not add code for this. Just test it. Our presubmits need at the very least 5 minutes to finish their testing, that is plenty of time for you to add a `hold` manually

cmd/prow-job-dispatcher/main.go
```diff
@@ -370,4 +424,61 @@ func main() {
...
+		if committed {
+			commitsCounter++
+		}
+	}
+	if commitsCounter == 0 {
+		logrus.Info("no new commits, existing ...")
+		os.Exit(0)
+	}
+
+	if err := bumper.GitPush(fmt.Sprintf("https://%s:%s@github.com/%s/%s.git", o.githubLogin, string(sa.GetTokenGenerator(o.PRCreationOptions.GitHubOptions.TokenPath)()), o.githubLogin, githubRepo), remoteBranch, stdout, stderr, ""); err != nil {
```
alvaroaleman, line 474: Would you mind using one line per argument? This is extremely confusing to read right now

cmd/prow-job-dispatcher/main.go
```diff
@@ -45,52 +62,58 @@ func gatherOptions() options {
...
-	fs.StringVar(&o.prometheusURL, "prometheus-url", "https://prometheus-prow-monitoring.apps.ci.l2s4.p1.openshiftapps.com", "The prometheus URL")
-	fs.StringVar(&o.prometheusUsername, "prometheus-username", "", "The Prometheus username.")
-	fs.StringVar(&o.prometheusPasswordPath, "prometheus-password-path", "", "The path to a file containing the Prometheus password")
 	fs.IntVar(&o.prometheusDaysBefore, "prometheus-days-before", 1, "Number [1,15] of days before. Time 00-00-00 of that day will be used as time to query Prometheus. E.g., 1 means 00-00-00 of yesterday.")
 	fs.IntVar(&o.maxConcurrency, "concurrency", 0, "Maximum number of concurrent in-flight goroutines to handle files.")
 
+	fs.BoolVar(&o.createPR, "create-pr", false, "Create a pull request to the change made with this tool.")
+	fs.StringVar(&o.githubLogin, "github-login", githubLogin, "The GitHub username to use.")
+	fs.StringVar(&o.targetDir, "target-dir", "", "The directory containing the target repo.")
+	fs.StringVar(&o.assign, "assign", githubTeam, "The github username or group name to assign the created pull request to.")
```
alvaroaleman, line 71: The default is to not assign anyone, as we automerge those anyways

cmd/prow-job-dispatcher/main.go
```diff
@@ -329,20 +352,51 @@ func dispatchJobs(ctx context.Context, prowJobConfigDir string, maxConcurrency i
...
-	}))
 	o := gatherOptions()
 	if err := o.validate(); err != nil {
 		logrus.WithError(err).Fatal("Failed to complete options.")
 	}
-	if err := o.complete(&secrets); err != nil {
-		logrus.WithError(err).Fatal("Failed to complete options.")
+
+	sa := &secret.Agent{}
+	if err := sa.Start([]string{o.PRCreationOptions.GitHubOptions.TokenPath, o.PrometheusOptions.PrometheusPasswordPath}); err != nil {
```
alvaroaleman, line 395: I don't think the `TokenPath` should always be mandatory, only `if o.CratePR`

cmd/prow-job-dispatcher/main.go
hongkailiu: Updated.

cmd/prow-job-dispatcher/main.go
hongkailiu: Removed.

cmd/prow-job-dispatcher/main.go
hongkailiu: Updated.

cmd/prow-job-dispatcher/main.go
hongkailiu: default to `ghost`?

cmd/prow-job-dispatcher/main.go
hongkailiu: Updated.

cmd/prow-job-dispatcher/main.go
```diff
@@ -370,4 +430,67 @@ func main() {
...
+		logrus.WithError(err).Fatal("Failed to change to root dir")
+	}
+
+	steps := []struct {
+		command   string
+		arguments []string
+	}{
+		{
+			// start with git commands without running any tools
+			command: "",
```
alvaroaleman, line 450: This is still super confusing, removing the `run` part out of the `runAndCommitIfNeeded` function and not trying to use a loop here would made everything more obvious

alvaroaleman (approved):

cmd/prow-job-dispatcher/main.go
hongkailiu: That was one of my commits and I felt using the loop is simpler.
I will follow up tomorrow.

PR #1653, opened 2021-01-29
backport-verifier: only comment about missing config when explicit

stevekuznetsov:
We want to be able to turn on this plugin for the entire org, so we
can't be spamming a comment for unconfigured repositories unless the
user asked us to do a backport verification explicitly.

/assign @alvaroaleman


stevekuznetsov:
part of Part of DPTP-1842

alvaroaleman (approved):

PR #1657, opened 2021-02-02
pkg/registry: merge test/workflow list/dict fields

bbguimaraes:
While technically correct according to the [specification][0],
completely overriding the workflow fields is sometimes not what is
desired when fields are added to the test in the `ci-operator`
configuration file.  [DPTP-1824][1] documents one of these cases: when
an `env` section is added to a test, the intent is likely to augment the
workflow section, but it is instead completely overridden.

The behavior of configuration resolution is changed so that the two
sections are now merged for list and dictionary fields, namely:

- `env`/`TestEnvironment`: parameters in the test override those in the
  workflow if they target the same environment variable.  Otherwise,
  the resulting parameter list is the combination of both sections.  An
  existing value can be cleared by setting `default` to the empty
  string.  It is not possible to revert the effects of the workflow
  section (i.e. use the original default value), but that is unlikely to
  be useful.
- `dependencies`/`TestDependencies`: same treatment as the previous item.
- `leases`/`[]StepLease`: resources in the test must not target an
  environment variable already present in the workflow.  Otherwise, the
  resulting lease list is the combination of both sections.

[0]: https://docs.ci.openshift.org/docs/architecture/step-registry/#ci-operator-test-configuration
[1]: https://issues.redhat.com/browse/DPTP-1824

TODO:

- [x] verify existing configuration files remain valid
- [x] verify existing tests are not negatively affected
- [x] update documentation

### DIFF
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -80,20 +80,12 @@ func (r *registry) Resolve(name string, config api.MultiStageTestConfiguration) 
 		if config.Post == nil {
 			config.Post = workflow.Post
 		}
-		if config.Environment == nil {
-			config.Environment = make(api.TestEnvironment, len(workflow.Environment))
-			for k, v := range workflow.Environment {
-				config.Environment[k] = v
-			}
-		}
-		if config.Dependencies == nil {
-			config.Dependencies = make(api.TestDependencies, len(workflow.Dependencies))
-			for k, v := range workflow.Dependencies {
-				config.Dependencies[k] = v
-			}
-		}
-		if config.Leases == nil {
-			config.Leases = append(config.Leases, workflow.Leases...)
+		mergeEnvironments(&config.Environment, workflow.Environment)
+		mergeDependencies(&config.Dependencies, workflow.Dependencies)
+		if l, err := mergeLeases(workflow.Leases, config.Leases); err != nil {
+			resolveErrors = append(resolveErrors, err)
+		} else {
+			config.Leases = l
 		}
 		if config.AllowSkipOnSuccess == nil {
 			config.AllowSkipOnSuccess = workflow.AllowSkipOnSuccess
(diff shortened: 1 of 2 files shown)


pkg/registry/resolver.go
```diff
@@ -147,6 +139,61 @@ func (r *registry) Resolve(name string, config api.MultiStageTestConfiguration)
...
+func mergeDependencies(dst *api.TestDependencies, src api.TestDependencies) {
+	mergeMaps((*map[string]string)(dst), (map[string]string)(src))
+}
+
+func mergeMaps(dst *map[string]string, src map[string]string) {
+	if src == nil {
+		return
+	}
+	if *dst == nil {
+		*dst = make(api.TestEnvironment, len(src))
```
stevekuznetsov, line 161: Why does this assume a type for `dst`?

bbguimaraes: Uh… thanks for catching it, but how is this legal?  Updated.

bbguimaraes: OK, so only the pointer types are incompatible.  Updated again to remove the casts in the other `merge*` functions.

bbguimaraes:
The only differences in resolution for all configuration files currently
(`53d533f070`) in `openshift/release` are:

- `openshift/cluster-version-operator` tests use the upgrade workflows and
  override `TEST_UPGRADE_OPTIONS`.  This causes the `TEST_SUITE` parameter of
  the workflow to also be overridden, reverting it to the default used in the
  step.  This is harmless as the value is ignored for upgrade tests.  If
  necessary, the variable could be set to another value in the test.
- The `e2e-44-stable-to-45-ci` test in the `ocp-4.5-ci` variant of
  `openshift/release` has an `env` section which unintentionally reverts the
  `TEST_TYPE` from `upgrade` to `suite`.  This is a bug in the job definition
  and will be corrected with the new resolution algorithm.

bbguimaraes:
https://github.com/openshift/ci-docs/pull/89

PR #1666, opened 2021-02-05
cvp-trigger: supply parameters to namespace hash

stevekuznetsov:

/assign @petr-muller 
/cc @dirgim

### DIFF
--- a/cmd/cvp-trigger/main.go
+++ b/cmd/cvp-trigger/main.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/ghodss/yaml"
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Solves DPTP-1903

cmd/cvp-trigger/main.go
```diff
@@ -223,6 +223,13 @@ func main() {
 	if o.targetNamespaces != "" {
 		envVars[steps.OOTargetNamespaces] = o.targetNamespaces
 	}
+	input := strings.Builder{}
+	input.WriteString("--input-hash=")
+	for key, value := range envVars {
+		input.WriteString(key)
+		input.WriteString(value)
+	}
```
petr-muller, line 231: /paranoid

This will still fail if CVP submits two identical jobs (same bundle, same base OCP...) in the same time window.  @dirgim is that possible?

petr-muller (commented):
/hold

cmd/cvp-trigger/main.go
dirgim: Submitting two identical jobs in the same time window should not be possible in general, at the very least the index image references should be different.
The only exception I can think of would be if the cvp-trigger loses connection to the Prow cluster and CVP retries re-running the job right after it.

petr-muller: /hold cancel

petr-muller: /paranoid cancel

petr-muller: /lgtm

petr-muller: /hold

oh wait, @stevekuznetsov this needs sorting to not flake tests

stevekuznetsov:
/test images

```
 error: error creating buildah builder: Error parsing image configuration: Error fetching blob: invalid status code from registry 503 (Service Unavailable) 
```

sigh

PR #1672, opened 2021-02-08
Use ocp/4.8:cli image for promotion pod

hongkailiu:


petr-muller (approved):
Leading edge :metal:

PR #1673, opened 2021-02-08
Remove code supporting obsolete templates

petr-muller:
There are no configs (even YAML-inline) that use these stanzas, so we
may remove all support for them in our tools:

- openshift_ansible_40
- openshift_installer_src

/cc @stevekuznetsov @AlexNPavel


petr-muller:
@AlexNPavel @stevekuznetsov ping

stevekuznetsov (approved):

PR #1679, opened 2021-02-09
Allow steps to create roles

jupierce:
Allow step serviceaccount to create roles. This will enable a step to,
for example, expose a resource within its namespace to another
serviceaccount.

The first use case is to enable a registry step which allows cluster-bot
to monitor a secrets within a launched cluster namespace and
extract the kubeconfig when a cluster is available. This kubeconfig
is shared back to the slack user who requested the cluster.

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -230,7 +230,7 @@ func (s *multiStageTestStep) setupRBAC(ctx context.Context) error {
 		ObjectMeta: m,
 		Rules: []rbacapi.PolicyRule{{
 			APIGroups: []string{"rbac.authorization.k8s.io"},
-			Resources: []string{"rolebindings"},
+			Resources: []string{"rolebindings", "roles"},
 			Verbs:     []string{"create", "list"},
 		}, {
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
Not opposed to this, but you could instead just use a rolebinding with one of the predefined clusterroles like admin

jupierce:
Thanks @alvaroaleman . I may not have crafted the step correctly, but the serviceaccount my steps run with isn't admin in the project so I could not bind with admin.
```
Error from server (Forbidden): rolebindings.rbac.authorization.k8s.io "ci-chat-bot-secret-reader-binding" is forbidden: user "system:serviceaccount:ci-op-r39jkw1s:launch" (groups=["system:serviceaccounts" "system:serviceaccounts:ci-op-r39jkw1s" "system:authenticated"]) is attempting to grant RBAC permissions not currently held:
```

https://github.com/openshift/ci-tools/pull/1679/files#diff-da42b74c690675134a34fba17070b2705bdf546b16b6e84e06ef9440f637a166R231-R244 is the comprehensive set of rbac for the serviceaccount created in my ci namespace.

PR #1682, opened 2021-02-10
ci-operator: set TTL annotations before everything else

petr-muller:
ci-operator was setting up SSAR and pull secrets between the namespace
creation and setting up the annotations, so if one of them failed, the
annotations were not set and we leak the namespace.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -857,32 +857,12 @@ func (o *options) initializeNamespace() error {
 		return errors.New("timed out waiting for RBAC")
 	}
 
-	pullStart := time.Now()
-	var imagePullSecretsMinted bool
-	for i := 0; i < 30; i++ {
-		imagePullSecretsMinted = true
-		serviceAccounts := map[string]*coreapi.ServiceAccount{
-			"builder": {},
-			"default": {},
-		}
-		for name := range serviceAccounts {
-			if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: o.namespace, Name: name}, serviceAccounts[name]); err != nil && !kerrors.IsNotFound(err) {
-				return fmt.Errorf("failed to fetch service account %s: %w", name, err)
-			}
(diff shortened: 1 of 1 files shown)


cmd/ci-operator/main.go
```diff
@@ -943,6 +892,57 @@ func (o *options) initializeNamespace() error {
 		}
 	}
 
+	ssarStart := time.Now()
```
alvaroaleman, line 895: Can we not simply set the annotation on creation or does the projects api not allow that?

petr-muller: Possibly, haven't tried. It's not entirely braindead because the annotations are technically optional and you may still need to update afterwards because you are reusing an existing project but it would be a safest thing to do.

I'm not too keen on doing it myself though - reordering stuff seemed easy enough to just do it, but I'd really like to avoid another rabbit hole :)

alvaroaleman: I guess the point is, this might just fail because the SSAR succeeded and is generally still prone to racing

alvaroaleman: The ProjectRequest takes an ObjectMeta, lets just add the annotation there and then do the update after SSAR. We still have to do the update in case we didn't create the project

petr-muller: OK, will fixup the PR, stay tuned.

smarterclayton: You can't set annotations on project (historical limitation).  So if you send them they'll be ignored I think (but if you see in your test they don't let me know)

stevekuznetsov: Let's add a comment, when I wrote this it was common knowledge

stevekuznetsov:
/hold

wait

We can't do *anything* before the SSAR check goes through, as that's when we actually have RBAC in the namespace to do anything in the first place. Please add the annotation logic after this check.

stevekuznetsov (approved):
/hold cancel

petr-muller:
/hold cancel

Addressed. Moved the annotation setup after SSAR but before pull secret setup. Added a comment explaining the quirks involved.

I confirmed that adding annotations/labels does not work in the initial `Create` API call.

PR #1676, opened 2021-02-09
prowgen: generate flag --lease-server-credentials-file

hongkailiu:
https://issues.redhat.com/browse/DPTP-1913

/hold

require

https://github.com/openshift/ci-tools/pull/1675

PR #1683, opened 2021-02-10
ci-operator: support credentials file for boskos

hongkailiu:
Follow up https://github.com/openshift/ci-tools/pull/1676#issuecomment-776017085

/cc @petr-muller @alvaroaleman

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -44,6 +44,7 @@ import (
 	"k8s.io/client-go/util/retry"
 	"k8s.io/klog/v2"
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/config/secret"
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 	"k8s.io/test-infra/prow/version"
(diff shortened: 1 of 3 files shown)


cmd/ci-operator/main.go
```diff
@@ -1385,18 +1386,41 @@ func (o *options) saveNamespaceArtifacts() {
...
+		username := splits[0]
+		passwordGetter := func() []byte {
+			splits := strings.Split(string(sa.GetSecret(leaseServerCredentialsFile)), ":")
+			if len(splits) != 2 {
+				log.Printf("got invalid content of lease server credentials file which must be of the form '<username>:<passwrod>'")
+				return nil
+			}
+			return []byte(splits[1])
+		}
+		return username, passwordGetter, nil
```
petr-muller, line 1408: Hm what is fishy here is that the password getter is IMO unnecessarily dynamic - if the underlying file can change (which it IMO not expected for ci-operator), then the username can change too. So I'd just read the file once here and made the password getter a simple 

```go
func() []byte {
  return []byte(splits[1])
}
```

and not pretend to allow credentials reloading when it really doesn't.

hongkailiu: Updated.

petr-muller (approved):

PR #1685, opened 2021-02-12
release: use app.ci URLs for release-controllers

petr-muller:
This was also a last usage of `svc.ci.openshift.org` except for registry URLs.


petr-muller:
So why does asking the *correct* RC result in imagepullbackoffs FFS

stevekuznetsov:
I mean we pulled the `release` image from the registry, not sure why the others would not also pull ...

stevekuznetsov:
As expected, flake :)

PR #1689, opened 2021-02-16
result-aggregator: support --passwd-file

hongkailiu:
https://issues.redhat.com/browse/DPTP-1970

### DIFF
--- a/cmd/result-aggregator/main.go
+++ b/cmd/result-aggregator/main.go
@@ -45,6 +45,7 @@ type options struct {
 	gracePeriod time.Duration
 	username    string
 	password    string
+	passwdFile  string
 }
 
 func gatherOptions() (options, error) {
@@ -55,6 +56,7 @@ func gatherOptions() (options, error) {
 	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
 	fs.StringVar(&o.username, "username", "", "Username to trust for clients.")
 	fs.StringVar(&o.password, "password-file", "", "File holding the password for clients.")
+	fs.StringVar(&o.passwdFile, "passwd-file", "", "Authenticate against a file. Each line of the file is with the form `<username>:<password>`. This will override --username and --password-file")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		return o, fmt.Errorf("failed to parse flags: %w", err)
 	}
@@ -66,11 +68,11 @@ func validateOptions(o options) error {
 	if err != nil {
 		return fmt.Errorf("invalid --log-level: %w", err)
 	}
-	if o.username == "" {
-		return errors.New("--username is required")
+	if (o.username == "") != (o.password == "") {
+		return errors.New("--username and --password-file must be specified together")
 	}
-	if o.password == "" {
-		return errors.New("--password-file is required")
+	if (o.username == "") && (o.passwdFile == "") {
+		return errors.New("--username or --htpasswd-file must be specified")
 	}
 	return nil
 }
@@ -110,15 +112,24 @@ func withErrorRate(request *results.Request) {
 	errorRate.With(labels).Inc()
 }
 
-func handleCIOperatorResult(username string, password func() []byte) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		start := time.Now()
+type validator interface {
+	Validate(username, password string) bool
+}
+
+func loginHandler(validator validator, next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		user, pass, ok := r.BasicAuth()
-		if !ok || user != username || pass != string(password()) {
-			w.WriteHeader(http.StatusUnauthorized)
-			fmt.Fprint(w, "Unauthorized")
+		if !ok || !validator.Validate(user, pass) {
+			http.Error(w, "Unauthorized", http.StatusUnauthorized)
 			return
 		}
+		next.ServeHTTP(w, r)
+	})
+}
+
+func handleCIOperatorResult() http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		start := time.Now()
 
 		bytes, err := ioutil.ReadAll(r.Body)
 		if err != nil {
@@ -165,7 +176,13 @@ func main() {
 	}
 
 	http.HandleFunc("/", http.NotFound)
-	http.HandleFunc("/result", handleCIOperatorResult(o.username, secretAgent.GetTokenGenerator(o.password)))
+
+	validator := &multi{delegates: []validator{
+		&passwdFile{file: o.passwdFile},
+		&literal{username: o.username, password: secretAgent.GetTokenGenerator(o.password)}},
+	}
+
+	http.Handle("/result", loginHandler(validator, handleCIOperatorResult()))
 	metrics.ExposeMetrics("result-aggregator", prowConfig.PushGateway{}, flagutil.DefaultMetricsPort)
 
 	interrupts.ListenAndServe(&http.Server{Addr: o.address}, o.gracePeriod)
--- a/cmd/result-aggregator/main_test.go
+++ b/cmd/result-aggregator/main_test.go
@@ -1,0 +1,150 @@
+package main
+
+import (
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/http/httptest"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestValidator(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() {
+		if err := os.RemoveAll(dir); err != nil {
+			t.Errorf("failed to remove the temp dir: %s", dir)
+		}
+	}()
+
+	passwdFileRaw := filepath.Join(dir, "passwdFile")
+	if err := ioutil.WriteFile(passwdFileRaw, []byte(`a:b
+c:d
+:`), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	passwdFileRawAbnormalContent := filepath.Join(dir, "passwdFileRawAbnormalContent")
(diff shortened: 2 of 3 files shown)


cmd/result-aggregator/main.go
```diff
@@ -55,6 +57,7 @@ func gatherOptions() (options, error) {
 	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
 	fs.StringVar(&o.username, "username", "", "Username to trust for clients.")
 	fs.StringVar(&o.password, "password-file", "", "File holding the password for clients.")
+	fs.StringVar(&o.htpasswdFile, "htpasswd-file", "", "Authenticate against a htpasswd file. Entries must be created with \"htpasswd -p\" for no encryption.")
```
alvaroaleman, line 60: Nit: This is not encryption, this is hashing

cmd/result-aggregator/htpasswd.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/csv"
+	"fmt"
+	"io"
+	"os"
+)
+
+// lookup passwords in a htpasswd file
+// The entries must have been created with -s for SHA encryption
+
+type HtpasswdFile struct {
```
alvaroaleman, line 13: None of this needs exporting

cmd/result-aggregator/htpasswd.go
```diff
@@ -0,0 +1,53 @@
...
+	csvReader.TrimLeadingSpace = true
+
+	records, err := csvReader.ReadAll()
+	if err != nil {
+		return nil, err
+	}
+	h := &HtpasswdFile{Users: make(map[string]string)}
+
+	for _, record := range records {
+		if len(record) < 2 {
```
alvaroaleman, line 39: This will fail the result aggregator

cmd/result-aggregator/htpasswd.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	return h, nil
+}
+
+func (h *HtpasswdFile) Validate(user string, password string) bool {
+	realPassword, exists := h.Users[user]
+	if !exists {
+		return false
+	}
+	return realPassword == password
```
alvaroaleman, line 52: The password in a htpasswd file is hashed, you can not compare it like this, you need to do something like `bcrypt.CompareHashAndPassword` after figuring out the hash format in the htpasswd file, not sure how to do that

cmd/result-aggregator/main.go
```diff
@@ -110,15 +116,43 @@ func withErrorRate(request *results.Request) {
...
 }
 
-func handleCIOperatorResult(username string, password func() []byte) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		start := time.Now()
+type validator interface {
+	Validate(username, password string) bool
+}
+
+func newValidator(username string, password string, htpasswdFileRaw string) (validator, error) {
```
alvaroaleman, line 123: Would be good if the validator re-read the file to be able to pick up changes to it without a restart

cmd/result-aggregator/main_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+func TestNewValidator(t *testing.T) {
+	dir, err := ioutil.TempDir("", "test")
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer os.RemoveAll(dir)
+
+	htpasswdFileRaw := filepath.Join(dir, "htpasswdFileRaw")
+	if err := ioutil.WriteFile(htpasswdFileRaw, []byte(`a:b
```
alvaroaleman, line 27: This is not at all how such a file looks, the password is always hashed:
```
$ htpasswd -bnB  username password 
username:$2y$05$8QGlBkiCieIDYJ8AWRl0UOgkXgj.LQASoqplEKRwwCh/xVC74whTO
```

cmd/result-aggregator/main.go
hongkailiu: I know. 

I copied the code over from (removing the hashing/encription part)
https://github.com/openshift/oauth-proxy/blob/master/htpasswd.go

https://github.com/openshift/oauth-proxy/blob/4bd4705011c19a7556551c9a1dfaaa5eacc7898c/main.go#L64
```
	flagSet.String("htpasswd-file", "", "additionally authenticate against a htpasswd file. Entries must be created with \"htpasswd -s\" for SHA encryption")
```

I feel the encryption is not needed since the user/password is generated (the original way is plain-text too). So

> Entries must be created with \"htpasswd -p\" for no encryption."

```bash
$ htpasswd -bnp username password
Warning: storing passwords as plain text might just not work on this platform.
username:password
```

my bad choice again?

cmd/result-aggregator/main.go
hongkailiu: Do we have an API for this already?

cmd/result-aggregator/main.go
alvaroaleman: I guess that is fine, but then do not name this `htpasswd`, because its not

cmd/result-aggregator/main.go
alvaroaleman: an api? Just put the file reading in the verification closure

cmd/result-aggregator/main.go
hongkailiu: Will rename.

cmd/result-aggregator/main.go
hongkailiu: Hmm. Each http request will need to read the file again?
I do not think we rotate that often.
Maybe just restarting the pod is a better choice?

alvaroaleman: It will end up in the page cache if we do the request often. I don't think the performance is a big deal here. You could also create a goroutine that uses fsnotify to wach, but its not trivial, @AlexNPavel had some fun getting that to work properly for configmap/secret mounts

hongkailiu: Ack. Will update.

cmd/result-aggregator/main.go
```diff
@@ -110,15 +116,43 @@ func withErrorRate(request *results.Request) {
...
+	var htpasswdFile *HtpasswdFile
+	if htpasswdFileRaw != "" {
```
stevekuznetsov, line 125: nit: a method that takes a list of inputs and uses `if` to do something different with each is a code smell

just call different methods originally

cmd/result-aggregator/main.go
hongkailiu: Updated.

cmd/result-aggregator/htpasswd.go
hongkailiu: Ignored the record if `len(record)!=2`

cmd/result-aggregator/main.go
hongkailiu: Updated.

cmd/result-aggregator/htpasswd.go
hongkailiu: Updated.

cmd/result-aggregator/main.go
hongkailiu: Did not get rid of it completely.
Should look better after removing the u/p flag in the future.

cmd/result-aggregator/passwd.go
```diff
@@ -0,0 +1,54 @@
...
+	"io"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+)
+
+type passwdFile struct {
+	username      string
+	password      func() []byte
```
stevekuznetsov, line 15: This is not part of the file. Just have two separate validators that implement your interface. One uses a literal password, one reads the htpasswd file. Combine them with logical OR.

stevekuznetsov: Similar (but not the same) problem is solved [here](https://github.com/openshift/ci-tools/blob/aa5a6f46442730c79c61708c61b60d2c7b6f4d55/pkg/slack/events/handler.go#L80-L93)

stevekuznetsov: So I'd expect to see:

```go

type validator interface {
	Validate(username, password string) bool
}

type htpasswd struct {
	file string
}

func (h *htpasswd) Validate(username, password string) bool {
	reader, err := os.Open(h.file)
	if err != nil {
		logrus.WithField("file", h.file).Error("failed to open file")
		return false
	}
	defer func() {
		if err := reader.Close(); err != nil {
			logrus.WithError(err).Warn("Could not close htpasswd file reader.")
		}
	}()

	csvReader := csv.NewReader(reader)
	csvReader.Comma = ':'
	csvReader.Comment = '#'
	csvReader.TrimLeadingSpace = true

	records, err := csvReader.ReadAll()
	if err != nil {
		logrus.Error("failed to read csv")
	}

	for _, record := range records {
		if len(record) == 2 && record[0] == username && record[1] == password {
			return true
		}
	}
	return false
}

type literal struct {
	username, password string
}

func (l *literal) Validate(username, password string) bool {
	return username == l.username && password == l.password
}

type multi struct {
	delegates []validator
}

func (m *multi) Validate(username, password string) bool {
	for _, delegate := range m.delegates {
		if delegate.Validate(username, password) {
			return true
		}
	}
	return false
}
```

NOTE: handle errors for `Close()` in your `defer`. Change your editor to catch this.

hongkailiu: I see. Updated.

> NOTE: handle errors for Close() in your defer. Change your editor to catch this.

It is enabled by default in GoLand. And my eyes did not catch that out.

cmd/result-aggregator/main.go
```diff
@@ -165,7 +176,13 @@ func main() {
 	}
 
 	http.HandleFunc("/", http.NotFound)
-	http.HandleFunc("/result", handleCIOperatorResult(o.username, secretAgent.GetTokenGenerator(o.password)))
+
+	validator := &multi{delegates: []validator{
+		&passwdFile{file: o.passwdFile},
+		&literal{username: o.username, password: secretAgent.GetTokenGenerator(o.password)}},
```
stevekuznetsov, line 182: if `o.username` and `o.password` are empty this will validate logins with `"", ""`

hongkailiu: Updated.

stevekuznetsov (approved):

PR #1691, opened 2021-02-17
secret-wrapper: handle writable home

stevekuznetsov:
There are a number of changes we make to the environment of the process
under test. Some occur in the `secret-wrapper` tool, while some form an
ugly Bash preable. As the content of these changes continues to grow and
become more complex, we should aim to make all of them in the Go tool.

/assign @bbguimaraes 
/cc @alvaroaleman 

This is backwards  compatible and will work with *both* `secret-wrapper` and the shell. I'll remove the shell later.

### DIFF
--- a/cmd/secret-wrapper/main.go
+++ b/cmd/secret-wrapper/main.go
@@ -172,6 +172,7 @@ func execCmd(argv []string) error {
 	proc := exec.Command(argv[0], argv[1:]...)
 	proc.Stdout = os.Stdout
 	proc.Stderr = os.Stderr
+	manageHome(proc)
 	sig := make(chan os.Signal, 1)
 	signal.Notify(sig, syscall.SIGINT, syscall.SIGTERM)
 	ctx, cancel := context.WithCancel(context.Background())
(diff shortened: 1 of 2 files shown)


cmd/secret-wrapper/main.go
```diff
@@ -172,6 +172,18 @@ func execCmd(argv []string) error {
 	proc := exec.Command(argv[0], argv[1:]...)
 	proc.Stdout = os.Stdout
 	proc.Stderr = os.Stderr
+	// provide a writeable home so kubectl discovery can be cached
```
alvaroaleman, line 175: Can we rename the tool if it now does stuff that is entirely unrelated to secret wrapping?

stevekuznetsov: Sure, can I do that after I land the useful bits? :)

alvaroaleman: I am afraid that won't happen then :P Test failure looks genuine btw

stevekuznetsov:
Needed for DPTP-1668

cmd/secret-wrapper/main.go
```diff
@@ -172,6 +172,18 @@ func execCmd(argv []string) error {
...
+	home, set := os.LookupEnv("HOME")
+	needHome := !set
+	if set {
+		if err := syscall.Access(home, syscall.O_RDWR); err != nil {
+			// $HOME is set but not writeable
+			needHome = true
+		}
+	}
+	if needHome {
+		proc.Env = append(os.Environ(), "HOME=/alabama")
```
bbguimaraes, line 185: To save time for other potential paranoid reviewers:
```
// If Env contains duplicate environment keys, only the last
// value in the slice for each duplicate key is used.
```

cmd/secret-wrapper/main.go
```diff
@@ -172,6 +172,18 @@ func execCmd(argv []string) error {
...
+	}
```
bbguimaraes, line 186: This function will get quite involved, specially with the following PR.  Can we leave only the execution bits here and move the new code to a separate one?  Adding the environment as a parameter looks like a good interface.

cmd/secret-wrapper/main.go
stevekuznetsov: Fixed in the following PR

cmd/secret-wrapper/main.go
stevekuznetsov: Sure

stevekuznetsov:
```
         * 3x kubelet: Failed to create pod sandbox: rpc error: code = DeadlineExceeded desc = context deadline exceeded 
```

/test e2e

PR #1693, opened 2021-02-17
secret-wrapper: populate a unique, writable KUBECONFIG

stevekuznetsov:
As we upload `$KUBECONFIG` as soon as it's available and some consumers
of the `$SHARED_DIR` (observers) will want to be able to consume it as
soon as it's uploaded, we need to handle the case where it does not
exist at the moment that a process starts reading the `$SHARED_DIR`, and
furthermore we can handle the creation of a unique, writeable copy of
the `$KUBECONFIG` for every consumer.

/assign @bbguimaraes 
/cc @alvaroaleman 

follow-up to #1691

### DIFF
--- a/cmd/secret-wrapper/main.go
+++ b/cmd/secret-wrapper/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Needed for DPTP-1668

bbguimaraes:
LGTM

PR #1694, opened 2021-02-17
e2e: unify artifact dir location

stevekuznetsov:
Previously, logs and other artifacts from acessory processes were not
stored in the output artifacts for a test.

/assign @alvaroaleman


stevekuznetsov:
Needed for DPTP-1668

test/e2e/framework/ci-operator.go
```diff
@@ -116,10 +107,10 @@ func (c *CiOperatorCommand) Run() ([]byte, error) {
 			c.t.Errorf("failed to close logfile: %v", err)
 		}
 	}()
-	log := bytes.Buffer{}
-	tee := io.TeeReader(&log, logFile)
-	c.cmd.Stdout = &log
-	c.cmd.Stderr = &log
+	outputBuffer := bytes.Buffer{}
+	log := io.MultiWriter(logFile, &outputBuffer)
```
alvaroaleman, line 111: Why the buffer instead of just os.Stdout?

alvaroaleman (approved):

test/e2e/framework/ci-operator.go
stevekuznetsov: Streaming 100 parallel tests to stdout is not useful. Buffer is needed since we allow tests to check that the output of the command contains things.

stevekuznetsov:
.... how did this break vendor?

PR #1696, opened 2021-02-17
Add vault subpath proxy

alvaroaleman:
Adds a vault subpath proxy that shows directories in a kv engine the user has access to. Apart from the included tests I have also veriified via the UI and CLI that everything works.

The tests will fail until we add a vault binary in there.

Ref https://issues.redhat.com/browse/DPTP-1890


stevekuznetsov (commented):
Code and tests look good, can we add a quick README that explains the problem we're trying to solve and the approach (resultant ACL API) we're using to solve it?

alvaroaleman:
@stevekuznetsov updated, PTAL

stevekuznetsov (approved):

PR #1703, opened 2021-02-18
blocking-issue-creator: allow using GH proxy

petr-muller:
The tool is recently hitting API abuse limits (https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-openshift-release-merge-blockers/1362446851738963968#1:build-log.txt%3A93), going through the proxy should help.

Unfortunately the client does not seem to have any client throttling nor does it expose the rate limit error in any reasonable way for us to act on.

### DIFF
--- a/cmd/blocking-issue-creator/main.go
+++ b/cmd/blocking-issue-creator/main.go
@@ -22,8 +22,9 @@ import (
 
 type options struct {
 	promotion.FutureOptions
-	username  string
-	tokenPath string
+	username          string
+	tokenPath         string
+	ghGraphqlEndpoint string
 }
 
 func (o *options) Validate() error {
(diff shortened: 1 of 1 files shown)


cmd/blocking-issue-creator/main.go
```diff
@@ -44,6 +45,7 @@ func gatherOptions() options {
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	fs.StringVar(&o.username, "username", "", "Username to use when communicating with GitHub.")
 	fs.StringVar(&o.tokenPath, "token-path", "", "Path to token to use when communicating with GitHub.")
+	fs.StringVar(&o.ghGraphqlEndpoint, "github-graphql-endpoint", "https://api.github.com/graphql", "GH GraphQL endpoint URL")
```
alvaroaleman, line 48: Those are not cacheable at all, this will only help with getting metrics. Ratelimiting might help to not hit the abuse limit, but if possible it might be better to replace the graphql requests with v3 api requests so we can cache

stevekuznetsov:
Can't cache v4 api :(

cmd/blocking-issue-creator/main.go
petr-muller: Avoiding hitting the abuse limit is my primary goal and being counted in metrics is a good thing too. Given how trivial this PR is compared to making it use a differnt client, can we try merging this?

alvaroaleman: yeah but this change doesn't do anything that would help with not hitting the abuse limit

petr-muller: I lived under impression that the proxy itself does some throttling to avoid abuse limits :facepalm:

PR #1704, opened 2021-02-18
Add DockerfileLiteral option to images

alvaroaleman:
This is very useful for trivial dockerfiles like "copy one thing from a
different image" or "install one package"


alvaroaleman:
Came up in the context of https://issues.redhat.com/browse/DPTP-1890 because that makes us need an additional binary in the unit test image and being able to do that just in ci-operator config woul be nice

stevekuznetsov (approved):

PR #1705, opened 2021-02-18
[wip]secret-syncer: ensure the target namespace exists

hongkailiu:
Follow up https://coreos.slack.com/archives/CHY2E1BL4/p1613674179255500

/cc @alvaroaleman @petr-muller

### DIFF
--- a/pkg/controller/secretsyncer/secretsyncer.go
+++ b/pkg/controller/secretsyncer/secretsyncer.go
@@ -188,6 +188,19 @@ func (r *reconciler) mirrorSecret(ctx context.Context, log *logrus.Entry, cluste
 		}
 		data[k] = source.Data[k]
 	}
+
+	if err := c.Get(ctx, types.NamespacedName{Name: to.Namespace}, &corev1.Namespace{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to check if namespace %s exists: %w", to.Namespace, err)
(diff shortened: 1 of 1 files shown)


pkg/controller/secretsyncer/secretsyncer.go
```diff
@@ -188,6 +188,19 @@ func (r *reconciler) mirrorSecret(ctx context.Context, log *logrus.Entry, cluste
 		}
 		data[k] = source.Data[k]
 	}
+
+	if err := c.Get(ctx, types.NamespacedName{Name: to.Namespace}, &corev1.Namespace{}); err != nil {
```
alvaroaleman, line 192: Can we instead add a config option to limit the target clusters? Users are not supposed to use this to create namesapces. The secret in question is only needed for app.ci

hongkailiu: Ack. Will do something like cluster group.

hongkailiu: Hu.
It has been implemented.
https://github.com/openshift/ci-tools/blob/b0351c25136f3e2695d5f2364dada541010230fe/pkg/controller/secretsyncer/config/config.go#L42

hongkailiu: /close

https://github.com/openshift/release/pull/16036

PR #1709, opened 2021-02-19
ci-operator: remove 2min from every test's runtime

stevekuznetsov:


stevekuznetsov:
those are very nice doggos

PR #1711, opened 2021-02-19
ci-operator: be more intelligent with grace periods

stevekuznetsov:
Today, the test steps' grace periods are simply inherited from their
parent job. Test steps have their own Pod grace periods (either default
or configured by the user), so we should set our test process to have a
grace period shy of that.

In the future we should deprecate the existing active deadline seconds
and grace period fields on the test steps and use the Prow timeout and
grace period concepts, as they are more friendly to users.

/assign @alvaroaleman @bbguimaraes

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -14,6 +14,7 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
+	v1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/entrypoint"
 	utilpointer "k8s.io/utils/pointer"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
Part of DPTP-1668

pkg/steps/multi_stage.go
```diff
@@ -407,6 +408,15 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
...
 		if s.artifactsViaPodUtils {
 			artifactDir = fmt.Sprintf("%s/%s", s.name, step.As)
+			gracePeriod := 30 * time.Second
+			if step.TerminationGracePeriodSeconds != nil {
+				gracePeriod = time.Duration(*step.TerminationGracePeriodSeconds) * time.Second
+			}
+			// We want upload to have some time to do what it needs to do, so set
+			// the grace period for the process to be just shy of that, assuming
+			// an 80/20 distribution of work.
+			gracePeriod = (gracePeriod * 4) / 5
```
alvaroaleman, line 418: Isn't this redundant with the test-infra change?

alvaroaleman: No wait, we are removing it here again? Why?

stevekuznetsov: No, for two reasons:

 1. `test-infra` controls the `ci-operator` Pod that we start for the Prow job, and the user controls there the execution time and grace period of the test process (we increase the amount to set the grace period of the Pod)
 2. `ci-tools` here controls the test pods for a step in a workflow, and the user today sets the overall grace period for the Pod (will fix this), so we reduce the given time to get the grace period for the process

stevekuznetsov: (So the two parts of the code are unrelated and act on separate Pods with separate user inputs. My goal is to make users set timeout/grace period on step pods the same way in the future)

alvaroaleman (approved):

PR #1715, opened 2021-02-22
ci-operator: improve user experience for timeouts

stevekuznetsov:
Today, timeouts and grace periods are in theory possible for test steps
in a multi-stage workflow. However, there are two major drawbacks with
the implementation:

 - pod.spec.activeDeadlineSeconds constrains the "active time" of a Pod
   on a node, which *includes* the time to mount volumes, pull images,
   set up the sandbox, etc. The implication is therefore that a user's
   setting for "timeout" may expire without the process they wish to
   constrain ever even starting. As we allow a 15 minute image pull
   timeout, no setting of active deadline shorter than that is
   guaranteed to be applied sensibly.
 - pod.spec.terminationGracePeriodSeconds constrains the full window
   between the Pod being asked to delete (and getting SIGTERM) and the
   time at which the kubelet sends SIGKILL. Test steps often execute two
   distinct phases during this time: in-process cleanup and post-process
   log and artifact upload. The user is *only* in control of the former,
   but this grace period must be set long enough for both to succeed.
   Users will nto get this right.

Now that we inject the Prow pod-utils into each test step (modulo some
that are actively being removed as we deprecate running without them),
we can use the entrypoint utility to control the timeout and grace
period for steps in a multi-stage workflow. This fixes both of the above
issues.

/assign @alvaroaleman @wking

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strings"
 
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/repoowners"
 )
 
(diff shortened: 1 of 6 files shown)


stevekuznetsov:
Part of DPTP-1668

alvaroaleman (approved):

PR #1720, opened 2021-02-23
[DPTP-2014] Refactor blocking-issue-creator to use the prow github client

droslean:
The old implementation creates flakes like [job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-openshift-release-merge-blockers/1364198484130926592)

This PR refactors the `blocking-issue-creator` tool to use the Prow's GitHub client

follow-up https://github.com/openshift/release/pull/16280

### DIFF
--- a/cmd/blocking-issue-creator/main.go
+++ b/cmd/blocking-issue-creator/main.go
@@ -1,19 +1,16 @@
 package main
 
 import (
-	"context"
-	"errors"
 	"flag"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strings"
 
-	githubql "github.com/shurcooL/githubv4"
 	"github.com/sirupsen/logrus"
-	"golang.org/x/oauth2"
 
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/config/secret"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
@@ -22,19 +19,17 @@ import (
 
 type options struct {
 	promotion.FutureOptions
-	username  string
-	tokenPath string
+	github prowflagutil.GitHubOptions
+
+	dryRun bool
 }
 
 func (o *options) Validate() error {
 	if err := o.FutureOptions.Validate(); err != nil {
 		return err
 	}
-	if o.username == "" {
-		return errors.New("--username is required")
-	}
-	if o.tokenPath == "" {
-		return errors.New("--token-path is required")
+	if err := o.github.Validate(o.dryRun); err != nil {
+		return err
 	}
 	return nil
 }
@@ -42,9 +37,12 @@ func (o *options) Validate() error {
 func gatherOptions() options {
 	o := options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	fs.StringVar(&o.username, "username", "", "Username to use when communicating with GitHub.")
-	fs.StringVar(&o.tokenPath, "token-path", "", "Path to token to use when communicating with GitHub.")
-	o.Bind(fs)
+
+	fs.BoolVar(&o.dryRun, "dry-run", false, "Dry run for testing. Uses API tokens but does not mutate.")
+
+	o.github.AddFlags(fs)
+	o.FutureOptions.Bind(fs)
+
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("could not parse input")
 	}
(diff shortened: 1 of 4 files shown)


cmd/blocking-issue-creator/main.go
```diff
@@ -190,6 +191,7 @@ func main() {
 			}
 
 			logger.Infof("Updated issue %d", updateIssue.UpdateIssue.Issue.Number)
+			time.Sleep(time.Second * 1)
```
alvaroaleman, line 194: Can we just use the prow gh client and enable the throttler on it?

droslean: Yea I will take a look and do the changes. I hold the PR

stevekuznetsov: Just did the math - we need to make ~200 calls once an hour, so a sleep is trivially correct (although the sleeps in this PR are not - we need to wait after the search). If it's easy to do the throttler, I'd prefer that

droslean:
/hold

Holding until https://github.com/openshift/ci-tools/pull/1728 lands.

stevekuznetsov (changes requested):
I'm almost tempted to say I won't review this. By changing the entire file, it's almost impossible to reason about your changes. You should be able to achieve this without refactoring the entire tool.

Can you try re-applying a _minimal_ set of changes, please?

cmd/blocking-issue-creator/main.go
```diff
@@ -87,187 +104,77 @@ func main() {
...
 
-			if !o.Confirm {
-				logger.Info("Would create issue.")
-				return nil
+	if len(issues) > 1 {
+		logger.Warnf("Found more than one merge blocking issue by the bot: %v", len(issues))
+		for _, issue := range issues[1:] {
+			if err := client.CloseIssue(repoInfo.Org, repoInfo.Repo, issue.Number); err != nil {
+				logger.WithError(err).Error("Failed to close issue.")
+				return err
```
stevekuznetsov, line 143: For instance, your refactor makes this flow change. Now, we fail and exit early if we can't close an issue. That is not intended.

droslean:
> Can you try re-applying a minimal set of changes, please?

@stevekuznetsov I split the changes into two commits.
- Remove the graphql client and use the prow github one
- Put the whole logic in a function and add some unittests. 

It should be easier to review now. I should have done this in the first place, sorry about that.

cmd/blocking-issue-creator/main.go
```diff
@@ -98,71 +106,76 @@ func main() {
...
+	}); err != nil || failed {
+		logrus.WithError(err).Fatal("Could not publish merge blocking issues.")
+	}
+}
 
-		// we have an existing issue that needs to be up to date
-		if len(issues) != 0 {
-			logger = logger.WithField("merge-blocker", issues[0])
-			existing := issues[0]
+func operateOnIssues(client githubClient, githubLogin string, repoInfo *config.Info, branches sets.String, logger *logrus.Entry) error {
```
petr-muller, line 120: "operate on issues" is not a useful name. I would expect such name for a function that iterates over all issues and does a custom "some operation" on each, where the custom operation is e.g. a parameter function or something similar.

cmd/blocking-issue-creator/main_test.go
```diff
@@ -0,0 +1,192 @@
...
+			fgh.FakeClient.Issues = tc.issues
+
+			if err := operateOnIssues(fgh, "", tc.repoInfo, tc.branches, logrus.WithField("id", tc.id)); err != nil {
+				t.Fatal(err)
+			}
+
+			openedIssues, _ := fgh.ListOpenIssues(tc.repoInfo.Org, tc.repoInfo.Repo)
+			sort.Slice(openedIssues, func(i, j int) bool { return openedIssues[i].ID < openedIssues[j].ID })
+
+			if !reflect.DeepEqual(openedIssues, tc.expectedIssues) {
```
petr-muller, line 187: We have an established pattern for these checks:

```
if diff = cmp.Diff(a,b); diff != "" {
  t.Errorf("...%s", diff)
}
```

cmd/blocking-issue-creator/main.go
```diff
@@ -22,29 +20,36 @@ import (
...
 
 func gatherOptions() options {
 	o := options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	fs.StringVar(&o.username, "username", "", "Username to use when communicating with GitHub.")
-	fs.StringVar(&o.tokenPath, "token-path", "", "Path to token to use when communicating with GitHub.")
-	o.Bind(fs)
+
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Dry run for testing. Uses API tokens but does not mutate.")
+	fs.StringVar(&o.githubLogin, "github-login", "", "The GitHub username to use.")
```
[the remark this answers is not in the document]
stevekuznetsov: Use the client's `BotUser()` call

cmd/blocking-issue-creator/main.go
```diff
@@ -22,29 +20,36 @@ import (
...
 }
 
 func gatherOptions() options {
 	o := options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	fs.StringVar(&o.username, "username", "", "Username to use when communicating with GitHub.")
-	fs.StringVar(&o.tokenPath, "token-path", "", "Path to token to use when communicating with GitHub.")
-	o.Bind(fs)
+
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Dry run for testing. Uses API tokens but does not mutate.")
```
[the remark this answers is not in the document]
droslean: Yea I have a [follow-up](https://github.com/openshift/release/pull/16280) for that

cmd/blocking-issue-creator/main.go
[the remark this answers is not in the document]
droslean: To use it in the query for finding the issues. The tool was searching for issues authored by this name even before, so I didn't want to change the behavior here.

cmd/blocking-issue-creator/main.go
```diff
@@ -96,178 +107,62 @@ func main() {
...
-				ID:    issues[0].ID,
-				Title: githubql.String(title),
-				Body:  githubql.String(body),
-			}
-			if !o.Confirm {
-				logger.Info("Would update issue.")
-				return nil
-			}
-			if err := client.Mutate(context.Background(), &updateIssue, input, nil); err != nil {
+			toBeUpdated := existing
```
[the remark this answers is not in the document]
droslean: Either way, it doesn't matter since we don't do anything else with the `existing` variable. I made it this way to make the code more understandable.

cmd/blocking-issue-creator/main.go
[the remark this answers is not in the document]
petr-muller: Steve's point is that there will be a period of time where the job is broken (between merging the code change and manifest change). The usual way (which we should do by default) to avoid breakages in cases like this is:

1) Add a new option, keep the old option
2) Move the users to use the new options
3) Remove old option

petr-muller (approved):

PR #1723, opened 2021-02-23
Check sanitize's config if it can dispatch jobs determinically

hongkailiu:
https://issues.redhat.com/browse/DPTP-1984

We have 3 places to use the config:
- `sanitize`: read the config as input (run in a presubmit)
- `dispatcher`: read the config as input, query promotheus and then output the config (run in a periodic)
- us: change the config to move jobs

We ensure
- no job name will occur more than once in the config.
- no path of the job yaml matches more than one regex in the config

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -170,7 +170,7 @@ type clusterVolume struct {
 // findClusterForJobConfig finds a cluster running on a preferred cloud provider for the jobs in a Prow job config.
 // The chosen cluster will be the one with minimal workload with the given cloud provider.
 // If the cluster provider is empty string, it will choose the one with minimal workload across all cloud providers.
-func (cv *clusterVolume) findClusterForJobConfig(cloudProvider string, jc *prowconfig.JobConfig, path string, config *dispatcher.Config, jobVolumes map[string]float64) string {
+func (cv *clusterVolume) findClusterForJobConfig(cloudProvider string, jc *prowconfig.JobConfig, path string, config *dispatcher.Config, jobVolumes map[string]float64) (string, error) {
 	//no cluster in the build farm is from the targeting cloud provider
 	if _, ok := cv.clusterVolumeMap[cloudProvider]; !ok {
 		cloudProvider = ""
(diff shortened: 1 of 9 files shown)


cmd/prow-job-dispatcher/main.go
```diff
@@ -392,6 +392,9 @@ func main() {
 	if err := dispatchJobs(context.TODO(), o.prowJobConfigDir, o.maxConcurrency, config, jobVolumes); err != nil {
 		logrus.WithError(err).Fatal("Failed to dispatch")
 	}
+	if err := config.AreDeterministic(o.prowJobConfigDir); err != nil {
+		logrus.WithError(err).Fatal("Failed to check if the config can determinize jobs")
+	}
```
petr-muller, line 397: This looks a bit weird to me, I would expect the dispatcher tool to never generate an incorrect config, not check and fail when it does. Can we detect the situation and fix it instead of failing -- like select one of the options or something else.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -140,6 +140,9 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatalf("Failed to load config from %q", opt.configPath)
 	}
+	if err := config.AreDeterministic(opt.prowJobConfigDir); err != nil {
+		logrus.WithError(err).Fatal("Failed to check if the config can determinize jobs")
+	}
```
petr-muller, line 145: So this ends up doing a second full walkthrough? That sounds excessive.

Perhaps we can, while sanitizing a single job, walk through the whole config instead of selecting the first match? We would always iterate through all possibilities and if more than one matched, we would record a failure.

petr-muller (commented):
I only skimmed the PR because it's still a WIP but the approach looks a bit different that I would expect.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Let me try.

cmd/prow-job-dispatcher/main.go
hongkailiu: Removed from dispatcher.
If dispatcher generates a bad config (that would be a bug), then sanitize (presubmit) will failed and the PR from dispatcher wont be merged.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Updated.

hongkailiu: Why github does not show `Outdated`? I updated the PR.

hongkailiu:
/hold

Need to run dispatcher once.

pkg/dispatcher/config.go
```diff
@@ -210,6 +211,54 @@ func LoadConfig(configPath string) (*Config, error) {
 	return config, nil
 }
 
+func (config *Config) JobsWithMultiOccurrence() []string {
+	records := map[string]int{}
+	for _, group := range config.Groups {
+		for _, job := range group.Jobs {
+			records[job] = records[job] + 1
```
petr-muller, line 218: golang has `++`, why not use it?

pkg/dispatcher/config.go
```diff
@@ -210,6 +211,54 @@ func LoadConfig(configPath string) (*Config, error) {
 	return config, nil
 }
 
+func (config *Config) JobsWithMultiOccurrence() []string {
```
petr-muller, line 214: if itr's exported, it should have GoDoc. Also, I think that because this takes no input, this should be just called `Validate`, and return an appropriate error, not let the callsite do it.

cmd/sanitize-prow-jobs/main.go
```diff
@@ -59,6 +59,11 @@ func determinizeJobs(prowJobConfigDir string, config *dispatcher.Config) error {
 			return nil
 		}
 
+		matches := config.AllMatchingPathRegex(path)
+		if len(matches) > 1 {
+			errChan <- fmt.Errorf("path %s matches more than 1 regex: %s", path, matches)
+		}
+
```
petr-muller, line 66: It's better, but I think this should jsut be somewhere under `GetClusterForJob`. Somewhere under that method, we end up doing the same thing (iterate over regexes), and that method should IMO do the check itself, not rely on some code doing it first.

The `DetermineClusterForJob` method seems like a good place. That method should not rely on the config being pre-checked, it should detect the possible duplicity and end with an error (which needs to be propagate ot the `defaultJobConfig` here.

pkg/dispatcher/config.go
hongkailiu: I cannot see `Outdated`, although i have updated the line.
/shrug

pkg/dispatcher/config.go
hongkailiu: Updated.

cmd/sanitize-prow-jobs/main.go
hongkailiu: Done.

Propagate the error all the way up.

hongkailiu:
gee. shrug has a label.

/unshrug

cmd/prow-job-dispatcher/main.go
```diff
@@ -191,45 +191,62 @@ func (cv *clusterVolume) findClusterForJobConfig(cloudProvider string, jc *prowc
...
-		cluster = cv.findClusterForJobConfig(cloudProvider, jc, path, config, jobVolumes)
+		if cluster, err = cv.findClusterForJobConfig(cloudProvider, jc, path, config, jobVolumes); err != nil {
+			return "", fmt.Errorf("fail to find cluster for job config: %w", err)
+		}
+
 	} else {
-		cluster = cv.findClusterForJobConfig("", jc, path, config, jobVolumes)
+		if cluster, err = cv.findClusterForJobConfig("", jc, path, config, jobVolumes); err != nil {
+			return "", fmt.Errorf("fail to find cluster for job config: %w", err)
+		}
```
petr-muller, line 247: Looks like this can be simplified even more to

```go
var cloudProvider string
if cloudProvidersForE2ETests.Len() == 1 {
  cloudProvider, _ = cloudProvidersForE2ETests.PopAny()
}
if cluster, err = cv.findClusterForJobConfig(cloudProvider, jc, path, config, jobVolumes); err != nil {
   return "", fmt.Errorf("fail to find cluster for job config: %w", err)
}
```
```

pkg/dispatcher/config.go
petr-muller: I see you updated the occurrence below (L247) but not here

petr-muller (commented):
We're getting there, few more comments and we're good to go. Thanks for taking the effort.

pkg/dispatcher/config.go
petr-muller: I checked out the code form GH to be sure it does not lie ;)

cmd/prow-job-dispatcher/main.go
hongkailiu: Updated.

petr-muller (approved):

PR #1727, opened 2021-02-25
cmd/testgrid-config-generator: All release periodics should appear in testgrid

smarterclayton:
Release periodics are all informing, and testgrid visibility is a key part of TRT automation. Jobs that match the standard pattern should appear in testgrid unless there is a reason. We would never allow a release periodic in the standard set that do not benefit our ability to understand what is in the release. Whether a job is old or new is irrelevant, it's input to our decision processes.

This commit allows the allow list to simply become an override of which dashboard a small set of jobs should appear in. After this change the allow list should be a very, very small set.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -230,7 +230,7 @@ func main() {
 			var dashboardType string
 			switch {
 			case job.Upgrade:
-				dashboardType = "generic-informing"
+				dashboardType = "informing"
 			case job.Optional:
 				if existing == "generic-informing" || existing == "blocking" {
 					continue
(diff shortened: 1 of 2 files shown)


cmd/testgrid-config-generator/main.go
```diff
@@ -309,7 +325,7 @@ func main() {
 		default:
 			var stream string
 			switch {
-			case strings.Contains(name, "-ocp-") || strings.Contains(name, "-origin-"):
+			case strings.Contains(name, "-ocp-"), strings.Contains(name, "-origin-"), strings.Contains(name, "-ci-"), strings.Contains(name, "-nightly-"):
```
smarterclayton, line 328: /hold

the CI one has to be more specific

smarterclayton:
/hold cancel

@petr-muller can you review?

smarterclayton:
Note this will be a 15-30% bump in number of testgrid jobs we add, but if we start adding too many i think we can tune the retention a bit.

petr-muller (approved):

PR #1731, opened 2021-02-25
ci-operator: stop allowing users to set artifact dir

stevekuznetsov:
No longer are users able to set the explicit directory for:
 - `ci-operator` outputs
 - simple Pod tests
 - multi-stage test steps

`ci-operator` itself will read the `${ARTIFACTS}` value to determine
where to put its own artifacts (like the output jUnit files). This is in
effect what already happens, as this is the fallback to when the flag is
not passed, and we do not pass it in generated jobs. The flag will
remain in a deprecated manner so as to not break users.

All test steps *except* for `Template` tests are already using the
test-infra pod-utils to upload, and therefore are not allowed to set a
non-standard directory, anyway. Nothing changes for these steps. It is
now not possible to opt out of this by telling `ci-operator` not to use
pod-utils.

We continue to use the deprecated artifact fetching sidecar code for
Templates, but only for templates.

This will cause a breaking change to prowgen as our previous automation
explicitly wrote our defaulted value into all configs even if they never
set it.


stevekuznetsov:
Part of DPTP-1668

pkg/steps/artifacts.go
```diff
@@ -366,27 +367,6 @@ func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.
 	return nil
 }
 
-func addArtifacts(pod *coreapi.Pod, artifactDir string) {
```
stevekuznetsov, line 369: This is suspicious but I think templates don't actually use this, they use `addArtifactsToPod` ...

stevekuznetsov:
Hmm I think I can split this up

stevekuznetsov:
OK, so it theoretically should be possible to split this into steps:

 1. remove the feature from basic container jobs
 2. remove it from step registry stuff
 3. remove it from the ci-operator top-level

I'll try to do that - there are areas where we infer 1&2 from 3 randomly so it's a bit messy but probably doable.

stevekuznetsov:
@alvaroaleman @petr-muller updated this to be minimal now that the rest of the changes have come in. Notably, this *removes* the `--artifact-dir` setting from `ci-operator` itself. I think this is a useful thing to do as:

1. we never set it explicitly
2. we only care about using it when Prow's running us

In `openshift/release` today, we have this:

```
$ git grep -h artifact-dir | head -n 1
        - --artifact-dir=$(ARTIFACTS)
$ git grep artifact-dir | wc -l
563
$ git grep artifact-dir | grep -v ARTIFACTS | wc -l
0
```

So there's 563 places where we set it to what would otherwise be the default value anyway, so there's zero utility to this flag. In addition, having to pass this directory around _everywhere_ in our codebase ended up being a bad cargo-cult ... half the places didn't even end up using it, and in some cases like multi-stage it looks like we accidentally used this instead of the per-step one ...

images/entrypoint-wrapper/Dockerfile
```diff
@@ -3,3 +3,4 @@ LABEL maintainer="bbarcaro@redhat.com"
 
 ADD entrypoint-wrapper /usr/bin/entrypoint-wrapper
 ENTRYPOINT ["/usr/bin/entrypoint-wrapper"]
+
```
alvaroaleman, line 6: ?

pkg/util/env.go
```diff
@@ -0,0 +1,10 @@
+package util
+
+import "os"
+
+// prowArtifactsEnv is the directory Prow wants us to put artifacts into for upload
+const prowArtifactsEnv string = "ARTIFACTS"
```
alvaroaleman, line 6: IMHO this is not util, this is `api`

images/entrypoint-wrapper/Dockerfile
stevekuznetsov: fat fingers

cmd/ci-operator/main.go
```diff
@@ -186,6 +183,9 @@ func main() {
 		if err := flag.CommandLine.Set("v", "10"); err != nil {
 			logrus.WithError(err).Error("Failed to set flag -v=10")
 		}
+		logrus.SetLevel(logrus.TraceLevel)
```
alvaroaleman, line 186: ?

cmd/ci-operator/main.go
```diff
@@ -345,7 +345,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.BoolVar(&opt.promote, "promote", false, "When all other targets complete, publish the set of images built by this job into the release configuration.")
 
 	// output control
-	flag.StringVar(&opt.artifactDir, "artifact-dir", "", "If set grab artifacts from test and template jobs. Defaults to $ARTIFACTS if set.")
+	flag.StringVar(&opt.artifactDir, "artifact-dir", "", "DEPRECATED.")
```
alvaroaleman, line 348: Please also add the info that it doesn't do anything

images/entrypoint-wrapper/Dockerfile
```diff
@@ -2,4 +2,4 @@ FROM centos:8
 LABEL maintainer="bbarcaro@redhat.com"
 
 ADD entrypoint-wrapper /usr/bin/entrypoint-wrapper
-ENTRYPOINT ["/usr/bin/entrypoint-wrapper"]
+ENTRYPOINT ["/usr/bin/entrypoint-wrapper"]
```
alvaroaleman, line 5: ?

cmd/ci-operator/main.go
stevekuznetsov: This was part of my commit to add logging to the artifacts thing. The default log library we use has no concept of levels, and we use logrus anyway, so I just added logging to it in Trace level, and setting ci-operator verbose mode opens that up.

images/entrypoint-wrapper/Dockerfile
stevekuznetsov: sigh

stevekuznetsov:
@alvaroaleman updated

stevekuznetsov:
Seeing a lot of this lately

```
 error: error creating buildah builder: Error writing blob: error storing blob to file "/var/tmp/storage350157992/5": unexpected EOF 
```

/retest

stevekuznetsov:
Rebased and ready @alvaroaleman @petr-muller

cmd/ci-operator/main.go
alvaroaleman: I think my larger point was that almost everywhere we run logrus with Debug level by default and reserve trace level logs for things that you need to opt in to get them, I think it is valuable to be consistent with that

petr-muller: Well this *is* opt-in, it's under verbose so I guess this is fine?

petr-muller (approved):
LGTM

/hold
Holding to allow resolving Alvaro's comment

cmd/ci-operator/main.go
stevekuznetsov: Yeah you need to `--verbose` the `ci-operator` binary to get this, which you'd only be doing in dev.

stevekuznetsov:
/hold cancel

You need to opt in via flag

PR #1732, opened 2021-02-25
ci-operator: disallow opting out of pod-utils

stevekuznetsov:
We already opt everyone in by default and as of yesterday nobody opts
back out, so this should be a no-op.

Part 1 of superseding https://github.com/openshift/ci-tools/pull/1731
Part of DPTP-1668

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -366,27 +366,6 @@ func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.
 	return nil
 }
 
-func addArtifacts(pod *coreapi.Pod, artifactDir string) {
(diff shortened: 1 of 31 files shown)


pkg/steps/artifacts.go
```diff
@@ -366,27 +366,6 @@ func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.
 	return nil
 }
 
-func addArtifacts(pod *coreapi.Pod, artifactDir string) {
```
stevekuznetsov, line 369: This _looks_ super suspect, but it turns out while Template code does the same thing, it uses a separate set of methods to do it. Sigh.

alvaroaleman (approved):

PR #1735, opened 2021-02-26
ci-operator: stop allowing users to set artifact dir for pods

stevekuznetsov:
As can be seen by the validation that was removed in this, it's already
not valid to set this to anything other than the default and the value
is unused.

Part 2 of superseding #1731
Part of [DPTP-1668](https://issues.redhat.com/browse/DPTP-1668)

Will introduce breaking changes to prowgen since we serialize the default even when it's unset.


stevekuznetsov:
You can see by inspection that we don't actually use this field at all, I'll self-merge so this can be wrapped up this sprint.

PR #1742, opened 2021-03-02
prowgen: generate kvm label

hongkailiu:
https://issues.redhat.com/browse/DPTP-2017

/cc @openshift/openshift-team-developer-productivity-test-platform @cgwalters

### DIFF
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -178,7 +178,12 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 			postsubmit.MaxConcurrency = 1
 			postsubmits[orgrepo] = append(postsubmits[orgrepo], *postsubmit)
 		} else {
-			presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(element.As, info, podSpec, configSpec.CanonicalGoRepository, jobRelease, skipCloning))
+			presubmit := *generatePresubmitForTest(element.As, info, podSpec, configSpec.CanonicalGoRepository, jobRelease, skipCloning)
+			v, requestingKVM := configSpec.Resources.RequirementsForStep(element.As).Requests[cioperatorapi.KVMDeviceLabel]
+			if requestingKVM {
+				presubmit.Labels[cioperatorapi.KVMDeviceLabel] = v
+			}
+			presubmits[orgrepo] = append(presubmits[orgrepo], presubmit)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/prowgen/prowgen.go
```diff
@@ -178,7 +178,12 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 			postsubmit.MaxConcurrency = 1
 			postsubmits[orgrepo] = append(postsubmits[orgrepo], *postsubmit)
 		} else {
-			presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(element.As, info, podSpec, configSpec.CanonicalGoRepository, jobRelease, skipCloning))
+			presubmit := *generatePresubmitForTest(element.As, info, podSpec, configSpec.CanonicalGoRepository, jobRelease, skipCloning)
+			v, requestingKVM := configSpec.Resources.RequirementsForStep(element.As).Requests["devices.kubevirt.io/kvm"]
```
alvaroaleman, line 182: Add a test, please

hongkailiu: Added.

pkg/prowgen/prowgen_test.go
```diff
@@ -469,6 +469,21 @@ func TestGenerateJobs(t *testing.T) {
 				Repo:   "repository",
 				Branch: "branch",
 			}},
+		}, {
+			id: "kvm label",
+			config: &ciop.ReleaseBuildConfiguration{
+				Resources: map[string]ciop.ResourceRequirements{
+					"*": {Requests: ciop.ResourceList{"devices.kubevirt.io/kvm": "1"}},
```
stevekuznetsov, line 476: very clever, or too clever? only time will tell

alvaroaleman (approved):

pkg/prowgen/prowgen_test.go
hongkailiu: Not sure if I get your point.
That is how our user uses it for now.

https://github.com/openshift/release/blob/0f651ff1091687662a4799e57f5455504a45a4d3/ci-operator/config/openshift/os/openshift-os-master.yaml#L13

I cannot see what it means to build pods though.
Is it what you think only time will tell?

hongkailiu: avoid to apply to build pods: the 2nd use case
https://github.com/openshift/release/blob/f896da9216d4742ce591f04fad80331dfc8f47c4/ci-operator/config/coreos/rpm-ostree/coreos-rpm-ostree-master.yaml#L26-L37

PR #1744, opened 2021-03-03
[DPTP-1969] Implement a more sophisticated way to chooses a subset of jobs when there is a rehearsal limit

droslean:
Currently, the design of the tool is really fragile when picking all the presubmits to rehearse. In the effort of avoiding bad refactoring, this PR introduces a new label `pj-rehearse.openshift.io/source-type` that is injected in all rehearsals, and based on its value the tool will sophisticated choose a subset of the jobs to be rehearsed when the jobs are more than the rehearsal limit. 

Pros: We can use this label to gather Prometheus metrics from the rehearsals and create grafana dashboard to have a clear picture of which changes are triggering rehearsals. 

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -5,9 +5,9 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"math/rand"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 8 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -572,3 +571,37 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+	dropped := []*prowconfig.Presubmit{}
+
+	for _, jobs := range presubmitsBySourceType {
+		if len(jobs) > maxJobsPerSourceType {
+			toRehearse = append(toRehearse, jobs[:maxJobsPerSourceType]...)
+			dropped = append(dropped, jobs[maxJobsPerSourceType:]...)
+		}
+	}
+
+	if len(toRehearse) <= rehearsalLimit {
```
stevekuznetsov, line 601: Can you add a comment here that this might happen due to the rounding inherent in integer division? Also, can we get a test for when you need to add more?

stevekuznetsov (commented):
One comment, seems like a nice way to do this +1

cmd/pj-rehearse/main.go
droslean: Comment added as well as a new test case for this.

stevekuznetsov (approved):

PR #1746, opened 2021-03-04
pkg/registry: emit less arcane error messages

bbguimaraes:
This addresses https://issues.redhat.com/browse/DPTP-2043, among other related
improvements.

See the tests in the final `diff` for examples.

### DIFF
--- a/pkg/registry/stack.go
+++ b/pkg/registry/stack.go
@@ -1,0 +1,119 @@
+package registry
+
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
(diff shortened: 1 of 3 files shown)


pkg/registry/stack.go
```diff
@@ -0,0 +1,137 @@
...
+					r.unusedDeps.Delete(e.Env)
+				}
+				return r.deps[j].Name
+			}
+		}
+	}
+	return ""
+}
+
+func (s *stack) checkUnused(r *stackRecord) (ret []error) {
```
bbguimaraes, line 78: The cubic loops here are horrible, but they're exceptional cases and the depth of the stack is never too big.

stevekuznetsov (approved):

PR #1748, opened 2021-03-04
ci-operator: add support for operator upgrade tests

AlexNPavel:
This PR adds support for testing optional operator upgrades in openshift
CI.

To allow optional operator upgrade tests, 3 fields have been added to
the bundle definition field (`operator.bundles`). They are:
- `as`: the name of the bundle image. If this is set, the built bundle
  will be tagged as `pipeline:as` and a corresponding index will be
  built called `pipeline:ci-index-as`. This will also be useful for
  future multistage tests that will pass the bundle pullspec directly
  to the `operator-sdk` command
- `base_index`: define an existing index to add the new bundle to for
  the test instead of creating a new index
- `update_graph`: define the mode to add the built bundle to the index.
  Can be `semver`, `semver-skippatch`, or `replaces`

Any tests that had to be changed and did not follow our current test
conventions were restructured to match.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -141,7 +141,7 @@ func (config ReleaseBuildConfiguration) IsPipelineImage(name string) bool {
 		string(PipelineImageStreamTagReferenceIndexImage):
 		return true
 	}
-	return IsBundleImage(name)
+	return config.IsBundleImage(name)
 }
 
 // ResourceConfiguration defines resource overrides for jobs run
@@ -1136,10 +1136,28 @@ type OperatorStepConfiguration struct {
 	Substitutions []PullSpecSubstitution `json:"substitutions,omitempty"`
 }
 
-// Bundle contains the data needed to build a bundle from the bundle source image
+// IndexUpdate specifies the update mode for an operator being added to an index
+type IndexUpdate string
+
+const (
+	IndexUpdateSemver          = "semver"
+	IndexUpdateReplaces        = "replaces"
+	IndexUpdateSemverSkippatch = "semver-skippatch"
+)
+
+// Bundle contains the data needed to build a bundle from the bundle source image and update an index to include the new bundle
 type Bundle struct {
+	// As defines the name for this bundle. If not set, a name will be automatically generated for the bundle.
+	As string `json:"as,omitempty"`
+	// DockerfilePath defines where the dockerfile for build the bundle exists relative to the contextdir
 	DockerfilePath string `json:"dockerfile_path,omitempty"`
-	ContextDir     string `json:"context_dir,omitempty"`
+	// ContextDir defines the source directory to build the bundle from relative to the repository root
+	ContextDir string `json:"context_dir,omitempty"`
+	// BaseIndex defines what index image to use as a base when adding the bundle to an index
+	BaseIndex string `json:"base_index,omitempty"`
+	// UpdateGraph defines the update mode to use when adding the bundle to the base index.
+	// Can be: semver (default), semver-skippatch, or replaces
+	UpdateGraph IndexUpdate `json:"update_graph,omitempty"`
 }
 
 // IndexGeneratorStepConfiguration describes a step that creates an index database and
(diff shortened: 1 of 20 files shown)


AlexNPavel:
I think the code in general is complete. I am attempting to find a way to actually test that the index image is being built correctly, but currently our tests assume that images have `/bin/bash`, which the index images don't as they're alpine based, so we can't simply run commands in the container to test that the index has the correct information. I will work on a way to allow us to run commands in images without bash. I will keep this PR as a WIP until then.

stevekuznetsov:
What makes the index valid or invalid? Can you write an e2e test that depends on the image (`dependencies`) and uses `skopeo inspect` or something?

stevekuznetsov (commented):
Can you open a PR to `ci-docs` while this gets reviewed?

pkg/api/types.go
```diff
@@ -1179,11 +1215,22 @@ type BundleSourceStepConfiguration struct {
 // PipelineImageStreamTagReferenceBundleSourceName is the name of the bundle source image built by the CI
 const PipelineImageStreamTagReferenceBundleSource PipelineImageStreamTagReference = "src-bundle"
 
-// BundlePrefix is the prefix used by ci-operator for bundle images
+// BundlePrefix is the prefix used by ci-operator for bundle images without an explicitly configured name
```
stevekuznetsov, line 1218: Do we have validation that ensures this name does not collide with any `base_images` or `images` or baseline `pipeline` images?

pkg/steps/index_generator.go
```diff
@@ -94,7 +94,12 @@ func (s *indexGeneratorStep) indexGenDockerfile() (string, error) {
 		}
 		bundles = append(bundles, fullSpec)
 	}
-	dockerCommands = append(dockerCommands, fmt.Sprintf(`RUN ["opm", "index", "add", "--mode", "semver", "--bundles", "%s", "--out-dockerfile", "%s", "--generate"]`, strings.Join(bundles, ","), IndexDockerfileName))
+	opmCommand := fmt.Sprintf(`RUN ["opm", "index", "add", "--mode", "%s", "--bundles", "%s", "--out-dockerfile", "%s", "--generate"`, s.config.UpdateGraph, strings.Join(bundles, ","), IndexDockerfileName)
```
stevekuznetsov, line 97: You use `UpdateGraph` without checking for empty - do we have validation or defaulting on it?

stevekuznetsov (commented):
These changes look fairly straightforward, couple comments.

pkg/api/types.go
```diff
@@ -1143,10 +1143,28 @@ type OperatorStepConfiguration struct {
...
+const (
+	IndexUpdateSemver          = "semver"
+	IndexUpdateReplaces        = "replaces"
+	IndexUpdateSemverSkippatch = "semver-skippatch"
+)
+
+// Bundle contains the data needed to build a bundle from the bundle source image and update an index to include the new bundle
 type Bundle struct {
+	// As defines the name for this bundle
+	As string `json:"as,omitempty"`
```
petr-muller, line 1158: I think `As` being optional (omitempty) may be a bit confusing, we may want to point out in the GoDoc that it's intentional.

petr-muller (commented):
Minor non-blocking comment

pkg/api/types.go
AlexNPavel: Added

pkg/api/types.go
AlexNPavel: I had some new validation. I have updated this PR to add a lot more validation. It should now detect any collisions between bundle+index names and base and built images.

Something I noticed is that we don't seem to validate for collisions between built images and base images. Should we add validation for that (in a separate PR)?

pkg/steps/index_generator.go
AlexNPavel: We default it in `defaults.go`

AlexNPavel:
> What makes the index valid or invalid? Can you write an e2e test that depends on the image (`dependencies`) and uses `skopeo inspect` or something?

My idea is that I would
1. Start the registry server (`opm registry serve --database /database/index.db &`)
2. `wget` and `tar -xf` the `grpcurl` utility (https://github.com/fullstorydev/grpcurl/releases)
3. Verify that the bundles we want exist (this includes both the ones we built and any from a `base_index`) using `grpcurl`:
```
grpcurl -plaintext -d '{"pkgName":"metering-ocp","channelName":"test-channel","csvName":"metering-operator.v4.6.0"}' localhost:50051 api.Registry/GetBundle
```

AlexNPavel:
Docs PR: https://github.com/openshift/ci-docs/pull/107

pkg/api/types.go
stevekuznetsov: Wow, yeah we should, since they both target `pipeline:`

stevekuznetsov (approved):

PR #1751, opened 2021-03-05
steps: run scripts with /bin/sh

AlexNPavel:
The current method of running commands in either container or multistage
tests is essentially by running `/bin/bash -c ""#!/bin/bash\nset
-eu\nCommandProvidedByUser"`, where `CommandProvidedByUser` is command
that the user specified in their container or multistage test step's
command field. This has 2 problems:
- Requires `bash`. I am not sure if this has been complained about
  before, but ci-tools has. Specifically, ci-tools has not been able to
  test whether the index images generated for operators are generated
  correctly. Since the opm image used to serve the index is an alpine
  based image, there is no bash, only busybox (available at `/bin/sh`).
  This means the only testing our e2e can currently do is check whether
  the build step for the index succeeded or failed. Allowing a test to run
  in the generated index wouls allow us to check the contents of the index
  as well.
- It does not allow a user to specify their own shell. We currently have
  users set the shebang for their scripts in the step registry. However,
  this has no effect. This is because the shebang is completely ignored by
  shells when the command is provided in string form. The shebang is only
  honored when the OS executes the command and chooses the shell to use.
  This means that the `#!/bin/bash` that the ci-operator sets for all
  commands as well as any shebang that a user sets are ignored. This could
  be a problem is a user wants to execute any non-bash compatible script
  (for example, `#!/bin/python3`).

The solution to this is to execute the script directly. The simplest way
is to `echo` the user's script to disk, mark that as executable, and
then execute it. To prevent the shell from trying to execute code in the
provided command during the `echo`, we can encode the original command
as base64. This result in a command like this:
```
/bin/sh -c "echo B64== | base64 -d > test.sh; chmod +x ./test.sh; ./test.sh
```

This allows us to run in simpler container images lacking `bash` as well
as respect script shebangs.

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -41,8 +41,6 @@ const (
 	CliMountPath = "/cli"
 	// CliEnv if the env we use to expose the path to the cli
 	CliEnv = "CLI_DIR"
-	// CommandPrefix is the prefix we add to a user's commands
-	CommandPrefix = "#!/bin/bash\nset -eu\n"
 )
 
 var envForProfile = []string{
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
I think I'd much rather see us create a ConfigMap and mount the file into the contaiiner, then run it.
/hold

AlexNPavel:
/close

Will handle via configmaps: https://github.com/openshift/ci-tools/pull/1776

PR #1752, opened 2021-03-05
Vault secret manager: Add request logging

alvaroaleman:

### DIFF
--- a/cmd/vault-secret-collection-manager/main.go
+++ b/cmd/vault-secret-collection-manager/main.go
@@ -13,12 +13,12 @@ import (
 	"time"
 
 	"github.com/julienschmidt/httprouter"
-	uuid "github.com/satori/go.uuid"
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


cmd/vault-secret-collection-manager/main.go
```diff
@@ -47,7 +47,9 @@ func parseOptions() (*option, error) {
 }
 
 func main() {
+	version.Name = "vault-secret-collection-manager"
 	logrusutil.ComponentInit()
+	logrus.SetLevel(logrus.DebugLevel)
```
stevekuznetsov, line 52: Do we not have a default flag for this?

stevekuznetsov (approved):

cmd/vault-secret-collection-manager/main.go
alvaroaleman: I don't think so, in the prow components this comes from the config

PR #1753, opened 2021-03-05
Vault secret collection manager: Add backend for Deletion

alvaroaleman:

### DIFF
--- a/cmd/vault-secret-collection-manager/main.go
+++ b/cmd/vault-secret-collection-manager/main.go
@@ -133,6 +133,9 @@ func (c *idNameCache) set(name, id string) {
 
 func (m *secretCollectionManager) mux() *httprouter.Router {
 	router := httprouter.New()
+	// Do not redirect something like POST secretcollection/ where someone tried to
+	// create a nameless secret collection to secretcollection
+	router.RedirectTrailingSlash = false
 	router.GET("/", simpleLoggingWrapper(redirectHandler("/secretcollection?ui=true")))
 	router.GET("/style.css", simpleLoggingWrapper(staticFileHandler(styleCSS, "text/css")))
 	router.GET("/index.js", simpleLoggingWrapper(staticFileHandler(indexJS, "text/javascript")))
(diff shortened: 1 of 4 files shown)


cmd/vault-secret-collection-manager/main.go
```diff
@@ -133,13 +133,17 @@ func (c *idNameCache) set(name, id string) {
 
 func (m *secretCollectionManager) mux() *httprouter.Router {
 	router := httprouter.New()
+	// Do not rediurect something like POST secretcollection/ where someone tried to
```
stevekuznetsov, line 136:
```suggestion
	// Do not redirect something like POST secretcollection/ where someone tried to
```

cmd/vault-secret-collection-manager/main_test.go
```diff
@@ -174,6 +174,27 @@ func TestSecretCollectionManager(tt *testing.T) {
 			}},
 			expectedVaultPolicies: []string{"default", "secret-collection-manager-managed-mine-alone", "root"},
 		},
+		{
+			name:               "User-2 tries to delete, 404",
```
stevekuznetsov, line 178: These names mean nothing without context, something like this would be better:

```
user gets 404 when deleting group they are not a part of
```

cmd/vault-secret-collection-manager/main_test.go
```diff
@@ -193,8 +214,12 @@ func TestSecretCollectionManager(tt *testing.T) {
 			if err != nil {
 				t.Fatalf("failed to read response body: %v", err)
 			}
-			if diff := cmp.Diff(tc.expectedBody, string(bodyData)); diff != "" {
-				t.Errorf("expected body differs from actual: %s", diff)
+
+			// Do not check response body on errors
```
stevekuznetsov, line 218: Why?

pkg/vaultclient/vaultclient.go
```diff
@@ -57,6 +57,34 @@ func (v *VaultClient) ListKV(path string) ([]string, error) {
...
+		children, err := v.ListKV(path)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list %s: %w", path, err)
+		}
+		for _, child := range children {
+			if strings.HasSuffix(child, "/") {
+				// staticcheck complains `this result of append is never used, except maybe in other appends` but
+				// we do use it in the initial range.
+				// nolint: staticcheck
+				paths = append(paths, child)
```
stevekuznetsov, line 73: is it valid to change the slice you're looping over?

cmd/vault-secret-collection-manager/main_test.go
alvaroaleman: Added an explanation

pkg/vaultclient/vaultclient.go
alvaroaleman: I wondered the same, hence the test. Seems to be working fine

PR #1756, opened 2021-03-08
Vault secret manager: Add frontend for deletions

alvaroaleman:
This change adds the frontend part to allow secret collection deletion, ref: https://issues.redhat.com/browse/DPTP-1891
Also includes a couple of refactoring:
* Rename the modal, its main content div and the error div to clarify that it's not only used to create a secret collection but
* Move hiding of the modal into a func that will hide the modal, the error and content sub-div and all children of the latter, so that funcs that put something into there do not first need to do that (and know about everything else that could put content in there)
* Fix the http requests to not always create a modal error when there is any body and to use exception handling for both "request fails entirely" and "request didn't get a 2XX status", rather than only for the first
* Fix `fetchAndRenderSecretCollections` to be able to cope with no secret collections being there, rather than simply doing nothing in that case which would keep deleted collections on the screen
* Consistently use candy crush theming for buttons rather than combining it with some Windows 98 styles
* Handle enter key press and use it to confirm secret collection creation
* Move code around so things only get used after they were defined

Preview is here: https://vault-secret-collection-manager-ci-tools-1756.apps.ci.l2s4.p1.openshiftapps.com

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -121,7 +121,7 @@ install:
 .PHONY: install
 
 cmd/vault-secret-collection-manager/index.js: cmd/vault-secret-collection-manager/index.ts
-	tsc cmd/vault-secret-collection-manager/index.ts
+	tsc --lib ES2015,dom cmd/vault-secret-collection-manager/index.ts
 
 # Install Go binaries to $GOPATH/bin.
 # Set version and name variables.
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Does not compile, which I think also makes the preview fail>

stevekuznetsov:
Is there any way to make the text input highlighted when I hit "new collection" so I can just hit the button and start typing immediately?

PR #1763, opened 2021-03-09
vendor: bump k8s.io/test-infra

stevekuznetsov:
We need this to handle the new config fields for [DPTP-1962](https://issues.redhat.com/browse/DPTP-1962)


stevekuznetsov:
Good lord why on _earth_ did we duplicate the censoring tests from upstream here. We've duplicated them _three times_ by my count. You don't need to write unit tests to test that an upstream library does the right thing when the library has literally the same tests. It's insanity.

petr-muller:
LGTM but needs `make generate`

/retitle vendor: bump k8s.io/test-infra

PR #1764, opened 2021-03-09
fix the go:embed wrong file reference

droslean:


petr-muller:
@droslean IMO we can close this as discussed on Slack
/cloes

PR #1766, opened 2021-03-09
ci-operator: censor all secret data in sidecar

stevekuznetsov:
The Prow pod utilities can now censor secret data before uploading
anything to the cloud, both in plain-text files and in archives that the
user creates during a test run. We know that the set of sensitive data
available to test Pods for a job is limited to the set of Secrets in the
Namespace that ci-operator created for the job, so by mounting them all
into the test Pods we can effectively guard against all secret leaks in
the future.

---

/assign @petr-muller @alvaroaleman @bbguimaraes

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -343,7 +343,7 @@ func removeFile(podClient PodClient, ns, name, containerName string, paths []str
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


stevekuznetsov:
This is the last pull to implement DPTP-1962

petr-muller (approved):
LGTM but needs rebase

PR #1771, opened 2021-03-10
Add the OO_BUNDLE env var to the optional_operators step

dirgim:


petr-muller:
/hold

I'm not sure if this is that easy, this can interact with ci-op building bundles itself (ci-op does not assume there is a single bundle). I will need to look more closely at this.

pkg/steps/optional_operators.go
```diff
@@ -63,6 +67,9 @@ func resolveOptionalOperator(parameters getter) (*optionalOperator, error) {
 		*valuePointer = value
 	}
 
+	if oo.Bundle, err = parameters.Get(OOBundle); err != nil {
+		oo.Bundle = ""
+	}
```
petr-muller, line 72: Hm, from the changes I saw in the registry I'd say the bundle would be required and not optional, no? It's consumed by the new scorecard step which would be always present, so withtout a bundle the test would always fail....

petr-muller (approved):
/hold

LGTM and should not interfere with existing OO code. I have a question on the required-ness of the bundle param but if you're sure the param can be optional feel free to merge.

pkg/steps/optional_operators.go
dirgim: That's a good point - adding the change to make it required.

dirgim: @petr-muller I made the changes to have the bundle required, ptal

petr-muller (approved):

PR #1772, opened 2021-03-10
Vault secret collection manager: Add edit members functionality

alvaroaleman:

### DIFF
--- a/cmd/vault-secret-collection-manager/docker-compose.yaml
+++ b/cmd/vault-secret-collection-manager/docker-compose.yaml
@@ -34,9 +34,11 @@ services:
       # Need a second backend to be able to test
       # the interaction with aliases
       vault auth enable userpass
-      # User must exist
-      vault write auth/userpass/users/admin password=false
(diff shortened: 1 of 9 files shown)


alvaroaleman:
Preview is at https://vault-secret-collection-manager-ci-tools-1772.apps.ci.l2s4.p1.openshiftapps.com

stevekuznetsov:
How easy would it be to switch the two columns in the UI? In my experience it's most common to have the flow be from left to right, so the right side is the list of members and you pull people into or out of the right side

stevekuznetsov (approved):
Testing the UI out looks good
/hold

alvaroaleman:
>  How easy would it be to switch the two columns in the UI? In my experience it's most common to have the flow be from left to right, so the right side is the list of members and you pull people into or out of the right side

Updated, however I think it is no difference in term of untuitivity because one of "Add a new member" and "Remove an existing member" will always be from right to left and thus unintuitive.

PR #1773, opened 2021-03-10
ci-operator: remove support for legacy promotion

stevekuznetsov:
Nobody uses this path any longer, we do not need to support it.


stevekuznetsov:
Part of DPTP-2051

PR #1774, opened 2021-03-10
ci-operator: unite the logic for choosing images to promote

stevekuznetsov:
There's no reason that the logic for choosing which images are promoted
and how they are mapped to the outputs should vary between different
parts of our codebase.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

Depends on #1773 
Part of [DPTP-2051](https://issues.redhat.com/browse/DPTP-2051)
/assign @alvaroaleman @hongkailiu

PR #1775, opened 2021-03-10
ci-operator: always promote the build cache

stevekuznetsov:
Lots of valuable work is done when binaries are built from the source
code in a test. We can promote the resulting image to a build cache
repository and use it as the `build_root` for subsequent jobs (if the
underlying `build_root` that we originate from has not changed) and
thereby save enormous amounts of time in both incremental cloning and
incremental compilation.

Depends on #1774 
Part of [DPTP-2051](https://issues.redhat.com/browse/DPTP-2051)

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -272,11 +272,10 @@ func fromConfig(
 		if pushSecret == nil {
 			return nil, nil, errors.New("--image-mirror-push-secret is required for promoting images")
 		}
-		cfg, err := promotionDefaults(config)
-		if err != nil {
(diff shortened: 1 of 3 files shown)


smarterclayton:
Seems reasonable, is there a test run anywhere that shows this correctly inheriting the build cache (i.e. can i spot inspect this, or have we not got it all together)?

stevekuznetsov:
This is not all together - this pull just gets the images out for promotion, the next step will be to use them conditionally to actually make use of the cache. We'll need to make sure that changes to the "real" build root mean we don't use the cache (underlying builder image).

PR #1776, opened 2021-03-10
[DPTP-2057] multi_stage: write commands.sh to configmap

AlexNPavel:
Instead of running commands directly with `bash` when running multistage
tests, `ci-operator` should instead execute the script directly so that
the shebang specifying what shell to use is followed (it is ignored when
run directly by the shell) and to remove the multistage test steps'
dependance on `bash`, which does not exist in certain images, usually
alpine based.

This PR now runs commands by creating a configmap and mounting it to a
known location in the pod. Then, the script is executed by doing
`/bin/sh -c /path/to/script.sh`. At a bare minimum, all images should
have a `/bin/sh`, and in order to continue supporting injecting the cli
binary into images, the ci-operator needs to execute commands via
`/bin/sh` so the `$PATH` env var can be updated to include the path to
the cli before running the user's script.

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -46,6 +46,8 @@ const (
 	CliEnv = "CLI_DIR"
 	// CommandPrefix is the prefix we add to a user's commands
 	CommandPrefix = "#!/bin/bash\nset -eu\n"
+	// CommandScriptMountPath is where we mount the command script
+	CommandScriptMountPath = "/var/run/configmaps/ci.openshift.io/multi-stage"
 )
 
 var envForProfile = []string{
@@ -131,6 +133,9 @@ func (s *multiStageTestStep) run(ctx context.Context) error {
 	}
 	if err := s.createCredentials(); err != nil {
 		return fmt.Errorf("failed to create credentials: %w", err)
+	}
+	if err := s.createCommandConfigMaps(ctx); err != nil {
+		return fmt.Errorf("failed to create command configmap: %w", err)
 	}
 	if err := s.setupRBAC(ctx); err != nil {
 		return fmt.Errorf("failed to create RBAC objects: %w", err)
@@ -333,6 +338,36 @@ func (s *multiStageTestStep) createCredentials() error {
 	return nil
 }
 
+func (s *multiStageTestStep) createCommandConfigMaps(ctx context.Context) error {
+	log.Printf("Creating multi-stage test commands configmap for %q", s.name)
+	data := make(map[string]string)
+	for _, step := range append(s.pre, append(s.test, s.post...)...) {
+		data[step.As] = step.Commands
+	}
+	name := commandConfigMapForTest(s.name)
+	yes := true
+	commands := &coreapi.ConfigMap{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      name,
+			Namespace: s.jobSpec.Namespace(),
+		},
+		Data:      data,
+		Immutable: &yes,
+	}
+	// delete old command configmap if it exists
+	if err := s.client.Delete(ctx, commands); err != nil && !kerrors.IsNotFound(err) {
+		return fmt.Errorf("could not delete command configmap %s: %w", name, err)
+	}
+	if err := s.client.Create(ctx, commands); err != nil {
+		return fmt.Errorf("could not create command configmap %s: %w", name, err)
+	}
+	return nil
+}
+
+func commandConfigMapForTest(testName string) string {
+	return fmt.Sprintf("%s-commands", testName)
+}
+
 func (s *multiStageTestStep) runSteps(
 	ctx context.Context,
 	steps []api.LiteralTestStep,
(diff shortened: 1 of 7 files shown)


pkg/steps/multi_stage.go
```diff
@@ -45,10 +45,15 @@ const (
 	CliMountPath = "/cli"
 	// CliEnv if the env we use to expose the path to the cli
 	CliEnv = "CLI_DIR"
-	// CommandPrefix is the prefix we add to a user's commands
-	CommandPrefix = "#!/bin/bash\nset -eu\n"
+	// CommandScriptFilename is the filename for the command script
+	CommandScriptFilename = "commands.sh"
```
stevekuznetsov, line 49: Can we get this from the step registry? Will non-shell files be unhappy about being called `.sh` ? Does `commands.sh` run in Python fine?

pkg/steps/multi_stage.go
```diff
@@ -334,6 +342,37 @@ func (s *multiStageTestStep) createCredentials() error {
...
+		}
+		if err := s.client.Create(context.TODO(), toCreate[name]); err != nil {
+			return fmt.Errorf("could not create command configmap: %w", err)
+		}
+	}
+	return nil
+}
+
+func commandConfigMapForStep(testName, stepName string) string {
+	return fmt.Sprintf("%s-%s-command-script", testName, stepName)
```
stevekuznetsov, line 373: Will this ever be longer than the maximum length of a k8s object name?

pkg/steps/multi_stage.go
```diff
@@ -420,7 +459,7 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 		// the grace period for the Pod to be just larger than the grace period
 		// for the process, assuming an 80/20 distribution of work.
 		terminationGracePeriodSeconds := p(int64(gracePeriod.Seconds() * 5 / 4))
-		pod, err := generateBasePod(s.jobSpec, name, multiStageTestStepContainerName, []string{"/bin/bash", "-c", CommandPrefix + step.Commands}, image, resources, artifactDir, s.jobSpec.DecorationConfig, s.jobSpec.RawSpec(), secretVolumeMounts)
+		pod, err := generateBasePod(s.jobSpec, name, multiStageTestStepContainerName, []string{"/bin/sh", "-c", CommandExec}, image, resources, artifactDir, s.jobSpec.DecorationConfig, s.jobSpec.RawSpec(), secretVolumeMounts)
```
stevekuznetsov, line 462: Why are we using a shell to execute this? Shouldn't we just ensure that the files have a shebang and are executable?

stevekuznetsov (commented):
Mostly looks ok - we will want this to be opt-in at first, though.

pkg/steps/multi_stage.go
AlexNPavel: As long as the shebang points to a valid executable, the filename doesn't matter.

pkg/steps/multi_stage.go
AlexNPavel: If a user injects the CLI into their test, we need to export a new path that includes the CLI_DIR. Using `/bin/sh` allows us to run the export command to update the env var before running the script: https://github.com/openshift/ci-tools/pull/1776/files#diff-da42b74c690675134a34fba17070b2705bdf546b16b6e84e06ef9440f637a166R687-R688

pkg/steps/multi_stage.go
AlexNPavel: Quite unlikely, but I assume it's possible. I could change the name to `testName-stepName`, which is the name used for the pod as well, to slightly shorten it.

AlexNPavel:
PR has been updated. There are 2 main changes:
1. Instead of moving all steps over to the new mode immediately, it is a non-default optional setting configured via a `run_as_script` bool configured in the step reference.
2. Instead of making one configmap per step, one configmap is used for the entire test and each step's command is a separate entry in the configmap's data field.

pkg/steps/multi_stage.go
stevekuznetsov: Can we move that to the `entrypoint-wrapper` like I did for `$HOME` handling (in a separate PR, first)? The less weird Bash mangling we have the better.

pkg/steps/multi_stage.go
```diff
@@ -334,6 +344,34 @@ func (s *multiStageTestStep) createCredentials() error {
...
 }
 
+func (s *multiStageTestStep) createCommandConfigMaps() error {
+	log.Printf("Creating multi-stage test commands configmap for %q", s.name)
+	data := make(map[string]string)
+	for _, step := range append(s.pre, append(s.test, s.post...)...) {
+		data[step.As] = step.Commands
+	}
+	name := commandConfigMapForTest(s.name)
+	commands := &coreapi.ConfigMap{
```
stevekuznetsov, line 354: Can we make sure to mark this as immutable?

pkg/steps/multi_stage.go
```diff
@@ -334,6 +344,34 @@ func (s *multiStageTestStep) createCredentials() error {
...
+		ObjectMeta: meta.ObjectMeta{
+			Name:      name,
+			Namespace: s.jobSpec.Namespace(),
+		},
+		Data: data,
+	}
+	// delete old command configmap if it exists
+	if err := s.client.Delete(context.TODO(), commands); err != nil && !kerrors.IsNotFound(err) {
```
stevekuznetsov, line 362: Please use the context from `run()`

pkg/steps/multi_stage.go
```diff
@@ -420,7 +458,12 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 		// the grace period for the Pod to be just larger than the grace period
 		// for the process, assuming an 80/20 distribution of work.
 		terminationGracePeriodSeconds := p(int64(gracePeriod.Seconds() * 5 / 4))
-		pod, err := generateBasePod(s.jobSpec, name, multiStageTestStepContainerName, []string{"/bin/bash", "-c", CommandPrefix + step.Commands}, image, resources, artifactDir, s.jobSpec.DecorationConfig, s.jobSpec.RawSpec(), secretVolumeMounts)
+		var pod *coreapi.Pod
+		if step.RunAsScript != nil && *step.RunAsScript {
```
stevekuznetsov, line 462: Can we just factor out the script instead of duping the rest of this callsite?

test/e2e/multi-stage/config.yaml
```diff
@@ -15,7 +19,8 @@ tests:
       test:
         - as: produce-content
           commands: echo -n "message" > ${SHARED_DIR}/file
-          from: os
+          from: alpine
+          run_as_script: true
```
stevekuznetsov, line 23: nit: can we make this a dedicated test, please?

AlexNPavel:
/hold
until CLI path configuration is moved to entrypoint-wrapper in separate PR

AlexNPavel:
PR handling the CLI path via `entrypoint-wrapper`: https://github.com/openshift/ci-tools/pull/1802

stevekuznetsov (approved):
This looks good but will need a rebase on top of #1802 right?

/hold

stevekuznetsov (approved):
/hold cancel

PR #1777, opened 2021-03-11
ci-operator: allow overriding registry for promotions

stevekuznetsov:
This is a useful feature for the ci-chat-bot when working with multiple
PRs. It is not, however, actually useful to normal test flows, so we
document it as such and validate that checked-in configuration does not
set it.

---

Depends on #1775 
/assign @smarterclayton

### DIFF
--- a/pkg/steps/release/promote_test.go
+++ b/pkg/steps/release/promote_test.go
@@ -560,8 +560,34 @@ func TestGetImageMirror(t *testing.T) {
 
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			if actual, expected := getImageMirrorTarget(testCase.tags, testCase.pipeline), testCase.expected; !reflect.DeepEqual(actual, expected) {
+			if actual, expected := getImageMirrorTarget(testCase.tags, testCase.pipeline, "registry.ci.openshift.org"), testCase.expected; !reflect.DeepEqual(actual, expected) {
 				t.Errorf("%s: got incorrect ImageMirror mapping: %v", testCase.name, diff.ObjectDiff(actual, expected))
+			}
+		})
+	}
+}
+
+func TestRegistryDomain(t *testing.T) {
+	var testCases = []struct {
+		name     string
+		config   *api.PromotionConfiguration
+		expected string
+	}{
+		{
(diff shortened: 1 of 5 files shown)


smarterclayton:
I did not see anything here that made me think this wouldn't work for me.  So lgtm but I didn't understand the binary promotion part and i'm a bit rusty on the logic.    Did you want to add a unit test for registry_override?

stevekuznetsov:
@smarterclayton  binary thing is just that this is in a series of pulls and that is a pre-req - would rather not have to mess with this in two forks at once. There's a unit test already, no?

smarterclayton:
I didn't see a unit test for the bit that was added (setting it bypasses registry.ci.openshift.org and goes to whatever is specified), but maybe i missed it.

pkg/steps/release/promote_test.go
```diff
@@ -305,19 +294,48 @@ func TestGetImageMirror(t *testing.T) {
 					},
 				},
 			},
-			expected: map[string]string{"docker-registry.default.svc:5000/ci-op-y2n8rsh3/pipeline@sha256:bbb": "registry.ci.openshift.org/ci/name:a", "docker-registry.default.svc:5000/ci-op-y2n8rsh3/pipeline@sha256:ddd": "registry.ci.openshift.org/ci/name:c"},
+			expected: map[string]string{
+				"docker-registry.default.svc:5000/ci-op-y2n8rsh3/pipeline@sha256:bbb": "registry.ci.openshift.org/ci/a:latest",
+				"docker-registry.default.svc:5000/ci-op-y2n8rsh3/pipeline@sha256:ddd": "registry.ci.openshift.org/ci/c:latest",
```
smarterclayton, line 299: To clarify, the unit test i was expecting was here pulling from the promotion config but it looks like you killed the combined one.  That's fine

smarterclayton:
Does this need to be rebased?

stevekuznetsov:
@smarterclayton yeah I am not sure what GitHub is doing /shrug

stevekuznetsov:
There was a revert last week nobody told me about ... sigh

stevekuznetsov:
This now depends on https://github.com/openshift/ci-tools/pull/1792

PR #1778, opened 2021-03-11
ci-secret-generator: add a target parameter

bbguimaraes:
Unify the different modes of execution under a single parameter,
simplifying the interface and allowing for future additional targets.
Some backwards-compatible code is kept while CI jobs are changed.

A summary of the changes, more details for each target can be found in
the README:

- `--dry-run` becomes `--target file` (and continues being the default)
- `--validate-only` becomes `--target validate`
- `--dry-run false` becomes `--target bitwarden`

### DIFF
--- a/cmd/ci-secret-generator/README.md
+++ b/cmd/ci-secret-generator/README.md
@@ -1,13 +1,40 @@
 # CI-Secret-Generator
 
-This tool aims to automate the process of deployment of secrets to Bitwarden while also providing a platform to document the commands used to generate these secrets. 
+This tool automates the process of deployment of secrets while also providing a
+platform to document the commands used to generate them.
 
-## Args and config.yaml
+## Target
 
-The tool expects a configuration like the one below which specifies the mapping between the `itemName`+`attributeName`/`attachmentName`/`fieldName` and the command used to generate the secret.
-The output of the command is stored into Bitwarden as the contents of the field/attachment/password.
+Different output targets are supported, which can be specified with the
+`--target` command-line option.
 
-`password` is the only valid name which is accepted for an attribute
+### `validate`
+
+Verifies that the configuration is well-formed in respect to itself and the
+[`ci-secret-bootstrap`](../ci-secret-bootstrap) configuration, which has to be
+specified with the `--bootstrap-config` parameter.
+
+### `file`
+
+Executes the input commands and writes debug output to a local file.  A
+temporary file is used unless the `--output-file` parameter is specified.
+
+### `bitwarden`
+
+Currently the secret back end used in production.  Secrets are stored in
+field/attachment/password elements in Bitwarden depending on their size
+requirements.
+
+For attributes, `password` is the only valid name which is accepted.
+
+## Arguments and `config.yaml`
+
+The tool expects a configuration like the one below which specifies the mapping
+between the `itemName`+`attributeName`/`attachmentName`/`fieldName` and the
+command used to generate the content of the item.
+
+The output of the command is stored into the secret back end as the contents of
+the field/attachment/password.
 
 ```yaml
 - item_name: first_item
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-generator/README.md
```diff
@@ -1,13 +1,40 @@
...
-## Args and config.yaml
+## Target
 
-The tool expects a configuration like the one below which specifies the mapping between the `itemName`+`attributeName`/`attachmentName`/`fieldName` and the command used to generate the secret.
-The output of the command is stored into Bitwarden as the contents of the field/attachment/password.
+Different output targets are supported, which can be specified with the
+`--target` command-line option.
 
-`password` is the only valid name which is accepted for an attribute
+### `validate`
```
alvaroaleman, line 11: Not sure if `validate` really makes sense here, as it isn't mutually exclusive with the other targets

cmd/ci-secret-generator/main.go
```diff
@@ -367,14 +400,16 @@ func main() {
...
-	} else {
+	case targetBitwarden:
 		var err error
 		client, err = bitwarden.NewClient(o.bwUser, o.bwPassword, func(s string) {
 			secrets.Insert(s)
 		})
 		if err != nil {
 			logrus.WithError(err).Fatal("failed to create Bitwarden client")
 		}
+	default:
```
alvaroaleman, line 411: Let's not panic

cmd/ci-secret-generator/README.md
alvaroaleman: And the other targets are probably more of an output or a destination. Also, bitwarden/vault can be inferred from what credentials get passed. File probably as well?

bbguimaraes: > Not sure if validate really makes sense here, as it isn't mutually exclusive with the other targets

One of the many reasons for the change is that there are *two* arguments related to validation.  `--validate-only` is what is being substituted by `--target validate`, and is a mode of execution that stops after the configuration is loaded.  `--validate` is a separate argument (which is kept in this PR) that can be used with any other mode and validates the generator's configuration against `ci-secret-bootstrap`'s configuration.

Amusingly, `--validate-only --validate=false` is a possible combination which does only the implicit (generator) configuration validation when it is loaded.

bbguimaraes: > And the other targets are probably more of an output or a destination. Also, bitwarden/vault can be inferred from what credentials get passed. File probably as well?

That is roughly how the tool already works, but I tried to make the interface a bit more explicit.  A comparison of the before/after combination of arguments (partially described in the commit message):

Validate bootstrap/generator configuration files:

`ci-secret-generator --config "$config" --bootstrap-config "$bootstrap_config" --validate-only`

`ci-secret-generator --config "$config" --bootstrap-config "$bootstrap_config" --target validate`

Dry-run mode (the default), output to a file:

`ci-secret-generator --config "$config" --output-file /dev/stdout`

`ci-secret-generator --config "$config" --target file --output-file /dev/stdout`

Bitwarden:

`ci-secret-generator --config "$config" --dry-run=false --bw-user test --bw-password-path /dev/null`

`ci-secret-generator --config "$config" --target bitwarden --bw-user test --bw-password-path /dev/null`

Vault:

`N/A`

`ci-secret-generator --config "$config" --target vault --vault-addr http://localhost:8200 --vault-token /dev/null --vault-prefix prefix`

cmd/ci-secret-generator/main.go
bbguimaraes: Hmm, my impression was it is the `assert` equivalent in Go.  That is an unreachable path unless there is a logic error in the program, what would be a better alternative?

cmd/ci-secret-generator/README.md
alvaroaleman: yeah but all of file, bitwarden, and vault can be derived from other options, so the "more explicit" here means "redundant". As a user, I expect that after I pass credentials for Vault it will write to Vault and not also expect me to set another option. I agree that the `--validate-only` is a bit misleading, technically its more of a `--dont-write-to-secretstore`

cmd/ci-secret-generator/main.go
alvaroaleman: Not sure what exactly assert is, but the only places that are ok for panics are:
* func main
* func init
* Package-level variable declarations
* Testcode

It doesn't really make much of a difference there if the reason for the error is the programmers fault or someone elses IMHO.

bbguimaraes:
Superseded by the shared client approach in https://github.com/openshift/ci-tools/pull/1826.

/close

PR #1782, opened 2021-03-11
Add vault support to secret bootstrapper

alvaroaleman:
This is very straightforward except for the getUnusedBWItems check. That
check verifies that all of fields, attachments and password are used, if
present. As Vault just has a k-v map and doesn't know to what bitwarden
type a field maps we do the following:
* We store all keys vault has
* For all of field/attachment/password in the config file, we ask the
  secret provider if it has them
* This will make the vault provider mark those fields as in-use
* At the end, we add another check, "superfluous fields". The Vault
  implementation will use that to report all fields it didn't get asked
  about

This allows us to re-use all tests and thus be confident that it works
for Vault as well. The error message however is slightly different for
Vault.

It is recommended to review this PR with whitespace disabled.

Ref https://issues.redhat.com/browse/DPTP-1975

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -36,6 +37,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/bitwarden"
 	"github.com/openshift/ci-tools/pkg/kubernetes/pkg/credentialprovider"
 	"github.com/openshift/ci-tools/pkg/util"
+	"github.com/openshift/ci-tools/pkg/vaultclient"
 )
 
 type options struct {
@@ -51,6 +53,10 @@ type options struct {
 	logLevel        string
 	impersonateUser string
 	bwPassword      string
+	vaultToken      string
+	vaultTokenFile  string
+	vaultAddr       string
+	vaultPrefix     string
 
 	maxConcurrency int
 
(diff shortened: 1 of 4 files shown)


pkg/vaultclient/vaultclient.go
```diff
@@ -87,7 +87,10 @@ func (v *VaultClient) DestroyKVIrreversibly(path string) error {
 
 func (v *VaultClient) GetKV(path string) (*KVData, error) {
 	var response KVData
-	return &response, v.readInto(InsertDataIntoPath(path), &response)
+	if err := v.readInto(InsertDataIntoPath(path), &response); err != nil {
+		return nil, fmt.Errorf("failed to get item at path %q: %w", path, err)
+	}
+	return &response, nil
```
petr-muller, line 93: Did you just find this too smart or was it actually problematic in some way?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -87,37 +96,65 @@ func parseOptions() options {
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Errorf("cannot parse args: %q", os.Args[1:])
 	}
+	if o.vaultAddr == "" {
+		o.vaultAddr = os.Getenv("VAULT_ADDR")
+	}
+	o.vaultToken = os.Getenv("VAULT_TOKEN")
```
petr-muller, line 102: Should this be conditional too? At least the help says "defaults to the VAULT_TOKEN env var if unset"

petr-muller: Oh, I see, nevermind.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -87,37 +96,65 @@ func parseOptions() options {
...
+	var credentialsProviderConfigured []string
+	if (o.bwUser == "") != (o.bwPasswordPath == "") {
+		errs = append(errs, fmt.Errorf("--bw-user and --bw-password-path must be specified together"))
+	} else if o.bwUser != "" {
+		credentialsProviderConfigured = append(credentialsProviderConfigured, "bitwarden")
 	}
-	if o.bwPasswordPath == "" {
-		return fmt.Errorf("--bw-password-path is empty")
+
+	if vals := sets.NewString(o.vaultAddr, o.vaultTokenFile, o.vaultPrefix); len(vals) != 1 && len(vals) != 3 {
```
petr-muller, line 121: This is not doing exactly what it should as it will allow one of the values to be unset (one of the items in he set will be a `""`)

cmd/ci-secret-bootstrap/vault.go
```diff
@@ -0,0 +1,109 @@
...
+	"fmt"
+	"strings"
+	"time"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/openshift/ci-tools/pkg/vaultclient"
+)
+
+var _ secretStoreClient = &vaultSecretStoreWrapper{}
```
petr-muller, line 13: What's this for?

pkg/vaultclient/vaultclient.go
alvaroaleman: I wanted to have the path in the error message

cmd/ci-secret-bootstrap/vault.go
alvaroaleman: It checks that the &vaultsecretStoreWrapper impelements the secretStoreClient interface

cmd/ci-secret-bootstrap/main.go
alvaroaleman: Fixed

droslean:
/lgtm
/approve

/hold
LGTM from my side. @petr-muller Is anything left that you want to take a look at here?

petr-muller (approved):

PR #1784, opened 2021-03-12
use a watch to wait for pods instead of a tight poll

deads2k:
Updates the ci-operator to use standard kuberentes principle established in the upstream client (not controller-runtime) which have successfully scalable systems like
1. kube-apiserver RBAC
2. every apps controller
3. every controller in kube-controller-manager
4. all openshift-controllers
5. kubelet
6. scheduler
7. network plugins
8. csi drivers
9. more than a third of all operators

The backing library itself is well tested (see references above), so we should consider using the standard client library that is prove scalable and not a shim which has not been.

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -270,6 +270,15 @@ func waitForContainer(podClient PodClient, ns, name, containerName string) error
 
 	waitTimeout, _ := watchwait.ContextWithOptionalTimeout(ctx, 300*5*time.Second /*weird, but this is what it was*/)
 	_, err := watchwait.UntilWithSync(waitTimeout, lw, &corev1.Pod{}, podExistsPrecondition, waitForContainerStatus)
+
+	// TODO this looks like an attempt to have an informer.  Even with a punch through, this client would then miss external
+	//  modifications, so it's notion of current state is flawed.
+	// there is a side-effect based client, which appears to cache state based on a GET request and is apparently used
+	// as a side-cache by other parts of this system. While there are no tests catching this today, removing such side-effects
+	// to create an externally verifiable system is outside the scope of this PR.  I might recommend using an existing
+	// informer to drive such decisions, but this will place the current state into the client
+	_ = podClient.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: ns, Name: name}, &corev1.Pod{})
+
 	return err
 }
 
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -18,13 +18,20 @@ import (
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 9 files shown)


pkg/steps/artifacts.go
```diff
@@ -169,16 +176,23 @@ type PodClient interface {
 	WithNewLoggingClient() PodClient
 	Exec(namespace, pod string, opts *coreapi.PodExecOptions) (remotecommand.Executor, error)
 	GetLogs(namespace, name string, opts *coreapi.PodLogOptions) *rest.Request
+	RealClient() kubernetes.Interface
 }
 
 func NewPodClient(ctrlclient loggingclient.LoggingClient, config *rest.Config, client rest.Interface) PodClient {
-	return &podClient{LoggingClient: ctrlclient, config: config, client: client}
+	return &podClient{
```
alvaroaleman, line 183: This entirely breaks the loggingclient abstraction for pods

alvaroaleman:
> The backing library itself is well tested (see references above), so we should consider using the standard client library that is prove scalable and not a shim which has not been.

This is a useless comment and just bashing based on personal taste lol. The reason one "scales" and the other doesn't is that one uses a wach and the other doesn't.

pkg/steps/artifacts.go
deads2k: > This entirely breaks the loggingclient abstraction for pods

I doubt that it breaks a logging abstraction since the logging client was not touched.

deads2k:
> The backing library itself is well tested (see references above), so we should consider using the standard client library that is prove scalable and not a shim which has not been.

> This is a useless comment and just bashing based on personal taste lol. The reason one "scales" and the other doesn't is that one uses a wach and the other doesn't.


This list is provided to encourage confidence in the underlying library chosen here by referencing its heavy usage by clients which currently have good scaling qualities.  I think you astutely noticed that using watches allows for high scaling with less load, let's switch to using that mechanism with a well tested library (see the references I provided).

I don't believe the scaling properties of a watch API are a matter of personal taste.

pkg/steps/artifacts.go
alvaroaleman: That's the problem: https://github.com/openshift/ci-tools/blob/252a894afa981c86712afdf0bc826adc73cf5d81/pkg/steps/loggingclient/loggingclient.go#L43

pkg/steps/artifacts.go
```diff
@@ -194,46 +208,69 @@ func (c podClient) GetLogs(namespace, name string, opts *coreapi.PodLogOptions)
...
 func (c podClient) WithNewLoggingClient() PodClient {
 	return podClient{
 		LoggingClient: c.New(),
 		config:        c.config,
 		client:        c.client,
+		realClient:    c.realClient,
 	}
 }
 
-// Allow tests to accelerate time
```
cgwalters, line 205: Ah phew, glad we're not allowing this anymore.

pkg/steps/artifacts.go
deads2k: Thanks.  Added a commit to do a single GET to keep the cache as up to date as possible for future usage.

PR #1787, opened 2021-03-12
Import release: Create an immutable ConfigMap

alvaroaleman:
This reduces the load on the apiserver.


stevekuznetsov (approved):
Yikes, is there really no flag for this?

PR #1788, opened 2021-03-12
WIP Debug locking

alvaroaleman:


bparees:
the issue isn't the unlock itself, the issue is that the channel is closed while the code is trying to write to it, which triggers a panic that exits the function which trips the defer which then attempts to unlock, but the lock isn't held since the code released the lock prior to trying to write to the channel.  Unfortunately the stack masks the original panic.  See this recreate which reproduces the stack seen in my jobs:

https://play.golang.org/p/TmWfDMhbxnk

PR #1790, opened 2021-03-15
secretsyncer: tolerate conflicts without polluting logs

petr-muller:
`secretsyncer` can race with other thins that manage secrets (like ci-secret-bootstrap) and then it unnecessarily polluted logs:

```
failed to upsert secret: Operation cannot be fulfilled on secrets "cluster-secrets-packet": the object has been modified; please apply your changes to the latest version and try again
```
Part of this secret above is managed by bootstrap and part by the syncer.

### DIFF
--- a/pkg/controller/secretsyncer/secretsyncer.go
+++ b/pkg/controller/secretsyncer/secretsyncer.go
@@ -125,7 +125,11 @@ func (r *reconciler) Reconcile(ctx context.Context, request reconcile.Request) (
 	log := r.log.WithField("request", request.String())
 	err := r.reconcile(ctx, log, request)
 	if err != nil {
-		log.WithError(err).Error("Reconciliation failed")
+		if apierrors.IsConflict(err) {
+			log.WithError(err).Trace("Reconciliation failed because of conflicts only")
+		} else {
+			log.WithError(err).Error("Reconciliation failed")
+		}
 	} else {
 		log.Trace("Finished reconciliation successfully")
 	}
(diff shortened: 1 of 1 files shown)


pkg/controller/secretsyncer/secretsyncer.go
```diff
@@ -162,17 +165,28 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, encodedRe
...
+				conflictErrors = append(conflictErrors, err)
+			} else {
+				mirrorErrors = append(mirrorErrors, err)
+			}
+		}
 	}
 
-	return utilerrors.NewAggregate(mirrorErrors)
+	// If we *only* hit conflict errors return a "representative" sample of them to pass a IsConflict check
+	if len(conflictErrors) > 0 && len(mirrorErrors) == 0 {
```
petr-muller, line 186: This is not the prettyest way to do it. Ideally I'd like to track the "we only saw conflict errors" as a boolean and if true, then I'd wrap the aggregate in a new conflict error, but I have found no way how to easily construct an error that would pass `IsConflict`. The only way I found was to build a `StatusError` instance myself which I don't see is much better than what I have here.

stevekuznetsov:
Why do we want to do this on conflict instead of retrying? We did not achieve desired state, but a fresh fetch and update should do it.

petr-muller:
> Why do we want to do this on conflict instead of retrying? We did not achieve desired state, but a fresh fetch and update should do it.

That's already happening in an outer level - when a reconciler returns an error the request gets requeued, and from what our other controllers do I understand that's how controller-runtime controllers are supposed to handle retrying. The PR just makes the controller *not* log an error when we only encounter conflict errors and nothing else.

alvaroaleman (approved):

PR #1800, opened 2021-03-17
ci-operator: add support for operator upgrade tests

AlexNPavel:
This re-adds the content from #1748 and fixes a bug where a dependency on `pipeline:` was being erroneously created and breaking all tests that had an `operator` section. The fix is in the second commit of this PR. This is the message from that commit:

types: empty string is not a bundle image

This commit improves checking of the `IsPipelineImage` function by
identifying all index images instead of just unnamed ones and preventing
`IsBundleImage` from returning `true` for empty image names, which was
causing broken dependencies when the PR containing the previous commit
in this PR merged. Further, this commit prevents the MultiStage step's
run function from adding a dependency for the `from` image if `from` is
empty (which is why `IsBundleImage` was given an empty `imageName`
string in the first place).

/cc @petr-muller

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -167,14 +167,16 @@ func (s *multiStageTestStep) SubSteps() []api.CIOperatorStepDetailInfo {
 func (s *multiStageTestStep) Requires() (ret []api.StepLink) {
 	var needsReleaseImage, needsReleasePayload bool
 	for _, step := range append(append(s.pre, s.test...), s.post...) {
-		dependency := api.StepDependency{Name: step.From}
-		imageStream, name, explicit := s.config.DependencyParts(dependency)
-		if explicit {
-			ret = append(ret, api.LinkForImage(imageStream, name))
-		} else {
-			// if the user did not specify an explicit namespace for this image,
-			// it's likely coming from an imported release we need to wait for
-			needsReleaseImage = true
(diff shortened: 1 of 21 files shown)


stevekuznetsov:
Can we add an e2e test for this?

pkg/steps/multi_stage.go
```diff
@@ -167,14 +167,16 @@ func (s *multiStageTestStep) SubSteps() []api.CIOperatorStepDetailInfo {
...
 	for _, step := range append(append(s.pre, s.test...), s.post...) {
-		dependency := api.StepDependency{Name: step.From}
-		imageStream, name, explicit := s.config.DependencyParts(dependency)
-		if explicit {
-			ret = append(ret, api.LinkForImage(imageStream, name))
-		} else {
-			// if the user did not specify an explicit namespace for this image,
-			// it's likely coming from an imported release we need to wait for
-			needsReleaseImage = true
+		if step.From != "" {
```
stevekuznetsov, line 170: When is it valid for `step.From` to be empty?

AlexNPavel: When `step.FromImage` is set

AlexNPavel:
Added an e2e test that uses `from_image` in the optional operators config and would have triggered the error that we were experiencing before.

stevekuznetsov (approved):

PR #1802, opened 2021-03-18
multi_stage: update PATH for cli using entrypoint-wrapper

AlexNPavel:
Instead of modifying the script that a multistage step runs in order to
update the `PATH` to point to the copied CLI, `entrypoint-wrapper`
checks for the `CLI_DIR` env var and, if it exists, appends that to the
`PATH` for the wrapped entrypoint.

### DIFF
--- a/cmd/entrypoint-wrapper/main.go
+++ b/cmd/entrypoint-wrapper/main.go
@@ -28,6 +28,7 @@ import (
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 
+	"github.com/openshift/ci-tools/pkg/steps"
 	"github.com/openshift/ci-tools/pkg/util"
 )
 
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):

bbguimaraes:
E2E uses the production `entrypoint-wrapper` image.

We can break this into two PRs and merge them separately, a redundant entry in `PATH` is harmless.

petr-muller:
Or just `/override` if we know it's safe?

AlexNPavel:
The way I'll handle this is splitting into 2 PRs:
- This PR will have the changes to `entrypoint-wrapper`. I'll also update the `entrypoint-wrapper` integration script to test the new function.
- A PR after this one that removes the old CLI path injection from `pkg/steps/multistage`.

AlexNPavel:
Updated to only include the changes to the `entrypoint-wrapper` command and it's integration test

stevekuznetsov (approved):

PR #1803, opened 2021-03-19
ci-operator: label digest of image we built on top of

stevekuznetsov:
In the future, `ci-operator` will optionally re-use a `bin` image to get
access to the clone and build cache that is stored there after the image
is built. In order to be able to tell that a cached image is
out-of-date, we need to label the version of the root it's built on top
of and check that against the version of the root we resolved at
runtime.

part of [DPTP-2051](https://issues.redhat.com/browse/DPTP-2051)

### DIFF
--- a/pkg/steps/project_image.go
+++ b/pkg/steps/project_image.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"path"
 
 	coreapi "k8s.io/api/core/v1"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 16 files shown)


pkg/steps/project_image.go
```diff
@@ -83,10 +64,16 @@ func (s *projectDirectoryImageBuildStep) run(ctx context.Context) error {
...
 			},
 			Paths: []buildapi.ImageSourcePath{{
-				SourcePath:     fmt.Sprintf("%s/%s/.", workingDir, s.config.ContextDir),
+				SourcePath:     fmt.Sprintf("%s/.", path.Join(baseDir, contextDir)),
 				DestinationDir: ".",
 			}},
 		})
+	} else {
+		sourceTag = s.config.From
+	}
```
petr-muller, line 73: I *think* this would be slightly more readable in the following form:

```go
var sourceTag api.PipelineImageStreamTagReference = s.config.From
// Three options to set `sourceTag` to somethign different
...
if sourceTag != s.config.From { 
  // if we are changing the source, we need to make sure it's mounted in
}
```

petr-muller (commented):
Concept is sound, implementation looks good, one structure comment.

stevekuznetsov:
@petr-muller updated
@AlexNPavel PTAL as well

petr-muller:
/lgtm
/hold
Holding for Alex, merge at will

PR #1804, opened 2021-03-19
Conform to the new `secretutil` semantics

bbguimaraes:


stevekuznetsov (changes requested):
Please use test-infra secretutil.Censorer

/hold

bbguimaraes:
@stevekuznetsov: I'm not sure what you are suggesting.  This is merely a refactoring of existing code:

- `f7b492839 pkg/secrets: extract applyconfig censor` moves the censoring code in `applyconfig` to a shared package.  That code is already based on `logrusutil` and consequently on `secretutil`.
- `e3d2d10e3 Use shared censor` uses the shared package in `ci-secret-bootstrap` and `ci-secret-generator`, removing existing similar code.
- `1efab2987 ci-secret-bootstrap: censor Vault secrets` uses the shared package in `ci-secret-bootstrap` for the Vault client.  This was missed in both my `ci-secret-generator` work and the `ci-secret-bootstrap` PR.

bbguimaraes:
On deeper inspection, there is something wrong with the code that's being reused here.

I'll continue the discussion elsewhere.  The hold should remain.

pkg/secrets/censor.go
```diff
@@ -0,0 +1,45 @@
...
+// GetSecrets returns the current list of secrets to censor.
+// The return value should not be modified.
+func (c *DynamicCensor) GetSecrets() sets.String {
+	c.RLock()
+	defer c.RUnlock()
+	return c.secrets
+}
+
+// Formatter creates a new formatter to be used to filter output.
+func (c *DynamicCensor) Formatter(f logrus.Formatter) logrus.Formatter {
```
stevekuznetsov, line 43: We do not need another implementation of this. Why don't we (simply) just use a `secretutil.Censorer` ? I'll add another call upstream to update instead of overwrite secrets...

stevekuznetsov: Or just do this:

```go
type DynamicCensor struct {
	sync.RWMutex
	delegate secretutil.Censorer
	secrets sets.String
}

func (c *DynamicCensor) AddSecrets(s ...string) {
	c.Lock()
	defer c.Unlock()
	c.secrets.Insert(s...)
	c.delegate.Refresh(c.secrets.List()...)
}

func (c *DyanmicCensor) Censor(input *[]byte) {
	c.delegate.Censor(input)
}
```

That's all you need here.

cmd/applyconfig/applyconfig.go
```diff
@@ -459,30 +458,13 @@ func fileFilter(info os.FileInfo, path string) (bool, error) {
...
 var (
-	secrets *secretGetter
+	censor secrets.DynamicCensor
 )
 
 func init() {
-	secrets = &secretGetter{secrets: sets.NewString()}
-	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, secrets.getSecrets))
+	censor = secrets.NewDynamicCensor()
+	logrus.SetFormatter(censor.Formatter(logrus.StandardLogger().Formatter))
```
stevekuznetsov, line 467: Why are we making our own call to this? We do `LogrusInit` pretty much everywhere else, and it's applicable here, too.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -798,22 +798,18 @@ func main() {
 		logrus.Infof("the config file %s has been validated", o.configPath)
 		return
 	}
-	secrets := sets.NewString()
-	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, func() sets.String {
-		return secrets
-	}))
+	censor := secrets.NewDynamicCensor()
+	logrus.SetFormatter(censor.Formatter(logrus.StandardLogger().Formatter))
```
stevekuznetsov, line 802: This should also use the shared init.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -508,7 +509,7 @@ func TestCompleteOptions(t *testing.T) {
 				}
 				sort.Strings(actualClusters)
 				equal(t, "clusters", tc.expectedClusters, actualClusters)
-				equal(t, "some set", sets.NewString("topSecret"), secrets)
+				equal(t, "some set", []string{"topSecret"}, censor.GetSecrets())
```
stevekuznetsov, line 512: I don't think this is that useful of a test

cmd/ci-secret-generator/main.go
```diff
@@ -313,14 +313,12 @@ func updateSecrets(bwItems []bitWardenItem, bwClient bitwarden.Client) error {
 func main() {
 	// CLI tool which does the secret generation and uploading to bitwarden
 	o := parseOptions()
-	secrets := sets.NewString()
-	logrus.SetFormatter(logrusutil.NewCensoringFormatter(logrus.StandardLogger().Formatter, func() sets.String {
-		return secrets
-	}))
+	censor := secrets.NewDynamicCensor()
+	logrus.SetFormatter(censor.Formatter(logrus.StandardLogger().Formatter))
```
stevekuznetsov, line 317: Shared init please

pkg/secrets/censor.go
```diff
@@ -0,0 +1,49 @@
...
+func (c *DynamicCensor) AddSecrets(s ...string) {
+	c.Lock()
+	defer c.Unlock()
+	c.secrets.Insert(s...)
+	c.ReloadingCensorer.Refresh(c.secrets.List()...)
+}
+
+// GetSecrets returns the current list of secrets to censor.
+// For informational purposes only.
+func (c *DynamicCensor) GetSecrets() []string {
```
stevekuznetsov, line 40: This is only used in tests, let's delete it

stevekuznetsov:
@bbguimaraes note on the `GetSecrets()` call and the test - this is exposing internal state of an object in a public interface for no reason other than to test, right? A test should exist at the _unit_ level that `AddSecret()` adds to the set inside, but for consumers of the interface we should only test the actual interface of the thing - that adding new secrets changes the censoring mechanisms.

cmd/applyconfig/applyconfig.go
```diff
@@ -545,6 +528,7 @@ func detectDryRunMethod(kubeconfig, context, username string) dryRunMethod {
 }
 
 func main() {
+	logrusutil.ComponentInit()
```
stevekuznetsov, line 531: Sorry, I should have said this before, too -- there is a subtle bug here in that `init()` fires first, so this overwrites that formatter.

IMHO we should not use `init()` or global `var`s here - it does not seem to have any utility over doing this in `main()`. A minimal fix would be to call `ComponentInit()` in `init()` and I can follow-up to stop using `init()` and globals.

bbguimaraes: Oh, good, I missed that.  Updated.

stevekuznetsov (approved):

bbguimaraes:

PR #1808, opened 2021-03-22
ci-operator: optionally use build cache for the root

stevekuznetsov:
ci-operator: validate that build_root ISTags are fully formed

---

ci-operator: clean up build root step generation

There's no place where any of the weird "test-base" stuff is used and
the previous commit validates as much. I'm not sure what it was for and
it's not documented anywhere. I removed that handler and fixed the git
base / project build options to pass through the full struct, instead of
cherry-picking some options and silently dropping the rest.

---

ci-operator: optionally use build cache for the root

When the user asks us to, we can determine if the build cache we have
published for the repo is up-to-date. If it is, we should use it for our
build root instead of the build root that would otherwise be used. In
cases where we can't resolve it or the cache is stale, fall back to the
original root.

---

need to think about how this might be e2e tested or it's worth it

### DIFF
--- a/pkg/api/helper/imageextraction.go
+++ b/pkg/api/helper/imageextraction.go
@@ -35,6 +35,9 @@ func TestInputImageStreamTagsFromResolvedConfig(cfg api.ReleaseBuildConfiguratio
 	imageStreamTagReferenceMapIntoMap(cfg.BaseRPMImages, result)
 	if cfg.BuildRootImage != nil && cfg.BuildRootImage.ImageStreamTagReference != nil {
 		insert(*cfg.BuildRootImage.ImageStreamTagReference, result)
+		if cfg.InputConfiguration.BuildRootImage.UseBuildCache {
+			insert(api.BuildCacheFor(cfg.Metadata), result)
+		}
 	}
 
 	var errs []error
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -1,6 +1,8 @@
 package defaults
 
 import (
+	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io/ioutil"
@@ -9,11 +11,15 @@ import (
 	"strings"
 
 	coreapi "k8s.io/api/core/v1"
+	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
 
+	"github.com/openshift/api/image/docker10"
+	imagev1 "github.com/openshift/api/image/v1"
 	templateapi "github.com/openshift/api/template/v1"
 	buildclientset "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	templateclientset "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
@@ -41,6 +47,7 @@ type inputImageSet map[api.InputImageTagStepConfiguration]struct{}
 // build, including defaulted steps, generated steps and
 // all raw steps that the user provided.
 func FromConfig(
+	ctx context.Context,
 	config *api.ReleaseBuildConfiguration,
 	jobSpec *api.JobSpec,
 	templates []*templateapi.Template,
@@ -75,10 +82,11 @@ func FromConfig(
 	}
 
 	podClient := steps.NewPodClient(client, clusterConfig, coreGetter.RESTClient())
-	return fromConfig(config, jobSpec, templates, paramFile, promote, client, buildClient, templateClient, podClient, leaseClient, &http.Client{}, requiredTargets, cloneAuthConfig, pullSecret, pushSecret, api.NewDeferredParameters(nil))
+	return fromConfig(ctx, config, jobSpec, templates, paramFile, promote, client, buildClient, templateClient, podClient, leaseClient, &http.Client{}, requiredTargets, cloneAuthConfig, pullSecret, pushSecret, api.NewDeferredParameters(nil))
 }
 
 func fromConfig(
+	ctx context.Context,
 	config *api.ReleaseBuildConfiguration,
 	jobSpec *api.JobSpec,
 	templates []*templateapi.Template,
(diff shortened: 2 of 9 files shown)


alvaroaleman:
@stevekuznetsov can this be split up into "Improvements to existing logic" and "Add support to use a build cache"? Can you also write three sentences high level description of how the build cache is intended to works?

pkg/defaults/defaults.go
```diff
@@ -397,9 +406,44 @@ func leasesForTest(s *api.MultiStageTestConfigurationLiteral) (ret []api.StepLea
...
+// the root image.
+func rootImageResolver(client loggingclient.LoggingClient, ctx context.Context) func(root, cache *api.ImageStreamTagReference) (*api.ImageStreamTagReference, error) {
+	return func(root, cache *api.ImageStreamTagReference) (*api.ImageStreamTagReference, error) {
+		cacheTag := &imagev1.ImageStreamTag{}
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: cache.Namespace, Name: fmt.Sprintf("%s:%s", cache.Name, cache.Tag)}, cacheTag); err != nil {
+			if kapierrors.IsNotFound(err) {
+				// no build cache, use the normal root
+				return root, nil
+			}
+			return nil, fmt.Errorf("could not resolve build cache image stream tag %s: %w", cache.ISTagName(), err)
```
petr-muller, line 420: Maybe we could be robust against random flakes here and just use the normal root. If the cluster is in serious trouble than something will fail later, but cache flakes should not fail a job.

pkg/defaults/defaults.go
```diff
@@ -397,9 +406,44 @@ func leasesForTest(s *api.MultiStageTestConfigurationLiteral) (ret []api.StepLea
...
+		}
+		metadata := &docker10.DockerImage{}
+		if len(cacheTag.Image.DockerImageMetadata.Raw) == 0 {
+			return nil, fmt.Errorf("could not fetch Docker image metadata build cache %s", cache.ISTagName())
+		}
+		if err := json.Unmarshal(cacheTag.Image.DockerImageMetadata.Raw, metadata); err != nil {
+			return nil, fmt.Errorf("malformed Docker image metadata on build cahce %s: %w", cache.ISTagName(), err)
```
petr-muller, line 428: typo: chace->cache

pkg/defaults/defaults.go
```diff
@@ -397,9 +406,44 @@ func leasesForTest(s *api.MultiStageTestConfigurationLiteral) (ret []api.StepLea
...
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: cache.Namespace, Name: fmt.Sprintf("%s:%s", cache.Name, cache.Tag)}, cacheTag); err != nil {
+			if kapierrors.IsNotFound(err) {
+				// no build cache, use the normal root
+				return root, nil
+			}
+			return nil, fmt.Errorf("could not resolve build cache image stream tag %s: %w", cache.ISTagName(), err)
+		}
+		metadata := &docker10.DockerImage{}
+		if len(cacheTag.Image.DockerImageMetadata.Raw) == 0 {
+			return nil, fmt.Errorf("could not fetch Docker image metadata build cache %s", cache.ISTagName())
```
petr-muller, line 425: Same here -> Is there a way how we easily discover problems like this without failing jobs for the user whose job would be fine if we didn't try to be smart?

stevekuznetsov:
@alvaroaleman broke out the tech debt into https://github.com/openshift/ci-tools/pull/1811

I'll add docs to the site on the build-root later but the basic idea is:

1. publish `bin` for repositories, labeling it with the version of `root` used to build it
2. use this published image instead of `root` when the published image is based on the same `root`

Ideally, this means that:

1. `clonerefs` runs in a pre-populated repo from the last push and only does incremental fetches
2. all of the `go` build cache exists as it needs to and we get incremental builds
3. any changes to the `root` will propagate as needed, allowing people to seamlessly update their root without caring about invalidating the cache on purpose

petr-muller (commented):
I was also thinking if there are some corner cases when people change their `build_root` configs but I was not able to come up with an idea of what could go wrong so it's probably fine :shrug: 

How does this work in the multi-cluster environment? Do I understand correctly that the `bin` image gets promoted in the standard promotion job (after each merge) which gets the image to the `app.ci` registry. What then distributes these images to build farms?

stevekuznetsov:
@petr-muller 

> I was also thinking if there are some corner cases when people change

I don't think this is possible as we will only ever use the cache when it's based on the same root image digest as `root` would otherwise have been for that run.

> Do I understand correctly that

Yes.

> What then distributes these images to build farms?

That's what the `imageextraction.go` changes are for - that's the code that determines which images are required to be distributed across clusters for a given config.

pkg/defaults/defaults.go
stevekuznetsov: I thought about that, but I'm not sure that there's any difference between failing on a GET ISTag call for the cache or the root - both are against the same API server, so /shrug

pkg/defaults/defaults.go
stevekuznetsov: I thought about making this entirely best-effort but yes as you note it means that failures will be silent and hard to detect.

I also don't know that this really increases the burden on the job for success. We have three things that can fail here:

1. two GET ISTag calls
2. a JSON unmarshal

One of the two GET calls are calls we'd be doing anyway when we resolve `root`, so we only add the cost of potentially flaking here. I don't know that we really see flakes in the API server for our clusters, they are either up or not.

The JSON unmarhsal can never fail in a working system. I don't think it's possible for the build API to produce bad metadata here in the first place.

IMHO as-is gives us the best ratio of usefulness/discoverability/etc.

stevekuznetsov:
/hold

For https://github.com/openshift/ci-tools/pull/1811

pkg/api/helper/imageextraction.go
```diff
@@ -35,6 +35,9 @@ func TestInputImageStreamTagsFromResolvedConfig(cfg api.ReleaseBuildConfiguratio
 	imageStreamTagReferenceMapIntoMap(cfg.BaseRPMImages, result)
 	if cfg.BuildRootImage != nil && cfg.BuildRootImage.ImageStreamTagReference != nil {
 		insert(*cfg.BuildRootImage.ImageStreamTagReference, result)
+		if cfg.InputConfiguration.BuildRootImage.UseBuildCache {
```
stevekuznetsov, line 38: We need to land this to test it in rehearsal even :|

pkg/defaults/defaults.go
petr-muller: Makes sense!

petr-muller (approved):

stevekuznetsov:
Just rebased to remove the first two commits.

PR #1810, opened 2021-03-23
remove ipi-deprovision tool and its related image 

droslean:

This tool is deprecated. See [here](https://github.com/openshift/release/tree/master/core-services/ipi-deprovision). 
This PR also removes the script from the ipi-deprovision image.


images/ipi-deprovision/Dockerfile
```diff
@@ -5,18 +5,12 @@ ARG JQ_VERSION=1.6
```
alvaroaleman, line 5: Why do we need to keep this image here, shouldn't we simply make this a BuildConfig instead?

alvaroaleman: All imagestreamtags in the CI get automatically distributed to all build clusters already: https://github.com/openshift/release/blob/8568b711c7374cd765c75d3086c92140e21a49d8/clusters/app.ci/assets/dptp-controller-manager.yaml#L126

So if you just put it there, nothing more to do.

droslean: Also, there are 2 repo files that this image is using. I am not sure if we can inject them in the buildConfig.

droslean: Let's take this approach for now. I will open a card to completely remove that image and use a bc.

alvaroaleman: > Also, there are 2 repo files that this image is using. I am not sure if we can inject them in the buildConfig.

Yes, you can use `echo` in a `docker run` directive, I would assume a heredoc works as well. This image has literally nothing to do with this repo so I would very much prefer if we didn't build it on every PR here.

droslean: https://github.com/openshift/release/pull/17065

alvaroaleman (approved):

PR #1813, opened 2021-03-24
ci-operator: initialize a more intelligent logger

stevekuznetsov:
Default censoring that is provided by `sidecar` has one draw-back: the
set of secrets that must be censored is loaded at start-up and must be
known ahead of time in order to be provided to the process. Inside of
the `ci-operator`, we routinely touch secret values via the Kubernetes
API at runtime. This means that the total set of secret values we may
possibly leak grows over time. Due to this, we cannot solely rely on the
censoring implemenation that runs out-of-process.

Also, today the `ci-operator` uses the `log` library for logging, in
contrast to pretty must every other part of the codebase and larger
ecosystem (Prow, Boskos, etc...). Censoring for the `logrus` library is
implemented, tried, and tested in production for a long time. We are
better off by migrating to `logrus` over time in order to consolidate
our codebase.

We have a similar issue in the `ci-secret-bootstrap` tool and already
have the `DynamicCensorer` utlity to handle this case. This patch does
two things:

- create an initial implementation of a `logrus` logger for
  `ci-operator` that not only censors by default but also shows only the
  most important messages to users in stdout, while at the same time
  funneling more specifc/spammy messages to a file which an
  administrator can read after a job failed.
- wire through a client shim that records the values of Secrets that
  `ci-operator` touches in the Kuberntes API while it runs, in order to
  ensure that the censoring provided by the logger cannot leak anything
  sensitive

Follow-up PRs will begin to move usage of `log` to `logrus` in our
code-base.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
+	"github.com/sirupsen/logrus/hooks/writer"
 	"go.uber.org/zap/zapcore"
 
 	appsv1 "k8s.io/api/apps/v1"
@@ -46,6 +47,7 @@ import (
 	"k8s.io/klog/v2"
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	"k8s.io/test-infra/prow/config/secret"
+	"k8s.io/test-infra/prow/logrusutil"
 	"k8s.io/test-infra/prow/pod-utils/downwardapi"
 	"k8s.io/test-infra/prow/version"
 	utilpointer "k8s.io/utils/pointer"
@@ -72,6 +74,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/lease"
 	"github.com/openshift/ci-tools/pkg/load"
 	"github.com/openshift/ci-tools/pkg/results"
+	"github.com/openshift/ci-tools/pkg/secrets"
 	"github.com/openshift/ci-tools/pkg/steps"
 	"github.com/openshift/ci-tools/pkg/util"
 	"github.com/openshift/ci-tools/pkg/validation"
@@ -172,11 +175,23 @@ var (
 const CustomProwMetadata = "custom-prow-metadata.json"
 
 func main() {
+	censor, closer, err := setupLogger()
+	if err != nil {
+		logrus.WithError(err).Fatal("Could not set up logging.")
+	}
+	if closer != nil {
+		defer func() {
+			if err := closer.Close(); err != nil {
+				logrus.WithError(err).Warn("Could not close ci-operator log file.")
+			}
+		}()
+	}
 	// "i just doin't want spam"
 	klog.LogToStderr(false)
 	log.Printf("%s version %s", version.Name, version.Version)
 	flagSet := flag.NewFlagSet("", flag.ExitOnError)
 	opt := bindOptions(flagSet)
+	opt.censor = censor
 	if err := flagSet.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("failed to parse flags")
 	}
@@ -241,6 +256,38 @@ func main() {
 		os.Exit(1)
 	}
 	opt.Report()
+}
+
+// setupLogger sets up logrus to print all logs to a file and user-friendly logs to stdout
+func setupLogger() (*secrets.DynamicCensor, io.Closer, error) {
+	logrusutil.ComponentInit()
+	logrus.SetLevel(logrus.TraceLevel)
+	censor := secrets.NewDynamicCensor()
+	logrus.SetFormatter(logrusutil.NewFormatterWithCensor(logrus.StandardLogger().Formatter, &censor))
+	logrus.SetOutput(ioutil.Discard)
+	logrus.AddHook(&writer.Hook{
+		Writer: os.Stdout,
+		LogLevels: []logrus.Level{
+			logrus.InfoLevel,
+			logrus.WarnLevel,
+			logrus.ErrorLevel,
+			logrus.FatalLevel,
+			logrus.PanicLevel,
+		},
+	})
+	artifactDir, set := api.Artifacts()
+	if !set {
+		return &censor, nil, nil
+	}
+	verboseFile, err := os.Open(filepath.Join(artifactDir, "ci-operator.log"))
+	if err != nil {
+		return nil, nil, err
+	}
+	logrus.AddHook(&writer.Hook{
+		Writer:    verboseFile,
+		LogLevels: logrus.AllLevels,
+	})
+	return &censor, verboseFile, nil
 }
 
 type stringSlice struct {
@@ -320,6 +367,8 @@ type options struct {
 	cloneAuthConfig *steps.CloneAuthConfig
 
 	resultsOptions results.Options
+
+	censor *secrets.DynamicCensor
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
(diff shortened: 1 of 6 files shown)


pkg/steps/censoringclient/client.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (c *client) DeleteAllOf(ctx context.Context, obj ctrlruntimeclient.Object, opts ...ctrlruntimeclient.DeleteAllOfOption) error {
+	return c.upstream.DeleteAllOf(ctx, obj, opts...)
+}
+
+func (c *client) Watch(ctx context.Context, obj ctrlruntimeclient.ObjectList, opts ...ctrlruntimeclient.ListOption) (watch.Interface, error) {
+	return c.upstream.Watch(ctx, obj, opts...)
+}
+
+func (c *client) recordList(obj ctrlruntimeclient.ObjectList) {
```
stevekuznetsov, line 85: @deads2k could you please review from this line down?

alvaroaleman: The `apiutil.GVKForObjece` dance is not wrong but not needed, because the client will stop working entirely if the same go type is registered in more than one api group, so just type asserting is enough

bbguimaraes:
Nice.

/lgtm
/hold

for the other review.

pkg/steps/censoringclient/client.go
```diff
@@ -0,0 +1,138 @@
+package censoringclient
```
alvaroaleman, line 1: The name is IMHO pretty confusing, because the client doesn't censor anything, it just records what to censor. I can not immediately come up with a great name, but maybe something like `secretrecordingclient` might be better?

Also, one sentence of godoc explaining this would be great

cmd/ci-operator/main.go
```diff
@@ -243,6 +258,38 @@ func main() {
...
 }
 
+// setupLogger sets up logrus to print all logs to a file and user-friendly logs to stdout
+func setupLogger() (*secrets.DynamicCensor, io.Closer, error) {
+	logrusutil.ComponentInit()
+	logrus.SetLevel(logrus.TraceLevel)
+	censor := secrets.NewDynamicCensor()
+	logrus.SetFormatter(logrusutil.NewFormatterWithCensor(logrus.StandardLogger().Formatter, &censor))
+	logrus.SetOutput(ioutil.Discard)
+	logrus.AddHook(&writer.Hook{
```
alvaroaleman, line 268: Why do we need to discard the output and use a hook when the formatter already uses the dynamic censor?

cmd/ci-operator/main.go
```diff
@@ -243,6 +258,38 @@ func main() {
...
+			logrus.ErrorLevel,
+			logrus.FatalLevel,
+			logrus.PanicLevel,
+		},
+	})
+	artifactDir, set := api.Artifacts()
+	if !set {
+		return &censor, nil, nil
+	}
+	verboseFile, err := os.Open(filepath.Join(artifactDir, "ci-operator.log"))
```
alvaroaleman, line 282: Why, doesn't Prow give us this already?

pkg/steps/censoringclient/client.go
```diff
@@ -0,0 +1,138 @@
...
+func (c *client) DeleteAllOf(ctx context.Context, obj ctrlruntimeclient.Object, opts ...ctrlruntimeclient.DeleteAllOfOption) error {
+	return c.upstream.DeleteAllOf(ctx, obj, opts...)
+}
+
+func (c *client) Watch(ctx context.Context, obj ctrlruntimeclient.ObjectList, opts ...ctrlruntimeclient.ListOption) (watch.Interface, error) {
+	return c.upstream.Watch(ctx, obj, opts...)
+}
+
+func (c *client) recordList(obj ctrlruntimeclient.ObjectList) {
+	gvk, err := apiutil.GVKForObject(obj, c.Scheme())
```
deads2k, line 86: This assumes the type you have is registered in the scheme.  I didn't read to see if that is always true.

If it is possible that the type is not registered in the scheme, you may need to inspect the GVK using GetObjectKind to handle unstructured.

pkg/steps/censoringclient/client.go
```diff
@@ -0,0 +1,138 @@
...
+
+func (c *client) recordSecret(secret *v1.Secret) {
+	var values []string
+	for _, value := range secret.Data {
+		values = append(values, string(value))
+	}
+	for _, value := range secret.StringData {
+		values = append(values, value)
+	}
+	c.censor.AddSecrets(values...)
```
deads2k, line 125: recall the secrets in annotations.  See https://github.com/openshift/oc/blob/master/pkg/cli/admin/inspect/secret.go#L76-L91 for instance.

pkg/steps/censoringclient/client.go
stevekuznetsov: IIRC David said if it was `unstructured.Unstructured` we needed it?

cmd/ci-operator/main.go
stevekuznetsov: This is the way that `logrus` wants it done lol, discard the default output and then hook it into the spots you want. If you didn't do this, the default output (to stdout) would go with Trace level and spam. This is unfortunately their architecture for this /shrug

cmd/ci-operator/main.go
stevekuznetsov: I don't think Prow has any "user-facing" components like `ci-operator`. Perhaps there's something like this in `kubetest`?

pkg/steps/censoringclient/client.go
alvaroaleman: Well the code doesn't handle unstructured.Unstructured in any way, as it will just return if the object isn't a `Secret{,List}`. It seems a tad overkill to make it handle that to me, because I doubt we will ever request a Secret through an unstructured. If we really want to handle that case, we can just type assert to unstructured.Unstructured as well and then use the TypeMeta to check if this is for a Secret.

stevekuznetsov: Makes sense - I do not understand in what cases we'd be getting one of those. Would we have to opt into it somehow on the client side?

stevekuznetsov:
@alvaroaleman @deads2k updated for review comments.

pkg/steps/censoringclient/client.go
alvaroaleman: well, if our code did something like
```
u := &unstructured.Unstructured{}
u.SetApiVersion("v1")
u.SetKind("Secret")
client.Get(ctx, someSecretName, u)
```

But I don't see any reason why we would do that instead of using a Secret directly.

stevekuznetsov:
Sigh, you trust GoLand to manage the imports ,,,

pkg/steps/secretrecordingclient/client.go
```diff
@@ -0,0 +1,132 @@
...
+
+func (c *client) recordSecret(secret *v1.Secret) {
+	var values []string
+	for _, value := range secret.Data {
+		values = append(values, string(value))
+	}
+	for _, value := range secret.StringData {
+		values = append(values, value)
+	}
+	for _, key := range []string{"openshift.io/token-secret.value", "kubectl.kubernetes.io/last-applied-configuration"} {
```
alvaroaleman, line 114: wat, how is the value of the `kubectl.kubernetes.io/last-applied-configuration` annotation secret?

pkg/steps/censoringclient/client.go
stevekuznetsov: Yeah, we can enforce that in review.

pkg/steps/secretrecordingclient/client.go
stevekuznetsov: It contains the whole Secret itself, so ... :)

PR #1821, opened 2021-03-25
test: add index content test for optional operator e2e

AlexNPavel:
This PR adds a content test for the operator indices built by the
optional-operators e2e test. The way it is done is by running the
serving the opm registry in the index image and then using `grpcurl` to
identify whether certain CSVs exist. For the `ci-index-named-bundle`
image which has a `base_index`, we verify that both existing bundles and
the new bundle exist in the index. For the unnamed bundle, we verify
that the new bundle is in the index and a bundle that exists in the
`base_index` used for the other index image does not.

Due to a permission error that occurs during the test that doesn't occur
when the index image is actually run in the cluster, the `index.db` file
has to be moved to the `$HOME` directory, otherwise a permissions error
occurs when the database is being read.

### DIFF
--- a/test/e2e/simple/e2e_test.go
+++ b/test/e2e/simple/e2e_test.go
@@ -259,14 +259,17 @@ func TestOptionalOperators(t *testing.T) {
 		name       string
 		indexName  string
 		bundleName string
+		target     string
 	}{{
 		name:       "unnamed bundle",
 		indexName:  "ci-index",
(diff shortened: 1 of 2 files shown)


stevekuznetsov (approved):
This seems reasonable - the permissions is likely due to the fact that a Build is root?

stevekuznetsov (approved):

PR #1823, opened 2021-03-25
Wait for pod deletion during cleanup

jupierce:
Noted this condition with the current implementation:
```
$ oc get pods -n ci-op-bn8cmi7b
NAME                                READY   STATUS        RESTARTS   AGE
launch-openshift-cluster-bot-rbac   0/2     Completed     0          7m5s
launch-osd-create-create            2/2     Terminating   0          6m55s
launch-osd-delete-delete            0/2     Completed     0          60s
```

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -367,6 +367,11 @@ func (s *multiStageTestStep) runSteps(
 				}
 				if err := s.client.Delete(cleanupCtx, &pod); err != nil && !kerrors.IsNotFound(err) {
 					errs = append(errs, fmt.Errorf("failed to delete pod %s with label %s=%s: %w", pod.Name, MultiStageTestLabel, s.name, err))
+					continue
+				}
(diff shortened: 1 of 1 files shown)


pkg/steps/multi_stage.go
```diff
@@ -367,6 +367,10 @@ func (s *multiStageTestStep) runSteps(
 				}
 				if err := s.client.Delete(cleanupCtx, &pod); err != nil && !kerrors.IsNotFound(err) {
 					errs = append(errs, fmt.Errorf("failed to delete pod %s with label %s=%s: %w", pod.Name, MultiStageTestLabel, s.name, err))
+				} else {
+					if err := waitForPodDeletion(s.client, s.jobSpec.Namespace(), pod.Name, pod.UID); err != nil {
```
petr-muller, line 371: Using `continue` above  would allow us to reduce indent level and IMO the block would "flow" better (there's a similar `if cond { continue }` statement above.

jupierce: Sounds good. Change pushed.

stevekuznetsov (approved):

PR #1826, opened 2021-03-29
ci-secret-bootstrap: share clients

bbguimaraes:
Apologies for the resulting size, althgouh most of it is just shuffling.  Most
of the PR is concerned with extracting (extirpating, really) the
`ci-secret-bootstrap` clients from `cmd` to `pkg/secrets` (`client.go`,
`bitwarden.go`, and `vault.go`).

This is so we can stop duplicating code in both `bootstrap` and `generator`,
including the command-line processing and secret censoring implementations.


pkg/secrets/censor.go
```diff
@@ -31,3 +34,23 @@ func (c *DynamicCensor) AddSecrets(s ...string) {
...
+}
+
+// ReadFromFile loads content from a file and adds it to the list.
+func (c *DynamicCensor) ReadFromFile(path string) (string, error) {
+	bytes, err := ioutil.ReadFile(path)
+	if err != nil {
+		return "", err
+	}
+	ret := strings.TrimSpace(string(bytes))
+	c.AddSecrets(ret)
```
petr-muller, line 54: Should we also guard against an empty string (a file that contains an empty string or just whitespace...). I remember that empty string as a censored secret caused mayhem in the past (perhaps `AddSecrets` should guard agaist it? but then the check in `ReadFromEnv` is useless)

petr-muller (commented):
Ugh, not the easiest PR to review ;)
I don't claim to read this line-by-line (a lot of it is just code moved around) but the basics look solid.

pkg/secrets/censor.go
stevekuznetsov: No, the censorer handles this.

pkg/secrets/censor.go
```diff
@@ -31,3 +34,23 @@ func (c *DynamicCensor) AddSecrets(s ...string) {
 	c.secrets.Insert(s...)
 	c.ReloadingCensorer.Refresh(c.secrets.List()...)
 }
+
+// ReadFromEnv loads content from the environment and adds it to the list.
```
stevekuznetsov, line 38: Why is this a method on `DynamicCensor` ?

stevekuznetsov (commented):
Mostly LGTM - not sure why we want to add reading from env/file to the dynamic censorer though instead of just a helper that reads & updates the thing? Seems like a weird encapsulation

pkg/secrets/censor.go
bbguimaraes: The idea was to centralize the methods for loading secrets to:

- reduce the code necessary in clients
- hopefully remind them that secrets should be censored

similar to `test-infra`'s secret `Agent` and `secretrecordingclient`.  Assuming that is useful, the choice between:

1. `secret, err := censor.ReadFromFile(path)`
2. `secret, err := secrets.ReadFromFile(path, &censor)`

was decided purely by character count.  I can't think of an advantage of the second form, but I'm also not opposed to it.

stevekuznetsov: Right, my comment was on the second form - I guess I prefer the composition there, it's weird for a secret recorder to close over file acces.

bbguimaraes:

PR #1827, opened 2021-03-29
ci-operator: create logfile

stevekuznetsov:

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -279,7 +279,10 @@ func setupLogger() (*secrets.DynamicCensor, io.Closer, error) {
 	if !set {
 		return &censor, nil, nil
 	}
-	verboseFile, err := os.Open(filepath.Join(artifactDir, "ci-operator.log"))
+	if err := os.MkdirAll(artifactDir, 0777); err != nil {
+		return nil, nil, err
+	}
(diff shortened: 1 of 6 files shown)


cmd/ci-operator/main.go
```diff
@@ -279,7 +279,7 @@ func setupLogger() (*secrets.DynamicCensor, io.Closer, error) {
 	if !set {
 		return &censor, nil, nil
 	}
-	verboseFile, err := os.Open(filepath.Join(artifactDir, "ci-operator.log"))
+	verboseFile, err := os.Create(filepath.Join(artifactDir, "ci-operator.log"))
```
alvaroaleman, line 282: What if it already exists?

stevekuznetsov: Well, it clearly doesn't :)

alvaroaleman: But how did it go through e2e then?

stevekuznetsov: So `Create` does not fail when the file exists, it just truncates it. I think this is fine behavior. Added code here to make the dir if it does not exist, perhaps that is the cause here. Not entirely sure, but we should be OK to merge the safer code ...

alvaroaleman (approved):

PR #1831, opened 2021-03-30
rehearse: discover periodics affected by step registry changes

petr-muller:
As a whole, this is probably not easy to review, sorry for that. The actual, quite dumb functional change is present in the first commit. The implementation just makes everything done for presubmits done again for periodics. I noticed some flaws of the existing implementation and fixed them one-by-one in the following commits, so per-commit review should be not that hard. The PR brings some efficiency improvements, notably removes useless sorts over the same data that were done for each changed step, and also full ci-op config re-read from disk.

The meat of the change is pretty much isolated in the `SelectJobsForChangedRegistry` and `selectJobsForRegistryStep` methods in the `pkg/rehearse/jobs.go` file, the rest is mostly refactor-related noise.

I still want to write some tests but I consider this to be reviewable in the current state.

Resolves [DPTP-1098](https://issues.redhat.com/browse/DPTP-1098) and contributes to [DPTP-1027](https://issues.redhat.com/browse/DPTP-1027)


pkg/rehearse/jobs.go
```diff
@@ -589,28 +589,22 @@ func getAllAncestors(changed []registry.Node) []registry.Node {
...
-		configs = append(configs, &config.DataWithInfo{Configuration: *cfg, Info: *info})
-		return nil
-	}); err != nil {
-		loggers.Debug.Errorf("Failed to load config by filename in SelectJobsForChangedRegistry: %v", err)
-		return selectedPresubmits, selectedPeriodics
+	// We need a sorted index of ci-operator configs for deterministic behavior
+	var sortedConfigs []*config.DataWithInfo
+	for idx := range ciopConfigs {
+		cfg := ciopConfigs[idx]
+		sortedConfigs = append(sortedConfigs, &cfg)
```
stevekuznetsov, line 609: IIRC this is going to be putting the pointers to the location in the `ciopConfigs` slice into `sortedConfigs` ... can we get `config.DataByFilename` to have pointers itself inside?

pkg/registry/graph.go
```diff
@@ -21,13 +21,13 @@ type Node interface {
...
 	Ancestors() []Node
-	// DescendantNames returns a set of strings containing the names of all of the node's descendants
+	// Descendants returns a set of strings containing the names of all of the node's descendants
 	Descendants() []Node
-	// ParentNames returns a set of strings containing the names of all the node's parents
+	// Parents returns a set of strings containing the names of all the node's parents
 	Parents() []Node
-	// ChildrenNames returns a set of strings containing the names of all the node's children
+	// Childrens returns a set of strings containing the names of all the node's children
 	Childrens() []Node
```
stevekuznetsov, line 31: `Children`  ?

pkg/rehearse/jobs.go
```diff
@@ -517,17 +517,20 @@ type periodicsByName map[string]prowconfig.Periodic
...
 
 // selectJobsForRegistryStep returns a sample from all jobs affected by the provided registry node.
-func selectJobsForRegistryStep(node registry.Node, configs []*config.DataWithInfo, allPresubmits presubmitsByName, allPeriodics periodicsByName, skipJobs sets.String) (presubmitsByRepo, periodicsByRepo) {
+func selectJobsForRegistryStep(node registry.Node, configs []*config.DataWithInfo, allPresubmits presubmitsByName, allPeriodics periodicsByName, skipJobs sets.String, loggers Loggers) (presubmitsByRepo, periodicsByRepo) {
 	selectedPresubmits := make(map[string][]prowconfig.Presubmit)
 	selectedPeriodics := make(map[string][]prowconfig.Periodic)
 
+	loggers.Debug.WithField("node-name", node.Name()).Debug("Searching for jobs affected by changed node")
 	for _, cfg := range configs {
+		logFields := cfg.Info.LogFields()
```
stevekuznetsov, line 526: We don't generally do this, but instead `WithFields()` right away to add to the logger. Why keep the fields around without adding them in this case?

pkg/rehearse/jobs.go
```diff
@@ -576,6 +587,7 @@ func selectJobsForRegistryStep(node registry.Node, configs []*config.DataWithInf
 			}
 		}
 	}
+	loggers.Debug.WithField("node-name", node.Name()).Debug("Found no jobs using node")
```
stevekuznetsov, line 590: Yeah this would really benefit from the logrus hooks impl, no reason to ask the devs to choose which logger to use :|

stevekuznetsov (commented):
Mostly LGTM, couple questions

pkg/rehearse/jobs.go
petr-muller: I did it intentionally as a slice of pointers - there's no need to copy full structs around and the configs are just read, never written. The only point of the slice is to get a sorted view of the map, so pointers seemed like a reasonably lightweight way to do that.

`config.DataByFilename`  having pointers sounds risky, the struct is afaik used a lot elsewhere and we could get some undesired modifications? But I can try it and see what happens.

pkg/registry/graph.go
petr-muller: I'll fix this in a followup if it's okay - I fixed up the godocs to silence a warning but changing unrelated interfaces in a feature PR is excessive.

pkg/rehearse/jobs.go
petr-muller: Will fix

petr-muller:
The intergration test failure is suspicious, leaving out a job... Need to inspect.

pkg/registry/graph.go
petr-muller: I extracted this to https://github.com/openshift/ci-tools/pull/1833

pkg/rehearse/jobs.go
petr-muller: I tried to change `config.DataByFilename` to work with pointers right from the start, but the change is quite invasive -- there is code in other tools that relies on zero values from that map, working around that does not seem to be worth the effort, IMO.

pkg/rehearse/jobs.go
petr-muller: I'm not sure, is this a suggestion for a change that I can make here, or does this mean "we need a followup to rework logging in rehearsals"?

stevekuznetsov: The latter.

stevekuznetsov:
@petr-muller OK - LMK when ready to re-review

petr-muller:
> The intergration test failure is suspicious, leaving out a job... Need to inspect.

I misread the logs: we now rehearse more jobs, not less. This is a consequence of https://github.com/openshift/ci-tools/pull/1831/commits/59b4b5e97f7d61821275814d7b6d0547984d4dc1: previously, some jobs could have been filtered out based on the fact that the `MultiStageTestConfiguration` instance was identical to a one that was previously seen, even for another job. This has *some* merit, as identical ci-operator test stanza is probably a very similar job to a one already rehearsed, but I believe the added complexity is not worth it.

But after this, I added better logging to understand what jobs are selected for what steps, and discovered that we actually process a single registry node multiple times which leads to excessive rehearsals, and I fixed that too (I can split the fix to a separate PR)

PR #1829, opened 2021-03-30
ci-operator: use logrus in main.go

stevekuznetsov:
This change does a couple of things:

 - use logrus instead of log
 - change log messages like "warning: ..." into first-class logrus
   Warn() calls
 - capitalize first letters, add periods as needed
 - change to use WithError() instead of printing it into the message

Notably, I translate `log.Print()` to `logrus.Info()`, so there should
be no change to the lines output to the user-facing log. In my opinion,
some of these meessages are spammy and not useful for a user, but we can
change that in the future.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @openshift/openshift-team-developer-productivity-test-platform

PR #1832, opened 2021-03-30
ci-operator: use logrus everywhere

stevekuznetsov:
This patch ubiquitously:

- uses `logrus.WithError` where appropriate
- uses `logrus.Warn` and `logrus.Error` instead of prepending string
  prefixes to log lines
- moves `log.Print` to `logrus.Info` otherwise

Follow-up to #1829 
I'll next move lines to debug level that don't need to be shown to users.


pkg/defaults/defaults.go
```diff
@@ -415,17 +416,17 @@ func leasesForTest(s *api.MultiStageTestConfigurationLiteral) (ret []api.StepLea
 // the root image.
 func rootImageResolver(client loggingclient.LoggingClient, ctx context.Context) func(root, cache *api.ImageStreamTagReference) (*api.ImageStreamTagReference, error) {
 	return func(root, cache *api.ImageStreamTagReference) (*api.ImageStreamTagReference, error) {
-		log.Printf("Determining if build cache %s can be used in place of root %s\n", cache.ISTagName(), root.ISTagName())
+		logrus.Debugf("Determining if build cache %s can be used in place of root %s\n", cache.ISTagName(), root.ISTagName())
```
droslean, line 419: Do we already have a way to enable debug level?

stevekuznetsov: Yes. Debug and trace go into `$ARTIFACT_DIR/ci-operator.log` instead of stdout.

petr-muller (approved):

PR #1835, opened 2021-03-31
Revert "CI-Operator: Stop writing step graph out"

stevekuznetsov:
ci-operator: add a utility to write censored artifacts

We want to make sure that we censor anything we're writing to artifacts
from this process as we often interact with secrets during runtime for
the first time, so we can't be certain that out-of-process censoring
will do anything sensible. A helper makes it easier for an engineer to
do the right thing.

---

Revert "CI-Operator: Stop writing step graph out"

This reverts commit 4b978f02018eb00ed5dccb1850a8f295d57bd4d9.

---

ci-operator: use the artifacts helper to write the graph

---

/assign @alvaroaleman 

Tested locally, produces output like this:

```
$ cat /tmp/artifacts/TestMultiStage_step_with_credentials/ci-operator-step-graph.json | jq '.[0].manifests[] | select(.metadata.name=="ci-operator") | select(.kind=="Secret")'
{
  "kind": "Secret",
  "apiVersion": "v1",
  "metadata": {
    "name": "ci-operator",
    "namespace": "test-credentials",
    "selfLink": "/api/v1/namespaces/test-credentials/secrets/ci-operator",
    "uid": "c2ff507b-154a-4093-b55c-51d4ed82f628",
    "resourceVersion": "538689868",
    "creationTimestamp": "2020-05-20T21:03:46Z",
    "labels": {
      "dptp.openshift.io/requester": "ci-secret-bootstrap"
    }
  },
  "data": {
    "kubeconfig": "********************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************"
  },
  "type": "Opaque"
}
```

DO NOT MERGE BEFORE https://github.com/kubernetes/test-infra/pull/21593
/hold

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -696,20 +696,15 @@ func (o *options) Run() []error {
 	}
 
 	graph := calculateGraph(nodes)
-	if o.artifactDir != "" {
-		defer func() {
-			serializedGraph, err := json.Marshal(graph)
-			if err != nil {
-				logrus.WithError(err).Error("Failed to marshal graph")
-				return
-			}
+	defer func() {
+		serializedGraph, err := json.Marshal(graph)
+		if err != nil {
+			logrus.WithError(err).Error("Failed to marshal graph")
+			return
+		}
 
-			dest := filepath.Join(o.artifactDir, api.CIOperatorStepGraphJSONFilename)
-			if err := ioutil.WriteFile(dest, serializedGraph, 0644); err != nil {
-				logrus.WithError(err).Error("Failed to write serialized step graph")
-			}
-		}()
-	}
+		_ = api.SaveArtifact(o.censor, api.CIOperatorStepGraphJSONFilename, serializedGraph)
+	}()
 
 	if err := validateGraph(nodes); err != nil {
 		return err
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -25,7 +25,6 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 	"go.uber.org/zap/zapcore"
-
 	appsv1 "k8s.io/api/apps/v1"
 	authapi "k8s.io/api/authorization/v1"
 	coreapi "k8s.io/api/core/v1"
@@ -1196,13 +1195,15 @@ type prowResultMetadata struct {
 	Metadata      map[string]string `json:"metadata"`
 }
 
+const metadataJSONfile = "metadata.json"
+
 func (o *options) writeMetadataJSON() error {
 	artifactDir, set := api.Artifacts()
 	if !set {
 		return nil
 	}
 
-	metadataJSONPath := filepath.Join(artifactDir, "metadata.json")
+	metadataJSONPath := filepath.Join(artifactDir, metadataJSONfile)
 
 	customProwMetadataFile, err := o.findCustomMetadataFile(artifactDir)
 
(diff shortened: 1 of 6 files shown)


pkg/steps/secretrecordingclient/client.go
```diff
@@ -104,11 +104,20 @@ func (c *client) record(obj ctrlruntimeclient.Object) {
 }
 
 func (c *client) recordSecret(secret *v1.Secret) {
+	_, isServiceAccountSecret := secret.ObjectMeta.Annotations["kubernetes.io/service-account.name"]
 	var values []string
-	for _, value := range secret.Data {
+	for key, value := range secret.Data {
+		if key == "namespace" && isServiceAccountSecret {
+			// this is just never a useful thing to censor
```
stevekuznetsov, line 111: I don't want to do this, open to other ideas. We list all secrets in the namespace in order to figure out which we want to censor for the test pods, so we end up reading all of the values in this recording client even if we end up filtering the list of secrets after listing. Censoring the namespace ends up mangling e.g. the jUnit, etc.

stevekuznetsov:
Note that we have to merge the upstream pull *and* vendor it in here before merging this - that will also mean that we need to bump the vendor in `release-controller` first.

alvaroaleman:
What about omitting secrets from this?

stevekuznetsov:
@alvaroaleman We could do that as well, but not needed per se

stevekuznetsov:
Needs https://github.com/openshift/ci-tools/pull/1839

stevekuznetsov:
To expand on the secret thing - we can certainly remove secret collection from this afterword, but what i wanted was a proof that the new censoring approach would stop us from having a leak in the future, so the intent is that reverting the previous revert should leave us in a sensible place, from which we can then change which objects we collect.

cmd/ci-operator/main.go
```diff
@@ -1387,58 +1395,38 @@ func inputHash(inputs api.InputDefinition) string {
...
-			logrus.WithError(err).Errorf("Failed to write %s", path)
-		}
+		_ = api.SaveArtifact(o.censor, path, data)
 		events, _ := kubeClient.Events(o.namespace).List(context.TODO(), meta.ListOptions{})
 		data, _ = json.MarshalIndent(events, "", "  ")
 		path = filepath.Join(namespaceDir, "events.json")
-		if err := ioutil.WriteFile(path, data, 0644); err != nil {
-			logrus.WithError(err).Errorf("Failed to write %s", path)
-		}
+		_ = api.SaveArtifact(o.censor, path, data)
```
alvaroaleman, line 1407: Why do we ignore all the errors? Can we at least logg them?

stevekuznetsov: We log all errors, then the caller chooses to do something with them or not. Wasn't sure how to DRY this out well, in half the cases our artifact stuff is best-effort, in other cases we seem to want to have it impact control flow.

pkg/steps/secretrecordingclient/client.go
alvaroaleman: I would probably check on the type instead of the annotation to see if its a SA secret, but I don't really see a way around special casing the namespace

alvaroaleman (approved):

pkg/steps/secretrecordingclient/client.go
stevekuznetsov: Done

PR #1836, opened 2021-03-31
ci-operator: format user output and debug log differently

stevekuznetsov:
Originally I thought this was going to be really hard to do, and it does
seem to kind of break the architecture of logrus since each entry has an
embedded formatter, but it works, so ... ?


stevekuznetsov:
(This puts JSON into the debug/admin log and keeps normal text for users)

cmd/ci-operator/main.go
```diff
@@ -263,9 +262,10 @@ func setupLogger() (*secrets.DynamicCensor, io.Closer, error) {
 	censor := secrets.NewDynamicCensor()
 	logrus.SetFormatter(logrusutil.NewFormatterWithCensor(logrus.StandardLogger().Formatter, &censor))
 	logrus.SetOutput(ioutil.Discard)
-	logrus.AddHook(&writer.Hook{
-		Writer: os.Stdout,
-		LogLevels: []logrus.Level{
+	logrus.AddHook(&formattingHook{
```
droslean, line 265: That won't respect our censor formatter and we will end up leaking a secret. Can you check this or write a test that ensures that?

stevekuznetsov: Good point ...

droslean: I remember I hit that issue before and I opened https://github.com/sirupsen/logrus/issues/1153. It worth taking a look at it.

stevekuznetsov: I don't think *that* particular bug will hit me, since I'm forcefully formatting in this code :|

stevekuznetsov:
I'll just pull this one in - it's a formatting/style change and everyone's out for the day anyway.

PR #1838, opened 2021-04-01
Expose tests.cluster_claim in ci-op's configuration

hongkailiu:
https://issues.redhat.com/browse/DPTP-2098

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -514,6 +514,9 @@ type TestStepConfiguration struct {
 	// Postsubmit configures prowgen to generate the job as a postsubmit rather than a presubmit
 	Postsubmit bool `json:"postsubmit,omitempty"`
 
+	// ClusterClaim claims an OpenShift cluster and exposes environment variable ${KUBECONFIG} to the test container
+	ClusterClaim *ClusterClaim `json:"cluster_claim,omitempty"`
+
 	// Only one of the following can be not-null.
 	ContainerTestConfiguration                                *ContainerTestConfiguration                                `json:"container,omitempty"`
 	MultiStageTestConfiguration                               *MultiStageTestConfiguration                               `json:"steps,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/api/types.go
```diff
@@ -514,6 +514,10 @@ type TestStepConfiguration struct {
 	// Postsubmit configures prowgen to generate the job as a postsubmit rather than a presubmit
 	Postsubmit bool `json:"postsubmit,omitempty"`
 
+	// ClusterClaim claims an OpenShift cluster and exposes environment variable ${KUBECONFIG} to the test container
+	// if there is a cluster pool provides the required cluster
```
stevekuznetsov, line 518: What does this `if ... ` mean?

stevekuznetsov (commented):
Where do we use the new config?

pkg/api/types.go
hongkailiu: I was trying to express the condition of getting a cluster via the pool.
A user won't get any cluster if it claims an OCP version `1` because no pools provides it.
Maybe I do not need to focus on the case like that.

Removed the `if` part.

pkg/api/types.go
```diff
@@ -527,6 +530,32 @@ type TestStepConfiguration struct {
...
+	CloudAWS Cloud = "aws"
+)
+
+// ClusterClaim claims an OpenShift cluster for the job.
+type ClusterClaim struct {
+	// Product is the name of the product being released.
+	// Defaults to ocp.
+	Product ReleaseProduct `json:"product,omitempty"`
+	// Version is the version of the product
+	Version string `json:"version,omitempty"`
```
petr-muller, line 546: The non-optional fields should probablky not be omitempty?

hongkailiu: Updated.

stevekuznetsov (approved):

PR #1840, opened 2021-04-04
prowgen: generate the hive-kubeconfig arg

hongkailiu:
https://issues.redhat.com/browse/DPTP-2099

/hold

wait for https://github.com/openshift/ci-tools/pull/1838 and https://github.com/openshift/ci-tools/pull/1837

### DIFF
--- a/pkg/api/constant.go
+++ b/pkg/api/constant.go
@@ -32,6 +32,11 @@ const (
 	HiveAdminPasswordSecret = "hive-admin-password"
 	// HiveAdminPasswordSecretKey is the key to the password in the secret HiveAdminKubeconfigSecret
 	HiveAdminPasswordSecretKey = "password"
+
+	// HiveControlPlaneKubeconfigSecret is the name of the secret that stores kubeconfig to contact the cluster where Hive is deployed
+	HiveControlPlaneKubeconfigSecret = "app.ci-hive-credentials"
+	// HiveControlPlaneKubeconfigSecretArg is the flag to ci-operator
+	HiveControlPlaneKubeconfigSecretArg = "--hive-kubeconfig=/secrets/app.ci-hive-credentials/kubeconfig"
 )
 
 var (
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -75,14 +75,15 @@ func generatePodSpec(info *ProwgenInfo, secrets []*cioperatorapi.Secret) *corev1
 	}
 
 	for _, secret := range secrets {
+		name := strings.ReplaceAll(secret.Name, ".", "-")
 		volumeMounts = append(volumeMounts, corev1.VolumeMount{
-			Name:      secret.Name,
+			Name:      name,
 			MountPath: fmt.Sprintf("/secrets/%s", secret.Name),
 			ReadOnly:  true,
 		})
 
 		volumes = append(volumes, corev1.Volume{
-			Name: secret.Name,
+			Name: name,
 			VolumeSource: corev1.VolumeSource{
 				Secret: &corev1.SecretVolumeSource{SecretName: secret.Name},
 			},
@@ -153,7 +154,13 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 			element.Secrets = append(element.Secrets, element.Secret)
 		}
 		if element.ContainerTestConfiguration != nil {
-			podSpec = generateCiOperatorPodSpec(info, element.Secrets, []string{element.As})
+			var additionalArgs []string
+			secrets := element.Secrets
+			if element.ClusterClaim != nil {
+				additionalArgs = []string{cioperatorapi.HiveControlPlaneKubeconfigSecretArg}
+				secrets = append(secrets, &api.Secret{Name: api.HiveControlPlaneKubeconfigSecret})
+			}
+			podSpec = generateCiOperatorPodSpec(info, secrets, []string{element.As}, additionalArgs...)
 		} else if element.MultiStageTestConfiguration != nil {
 			podSpec = generatePodSpecMultiStage(info, &element, configSpec.Releases != nil || element.ClusterClaim != nil)
 		} else {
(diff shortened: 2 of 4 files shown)


pkg/api/constant.go
```diff
@@ -15,4 +15,8 @@ const (
 	DPTPRequesterLabel = "dptp.openshift.io/requester"
 
 	KVMDeviceLabel = "devices.kubevirt.io/kvm"
+
+	HiveControlPlaneKubeconfigSecret          = "hive-kubeconfig"
```
stevekuznetsov, line 19: There will be dozens of these

pkg/api/types.go
```diff
@@ -527,6 +532,24 @@ type TestStepConfiguration struct {
...
+	// Product is the name of the product being released.
+	// Defaults to ocp.
+	Product ReleaseProduct `json:"product,omitempty"`
+	// Version is the version of the product
+	Version string `json:"version,omitempty"`
+	// Architecture is the architecture for the product.
+	// Defaults to amd64.
+	Architecture ReleaseArchitecture `json:"architecture,omitempty"`
+	// Cloud is the cloud where the product is installed, e.g., aws.
+	Cloud ClusterProfile `json:"cloud,omitempty"`
```
stevekuznetsov, line 545: I don't know that we necessarily want to use this type here

pkg/api/types.go
```diff
@@ -514,6 +514,11 @@ type TestStepConfiguration struct {
 	// Postsubmit configures prowgen to generate the job as a postsubmit rather than a presubmit
 	Postsubmit bool `json:"postsubmit,omitempty"`
 
+	// Please do not use: Still under implementation.
```
stevekuznetsov, line 517: probably don't want to check this in

pkg/api/constant.go
hongkailiu: This is the kubeconfig file that ci-operator (running on a build-farm cluster) uses to contact Hive (running on `app.ci`), instead of the admin kubeconfig file for the test contacting the ephemeral cluster that Hive creates.
We should have exactly one secret like this.

pkg/api/types.go
hongkailiu: Will remove it in https://github.com/openshift/ci-tools/pull/1838

pkg/api/types.go
hongkailiu: I will define a new type for this.
But the change will be in https://github.com/openshift/ci-tools/pull/1838 and this PR will be rebased.

pkg/api/constant.go
hongkailiu: Renamed.

bbguimaraes: I think the point here is that this is only one of the Hive instances we are planning to support.  Once we have third-party instances, a single boolean value based on `element.ClusterClaim` will not be sufficient.

This is fine as a provisional implementation if it's meant as a POC, but how are we planning to determine which `kubeconfig` to use in the future?

alvaroaleman (commented):
Instead of going through this dance with the flag and adding a new bool param everywhere that is  hardcoded to `false` in the majority of cases, why not one of:
* Setting an env var that tells the ci-operator where to find this
* Just put it into a default location the ci-operator knows about

pkg/prowgen/prowgen.go
```diff
@@ -268,14 +283,17 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	}
 }
 
-func generateCiOperatorPodSpec(info *ProwgenInfo, secrets []*cioperatorapi.Secret, targets []string, additionalArgs ...string) *corev1.PodSpec {
+func generateCiOperatorPodSpec(info *ProwgenInfo, secrets []*cioperatorapi.Secret, targets []string, needHiveKubeconfig bool, additionalArgs ...string) *corev1.PodSpec {
+	if needHiveKubeconfig {
+		additionalArgs = []string{cioperatorapi.HiveControlPlaneKubeconfigSecretArg}
```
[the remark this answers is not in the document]
hongkailiu: Should be `append` instead.

hongkailiu:
> Instead of going through this dance with the flag and adding a new bool param everywhere that is hardcoded to `false` in the majority of cases, why not one of:
> 
> * Setting an env var that tells the ci-operator where to find this
> * Just put it into a default location the ci-operator knows about

Without this boolean, we would mount the Hive kubeconfig onto every job even though the job does not need it, right?
Maybe I did see through your suggestions?

pkg/prowgen/prowgen.go
```diff
@@ -312,7 +330,7 @@ func generatePodSpecMultiStage(info *ProwgenInfo, test *cioperatorapi.TestStepCo
 			Name: "ci-pull-credentials",
 		})
 	}
-	podSpec := generateCiOperatorPodSpec(info, secrets, []string{test.As})
+	podSpec := generateCiOperatorPodSpec(info, secrets, []string{test.As}, test.ClusterClaim != nil)
```
alvaroaleman, line 333: What I meant with my other comment about the boolean flag is: At this point here in the code we know if we need the hive kubeconfig or not. So instead of adding another boolean flag, we can just append to the existing `secrets` arg

pkg/prowgen/prowgen.go
```diff
@@ -269,13 +288,17 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 }
 
 func generateCiOperatorPodSpec(info *ProwgenInfo, secrets []*cioperatorapi.Secret, targets []string, additionalArgs ...string) *corev1.PodSpec {
+	var needHiveKubeconfig bool
```
alvaroaleman, line 291: What I meant was: Remove the boolean arg from everywhere, including `generatePodSpec`. Just add the secret to the `secrets` slice. Make the ci-operator not need a flag but derive from the steps if it needs a hive kubeconfig. Default the location for the hive kubeconfig. If the ci-operator derived from the steps that it needs the hive kubeconfig, it should look at the default location and fail if its not present there.

hongkailiu: Updated.

pkg/prowgen/prowgen_test.go
```diff
@@ -111,7 +111,6 @@ func TestGeneratePodSpecMultiStage(t *testing.T) {
 				As:           "test",
 				ClusterClaim: &ciop.ClusterClaim{},
 				MultiStageTestConfiguration: &ciop.MultiStageTestConfiguration{
-					ClusterProfile: ciop.ClusterProfileAWS,
```
hongkailiu, line 114: ClusterClaim and ClusterProfile are exclusive.

alvaroaleman (approved):

PR #1843, opened 2021-04-06
[DPTP-2094] ci-secret-bootstrap: validate if the given items/field/attachments exist

droslean:

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -741,23 +741,59 @@ func getUnusedBWItems(config secretbootstrap.Config, client secrets.ReadOnlyClie
 	return utilerrors.NewAggregate(errs)
 }
 
+func validateBWItems(client secrets.ReadOnlyClient, secretConfigs []secretbootstrap.SecretConfig) error {
+	var errs []error
+
+	for _, config := range secretConfigs {
+		for _, item := range config.From {
+
+			if item.DockerConfigJSONData != nil {
+				for _, data := range item.DockerConfigJSONData {
+					if !client.HasItem(data.BWItem) {
+						errs = append(errs, fmt.Errorf("item %s doesn't exist", data.BWItem))
+						break
+					}
+					if _, err := client.GetAttachmentOnItem(data.BWItem, data.AuthBitwardenAttachment); err != nil {
+						errs = append(errs, fmt.Errorf("attachment %s in item %s doesn't exist", data.AuthBitwardenAttachment, data.BWItem))
+					}
+				}
+			} else {
+				if !client.HasItem(item.BWItem) {
+					errs = append(errs, fmt.Errorf("item %s doesn't exist", item.BWItem))
+					continue
+				}
+
+				if item.Field != "" {
+					if _, err := client.GetFieldOnItem(item.BWItem, item.Field); err != nil {
+						errs = append(errs, fmt.Errorf("field %s in item %s doesn't exist", item.Field, item.BWItem))
+					}
+				}
+
+				if item.Attachment != "" {
+					if _, err := client.GetAttachmentOnItem(item.BWItem, item.Attachment); err != nil {
+						errs = append(errs, fmt.Errorf("attachment %s in item %s doesn't exist", item.Attachment, item.BWItem))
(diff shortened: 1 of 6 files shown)


stevekuznetsov (commented):
Technically this is an avenue for attack (to list items in BW) if we run in a presubmit but I think it's useful enough to have?

droslean:
> Technically this is an avenue for attack (to list items in BW) if we run in a presubmit but I think it's useful enough to have?

Well, yes. If someone opens a PR adding a trillion secrets in the config, we will brute force Bitwarder, but I think this is fine since `ok-to-test` is protecting us. My main concern is if we still care about Bitwarden, since we are planning to migrate our secrets to vault.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -672,23 +672,63 @@ func getUnusedBWItems(config secretbootstrap.Config, client secrets.Client, bwAl
...
+			} else {
+				if !client.HasItem(item.BWItem) {
+					errs = append(errs, fmt.Errorf("item %s doesn't exist", item.BWItem))
+					continue
+				}
+
+				if item.Field != "" {
+					if _, err := client.GetFieldOnItem(item.BWItem, item.Field); err != nil {
+						errs = append(errs, fmt.Errorf("field %s in item %s doesn't exist", item.Field, item.BWItem))
+						continue
```
petr-muller, line 701: These `continue`s` (this one and the ones below) do not make sense to me. If we expect exactly one of the `if`s to apply, then the three `if` statements should be a switch, or we expect multiple ones to apply, and then we should not abort on finding a first error.

droslean: Nice catch. I will update it. Thanks

droslean: updated

cmd/ci-secret-bootstrap/main.go
```diff
@@ -741,23 +741,59 @@ func getUnusedBWItems(config secretbootstrap.Config, client secrets.ReadOnlyClie
...
+	var errs []error
+
+	for _, config := range secretConfigs {
+		for _, item := range config.From {
+
+			if item.DockerConfigJSONData != nil {
+				for _, data := range item.DockerConfigJSONData {
+					if !client.HasItem(data.BWItem) {
+						errs = append(errs, fmt.Errorf("item %s doesn't exist", data.BWItem))
+						continue
```
stevekuznetsov, line 754: This `continue` does not make sense - the next check is for an attachment to this item, so if the item does not exist there's no point in checking for the attachment.

droslean: :facepalm:  Let me fix this

droslean: updated

pkg/secrets/vault.go
```diff
@@ -75,6 +75,12 @@ func (c *vaultClient) setItemAtPath(path, field string, content string) error {
 	return c.upstream.UpsertKV(path, data)
 }
 
+func (c *vaultClient) HasItem(itemName string) bool {
+	path := c.pathFor(itemName)
+	_, err := c.upstream.GetKV(path)
+	return err != nil
```
alvaroaleman, line 81: Please make this a IsNotFound check instead

droslean: updated.

pkg/secrets/vault.go
```diff
@@ -75,6 +75,12 @@ func (c *vaultClient) setItemAtPath(path, field string, content string) error {
 	return c.upstream.UpsertKV(path, data)
 }
 
+func (c *vaultClient) HasItem(itemName string) bool {
+	path := c.pathFor(itemName)
+	_, err := c.upstream.GetKV(path)
+	return vaultclient.IsNotFound(err)
```
alvaroaleman, line 81: THis needs to be negated, if it is a notfound, you need to return false, not true

droslean: :man_facepalming:

alvaroaleman:
Gimme link for DPTP-2094

alvaroaleman (approved):

PR #1844, opened 2021-04-06
Registry-Replacer: Ingore ocp-buil-data configs that set from.image

alvaroaleman:


alvaroaleman:
/hold
likely not needed

PR #1850, opened 2021-04-07
prowgen: add setting to skip rehearsal for a presubmit job

ray-harris:
We have several pre-submit jobs that, while they can be rehearsed, will always fail rehearsal.

Our end-to-end tests checkout and spin up a pretty large kubernetes cluster and then installs our ACM product on it - just for the rehearsal to fail.

I looked through the ci-tools code to see if there was an option to prevent a test from being rehearsed, but couldn't find anything.

This PR adds a "SkipRehearsal" option to TestStepConfiguration that, if set, would not put the "can-be-rehearsed" label on a pre-submit job.

This setting is ignored for periodic and postsubmit jobs. Periodic jobs will always be rehearsed while postsubmit jobs will never be rehearsed.

I ran `make generate` locally to update the documentation, but it didn't seem to change anything. I'm using a Mac, so that may affect it.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -485,9 +485,16 @@ type PipelineImageCacheStepConfiguration struct {
 type TestStepConfiguration struct {
 	// As is the name of the test.
 	As string `json:"as"`
+
 	// Commands are the shell commands to run in
 	// the repository root to execute tests.
 	Commands string `json:"commands,omitempty"`
+
+	// SkipRehearsal indicates that the "can-be-rehearsed"
+	// label is to be omitted. This only applies to presubmit
+	// jobs. Postsubmit jobs will never be rehearsed. Periodic
+	// jobs will always be rehearsed.
+	SkipRehearsal bool `json:"skip_rehearsal,omitempty"`
 
 	// Secret is an optional secret object which
 	// will be mounted inside the test container.
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -179,7 +179,8 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 			postsubmit.MaxConcurrency = 1
 			postsubmits[orgrepo] = append(postsubmits[orgrepo], *postsubmit)
 		} else {
-			presubmit := *generatePresubmitForTest(element.As, info, podSpec, configSpec.CanonicalGoRepository, jobRelease, skipCloning)
+			rehearsable := !element.SkipRehearsal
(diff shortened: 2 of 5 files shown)


pkg/api/types.go
```diff
@@ -485,10 +485,17 @@ type PipelineImageCacheStepConfiguration struct {
...
+
 	// Commands are the shell commands to run in
 	// the repository root to execute tests.
 	Commands string `json:"commands,omitempty"`
 
+	// SkipRehearsal indicates that the "can-be-rehearsed"
+	// label is to be omitted. This only applies to presubmit
+	// jobs. Postsubmit jobs will never be rehearsed. Periodic
+	// jobs will always be rehearsed.
+	SkipRehearsal bool `json:"skip_rehearsal,omitempty"`
```
alvaroaleman, line 497: How about alternatively change your jobs so they can be rehearsed?

ray-harris: The current default behavior is to always rehearse presubmit jobs. If I negated the logic here, none of the existing presubmit jobs would be rehearsed and all test configurations across all orgs and repos would need to be reviewed and updated to enable rehearsing jobs.

I can do this as it's not truly a "breaking" change, but I felt it would be better to incrementally change the behavior rather than completely upend it.

Just let me know.

ray-harris: Sorry. I misread your comment earlier. I understand it now.

The problem with rehearsing these jobs is that they require $REPO_OWNER, $REPO_NAME, and $PULL_NUMBER to be set correctly to work. But when the job is rehearsed, these are set to `openshift`, `release`, and whatever the PR number is where the rehearsal is taking place.

Our jobs that run the SonarCloud scan fail because the $PULL_NUMBER in the rehearsal isn't in the actual component repo and so it fails with this type of error:
```
DEBUG: GET 404 https://sonarcloud.io/api/alm_integration/show_pullrequest?project=open-cluster-management_console_api&pullrequestKey=13425
```
where `13425` is the PR number for `openshift/release` that doesn't exist in the `open-cluster-management/console-api` repo. This isn't the primary concern though as this doesn't really "cost" us anything, but the rehearsal will never pass.

Our e2e tests are going to be the ones that cause the most problems (and cost). Here's the entry for the workflow:
https://steps.ci.openshift.org/workflow/ocm-e2e-clusterpool

I know the docs say that using these Prow variables "breaks encapsulation" but our e2e workflow (in a pre step command script) needs to know which component is being tested. It has to checkout an OCP cluster, partially deploy ACM, patch the deployment CSV for ACM with the image ref for that component, and finish the deployment. During a pre step, the only place the component info under test seems to be available is in $REPO_OWNER and $REPO_NAME.

PR #1851, opened 2021-04-07
Vault subpath proxy: Fix secret key validation

alvaroaleman:
The current validation is way too strict and not what kube itself does.

### DIFF
--- a/cmd/vault-subpath-proxy/kv_key_validator.go
+++ b/cmd/vault-subpath-proxy/kv_key_validator.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


cmd/vault-subpath-proxy/kv_key_validator.go
```diff
@@ -15,6 +16,11 @@ import (
 	"github.com/openshift/ci-tools/pkg/api/vault"
 )
 
+// https://github.com/openshift/ci-tools/blob/7af2e075f381ecae1562d1406bad2c86a23e72a3/vendor/k8s.io/api/core/v1/types.go#L5748-L5749
+const secretKeyValidationRegexString = `^[a-zA-Z0-9\.\-_]+$`
```
hongkailiu, line 20: nit: I would expect upstream has such a `const` already.

stevekuznetsov (approved):

hongkailiu (approved):

PR #1852, opened 2021-04-07
ci-operator: update the label set we use on Pods

stevekuznetsov:
The intent of adding labels to the Pods we create was always two-fold:
first, help administrators identify Pods as they relate to jobs and
steps. Second, allow for the labels to end up in Prometheus to help us
track metrics across job runs. Unfortunately, the previous label set
fell short in the second case:

 - we added the full job name as a label, which was trimmed to fit in
   most cases and in some cases the trimming left only the org/repo/etc
   parts, making it not actually useful as an identifier
 - we did not label the target that ci-operator executed, just fields
   like the full job name which might include it as a suffix
 - we did not label the multi-stage step name, now allowing us to
   discern individual usage metrics
 - we labeled the build and Prow job IDs, which are unique and not
   ultimately useful to selecting anything

This patch changes the label set slightly to make it more useful.

Needed for [DPTP-1822](https://issues.redhat.com/browse/DPTP-1822)

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -484,7 +484,12 @@ func (o *options) Complete() error {
 		jobSpec.Refs = spec.Refs
 	}
 	jobSpec.BaseNamespace = o.baseNamespace
+	target := "all"
+	if len(o.targets.values) > 0 {
+		target = o.targets.values[0]
+	}
 	o.jobSpec = jobSpec
+	o.jobSpec.Target = target
 
 	info := o.getResolverInfo(jobSpec)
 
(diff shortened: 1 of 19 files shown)


cmd/ci-operator/main.go
```diff
@@ -503,6 +508,7 @@ func (o *options) Complete() error {
 		o.jobSpec.Refs.PathAlias = *config.CanonicalGoRepository
 	}
 	o.configSpec = config
+	o.jobSpec.Metadata = config.Metadata
```
stevekuznetsov, line 511: So there's a class of jobs (looking at you, release informers...) which have neither Prow refs nor metadata as they use `$RESOLVED_CONFIG` or whatever and don't actually come from the server. We can phase in requiring those jobs to explicitly set `zz_generated_metadata` manually so that their Pods end up getting automatic requests & limits?

AlexNPavel: Shouldn't be too difficult to make a script that adds `zz_generated_metadata` to all `CONFIG_SPEC` and `UNRESOLVED_CONFIG` sections for handwritten release jobs. I don't think we need anything more complicated for those jobs.

AlexNPavel (approved):
LGTM

stevekuznetsov:
... ok yet?

PR #1854, opened 2021-04-08
ci-operator: don't record SA namespaces for censoring

stevekuznetsov:
This was somehow lost from the previous commits which handled this case.
There is no world in which the Namespace in which we work is secret, as
we create it for the test and it's entirely a derivative value from our
inputs. This is a silly hack but there's no real better way of ignoring
it.

/assign @alvaroaleman


stevekuznetsov:
Both flakes,


/retest

alvaroaleman (approved):

PR #1855, opened 2021-04-08
pkg/secrets: properly default/validate arguments

bbguimaraes:
https://github.com/openshift/release/pull/17537#discussion_r609096514

### DIFF
--- a/pkg/secrets/flags.go
+++ b/pkg/secrets/flags.go
@@ -3,7 +3,6 @@ package secrets
 import (
 	"flag"
 	"fmt"
-	"os"
 
 	"k8s.io/apimachinery/pkg/util/sets"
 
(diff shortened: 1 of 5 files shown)


pkg/secrets/flags.go
```diff
@@ -38,11 +42,16 @@ func (o *CLIOptions) Validate() []error {
 	} else if o.BwUser != "" {
 		credentialsProviderConfigured = append(credentialsProviderConfigured, "bitwarden")
 	}
-
-	if vals := sets.NewString(o.VaultAddr, o.VaultTokenFile, o.VaultPrefix); len(vals) != 1 && ((len(vals) != 3) || vals.Has("")) {
-		errs = append(errs, fmt.Errorf("--vault-addr, --vault-token and --vault-prefix must be specified together"))
-	} else if len(vals) == 3 {
+	b2i := func(b bool) int {
```
bbguimaraes, line 45: Why does anyone use this silly language?

alvaroaleman: can we instead just if any of addr, token or prefix is set verify that all are set? This is not good code.

alvaroaleman: (and I am aware that I started it)

bbguimaraes: That is the verification it performs (see tests).

Other than the ridiculous step required in Go to have access to fundamental operations, I don't see what is not good about it.  It checks the required condition, and gets reasonably close to maximum efficiency:

<details>

``` go
	return b2i(o.s0 != "") + b2i(o.s1 != "" || o.s2 != "") + b2i(o.s3 != "")
```

```asm
        movq    "".o+8(SP), AX
        movq    8(AX), CX
        cmpq    24(AX), $0
        jeq     _O_f_pc61
        testq   CX, CX
        movl    $1, DX
_O_f_pc24:
        movq    56(AX), AX
        testq   CX, CX
        setne   CL
        testq   AX, AX
        setne   AL
        movblzx DL, DX
        movblzx CL, CX
        addq    DX, CX
        movblzx AL, AX
        addq    CX, AX
        movq    AX, "".~r0+16(SP)
        ret
_O_f_pc61:
        cmpq    40(AX), $0
        setne   DL
        testq   CX, CX
        jmp     _O_f_pc24
```

</details>

https://godbolt.org/z/PfnqK9vxn

pkg/secrets/flags.go
```diff
@@ -38,11 +42,16 @@ func (o *CLIOptions) Validate() []error {
...
+		if b {
+			return 1
+		}
+		return 0
+	}
+	if n := b2i(o.VaultAddr != "") + b2i(o.VaultToken != "" || o.VaultTokenFile != "") + b2i(o.VaultPrefix != ""); n == 3 {
```
alvaroaleman, line 51:
```suggestion
	if o.VaultAddr != "" || o.VaultTokenFile != "" || o.VaultPrefix != "" {
	    if o.VaultAddr == "" || o.vaultToken == "" || o.VaultPrefix == "" {
	        return err
	    } else { credentialsProvidersConfigured {...}
```

My point was not that your change is incorrect, but that it is unreadable

alvaroaleman: @bbguimaraes I've included your commit and a fixup into https://github.com/openshift/ci-tools/pull/1857 because it would otherwise conflict with this PR

bbguimaraes:
Added in https://github.com/openshift/ci-tools/pull/1857.

/close

PR #1865, opened 2021-04-12
[WIP] pod-scaler: a server for automated resource requests

stevekuznetsov:
This commit introduces a new server to automate setting resource

recommendations and can also serve a web UI.

You can see the data that the producer creates (and that the consumers read) with:

```
gsutil ls gs://origin-ci-resource-usage-data/*.json
```

If you want to run it locally, use the `resource-usage-data-publisher@@openshift-ci-infra.iam.gserviceaccount.com` credentials.

Items that are still TODO:

 - [ ] end-to-end tests with dummy data for the consumers
 - [ ] end-to-end tests with dummy Prometheus server for the producer
 - [x] gather data for raw Prow Pods (not spawned by ci-operator but by plank itself)
 - [x] actually make the UI styling and layout work, form fills need to reset, etc
 - [x] figure out the best way to do a histogram heatmap style presentation/visualization of the data
 - [ ] determine how we're going to allow users to override this with config
 - [x] plumb it through to actually be a mutating admission webhook
 - [ ] (??) write some sort of tool to dump this data into `openshift/release` config for stale branches so the server does not need to load quite so much
 - [ ] write some sort of tool to prune configuration values when the server knows good values, so our config is not cluttered with useless requests


stevekuznetsov:
UI TODOs:

 - [ ] real header/navigation/search bar at the top
 - [ ] scale to height
 - [ ] rescale on resize
 - [ ] tabs for specific/step-generic views
 - [ ] show cutoffs & suggested requests
 - [ ] blurb on how to read the charts

cmd/pod-scaler/producer.go
```diff
@@ -0,0 +1,494 @@
...
+
+	// MaxSamplesPerRequest is the maximum number of samples that Prometheus will allow a client to ask for in
+	// one request. We also use this to approximate the maximum number of samples we should be asking any one
+	// Prometheus server for at once from many requests.
+	MaxSamplesPerRequest = 11000
+)
+
+func Produce(clients map[string]prometheusapi.API, dataCache Cache) {
+	ticker := time.NewTicker(3 * time.Hour)
+	interrupts.Run(func(ctx context.Context) {
```
alvaroaleman, line 41: Why not `interrupts.Tick`?

cmd/pod-scaler/storage.go
```diff
@@ -0,0 +1,141 @@
...
+func (l *localCache) LastUpdated(name string) (time.Time, error) {
+	info, err := os.Stat(path.Join(l.dir, name))
+	if err != nil {
+		return time.Time{}, fmt.Errorf("could not query cache for attributes: %w", err)
+	}
+	return info.ModTime(), nil
+}
+
+// LoadCache loads cached query data from the canonical location in GCS.
+func LoadCache(loader Loader, metricName string, logger *logrus.Entry) (*CachedQuery, error) {
```
alvaroaleman, line 89: Is there a need for this to be exported?

cmd/pod-scaler/producer.go
```diff
@@ -0,0 +1,494 @@
...
+	r := prometheusapi.Range{
+		Start: until.Add(-time.Duration(retention)),
+		End:   until,
+		Step:  1 * time.Minute,
+	}
+
+	errLock := &sync.Mutex{}
+	var errs []error
+	go func() {
+		errLock.Lock()
```
alvaroaleman, line 180: Nit: Not using `defer` can easily lead to hard to debug locking issues down the line if someone changes the content of the closure

cmd/pod-scaler/producer.go
```diff
@@ -0,0 +1,494 @@
...
+			}
+		}
+		logger.WithError(err).Error("Failed to query Prometheus API.")
+		c.errors <- fmt.Errorf("failed to query Prometheus API: %w", err)
+		return
+	}
+	if len(warnings) > 0 {
+		logger.WithField("warnings", warnings).Warn("Got warnings from Prometheus.")
+	}
+	logger.Debugf("Queried Prometheus API after %s.", time.Since(queryStart))
```
alvaroaleman, line 297: Maybe something like `Took $duration to query Prometheus api`?

cmd/pod-scaler/storage.go
stevekuznetsov: Kind of - this is all in an unimportable `main` right now anyway, but I figured we could move things out into packages before merge? WDYT?

stevekuznetsov:
Part of DPTP-1822

stevekuznetsov:
Depends on https://github.com/openshift/ci-tools/pull/1911

PR #1869, opened 2021-04-13
Determinize Prow config: Shard prow config

alvaroaleman:
This PR makes determinize-prow-config shard by org repo. It is written
under the assumption that we will add a knob upstream to make the
filename for the supplemental files configurable because:
* We want both Prow and plugin configs in there
* We want to add the basedir as an additional folder, but we do not want
  the main config to be considered a supplemental config, as that will
  fail the merging. Not adding the basedir would mean to add each org
  folder seperately.

### DIFF
--- a/cmd/determinize-prow-config/main.go
+++ b/cmd/determinize-prow-config/main.go
@@ -10,10 +10,13 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
+	"path/filepath"
 	"sort"
 
 	"github.com/sirupsen/logrus"
+	"github.com/spf13/afero"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/plugins"
@@ -23,7 +26,8 @@ import (
 )
 
 type options struct {
-	prowConfigDir string
+	prowConfigDir            string
+	shardedProwConfigBaseDir string
 }
 
 func (o *options) Validate() error {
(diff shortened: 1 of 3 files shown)


cmd/determinize-prow-config/main.go
```diff
@@ -181,3 +198,54 @@ func updatePluginConfig(configDir string) error {
 
 	return ioutil.WriteFile(configPath, data, 0644)
 }
+
+// prowConfigWithPointers mimics the upstream prowConfig but has pointer fields only
+// in order to avoid serializing empty structs.
+type prowConfigWithPointers struct {
```
alvaroaleman, line 204: So it turns out we do need a different type, otherwise we end up marshalling a bunch of empty config in every file. This will get more complicated when we have this issue with substructs.

stevekuznetsov: Isn't this already an issue if we didn't have any BP for instance? Why not change upstream?

cmd/determinize-prow-config/main_test.go
```diff
@@ -70,3 +76,109 @@ func TestDeduplicateTideQueries(t *testing.T) {
...
+							Repos: map[string]config.Repo{
+								"release": {Policy: config.Policy{Protect: utilpointer.BoolPtr(false)}},
+							},
+						},
+					},
+				},
+			},
+
+			expectedShardFiles: map[string]string{
+				"openshift/_prowconfig.yaml": strings.Join([]string{
```
stevekuznetsov, line 104: Why not back-tick string literals for this?

cmd/determinize-prow-config/main.go
alvaroaleman: Because changing upstream means to change a lot of code as now we have to nilcheck everywhere and it might end up sparkling a lot of debate.

cmd/determinize-prow-config/main_test.go
alvaroaleman: Not a super strong opinion one way or another, but I find backtick literals more annoying to read because the second line onwards starts at the beginning of the file without any indent. `cmp` spat this out so I didn't hand write it and it looks quite nice to me.

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-2072

PR #1873, opened 2021-04-14
prowgen: incldue ci-pull-credentials for tests with cluster-claim

AlexNPavel:
Currently, multistage tests only include `ci-pull-credentials` if the
`configSpec` has a release. It should also include the credentials if
the test specifies a cluster claim.

/cc @hongkailiu


hongkailiu (commented):
/lgtm

PR #1874, opened 2021-04-14
cluster claim: use watch API

hongkailiu:
/cc @stevekuznetsov

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -88,9 +88,9 @@ func FromConfig(
 
 	podClient := steps.NewPodClient(client, clusterConfig, coreGetter.RESTClient())
 
-	var hiveClient ctrlruntimeclient.Client
+	var hiveClient ctrlruntimeclient.WithWatch
 	if hiveKubeconfig != nil {
-		hiveClient, err = ctrlruntimeclient.New(hiveKubeconfig, ctrlruntimeclient.Options{})
+		hiveClient, err = ctrlruntimeclient.NewWithWatch(hiveKubeconfig, ctrlruntimeclient.Options{})
 		if err != nil {
 			return nil, nil, fmt.Errorf("could not get Hive client for Hive kube config: %w", err)
 		}
@@ -111,7 +111,7 @@ func fromConfig(
 	templateClient steps.TemplateClient,
 	podClient steps.PodClient,
 	leaseClient *lease.Client,
-	hiveClient ctrlruntimeclient.Client,
+	hiveClient ctrlruntimeclient.WithWatch,
 	httpClient release.HTTPClient,
 	requiredTargets []string,
 	cloneAuthConfig *steps.CloneAuthConfig,
@@ -320,7 +320,7 @@ func stepForTest(
 	leaseClient *lease.Client,
 	templateClient steps.TemplateClient,
 	client loggingclient.LoggingClient,
-	hiveClient ctrlruntimeclient.Client,
+	hiveClient ctrlruntimeclient.WithWatch,
 	jobSpec *api.JobSpec,
 	inputImages inputImageSet,
 	c *api.TestStepConfiguration,
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -207,6 +207,55 @@ func (c podClient) WithNewLoggingClient() PodClient {
 	}
 }
 
+type evaluator func(obj runtime.Object) (bool, error)
+
+// waitForConditionOnObject uses a watch to wait for a condition to be true on an object.
+// When the condition is satisfied or the timeout expires, the object is returned along
+// with any errors encountered.
+func waitForConditionOnObject(ctx context.Context, client ctrlruntimeclient.WithWatch, identifier ctrlruntimeclient.ObjectKey, list ctrlruntimeclient.ObjectList, into ctrlruntimeclient.Object, evaluate evaluator, timeout time.Duration) error {
+	lw := &cache.ListWatch{
+		ListFunc: func(options metav1.ListOptions) (object runtime.Object, e error) {
+			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", identifier.Name).String()
+			res := list
+			return res, client.List(ctx, res, &ctrlruntimeclient.ListOptions{Namespace: identifier.Namespace, Raw: &options})
+		},
+		WatchFunc: func(options metav1.ListOptions) (i watch.Interface, e error) {
+			opts := &ctrlruntimeclient.ListOptions{
+				Namespace:     identifier.Namespace,
+				FieldSelector: fields.OneTermEqualSelector("metadata.name", identifier.Name),
+				Raw:           &options,
+			}
+			res := list
+			return client.Watch(ctx, res, opts)
+		},
+	}
+
+	// objects in this call here are always expected to exist in advance
+	var existsPrecondition toolswatch.PreconditionFunc
+
+	waitForObjectStatus := func(event watch.Event) (bool, error) {
+		if event.Type == watch.Deleted {
+			return false, fmt.Errorf("%s was deleted", identifier.String())
+		}
+		// the outer library will handle errors, we have no pod data to review in this case
+		if event.Type != watch.Added && event.Type != watch.Modified {
+			return false, nil
+		}
+		return evaluate(event.Object)
+	}
+
+	waitTimeout, cancel := toolswatch.ContextWithOptionalTimeout(ctx, timeout)
+	defer cancel()
+	_, syncErr := toolswatch.UntilWithSync(waitTimeout, lw, into, existsPrecondition, waitForObjectStatus)
+
+	// Hack to make sure this ends up in the logging client
+	if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: identifier.Namespace, Name: identifier.Name}, into); err != nil {
+		logrus.WithError(err).Warn("failed to get object after finishing watch")
+	}
+
+	return syncErr
+}
+
 func waitForContainer(podClient PodClient, ns, name, containerName string) error {
 	logrus.WithFields(logrus.Fields{
 		"namespace": ns,
@@ -216,34 +265,8 @@ func waitForContainer(podClient PodClient, ns, name, containerName string) error
 
 	ctx := context.TODO()
 
-	lw := &cache.ListWatch{
-		ListFunc: func(options metav1.ListOptions) (object runtime.Object, e error) {
-			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", name).String()
-			res := &corev1.PodList{}
-			return res, podClient.List(ctx, res, &ctrlruntimeclient.ListOptions{Namespace: ns, Raw: &options})
-		},
-		WatchFunc: func(options metav1.ListOptions) (i watch.Interface, e error) {
-			opts := &ctrlruntimeclient.ListOptions{
-				Namespace:     ns,
-				FieldSelector: fields.OneTermEqualSelector("metadata.name", name),
-				Raw:           &options,
-			}
-			return podClient.Watch(ctx, &corev1.PodList{}, opts)
-		},
-	}
-
-	// pods in this call here are always expected to exist in advance
-	var podExistsPrecondition toolswatch.PreconditionFunc
-
-	waitForContainerStatus := func(event watch.Event) (bool, error) {
-		if event.Type == watch.Deleted {
-			return false, fmt.Errorf("pod/%v ns/%v was deleted", name, ns)
-		}
-		// the outer library will handle errors, we have no pod data to review in this case
-		if event.Type != watch.Added && event.Type != watch.Modified {
-			return false, nil
-		}
-		switch pod := event.Object.(type) {
+	evaluatorFunc := func(obj runtime.Object) (bool, error) {
+		switch pod := obj.(type) {
 		case *corev1.Pod:
 			for _, container := range pod.Status.ContainerStatuses {
 				if container.Name == containerName {
@@ -254,21 +277,12 @@ func waitForContainer(podClient PodClient, ns, name, containerName string) error
 				}
 			}
 		default:
-			return false, fmt.Errorf("pod/%v ns/%v got an event that did not contain a pod: %v", name, ns, event.Object)
+			return false, fmt.Errorf("pod/%v ns/%v got an event that did not contain a pod: %v", name, ns, obj)
 		}
 		return false, nil
 	}
 
-	waitTimeout, cancel := toolswatch.ContextWithOptionalTimeout(ctx, 300*5*time.Second /*weird, but this is what it was*/)
-	defer cancel()
-	_, err := toolswatch.UntilWithSync(waitTimeout, lw, &corev1.Pod{}, podExistsPrecondition, waitForContainerStatus)
-
-	// Hack to make sure this ends up in the logging client
-	if err := podClient.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: ns, Name: name}, &corev1.Pod{}); err != nil {
-		logrus.WithError(err).Warn("faild to get pod after finishing watch")
-	}
-
-	return err
+	return waitForConditionOnObject(ctx, podClient, ctrlruntimeclient.ObjectKey{Namespace: ns, Name: name}, &corev1.PodList{}, &corev1.Pod{}, evaluatorFunc, 300*5*time.Second)
 }
 
 func copyArtifacts(podClient PodClient, into, ns, name, containerName string, paths []string) error {
--- a/pkg/steps/artifacts_test.go
+++ b/pkg/steps/artifacts_test.go
@@ -1,6 +1,7 @@
 package steps
 
 import (
+	"context"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
@@ -14,10 +15,13 @@ import (
 	coreapi "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/diff"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/remotecommand"
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakectrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	"github.com/openshift/ci-tools/pkg/junit"
@@ -711,4 +715,102 @@ func TestAddPodUtils(t *testing.T) {
 		t.Errorf("failed to decorate: %v", err)
 	}
 	testhelper.CompareWithFixture(t, base)
+}
+
+func aPod() *coreapi.Pod {
+	return &coreapi.Pod{
+		ObjectMeta: meta.ObjectMeta{
+			Name:      "p",
+			Namespace: "ns",
+		},
+	}
+}
+
+func TestWaitForConditionOnObject(t *testing.T) {
+	containerName := "c"
+	podName := "p"
+	ns := "ns"
+
+	evaluateFunc := func(obj runtime.Object) (bool, error) {
+		switch pod := obj.(type) {
+		case *coreapi.Pod:
+			for _, container := range pod.Status.ContainerStatuses {
+				if container.Name == containerName {
+					if container.State.Running != nil || container.State.Terminated != nil {
+						return true, nil
+					}
+					break
+				}
+			}
+		default:
+			return false, fmt.Errorf("pod/%v ns/%v got an event that did not contain a pod: %v", podName, ns, obj)
+		}
+		return false, nil
+	}
+
+	testCases := []struct {
+		name          string
+		expected      error
+		client        ctrlruntimeclient.WithWatch
+		containerName string
+		objectFunc    func(client ctrlruntimeclient.Client) error
+	}{
+		{
+			name:   "happy path: pod",
+			client: fakectrlruntimeclient.NewFakeClient(aPod()),
+			objectFunc: func(client ctrlruntimeclient.Client) error {
+				// wait for watch being ready
+				time.Sleep(10 * time.Millisecond)
+				ctx := context.TODO()
+				pod := &coreapi.Pod{}
+				if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: podName, Namespace: ns}, pod); err != nil {
+					return err
+				}
+				pod.Status.ContainerStatuses = append(pod.Status.ContainerStatuses, coreapi.ContainerStatus{
+					Name: "c",
+					State: coreapi.ContainerState{
+						Running: &coreapi.ContainerStateRunning{},
+					},
+				})
+				if err := client.Update(ctx, pod); err != nil {
+					return err
+				}
+				return nil
+			},
+		},
+		{
+			name:       "timeout",
+			client:     fakectrlruntimeclient.NewFakeClient(aPod()),
+			objectFunc: func(client ctrlruntimeclient.Client) error { return nil },
+			expected:   fmt.Errorf("timed out waiting for the condition"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			readingDone := make(chan struct{})
+			errChan := make(chan error)
+			var errs []error
+			go func() {
+				for err := range errChan {
+					errs = append(errs, err)
+				}
+				close(readingDone)
+			}()
+			go func() {
+				if err := tc.objectFunc(tc.client); err != nil {
+					errChan <- err
+				}
+			}()
+			actual := waitForConditionOnObject(context.TODO(), tc.client, ctrlruntimeclient.ObjectKey{Name: podName, Namespace: ns}, &coreapi.PodList{}, &coreapi.Pod{}, evaluateFunc, 30*time.Millisecond)
+			close(errChan)
+			<-readingDone
+			if diff := cmp.Diff(tc.expected, actual, testhelper.EquateErrorMessage); diff != "" {
+				t.Errorf("actualError does not match expectedError, diff: %s", diff)
+			}
+			if len(errs) > 0 {
+				t.Errorf("unexpected error occurred: %v", utilerrors.NewAggregate(errs))
+			}
+		})
+	}
 }
--- a/pkg/steps/cluster_claim.go
+++ b/pkg/steps/cluster_claim.go
@@ -10,7 +10,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/test-infra/prow/kube"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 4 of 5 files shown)


pkg/steps/cluster_claim.go
```diff
@@ -180,7 +178,67 @@ func mutate(secret *corev1.Secret, name, namespace string) (*corev1.Secret, erro
 	}, nil
 }
 
-func ClusterClaimStep(as string, clusterClaim *api.ClusterClaim, hiveClient ctrlruntimeclient.Client, client loggingclient.LoggingClient, jobSpec *api.JobSpec) api.Step {
+var (
+	waitForClaim = func(client watchingclient.Client, ns, name string, timeout time.Duration) error {
```
hongkailiu, line 182: I am not sure how to unit testing this.

/hold

Use built image for rehearsals.

pkg/steps/cluster_claim.go
```diff
@@ -180,7 +178,67 @@ func mutate(secret *corev1.Secret, name, namespace string) (*corev1.Secret, erro
...
+			}
+			// the outer library will handle errors, we have no clusterClaim data to review in this case
+			if event.Type != watch.Added && event.Type != watch.Modified {
+				return false, nil
+			}
+			switch clusterClaim := event.Object.(type) {
+			case *hivev1.ClusterClaim:
+				// https://github.com/openshift/hive/blob/a535d29c4d2dc4f7f9bf3f30098c69d9334bee2e/apis/hive/v1/clusterclaim_types.go#L18-L22
+				for _, condition := range clusterClaim.Status.Conditions {
+					logrus.WithField("claim.Name", name).WithField("claim.Namespace", ns).
```
stevekuznetsov, line 221: This log line will just be spammy ... are you sure we need it? The logic for this watch processing is one `if` statement. It either works or it doesn't ;)

pkg/steps/cluster_claim.go
stevekuznetsov: Why is this in a global?

cc @alvaroaleman for how to unit-test

pkg/steps/cluster_claim.go
hongkailiu: Removed.

pkg/steps/cluster_claim.go
hongkailiu: To be precise, I'd like to unit-test the caller `func (s *clusterClaimStep) run(ctx context.Context) error` of this function.
The function is global because I want to override it in the unit tests of the caller to avoid timeout.

hongkailiu: Make it as step's field. See if it is better.

pkg/steps/cluster_claim.go
```diff
@@ -205,14 +201,69 @@ func releaseCluster(ctx context.Context, hiveClient ctrlruntimeclient.Client, cl
...
-func ClusterClaimStep(as string, clusterClaim *api.ClusterClaim, hiveClient ctrlruntimeclient.Client, client loggingclient.LoggingClient, jobSpec *api.JobSpec, wrapped api.Step) api.Step {
+func ClusterClaimStep(as string, clusterClaim *api.ClusterClaim, hiveClient watchingclient.Client, client loggingclient.LoggingClient, jobSpec *api.JobSpec, wrapped api.Step) api.Step {
 	ret := clusterClaimStep{
 		as:           as,
 		clusterClaim: clusterClaim,
 		hiveClient:   hiveClient,
 		client:       client,
 		jobSpec:      jobSpec,
 		wrapped:      wrapped,
+		waitForClaim: func(client watchingclient.Client, ns, name string, timeout time.Duration) error {
```
stevekuznetsov, line 212: In every case that we write code that looks identical in two places we need to ask if it's attempting to do the same thing, and, if so, how we can make it share the core structure. This code is identical to that which David (Alvaro?) wrote for containers. Consider something like the following:

```go

type evaluator func(obj runtime.Object) (bool, error)

// waitForConditionOnObject uses a watch to wait for a condition to be true on an object.
// When the condition is satisfied or the timeout expires, the object is returned along
// with any errors encountered.
func waitForConditionOnObject(ctx context.Context, client watchingclient.Client, identifier ctrlruntimeclient.ObjectKey, list *ctrlruntimeclient.ObjectList, into ctrlruntimeclient.Object, evaluate evaluator, timeout time.Duration) error {
	lw := &cache.ListWatch{
		ListFunc: func(options metav1.ListOptions) (object runtime.Object, e error) {
			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", identifier.Name).String()
			res := *list
			return res, client.List(ctx, res, &ctrlruntimeclient.ListOptions{Namespace: identifier.Namespace, Raw: &options})
		},
		WatchFunc: func(options metav1.ListOptions) (i watch.Interface, e error) {
			opts := &ctrlruntimeclient.ListOptions{
				Namespace:     identifier.Namespace,
				FieldSelector: fields.OneTermEqualSelector("metadata.name", identifier.Name),
				Raw:           &options,
			}
			res := *list
			return client.Watch(ctx, res, opts)
		},
	}

	// pods in this call here are always expected to exist in advance
	var podExistsPrecondition toolswatch.PreconditionFunc

	waitForContainerStatus := func(event watch.Event) (bool, error) {
		if event.Type == watch.Deleted {
			return false, fmt.Errorf("%s was deleted", identifier.String())
		}
		// the outer library will handle errors, we have no data to review in this case
		if event.Type != watch.Added && event.Type != watch.Modified {
			return false, nil
		}
		return evaluate(event.Object)
	}

	waitTimeout, cancel := toolswatch.ContextWithOptionalTimeout(ctx, timeout)
	defer cancel()
	res := *list
	_, syncErr := toolswatch.UntilWithSync(waitTimeout, lw, res, podExistsPrecondition, waitForContainerStatus)

	// We need to make sure this ends up in the logging client
	if err := client.Get(ctx, identifier, into); err != nil {
		logrus.WithError(err).Warn("failed to get object after finishing watch")
	}
	return syncErr
}
```

hongkailiu: After 2 abstraction that you have given me, I hope I can learn the level of abstraction.

stevekuznetsov:
Part of DPTP-2099

pkg/steps/artifacts.go
```diff
@@ -207,43 +208,66 @@ func (c podClient) WithNewLoggingClient() PodClient {
...
+// When the condition is satisfied or the timeout expires, the object is returned along
+// with any errors encountered.
+func waitForConditionOnObject(ctx context.Context, client watchingclient.Client, identifier ctrlruntimeclient.ObjectKey, list ctrlruntimeclient.ObjectList, into ctrlruntimeclient.Object, evaluate evaluator, timeout time.Duration) error {
 	lw := &cache.ListWatch{
 		ListFunc: func(options metav1.ListOptions) (object runtime.Object, e error) {
-			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", name).String()
-			res := &corev1.PodList{}
-			return res, podClient.List(ctx, res, &ctrlruntimeclient.ListOptions{Namespace: ns, Raw: &options})
+			options.FieldSelector = fields.OneTermEqualSelector("metadata.name", identifier.Name).String()
+			res := list
```
stevekuznetsov, line 220: Passing in a pointer to the list and dereferencing it to copy in my example was important - I don't know what happens when you don't do that and use the same list object in all areas. cc @alvaroaleman perhaps you have a good suggestion on how to test this?

pkg/steps/artifacts.go
```diff
@@ -207,43 +208,66 @@ func (c podClient) WithNewLoggingClient() PodClient {
...
+				FieldSelector: fields.OneTermEqualSelector("metadata.name", identifier.Name),
 				Raw:           &options,
 			}
-			return podClient.Watch(ctx, &corev1.PodList{}, opts)
+			res := list
+			return client.Watch(ctx, res, opts)
 		},
 	}
 
 	// pods in this call here are always expected to exist in advance
```
stevekuznetsov, line 234: Comments should not be type-specific.

pkg/steps/artifacts.go
```diff
@@ -207,43 +208,66 @@ func (c podClient) WithNewLoggingClient() PodClient {
...
 		if event.Type != watch.Added && event.Type != watch.Modified {
 			return false, nil
 		}
-		switch pod := event.Object.(type) {
+		return evaluate(event.Object)
+	}
+
+	waitTimeout, cancel := toolswatch.ContextWithOptionalTimeout(ctx, timeout /*weird, but this is what it was*/)
+	defer cancel()
+	_, syncErr := toolswatch.UntilWithSync(waitTimeout, lw, &corev1.Pod{}, existsPrecondition, waitForContainerStatus)
```
stevekuznetsov, line 250: You cannot use `&corev1.Pod{}` here, this is a generic function.

pkg/steps/artifacts.go
```diff
@@ -207,43 +208,66 @@ func (c podClient) WithNewLoggingClient() PodClient {
...
 		}
 		// the outer library will handle errors, we have no pod data to review in this case
 		if event.Type != watch.Added && event.Type != watch.Modified {
 			return false, nil
 		}
-		switch pod := event.Object.(type) {
+		return evaluate(event.Object)
+	}
+
+	waitTimeout, cancel := toolswatch.ContextWithOptionalTimeout(ctx, timeout /*weird, but this is what it was*/)
```
stevekuznetsov, line 248: Remove David's comment please.

pkg/steps/cluster_claim.go
```diff
@@ -21,15 +21,17 @@ import (
...
 
 type clusterClaimStep struct {
 	as           string
 	clusterClaim *api.ClusterClaim
-	hiveClient   ctrlruntimeclient.Client
+	hiveClient   watchingclient.Client
 	client       loggingclient.LoggingClient
 	jobSpec      *api.JobSpec
 	wrapped      api.Step
+	waitForClaim func(client watchingclient.Client, ns, name string, timeout time.Duration) error
```
stevekuznetsov, line 34: I do not understand why this is a mutable field on the step. Please explain?

pkg/steps/cluster_claim.go
```diff
@@ -122,23 +124,12 @@ func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClie
...
-				WithField("condition.Type", condition.Type).WithField("condition.Status", condition.Status).
-				Debug("Checking the claim's status.")
-			if condition.Type == hivev1.ClusterRunningCondition && condition.Status == corev1.ConditionTrue {
-				return true, nil
-			}
-		}
-		return false, nil
-	}); err != nil {
-		return nil, fmt.Errorf("failed to wait for created cluster claim to become running: %w", err)
+	if err := hiveClient.Get(ctx, ctrlruntimeclient.ObjectKey{Name: claimName, Namespace: claimNamespace}, claim); err != nil {
```
stevekuznetsov, line 131: The `into` is set by the generic function. You do not need this.

pkg/steps/artifacts.go
hongkailiu: Updated.

pkg/steps/artifacts.go
hongkailiu: Updated.

pkg/steps/artifacts.go
hongkailiu: Updated.

pkg/steps/cluster_claim.go
hongkailiu: Updated.

pkg/steps/cluster_claim.go
hongkailiu: Originally, I need to override the function for testing `run` function of the step.
But now we test `acquireCluster` instead and the function `waitForClaim ` is passed as an input.

`waitForClaim` becomes now a hard-coded function in `run`.

pkg/steps/artifacts.go
hongkailiu: `list ctrlruntimeclient.ObjectList`

It is an interface. So a pointer already?

stevekuznetsov: Right, but the _dereference_ is the important part, since it will copy. Re-assigning an interface value to a new variable does not do that.

hongkailiu: `list *ctrlruntimeclient.ObjectList`: I cannot make the caller to compile if I use it as a type of a param in `waitForConditionOnObject` function.

hongkailiu: https://github.com/kubernetes-sigs/controller-runtime/pull/1510
Used it for unit test.

pkg/defaults/defaults.go
```diff
@@ -89,9 +89,9 @@ func FromConfig(
 
 	podClient := steps.NewPodClient(client, clusterConfig, coreGetter.RESTClient())
 
-	var hiveClient ctrlruntimeclient.Client
+	var hiveClient watchingclient.Client
```
alvaroaleman, line 92:
```suggestion
	var hiveClient ctrlruntimeclient.WithWatch
```

hongkailiu: Done.

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +719,108 @@ func TestAddPodUtils(t *testing.T) {
...
+		{
+			name:       "happy path: pod",
+			client:     fakectrlruntimeclient.NewFakeClient(aPod()),
+			identifier: ctrlruntimeclient.ObjectKey{Name: podName, Namespace: ns},
+			list:       &coreapi.PodList{},
+			into:       &coreapi.Pod{},
+			timeout:    time.Second,
+			evaluate:   evaluateFunc,
+			objectFunc: func(client ctrlruntimeclient.Client) {
+				_ = wait.Poll(100*time.Millisecond, time.Second,
```
alvaroaleman, line 775: Please somehow handle the error, e.G. by passing in a `*testing.T`

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +719,108 @@ func TestAddPodUtils(t *testing.T) {
...
+			evaluate:   evaluateFunc,
+			objectFunc: func(client ctrlruntimeclient.Client) {},
+			expected:   fmt.Errorf("timed out waiting for the condition"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			var wg sync.WaitGroup
+			wg.Add(1)
```
alvaroaleman, line 815: A `waitGroup` whose `Wait` is never called is useless. I don't see why we need synchronization here though

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +719,108 @@ func TestAddPodUtils(t *testing.T) {
...
+						}
+						return true, nil
+					},
+				)
+			},
+		},
+		{
+			name:       "timeout",
+			client:     fakectrlruntimeclient.NewFakeClient(aPod()),
+			identifier: ctrlruntimeclient.ObjectKey{Name: podName, Namespace: ns},
```
alvaroaleman, line 802: Just a nit, but all of `identifier`, `list`, `into`, `timeout` and `evaluate` or the same for all testcases, so it doesn't really make sense to make them a property

pkg/steps/artifacts_test.go
hongkailiu: I was jumping between implementation ideas.

pkg/steps/artifacts_test.go
hongkailiu: Updated.

pkg/steps/artifacts_test.go
hongkailiu: I have an impression that we can call `t.errorf` on in the main thread.
I will try to channel up the error.

alvaroaleman: No, you can call `Errorf` anywhere, you just cannot call `Fatalf` from a goroutine

hongkailiu: Updated.

pkg/steps/cluster_claim.go
```diff
@@ -61,7 +61,33 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
 		// should never happen
 		return fmt.Errorf("cannot claim a nil cluster")
 	}
-	clusterClaim, err := acquireCluster(ctx, *s.clusterClaim, s.hiveClient, s.client, *s.jobSpec)
+	waitForClaim := func(client ctrlruntimeclient.WithWatch, ns, name string, claim *hivev1.ClusterClaim, timeout time.Duration) error {
```
stevekuznetsov, line 64: Why does this need to be a closure as opposed to a method on `*clusterClaimStep`?

hongkailiu: Updated.

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +717,92 @@ func TestAddPodUtils(t *testing.T) {
...
+	}{
+		{
+			name:   "happy path: pod",
+			client: fakectrlruntimeclient.NewFakeClient(aPod()),
+			objectFunc: func(client ctrlruntimeclient.Client, t *testing.T) {
+				if err := wait.Poll(100*time.Millisecond, time.Second,
+					func() (bool, error) {
+						ctx := context.TODO()
+						pod := &coreapi.Pod{}
+						if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: podName, Namespace: ns}, pod); err != nil {
```
stevekuznetsov, line 767: What creates this Pod?

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +717,92 @@ func TestAddPodUtils(t *testing.T) {
...
+			expected:   fmt.Errorf("timed out waiting for the condition"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			go func() {
+				tc.objectFunc(tc.client, t)
+			}()
+			actual := waitForConditionOnObject(context.TODO(), tc.client, ctrlruntimeclient.ObjectKey{Name: podName, Namespace: ns}, &coreapi.PodList{}, &coreapi.Pod{}, evaluateFunc, time.Second)
```
stevekuznetsov, line 802: 1s is two or three orders of magnitude longer than the runtime for most unit tests.

pkg/steps/cluster_claim.go
```diff
@@ -61,33 +61,59 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
...
+	}
+
+	clusterClaim, err := s.acquireCluster(ctx, waitForClaim)
 	if err != nil {
 		acquireErr := results.ForReason("acquiring_cluster_claim").ForError(err)
 		// always attempt to delete claim if one exists
 		var releaseErr error
 		if clusterClaim != nil {
-			releaseErr = results.ForReason("releasing_cluster_claim").ForError(releaseCluster(ctx, s.hiveClient, clusterClaim, s.Name()))
+			releaseErr = results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim))
```
stevekuznetsov, line 96: Does `ForReason().ForError()` return `nil` when the error is `nil`?

pkg/steps/cluster_claim.go
```diff
@@ -61,33 +61,59 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
...
 
 	wrappedErr := results.ForReason("executing_test").ForError(s.wrapped.Run(ctx))
-	releaseErr := results.ForReason("releasing_cluster_claim").ForError(releaseCluster(ctx, s.hiveClient, clusterClaim, s.Name()))
+	releaseErr := results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim))
 
 	return aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr)
 }
 
-func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClient ctrlruntimeclient.Client, client loggingclient.LoggingClient, jobSpec api.JobSpec) (*hivev1.ClusterClaim, error) {
+func (s *clusterClaimStep) acquireCluster(ctx context.Context, waitForClaim func(client ctrlruntimeclient.WithWatch, ns, name string, claim *hivev1.ClusterClaim, timeout time.Duration) error) (*hivev1.ClusterClaim, error) {
```
stevekuznetsov, line 107: If `acquireCluster` is now a method on `*clusterClaimStep` why do we construct the waiting func outside of this and pass it in?

pkg/steps/cluster_claim.go
```diff
@@ -99,71 +125,59 @@ func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClie
 	}
 
 	clusterPool := clusterPools.Items[0]
-	claimName := jobSpec.ProwJobID
+	claimName := s.jobSpec.ProwJobID
 	claimNamespace := clusterPool.Namespace
-	claim := &hivev1.ClusterClaim{
+	ret := &hivev1.ClusterClaim{
```
stevekuznetsov, line 130: `ret` is a bad name as it's not descriptive. Why not `claim`?

pkg/steps/cluster_claim.go
```diff
@@ -99,71 +125,59 @@ func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClie
...
-		return false, nil
-	}); err != nil {
-		return claim, fmt.Errorf("failed to wait for created cluster claim to become running: %w", err)
+	claim := &hivev1.ClusterClaim{}
+	if err := waitForClaim(s.hiveClient, claimNamespace, claimName, claim, s.clusterClaim.Timeout.Duration); err != nil {
+		return ret, fmt.Errorf("failed to wait for created cluster claim to become ready: %w", err)
 	}
+	ret = claim
 	logrus.Info("The claimed cluster is ready.")
 	clusterDeploymentName := claim.Spec.Namespace
```
stevekuznetsov, line 154: What's the utility in this local variable?

pkg/steps/cluster_claim.go
```diff
@@ -99,71 +125,59 @@ func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClie
...
-	if err := hiveClient.Get(ctx, ctrlruntimeclient.ObjectKey{Name: clusterDeploymentName, Namespace: clusterDeploymentNamespace}, clusterDeployment); err != nil {
-		return claim, fmt.Errorf("failed to get cluster deployment %s in namespace %s: %w", clusterDeploymentName, clusterDeploymentNamespace, err)
+	if err := s.hiveClient.Get(ctx, ctrlruntimeclient.ObjectKey{Name: clusterDeploymentName, Namespace: clusterDeploymentNamespace}, clusterDeployment); err != nil {
+		return ret, fmt.Errorf("failed to get cluster deployment %s in namespace %s: %w", clusterDeploymentName, clusterDeploymentNamespace, err)
 	}
 	if clusterDeployment.Spec.ClusterMetadata == nil {
-		return claim, fmt.Errorf("got nil cluster metadata from cluster deployment %s in namespace %s", clusterDeploymentName, clusterDeploymentNamespace)
+		return ret, fmt.Errorf("got nil cluster metadata from cluster deployment %s in namespace %s", clusterDeploymentName, clusterDeploymentNamespace)
 	}
 	kubeconfigSecretName := clusterDeployment.Spec.ClusterMetadata.AdminKubeconfigSecretRef.Name
```
stevekuznetsov, line 163: What is the utility in these local variables?

pkg/steps/cluster_claim.go
```diff
@@ -99,71 +125,59 @@ func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClie
...
-			return claim, fmt.Errorf("failed to upsert immutable secret %s in namespace %s: %w", dstSecret.Name, dstSecret.Namespace, err)
+		if _, err := util.UpsertImmutableSecret(ctx, s.client, dstSecret); err != nil {
+			return ret, fmt.Errorf("failed to upsert immutable secret %s in namespace %s: %w", dst, dstNS, err)
 		}
 	}
-	return claim, nil
+	return ret, nil
 }
 
 func mutate(secret *corev1.Secret, name, namespace string) (*corev1.Secret, error) {
```
stevekuznetsov, line 183: nit: if you pass a pointer into a function called `mutate` you expect the underlying object to be mutated, not returned

stevekuznetsov: also: `mutate` is a poor function name, since it does not describe whatsoever _what_ is going on or _why_

pkg/steps/cluster_claim.go
```diff
@@ -186,24 +200,23 @@ func mutate(secret *corev1.Secret, name, namespace string) (*corev1.Secret, erro
 	}, nil
 }
 
-func releaseCluster(ctx context.Context, hiveClient ctrlruntimeclient.Client, clusterClaim *hivev1.ClusterClaim, testName string) error {
-	logrus.Infof("Releasing cluster claims for test %s", testName)
+func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hivev1.ClusterClaim) error {
+	logrus.Infof("Releasing cluster claims for test %s", s.Name())
 	logrus.WithField("clusterClaim.Namespace", clusterClaim.Namespace).WithField("clusterClaim.Name", clusterClaim.Name).Debug("Deleting cluster claim.")
-	if err := hiveClient.Delete(ctx, clusterClaim); err != nil {
+	if err := s.hiveClient.Delete(ctx, clusterClaim); err != nil {
```
stevekuznetsov, line 206: Likely worth it to retry a couple times on this?

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +717,92 @@ func TestAddPodUtils(t *testing.T) {
...
+								Running: &coreapi.ContainerStateRunning{},
+							},
+						})
+						if err := client.Update(ctx, pod); err != nil {
+							return false, err
+						}
+						return true, nil
+					},
+				); err != nil {
+					t.Errorf("unexpected error occurred in objectFunc: %v", err)
```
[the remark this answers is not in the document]
hongkailiu: 😄 
https://github.com/openshift/ci-tools/pull/1874#discussion_r627728178

pkg/steps/artifacts_test.go
hongkailiu: It comes with the fake client.

https://github.com/openshift/ci-tools/pull/1874/files#diff-86a943161d4f6ebe47a0c7efd74c4ac67371f1f4e99d4a32c8ea56e4d859f1ccR761

pkg/steps/artifacts_test.go
```diff
@@ -712,3 +717,92 @@ func TestAddPodUtils(t *testing.T) {
...
+		expected      error
+		client        ctrlruntimeclient.WithWatch
+		containerName string
+		objectFunc    func(client ctrlruntimeclient.Client, t *testing.T)
+	}{
+		{
+			name:   "happy path: pod",
+			client: fakectrlruntimeclient.NewFakeClient(aPod()),
+			objectFunc: func(client ctrlruntimeclient.Client, t *testing.T) {
+				if err := wait.Poll(100*time.Millisecond, time.Second,
```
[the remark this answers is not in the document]
hongkailiu: Updated.

pkg/steps/artifacts_test.go
hongkailiu: Updated.

pkg/steps/cluster_claim.go
```diff
@@ -61,33 +61,59 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
...
 		return fmt.Errorf("cannot claim a nil cluster")
 	}
-	clusterClaim, err := acquireCluster(ctx, *s.clusterClaim, s.hiveClient, s.client, *s.jobSpec)
+	waitForClaim := func(client ctrlruntimeclient.WithWatch, ns, name string, claim *hivev1.ClusterClaim, timeout time.Duration) error {
+		logrus.WithFields(logrus.Fields{
+			"namespace": ns,
+			"name":      name,
+		}).Trace("Waiting for claim to be running.")
+
+		ctx := context.TODO()
```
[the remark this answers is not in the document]
hongkailiu: Removed.

pkg/steps/cluster_claim.go
hongkailiu: We keep the created one to return because it will be deleted no matter what happens in the acquire function.
Renamed the var `into` which is passed to the acquire function

pkg/steps/cluster_claim.go
hongkailiu: For improve the readability of the code.
Removed.

pkg/steps/cluster_claim.go
hongkailiu: Renamed.

pkg/steps/cluster_claim.go
hongkailiu: Add retry.
Why it could fail in the first place? Do we do the retry on other object deleting?

hongkailiu:
https://issues.redhat.com/browse/DPTP-2099

PR #1878, opened 2021-04-15
pod-scaler: a server for automated resource requests

stevekuznetsov:
This commit introduces a new server that will automate setting resource

recommendations and can also serve a web UI.

This commit contains the producer code. In order to provide an estimate
of resource usage for containers in a CI job, this server analyzes
metrics from previous executions of similar containers. Aggregate
statistics are used to provide resource request recommendations by
digesting prior metrics. It is assumed that, for a sufficiently similar
container, resource usage will not vary much across executions - we
expect this to be true for e.g. all executions of unit tests for some
branch on a repository. This assumption allows for samples from all
executions to be treated as one dataset with a single underlying
distribution, so that aggregation can be done on the larger dataset to
yield higher-fidelity signal. The overall size of the raw data, however,
quickly grows unmanageable. In order to operate efficiently on this
dataset we therefore store compressed histograms for each execution
trace. This allows us to reduce the data footprint while continuing
to allow for dataset merging and aggregation. The Circonus log-linear
histogram is used as it's performant, accurate, efficient and open-
source.


cmd/pod-scaler/producer.go
```diff
@@ -0,0 +1,482 @@
...
+					// We hit a timeout asking for a known large value,
+					exitEarly = true
+				}
+				c.lock.Unlock()
+				if exitEarly {
+					subdivide()
+					return
+				}
+			}
+		}
```
petr-muller, line 278: This is quite hard to read. I played with it a bit in an editor and I think the flow can be simplified like this to reduce the amount of branching and indentation (proposing, I don't insist). Could be even more straightforward using a switch with fallthrough, but I just hate fallthrough in Go.

```go
		// Prometheus determined not to expose this programmatically ...
		if errors.As(err, &apiError) && strings.HasSuffix(apiError.Msg, "504") {
			c.lock.Lock()
			divide := func() { c.lock.Unlock(); subdivide() }
			if numSteps >= c.maxSize {
				// We hit a timeout asking for a known large value,
				divide()
				return
			}
			if numSteps > 250 {
				// We hit a timeout and are still asking for a reasonably "large" amount of
				// data at once, so halve the amount of data we are asking for in order to
				// have a higher chance of getting the data next time. If we're asking for
				// a small amount already it's likely the server is on the verge of falling
				// over, so just error out and try again later.
				logger.Debugf("Received 504 asking for %d samples, halving to %d.", numSteps, numSteps/2)
				c.maxSize = numSteps
				divide()
				return
			}
			logger.Debugf("Received 504 but only asking for %d samples, aborting.", numSteps)
			c.lock.Unlock()
		}
```

cmd/pod-scaler/producer.go
```diff
@@ -0,0 +1,482 @@
...
+					End:   input[i].End,
+				}
+			}
+			if coalesced != nil {
+				return append(input[:i], append(input[i+1:j], append(input[j+1:], *coalesced)...)...)
+			}
+		}
+	}
+	return input
+}
```
petr-muller, line 482: If you didn't make an off-by-one, unwanted overlap, out of bounds or something similar somewhere here I'll buy you a beer next time we meet.

petr-muller (commented):
I like the structure and interfaces, very clean and understandable. Some comments inline.

cmd/pod-scaler/storage.go
```diff
@@ -0,0 +1,136 @@
...
+	LastUpdated(name string) (time.Time, error)
+}
+
+type bucketCache struct {
+	bucket *storage.BucketHandle
+}
+
+var _ Cache = &bucketCache{}
+
+func (b *bucketCache) Load(name string) (io.ReadCloser, error) {
```
alvaroaleman, line 47: Shouldn't these take a `context`? I recall that we did have cases where we got stuck trying to read from GCS in prow occasioanlly.

cmd/pod-scaler/storage.go
```diff
@@ -0,0 +1,136 @@
...
+	handle := b.bucket.Object(name)
+	return handle.NewReader(interrupts.Context())
+}
+
+func (b *bucketCache) Store(name string) (io.WriteCloser, error) {
+	handle := b.bucket.Object(name)
+	return handle.NewWriter(context.Background()), nil
```
alvaroaleman, line 54: Same here re `context`, caller should give a reasonable timeout and retry

cmd/pod-scaler/storage.go
```diff
@@ -0,0 +1,136 @@
...
+	if err != nil {
+		return fmt.Errorf("could open cache for writing: %w", err)
+	}
+	if _, err := writer.Write(raw); err != nil {
+		return fmt.Errorf("could not write cached data: %w", err)
+	}
+	if err := writer.Close(); err != nil {
+		return fmt.Errorf("could not write cached data: %w", err)
+	}
+	logger.Infof("Flushed Prometheus data to cache after %s.", time.Since(flushStart))
```
alvaroaleman, line 134: Since this is going to run in a server we should use proper structured logging instead of `printf` IMHO

cmd/pod-scaler/producer.go
```diff
@@ -0,0 +1,482 @@
...
+					client: client,
+					lock:   &sync.RWMutex{},
+					// there's absolutely no chance Prometheus at the current scaling will ever be able
+					// to respond to large requests it's completely capable of creating, so don't even
+					// bother asking for anything larger than 1/10th of the largest request we can
+					maxSize: MaxSamplesPerRequest / 10,
+					errors:  make(chan error),
+					// there's also no chance that Prometheus will be able to handle any real concurrent
+					// request volume, so don't even bother trying to request more samples at once than
+					// a third of the maximum samples it can technically provide in one request
```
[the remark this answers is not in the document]
stevekuznetsov: This was stable across my testing but I don't know that it can generalize forever - I also don't know that we can do anything more intelligent, though.

cmd/pod-scaler/producer.go
stevekuznetsov: There are tests! It works!

cmd/pod-scaler/storage.go
stevekuznetsov: We implicitly use the `interrupts.Context()` - the only case where we want to exit early is if the binary is terminated.

cmd/pod-scaler/storage.go
stevekuznetsov: Timeout and retry are implicit given the periodic nature of the tool - do we need anything more?

cmd/pod-scaler/storage.go
stevekuznetsov: I agree - but durations should be metrics, not logs anyway. I at least use structured log fields for filtering, so e.g. `org` and `repo` are useful in Prow but not `duration`. I expect to extend this thing to expose all these durations as counters in Prom for inspection.

cmd/pod-scaler/storage.go
alvaroaleman: Retry maybe, timeout not. If this hangs it will hang forever which did happen in the past in tide. The timeout could be very high, but it can not be indefinite.

cmd/pod-scaler/storage.go
alvaroaleman: See below

cmd/pod-scaler/storage.go
alvaroaleman: No one is going to filter for a duration but they are going to filter for this log message. Which with this implementation requires regex/glob filtering.

cmd/pod-scaler/producer.go
stevekuznetsov: I think I went the way I did because there is one and only one return, which means you can more simply reason about locking and unlocking the critical section. I'll play around with making this `if` less nested though.

stevekuznetsov:
@petr-muller @alvaroaleman updated for comments

alvaroaleman (approved):

stevekuznetsov:
Random flake in e2e ... I don't *really* want to wait for another ...

stevekuznetsov:
There's no way this can break anything, it's a new tool. Waiting 2h+ for this to merge is going to ruin my productivity today ...

stevekuznetsov:
Part of DPTP-1822

PR #1882, opened 2021-04-16
prowgen: add release-informing job type

AlexNPavel:
This PR adds support for defining jobs as release-informing to prevent
confusion when a periodic job is configured via release-controller
instead of a prowjob or ci-operator config. A `release-controller` PR
will be created to validate that all jobs defined in the
release-controller periodics are appropriately labelled as release
informers and that all jobs with the label are in the release-controller
periodics configs.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -525,6 +525,12 @@ type TestStepConfiguration struct {
 	// create a periodic job instead of a presubmit
 	Interval *string `json:"interval,omitempty"`
 
+	// ReleaseController configures prowgen to create a periodic that
+	// does not get run by prow and instead is run by release-controller.
+	// The job must be configured as a verification or periodic job in a
+	// release-controller config file when this field is set to `true`.
+	ReleaseController bool `json:"release_controller,omitempty"`
+
 	// Postsubmit configures prowgen to generate the job as a postsubmit rather than a presubmit
 	Postsubmit bool `json:"postsubmit,omitempty"`
 
--- a/pkg/jobconfig/files.go
+++ b/pkg/jobconfig/files.go
@@ -21,14 +21,16 @@ import (
 )
 
 const (
-	CanBeRehearsedLabel = "pj-rehearse.openshift.io/can-be-rehearsed"
-	CanBeRehearsedValue = "true"
-	SSHBastionLabel     = "dptp.openshift.io/ssh-bastion"
-	ProwJobLabelVariant = "ci-operator.openshift.io/variant"
-	JobReleaseKey       = "job-release"
-	PresubmitPrefix     = "pull"
-	PostsubmitPrefix    = "branch"
-	PeriodicPrefix      = "periodic"
+	CanBeRehearsedLabel    = "pj-rehearse.openshift.io/can-be-rehearsed"
+	CanBeRehearsedValue    = "true"
+	SSHBastionLabel        = "dptp.openshift.io/ssh-bastion"
+	ProwJobLabelVariant    = "ci-operator.openshift.io/variant"
+	ReleaseControllerLabel = "ci-operator.openshift.io/release-controller"
+	ReleaseControllerValue = "true"
+	JobReleaseKey          = "job-release"
+	PresubmitPrefix        = "pull"
+	PostsubmitPrefix       = "branch"
+	PeriodicPrefix         = "periodic"
 )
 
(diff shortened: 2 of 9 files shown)


pkg/api/types.go
```diff
@@ -525,6 +525,12 @@ type TestStepConfiguration struct {
 	// create a periodic job instead of a presubmit
 	Interval *string `json:"interval,omitempty"`
 
+	// ReleaseInforming configures prowgen to create a periodic that
+	// does not get run by prow and instead is run by release-controller.
+	// The job must be configured as a periodic in a release-controller when
+	// this field is set to `true`.
+	ReleaseInforming bool `json:"release_informing,omitempty"`
```
petr-muller, line 532: Let's call it something else than "release informing"... That term already has a meaning and (already, too) confuses people. Some of these jobs will be blocking. How about simply `ReleaseController`, as in "This jobs is to be triggered by release controller?". 


Also, how about the jobs that are triggerd by release-controller but not periodically (ones that do get triggered for every release). Thease jobs do run even via horologium today, and therefore would have cron/interval, but they should also have this boolean set to true, no?

petr-muller (commented):
I am :grimacing: every time we add a new field to `TestStepConfiguration` that gets validated with criteria like "must be set iff _other fielld_ is not set". I think we're hitting the complexity limits on these fields, and should really think about how to make these structs extensible for the future, because adding more fields that are mutually exclustive with other fields are not going to cut it.

(this is not a critique of this individual PR)

Otherwise I jsut think we should name the thing differently.

pkg/api/types.go
AlexNPavel: The jobs that are triggered by releases (verification jobs) are renamed based on the release (the new name being "releaseName-testName", where "testName" is the key in the verification job map), so they don't conflict with what horologium or the release-controller periodics do. While the periodic job definition is still in a map like the verification jobs and thus each job has a test name in the config, we don't actually rename the job for periodics (that would be confusing for end users). So this conflict check is only necessary for the release-controller periodic jobs that would conflict with horologium, not the verification ones.

petr-muller: Huh, can you explain the renaming? That's new for me. I thought that r-c config simply refers to any existing jobs, and triggers them either per-release (verification jobs) or periodically (r-c triggered periodics). Do you mean we rename them in the migrator tool? 

I'm trying to look at it from the point of view of the user, who will need to bear the complexity of deciding when to set this field to `true`. An obvious thought is when you set up an informing job - which right now is considered to be pretty much any job in o/release. unless explicitly marked as "broken". They do not even need to add the job to the r-c config, many people are fine just running it as a Prow periodic (and the job still shows in testgrid, which people mostly associate with "informing jobs").

For example, consider the CI 4.8 `e2e-gcp-compact` job. It's a "normal" Prow periodic with a 48h interval. It's not triggered by r-c. It shows up on [informing testgrid](https://testgrid.k8s.io/redhat-openshift-ocp-release-4.8-informing#periodic-ci-openshift-release-master-ci-4.8-e2e-gcp-compact) though, so for a user, it's an informing job. Would that job have `release_informing: true`? From your explanation I understand that no, it would not. But how should the user know?

https://github.com/openshift/release/blob/080a3aca7b9cdafb7fb3896f5355913cb8dc1aab/ci-operator/config/openshift/release/openshift-release-master__ci-4.8.yaml#L112-L118

AlexNPavel: I got misremembered how verification job naming worked. The modified name was the object name (by default it would be UUID, release-controller hardcodes it). The job name remains the same. This actually does mean that there is a bit of a conflict between periodics and verification jobs, but quite minor (periodics that run on intervals will run X hours after the last run of the job, regardless of whether that run was a periodic or verification). This is something we can deal with at another time (we should probably disallow using the same job for both verification and periodic jobs; currently we only check for duplicates within the periodic jobs).

It makes sense to change the field to make it for all release controller jobs regardless of type though. I've updated this PR to have a `release_controller` field instead of `release_informing` and will update the release-controller PR to match

stevekuznetsov: Jobs that are pure periodics cannot be associated with specific releases by the release controller, so I don't think we should allow them to be in informing dashboards ...

stevekuznetsov: And, with the counterpart PR, this is enforced. Good.

PR #1883, opened 2021-04-16
ci-operator: disable caching on empty input SHAs

petr-muller:
Prowjobs with `extra_refs` do not contain SHAs (periodics and
rehearsals, among others), so we end up using an empty string as a hash
input ([example](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-odo-master-v4.5-integration-e2e-periodic/1383118121950253056#1:build-log.txt%3A3)) and sometimes hit cache when we should not.

We can use time as an artificial input in such case, effectively
disabling the cache.

Alternatively, we could fix up Prowjob producers to always resolve SHAs
for refs, but even in that case this chance protects us from silently
doing the wrong thing on empty input.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -536,6 +536,10 @@ func (o *options) Complete() error {
 		logrus.Info("No source defined")
 	}
 	for _, ref := range refs {
+		if ref.BaseSHA == "" {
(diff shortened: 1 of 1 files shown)


2uasimojo:
:eye:

alvaroaleman (approved):

PR #1893, opened 2021-04-20
Switching to base64.StdEncoding.DecodeString() for decoding binary file content

bradmwilliams:
The current method appears to be subtly changing the original encoded string just enough to make the resultant file not be recognized as `binary` and ultimately it fails to load via openpgp.  Switching to the `DecodeString` method produces the exact same output as the string value that is stored in Vault.


alvaroaleman (approved):

stevekuznetsov:
wat

PR #1895, opened 2021-04-21
pod-scaler: add labels to Build Pods via admission

stevekuznetsov:
We need labels on Pods to associate individual executions with their
semantic category. Today, we label Builds (as that's what we create) but
the Build subsystem does not carry those labels forward to the Pods that
implement the Build. This patch adds a new mutating admission webhook
server that will add the labels where necessary.


cmd/pod-scaler/admission.go
```diff
@@ -0,0 +1,211 @@
+package main
```
alvaroaleman, line 1: I am not a fan of handrolling stuff we could just use a lib for like e.G. in https://github.com/kubernetes-sigs/controller-runtime/blob/master/examples/builtins/mutatingwebhook.go

cmd/pod-scaler/admission.go
```diff
@@ -0,0 +1,94 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/rest"
+	"k8s.io/test-infra/prow/interrupts"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	buildv1 "github.com/openshift/api/build/v1"
+	buildclientset "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
+	buildclientv1 "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
```
petr-muller, line 20: nit: are these two lines the same thing?

cmd/pod-scaler/admission.go
```diff
@@ -0,0 +1,94 @@
...
+	mgr, err := manager.New(restConfig, manager.Options{Port: port})
+	if err != nil {
+		logrus.WithError(err).Fatal("Could not start manager.")
+	}
+	server := mgr.GetWebhookServer()
+	server.Register("/pods", &webhook.Admission{Handler: &podMutator{logger: logger, client: client}})
+	if err := mgr.Start(interrupts.Context()); err != nil {
+		logrus.WithError(err).Error("Failed to serve admission responses.")
+	}
+	logger.Debug("Ready to serve HTTP requests.")
```
alvaroaleman, line 44: Start is blocking, so once you see this log its actually not ready anymore

cmd/pod-scaler/testdata/zz_fixture_TestMutatePods_pod_associated_with_a_build_that_has_no_labels.yaml
```diff
@@ -0,0 +1 @@
+{"response":{"uid":"705ab4f5-6393-11e8-b7cc-42010a800002","allowed":true}}
```
alvaroaleman, line 1: All the fixtures are unused or not?

cmd/pod-scaler/admission.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+type podMutator struct {
+	logger  *logrus.Entry
+	client  buildclientv1.BuildV1Interface
+	decoder *admission.Decoder
+}
+
+func (m *podMutator) Handle(ctx context.Context, req admission.Request) admission.Response {
```
alvaroaleman, line 53: No tests at all?

petr-muller (approved):
/hold

cmd/pod-scaler/admission.go
stevekuznetsov: I really couldn't figure out how to sensibly test this without having to create my own Scheme, which I really would rather not have to do

cmd/pod-scaler/testdata/zz_fixture_TestMutatePods_pod_associated_with_a_build_that_has_no_labels.yaml
stevekuznetsov: Ah, forgot to remove

cmd/pod-scaler/admission.go
stevekuznetsov: Without the scheme for instance you end up hand-crafting AdmissionRequests with raw Object bodies, but also needing to craft the decoded objects, then you get weird diffs in the thing anyway ... what I would love is an e2e that the HTTP endpoint works TBH

alvaroaleman: how about just testing a `handle(ctx context.Context, p *corev1.Pod, req admission.Request) admission.Response` ?

stevekuznetsov:
/hold 

I want this: https://github.com/kubernetes-sigs/controller-runtime/pull/1429

cmd/pod-scaler/admission_test.go
```diff
@@ -0,0 +1,116 @@
...
+					Object:   runtime.RawExtension{Raw: []byte(`{"apiVersion": "v1","kind": "Pod","metadata": {"creationTimestamp": null, "labels": {"openshift.io/build.name": "withlabels"}, "name": "withoutlabels-build","namespace": "namespace"}, "spec":{"containers":[]}, "status":{}}`)},
+				},
+			},
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			response := mutator.Handle(context.Background(), testCase.request)
+			testhelper.CompareWithFixture(t, response)
```
alvaroaleman, line 113: Can we maybe restructure the tests to pass in a `runtime.Object` and at the end verify that the error in the response is as expected and (if no error) apply the patch and compare the result of that?

Most of the admission request in the testcases is duplicate and inferring from the raw jsonpatch if the resulting object looks like expected is pretty hard, especially since that depends on the original object, which isn't visible in the fixture

stevekuznetsov:
@petr-muller @alvaroaleman this is ready for review again now

alvaroaleman (approved):

alvaroaleman:
/hold
import order :upside_down_face:

stevekuznetsov:
lol
/hold cancel
@alvaroaleman updated

stevekuznetsov:
Part of DPTP-1822

alvaroaleman (approved):

PR #1897, opened 2021-04-21
autoowners: allow placing OWNERS files in other dirs

stevekuznetsov:
We need to place OWNERS also into the sharded Prow configuration
directory tree, so we can now pass the base of this tree to the
autoowners tool to do so.

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -54,7 +54,7 @@ func (orgRepo orgRepo) repoString() string {
 	return fmt.Sprintf("%s/%s", orgRepo.Organization, orgRepo.Repository)
 }
 
-func loadRepos(configRootDir string, blocked blocklist, configSubDirs []string, githubOrg string, githubRepo string) ([]orgRepo, error) {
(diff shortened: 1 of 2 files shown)


cmd/autoowners/main.go
```diff
@@ -87,10 +87,13 @@ func loadRepos(configRootDir string, blocked blocklist, configSubDirs []string,
 			if org == githubOrg && repo == githubRepo {
 				continue
 			}
-			repoConfigDirs := make([]string, 0, len(configSubDirectories))
+			var repoConfigDirs []string
 			for _, sourceSubDir := range configSubDirectories {
 				repoConfigDirs = append(repoConfigDirs, filepath.Join(sourceSubDir, org, repo))
 			}
+			for _, extraDir := range extraDirs {
+				repoConfigDirs = append(repoConfigDirs, filepath.Join(extraDir, org, repo))
```
alvaroaleman, line 95: What if it doesn't exist?

stevekuznetsov: It gets filtered out below, as with all the others. The test handles this - only a dir for one of the two repos exists in the extra directory tree.

alvaroaleman (approved):

PR #1898, opened 2021-04-21
Allow for custombuild strategy

darkmuggle:
The CoreOS team needs the ability to use custom build strategies to emit
ostree containers; these containers are not buildable using Docker
builds.

Signed-off-by: Ben Howard <ben.howard@redhat.com>

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1359,6 +1359,11 @@ type ProjectDirectoryImageBuildInputs struct {
 	// that will populate the build context for the Dockerfile or
 	// alter the input image for a multi-stage build.
 	Inputs map[string]ImageBuildInputs `json:"inputs,omitempty"`
+
+	// CustomBuildStrategy indicates that the image should be executed as
+	// a custom build strategy insttead of a Docker Build Strategy (default).
+	// The image must support the Custom Build API and emit an OCI Container.
+	CustomBuildStrategy bool `json:"custom_build_strategy,omitempty"`
 }
 
 // PullSpecSubstitution contains a name of a pullspec that needs to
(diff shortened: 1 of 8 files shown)


darkmuggle:
/assign @stevekuznetsov 

I *think* this is the idea that @stevekuznetsov and I spoke about in terms of creating a custom build strategy. The second piece for privileged service accounts seems a lot more involved; I'll deal with that in a separate PR.

pkg/api/types.go
```diff
@@ -1359,6 +1359,11 @@ type ProjectDirectoryImageBuildInputs struct {
 	// that will populate the build context for the Dockerfile or
 	// alter the input image for a multi-stage build.
 	Inputs map[string]ImageBuildInputs `json:"inputs,omitempty"`
+
+	// CustomBuildStrategy indicates that the image should be executed as
+	// a custom build strategy insttead of a Docker Build Strategy (default).
+	// The image must support the Custom Build API and emit an OCI Container.
+	CustomBuildStrategy bool `json:"custom_build_strategy,omitempty"`
```
stevekuznetsov, line 1366: Could we expose something more like

```go
type ProjectDirectoryImageBuildInputs struct {
  // ...
  BuildStrategyType buildapi.BuildStrategyType `json:"build_strategy_type,omitempty"
  
  BuildOptions *BuildOptions `json:"build_options,omitempty"
}

type BuildOptions struct {
  Docker *DockerOptions `json:"docker,omitempty"`
  Custom *CustomOptions `json:"custom,omitempty"`
}

type CustomOptions struct {
  ServiceAccountRoles []string `json:"serviceaccount_roles,omitempty"`
  Secrets []buildapi.SecretSpec `json:"secrets,omitempty"`
}
```

Then we can do validation that e.g. people don't set the context dir/dockerfile path when asking for a custom strategy, etc.

darkmuggle: We can probably just close this PR - it's going to take a bit to figure out the interface model and how to do this in an acceptable manner. No point in having this PR continue its decline into staleness.

PR #1904, opened 2021-04-22
Vault accessory: Use testing.T instead of testhelper.T

alvaroaleman:
testhelpers.T must be run from the testframeworks Run method to ensure
proper synchronization, otherwise it will not wait correctly and swallow
Errorf and Fatalf calls, making failing tests appear as if they
succeeded.

This changes the accessory to also allow testing.T, and adds a strongly
worded warning when using testhelper.T. The main drawback is that
failures in the accessory can not Fatalf the test anymore, as that is
not allowed from a distinct goroutine. In all practical terms this is of
little relevance for tests that have a runtime of <10 seconds like the
Vault tests.

### DIFF
--- a/pkg/testhelper/accessory.go
+++ b/pkg/testhelper/accessory.go
@@ -21,7 +21,7 @@ import (
 )
 
 // ArtifactDir determines where artifacts should for for a test case.
-func ArtifactDir(t *T) string {
+func ArtifactDir(t TestingTInterface) string {
 	var baseDir string
 	if dir, set := os.LookupEnv("ARTIFACT_DIR"); set {
 		baseDir = dir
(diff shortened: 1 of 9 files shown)


pkg/testhelper/accessory.go
```diff
@@ -120,8 +120,21 @@ type Accessory struct {
...
+// tests might pass incorrectly.
+func (a *Accessory) RunFromFrameworkRunner(t *T, parentCtx context.Context) {
+	a.run(parentCtx, t, t.Fatalf)
+}
+
+// run begins the accessory process. this call is not blocking.
+// Because testing.T does not allow to call Fatalf in a distinct
+// goroutine, this will use Errorf instead.
+func (a *Accessory) Run(t *testing.T) {
+	a.run(context.Background(), t, t.Errorf)
```
stevekuznetsov, line 134: So will this leak Vault processes or whatever?

alvaroaleman: It doesn't look like, I don't have any stray vault processes after running these tests locally

PR #1906, opened 2021-04-23
Support buildArgs for image building

hongkailiu:
https://issues.redhat.com/browse/DPTP-2157

/cc @openshift/openshift-team-developer-productivity-test-platform @vrutkovs

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1359,6 +1359,18 @@ type ProjectDirectoryImageBuildInputs struct {
 	// that will populate the build context for the Dockerfile or
 	// alter the input image for a multi-stage build.
 	Inputs map[string]ImageBuildInputs `json:"inputs,omitempty"`
+
+	// BuildArgs contains build arguments that will be resolved in the Dockerfile.
+	// See https://docs.docker.com/engine/reference/builder/#/arg for more details.
+	BuildArgs []BuildArg `json:"build_args,omitempty"`
+}
+
+type BuildArg struct {
+	// Name of the build arg.
+	Name string `json:"name,omitempty"`
+
+	// Value of the build arg.
+	Value string `json:"value,omitempty"`
 }
 
 // PullSpecSubstitution contains a name of a pullspec that needs to
--- a/pkg/steps/project_image.go
+++ b/pkg/steps/project_image.go
@@ -31,7 +31,7 @@ func (s *projectDirectoryImageBuildStep) Inputs() (api.InputDefinition, error) {
 	return nil, nil
 }
 
-func (*projectDirectoryImageBuildStep) Validate() error { return nil }
+func (s *projectDirectoryImageBuildStep) Validate() error { return nil }
 
 func (s *projectDirectoryImageBuildStep) Run(ctx context.Context) error {
 	return results.ForReason("building_project_image").ForError(s.run(ctx))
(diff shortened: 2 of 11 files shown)


pkg/steps/project_image.go
```diff
@@ -31,7 +31,14 @@ func (s *projectDirectoryImageBuildStep) Inputs() (api.InputDefinition, error) {
 	return nil, nil
 }
 
-func (*projectDirectoryImageBuildStep) Validate() error { return nil }
+func (s *projectDirectoryImageBuildStep) Validate() error {
+	for i, e := range s.config.BuildArgs {
+		if e.ValueFrom != nil {
+			fmt.Errorf("build_args[%d] not implemented: valueFrom", i)
```
alvaroaleman, line 37:
```suggestion
			fmt.Errorf("build_args[%d]: valueFrom must be set", i)
```

Is the implication that `Value` must be unset? If so, we should validate that. Do you have a reference for this?

hongkailiu: We support only `value` for now.

Confirmed by @vrutkovs 

https://coreos.slack.com/archives/CBN38N3MW/p1619182185023200?thread_ts=1619083815.472400&cid=CBN38N3MW

hongkailiu:
```
oc explain build.spec.strategy.dockerStrategy.buildArgs
KIND:     Build
VERSION:  build.openshift.io/v1

RESOURCE: buildArgs <[]Object>

DESCRIPTION:
     buildArgs contains build arguments that will be resolved in the Dockerfile.
     See https://docs.docker.com/engine/reference/builder/#/arg for more
     details.

     EnvVar represents an environment variable present in a Container.

FIELDS:
   name	<string> -required-
     Name of the environment variable. Must be a C_IDENTIFIER.

   value	<string>
     Variable references $(VAR_NAME) are expanded using the previous defined
     environment variables in the container and any service environment
     variables. If a variable cannot be resolved, the reference in the input
     string will be unchanged. The $(VAR_NAME) syntax can be escaped with a
     double $$, ie: $$(VAR_NAME). Escaped references will never be expanded,
     regardless of whether the variable exists or not. Defaults to "".

   valueFrom	<Object>
     Source for the environment variable's value. Cannot be used if value is not
     empty.
```

alvaroaleman (approved):

pkg/api/types.go
```diff
@@ -1359,6 +1360,10 @@ type ProjectDirectoryImageBuildInputs struct {
 	// that will populate the build context for the Dockerfile or
 	// alter the input image for a multi-stage build.
 	Inputs map[string]ImageBuildInputs `json:"inputs,omitempty"`
+
+	// BuildArgs contains build arguments that will be resolved in the Dockerfile.
+	// See https://docs.docker.com/engine/reference/builder/#/arg for more details.
```
alvaroaleman, line 1365: Please mention here that only `Value` is supported

hongkailiu: Updated.

pkg/api/types.go
```diff
@@ -1359,6 +1360,11 @@ type ProjectDirectoryImageBuildInputs struct {
...
+	// Currently only value is supported.
+	BuildArgs []corev1.EnvVar `json:"build_args,omitempty"`
```
stevekuznetsov, line 1367: Use a type that exposes the options we *actually* use, not some superset that misleads users and we need to validate, please.

stevekuznetsov (commented):
This needs tests. I don't think e2e, but unit for sure.

pkg/api/types.go
hongkailiu: Updated.

alvaroaleman (approved):

PR #1907, opened 2021-04-23
applyconfig: support ignore-file

hongkailiu:
Use case: `arm01` has its own version of cert-manager which is much newer than what we have.
So we will ignore the file in the common folder for arm01.

/cc @alvaroaleman

### DIFF
--- a/cmd/applyconfig/applyconfig.go
+++ b/cmd/applyconfig/applyconfig.go
@@ -35,6 +35,7 @@ type dryRunMethod string
 type options struct {
 	user        *nullableStringFlag
 	directories flagutil.Strings
+	ignoreFiles flagutil.Strings
 	context     string
 	kubeConfig  string
 	dryRun      dryRunMethod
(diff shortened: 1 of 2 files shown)


cmd/applyconfig/applyconfig.go
```diff
@@ -465,6 +467,10 @@ func fileFilter(info os.FileInfo, path string) (bool, error) {
 		return true, nil
 	}
 
+	if ignoreFiles.StringSet().Has(path) {
```
alvaroaleman, line 470:
```suggestion
	if ignoreFiles.StringSet().Has(filepath.BaseName(path)) {
```

Wouldn't that be more logical?

hongkailiu: Specially for `cert-manager.yaml`, we have several files with this name in different folders.

hongkailiu: https://github.com/openshift/release/blob/master/clusters/build-clusters/01_cluster/cert-manager/cert-manager.yaml
and
https://github.com/openshift/release/blob/master/clusters/build-clusters/common/cert-manager.yaml

alvaroaleman (approved):

PR #1909, opened 2021-04-23
ci-operator: label release pods with their release name

stevekuznetsov:
Release pods vary only by the CI Operator configuration, not the
specific target that ends up running them. It is important to be able to
detect that a Pod is being run for a release, in some manner that's less
fragile than searching for a prefix on the Pod name.


stevekuznetsov:
Part of DPTP-1822

alvaroaleman (approved):

PR #1912, opened 2021-04-24
Get build arg from a secret

hongkailiu:
https://issues.redhat.com/browse/DPTP-2157

/cc @openshift/openshift-team-developer-productivity-test-platform @vrutkovs

Not in the card.
But it is requested
https://coreos.slack.com/archives/CBN38N3MW/p1619182185023200?thread_ts=1619083815.472400&cid=CBN38N3MW


stevekuznetsov:
Is this https://issues.redhat.com/browse/DPTP-2150 ?

hongkailiu:
> Is this https://issues.redhat.com/browse/DPTP-2150 ?

No. See [the PR description](https://github.com/openshift/ci-tools/pull/1912#issue-622394154).

pkg/api/types.go
```diff
@@ -1369,8 +1369,42 @@ type BuildArg struct {
...
+}
+
+// SourceName represents the namespaced source name
+func (arg *BuildArgSource) SourceName() (string, error) {
+	if arg.Namespace == "" {
+		return "", fmt.Errorf("namespace must be set")
+	}
+	if arg.SecretKeyRef != nil {
+		if arg.SecretKeyRef.Name == "" {
+			return "", fmt.Errorf("secret_key_ref.name must be set")
```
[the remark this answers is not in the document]
hongkailiu: I did both.

This is the load time validation.

https://github.com/openshift/ci-tools/pull/1912/files#diff-1de6f92c6a7684ebc85283d5e494b7e26bd25dfbee9bf5f225ab7103ab7d171aR164

pkg/steps/project_image.go
```diff
@@ -64,6 +72,42 @@ func (s *projectDirectoryImageBuildStep) run(ctx context.Context) error {
 	return handleBuild(ctx, s.client, build)
 }
 
+func (s *projectDirectoryImageBuildStep) createSecrets() error {
```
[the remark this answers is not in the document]
hongkailiu: You caught me. I copy-and-modify. :)
Will give it a try.

pkg/steps/source.go
```diff
@@ -353,7 +353,24 @@ func buildFromSource(jobSpec *api.JobSpec, fromTag, toTag api.PipelineImageStrea
 
 func toEnv(args []api.BuildArg) []corev1.EnvVar {
 	var ret []corev1.EnvVar
-	for _, arg := range args {
+	for i, arg := range args {
+		if arg.ValueFrom != nil {
```
petr-muller, line 357: We could do 

```
if arg.ValueFrom == nil {
  ret = append ...
  continue
}
```

and switch the rest left, improving readability

pkg/steps/project_image.go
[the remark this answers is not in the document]
hongkailiu: Tried to abstract the logic but did not succeed.
The hard part is collecting `toCreates` because the source secrets are from different fields of ci-op's config.
Once we have `toCreates` there is nothing left except a simple Create call on the client.

pkg/steps/source.go
hongkailiu: Done.

pkg/api/types.go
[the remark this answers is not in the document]
stevekuznetsov: This seems weird. @petr-muller @alvaroaleman WDYT? I get defensive coding but at the same time emitting errors about configuration fields at runtime when it's guaranteed at config load already seems like a waste of time and confusing to read IMHO.

pkg/steps/project_image.go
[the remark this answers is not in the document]
stevekuznetsov: Why don't we write

```go
func CopySecretsIntoJobNamespace(ctx context.Context, jobSpec JobSpec, secrets []ctrlruntimeclient.ObjectKey) error {}
```

?

pkg/steps/project_image.go
```diff
@@ -64,6 +72,42 @@ func (s *projectDirectoryImageBuildStep) run(ctx context.Context) error {
...
+	}
+	toCreate := map[string]*coreapi.Secret{}
+	for i, args := range s.config.BuildArgs {
+		if args.ValueFrom != nil {
+			name, err := args.ValueFrom.SourceName()
+			if err != nil {
+				return fmt.Errorf("build_args[%d] failed to determine source name: %w", i, err)
+			}
+			raw := &coreapi.Secret{}
+			if err := s.secretClient.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: args.ValueFrom.Namespace, Name: args.ValueFrom.SecretKeyRef.Name}, raw); err != nil {
```
stevekuznetsov, line 88: Never use `context.TODO()`. All of our steps have `func (s *step) Run(ctx context.Context)`

pkg/steps/project_image.go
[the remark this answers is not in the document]
hongkailiu: Updated.

pkg/steps/project_image.go
hongkailiu: Done.

pkg/api/types.go
```diff
@@ -1375,8 +1375,42 @@ type BuildArg struct {
...
-	// Value of the build arg.
+	// Value of the build arg. Cannot be set if ValueFrom is set.
 	Value string `json:"value,omitempty"`
+
+	// ValueFrom specifies the source of the build Arg. Cannot be set if Value is set.
+	ValueFrom *BuildArgSource `json:"value_from,omitempty"`
+}
+
+// BuildArgSource represents a source for the value of a BuildArg.
+type BuildArgSource struct {
```
stevekuznetsov, line 1386: I don't understand this structure. Why is the keyref nested? Why are the key and secret name correlated in a separate struct? What does it mean for this:

```yaml
value_from:
  namespace: thing
```

?

This is implied as possible by having a pointer to `SecretKeySelector` and allowing to to be `omitempty`

pkg/api/types.go
```diff
@@ -1375,8 +1375,42 @@ type BuildArg struct {
...
+	// SecretKeyRef selects a key of a secret
+	SecretKeyRef *SecretKeySelector `json:"secret_key_ref,omitempty"`
+	// Namespace of the BuildArg source
+	Namespace string `json:"namespace,omitempty"`
```
stevekuznetsov, line 1390: Also why are these `omitempty` when we aggressively validate that they are not empty?

pkg/api/types.go
hongkailiu: Put them to the same level.
I was thinking the value could be taken from config as well in the future.

pkg/api/types.go
hongkailiu: Removed.

stevekuznetsov (approved):

PR #1914, opened 2021-04-25
testhelper: attempt to make global map global

stevekuznetsov:
The Go reference [1] says:

> If multiple packages import a package, the imported package will be
> initialized only once.

However, something seems wrong with this. It may be the case that the
map in this helper package is being initialized more than once. Perhaps
doing it in an `init()` will help?

[1] https://golang.org/ref/spec#Package_initialization


stevekuznetsov:
Might help DPTP-2159 ?

PR #1915, opened 2021-04-26
[DPTP-2155] Skip item existance validation that exist in ci-secret-generator configuration

droslean:

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -36,6 +36,7 @@ import (
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/api/secretbootstrap"
+	"github.com/openshift/ci-tools/pkg/api/secretgenerator"
 	vaultapi "github.com/openshift/ci-tools/pkg/api/vault"
 	"github.com/openshift/ci-tools/pkg/kubernetes/pkg/credentialprovider"
 	"github.com/openshift/ci-tools/pkg/secrets"
@@ -49,17 +50,20 @@ type options struct {
 	force                bool
 	validateBWItemsUsage bool
 
-	kubeConfigPath  string
-	configPath      string
-	cluster         string
-	logLevel        string
-	impersonateUser string
+	kubeConfigPath      string
+	configPath          string
+	generatorConfigPath string
+	cluster             string
+	logLevel            string
+	impersonateUser     string
 
 	maxConcurrency int
 
-	secretsGetters map[string]coreclientset.SecretsGetter
-	config         secretbootstrap.Config
-	bwAllowUnused  flagutil.Strings
+	secretsGetters  map[string]coreclientset.SecretsGetter
+	config          secretbootstrap.Config
+	generatorConfig secretgenerator.Config
+
+	bwAllowUnused flagutil.Strings
 
 	validateOnly bool
 }
@@ -80,6 +84,7 @@ func parseOptions(censor *secrets.DynamicCensor) (options, error) {
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
 	fs.StringVar(&o.kubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
 	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
+	fs.StringVar(&o.generatorConfigPath, "generator-config", "", "Path to the secret-generator config file.")
 	fs.StringVar(&o.cluster, "cluster", "", "If set, only provision secrets for this cluster")
 	fs.BoolVar(&o.force, "force", false, "If true, update the secrets even if existing one differs from Bitwarden items instead of existing with error. Default false.")
 	fs.StringVar(&o.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
@@ -117,6 +122,14 @@ func (o *options) completeOptions(censor *secrets.DynamicCensor) error {
 	var config secretbootstrap.Config
 	if err := secretbootstrap.LoadConfigFromFile(o.configPath, &config); err != nil {
 		return err
+	}
+
+	if o.generatorConfigPath != "" {
+		var err error
+		o.generatorConfig, err = secretgenerator.LoadConfigFromPath(o.generatorConfigPath)
+		if err != nil {
+			return err
+		}
 	}
 
 	var kubeConfigs map[string]*rest.Config
@@ -757,11 +770,12 @@ func getUnusedBWItems(config secretbootstrap.Config, client secrets.ReadOnlyClie
 	return utilerrors.NewAggregate(errs)
 }
 
-func validateBWItems(client secrets.ReadOnlyClient, secretConfigs []secretbootstrap.SecretConfig) error {
+func (o *options) validateBWItems(client secrets.ReadOnlyClient) error {
 	var errs []error
 
-	for _, config := range secretConfigs {
+	for _, config := range o.config.Secrets {
 		for _, item := range config.From {
+			logger := logrus.WithField("item", item.BWItem)
 
 			if item.DockerConfigJSONData != nil {
 				for _, data := range item.DockerConfigJSONData {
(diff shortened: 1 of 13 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -761,8 +761,14 @@ func validateBWItems(client secrets.ReadOnlyClient, secretConfigs []secretbootst
 	var errs []error
 
 	for _, config := range secretConfigs {
-		for _, item := range config.From {
+		if config.Generated {
```
alvaroaleman, line 764: /hold
This should be possible without additional config opts, ref https://coreos.slack.com/archives/GB7NB0CUC/p1619441793136000?thread_ts=1619440512.133900&cid=GB7NB0CUC

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if err := yaml.Unmarshal(cfgBytes, &config); err != nil {
+		return nil, err
+	}
+	return config.massiveItemGeneration()
+}
+
+type Config []SecretItem
+
+func (c Config) massiveItemGeneration() (Config, error) {
```
alvaroaleman, line 30:
```suggestion
func (c Config) expandParameters() (Config, error) {
```

"massive item" is kind of unclear IMHO.

pkg/api/secretgenerator/secretgenerator_test.go
```diff
@@ -0,0 +1,345 @@
+package secretgenerator
+
+import (
+	"sort"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestMassiveItemGeneration(t *testing.T) {
```
alvaroaleman, line 10: I think those would be easier to understand and maintain if they used fixtures and simply read a configfile, expand it, write the expanded file as yaml back.

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (c Config) IsItemGenerated(name string) bool {
+	byName := c.itemsByName()
+	if _, ok := byName[name]; ok {
+		return true
+	}
+	return false
+}
+
+func (c Config) IsComponentGenerated(name, component string) bool {
```
alvaroaleman, line 61: Could we rename this to `IsFieldGenerated`? I know it conflicts with the bitwarden notion of field, but we don't use bitwarden anymore

pkg/api/secretgenerator/secretgenerator.go
alvaroaleman: Also, what about instead implementing this as [json unmarshaler](https://golang.org/pkg/encoding/json/#Unmarshaler)? That has the advantage that it can't be forgotten

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,140 @@
...
+		processedItems = append(processedItems, items...)
+	}
+	return processedItems, utilerrors.NewAggregate(errs)
+}
+
+func (c Config) itemsByName() map[string][]SecretItem {
+	byName := make(map[string][]SecretItem)
+
+	for _, secretItem := range c {
+		byName[secretItem.ItemName] = append(byName[secretItem.ItemName], secretItem)
```
alvaroaleman, line 47: Do we validate somewhere that names are unique?

droslean: It doesn't seem so. I will add the validation.

pkg/api/secretgenerator/secretgenerator.go
droslean: That sounds intrested. I'll take a look.

pkg/api/secretgenerator/secretgenerator.go
droslean: oh.,.. it seems that intentionally have duplicated item names....

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,147 @@
...
+		return nil, err
+	}
+	return config, nil
+}
+
+type Config []SecretItem
+
+func (c *Config) UnmarshalJSON(data []byte) error {
+	var config []SecretItem
+	if err := yaml.Unmarshal(data, &config); err != nil {
```
alvaroaleman, line 31:
```suggestion
	if err := json.Unmarshal(data, &config); err != nil {
```

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,147 @@
...
+		return err
+	}
+
+	var err error
+
+	var newConfig Config
+	for _, si := range config {
+		i, _ := si.generateItemsFromParams()
```
alvaroaleman, line 39: why do we swallow the error here?

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,147 @@
...
+		newConfig = append(newConfig, i...)
+	}
+
+	*c = newConfig
+	if err != nil {
```
alvaroaleman, line 44: Its impossible for this error to be non-nil and `if err != nil { return err }; return nil ` is equal to `return err`

pkg/api/secretgenerator/secretgenerator_test.go
```diff
@@ -0,0 +1,225 @@
...
+
+func TestLoadConfigFromPath(t *testing.T) {
+	testcases := []struct {
+		name     string
+		config   string
+		expected string
+	}{
+		{
+			name: "No parameters",
+			config: `- item_name: Item1
```
alvaroaleman, line 23: What I meant was not to inline it but to read it from a literal file and write it out, i.E. using pkg/testhelper.CompareWithFixture

pkg/api/secretgenerator/secretgenerator.go
droslean: the data are in YAML format.

pkg/api/secretgenerator/secretgenerator.go
droslean: I ate that while debugging. Nice catch, thanks.

pkg/api/secretgenerator/secretgenerator.go
alvaroaleman: No, data that is given to an UnmarshalJSON must never be in yaml format

alvaroaleman: The yaml unmarshaler converts it to json and then calls the UnmarshalJSON IIRC

droslean: Yea that's why it's working now.  If we need to replace to `json.Unmarshal` we will have to convert the data to JSON before passing it here. But yaml already does that.

droslean: Or I can guess that the result will be the same

droslean: I just checked. Changing that to `json`  generates an error.

alvaroaleman: Then you are doing sth wrong, this definitely works, we have examples of this in our codebase like e.G.: https://github.com/openshift/ci-tools/blob/9f710b04bc1319b5374ea7ef26a8dd1e8dd97f8b/pkg/api/secretbootstrap/secretboostrap.go#L90-L95

where we implement a custom json unmarshaler that is used withing the yaml unmarshaling

droslean: Yea that makes sense. Let me check this and update the PR. Thanks

droslean: addressed

pkg/api/secretgenerator/secretgenerator.go
```diff
@@ -0,0 +1,148 @@
...
+
+func (si SecretItem) generateItemsFromParams() ([]SecretItem, error) {
+	var errs []error
+	var processedBwItems []SecretItem
+
+	replaceParameter := func(paramName, param, template string) string {
+		return strings.ReplaceAll(template, fmt.Sprintf("$(%s)", paramName), param)
+	}
+
+	itemsProcessingHolder := []SecretItem{si}
```
alvaroaleman, line 117:
```suggestion
	var result []SecretItem
```

pkg/api/secretgenerator/secretgenerator_test.go
```diff
@@ -0,0 +1,142 @@
...
+					Password:    "echo -n password1", Notes: "Note1",
+					Params: map[string][]string{
+						"FieldNum": {"1"},
+					},
+				},
+			},
+		},
+		{
+			name: "single parameter with multiple values",
+			expected: Config{
```
alvaroaleman, line 49: Well so ideally both the in and out of the test is a file, so we have a file `config_some_test.yaml` and `config_some_test_expanded.yaml` so you can just look at the files and also autogenerate them via `pkg/testhelper.CompareWithFixture`

pkg/api/secretgenerator/secretgenerator.go
droslean: That change will break it. We need `si` to be in the slice in order to update it multiple times after the first combination will finish and the next will start.

droslean: For the record, that code is a pretty much a copy-paste from https://github.com/openshift/ci-tools/blob/master/cmd/ci-secret-generator/main.go#L168-L203

alvaroaleman: > That change will break it. We need si to be in the slice in order to update it multiple times after the first combination will finish and the next will start.

We don't, the only write access to it is `itemsProcessingHolder = itemsProcessed` which overwrites it entirely.

droslean: Apparently, applying this change breaks the entire algorithm by returning an empty object. Either is something completely wrong here or I am missing something. For the record, I didn't change anything from our production code.

alvaroaleman: Yeah, because we loop over the `itemsProcessingHolder` in the beginning instead of over `si`

droslean: Yes, this seems intentional. I didn't touch anything. Let me check this out and refactor it from scratch.

droslean: Also, we can't loop over `si` since it's not a slice, and we don't want to change it at all. When I first saw this code I thought that maybe it's better to re-write it from scratch but I didn't do it.

alvaroaleman (approved):

PR #1916, opened 2021-04-26
Vaultclient: Refresh token when using kubernetes auth

alvaroaleman:
It's only valid for one hour, so long-running applications like the
secret manager currently just stop working :grimace:

### DIFF
--- a/pkg/vaultclient/vaultclient.go
+++ b/pkg/vaultclient/vaultclient.go
@@ -8,28 +8,59 @@ import (
 	"net/http"
 	"reflect"
 	"strings"
+	"sync"
+	"time"
 
 	"github.com/hashicorp/vault/api"
+	"github.com/sirupsen/logrus"
 )
 
-func NewFromKubernetesAuth(addr, role string) (*VaultClient, error) {
+func getKuberntesAuthToken(client *VaultClient, role string) (string, time.Duration, error) {
+
+	// Clone the client before resetting the token
+	var err error
+	client.Client, err = client.Client.Clone()
+	if err != nil {
+		return "", 0, fmt.Errorf("failed to clone client: %w", err)
+	}
+	client.SetToken("")
+
 	serviceAccountToken, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/token")
 	if err != nil {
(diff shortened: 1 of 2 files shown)


pkg/vaultclient/vaultclient.go
```diff
@@ -58,6 +88,26 @@ type VaultClient struct {
 	*api.Client
 }
 
+func (v *VaultClient) refreshTokenWhenNeeded(ttl time.Duration, refreshFn func(*VaultClient) (string, time.Duration, error)) {
```
stevekuznetsov, line 91: I think we could likely make this less fragile. Perhaps something in the k8s `wait` utilities already implements such a thing, by the way.

One suggestion: keep track of how many `backoff`s you've done and when the sum total there `>= tt/2` kill the process?

alvaroaleman: Okay so I re-implemented this using the apimachinerys wait.Package but I think that is actually worse and doesn't do some of the things it claims to do (i.E. wait.Exponential backoff doesn't use an exponential backoff, but wait * factor backoff, where you can not access wait anymore, so it can not be made exponential): https://github.com/openshift/ci-tools/compare/master...alvaroaleman:refresh-wait?expand=1

> One suggestion: keep track of how many backoffs you've done and when the sum total there >= tt/2 kill the process?

The problem is that we can not handle errors there anymore and I am not sure if its correct to assume that we always want the thing to be exit. Maybe vault is temporarorily and known down, so it will just recover once its up?

stevekuznetsov: Hm. What would be the better failure mode? Fall over and yell loudly so as to make it clear that something is up with Vault? With this backoff we could theoretically get to the point where we're waiting around for really long periods of time hoping vault comes back up, IMO maybe better to go into Pod CrashLoopBackOff so it's more visible?

alvaroaleman: Well, we'd have to plug "do you want the binary to crash when this fails" through. How about:
* We cap the retrying interval
* We expose the "did our token expire without us being able to renew it" somehow and plug it into a healthprobe?

stevekuznetsov (approved):

PR #1918, opened 2021-04-26
DPTP-2012: Best effort/timeout and trap/grace period validations.

emilvberglind:
With [DPTP-2012](https://issues.redhat.com/browse/DPTP-2012) we want to add validations that, when a test includes a `trap` command, a `grace_period` is also set, and that when `best_effort` is `true`, a `timeout` is also set.

### DIFF
--- a/pkg/registry/resolver.go
+++ b/pkg/registry/resolver.go
@@ -2,7 +2,7 @@ package registry
 
 import (
 	"fmt"
-
+	"github.com/openshift/ci-tools/pkg/validation"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 
@@ -24,15 +24,17 @@ type ObserverByName map[string]api.Observer
 func Validate(stepsByName ReferenceByName, chainsByName ChainByName, workflowsByName WorkflowByName, observersByName ObserverByName) error {
 	reg := registry{stepsByName, chainsByName, workflowsByName, observersByName}
 	var ret []error
+	var validationArgs = validation.NoOpValidationArgs()
+
 	for k := range chainsByName {
-		if _, err := reg.process([]api.TestStep{{Chain: &k}}, sets.NewString(), stackForChain()); err != nil {
+		if _, err := reg.process(validationArgs, []api.TestStep{{Chain: &k}}, sets.NewString(), stackForChain()); err != nil {
 			ret = append(ret, err...)
 		}
 	}
 	for k, v := range workflowsByName {
 		stack := stackForWorkflow(k, v.Environment, v.Dependencies)
 		for _, s := range [][]api.TestStep{v.Pre, v.Test, v.Post} {
-			if _, err := reg.process(s, sets.NewString(), stack); err != nil {
+			if _, err := reg.process(validationArgs, s, sets.NewString(), stack); err != nil {
 				ret = append(ret, err...)
 			}
 		}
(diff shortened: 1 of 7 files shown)


emilvberglind:
Note that the literal test step validations will now also be executed as part of the resolve process.

pkg/registry/resolver_test.go
```diff
@@ -698,6 +698,57 @@ func TestResolve(t *testing.T) {
...
+			Workflow:       &awsWorkflow,
+		},
+		workflowMap: WorkflowByName{
+			awsWorkflow: {
+				ClusterProfile: api.ClusterProfileAWS,
+				Pre: []api.TestStep{{
+					LiteralTestStep: &api.LiteralTestStep{
+						As:       "trapper-keeper",
+						From:     "installer",
+						Commands: "trap \"\" SIGINT",
```
alvaroaleman, line 714: Nit: You can avoid the need to escape by putting the string into backticks:
```suggestion
						Commands: `trap "" SIGINT`,
```

pkg/registry/resolver_test.go
```diff
@@ -698,6 +698,57 @@ func TestResolve(t *testing.T) {
...
+							Requests: api.ResourceList{"cpu": "1000m"},
+							Limits:   api.ResourceList{"memory": "2Gi"},
+						},
+						Environment: []api.StepParameter{
+							{Name: "STEP_ENV", Default: &stepEnv},
+						}},
+				}},
+			},
+		},
+		expectedErr: errors.New("test best-effort contains best_effort without timeout"),
```
alvaroaleman, line 751: Can you also add positive tests, i,E. `No error when no trap command & not best effort`?

pkg/validation/test.go
```diff
@@ -522,23 +554,51 @@ func validateLiteralTestStep(context context, stage testStage, step api.LiteralT
...
 }
 
+func validateCommands(test api.LiteralTestStep) []error {
+	var validationErrors []error
+
+	var commands = test.Commands
+	scanner := bufio.NewScanner(strings.NewReader(commands))
+
+	hasTrapCommand := false
+	for scanner.Scan() {
```
alvaroaleman, line 588:
```suggestion
	if trapPattern.MatchString(test.Commands) {...
```

pkg/validation/test.go
```diff
@@ -41,12 +45,34 @@ func newContext(fieldRoot string, env api.TestEnvironment, releases sets.String)
...
+	}
+}
+
+func NoOpValidationArgs() Args {
+	return Args{
+		context: nil,
+	}
+}
+
+func (v *Args) ShouldValidate() bool {
```
alvaroaleman, line 72: I don't think we do any runtime checking if we should validate elsewhere and just always do it. The assumption is that its not possible to get invalid config merged.

pkg/validation/test.go
```diff
@@ -41,12 +45,34 @@ func newContext(fieldRoot string, env api.TestEnvironment, releases sets.String)
...
+func NewValidationArgs(context Context, testStage TestStage) Args {
+	return Args{
+		context:   &context,
+		testStage: testStage,
+	}
+}
+
+func NoOpValidationArgs() Args {
+	return Args{
+		context: nil,
```
alvaroaleman, line 68: Pointers are implicitly nil by default, so this is equal to `return Args{}`

pkg/registry/resolver.go
```diff
@@ -2,7 +2,7 @@ package registry
```
alvaroaleman, line 2: Why does adding this validation require us to pull all validation into the resolver? I think the general assumption is that config that reaches the resolver is valid, we validate it before merging anything.

stevekuznetsov: IIRC when we looked through this, there was technically a gap here - we validate in ci-op after resolving, so if someone checked in a step that did not get used anywhere, it could remain invalid. There didn't seem to be any ideological reason as to why we only validate on one end of things *and* we likely wanted the validation to occur in both cases (however unlikely that edge was where an unused step was checked in).

pkg/validation/test.go
emilvberglind: There were some instances here, like in `resolver.Validate(stepsByName ReferenceByName, chainsByName ChainByName, workflowsByName WorkflowByName, observersByName ObserverByName)` where it is just doing a "shallow" validation where it just checks some high-level components rather than doing the full step validation. So, I just added this `ShouldValidate` as a convenience method for determining whether or not to perform that full step validation - perhaps I should rename it to something like `ShouldPerformLiteralStepValidation` to be more descriptive?

pkg/registry/resolver.go
alvaroaleman: Yeah I am not saying we don't want it, just that we don't need it in the configresolver but a presubmit instead. Right now this requires exporting a lot of code which isn't great

pkg/validation/test.go
emilvberglind: Good call. I think I was initially going to try to do something very different from what I wound up doing here, but in its current state it definitely is not very useful to loop over each line.

pkg/registry/resolver.go
emilvberglind: Can we assume that everything that reaches the resolver will for sure be valid? I.e. there is no way that it can change before it reaches the resolver? I don't yet know enough, so that's why I'm asking :). When I was playing around with it it appeared that, as @stevekuznetsov said, you could potentially check in an unused step and that step would then not be validated until it was actually used in a job, which didn't seem ideal.

alvaroaleman: > you could potentially check in an unused step and that step would then not be validated until it was actually used in a job, which didn't seem ideal.

That does sound like something we should fix.

The reason I am not much in favor of doing this validation in the configresolver is that the only ones who might ever see an error there is our team, even though we won't own the config that causes the error.

bbguimaraes: I agree, it's very unexpected to see this change in the middle of this PR.

We have `pull-ci-openshift-release-master-ci-operator-registry`, which loads the registry and validates the contents of all files (using `ci-operator-configresolver --validate-only`).  If there is a validation gap, I'd expected it to be fully stated and fixed separately.

bbguimaraes: (that said, this is definitely not the most straightforward part of `ci-tools`, I'd be happy to provide some guidance)

emilvberglind:
After discussion with various persons and getting better clarity on the problem with this PR I'm going to close it and open a new one with revised changes.

/close

PR #1834, opened 2021-03-31
ci-operator: reduce log spam

stevekuznetsov:
e2e: check verbose logfile for content

We don't put everything into the user-facing logfile on purpose, but we
*do* want to probe implementation details in the tests, so we look at
the verbose logfile for the snippets we care about.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

ci-operator: reduce log spam

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

Before:

```
time="2021-03-31T10:32:29Z" level=info msg="ci-operator version v20210331-a19452f"
time="2021-03-31T10:32:29Z" level=info msg="Resolved source https://github.com/test/test to master@6d231cc3, merging: #1234 538680df @droslean"
time="2021-03-31T10:32:30Z" level=info msg="Resolved release latest to quay.io/openshift-release-dev/ocp-release@sha256:a035dddd8a5e5c99484138951ef4aba021799b77eb9046f683a5466c23717738"
time="2021-03-31T10:32:30Z" level=info msg="Resolved openshift/centos:7 to sha256:e4ca2ed0202e76be184e75fb26d14bf974193579039d5573fb2348664deef76e."
time="2021-03-31T10:32:30Z" level=info msg="Using namespace https://console.build01.ci.openshift.org/k8s/cluster/projects/ci-op-bq78lp30"
time="2021-03-31T10:32:30Z" level=info msg="Running [input:os], configurable-leases-registry"
time="2021-03-31T10:32:30Z" level=info msg="Creating namespace ci-op-bq78lp30"
time="2021-03-31T10:32:31Z" level=info msg="Spent 4.508295ms waiting for RBAC to initialize in the new namespace."
time="2021-03-31T10:32:31Z" level=info msg="[0/30] Image pull secrets in namespace not yet ready, sleeping for a second..."
time="2021-03-31T10:32:32Z" level=info msg="Spent 1.03617484s waiting for image pull secrets to initialize in the new namespace."
time="2021-03-31T10:32:32Z" level=info msg="Creating rolebinding for user droslean in namespace ci-op-bq78lp30"
time="2021-03-31T10:32:32Z" level=info msg="Setting up pipeline ImageStream for the test"
time="2021-03-31T10:32:32Z" level=info msg="Created secret import-secret"
time="2021-03-31T10:32:32Z" level=info msg="Created secret success-cluster-profile"
[...]

PR #1921, opened 2021-04-27
pkg/steps/lease: Log acquired leases at info level

wking:
Partially rolling back ef63f31f5b (#1834).  This gets the "which resource type?" and "which acquired name?" information up into the build log, so we can see which jobs are running in which AWS regions and such, instead of having to drop down from [the job][1]:

    INFO[2021-04-16T19:55:46Z] Acquiring leases for test launch
    INFO[2021-04-16T19:55:47Z] Running multi-stage test launch

into [ci-operator assets][2]:

    {"level":"debug","msg":"Acquired lease(s) for osd-ephemeral-quota-slice: [9096423b-d59e-491d-8ee2-f7de51139df1]","time":"2021-04-16T19:55:47Z"}

[1]: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/release-openshift-origin-installer-launch-osd-ephemeral/1383146668190339072#1:build-log.txt%3A12
[2]: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-osd-ephemeral/1383146668190339072/artifacts/ci-operator.log

### DIFF
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -155,7 +155,7 @@ func acquireLeases(
 			errs = append(errs, results.ForReason(results.Reason("acquiring_lease:"+l.ResourceType)).WithError(err).Errorf("failed to acquire lease: %v", err))
 			break
 		}
-		logrus.Debugf("Acquired lease(s) for %s: %v", l.ResourceType, names)
+		logrus.Infof("Acquired %d lease(s) for %s: %v", l.Count, l.ResourceType, names)
 		l.resources = names
 	}
 	if errs != nil {
--- a/test/e2e/lease/e2e_test.go
+++ b/test/e2e/lease/e2e_test.go
@@ -58,7 +58,6 @@ func TestLeases(t *testing.T) {
 			success: true,
 			output: []string{
 				`Acquiring 1 lease(s) for aws-quota-slice`,
-				`Acquired lease(s) for aws-quota-slice`,
 				`Releasing leases for test success`,
 				`Releasing lease for aws-quota-slice`,
 			},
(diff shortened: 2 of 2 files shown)


alvaroaleman (approved):

petr-muller:
+1, also useful now when we have jobs backed by different cloud accounts

stevekuznetsov:
You would need to fix e2e. Though I really wonder - how useful is this? Why is it useful? In what cases are what types of users looking for this? Is there something more intelligent we can do? My gut instinct is that 99.99999% of users _never_ want this, so it's spam in the logs. The ones that do, are looking for something higher-level than "which lease did this use" and perhaps they are better suited with some other tool...

/approve cancel

petr-muller:
> The ones that do, are looking for something higher-level than "which lease did this use" and perhaps they are better suited with some other tool...

My use case: I'm aware of the fact that there are multiple accounts backing the jobs and I can read the lease names to very quickly figure out "what account is this job using" which is very relevant now as we are moving jobs between accounts. I agree that this mental mapping is something that probably only five to ten people in the org can do and it's useless for most users, but I can attest that not being able to eyeball this right from the log is annoying.

I'm all for solving this somehow better, but I'd really like this PR until we solve this better.

wking:
I've pushed a3f9318b5 -> 7984332a8 to hopefully fix the e2e, although it's not clear to me why we aren't capturing info-level logs to compare against in the e2e tests.

For use-cases, I agree with @petr-muller.  I also like to [use CI search][1] to be able to figure out which accounts/regions are having quota problem, etc.  Getting that data into the build log allows me to do that.  I also agree with @petr-muller that it would be awesome to have more structured access to this data, but until we do, this seems like a reasonable stopgap.

[1]: https://search.ci.openshift.org/?maxAge=24h&type=all&search=Acquired+lease.*+for

petr-muller (approved):

wking:
Still has a hold.  Have we been convincing enough, @stevekuznetsov ?

PR #1923, opened 2021-04-27
DPTP-2012: Best effort/timeout and trap/grace period validations.

emilvberglind:
With [DPTP-2012](https://issues.redhat.com/browse/DPTP-2012) we want to add validations that, when a test includes a trap command, a grace_period is also set, and that when best_effort is true, a timeout is also set.


stevekuznetsov:
You've got a lot of commits with the same name - consider [squashing](https://www.git-tower.com/learn/git/faq/git-squash/) them with `git` or ask our merge robot to do it with:

> /label tide/merge-method-squash

pkg/validation/test.go
```diff
@@ -539,6 +568,21 @@ func validateLiteralTestStep(context context, stage testStage, step api.LiteralT
 	return
 }
 
+func validateCommands(test api.LiteralTestStep) []error {
+	var validationErrors []error
+
+	hasTrapCommand := false
```
stevekuznetsov, line 574: nit: this can just be

```go
hasTrapCommand := trapPattern.MatchString(test.Commands)
```

pkg/validation/test_test.go
```diff
@@ -491,6 +491,8 @@ func TestValidateTestSteps(t *testing.T) {
 	myReference := "my-reference"
 	asReference := "as"
 	yes := true
+	defaultDuration := &prowv1.Duration{Duration: 1 * time.Minute}
+	trueRef := &[]bool{true}[0]
```
stevekuznetsov, line 495: this is amazing - we usually do

```
yes := true 

// ...
{
  something: &yes
}
```

stevekuznetsov (commented):
Couple comments but this looks good overall

pkg/validation/test_test.go
emilvberglind: Lol, yeah I'll just do that since the yes bool is already there. I initially had it inlined and then just made it a variable.

pkg/validation/test.go
```diff
@@ -515,18 +536,26 @@ func validateLiteralTestStep(context context, stage testStage, step api.LiteralT
...
 	ret = append(ret, validateCredentials(context.fieldRoot, step.Credentials)...)
-	if err := validateParameters(&context, step.Environment); err != nil {
-		ret = append(ret, err)
+	if context.env != nil {
+		if err := validateParameters(&context, step.Environment); err != nil {
+			ret = append(ret, err)
+		}
 	}
 	ret = append(ret, validateDependencies(context.fieldRoot, step.Dependencies)...)
 	ret = append(ret, validateLeases(context.forField(".leases"), step.Leases)...)
```
bbguimaraes, line 561: I missed this one: we'll also want to do a similar check in there to skip the duplicate test.

bbguimaraes (changes requested):
LGTM, but there's one conditional validation step I missed.

We also have some invalid configurations in the E2E tests now.

PR #1924, opened 2021-04-27
Shard tide merge method

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-2110

Requires https://github.com/kubernetes/test-infra/pull/21956 to merge first and us to update our test-infra dependency afterwards
/hold


stevekuznetsov (commented):
/lgtm
/hold

I still think that upstream - whether prow.k8s.io or anyone else - will want this functionality if they are to start using sharded config. As in, they will want to be able to blow up their monolith and will need better serialization. Would reduce our maintenance burden to try to have it there if feasible.

alvaroaleman:
Updated to include test-infra bump.

Requires:
* https://github.com/openshift/release-controller/pull/268
* https://github.com/openshift/ci-chat-bot/pull/153

PR #1926, opened 2021-04-27
pod-scaler: a server for automated resource requests

stevekuznetsov:
This commit introduces a new server to automate setting resource

recommendations and can also serve a web UI. The producer has previously
been merged, this patch provides the mutating admission webhook server
to set resource requests. The UI server for viewing the data will come
next.

While we roll out this feature, we need to be doing it in a
non-disruptive way. So, this patch will set requests and limits if and
only if the values we compute are larger than those already set on the
Pod by a user. We will then remove the configuration from the release
YAMLs, to allow the server to do what it needs to, always allowing users
to add configuration to explicitly turn up the requests past what we
think is appropriate.

/assign @alvaroaleman


stevekuznetsov:
Part of DPTP-1822

cmd/pod-scaler/admission.go
```diff
@@ -78,15 +84,94 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+			our := (*pair.ours)[field]
+			their := (*pair.theirs)[field]
+			if our.Cmp(their) == 1 {
+				(*pair.theirs)[field] = our
+			}
+		}
+	}
+}
+
+func mutatePodResources(pod *corev1.Pod, server *resourceServer) {
```
stevekuznetsov, line 126: note to self: we could handle rehearsals here by:

 - detecting with the rehearsal label
 - parsing the prowjob spec env as json
 - reading the extra_refs for the one with workdir:true
 - using those values for the meta we index into the cache with

cmd/pod-scaler/admission.go
```diff
@@ -78,15 +84,94 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+		{ours: &allOfOurs.Limits, theirs: &allOfTheirs.Limits},
+	} {
+		for _, field := range []corev1.ResourceName{corev1.ResourceCPU, corev1.ResourceMemory} {
+			our := (*pair.ours)[field]
+			their := (*pair.theirs)[field]
+			if our.Cmp(their) == 1 {
+				(*pair.theirs)[field] = our
+			}
+		}
+	}
```
stevekuznetsov, line 123: note to self: check that mem limits are still larger than requests, or set them to 120% of request if needed?

stevekuznetsov:
depends on https://github.com/openshift/ci-tools/pull/1929

ready for review

cmd/pod-scaler/admission.go
```diff
@@ -78,15 +93,151 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
 	return admission.PatchResponseFromRaw(req.Object.Raw, marshaledPod)
 }
 
-func mutatePod(pod *corev1.Pod, build *buildv1.Build) {
+// mutatePodMetadata updates metadata labels for Pods created by Prow for rehearsals,
+// where default metadata points to the release repo instead of the repo under test.
+// We can fix this by updating to use the values from the configuration that the job
+// ends up running with.
+func mutatePodMetadata(pod *corev1.Pod) error {
```
alvaroaleman, line 100: Do we really want to change the pod instead of only inferring that information for further processing?

cmd/pod-scaler/consumer.go
```diff
@@ -0,0 +1,128 @@
...
+func (c *cacheReloader) reload() {
+	// technically this can race as we read the attribute and data from the handle at
+	// different times, but there doesn't seem to be an atomic call to GCS for that anyway
+	lastUpdated, err := lastUpdated(c.cache, c.name)
+	if err != nil {
+		c.logger.WithError(err).Warn("Failed to query for last cache update time, won't reload this tick.")
+		return
+	}
+	c.lock.RLock()
+	lastSeen := c.lastUpdated
```
alvaroaleman, line 53: That doesn't seem to be a read

cmd/pod-scaler/main_test.go
```diff
@@ -0,0 +1 @@
+package main
```
alvaroaleman, line 1: ?

cmd/pod-scaler/testdata/zz_fixture_TestMutatePodResources_resources_to_add.yaml
```diff
@@ -0,0 +1,65 @@
+  &v1.Pod{
```
alvaroaleman, line 1: This doesn't seem to be yaml, where does it come from, come cmp output?

pkg/testhelper/testhelper.go
```diff
@@ -111,20 +109,8 @@ func CompareWithFixture(t *testing.T, output interface{}, opts ...Option) {
...
-		Context:  3,
-	}
-	diffStr, err := difflib.GetUnifiedDiffString(diff)
-	if err != nil {
-		t.Fatal(err)
-	}
-
-	if diffStr != "" {
-		t.Errorf("got diff between expected and actual result: \n%s\n\nIf this is expected, re-run the test with `UPDATE=true go test ./...` to update the fixtures.", diffStr)
+	if diff := cmp.Diff(expected, serializedOutput); diff != "" {
```
alvaroaleman, line 112: These need to be converted to strings first

cmd/pod-scaler/consumer.go
stevekuznetsov: What do you mean? We read `c.lastUpdated`

cmd/pod-scaler/admission.go
stevekuznetsov: I'm not sure - I think mentally thinking about "take a Pod, map it's labels to metric fields, then process the metric fields just as if it were coming from Prometheus" is valuable to reduce mental overhead, and mutating the metadata seemed like a way to add some optional complexity on top without changing the flow, but I'm open to other ideas. I like that this continues to feed everything into a metadata label --> metric label conversion.

cmd/pod-scaler/main_test.go
stevekuznetsov: lol

pkg/testhelper/testhelper.go
stevekuznetsov: Why? It seemed to work fine without that.

cmd/pod-scaler/testdata/zz_fixture_TestMutatePodResources_resources_to_add.yaml
stevekuznetsov: Yeah, it's a `cmp.Diff()` output - I was wondering the best way to do this ... basically the point of the `mutate` func is to apply some diff from a -> b ... it seemed like a good way to test that in a concise way would be to assert something about that diff. Asserting that the output object is in some specific shape runs into the issue you pointed out in the last PR - without the source object, it's hard to reason about.

petr-muller:
I did not do a line-by-line but the code LGTM

cmd/pod-scaler/testdata/zz_fixture_TestMutatePodResources_resources_to_add.yaml
alvaroaleman: Just serialize it at the end?

stevekuznetsov: If I serialize it, all the assert is about is the state of the Pod after we mutate the requests. That's not really useful, as you pointed out on the previous PR, as the important thing is the _change_ we make to the requests

cmd/pod-scaler/resources.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func (s *resourceServer) digestCPU(data *pod_scaler.CachedQuery) {
+	s.logger.Debugf("Digesting new CPU consumption metrics.")
+	s.digestData(data, cpuRequestQuantile, corev1.ResourceCPU, formatCPU())
+}
+
+const (
+	// memRequestQuantile is the quantile of memory usage data to use as the memory request
+	memRequestQuantile = 0.8
```
emilvberglind, line 73: Super pedantic, but do we need separate consts for cpu/mem since they are both `0.8`? Is the thought that they will diverge at some point?

cmd/pod-scaler/testdata/zz_fixture_TestMutatePodResources_resources_to_add.yaml
emilvberglind: It does seem a bit weird that it would exist as a yaml file since it isn't yaml. You couldn't just make the "expected" as a `v1.Pod` struct instance and diff that with the actual result?

cmd/pod-scaler/resources.go
stevekuznetsov: I do expect us to change these moving forward, and perhaps diverge them - we may be interested in making our CPU overcommit smaller (since it's much more fungible), etc.

cmd/pod-scaler/testdata/zz_fixture_TestMutatePodResources_resources_to_add.yaml
stevekuznetsov: The YAML bit is the diff library trying to be helpful - I can add a patch to make this a `.diff` if that would be useful.

cmd/pod-scaler/admission.go
stevekuznetsov: It's not really that - the labels relate to the commits under test, those that get reported to, etc - they are not "wrong" in the sense that they continue to convey the same information on rehearsals as they do on all jobs. It's just in the case of rehearsals, we can pull data from another source for the requests

cmd/pod-scaler/admission.go
```diff
@@ -87,15 +101,115 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
...
+			}
+		}
+	}
+}
+
+// reconcileLimits ensures that container resource limits do not set anything for CPU (as we
+// are fairly certain this is never a useful thing to do) and that the limits are >=200% of
+// requests (which they may not be any longer if we've changed requests)
+func reconcileLimits(resources *corev1.ResourceRequirements) {
+	if resources.Limits == nil {
```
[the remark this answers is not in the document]
stevekuznetsov: Otherwise the rest of it will NPE

cmd/pod-scaler/storage.go
```diff
@@ -205,3 +205,8 @@ func storeTo(storer storer, metricName string, data []byte) error {
 	}
 	return kerrors.NewAggregate(errs)
 }
+
+// lastUpdated determines the time at which the cache for this metric was last updated
+func lastUpdated(resolver attributeResolver, metricName string) (time.Time, error) {
+	return resolver.lastUpdated(interrupts.Context(), metricName+".json")
+}
```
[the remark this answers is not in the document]
stevekuznetsov: Empty file === i assert there is no diff

cmd/pod-scaler/admission.go
```diff
@@ -87,15 +101,115 @@ func (m *podMutator) Handle(ctx context.Context, req admission.Request) admissio
...
 		buildValue, buildHas := build.Labels[label]
 		_, podHas := pod.Labels[label]
 		if buildHas && !podHas {
 			pod.Labels[label] = buildValue
 		}
 	}
 }
+
+// useOursIfLarger updates fields in theirs when ours are larger
+func useOursIfLarger(allOfOurs, allOfTheirs *corev1.ResourceRequirements) {
```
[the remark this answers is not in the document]
stevekuznetsov: I don't see how changing the semantics to return the max instead of mutating changes much here. I didn't want to have to copy-paste the logic 4 times (requests, limits, etc) which is why the abstraction over the ResourceList was written. Should I just copy-paste it?

stevekuznetsov:
e2e finalized, comments addressed

stevekuznetsov:
Weird, not sure that e2e reproduces locally.

stevekuznetsov:
Finally passing all tests, ready for another review :)

cmd/pod-scaler/consumer.go
```diff
@@ -0,0 +1,130 @@
...
+		subscriber <- struct{}{}
+	}
+	c.lock.Unlock()
+	logger.Debug("Newer update loaded.")
+}
+
+func (c *cacheReloader) data() *pod_scaler.CachedQuery {
+	c.lock.RLock()
+	defer c.lock.RUnlock()
+	return c.lastLoaded
```
petr-muller, line 86: nit: I'm slightly triggered that you sometimes defer the unlock and sometimes you call it directly (`subscribe`), makes skimming harder when there's no pattern

petr-muller (approved):
/hold
LGTM, holding for others who did reviews here

PR #839, opened 2020-05-22
ci-operator: allow for dynamically resolved releases

stevekuznetsov:
Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @openshift/openshift-team-developer-productivity-platform @smarterclayton

PR #1927, opened 2021-04-27
pkg/release/official: Allow explicit version choices

wking:
With this commit, we extend that support to allow explicit version choices like:

```yaml
releases:
  firstz:
    release:
      channel: stable
      version: 4.7.0
latest:
  candidate:
    product: ocp
    stream: nightly
    version: "4.7"
```

to allow for tests updating between 4.7.0 and the most recent 4.7 nightly.  These tests will in turn alert us to issues like "something new has landed in recent 4.7 nightlies that breaks rollbacks to 4.7.0".

For reasons that are not clear to me, 33c2eaa49b (#839) did not just pass the configured channel along to Cincinnati, but instead appended the major/minor from the requested version.  Before this commit, that logic assumed that the requested version was just a major.minor, and then selected the largest SemVer version from that channel.  So this commit jumps through some hoops to extract the major.minor string from the configured release and append it to the configured channel.

`processVersionChannel` is set up so that if someone actually configured:

```yaml
releases:
  firstz:
    release:
      channel: stable-4.7
      version: 4.7.0
```

it would pass that `stable-4.7` channel through to Cincinnati.  We're unlikely to need that, because `pkg/validation`'s `validateRelease` restricts `channel` to `candidate`, `fast`, or `stable`.  But this package doesn't need to know that, and supporting `stable-4.7` puts us in a good position if `pkg/validation` relaxes a bit in the future.

I'm also fixing a 4.4 -> 4.3 typo in the tests, because real stable-4.4 results would not include 4.2.z releases.  I moved all of the test-cases over to 4.3 to keep them consistent.

### DIFF
--- a/pkg/defaults/defaults_test.go
+++ b/pkg/defaults/defaults_test.go
@@ -761,7 +761,7 @@ func (e environmentOverride) Get(name string) (string, error) {
 func TestFromConfig(t *testing.T) {
 	ns := "ns"
 	httpClient := release.NewFakeHTTPClient(func(req *http.Request) (*http.Response, error) {
-		content := `{"nodes": [{"version": "version", "payload": "payload"}]}`
+		content := `{"nodes": [{"version": "4.1.0", "payload": "payload"}]}`
 		return &http.Response{
 			StatusCode: http.StatusOK,
 			Body:       ioutil.NopCloser(bytes.NewBuffer([]byte(content))),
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
The previous work allowed such choices with the `prerelease` option and version selectors. Why is that insufficient?

stevekuznetsov:
Code looks reasonable, modulo test failures. Can we also get a ci-docs PR to show off this specific-version feature? Perhaps with a warning that it's a super-user option, not common and users should prefer floating tags/latest-in-channel

wking:
> The previous work allowed such choices with the `prerelease` option and version selectors. Why is that insufficient?

Because figuring out exclusive `version_bounds` is mental effort I don't want to invest, when I already know I want 4.7.1.

stevekuznetsov (approved):

PR #1929, opened 2021-04-28
rehearsal: label the context we're rehearsing

stevekuznetsov:
When a user looks at a Pod created by this tool, they cannot
programmatically get the context which the rehearsed job would have
triggered, as we change the context to prepend a rehearsal prefix, etc.
This change exposes the value of the context through labels.

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -37,7 +37,11 @@ import (
 )
 
 const (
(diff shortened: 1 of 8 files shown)


stevekuznetsov:
Part of DPTP-1822

alvaroaleman (approved):

alvaroaleman (approved):

PR #1931, opened 2021-04-29
ci-op: retry on conflicts when we patch ists

petr-muller:
https://github.com/openshift/ci-tools/pull/1928 assumed we do not need nto handle conflicts when we patch because they should not happen.

They happen.

This is suspected to be a bug in kube so I decided to keep this so that we have a repro (instead of switching it back to CreateOrUpdate where the conflict handling would not be surprising).


petr-muller:
build01 updates itself :/

PR #1905, opened 2021-04-22
[DPTP-2119] steps: clean up cluster claims when test ends

AlexNPavel:
This PR adds deletion functionality for cluster claims to be deleted
when a test using one ends. This is handled in the same way as leases,
where the test step is wrapped in the cluster claim step. This allows
the cluster claim step to delete the claim after the test ends.

/cc @hongkailiu

PR #1938, opened 2021-04-29
Dry out function aggregateWrappedErrorAndReleaseError

hongkailiu:
Follow up https://github.com/openshift/ci-tools/pull/1905#discussion_r622695659

/cc @stevekuznetsov

### DIFF
--- a/pkg/steps/lease.go
+++ b/pkg/steps/lease.go
@@ -118,8 +118,11 @@ func (s *leaseStep) run(ctx context.Context) error {
 	logrus.Infof("Releasing leases for test %s", s.Name())
 	releaseErr := results.ForReason("releasing_lease").ForError(releaseLeases(client, s.leases))
 
-	// we want a sensible output error for reporting, so we bubble up these individually
-	//if we can, as this is the only step that can have multiple errors
+	return aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr)
+}
+
+func aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr error) error {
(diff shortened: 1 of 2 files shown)


pkg/steps/lease.go
```diff
@@ -118,8 +118,11 @@ func (s *leaseStep) run(ctx context.Context) error {
 	logrus.Infof("Releasing leases for test %s", s.Name())
 	releaseErr := results.ForReason("releasing_lease").ForError(releaseLeases(client, s.leases))
 
-	// we want a sensible output error for reporting, so we bubble up these individually
-	//if we can, as this is the only step that can have multiple errors
+	return aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr)
+}
+
+func aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr error) error {
+	// we want a sensible output error for reporting, so we bubble up these individually if we can
```
alvaroaleman, line 125: The code below is useless, utilerrors.NewAggregate [filters out nil errors](https://github.com/openshift/ci-tools/blob/7ae0f5b757d5d9a0bedbcce80ee810460c3ce0bf/vendor/k8s.io/apimachinery/pkg/util/errors/errors.go#L53-L55) so you can just unconditionally do a `return utilerrors.NewAggreate([]error{wrapperdErr, releaseErr}`

stevekuznetsov: Not true. We do not want an aggregate, we want the wrapped error, if we can. That's the point of this code.

stevekuznetsov:
/lgtm
Part of DPTP-2099

PR #1943, opened 2021-04-30
Fix ci-secret-bootstrap user sync

alvaroaleman:
There were two issues:
* We had a Config struct in our options, but didn't unmarshal into that
  but into a different var, then filter the secrets, the copy only the
  secrets property in our config in the options, dropping all other
  settings including the UserSecretsTargetClusters along the way
* We operate under a prefix, but the user secrets are elsewhere. This
  adds a new option for a dptp prefix that gets appended to the normal
  prefix, which allows us to set the normal prefix to the basepath of
  the kv store

completed by https://github.com/openshift/release/pull/18216

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -119,8 +119,7 @@ func (o *options) completeOptions(censor *secrets.DynamicCensor) error {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


stevekuznetsov (approved):
We should likely rename some of these types, the BW stuff is pretty jarring to read and comprehend :)

PR #1948, opened 2021-05-03
DPTP-1463: Log the source from where an image was resolved

emilvberglind:
[DPTP-1463](https://issues.redhat.com/browse/DPTP-1463): Log the source from where an image was resolved. For example, if an image is resolved as a result of a step-ref test, then output the name of the test.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -452,6 +452,11 @@ type StepConfiguration struct {
 type InputImageTagStepConfiguration struct {
 	BaseImage ImageStreamTagReference         `json:"base_image"`
 	To        PipelineImageStreamTagReference `json:"to,omitempty"`
+
+	// The Source denotes what caused this image to get loaded. In the event
+	// that the image was loaded as a result of a test step, the source will
+	// contain the name of the test.
+	Source string `json:"source,omitempty"`
 }
 
 // OutputImageTagStepConfiguration describes a step that
@@ -1488,4 +1493,9 @@ const (
 	ReleaseImageStream = "release"
 
 	ComponentFormatReplacement = "${component}"
+
+	ImageStreamSourceRoot    = "root"
+	ImageStreamSourceBase    = "base_image"
+	ImageStreamSourceBaseRpm = "base_rpm_image"
+	ImageStreamSourceTest    = `test step "%s"`
 )
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -373,6 +373,7 @@ func stepsForStepImages(
 			config := api.InputImageTagStepConfiguration{
 				BaseImage: *subStep.FromImage,
 				To:        link,
+				Source:    fmt.Sprintf(api.ImageStreamSourceTest, subStep.As),
 			}
 			if _, ok := inputImages[config]; ok {
 				continue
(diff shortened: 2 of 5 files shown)


pkg/defaults/defaults.go
```diff
@@ -373,6 +373,7 @@ func stepsForStepImages(
 			config := api.InputImageTagStepConfiguration{
 				BaseImage: *subStep.FromImage,
 				To:        link,
+				Source:    fmt.Sprintf(api.ImageStreamSourceTest, subStep.As),
```
alvaroaleman, line 376: We use the `config` struct as map key to deduplicate, that will break now if different steps use the same `Source`

emilvberglind: Do you know where that happens so I can take a look and see about tweaking that deduplication?

alvaroaleman: two lines below, the `if _, ok := inputImages[config]; ok { continue }`

The implication is also that there could be more than one Source I think

emilvberglind: Cool, thanks! Still getting used to Go's syntax :)

When you say that there could be more than one Source, how would that happen?

alvaroaleman: Multiple steps that all need the same base image. The fact that we de-duplicate them seems to imply that is possible, but I am not super familiar with this part of the code base. @bbguimaraes might know more

petr-muller: Yes, what Alvaro says is correct. We may get duplicates by e.g. multiple multi-stage steps using the same image in their `from_image` staza, or having the same thing in a `base_images` item and as a multi-stage step image, etc. 

Not sure if we validate this, but this could be a minimal reproducer if we don't:

```yaml
base_images:                                                                                                                                                                                                                                                                              
  base:
    name: "4.8"
    namespace: ocp
     tag: base
  all-your-bases:
    name: "4.8"
    namespace: ocp
    tag: base
```

petr-muller: In reality this is most likely to happen when multiple steps need the same image

emilvberglind: Ah, I gotcha. Ok, that makes sense.

petr-muller: To super-clarify, here's a step that uses an image: https://github.com/openshift/release/blob/master/ci-operator/step-registry/gather/loki/gather-loki-ref.yaml#L4-L7.

We'll have duplicate sources if there are more steps with the idential `from_image`, or if a this step is used from a config that importes the same image in `base_images`.

pkg/api/types.go
```diff
@@ -452,6 +452,11 @@ type StepConfiguration struct {
 type InputImageTagStepConfiguration struct {
 	BaseImage ImageStreamTagReference         `json:"base_image"`
 	To        PipelineImageStreamTagReference `json:"to,omitempty"`
+
+	// The Source denotes what caused this image to get loaded. In the event
+	// that the image was loaded as a result of a test step, the source will
+	// contain the name of the test.
+	Source string `json:"source,omitempty"`
```
stevekuznetsov, line 459: IMO this should not be an externally-visible (exported, JSON-serialized) field, since it's set internally

PR #1949, opened 2021-05-03
Ci-Secret-Bootstrap: Fix unused check and avoid silently swallowing panics

alvaroaleman:


cmd/ci-secret-bootstrap/main.go
```diff
@@ -858,38 +866,43 @@ func main() {
...
 		logrus.WithError(err).Fatal("Failed to create client.")
 	}
-	logrus.RegisterExitHandler(func() {
+
+	if errs := reconcileSecrets(o, client); len(errs) > 0 {
+		logrus.WithError(utilerrors.NewAggregate(errs)).Fatalf("errors while updating secrets")
+	}
+}
+
+func reconcileSecrets(o options, client secrets.ReadOnlyClient) (errs []error) {
```
hongkailiu, line 881: Putting `if o.validateOnly {}` inside `reconcileSecrets` sounds a bit odd.
Maybe moving it out?

alvaroaleman: It uses the client, that's why. I just moved all the code inside the new `reconcileSecrets`. If you have an idea for a better name, I am all ears. But we can not move it out because we need this function to define the scope of the `defer Logout` call.

hongkailiu: > you have an idea for a better name

Oh No. That is really not my strength.

/lgtm

PR #1957, opened 2021-05-05
Update controller-runtime to the latest

hongkailiu:
This PR is generated from
```console
$ go get -u sigs.k8s.io/controller-runtime@master
```

I would like to use https://github.com/kubernetes-sigs/controller-runtime/commit/2a2f6b4ef26cd4201d84c051663c3bf0678477e3

/cc @alvaroaleman

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -66,7 +66,7 @@ require (
 	k8s.io/test-infra v0.0.0-20210504134511-af1ac03335dc
 	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009
 	sigs.k8s.io/boskos v0.0.0-20210210143059-9ac98d864d2a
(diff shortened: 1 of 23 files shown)


alvaroaleman:
err, did you run `go mod tidy`? That seems like a lot of changes, even though we updated controller-runtime recently. I usually just edit the go.mod file and put the sha after the dependency, then run `go mod tidy && go mod vendor`, that way we don't update everything

alvaroaleman (approved):

hongkailiu:
> err, did you run `go mod tidy`? That seems like a lot of changes, even though we updated controller-runtime recently. I usually just edit the go.mod file and put the sha after the dependency, then run `go mod tidy && go mod vendor`, that way we don't update everything

Yes I did run `go mod tidy && go mod vendor`. I run it via `make validate-vendor` which is why our check passed.

Did it your way this time.

PR #1958, opened 2021-05-06
Implement Watch in fake watching client

hongkailiu:
/cc @alvaroaleman


alvaroaleman (commented):
/hold

You can just delete this package, together with `watchingclient`, its upstream now

PR #1960, opened 2021-05-06
Adding nameservers to dns config

deepsm007:
DPTP-1260

signed-off-by: Deep Mistry


pkg/steps/multi_stage_test.go
```diff
@@ -132,7 +132,8 @@ func TestGeneratePods(t *testing.T) {
 			MultiStageTestConfigurationLiteral: &api.MultiStageTestConfigurationLiteral{
 				ClusterProfile: api.ClusterProfileAWS,
 				DNSConfig: &api.StepDNSConfig{
-					Searches: []string{"my.dns.search1", "my.dns.search2"},
+					Nameservers: []string{"nameserver1", "nameserver2"},
```
alvaroaleman, line 135: If this doesn't result in a change in the fixtures it doesn't work

pkg/validation/test.go
```diff
@@ -665,10 +665,13 @@ func validateDependencies(fieldRoot string, dependencies []api.StepDependency) [
...
 }
 
-func validateDNSConfig(fieldRoot string, searches []api.StepDNSConfig) (ret []error) {
+func validateDNSConfig(fieldRoot string, dnsConfig []api.StepDNSConfig) (ret []error) {
 	var errs []error
-	for i, search := range searches {
-		if search.Searches[i] == "" {
+	for i, dnsconfig := range dnsConfig {
+		if dnsconfig.Nameservers[i] == "" {
+			errs = append(errs, fmt.Errorf("%s.nameservers[%d] must be set", fieldRoot, i))
```
alvaroaleman, line 672: Didn't you say it is optional?

deepsm007: The field itself is optional but we require it to provide entry for our coredns pod running on sshd-bastion for step-registry migration of the MA jobs.

alvaroaleman: Yeah but right now if DNSConfig is set it must set both Nameservers and Searched. I would have thought that one of them can be enough, is that wrong?

deepsm007: Searches is only going to work if we pass in the ip to the openshift cluster's nameserver so it can resolve the coredns-ppc64le.bastion or coredns-z.bastion.... address

alvaroaleman (approved):

PR #1961, opened 2021-05-07
Refactor ci-secret-bootstrap dry logger to output data to a given file

droslean:
In the effort of [DPTP-2169](https://issues.redhat.com/browse/DPTP-2169)

Example output:

```yaml
app.ci:
- data:
    ssh_host_dsa_key: .....
  metadata:
    creationTimestamp: null
    labels:
      dptp.openshift.io/requester: ci-secret-bootstrap
    name: sshd-slack-bot-alpha-host-keys
    namespace: ci
  type: Opaque
- data:
    authorized_keys: .....==
  metadata:
    creationTimestamp: null
    labels:
      dptp.openshift.io/requester: ci-secret-bootstrap
    name: sshd-slack-bot-alpha-authorized-keys
    namespace: ci
  type: Opaque
build01:
- data:
    authorized_keys: .....=
  metadata:
    creationTimestamp: null
    labels:
      dptp.openshift.io/requester: ci-secret-bootstrap
    name: sshd-authorized-keys
    namespace: bastion-z
  type: Opaque
- data:
    authorized_keys: ......=
  metadata:
    creationTimestamp: null
    labels:
      dptp.openshift.io/requester: ci-secret-bootstrap
    name: sshd-authorized-keys
    namespace: bastion-ppc64le
  type: Opaque
```

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -8,7 +8,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"io"
 	"io/ioutil"
 	"os"
 	"runtime"
@@ -17,6 +16,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/semaphore"
 
@@ -24,11 +24,10 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	kubejson "k8s.io/apimachinery/pkg/runtime/serializer/json"
+	kuberuntime "k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/kubernetes/scheme"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	"k8s.io/test-infra/prow/flagutil"
@@ -56,6 +55,7 @@ type options struct {
 	cluster             string
 	logLevel            string
 	impersonateUser     string
+	dryLogOutput        string
 
 	maxConcurrency int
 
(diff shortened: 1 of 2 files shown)


alvaroaleman:
why?

droslean:
For better output that can be used in integration tests as well.

alvaroaleman:
> For better output that can be used in integration tests as well.

Why do they require a single file? This makes the output from our production config quite illegible.

droslean:
The output is only the list of secrets that will be created in the cluster, mapped by the cluster's name. This PR adds the ability to send the output to a given file with better output as well. Before we were just creating temporary files with a random prefix, and the contents were just the secret, following with `------` line, which makes it more confusable.

stevekuznetsov:
I think for an _integration_ test what I would have expected is Go code that:

1. starts a Vault server
2. creates fake k8s clients in code
2. runs the main `ci-secret-bootstrap` with the fake clients
3. validates that the fakes had the correct interaction

droslean:
I wasn't sure about the `3`. Because when the binary will stop, the fake clients will be gone...I assumed that it is better to just generate a dry-run log and then the integration tests can just compare the expected output.

stevekuznetsov:
I didn't mean running the binary, but running the methods, from Go code. An issue with testing for dry-run output from an end-to-end perspective is that you're testing by proxy - in an integration test with fake clients you can test that the objects we send to the client are actually what we expect - not only their names but their values, etc. In the case where we're asserting something about the logging output, we are only testing that the correct thing is printed to a file. It's possible to print the correct things but be 100% broken because we incorrectly use the client, so it's a test by proxy behavior and therefore fragile. Sometimes it's the best we can do, but it isn't a replacement.

droslean:
We can just land this code, but not use it for integration tests.

stevekuznetsov:
To note, we take that approach in the `ci-operator` end-to-end tests. The reason is two-fold, though:

 - we are testing really complex interactions between many controllers (ImageStream, kubelet, etc) and mocking them out in an integration fake is impractical
 - `ci-operator` is to some degree more focused on running user workloads than the specific objects that exist, at an end-to-end/user-focused level
 - we were lazy (!!)

To explain that little more, a good end-to-end test is going to assert something about the end-user behavior, the most outward-facing or general thing. So, for instance, "running `ci-operator` when I target a specific job ends up running my job's commands" or something - not saying anything specific about how those jobs are run, for instance not specifying that we have a specific namespace for it, or that we use Pods to run steps, etc. So from an idealistic perspective a good end-to-end test for `ci-operator` should _not_ say anything about the objects (in many cases). In some cases, like the `[images]` jobs, we do kind of want to say something about the output `ImageStreamTag`s. However, we did not invest in the effort necessary to test output objects. So in part because it's not a good idea and in part because we were lazy, we do not test output objects for `ci-operator` end-to-end and instead test by proxy in logs.

For the secret tools, however, you have to consider - what is the end-user behavior? The outward-facing effect of running this tool _is_ the `Secret` content. The `Secret`s are not an implementation detail of how the `ci-secret-bootstrap` does something else for users, they are the one and only output of the tool. So, a good end-to-end test would assert specifically what those `Secret`s are, what namespaces they are in, what their data is, etc. We *could* do that with an ephemeral OCP cluster (or, N clusters) to run against, but it's very heavy-weight. Asserting at the client fake level ensures that we call the client correctly, which gives us pretty good confidence that the tool is correct. We could still incorrectly create the client (or read the kubeconfig incorrectly) or something, but the possible errors are much smaller in scope.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -565,39 +570,35 @@ func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secret
...
 		}
+		objects[cluster] = s
 	}
-	return nil
-}
 
-func writeSecretsToFile(secrets []*coreapi.Secret, w io.Writer) error {
-	serializerOptions := kubejson.SerializerOptions{Yaml: true, Pretty: true, Strict: true}
-	serializer := kubejson.NewSerializerWithOptions(kubejson.DefaultMetaFactory, scheme.Scheme, scheme.Scheme, serializerOptions)
+	data, err := yaml.Marshal(objects)
```
stevekuznetsov, line 584: If we are serializing objects, we must use the correct scheme. This change is incorrect.

droslean: We don't serialize anything now, so the current code wasn't needed. That is why I changed it. Either way. I will close this PR and work on the integration tests. Thanks a lot for the explanation!

droslean: /close

PR #1962, opened 2021-05-07
DPTP-1463: Log the source from where an image was resolved

emilvberglind:
Maintain running list of resolved images along with their sources as we generate steps.
Track sources associated with a particular image on the `InputImageTagStepConfiguration`.
Create new `InputImage` interface to use for deduplication purposes.


pkg/api/types.go
```diff
@@ -450,10 +452,81 @@ type StepConfiguration struct {
 // if no explicit output tag is provided, the name
 // of the image is used as the tag.
 type InputImageTagStepConfiguration struct {
+	InputImage `json:",inline"`
+	Sources    *[]ImageStreamSource `json:"-"`
```
stevekuznetsov, line 456: Why is this a pointer to a slice?

pkg/api/types.go
```diff
@@ -450,10 +452,81 @@ type StepConfiguration struct {
...
+		if len(tests) > 0 {
+			var testNames []string
+			for k := range tests {
+				testNames = append(testNames, k)
+			}
+			sort.Strings(testNames)
+			formattedSources += fmt.Sprintf("test steps: %s", strings.Join(testNames, ","))
+		}
+
+		return strings.TrimSuffix(formattedSources, ",")
```
stevekuznetsov, line 485: nit: you could also keep `formattedSources` as a slice and return `strings.Join(formattedSources, ",")`

pkg/api/types.go
```diff
@@ -450,10 +452,81 @@ type StepConfiguration struct {
...
+	ImageStreamSourceBaseRpm ImageStreamSourceType = "base_rpm_image"
+	ImageStreamSourceTest    ImageStreamSourceType = "test step"
+)
+
+type ImageStreamSource struct {
+	SourceType ImageStreamSourceType
+	Name       string
+}
+
+func NewImageStreamSource(sourceType ImageStreamSourceType, source *string) ImageStreamSource {
```
stevekuznetsov, line 519: What's the utility of passing in a pointer for `source` here? If we don't pass that in, can users just create literals instead of calling this?

pkg/api/types_test.go
```diff
@@ -239,3 +240,106 @@ func TestIsBundleImage(t *testing.T) {
...
+			{
+				SourceType: ImageStreamSourceTest,
+				Name:       "test2",
+			},
+		},
+		expectedFormattedSources: "root,base_image,base_rpm_image,test steps: test1,test2",
+	}}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			if len(testCase.sources) > 0 {
```
stevekuznetsov, line 326: Why do we need this guard?

pkg/defaults/defaults_test.go
```diff
@@ -42,6 +42,9 @@ func TestStepConfigsForBuild(t *testing.T) {
 	noopResolver := func(root, cache *api.ImageStreamTagReference) (*api.ImageStreamTagReference, error) {
 		return root, nil
 	}
+	rootSources := []api.ImageStreamSource{api.NewImageStreamSource(api.ImageStreamSourceRoot, nil)}
```
stevekuznetsov, line 45: nit: in my personal mental framework for tests, there's no such thing as "DRY" and making separate, independent tests depend on the same construct is an antipattern as it reduces failure specificity 

(however, if we do not have a pointer to a slice in the sources member you can just inline these, which would fix it)

pkg/api/types.go
emilvberglind: Per our IM convo, I think this was me getting mixed up with some of the slice mutation things we talked about re: the slice headers. Going to try to change this so that the `config` on `inputImageTagStep` is a pointer so that when that config is mutated, it is reflected in any step that holds that config.

pkg/api/types.go
```diff
@@ -450,10 +452,70 @@ type StepConfiguration struct {
...
+}
+
+func (config InputImageTagStepConfiguration) Matches(other InputImage) bool {
+	return config.InputImage == other
+}
+
+func (config InputImageTagStepConfiguration) FormattedSources() string {
+	var formattedSources []string
+	tests := sets.String{}
+	if len(config.Sources) > 0 {
```
stevekuznetsov, line 466: You don't need this `if` guard

pkg/api/types.go
```diff
@@ -450,10 +452,70 @@ type StepConfiguration struct {
...
+				formattedSources = append(formattedSources, string(source.SourceType))
+			}
+		}
+
+		if len(tests) > 0 {
+			var testNames []string
+			for k := range tests {
+				testNames = append(testNames, k)
+			}
+			sort.Strings(testNames)
```
stevekuznetsov, line 481: nit: `sets.String()` returns a sorted list with `.List()` so you can simplify:

```go
if tests.Len() > 0 {
formattedSources = append(formattedSources, fmt.Sprintf("test steps: %s", strings.Join(tests.List(), ",")))
}
```

pkg/api/types.go
```diff
@@ -450,10 +452,70 @@ type StepConfiguration struct {
...
+		}
+
+		return strings.Join(formattedSources, ",")
+	} else {
+		return ""
+	}
+}
+
+func (config *InputImageTagStepConfiguration) AddSources(sources ...ImageStreamSource) *InputImageTagStepConfiguration {
+	if config.Sources == nil {
```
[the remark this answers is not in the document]
emilvberglind: Yeah, that was an artifact left there from when I had Sources as a pointer.

pkg/api/types.go
```diff
@@ -450,10 +452,70 @@ type StepConfiguration struct {
...
+			formattedSources = append(formattedSources, fmt.Sprintf("test steps: %s", strings.Join(testNames, ",")))
+		}
+
+		return strings.Join(formattedSources, ",")
+	} else {
+		return ""
+	}
+}
+
+func (config *InputImageTagStepConfiguration) AddSources(sources ...ImageStreamSource) *InputImageTagStepConfiguration {
```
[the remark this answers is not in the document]
emilvberglind: Makes sense.

stevekuznetsov (approved):

PR #1965, opened 2021-05-07
cluster_claim: always delete claim

AlexNPavel:
The current code immediately exits with an error if the `acquireClaim`
function fails without deleting the created claim. This PR makes the
`acquireClaim` function always return the claim if one is made, even on
error, and the `releaseCluster` function is run even when `aquireClaim`
fails.

/cc @hongkailiu

### DIFF
--- a/pkg/steps/cluster_claim.go
+++ b/pkg/steps/cluster_claim.go
@@ -63,7 +63,13 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
 	}
 	clusterClaim, err := acquireCluster(ctx, *s.clusterClaim, s.hiveClient, s.client, *s.jobSpec)
 	if err != nil {
-		return results.ForReason("acquiring_cluster_claim").ForError(err)
+		acquireErr := results.ForReason("acquiring_cluster_claim").ForError(err)
+		// always attempt to delete claim if one exists
+		var releaseErr error
+		if clusterClaim != nil {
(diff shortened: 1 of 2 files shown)


pkg/steps/cluster_claim.go
```diff
@@ -63,7 +63,13 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
 	}
 	clusterClaim, err := acquireCluster(ctx, *s.clusterClaim, s.hiveClient, s.client, *s.jobSpec)
 	if err != nil {
-		return results.ForReason("acquiring_cluster_claim").ForError(err)
+		acquireErr := results.ForReason("acquiring_cluster_claim").ForError(err)
+		// always attempt to delete claim if one exists
+		var releaseErr error
+		if clusterClaim != nil {
```
hongkailiu, line 69: Maybe push line76 (the logging line) to the function `releaseCluster` so that we can see it in both cases of release?

pkg/steps/cluster_claim.go
```diff
@@ -130,32 +136,32 @@ func acquireCluster(ctx context.Context, clusterClaim api.ClusterClaim, hiveClie
 		}
 		return false, nil
 	}); err != nil {
-		return nil, fmt.Errorf("failed to wait for created cluster claim to become running: %w", err)
+		return claim, fmt.Errorf("failed to wait for created cluster claim to become running: %w", err)
```
hongkailiu, line 139: Add a test for this? Checking it is not nil at error?

hongkailiu (approved):

PR #1967, opened 2021-05-09
Remove references of api.ci

hongkailiu:
User secret management is Vault now.

/cc @alvaroaleman


hongkailiu:
https://issues.redhat.com/browse/DPTP-2048

pkg/config/release.go
```diff
@@ -37,7 +37,7 @@ const (
 	TemplatesPath = "ci-operator/templates"
 	// ClusterProfilesPath is where profiles are stored in the release repo
 	ClusterProfilesPath = "cluster/test-deploy"
-	// StagingNamespace is the staging namespace in api.ci
+	// StagingNamespace is the staging namespace in app.ci
```
alvaroaleman, line 40: That ns doesn't exit on app.ci

pkg/controller/registrysyncer/registrysyncer.go
```diff
@@ -332,7 +332,7 @@ func dockerImageImportedFromTargetingCluster(cluster string, tag *imagev1.ImageS
 		return false
```
alvaroaleman, line 332: If we remove the plumbing in main, we should also remove the controller

alvaroaleman (commented):
If we remove the plumbing for the registrysyncer, secretsyncer and imagepushed, we should also remove the controllers entirely (But we don't have to, we can just deactivate them)

cmd/dptp-controller-manager/main.go
```diff
@@ -25,37 +25,27 @@ import (
...
-	"sigs.k8s.io/yaml"
 
 	imagev1 "github.com/openshift/api/image/v1"
 
-	"github.com/openshift/ci-tools/pkg/api/secretbootstrap"
-	"github.com/openshift/ci-tools/pkg/controller/imagepusher"
 	"github.com/openshift/ci-tools/pkg/controller/promotionreconciler"
-	"github.com/openshift/ci-tools/pkg/controller/registrysyncer"
 	"github.com/openshift/ci-tools/pkg/controller/secretsyncer"
-	secretsyncerconfig "github.com/openshift/ci-tools/pkg/controller/secretsyncer/config"
```
alvaroaleman, line 38: remove pkg/controller/secretsyncer if you remove the plumbing

pkg/config/release.go
hongkailiu: The constant is used in some of our tools.
I will handle this with another PR to keep the damage isolated.

cmd/dptp-controller-manager/main.go
hongkailiu: Updated.

alvaroaleman (approved):

PR #1971, opened 2021-05-10
Applying custom dnsPolicy for pod spec

deepsm007:
[DPTP-1260](https://issues.redhat.com/browse/DPTP-1260)
Signed-off-by: Deep Mistry

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -482,6 +482,9 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			}
 			pod.Spec.DNSConfig.Nameservers = append(pod.Spec.DNSConfig.Nameservers, step.DNSConfig.Nameservers...)
 			pod.Spec.DNSConfig.Searches = append(pod.Spec.DNSConfig.Searches, step.DNSConfig.Searches...)
+			if len(pod.Spec.DNSConfig.Nameservers) > 0 {
+				pod.Spec.DNSPolicy = coreapi.DNSNone
+			}
 		}
 		pod.Spec.Volumes = append(pod.Spec.Volumes, coreapi.Volume{Name: homeVolumeName, VolumeSource: coreapi.VolumeSource{EmptyDir: &coreapi.EmptyDirVolumeSource{}}})
 		pod.Spec.Volumes = append(pod.Spec.Volumes, secretVolumes...)
--- a/pkg/steps/multi_stage_test.go
+++ b/pkg/steps/multi_stage_test.go
@@ -125,18 +125,13 @@ func TestGeneratePods(t *testing.T) {
 					Commands: "command1",
 				}, {
 					As: "step2", From: "stable-initial:installer", Commands: "command2", RunAsScript: &yes,
+				}, {
+					As: "step3", From: "src", Commands: "command3", DNSConfig: &api.StepDNSConfig{
+						Nameservers: []string{"nameserver1", "nameserver2"},
+						Searches:    []string{"my.dns.search1", "my.dns.search2"},
+					},
 				}},
-			},
-		}, {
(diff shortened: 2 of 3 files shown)


pkg/validation/test.go
```diff
@@ -665,6 +665,15 @@ func validateDependencies(fieldRoot string, dependencies []api.StepDependency) [
 	return errs
 }
 
+func validateDNSPolicy(fieldRoot string, test api.TestStepConfiguration) (ret []error) {
+	var errs []error
+	if dnsPolicy := test.MultiStageTestConfiguration.DNSPolicy; dnsPolicy != "" && (api.DNSClusterFirstWithHostNet != dnsPolicy && api.DNSClusterFirst != dnsPolicy && api.DNSDefault != dnsPolicy && api.DNSNone != dnsPolicy) {
+		errs = append(errs, fmt.Errorf("%s.dnsPolicy is not a valid dns policy: %s", fieldRoot, string(dnsPolicy)))
+	}
+	fmt.Println(errs)
```
alvaroaleman, line 673: Debugging remainder?

pkg/steps/multi_stage.go
```diff
@@ -476,6 +476,7 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 		pod.Labels[MultiStageTestLabel] = s.name
 		pod.Spec.ServiceAccountName = s.name
 		pod.Spec.TerminationGracePeriodSeconds = terminationGracePeriodSeconds
+		pod.Spec.DNSPolicy = coreapi.DNSPolicy(step.DNSPolicy)
```
alvaroaleman, line 479: This needs to be conditional and not do anything if `step.DNSPolicy` is an empty string (please be sure to test this, it will break literally all jobs if this isn't correct)

alvaroaleman: Also, can't we derive this? If a DNSServer is set, it should be `None`, otherwise we don't set it?

pkg/validation/test.go
deepsm007: missed that one 😅

stevekuznetsov (commented):
Could you think of an e2e test for this behavior?

pkg/steps/multi_stage.go
```diff
@@ -480,6 +480,9 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			if pod.Spec.DNSConfig == nil {
 				pod.Spec.DNSConfig = &coreapi.PodDNSConfig{}
 			}
+			if len(pod.Spec.DNSConfig.Nameservers) > 0 {
+				pod.Spec.DNSPolicy = coreapi.DNSNone
```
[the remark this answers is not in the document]
deepsm007: https://github.com/openshift/ci-tools/blob/master/vendor/k8s.io/api/core/v1/types.go#L2607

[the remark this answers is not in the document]
alvaroaleman: Yeah, we can and should remove the config option, this just needs to set it to `None` when a DNSServer is configured

pkg/steps/multi_stage_test.go
```diff
@@ -125,18 +125,31 @@ func TestGeneratePods(t *testing.T) {
 					Commands: "command1",
 				}, {
 					As: "step2", From: "stable-initial:installer", Commands: "command2", RunAsScript: &yes,
+				}, {
+					As: "step3", From: "src", Commands: "command3", DNSConfig: &api.StepDNSConfig{
+						Nameservers: []string{"nameserver1", "nameserver2"},
+						Searches:    []string{"my.dns.search1", "my.dns.search2"},
```
alvaroaleman, line 131: Just the `step3` should be enough

alvaroaleman (approved):

PR #1972, opened 2021-05-10
secret bootstrap: create nonexistent namespaces

petr-muller:
Resolves failures such as [this one](https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-private/logs/periodic-ci-secret-bootstrap/1391791192831692800#1:build-log.txt%3A3866):

```
failed to update secrets: [error creating secret api.**:grafana-loki/grafana-config: namespaces \"grafana-loki\" not found ...
```

I'm convinced that creating namespaces is the right thing to do for this automation b/c we have no other way how to enforce a presence of certain namespace on all clusters.

I added a test to describe that this is an intentional behavior, but unfortunately the test passes even if I comment the new code out: the underlying fake client do not seem to be able to fake an error like that (or I did not find a way to coerce it).

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -59,7 +59,7 @@ type options struct {
 
 	maxConcurrency int
 
-	secretsGetters  map[string]coreclientset.SecretsGetter
+	secretsGetters  map[string]Getter
 	config          secretbootstrap.Config
 	generatorConfig secretgenerator.Config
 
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -512,7 +517,19 @@ func updateSecrets(secretsGetters map[string]coreclientset.SecretsGetter, secret
 			logger := logger.WithFields(logrus.Fields{"namespace": secret.Namespace, "name": secret.Name, "type": secret.Type})
 			logger.Debug("handling secret")
 
-			secretClient := secretsGetters[cluster].Secrets(secret.Namespace)
+			nsClient := getters[cluster].Namespaces()
+			if _, err := nsClient.Get(context.TODO(), secret.Namespace, metav1.GetOptions{}); err != nil {
```
droslean, line 521: IMHO we can just try to create the namespace and if it already exists, we can eat the error?

petr-muller: We can. Why is it better?

droslean: fewer API calls.

droslean (approved):

PR #1974, opened 2021-05-11
DPTP-1827  Output environment variables for chains and workflows

emilvberglind:
With [DPTP-1827](https://issues.redhat.com/browse/DPTP-1827):  we want to not only output the environment variables for specific steps, but also for chains and workflows


emilvberglind:
Test deployment at: https://server-ci-tools-1974.apps.ci.l2s4.p1.openshiftapps.com/

PR #1977, opened 2021-05-12
Fix an ineffective break and an incomplete nilcheck

alvaroaleman:


cmd/entrypoint-wrapper/main.go
```diff
@@ -193,7 +193,7 @@ func execCmd(argv []string) error {
 		for {
 			select {
 			case <-ctx.Done():
-				break
+				return
```
alvaroaleman, line 196: A `break` here just breaks the select and not the loop

bbguimaraes: :facepalm:

PR #1984, opened 2021-05-13
[DPTP-2169] Add integration tests for ci-secret-bootstrap pipeline

droslean:

### DIFF
--- a/cmd/ci-secret-bootstrap/main_test.go
+++ b/cmd/ci-secret-bootstrap/main_test.go
@@ -20,6 +20,7 @@ import (
 	coreapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/fake"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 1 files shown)


alvaroaleman:
DPTP-2169

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,435 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+	type vaultObject struct {
+		field string
+		data  []byte
+	}
+
+	testCases := []struct {
+		id              string
+		config          secretbootstrap.Config
+		secretGetters   map[string]coreclientset.SecretsGetter
+		vaultObjects    map[string][]vaultObject
```
alvaroaleman, line 2648:
```suggestion
		vaultObjects    map[string]map[string]string
```

The data structure for the content of an item used by vault is a` map[string]string`, so making it a `[]struct{field string, data []byte}` introduces some potential for writing incorrect and hard to understand tests

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,435 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+		expectedSecrets map[string][]coreapi.Secret
+		expectedErrors  bool
+	}{
+		{
+			id: "happy case",
```
alvaroaleman, line 2653:
```suggestion
			id: "Successfuly create secret from config",
```

There are other possible happy cases (e.G. `Update Secret from config`, `Create Secret from Vault User Secret`, `Update Secret from Vault User Secret`)

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,435 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+					{
+						ObjectMeta: metav1.ObjectMeta{Name: "prod-secret-1", Namespace: "namespace-1", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap"}},
+						Data:       map[string][]byte{"key-name-1": []byte("secret-data")},
+						Type:       coreapi.SecretTypeOpaque,
+					},
+				},
+			},
+		},
+		{
+			id: "happy case, multiple clusters",
```
alvaroaleman, line 2695:
```suggestion
			id: "Successflly creating secret from config in multiple clusters",
```

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,435 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+							"some-data-key":                []byte("a-secret"),
+							"secretsync-vault-source-path": []byte("secret/user-item-1"),
+						},
+						Type: coreapi.SecretTypeOpaque,
+					},
+				},
+			},
+		},
+		{
+			id: "sad case",
```
alvaroaleman, line 2972: Please describe how its sad

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,435 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+			},
+			vaultObjects: map[string][]vaultObject{
+				"no-one-wants-me": {
+					{
+						field: "field-name-1",
+						data:  []byte("secret-data"),
+					},
+				},
+			},
+			expectedErrors: true,
```
alvaroaleman, line 3003: Please do a better error check (error message, error substring). If we fail for an unrelated reason the test will incorrectly succeed.

alvaroaleman (commented):
There are no tests at all where we merge the content of a config secret with the one of a user secret and no tests at all where we update.

To avoid duplicating everything for the update case, you could do another layer of subtesting where you insert the `expectedSecrets` into the clients and manipulate their value

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,435 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+					t.Fatal("expected errors but got nothing")
+				}
+			} else {
+				if len(errs) > 0 {
+					t.Fatalf("errors weren't expected but got: %v", utilerrors.NewAggregate(errs))
+				}
+
+				actualSecretsByCluster := make(map[string][]coreapi.Secret)
+				for cluster, secretGetter := range tc.secretGetters {
+					actualSecrets, err := secretGetter.Secrets("namespace-1").List(context.Background(), metav1.ListOptions{})
```
alvaroaleman, line 3053:
```suggestion
					actualSecrets, err := secretGetter.Secrets("").List(context.Background(), metav1.ListOptions{})
```

cmd/ci-secret-bootstrap/main_test.go
droslean: Yea this was only to control the initial vault data. I didn't do it with `map[string]string` because I wanted to control also the secret data as well.

cmd/ci-secret-bootstrap/main_test.go
droslean: I tried to compare the error message, but vault returns some multi-line error messages and I was unable to compare them. I will check this out.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2633,3 +2634,621 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+					},
+				},
+			},
+			expectedSecrets: map[string][]coreapi.Secret{
+				"cluster-1": {
+					{
+						ObjectMeta: metav1.ObjectMeta{Name: "prod-secret-1", Namespace: "namespace-1", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap"}},
+						Data: map[string][]byte{
+							"key-name-1":    []byte("secret-data"),
+							"merge-with-me": []byte("old-secret-data"),
```
stevekuznetsov, line 3108: I'm lost - where does this come from?

cmd/ci-secret-bootstrap/main_test.go
stevekuznetsov: Hrm - I almost disagree here. If the type of error is important to the caller, the Vault functions should return errors that can be inspected with `errors.Is()`. Unwrapping an error to get at internal details like the message is an anti-pattern and opens the otherwise black-box approach. If we care specifically about the type, return typed errors and check with `errors.Is()`.

stevekuznetsov (commented):
Do we have a case where one Secret has keys both from user and bootstrap sources?

cmd/ci-secret-bootstrap/main_test.go
droslean: https://github.com/openshift/ci-tools/pull/1984/files#diff-5467c910200007c00123d9364d9d40c443678717319d06b6586374b03aaf8f18R3206

droslean: addressed.

cmd/ci-secret-bootstrap/main_test.go
alvaroaleman: First off you can simply convert between `map[string]string` and `map[string][]byte` and second off, the important point is that the map value needs to be a map of some sort and not a slice. Making it a slice with implicit map restrictions just makes everything more complicated and error-prone

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2658,3 +2659,635 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+					VaultPrefix: "secret",
+					VaultAddr:   "http://" + vaultAddr,
+					VaultToken:  testhelper.VaultTestingRootToken,
+				},
+			}
+
+			censor := secrets.NewDynamicCensor()
+			c, err := o.secrets.NewClient(&censor)
+			if err != nil {
+				t.Fatal("Failed to create client.")
```
alvaroaleman, line 3240: Log the error

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2658,3 +2659,635 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+
+			for item, objects := range tc.vaultObjects {
+				for _, object := range objects {
+					if err := c.SetFieldOnItem(item, object.field, object.data); err != nil {
+						t.Fatalf("couldn't populate vault item %s: %v", item, err)
+					}
+				}
+			}
+
+			if tc.initialData != nil {
```
alvaroaleman, line 3251: Sorry, but this doesn't make any sense. You already pass fakeclient arounds that can be prepopulated, there is absolutely no reason to have another `initialData` field.

Regardless of that as I explained in my earlier comments, please do _not_ add update tests in your existing `testCases` slice. Instead, iterate twice over it, once for the `Create` case and once for the `Update` case. For the `Update` case you can look at `tc.expectedSecrets` to know which secrets to pre-create. Do so, but manipluate their values to something arbitrary beforehand. I am happy to explain this in more detail over a call, but please do not duplicate all tests with a `Create` and an `Update` variation.

cmd/ci-secret-bootstrap/main_test.go
alvaroaleman: It is important to check that we got the error we expected and not any error. It doesn't really matter to me how exactly we do the checking (all of err message equality, err message substring, errors.Is/As is fine). 

We had plenty of tests in test-infra that succeeded because they did such a bool check but got an entirely different error because their test setup become invalid over time. Doing `expectErr: bool` style tests means they can very quickly and unnoticed rot under you and pass when they shouldn't.

cmd/ci-secret-bootstrap/main_test.go
droslean: Yea, my first thought was to create a map of a map but in order not to make things complicated I created the struct instead. I will make the changes. Thanks.

cmd/ci-secret-bootstrap/main_test.go
droslean: There is also the `merge` case, but I've got your point. I'll make the changes. Thanks!

cmd/ci-secret-bootstrap/main_test.go
droslean: I will try to check the error message, but it will be only a string comparison.

cmd/ci-secret-bootstrap/main_test.go
droslean: I was wrong about the `merge` case, so I removed the initial data and add them when the fake client is being created. I kept the update cases because it doesn't make sense to iterate again to the test cases to test the update case since we control that with the `--force` argument.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2752,3 +2753,498 @@ func (f *fakeVaultClient) ListKVRecursively(prefix string) ([]string, error) {
...
+					{
+						ObjectMeta: metav1.ObjectMeta{Name: "prod-secret-1", Namespace: "namespace-1", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap"}},
+						Data:       map[string][]byte{"key-name-1": []byte("secret-data")},
+						Type:       coreapi.SecretTypeOpaque,
+					},
+				},
+			},
+		},
+		{
+			id: "Successflly creating secret from config in multiple clusters",
```
alvaroaleman, line 2805:
```suggestion
			id: "Successfully creating secret from config in multiple clusters",
```

alvaroaleman (changes requested):
```
Regardless of that as I explained in my earlier comments, please do not add update tests in your existing testCases slice. Instead, iterate twice over it, once for the Create case and once for the Update case. For the Update case you can look at tc.expectedSecrets to know which secrets to pre-create. Do so, but manipluate their values to something arbitrary beforehand. I am happy to explain this in more detail over a call, but please do not duplicate all tests with a Create and an Update variation.
```

Ref https://github.com/openshift/ci-tools/pull/1984#discussion_r636939453

droslean:
@alvaroaleman Updated. I now check the update case in the same test case iteration, by changing the secrets that have been created in the Create case.

PR #1989, opened 2021-05-18
pkg/results: flatten error trees

bbguimaraes:
The `Aggregate` error type from `apimachinery` explicitly does not support
`Errors.As`:

```go
// Errors.As() is not supported, because the caller presumably cares
// about a specific error of potentially multiple that match the given
// type.
```

This breaks the causal chain in the `FullReason` method.  This change adds
a function that flattens trees of arbitrarily-nested
`results.Error` and `Aggregate`-like errors into a list while preserving
the hierarchy, so that all errors reported have the correct reason.

https://issues.redhat.com/browse/DPTP-2214

### DIFF
--- a/pkg/results/error.go
+++ b/pkg/results/error.go
@@ -52,6 +52,32 @@ func FullReason(err error) string {
 	return reasonedError.FullReason()
 }
 
+// Reasons provides the chains of error reasons.
+// Each item in the return value is a single chain divided by colons.  Aggregate
+// errors — those whose type provides an `Errors` method returning a list of
+// errors — are recursively expanded, generating a separate chain for each
+// child.
+func Reasons(errs ...error) (ret []string) {
+	for _, err := range errs {
+		switch err := err.(type) {
+		case *Error:
+			children := Reasons(err.Unwrap())
+			if len(children) == 0 {
+				ret = append(ret, string(err.reason))
+				break
+			}
+			for _, r := range children {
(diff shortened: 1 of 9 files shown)


bbguimaraes:
`codegen` and `validate-vendor` had registry failures, E2E was testing an error message from lease steps that I changed slightly.

pkg/results/error.go
```diff
@@ -33,23 +33,30 @@ func (e *Error) Is(target error) bool {
...
-func (e *Error) FullReason() string {
-	reasonedError := &Error{}
-	if !errors.As(e.wrapped, &reasonedError) {
-		return string(e.reason)
+// Reasons provides the chains of error reasons.
+// Each item in the return value is a single chain divided by colons.  Aggregate
+// errors — those whose type provides an `Errors` method returning a list of
+// errors — are recursively expanded, generating a separate chain for each
+// child.
+func Reasons(errs ...error) (ret []string) {
```
stevekuznetsov, line 41: Did we lose the unknown return>

stevekuznetsov (commented):
One comment, otherwise this looks really good!

pkg/results/error.go
bbguimaraes: My initial implementation kept it, but in the end I moved it [here](https://github.com/openshift/ci-tools/pull/1989/files#diff-d89ae970e1b23b3fe7065c1c228c08354612fe0ed0cc69db9dd5565dd16987b5R111) because the API made more sense to me and it simplified the recursive implementation of this function.

PR #1993, opened 2021-05-19
Bootstrap item validation: Correctly strip prefix

alvaroaleman:
We insert the VaultDPTPPrefix into all item names during deserialization
in the bootstrapper and give the generator the same prefix as CLI arg.

This means that the generator config has the items without that prefix
so we need to strip it before we check in the generator config if it
exists.

/cc @droslean @petr-muller 

Responsible for e.G. the incorrect failure here: https://github.com/openshift/release/pull/18643

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -833,7 +833,7 @@ func (o *options) validateBWItems(client secrets.ReadOnlyClient) error {
 					continue
 				}
 				if !hasItem {
-					if o.generatorConfig.IsItemGenerated(item.BWItem) {
+					if o.generatorConfig.IsItemGenerated(stripDPTPPrefixFromItem(item.BWItem, &o.config)) {
 						logrus.Warn("Item doesn't exist but it will be generated")
 					} else {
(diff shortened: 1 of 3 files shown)


droslean (approved):

cmd/ci-secret-bootstrap/main.go
```diff
@@ -833,7 +833,7 @@ func (o *options) validateBWItems(client secrets.ReadOnlyClient) error {
 					continue
 				}
 				if !hasItem {
-					if o.generatorConfig.IsItemGenerated(item.BWItem) {
+					if o.generatorConfig.IsItemGenerated(strings.TrimPrefix(item.BWItem, o.config.VaultDPTPPRefix+"/")) {
```
petr-muller, line 836: I'd consider giving the `strings.TrimPrefix(item.BWItem, o.config.VaultDPTPPRefix+"/")` a name and put it into a little helper so it does not need to be inline everywhere, for readability. But it's not a big deal.

alvaroaleman: Did that, mostly to have a location for an explanatory comment as this is pretty confusing. Also fixed the tests, PTAL

PR #1994, opened 2021-05-19
rehearse: compress CONFIG_SPEC

AlexNPavel:
This PR updates the `pj-rehearse` tool to compress and base64 encode the
`CONFIG_SPEC` env var in order to allow rehearsal of jobs have a
`CONFIG_SPEC` larger than ~128KB (max linux arglength).


pkg/rehearse/jobs_test.go
```diff
@@ -648,6 +665,16 @@ func TestExecuteJobsPositive(t *testing.T) {
...
+	compressedJob1, err := gzip.CompressString(testingCiOpCfgJob1YAML)
+	if err != nil {
+		t.Fatalf("Failed to compress config: %v", err)
+	}
+	job1Cfg := base64.StdEncoding.EncodeToString([]byte(compressedJob1))
+	compressedJob2, err := gzip.CompressString(testingCiOpCfgJob2YAML)
+	if err != nil {
+		t.Fatalf("Failed to compress config: %v", err)
+	}
+	job2Cfg := base64.StdEncoding.EncodeToString([]byte(compressedJob2))
```
alvaroaleman, line 677: Can we keep one uncompressed to make sure that continues to work?

AlexNPavel:
I've updated the PR to handle the `base64` encode as part of the compress string function and added a unit test loading compressed config from env.

pkg/rehearse/jobs_test.go
AlexNPavel: That variable is the expected config_spec output for that function, so it will always be compressed.

PR #1996, opened 2021-05-20
DPTP-2200: Support overriding a dependency with an external pull spec.

emilvberglind:
Allow a pull spec to be passed in to the ci-operator. This pull spec will override any dependency that matches the environment variable name. For example:

`--dependency-override-param=OO_INDEX=registry.mystuff.com:5000/pushed/myimage`

would override the `OO_INDEX` env param in any test step with the value `registry.mystuff.com:5000/pushed/myimage`. It is then up to the job to use that pull spec appropriately.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -407,6 +407,7 @@ type options struct {
 	hiveKubeconfig     *rest.Config
 
 	multiStageParamOverrides stringSlice
+	dependencyOverrides      stringSlice
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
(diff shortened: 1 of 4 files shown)


cmd/ci-operator/main.go
```diff
@@ -661,27 +663,42 @@ func (o *options) Complete() error {
 		return err
 	}
 
+	if err := overrideTestStepDependencyParams(o); err != nil {
```
stevekuznetsov, line 666: nit: 

```go
return overrideTestStepDependencyParams(o)
```

pkg/steps/multi_stage.go
```diff
@@ -192,10 +192,15 @@ func (s *multiStageTestStep) Requires() (ret []api.StepLink) {
...
 
 		for _, dependency := range step.Dependencies {
-			// we validate that the link will exist at config load time
-			// so we can safely ignore the case where !ok
-			imageStream, name, _ := s.config.DependencyParts(dependency)
-			ret = append(ret, api.LinkForImage(imageStream, name))
+			// if a fully-qualified pull spec was provided to the ci-operator for this dependency, then we don't need to
+			// create a step link since we won't do anything with this dependency other than passing the pull spec straight
+			// through in the environment variable.
+			if dependency.PullSpec == "" {
```
stevekuznetsov, line 198: nit: Go wants you to check the other case and `continue`

stevekuznetsov (commented):
Seems reasonable to me

cmd/ci-operator/main.go
```diff
@@ -701,6 +714,42 @@ func overrideMultiStageParams(o *options) error {
 	return nil
 }
 
+func overrideTestStepDependencyParams(o *options) error {
+	if len(o.dependencyOverrides.values) > 0 {
```
petr-muller, line 718: Consider using a guard clause and fast return, like in `overrideMultiStageParams`

petr-muller (approved):
LGTM with a nit, unhold or fix at will

PR #2001, opened 2021-05-24
[CVP-1467] add support for the kube_objects

jsztuka:
Adding new environment variable **PYXIS_URL** (optional), that will be used furthermore in optional_operator steps.

The cvp-trigger utility will need to be expanded to support sending the Pyxis partner project URL parameter to the Prow Job as an environment variable, and the OVP pipeline code will supply this parameter when calling the utility to trigger the Prow job.


petr-muller:
@jsztuka please pass tests.

How pressing is this work? We very recently landed some `ci-operator` improvements that will allow CVP to use standard multi-stage step parameters instead of the custom support in ci-op. I wanted to start reworking CVP steps to use that code this week, it would mean the ci-operator changes in this PR would be unnecessary and in general adding new paramaters would be simpler.

jsztuka:
> @jsztuka please pass tests.
> 
> How pressing is this work? We very recently landed some `ci-operator` improvements that will allow CVP to use standard multi-stage step parameters instead of the custom support in ci-op. I wanted to start reworking CVP steps to use that code this week, it would mean the ci-operator changes in this PR would be unnecessary and in general adding new paramaters would be simpler.

Hello Petr,
Looking into the task priority it is set to major. However it seems to me that this would need to be changed anyway, probably after your modifications. Can you provide the story/epic that corresponds with upcomming changes from your side so we can somehow watch the current state?

petr-muller:
@jsztuka It's DPTP-2248

cmd/cvp-trigger/README.md
```diff
@@ -70,6 +70,8 @@ CVP Trigger must load the configuration for the
 - `OO_CHANNEL`: Required. The name of the operator channel to track.
 - `OO_INSTALL_NAMESPACE`: Optional. The namespace into which the operator and
    catalog will be installed. If empty, a new namespace will be created.
+- `OO_PYXIS_URL`: Optional. Represents cvp product package name for specific ISV
```
nmars, line 73: Is the description accurate? The variable name has `URL` in it, but the description says it represents a package name

jsztuka: It is not currently. I need to update it. I thought we will be providing only the ISV_PID here, but we will provide whole URL together with the PID. Thanks for that catch!

petr-muller:
We just merged https://github.com/openshift/ci-tools/pull/2046 which changes the parameter passing interface between cvp-trigger and the job, using the standard mutli-stage parameters without ci-operator doing custom work. What this means for this PR:

- the `ci-operator` changes should be removed
- the cvp-trigger needs to pass this value via the new `appendMultiStageParams` method introduced in https://github.com/openshift/ci-tools/pull/2046
- the corresponding PR in o/release needs to be adjusted so that the new step receives the parameer as a multi-stage parameter and not the custom `OO_*` envvar.

Ideally, hold on with this for few days. We're in the intermediate period where both ci-operator and cvp-trigger temporarily support both ways to pass parameters so that we can rollout the cahges without breaking existing `cvp-trigger` usage. We're very close to using just the new one, then this PR (and the corresponding o/release one) should be simpler.

jsztuka:
If I try to rebase this current PR it registers 500+ changes so I would probably create a new PR and close this one.

PR #2004, opened 2021-05-25
Vault Secret Collection Manager Edit Members: Allow filtering

alvaroaleman:
This changes the secret collection manager to use a datalist for the all
members field instead of a multiselect. This allows ppl to either see
all entries as a suggestion or to just start typing which is useful in
our production case where there are lots of users.
datalist-0

datalist1
dl3

Ref https://issues.redhat.com/browse/DPTP-2192

### DIFF
--- a/cmd/vault-secret-collection-manager/index.template.html
+++ b/cmd/vault-secret-collection-manager/index.template.html
@@ -45,7 +45,10 @@
           <tbody>
             <tr class="user-edit-table">
               <td class="user-edit-table"><select name="CurrentMembers" id="currentMembersSelection" multiple></select></td>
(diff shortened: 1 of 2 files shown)


petr-muller (approved):

stevekuznetsov (commented):
Technically the user can still type in an option that's not actually one of the items in the list - do we want to validate that client-side?

alvaroaleman:
> Technically the user can still type in an option that's not actually one of the items in the list - do we want to validate that client-side?

If they do that it will not appear in the `Current Member` list because `moveSelectedOptions` iterates over all members and just skips the ones that are not selected. If the user is invalid, it will just end up doing nothing.

PR #2006, opened 2021-05-26
CI-Operator: Create imports for root image if FromRepository is enabled

alvaroaleman:
This change makes the CI-Operator create testimagestreamtag imports for
the root image if the FromRepository setting is set. This is best-efford
only and will log warnings on errors but not fail.

It is also insufficient by itself, we need to deploy the CRD, add RBAC
and extend the test-image-distribution-controller to watch
testimagestreamtagimports in all clusters.

Ref https://issues.redhat.com/browse/DPTP-1640

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -8,12 +8,15 @@ import (
 	"io/ioutil"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/wait"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
 	"k8s.io/client-go/rest"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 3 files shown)


pkg/defaults/defaults.go
```diff
@@ -803,3 +812,34 @@ func buildRootImageStreamFromRepository(readFile readFile) (*api.ImageStreamTagR
...
+		},
+	}
+	istImport.SetImageStreamLabels()
+
+	// Conflicts are expected
+	if err := client.Create(ctx, istImport); err != nil && !kapierrors.IsConflict(err) {
+		logrus.WithError(err).Warnf("Failed to create imagestreamtagimport for root %s", istImport.Name)
+	}
+
+	if err := wait.PollImmediate(5*second, 30*second, func() (bool, error) {
```
emilvberglind, line 834: If this is always going to use a duration of `second`, do we need to pass that in as a parameter?

alvaroaleman: It doesn't, the tests pass `time.Nanosecond` as "second" into `stepConfigsForBuild`, otherwise we'd have to wait number of unittests * unittestsrepetitions (25 in CI, to avoid flakes) * 30 seconds for the tests, which makes CI timeout

emilvberglind: Ah. I missed that the test passed in something different, sorry!

emilvberglind:
This all seems legit to me, but I don't think I'm qualified to leave an actual lgtm label yet.

pkg/defaults/defaults.go
```diff
@@ -803,3 +812,34 @@ func buildRootImageStreamFromRepository(readFile readFile) (*api.ImageStreamTagR
...
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKeyFromObject(istImport), &imagev1.ImageStreamTag{}); err != nil {
```
stevekuznetsov, line 835: Why is this not a watch?

alvaroaleman: Because the imagestreamtag api does not support the watch verb

alvaroaleman:
```
Error from server (MethodNotAllowed): the server does not allow this method on the requested resource (get imagestreamtags.image.openshift.io)
```

stevekuznetsov: Oh, right ...

stevekuznetsov (approved):

alvaroaleman:
The contents of this PR were merged in https://github.com/openshift/ci-tools/pull/2007

PR #2007, opened 2021-05-26
Test Image distributor: Consider TestImageImports from all clusters

alvaroaleman:
Based on top of https://github.com/openshift/ci-tools/pull/2006, please review that one first and only the second commit here.

Makes the TestImageDistributor react to testImageStreamImports in all clusters rather than just the ones in app.ci

Ref https://issues.redhat.com/browse/DPTP-1640

### DIFF
--- a/pkg/api/testimagestreamtagimport/v1/types.go
+++ b/pkg/api/testimagestreamtagimport/v1/types.go
@@ -21,7 +21,7 @@ type TestImageStreamTagImport struct {
 // creating multiple objects for the same import.
 func (t *TestImageStreamTagImport) SetDeterministicName() {
 	t.Name = fmt.Sprintf("%s-%s-%s", t.Spec.ClusterName, t.Spec.Namespace, strings.ReplaceAll(t.Spec.Name, ":", "."))
-	t.SetImageStreamLabels()
+	t.WithImageStreamLabels()
 }
 
 const (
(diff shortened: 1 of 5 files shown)


pkg/api/testimagestreamtagimport/v1/types.go
```diff
@@ -31,12 +31,13 @@ const (
 
 // SetImageStreamLabels sets namespace and name labels so we can easily
 // filter for specific imports
-func (t *TestImageStreamTagImport) SetImageStreamLabels() {
+func (t *TestImageStreamTagImport) SetImageStreamLabels() *TestImageStreamTagImport {
```
stevekuznetsov, line 34: While I see what you did in the test code with this, a setter that returns a reference to the object itself is definitely surprising code ...

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -74,6 +74,19 @@ func AddToManager(mgr manager.Manager,
 		}
 		buildClusters.Insert(buildClusterName)
 		r.buildClusterClients[buildClusterName] = imagestreamtagwrapper.MustNew(buildClusterManager.GetClient(), buildClusterManager.GetCache())
+
+		if buildClusterName == "app.ci" {
```
stevekuznetsov, line 78: Do we not have a constant for this?

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -108,6 +121,20 @@ func AddToManager(mgr manager.Manager,
 	return nil
 }
 
+func testImageStreamTagImportHandlerForNamedCluster(clusterName string) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+		testimagestreamtagimport, ok := o.(*testimagestreamtagimportv1.TestImageStreamTagImport)
+		if !ok {
+			logrus.WithField("type", fmt.Sprintf("%T", o)).Error("Got object that was not an ImageStram")
```
stevekuznetsov, line 128: `s/ImageStram/ImageStream/`

pkg/api/testimagestreamtagimport/v1/types.go
alvaroaleman: I've renamed it to `WithImageStreamLabels` to clarify the semantics, WDYT?

alvaroaleman: FWIW the `sets` from apimachinery also return themselves in their setter: https://github.com/openshift/ci-tools/blob/fa75e5c8447f70771bdf3346b6cfc0bec9875218/vendor/k8s.io/apimachinery/pkg/util/sets/string.go#L49-L54

stevekuznetsov: Wow, I had no idea.

PR #2008, opened 2021-05-27
Make the breaking-changes job to respect config shrading

droslean:


petr-muller (approved):
/hold

Holding so we can see the job's result

PR #2011, opened 2021-05-27
Remove/rename all bitwarden references

droslean:

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -44,9 +44,9 @@ import (
 type options struct {
 	secrets secrets.CLIOptions
 
-	dryRun               bool
-	force                bool
-	validateBWItemsUsage bool
+	dryRun             bool
+	force              bool
+	validateItemsUsage bool
 
 	kubeConfigPath      string
 	configPath          string
@@ -59,7 +59,7 @@ type options struct {
 	config          secretbootstrap.Config
 	generatorConfig secretgenerator.Config
 
-	bwAllowUnused flagutil.Strings
+	allowUnused flagutil.Strings
 
 	validateOnly bool
 }
@@ -73,10 +73,10 @@ const (
 func parseOptions(censor *secrets.DynamicCensor) (options, error) {
 	var o options
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	o.bwAllowUnused = flagutil.NewStrings()
+	o.allowUnused = flagutil.NewStrings()
 	fs.BoolVar(&o.validateOnly, "validate-only", false, "If set, the tool exists after validating its config file.")
-	fs.Var(&o.bwAllowUnused, "bw-allow-unused", "One or more bitwarden items that will be ignored when the --validate-bitwarden-items-usage is specified")
-	fs.BoolVar(&o.validateBWItemsUsage, "validate-bitwarden-items-usage", false, fmt.Sprintf("If set, the tool only validates if all attachments and custom fields that exist in BitWarden and were last modified before %d days ago are being used in the given config.", allowUnusedDays))
+	fs.Var(&o.allowUnused, "bw-allow-unused", "One or more items that will be ignored when the --validate-items-usage is specified")
+	fs.BoolVar(&o.validateItemsUsage, "validate-bitwarden-items-usage", false, fmt.Sprintf("If set, the tool only validates if all fields that exist in Vault and were last modified before %d days ago are being used in the given config.", allowUnusedDays))
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
 	fs.StringVar(&o.kubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
 	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
@@ -103,8 +103,8 @@ func (o *options) validateOptions() error {
 	if o.configPath == "" {
 		errs = append(errs, errors.New("--config is required"))
 	}
-	if len(o.bwAllowUnused.Strings()) > 0 && !o.validateBWItemsUsage {
-		errs = append(errs, errors.New("--bw-allow-unused must be specified with --validate-bitwarden-items-usage"))
+	if len(o.allowUnused.Strings()) > 0 && !o.validateItemsUsage {
+		errs = append(errs, errors.New("--bw-allow-unused must be specified with --validate-items-usage"))
 	}
 	return utilerrors.NewAggregate(errs)
 }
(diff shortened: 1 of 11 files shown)


droslean:
`secret-bootstrapper-validation` job should fail.

droslean:
/hold

A follow-up is required to change the `bootstrap` and `generator` config in the release repo.

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -18,23 +18,20 @@ const (
 	AttributeTypePassword AttributeType = "password"
 )
 
-type BitWardenContext struct {
-	BWItem               string                 `json:"bw_item"`
+type ItemContext struct {
+	Item                 string                 `json:"item"`
```
alvaroaleman, line 22: Can we maybe just do a first pr that adds new fields with new names, then change the config, then remove the old fields?

droslean: I was hoping that I can do a follow-up, after the changes in this PR merge, in the release repo changing the config. It won't cost anything. Worst case scenario we will have a failed periodic job.

droslean: I can disable the periodic jobs, land the changes in this PR and in release repo, then enable the jobs again. WDYT?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -77,10 +77,10 @@ const (
 func parseOptions(censor *secrets.DynamicCensor) (options, error) {
 	var o options
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	o.bwAllowUnused = flagutil.NewStrings()
+	o.allowUnused = flagutil.NewStrings()
 	fs.BoolVar(&o.validateOnly, "validate-only", false, "If set, the tool exists after validating its config file.")
-	fs.Var(&o.bwAllowUnused, "bw-allow-unused", "One or more bitwarden items that will be ignored when the --validate-bitwarden-items-usage is specified")
-	fs.BoolVar(&o.validateBWItemsUsage, "validate-bitwarden-items-usage", false, fmt.Sprintf("If set, the tool only validates if all attachments and custom fields that exist in BitWarden and were last modified before %d days ago are being used in the given config.", allowUnusedDays))
+	fs.Var(&o.allowUnused, "bw-allow-unused", "One or more bitwarden items that will be ignored when the --validate-bitwarden-items-usage is specified")
```
alvaroaleman, line 82: Maybe remove the mentioning of `bitwarden` here and below as well?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -277,30 +262,21 @@ func constructDockerConfigJSON(client secrets.ReadOnlyClient, dockerConfigJSONDa
 	for _, data := range dockerConfigJSONData {
 		authData := secretbootstrap.DockerAuth{}
 
-		registryURL := data.RegistryURL
-		if registryURL == "" {
```
alvaroaleman, line 281: Why is the registry url resolving logic being removed?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -804,69 +725,49 @@ func getUnusedBWItems(config secretbootstrap.Config, client secrets.ReadOnlyClie
...
 					}
 					if !hasItem {
-						errs = append(errs, fmt.Errorf("item %s doesn't exist", data.BWItem))
+						errs = append(errs, fmt.Errorf("item %s doesn't exist", data.Item))
 						break
 					}
-					if _, err := client.GetAttachmentOnItem(data.BWItem, data.AuthBitwardenAttachment); err != nil {
-						errs = append(errs, fmt.Errorf("attachment %s in item %s doesn't exist", data.AuthBitwardenAttachment, data.BWItem))
+					if _, err := client.GetFieldOnItem(data.Item, data.AuthField); err != nil {
+						errs = append(errs, fmt.Errorf("attachment %s in item %s doesn't exist", data.AuthField, data.Item))
```
alvaroaleman, line 747:
```suggestion
						errs = append(errs, fmt.Errorf("field %s in item %s doesn't exist", data.AuthField, data.Item))
```

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -791,91 +754,20 @@ func TestValidateCompletedOptions(t *testing.T) {
 				config: secretbootstrap.Config{
 					Secrets: []secretbootstrap.SecretConfig{
 						{
-							From: map[string]secretbootstrap.BitWardenContext{
```
alvaroaleman, line 794: Why was this test removed?

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -791,91 +754,20 @@ func TestValidateCompletedOptions(t *testing.T) {
...
-									Type:      "kubernetes.io/dockerconfigjson",
-								},
-							},
-						},
-					},
-				},
-			},
-		},
-		{
-			name: "sad dockerconfigJSON configuration: cannot set both RegistryURL and RegistryURLBitwardenField",
```
alvaroaleman, line 864: Why was this test removed?

pkg/api/secretbootstrap/secretboostrap.go
alvaroaleman: Not a fan of that, we also have presubmits that use this config and this is a gigantic PR that has a lot of potential to contain bugs, so I do not want to require it to be merged/reverted in lockstep with a config change. Please just leave the `json` annotation as it was, we can change it together with the config in a follow-up after this had some time to sink and we are confident it works correctly.

Also before this can merge, all usage of attachments and passwords in our config needs to be removed.

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -18,23 +18,20 @@ const (
...
 type DockerConfigJSONData struct {
-	BWItem                    string `json:"bw_item"`
-	RegistryURL               string `json:"registry_url"`
-	RegistryURLBitwardenField string `json:"registry_url_bw_field"`
-	AuthBitwardenAttachment   string `json:"auth_bw_attachment"`
-	EmailBitwardenField       string `json:"email_bw_field,omitempty"`
+	Item        string `json:"item"`
+	RegistryURL string `json:"registry_url"`
+	AuthField   string `json:"auth_field"`
+	EmailField  string `json:"email_field,omitempty"`
```
alvaroaleman, line 34: Same comment here regarding not changing the json annotations.

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -18,23 +18,20 @@ const (
...
 	DockerConfigJSONData []DockerConfigJSONData `json:"dockerconfigJSON,omitempty"`
 	// If the secret should be base64 decoded before uploading to kube. Encoding
 	// it is useful to be able to store binary data.
 	Base64Decode bool `json:"base64_decode,omitempty"`
 }
 
 type DockerConfigJSONData struct {
-	BWItem                    string `json:"bw_item"`
-	RegistryURL               string `json:"registry_url"`
-	RegistryURLBitwardenField string `json:"registry_url_bw_field"`
```
alvaroaleman, line 35: Why was this removed, do we not use it?

pkg/api/secretbootstrap/secretboostrap.go
droslean: Yea you are right. It's safer to do it like this. I will update the PR.

cmd/ci-secret-bootstrap/main.go
droslean: First we had only the registry-url field that it was supposed to be a field in bitwarden, but we added the `registryURL` that gets the value directly from the config. We kept the old one to keep backwards compatibility and now it is not being used at all. So it is safe to remove it as well.

cmd/ci-secret-bootstrap/main_test.go
droslean: `RegistryURLBitwardenField` is removed, so there is no need for that validation at all.

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -18,7 +18,8 @@ const (
 	AttributeTypePassword AttributeType = "password"
 )
 
-type BitWardenContext struct {
+type ItemContext struct {
+	Item                 string                 `json:"item"`
```
alvaroaleman, line 22: Why did we add a new field? Please just rename the existing field without changing the json annotation

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -30,11 +31,13 @@ type BitWardenContext struct {
 }
 
 type DockerConfigJSONData struct {
-	BWItem                    string `json:"bw_item"`
-	RegistryURL               string `json:"registry_url"`
-	RegistryURLBitwardenField string `json:"registry_url_bw_field"`
-	AuthBitwardenAttachment   string `json:"auth_bw_attachment"`
-	EmailBitwardenField       string `json:"email_bw_field,omitempty"`
+	Item                    string `json:"item"`
```
alvaroaleman, line 34: Please leave the json tag as it was

pkg/api/secretbootstrap/secretboostrap.go
droslean: I wanted to create a new json but still support the old one to keep backwards compatibility, and change the config without breaking anything.

pkg/api/secretgenerator/testdata/zz_fixture_TestLoadConfigFromPath_single_parameter_with_multiple_values.yaml
```diff
@@ -1,7 +1,4 @@
-- attachments:
-  - cmd: echo -n Attachment1
-    name: Attachment1
-  fields:
+- fields:
```
alvaroaleman, line 1: I think I would feel more comfortable if we didn't change any existing fixtures to proof that the current as we have it right now will continue to work (only add some more to show that the new config will also work)

pkg/api/secretbootstrap/secretboostrap.go
alvaroaleman: Ah sorry, I missed that, the fact that you changed all existing fixtures and tests to use the new format confused me. Can we not do that to:
* Reduce the size of this PR
* Proof that the old format our config is in still works?

droslean: Well, eventually we will end up changing them, so in this PR I only change their names and load them from the old values during the unmarshaling. The old version should be working fine.  I can try to see what it takes to make this PR smaller and better reviweablt.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -106,22 +106,22 @@ const (
 secret_configs:
 - from:
     key-name-1:
-      bw_item: item-name-1
+      item: item-name-1
```
alvaroaleman, line 109: Can we not change anything here to show that the existing config works as before?

droslean: Let me add both cases then.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -106,23 +106,29 @@ const (
...
```
alvaroaleman, line 109: Why is this conversation resolved? All the tests got changed

droslean: I added below an item using the old fields as well. That should be enough, right?

droslean:
@alvaroaleman Is there anything left here? Sorry I didn't focus on this PR lately.

cmd/ci-secret-bootstrap/main_test.go
alvaroaleman: Please just leave the existing tests unmodified and add new ones for the new functionality.

droslean: Got it. I'll make the changes. Thanks.

alvaroaleman (approved):

PR #2016, opened 2021-06-01
rehearse periodics affected by changed ci-op configs

petr-muller:
Resolves: [DPTP-1095](https://issues.redhat.com/browse/DPTP-1095)

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -289,8 +289,9 @@ func rehearseMain() error {
 	changedPresubmits := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
 	toRehearse.AddAll(changedPresubmits, config.ChangedPresubmit)
 
-	presubmitsForCiopConfigs := diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigData, affectedJobs, logger)
+	presubmitsForCiopConfigs, periodicsForCiopConfigs := diffs.GetJobsForCiopConfigs(prConfig.Prow, changedCiopConfigData, affectedJobs, logger)
 	toRehearse.AddAll(presubmitsForCiopConfigs, config.ChangedCiopConfig)
+	periodicsToRehearse.AddAll(periodicsForCiopConfigs)
 
 	presubmitsForClusterProfiles := diffs.GetPresubmitsForClusterProfiles(prConfig.Prow, rehearsalClusterProfiles.ProductionNames, logger)
 	toRehearse.AddAll(presubmitsForClusterProfiles, config.ChangedClusterProfile)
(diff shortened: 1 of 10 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -289,8 +289,9 @@ func rehearseMain() error {
 	changedPresubmits := diffs.GetChangedPresubmits(masterConfig.Prow, prConfig.Prow, logger)
 	toRehearse.AddAll(changedPresubmits, config.ChangedPresubmit)
 
-	presubmitsForCiopConfigs := diffs.GetPresubmitsForCiopConfigs(prConfig.Prow, changedCiopConfigData, affectedJobs, logger)
+	presubmitsForCiopConfigs, periodicsForCiopConfigs := diffs.GetJobsForCiopConfigs(prConfig.Prow, changedCiopConfigData, affectedJobs, logger)
 	toRehearse.AddAll(presubmitsForCiopConfigs, config.ChangedCiopConfig)
+	periodicsToRehearse.AddAll(periodicsForCiopConfigs)
```
stevekuznetsov, line 294: Why does this not take the `config.ChangedCiopConfig` reason like the presubmits do?

pkg/diffs/diffs.go
```diff
@@ -161,33 +161,58 @@ func GetImagesPostsubmitsForCiopConfigs(prowConfig *prowconfig.Config, ciopConfi
...
-			if !strings.HasPrefix(job.Name, jobNamePrefix) {
-				continue
-			}
-			testName := strings.TrimPrefix(job.Name, jobNamePrefix)
 
-			affectedJob, ok := affectedJobs[data.Info.Basename()]
-			if ok && !affectedJob.Has(testName) {
+			selectionFields := logrus.Fields{LogRepo: orgRepo, LogJobName: job.Name, LogReasons: "ci-operator config changed"}
+			logger.WithFields(selectionFields).Info(ChosenJob)
+			if presubmits == nil {
```
stevekuznetsov, line 195: Why bother initializing them to `nil` if you're just going to do this?

stevekuznetsov (commented):
Two small comments, otherwise LGTM

pkg/diffs/diffs.go
petr-muller: We may skip all jobs and then we return nil. But I guess it's simpler to initialize to an empty map and return it instead of nil, easier code :+1:

cmd/pj-rehearse/main.go
petr-muller: That's... a good question. I did not touch *that* code and just did what other sources of rehearsed periodics do. Let me look into this.

pkg/diffs/diffs.go
stevekuznetsov: I don't think that I would personally expect to personally distinguish between a map that's `nil` versus one that has no elements ...

petr-muller: It's just annoying in tests where you cannot leave the expected "empty" item initialized ;)

petr-muller:
@stevekuznetsov addressed the feeback, simplified some code, made `periodics.AddAll` consistent with `presubmits.AddAll`. Noticed `periodics.AddAll` is not tested, so added tests.

stevekuznetsov (approved):

PR #2017, opened 2021-06-01
DPTP-2243: provide CLI injection for cluster claim tests

AlexNPavel:
This PR adds a new field to multistage test steps to allow injecting the
CLI for tests using cluster claims. It also adds extra config validation
and unit tests for both the new and existing CLI injection support.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -768,6 +768,9 @@ type LiteralTestStep struct {
 	// Cli is the (optional) name of the release from which the `oc` binary
 	// will be injected into this step.
 	Cli string `json:"cli,omitempty"`
+	// ClusterClaimCLI defines if this step should inject the `oc` binary for
+	// the cluster claimed by this step. Requires test to define a cluster claim.
+	ClusterClaimCLI *bool `json:"cluster_claim_cli,omitempty"`
 	// Observers are the observers that should be running
 	Observers []string `json:"observers,omitempty"`
 	// RunAsScript defines if this step should be executed as a script mounted
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -332,7 +332,7 @@ func stepForTest(
 ) ([]api.Step, error) {
 	if test := c.MultiStageTestConfigurationLiteral; test != nil {
 		leases := leasesForTest(test)
-		if len(leases) != 0 {
+		if len(leases) != 0 || c.ClusterClaim != nil {
 			params = api.NewDeferredParameters(params)
 		}
 		step := steps.MultiStageTestStep(*c, config, params, podClient, jobSpec, leases)
@@ -342,6 +342,7 @@ func stepForTest(
 		}
 		if c.ClusterClaim != nil {
 			step = steps.ClusterClaimStep(c.As, c.ClusterClaim, hiveClient, client, jobSpec, step)
+			addProvidesForStep(step, params)
 		}
 		newSteps := stepsForStepImages(client, jobSpec, inputImages, test, imageConfigs)
 		return append([]api.Step{step}, newSteps...), nil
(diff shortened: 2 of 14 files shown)


pkg/api/types.go
```diff
@@ -768,6 +768,9 @@ type LiteralTestStep struct {
 	// Cli is the (optional) name of the release from which the `oc` binary
 	// will be injected into this step.
 	Cli string `json:"cli,omitempty"`
+	// ClusterClaimCLI defines if this step should inject the `oc` binary for
+	// the cluster claimed by this step. Requires test to define a cluster claim.
+	ClusterClaimCLI *bool `json:"cluster_claim_cli,omitempty"`
```
petr-muller, line 773: This sounds really unfortunate to me. This would require step authors to decide whether the step is usable on either claimed clusters or installed clusters, but never both, no?

Could we infer `true` here internally from nonempty `Cli`?

stevekuznetsov: Maybe infer that `cli: latest` would allow for CLI from the claim as well? I think we *do* need to make authors choose if they are using some weird custom CLI, since that will be doing something different ....

pkg/defaults/defaults.go
```diff
@@ -332,7 +332,7 @@ func stepForTest(
 ) ([]api.Step, error) {
 	if test := c.MultiStageTestConfigurationLiteral; test != nil {
 		leases := leasesForTest(test)
-		if len(leases) != 0 {
+		if len(leases) != 0 || c.ClusterClaim != nil {
 			params = api.NewDeferredParameters(params)
```
stevekuznetsov, line 336: Why is this conditional on anything?

pkg/steps/cluster_claim.go
```diff
@@ -50,7 +53,16 @@ func (s *clusterClaimStep) Description() string                 { return s.wrapp
 func (s *clusterClaimStep) Requires() []api.StepLink            { return s.wrapped.Requires() }
 func (s *clusterClaimStep) Creates() []api.StepLink             { return s.wrapped.Creates() }
 func (s *clusterClaimStep) Objects() []ctrlruntimeclient.Object { return s.wrapped.Objects() }
-func (s *clusterClaimStep) Provides() api.ParameterMap          { return s.wrapped.Provides() }
+func (s *clusterClaimStep) Provides() api.ParameterMap {
+	parameters := s.wrapped.Provides()
+	if parameters == nil {
+		parameters = api.ParameterMap{}
+	}
+	parameters[ClusterClaimCLIParam] = func() (string, error) { //nolint:unparam
```
stevekuznetsov, line 61: What is `unparam`?

pkg/steps/multi_stage.go
```diff
@@ -758,17 +762,23 @@ func addCommandScript(name string, pod *coreapi.Pod) {
 	})
 }
 
-func addCliInjector(release string, pod *coreapi.Pod) {
+func addCliInjector(release, pullSpec string, pod *coreapi.Pod) {
```
stevekuznetsov, line 765: Why not just always pass the full `pullSpec` - that's the only thing it looks like the `release` gets used for. Passing in two options to a method and having it always choose which to use is a code smell

stevekuznetsov (commented):
Implementation here looks good but the UX is not ideal. Perhaps we could implicitly import the release payload as `latest` when using a claim? Unclear what is best.

pkg/steps/cluster_claim.go
AlexNPavel: At least on my end, the linter was complaining that we never return anything except for `nil` for `error` in this function and was saying we should remove it. But that would break the code because parameters needs the `(string, error)` return type. So I'm just telling the linter to not complain about it.

AlexNPavel:
/close

This issue will instead be resolved by giving users the ability to import the release that a cluster claim was built from. Being tracked by https://issues.redhat.com/browse/DPTP-2262

PR #2026, opened 2021-06-07
CI-Operator-Yaml-Creator: Set build_root.from_repository if config is current

alvaroaleman:
This change makes the CI-Operator-Yaml-Creator also update the release
repo config if the .ci-operator.yaml is already current. I initially
planned to write a second tool for this but realized it is much easier
to just add it here.

/assign @jupierce 

Is this intended to be run as an one-off btw or should it run periodically?


jupierce:
@alvaroaleman - it seems like something to run periodically (at least until the migration PRs have a chance to successfully merge).

jupierce (approved):

PR #2034, opened 2021-06-09
DPTP-2220: Workaround to retrieve reporter from JIRA.

emilvberglind:
DPTP-2220

The JIRA API doesn't work as documented and this is causing a 400 when attempting to find the reporter. This is a workaround for that.


pkg/jira/issues.go
```diff
@@ -38,7 +39,14 @@ type jiraAdapter struct {
...
 func (a *jiraAdapter) FindUser(property string) ([]jira.User, *jira.Response, error) {
-	return a.delegate.User.Find(property) // ignore tweaks
+	// the JIRA API does not work as documented, and we can therefore not use the jira.Client.User.Find function here.
+	// This supplies a query parameter to the search endpoint which will result in a 400 as you are required to pass the
+	// username parameter which strangely behaves as the query parameter should.
+	req, _ := a.delegate.NewRequest("GET", fmt.Sprintf("rest/api/2/user/search?username=%s", url.QueryEscape(property)), nil)
+	users := new([]jira.User)
+	response, err := a.delegate.Do(req, users)
+
+	return *users, response, err
```
alvaroaleman, line 49: Nit: Imho easier to read as:
```suggestion
	var users []jira.User
	response, err := a.delegate.Do(req, &users)

	return users, response, err
```

pkg/jira/issues.go
```diff
@@ -38,7 +39,14 @@ type jiraAdapter struct {
 }
 
 func (a *jiraAdapter) FindUser(property string) ([]jira.User, *jira.Response, error) {
-	return a.delegate.User.Find(property) // ignore tweaks
+	// the JIRA API does not work as documented, and we can therefore not use the jira.Client.User.Find function here.
+	// This supplies a query parameter to the search endpoint which will result in a 400 as you are required to pass the
```
alvaroaleman, line 43: This is slightly confusing to read :P By the `This` you mean `a.delegate.User.Find(`?

emilvberglind: Yeah, I can clear that up a bit.

stevekuznetsov (approved):
:facepalm:

PR #2038, opened 2021-06-09
DPTP-2244: Registry-replacer should also prune unused base images.

emilvberglind:
With [DPTP-2244](https://issues.redhat.com/browse/DPTP-2244) we want to expand the registry-replacer so it'll also prune unused base images. In order to do this we need to resolve the full config so we can verify if a specific base image is used in any of the tests

### DIFF
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -33,6 +33,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/api/ocpbuilddata"
 	"github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/github"
+	"github.com/openshift/ci-tools/pkg/load/agents"
 	"github.com/openshift/ci-tools/pkg/steps/release"
 )
 
@@ -48,6 +49,7 @@ type options struct {
 	pruneUnusedReplacements                      bool
 	pruneOCPBuilderReplacements                  bool
 	ensureCorrectPromotionDockerfileIngoredRepos *flagutil.Strings
+	registryPath                                 string
 	flagutil.GitHubOptions
 }
 
@@ -65,6 +67,7 @@ func gatherOptions() (*options, error) {
 	flag.StringVar(&o.currentRelease.Minor, "current-release-minor", "6", "The minor version of the current release that is getting forwarded to from the master branch")
 	flag.BoolVar(&o.pruneUnusedReplacements, "prune-unused-replacements", false, "If replacements that match nothing should get pruned from the config")
 	flag.BoolVar(&o.pruneOCPBuilderReplacements, "prune-ocp-builder-replacements", false, "If all replacements that target the ocp/builder imagestream should be removed")
+	flag.StringVar(&o.registryPath, "registry", "", "Path to the step registry directory")
 	flag.Parse()
 
 	var errs []error
@@ -133,6 +136,11 @@ func main() {
 		}
 	}
 
+	agent, err := agents.NewRegistryAgent(opts.registryPath)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to create RegistryAgent")
+	}
+
 	var errs []error
 	errLock := &sync.Mutex{}
 	sem := semaphore.NewWeighted(int64(opts.maxConcurrency))
@@ -157,6 +165,7 @@ func main() {
 					promotionTargetToDockerfileMapping,
 					opts.currentRelease,
 					credentials,
+					agent.ResolveConfig,
 				)(config, info); err != nil {
 					errLock.Lock()
 					errs = append(errs, err)
@@ -200,8 +209,8 @@ func replacer(
 	ensureCorrectPromotionDockerfile bool,
 	ensureCorrectPromotionDockerfileIgnoredrepos sets.String,
 	promotionTargetToDockerfileMapping map[string]dockerfileLocation,
-	majorMinor ocpbuilddata.MajorMinor,
-	credentials *usernameToken,
+	majorMinor ocpbuilddata.MajorMinor, credentials *usernameToken,
+	configResolver func(config api.ReleaseBuildConfiguration) (api.ReleaseBuildConfiguration, error),
 ) func(*api.ReleaseBuildConfiguration, *config.Info) error {
 	return func(config *api.ReleaseBuildConfiguration, info *config.Info) error {
 		if len(config.Images) == 0 {
@@ -277,6 +286,18 @@ func replacer(
 		if pruneUnusedReplacementsEnabled && hasNonEmptyDockerfile {
 			if err := pruneUnusedReplacements(config, allReplacementCandidates); err != nil {
 				return fmt.Errorf("failed to prune unused replacements: %w", err)
+			}
+
+			// to prune base images we'll need the fully-resolved config.
+			resolvedConfig, err := configResolver(*config)
+			if err != nil {
+				// TODO this should stop execution rather than just logging an error, but since we want a smooth transition with
+				// the new registry-replacer enhancements we'll add that in later, after this has been rolled out.
+				logrus.WithError(err).Error("failed to resolve config. This means that base image pruning will not function as expected and will therefore be skipped")
+			} else {
+				if err := pruneUnusedBaseImages(config, &resolvedConfig); err != nil {
+					return fmt.Errorf("failed to prune unused base images: %w", err)
+				}
 			}
 		} else if pruneUnusedReplacementsEnabled {
 			logrus.WithField("org", info.Org).WithField("repo", info.Repo).WithField("branch", info.Branch).Info("Not purging unused replacements because we got an empty dockerfile")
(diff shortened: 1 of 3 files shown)


cmd/registry-replacer/main.go
```diff
@@ -278,6 +304,15 @@ func replacer(
 			if err := pruneUnusedReplacements(config, allReplacementCandidates); err != nil {
 				return fmt.Errorf("failed to prune unused replacements: %w", err)
 			}
+
+			// to prune base images we'll need the fully-resolved config.
+			resolvedConfig, err := configResolver()
+
+			if err != nil {
```
alvaroaleman, line 311: I think it would be good to return the error, otherwise we won't notice if this starts failing

cmd/registry-replacer/main.go
```diff
@@ -600,6 +635,88 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+var pipelineDependencyRegex = regexp.MustCompile(fmt.Sprintf(`(%s:)([\w.\-]+)`, api.PipelineImageStream))
+
+func getTestImages(images *sets.String, steps []api.LiteralTestStep) {
+	for _, step := range steps {
+		if step.From != "" {
+			images.Insert(step.From)
+		}
+
+		// see if we have any pipeline dependency images that also need to be included.
+		if len(step.Dependencies) == 0 {
```
alvaroaleman, line 703: Nit: looping over a potentially empty list is fine and there isn't really a need for the extra condition here

cmd/registry-replacer/main.go
```diff
@@ -278,6 +304,15 @@ func replacer(
 			if err := pruneUnusedReplacements(config, allReplacementCandidates); err != nil {
 				return fmt.Errorf("failed to prune unused replacements: %w", err)
 			}
+
+			// to prune base images we'll need the fully-resolved config.
+			resolvedConfig, err := configResolver()
+
```
alvaroaleman, line 310: Nit: The empty line between the call and the error handling bothers me, the two are usually directly after each other as they belong together :P

cmd/registry-replacer/main.go
```diff
@@ -189,6 +202,18 @@ type usernameToken struct {
 	token    string
 }
 
+func configResolver(info *config.Info, resolverAddress, registryPath string) (*api.ReleaseBuildConfiguration, error) {
```
alvaroaleman, line 205: This means that we have to do a http call for every single config. I would prefer it if we instead just load the config from the filesystem (we operate on o/release anyways), you can use the RegistryAgent constructed through pkg/load/agents.NewRegistryAgent for that and then use its `ResolveConfig`

cmd/registry-replacer/main.go
```diff
@@ -600,6 +635,88 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
 	return utilerrors.NewAggregate(errs)
 }
 
+// when pruning unused base images we'll check the fully-resolved config to make sure an image is not used directly in
```
alvaroaleman, line 638: Nit: It is convention to start comments on named things with the name of the thing, ref https://golang.org/doc/effective_go#commentary

> If every doc comment begins with the name of the item it describes, you can use the doc subcommand of the go 
> tool and run the output through grep

cmd/registry-replacer/main.go
```diff
@@ -600,6 +635,88 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+// the config, or within any of the tests.
+func pruneUnusedBaseImages(config *api.ReleaseBuildConfiguration, resolvedConfig *api.ReleaseBuildConfiguration) {
+	usedBaseImages := sets.NewString()
```
alvaroaleman, line 641: Intuitively I would have thought that this won't need the resolved config, as steps shouldn't really reference images from base_images, am I wrong about that?

cmd/registry-replacer/main.go
```diff
@@ -600,6 +635,88 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+	for _, step := range resolvedConfig.RawSteps {
```
alvaroaleman, line 642: What about
* PipelineImageCacheStepConfiguration
* SourceStepConfiguration
* ProjectDirectoryImageBuildStepConfiguration
* RPMServeStepConfiguration
* OutputImageTagStepConfiguration
?

/cc @bbguimaraes 
As I am not 100% familiar with the intrinsics of all steps types either

cmd/registry-replacer/main.go
```diff
@@ -600,6 +635,88 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+		usedBaseImages.Insert(string(image.From))
+		for input := range image.Inputs {
+			usedBaseImages.Insert(input)
+		}
+	}
+
+	for sourceImage := range config.BaseImages {
+		var keep bool
+		for _, candidate := range usedBaseImages.List() {
+			orgRepoTag, _ := orgRepoTagFromPullString(candidate)
```
alvaroaleman, line 663: Why is the error not being handled?

cmd/registry-replacer/main.go
stevekuznetsov: Yep, they can and do - but it's not ideal. More often they refer to images from the release.

cmd/registry-replacer/main.go
emilvberglind: I guess I was thinking that we wouldn't want to abort and stop up the rest of the process just because we couldn't resolve the config and therefore not prune the base images, though maybe I was just being overly paranoid. I can definitely return the error here

cmd/registry-replacer/main.go
emilvberglind: Right, I keep forgetting that :)

cmd/registry-replacer/main.go
emilvberglind: Lol, I can remove that space. My OCD wanted it there, but I definitely see your point.

cmd/registry-replacer/main.go
emilvberglind: Ok, that makes sense. I wasn't sure if the `registry-replacer` would always have access to the registry locally but if it does it definitely makes sense to avoid having to make an unnecessary http call.

cmd/registry-replacer/main.go
emilvberglind: My thought was that it might not be a well-formed pull string and could therefore error out, but we wouldn't care. However, looking more into the `orgRepoTagFromPullString` it should work for all of the valid cases, so if there is an error then it would be a valid error and should be handled.

cmd/registry-replacer/main.go
emilvberglind: I looked through all of the steps and did not think that any of those steps would need to be taken into consideration here, but I could definitely be wrong.

cmd/registry-replacer/main.go
emilvberglind: Ah, I thought that only applied to exported functions.

cmd/registry-replacer/main.go
alvaroaleman: I mean in the beginning this might be ok but once everything worked we need to remove it because otherwise we will not notice errors here, as no one looks at successful jobs. Maybe initially merge it like this and do a follow-up when the job is green?

cmd/registry-replacer/main.go
alvaroaleman: We run it as part of the autoconfigbrancher tool which is essentially a tool that runs a bunch of other tools to somehow fix config periodically and then creates an automerging PR. Initially it was for branching, nowadays for all kinds of things: https://github.com/openshift/ci-tools/blob/d861eedff2eb28b479d20d465f9b670d3097691d/cmd/autoconfigbrancher/main.go#L157

The job that runs it has both the openshift/release and the ocp-build-data repo: https://github.com/openshift/release/blob/425ad253347fda93e6079dcc148c3e865463d917/ci-operator/jobs/infra-periodics.yaml#L631-L637

cmd/registry-replacer/main.go
emilvberglind: That sounds good to me. I definitely agree that it _should_ return the actual error. I'll merge it like this and follow up with that like you said.

cmd/registry-replacer/main.go
```diff
@@ -278,6 +297,19 @@ func replacer(
 			if err := pruneUnusedReplacements(config, allReplacementCandidates); err != nil {
 				return fmt.Errorf("failed to prune unused replacements: %w", err)
 			}
+
+			// to prune base images we'll need the fully-resolved config.
+			resolvedConfig, err := configResolver()
+			if err != nil {
+				// TODO this should stop execution rather than just logging an error, but since we want a smooth transition with
+				// the new registry-replacer enhancements we'll add that in later, after this has been rolled out.
+				logrus.Error(fmt.Errorf("failed to resolve config. This means that base image pruning will not function as expected and will therefore be skipped: %w", err))
```
alvaroaleman, line 306:
```suggestion
				logrus.WithError(err).Error("failed to resolve config. This means that base image pruning will not function as expected and will therefore be skipped"))
```

cmd/registry-replacer/main.go
```diff
@@ -278,6 +297,19 @@ func replacer(
...
+			} else {
+				err := pruneUnusedBaseImages(config, resolvedConfig)
```
alvaroaleman, line 308: Nit: Doing it like below is more concise and reduces the scope of the `err` variable (although the latter is irrelevant here):
```suggestion
				if err := pruneUnusedBaseImages(config, resolvedConfig); err != nil {
				   handle
				 }
```

cmd/registry-replacer/main.go
alvaroaleman: Ok, this is something that _looks_ to me like it is needed but I am not really sure, lets wait for @bbguimaraes or @stevekuznetsov to chime in

cmd/registry-replacer/main.go
```diff
@@ -189,6 +202,11 @@ type usernameToken struct {
 	token    string
 }
 
+func configResolver(config *api.ReleaseBuildConfiguration, agent agents.RegistryAgent) (*api.ReleaseBuildConfiguration, error) {
+	resolvedConfig, err := agent.ResolveConfig(*config)
```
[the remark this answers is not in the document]
emilvberglind: The main reason was that we only need to resolve the config if we needed the fully resolved config, which is when pruning images.

cmd/registry-replacer/main.go
```diff
@@ -202,6 +220,7 @@ func replacer(
 	promotionTargetToDockerfileMapping map[string]dockerfileLocation,
 	majorMinor ocpbuilddata.MajorMinor,
 	credentials *usernameToken,
+	configResolver func() (*api.ReleaseBuildConfiguration, error),
```
alvaroaleman, line 223:
```suggestion
	configResolver func(*api.ReleaseBuildConfiguration) (*api.ReleaseBuildConfiguration, error),
```

alvaroaleman (commented):
That is what I meant by https://github.com/openshift/ci-tools/pull/2038/files#r649995357

cmd/registry-replacer/main.go
emilvberglind: I keep forgetting that go has this, definitely nicer!

cmd/registry-replacer/main.go
bbguimaraes: We'd definitely have to look at all steps that can reference one of the base images, since not doing so and removing the images would result in an invalid configuration.  Some of those steps are even in use in `openshift/release`:

```
$ jq -r < config.json '[.[].raw_steps[]?|keys]|flatten|sort|unique[]'
output_image_tag_step
pipeline_image_cache_step
project_directory_image_build_step
rpm_image_injection_step
```

The annoyance of having to reinvent code to do this every time is what I was complaining about in the context of `DPTP-2259`.  There is no good way to enumerate these steps statically, but any one whose `Requires` method returns an `InternalImageLink` with a reference to the `pipeline` stream qualifies.

emilvberglind: Ah ok, thanks for letting me know! I did run the `configresolver` after the changes and all of the configs in `openshift/release` did pass, so I assumed the changes were legit. I'll go back and ensure that all of those other steps are taken into account.

cmd/registry-replacer/main.go
```diff
@@ -200,9 +209,8 @@ func replacer(
 	ensureCorrectPromotionDockerfile bool,
 	ensureCorrectPromotionDockerfileIgnoredrepos sets.String,
 	promotionTargetToDockerfileMapping map[string]dockerfileLocation,
-	majorMinor ocpbuilddata.MajorMinor,
-	credentials *usernameToken,
-) func(*api.ReleaseBuildConfiguration, *config.Info) error {
+	majorMinor ocpbuilddata.MajorMinor, credentials *usernameToken,
+	configResolver func(config api.ReleaseBuildConfiguration) (api.ReleaseBuildConfiguration, error)) func(*api.ReleaseBuildConfiguration, *config.Info) error {
```
alvaroaleman, line 213: Could you keep the new line before the return signature, this is pretty confusing to read :grimacing:

cmd/registry-replacer/main.go
emilvberglind: I added some additional logic to account for the other step configurations. Hopefully this takes care of it, though I'm still a bit fuzzy on all of the various step types, so let me know if I'm still missing something.

cmd/registry-replacer/main.go
```diff
@@ -600,6 +621,104 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+			usedBaseImages.Insert(string(step.OutputImageTagStepConfiguration.From))
+		case step.PipelineImageCacheStepConfiguration != nil:
+			usedBaseImages.Insert(string(step.PipelineImageCacheStepConfiguration.From))
+		case step.RPMImageInjectionStepConfiguration != nil:
+			usedBaseImages.Insert(string(step.RPMImageInjectionStepConfiguration.From))
+		case step.SourceStepConfiguration != nil:
+			usedBaseImages.Insert(string(step.SourceStepConfiguration.From))
+		case step.TestStepConfiguration != nil:
+			getTestStepImages(&usedBaseImages, step.TestStepConfiguration)
+
```
stevekuznetsov, line 642: return an error in the default case since it's a programming error not to check every possible type - i wish Go had better enum handling :|

stevekuznetsov (commented):
I think this looks pretty good!

stevekuznetsov (approved):

PR #2044, opened 2021-06-10
Misc updates to the pod-scaler

stevekuznetsov:
pod-scaler: support http for prometheus

---

pod-scaler: close over different not exists errors

---

pod-scaler: create directory for cache when necessary

---

pod-scaler: query serially

We don't *need* to send off all of our requests all at once to every
single build farm at the same time, since we have no real requirement to
have the most up-to-date data at any point. In that vein, if we query
sequentially we can reduce the overall memory footprint of the producer
by only loading one dataset into memory at once.

---


alvaroaleman (approved):

stevekuznetsov:
Part of DPTP-2349

PR #2045, opened 2021-06-10
Repo-Init: Write out sharded plugin config

alvaroaleman:
Repo-Init updadates the plugin config and then writes it out. This means
  that it has to write it out sharded, otherwise the config becomes
  unloadable, as there is a central one and the sharded ones, conflicting
  with each other.

Ref https://issues.redhat.com/browse/DPTP-2258


pkg/prowconfigsharding/prowconfigsharding.go
```diff
@@ -0,0 +1,48 @@
...
+func MkdirAndWrite(fs afero.Fs, path string, content interface{}) error {
+	dir := filepath.Dir(path)
+	if err := fs.MkdirAll(dir, 0755); err != nil {
+		return fmt.Errorf("failed to make dir %s: %w", dir, err)
+	}
+	serialized, err := yaml.Marshal(content)
+	if err != nil {
+		return fmt.Errorf("failed to serialize: %w", err)
+	}
+	if err := afero.WriteFile(fs, path, serialized, 0644); err != nil {
```
droslean, line 44: why do we need to use the `afero` library here?

alvaroaleman: To make the whole thing testable

droslean (approved):

PR #2046, opened 2021-06-10
DPTP-2266: Pass OO_ params to ci-operator as multi-stage-param.

emilvberglind:
With [DPTP-2266](https://issues.redhat.com/browse/DPTP-2266) we're taking the first step towards removing special logic from the CVP jobs by passing the OO_ params as multi-stage-param entries to the ci-operator.

### DIFF
--- a/cmd/cvp-trigger/main.go
+++ b/cmd/cvp-trigger/main.go
@@ -14,6 +14,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 
+	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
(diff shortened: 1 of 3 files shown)


cmd/cvp-trigger/main.go
```diff
@@ -350,3 +356,19 @@ func writeResultOutput(prowjobResult jobResult, outputPath string) error {
 
 	return nil
 }
+
+// appendMultiStageParams passes all of the OO_ params to ci-operator as multi-stage-params.
+// TODO: After all of the CVP changes required to accept the multi-stage-params have been made, we need to go back and
+// remove the OO_ env params from here.
```
petr-muller, line 362: I think the TODO is a bit misleading: this needs to stay, what will need to be removed is setting them as envvars. Basically the `envvars` on L225 will be initialized empty rather than from `params`.

No need to fix ofc, just wanted to clarify.

petr-muller (approved):

cmd/cvp-trigger/main.go
emilvberglind: Yeah, you're right - that comment is a little misleading. I must've written it when I was tired, lol

PR #2047, opened 2021-06-11
ci-operator: propagate contexts as necessary

stevekuznetsov:
It is not clear why these were never propagated in the first place. We
use the central context in normal flows and the cleanup context in
post-interrupt code.

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -135,7 +135,7 @@ func (s *templateExecutionStep) run(ctx context.Context) error {
 	}()
 
 	logrus.Debugf("Creating or restarting template instance")
-	_, err := createOrRestartTemplateInstance(s.client, instance)
+	_, err := createOrRestartTemplateInstance(ctx, s.client, instance)
 	if err != nil {
 		return fmt.Errorf("could not create or restart template instance: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


stevekuznetsov:
```
 error: error creating buildah builder: Error writing blob: error storing blob to file "/var/tmp/storage287682306/6": error happened during read: unexpected EOF 
```

/retest

pkg/steps/template.go
```diff
@@ -442,30 +442,27 @@ func createOrRestartPod(podClient ctrlruntimeclient.Client, pod *coreapi.Pod) (*
 	return pod, nil
 }
 
-func waitForPodDeletion(podClient ctrlruntimeclient.Client, namespace, name string, uid types.UID) error {
-	timeout := 600
-	for i := 0; i < timeout; i += 2 {
+func waitForPodDeletion(ctx context.Context, podClient ctrlruntimeclient.Client, namespace, name string, uid types.UID) error {
+	return wait.ExponentialBackoffWithContext(ctx, wait.Backoff{Duration: 10 * time.Second, Factor: 2, Steps: 10}, func() (done bool, err error) {
```
alvaroaleman, line 446: 10 times duplicating the duration, starting with 10 gives a total of about 10k seconds intead of the 600 seconds we had before or am I misunderstanding this?
/hold

alvaroaleman (commented):
One comment, rest looks good

pkg/steps/template.go
stevekuznetsov: Oops, I wanted 1024, sorry -.-

stevekuznetsov:
@alvaroaleman updated for an initial delay of 2s, so 2^10=1024s

alvaroaleman (approved):

PR #2048, opened 2021-06-11
pod-scaler: misc flag improvements

stevekuznetsov:
pod-scaler: allow bounding query upper time limit

With --ignore-latest <duration> the pod scaler can now ignore the latest
<duration> of data from Prometheus. This may be useful in production to
not read data from tests still in progress, but is more useful in tests
to configure which data should be read in any one execution.

---

pod-scaler: allow producing just once

Running through all the queries and writing out the cached data just
once instead of on an interval is useful in testing.

---

### DIFF
--- a/cmd/pod-scaler/main.go
+++ b/cmd/pod-scaler/main.go
@@ -45,8 +45,8 @@ type options struct {
 }
 
 type producerOptions struct {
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
Part of DPTP-2349

PR #2050, opened 2021-06-14
[CVP-1467] add support for the kube objects

jsztuka:
Creation of optional parameter PYXIS_URL that will be used in newly created step `optional-operators-cvp-common-apply-secrets` in o/release.
I'm not 100% sure if this is the correct way of implementing the variable in the new way.


jsztuka:
@petr-muller can you please confirm/deny the way that the `OOPyxisUrl` variable is implemented?
Will be the variable `OOPyxisUrl` accessible in o/release steps?

petr-muller:
@jsztuka so, again. please hold on until we make the full switch to multi-stage parameters. The step needs to have such multi-stage parameter (https://docs.ci.openshift.org/docs/architecture/step-registry/#parameters), then cvp-trigger can set its value when triggering the jobs.

petr-muller:
@emilvberglind works on the switch

PR #2054, opened 2021-06-14
Prow Yaml creator: Fix chdir issues

alvaroaleman:
The PR creation changes the working dir, so we can not do it in parallel
or bad things will happen.

Also, we need to chdir before removing a repo, otherwise git clone will
fail as it requires to be in a valid dir.

Sample failure: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-operator-yaml-creator/1404544266985803776#1:build-log.txt%3A309

### DIFF
--- a/cmd/ci-operator-yaml-creator/main.go
+++ b/cmd/ci-operator-yaml-creator/main.go
@@ -1,18 +1,17 @@
 package main
 
 import (
-	"context"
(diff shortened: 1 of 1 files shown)


stevekuznetsov (approved):
sad that the PR library does that

PR #2055, opened 2021-06-15
PoC: tool to monitor/enforce CI job presence in repos

petr-muller:
/cc @ravisantoshgudimetla @dmage @stbenjam @wking 

Prototype/PoC, submitting just to get feedback.

This is an amalgamation of the custom code I used when I worked on the "missing gating CI jobs" and the code that Oleg wrote (https://github.com/dmage/openshift-trt/blob/master/main.go). Given a release image, it gets a list of images from the release, walks the CI config, finds the configs that promote these images (this is different from Oleg's code which gets repos straight from releas image metadata- this may be outdated, but it should be interesting to crosscheck). It then checks if the repository has jobs configured that implement the desired tests. It does not check test names, but it checks if the jobs use an allowlisted workflow and does not modify it.

### DIFF
--- a/cmd/trt/main.go
+++ b/cmd/trt/main.go
@@ -1,0 +1,283 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"sort"
+	"strings"
+	"sync"
+
+	"github.com/hashicorp/go-version"
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
+	"golang.org/x/sync/semaphore"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	repoAnnotation = "io.openshift.build.source-location"
+)
+
+// ocpImages is image -> repo
+type ocpImages map[string]string
+
+// getImageFromReleaseImage return version, images->repo map, error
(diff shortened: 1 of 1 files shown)


cmd/trt/main.go
```diff
@@ -0,0 +1,283 @@
...
+		cov := coverage[repo]
+		if cov.Requirements.Missing() > current {
+			current = cov.Requirements.Missing()
+			fmt.Printf("\n=== Repositories that miss %d jobs:\n", current)
+		}
+		fmt.Printf(" * %s\n", cov.SummaryLine(repo))
+	}
+}
+
+var parallelWorflows = sets.NewString(
```
dmage, line 208:
```suggestion
var parallelWorkflows = sets.NewString(
```

cmd/trt/main.go
```diff
@@ -0,0 +1,283 @@
...
+
+// getImageFromReleaseImage return version, images->repo map, error
+func getImagesFromReleaseImage(releaseImage string) (string, ocpImages, error) {
+	images := ocpImages{}
+	releaseInfoRaw, err := exec.Command("oc", "adm", "release", "info", releaseImage, "--commits", "--output=json").CombinedOutput()
+	if err != nil {
+		return "", nil, err
+	}
+
+	// TODO: There's probably a struct for this somewhere
```
stbenjam, line 37: Seems to be here https://github.com/openshift/oc/blob/master/pkg/cli/admin/release/info.go#L680-L703

stevekuznetsov: not sure if it's kosher to import the `oc` CLI instead of the API repos, should check if it a) messes with go mod and b) increases out compilation time by an hour before we do it ;)

PR #2057, opened 2021-06-15
Test-Image-Distributor: Distribute images for new config

alvaroaleman:
This change makes the test-image-distributor distribute images for new config as soon as new config is present, rather than just allowlisting the image in the filter, but not actually distributing it until it changes or the controller is restarted. It is divided in two commits:

* The first commit extends the ci-operator config agent to allow to subscribe to index changes
* The second commit makes the controller subscribe to changes go its index and trigger a reconciliation of all images for new config
Ref https://issues.redhat.com/browse/DPTP-2218

### DIFF
--- a/pkg/controller/test-images-distributor/test_images_distributor_test.go
+++ b/pkg/controller/test-images-distributor/test_images_distributor_test.go
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"sync"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
@@ -23,6 +24,7 @@ import (
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	fakeclient "sigs.k8s.io/controller-runtime/pkg/client/fake"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	imagev1 "github.com/openshift/api/image/v1"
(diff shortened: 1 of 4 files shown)


pkg/controller/test-images-distributor/test_images_distributor_test.go
```diff
@@ -784,3 +786,81 @@ type noOpRegistryResolver struct{}
...
+				t.Fatalf("failed to start source: %v", err)
+			}
+			changeChannel <- tc.change
+			close(changeChannel)
+
+			// The source copies the data from the input channel to the workqueue asynchronously
+			// and there is no way for us to properly synchronize that so instead we just wait.
+			// We can not poll because for negative testcases (expect no request) that would instanly
+			// succeed and later events won't make the test fail correctly.
+			time.Sleep(10 * time.Second)
```
stevekuznetsov, line 845: I guess technically you could have `sourceForConfigChangeChannel` send the events to both the source and to another optional given channel, which would be set only in test code, to allow synchronization. Since this is all in parallel it should not matter to overall unit test runtime though and the current impl is good

pkg/load/agents/configAgent.go
```diff
@@ -213,6 +235,8 @@ func (a *configAgent) loadFilenameToConfig() error {
 }
 
 func (a *configAgent) buildIndexes() {
+	oldIndexes := a.indexes
```
stevekuznetsov, line 238: nit: plural of index is indices ;)

pkg/load/agents/configAgent.go
```diff
@@ -239,5 +263,89 @@ func (a *configAgent) buildIndexes() {
...
+		// Building the diff is expensive, so cache it in case we have multiple
+		// subscribers.
+		var changes []IndexDelta
+		for _, channel := range a.indexSubscribers[indexName] {
+			if changes == nil {
+				changes = buildIndexDelta(oldIndexes[indexName], a.indexes[indexName])
+			}
+
+			// This might block, so do it in a new goroutine
+			channel := channel
```
stevekuznetsov, line 276: nit: totally a style thing but I think making this a named parameter of the anonymous func in the routine and passing it is a little more explicit as to what the code is doing

```go
func(channel chan <-IndexDelta) {
    //...
}(channel)
```

stevekuznetsov (approved):
The delta subscription and reaction code both look good

pkg/load/agents/configAgent.go
alvaroaleman: I hate that notation because you can not read the code from top to bottom anymore, you have to go to the start of the closure, then to the end to see what  arguments it gets, then to the start again to read it.

This notation here allows to simply read the code from top to bottom.

PR #2064, opened 2021-06-16
ci-operator-yaml-creator: Ignore variants

alvaroaleman:
Not doing this makes us race when a variant is present and update the PR
with the first variant and then the second on, visible e.G. on
https://github.com/openshift/origin/pull/26227


alvaroaleman:

hongkailiu (approved):

PR #2062, opened 2021-06-16
pkg/validation: validate all pipeline image inputs

bbguimaraes:
https://issues.redhat.com/browse/DPTP-2259

`ci-operator` currently has non-existent or inconsitent checks for whether the
`ImageStreamTag`s in the `pipeline` `ImageStream` conflict.  It is possible to
create configurations that result execution graphs with surprising and incorrect
behavior (see the Jira issue for examples).

A consistent set of validations is added, replacing existing ones where
appropriate, so that those configurations are not accepted.  Once we have proper
checking (see the task list below), the incorrect configuration presented in the
Jira issue will be rejected:

```
$ ci-operator-checkconfig --config-dir ci-operator/config --registry ci-operator/step-registry
ERRO[0000] Failed to load CI Operator configuration      error="invalid ci-operator config: invalid configuration: images[0]: duplicate image name 'elasticsearch-proxy' (previously defined by field 'base_images[elasticsearch-proxy]')" source-file=/home/bbguimaraes/src/release/ci-operator/config/openshift/elasticsearch-proxy/openshift-elasticsearch-proxy-master.yaml
error validating configuration files: invalid ci-operator config: invalid configuration: images[0]: duplicate image name 'elasticsearch-proxy' (previously defined by field 'base_images[elasticsearch-proxy]')
$ find ci-operator/config/openshift/elasticsearch-proxy/ -type f -exec sed -i '/^  elasticsearch-proxy:$/,+3d' {} +
$ ci-operator-checkconfig --config-dir ci-operator/config --registry ci-operator/step-registry
```

/hold

Issues still remaining before this can be adopted:

- [ ] fix configurations in `openshift/release`
- [ ] make these validations blocking in `ci-tools` and `openshift/release`
- [ ] make sure documentation reflects the validation

PR #2069, opened 2021-06-16
pkg/validation: add a `struct` for field paths

bbguimaraes:
In preparation to address the inconsistency highlighted in:

https://github.com/openshift/ci-tools/pull/2062#discussion_r652757992

### DIFF
--- a/pkg/validation/test.go
+++ b/pkg/validation/test.go
@@ -30,19 +30,20 @@ const (
 )
 
 // context contains the information from parent components.
-// All but `fieldRoot` can be nil if the validation being performed in
+// All but `field` can be nil if the validation being performed in
 // context-independent.
 type context struct {
-	fieldRoot  string
+	field      fieldPath
 	env        api.TestEnvironment
 	seen       sets.String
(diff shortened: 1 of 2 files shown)


pkg/validation/test.go
```diff
@@ -30,30 +30,39 @@ const (
...
 	seen       sets.String
 	leasesSeen sets.String
 	releases   sets.String
 }
 
-func newContext(fieldRoot string, env api.TestEnvironment, releases sets.String) context {
-	return context{
-		fieldRoot:  fieldRoot,
+// newContext creates a top-level context.
+func newContext(field fieldPath, env api.TestEnvironment, releases sets.String) *context {
```
bbguimaraes, line 44: This is a value type in my mind, but Go would not let me use `newContext(…).errorf` because of the pointer receiver.

I gave up trying to fight it and changed all return types to pointers.

bbguimaraes: Hopefully things get reasonably optimized since it's private to the module.

stevekuznetsov (approved):

PR #2074, opened 2021-06-17
DPTP-2262: Support importing cluster claim release streams

AlexNPavel:
This PR adds support for importing the release streams of cluster claims
for tests that use cluster claims. To avoid conflicts that can result in
naming collisions, the test step with the cluster claim creates an
import step using the name provided by the user appended by the name of
the test.

The `DependencyParts` function has been modified to accept a struct
called `ClaimRelease` containing this name (`ReleaseName`) and the name
of the release that it overrides (`OverrideName`). If the release stream
for the `OverrideName` in the `ClaimRelease` matches the imagestream
that the user requests, `DependencyParts` returns the image stream for
the release in `ClaimRelease.ReleaseName`. `DependencyParts` is used
extensively across the project when resolving image streams, which
simplifies overriding image names across the project.


pkg/defaults/defaults.go
```diff
@@ -330,45 +335,60 @@ func stepForTest(
...
 		}
+		var testSteps []api.Step
 		step := steps.MultiStageTestStep(*c, config, params, podClient, jobSpec, leases)
 		if len(leases) != 0 {
 			step = steps.LeaseStep(leaseClient, leases, step, jobSpec.Namespace)
 			addProvidesForStep(step, params)
 		}
 		if c.ClusterClaim != nil {
 			step = steps.ClusterClaimStep(c.As, c.ClusterClaim, hiveClient, client, jobSpec, step)
+			if c.ClusterClaim.As != "" {
```
stevekuznetsov, line 354: Are we going to default this to something? Or do we need users to explicitly opt into `as: latest` to get their CLI injection to work?

pkg/defaults/defaults.go
```diff
@@ -845,3 +865,34 @@ func ensureImageStreamTag(ctx context.Context, client ctrlruntimeclient.Client,
...
+	}
+	if err := hiveClient.List(ctx, clusterPools, listOption); err != nil {
+		return "", fmt.Errorf("failed to list cluster pools with list option %v: %w", listOption, err)
+	}
+
+	l := len(clusterPools.Items)
+	if l == 0 {
+		return "", fmt.Errorf("failed to find a cluster pool providing the cluster: %v", listOption)
+	} else if l > 1 {
+		return "", fmt.Errorf("find %d cluster pools providing the cluster (%v): should be only one", len(clusterPools.Items), listOption)
```
stevekuznetsov, line 887:
```suggestion
		return "", fmt.Errorf("found %d cluster pools that could provide the cluster (%v), but there should be only one", len(clusterPools.Items), listOption)
```

pkg/defaults/defaults.go
```diff
@@ -845,3 +865,34 @@ func ensureImageStreamTag(ctx context.Context, client ctrlruntimeclient.Client,
...
+	}
+
+	clusterPool := clusterPools.Items[0]
```
stevekuznetsov, line 890: Why do we need this interim variable?

pkg/api/config.go
```diff
@@ -75,13 +75,20 @@ func (config *ReleaseBuildConfiguration) ImageStreamFor(image string) (string, b
 }
 
 // DependencyParts returns the imageStream and tag name from a user-provided
-// reference to an image in the test namespace
-func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependency) (string, string, bool) {
+// reference to an image in the test namespace. If a ClaimRelease is provided
```
stevekuznetsov, line 78: Ah, so you want to be able to call this "latest" but we need the _actual_ release name to be something unique so it does not collide in the namespace with the latest release that may be configured in the config for other job on the branch? Let's make the comment a bit more exhaustively clear please :)

pkg/steps/multi_stage.go
```diff
@@ -565,7 +577,13 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			addProfile(s.profileSecretName(), s.profile, pod)
 		}
 		if step.Cli != "" {
-			addCliInjector(step.Cli, pod)
+			release := step.Cli
+			if s.clusterClaim != nil {
+				if claimRelease.OverrideName == release {
```
stevekuznetsov, line 582: Could we use DependencyParts here? Maybe call with `fmt.Sprintf("%s:cli", step.Cli)`?

pkg/validation/config_test.go
```diff
@@ -780,6 +781,15 @@ func TestReleaseBuildConfiguration_DependencyParts(t *testing.T) {
 			expectedTag:    "installer",
 			explicit:       true,
 		},
+		{
+			name:           "explicit, long-form for stable, overridden by cluster claim via latest release name",
```
stevekuznetsov, line 785: we want test cases with non-nil ClaimRelease that show current behavior does not change

stevekuznetsov (commented):
I like the approach!

pkg/defaults/defaults.go
AlexNPavel: I didn't think of that, but I don't see any reason we shouldn't default the cluster claim to be imported as `latest` if not specified by the user. I'll update the PR to do that.

AlexNPavel:
Docs PR: https://github.com/openshift/ci-docs/pull/155

/hold

stevekuznetsov (approved):

AlexNPavel:
This PR needs to be rebased but neither github nor prow are complaining about it? Strange...

AlexNPavel:
Rebased

PR #2083, opened 2021-06-17
Name claim's secrets by test

hongkailiu:
https://issues.redhat.com/browse/DPTP-2332

/cc @petr-muller

### DIFF
--- a/pkg/steps/cluster_claim.go
+++ b/pkg/steps/cluster_claim.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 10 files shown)


hongkailiu:
/hold
expecting e2e to fail.

pkg/steps/cluster_claim.go
```diff
@@ -175,7 +175,11 @@ func (s *clusterClaimStep) acquireCluster(ctx context.Context, waitForClaim func
 	return claim, nil
 }
 
-func getHiveSecret(src *corev1.Secret, name, namespace string) (*corev1.Secret, error) {
+func namePerTest(name, testName string) string {
+	return fmt.Sprintf("%s-%s", testName, name)
+}
```
petr-muller, line 180: I'm worried about the length - secret names have limited length. Job names are sometimes fairly long and our suffix is too. I think we can just prune the testname to be sure we can append a suffix with known length safely?

hongkailiu: Updated.
Also replace "." by "-".

petr-muller (approved):

PR #2091, opened 2021-06-22
[CVP-2012] Make the cvp-trigger utility resistent to refused Prow connections

14rcole:
Currently, the cvp-trigger utility is used to trigger and monitor Prow jobs that are used in the CVP operator bundle validation pipeline.  Because of a known OpenShift/AWS network interaction bug, an intermittent failure in connection can happen while the Prow Job is running. It has been determined that, while the connection from cvp-trigger gets refused, the Prow Job continues to run without issue and finishes to completion, while the cvp-trigger run gets re-triggered on the CVP side, causing unnecessary delays in the operator testing pipeline.

This MR adds a retry feature to the prow job creation code that retries up to ten times with exponential backoff.

### DIFF
--- a/cmd/cvp-trigger/main.go
+++ b/cmd/cvp-trigger/main.go
@@ -9,14 +9,17 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/watch"
 	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	pjclientset "k8s.io/test-infra/prow/client/clientset/versioned"
 	prowconfig "k8s.io/test-infra/prow/config"
@@ -305,8 +308,16 @@ func main() {
 	selector := fields.SelectorFromSet(map[string]string{"metadata.name": created.Name})
 
 	for {
-		w, err := pjclient.Watch(context.TODO(), metav1.ListOptions{FieldSelector: selector.String()})
-		if err != nil {
+		var w watch.Interface
+		if err = wait.ExponentialBackoff(wait.Backoff{Steps: 10, Duration: 1 * time.Second, Factor: 2}, func() (bool, error) {
+			var err2 error
+			w, err2 = pjclient.Watch(interrupts.Context(), metav1.ListOptions{FieldSelector: selector.String()})
+			if err2 != nil {
+				logrus.Error(err2)
+				return false, nil
+			}
+			return true, nil
+		}); err != nil {
 			logrus.WithError(err).Fatal("failed to create watch for ProwJobs")
 		}
 


cmd/cvp-trigger/main.go
```diff
@@ -383,3 +374,31 @@ func appendMultiStageParams(podSpec *v1.PodSpec, params map[string]string) {
...
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load cluster configuration")
+	}
+
+	logrus.WithFields(pjutil.ProwJobFields(prowjob)).Info("submitting a new prowjob")
+	var pjcset *pjclientset.Clientset
+	retryOptions := retry.Attempts(10)
+	err = retry.Do(
+		func() error {
+			pjcset, err = pjclientset.NewForConfig(clusterConfig)
```
petr-muller, line 391: I don't think *this* is the thing that should be that robust? At least it's not what you describe in PR description, and it's not the issue I remember we were seeing. Your description says the problem is when the prowjob is submitted and *then* there's disruption (which can happen) betweeen `cvp-trigger` and the cluster, but prowjob runs fine. This means that what you need to make robust is handling the errors in setting up `Watch().

cmd/cvp-trigger/main_test.go
```diff
@@ -106,6 +107,75 @@ func Test_getProwjob(t *testing.T) {
...
+				config:  config,
+			},
+			want:    nil,
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_ = getProwJobClient(tt.args.prowjob, tt.args.config)
+			// TODO: test to ensire the client is as expected
```
petr-muller, line 174: :trollface:

cmd/cvp-trigger/main.go
```diff
@@ -383,3 +387,17 @@ func appendMultiStageParams(podSpec *v1.PodSpec, params map[string]string) {
 		podSpec.Containers[0].Args = append(podSpec.Containers[0].Args, fmt.Sprintf("--multi-stage-param=%s=%s", key, params[key]))
 	}
 }
+
+func getProwWatcher(pjclient pjtypes.ProwJobInterface, selector fields.Selector) (watch.Interface, error) {
+	var w watch.Interface
+	var err error
+	err = wait.ExponentialBackoff(wait.Backoff{Steps: 10, Duration: 1 * time.Second, Factor: 2}, func() (bool, error) {
+		w, err = pjclient.Watch(context.TODO(), metav1.ListOptions{FieldSelector: selector.String()})
```
stevekuznetsov, line 395: While you assign to `err` here it will always be overwritten by the assignment from `wait.ExponentialBackoff()` - so it would be good to use a different variable here, as this is confusing as written.

stevekuznetsov: I know it predates this PR but `context.TODO()` is not production-ready code. I see you're importing the `interrupts` package - if you wait for graceful shutdown, you could use `interrupts.Context()` to cancel on process exit. Or, use `context.Background()` if you never want to handle cancellation, explicitly.

cmd/cvp-trigger/main_test.go
```diff
@@ -1,9 +1,9 @@
 package main
-
-import (
-	"flag"
-	"fmt"
-	"os"
+ 
+pimort (
```
stevekuznetsov, line 3: I imagine this does not compile.

cmd/cvp-trigger/main_test.go
```diff
@@ -683,3 +686,33 @@ func TestAppendMultiStageParams(t *testing.T) {
 		})
 	}
 }
+
+func Test_getProwWatcher(t *testing.T) {
```
stevekuznetsov, line 690: In my opinion, the `getProwWatcher()` logic a) does not need to be in a method, as it's not reused and b) is a thin wrapper over two libraries - `wait` and the ProwJob client. Neither really benefit much from testing at a unit level like this.

stevekuznetsov (commented):
It looks like dependencies need cleanup here.

cmd/cvp-trigger/main.go
14rcole: You're right, I've updated the code to retry the `pjclient.Watch()` function instead

cmd/cvp-trigger/main.go
```diff
@@ -383,3 +374,31 @@ func appendMultiStageParams(podSpec *v1.PodSpec, params map[string]string) {
...
+	logrus.Info("getting cluster config")
+	clusterConfig, err := util.LoadClusterConfig()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load cluster configuration")
+	}
+
+	logrus.WithFields(pjutil.ProwJobFields(prowjob)).Info("submitting a new prowjob")
+	var pjcset *pjclientset.Clientset
+	retryOptions := retry.Attempts(10)
+	err = retry.Do(
```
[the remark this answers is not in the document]
14rcole: Fixed

cmd/cvp-trigger/main.go
14rcole: Got it, thanks

14rcole: Aside from code clarity, is there any value with using `err` twice here?  The inner `err` is will be used on L396-397, and the outer `err` will be used in the return statement on L402.  I'm happy to change it for clarity, I just want to make sure there's no weird scoping behaviors in Go I'm unaware of.

I've fixed the `context.TODO()` and moved the code back into the `main()` function, but I wanted to ask the question before I push the updated commit and the diff changes.

14rcole:
Correct me if I'm wrong, but I don't believe that either of the test failures above are related to the changes in this MR

cmd/cvp-trigger/main.go
stevekuznetsov: No, it should be functional with the way Go handles scoping/shadowing, just confusing.

stevekuznetsov (approved):

PR #2098, opened 2021-06-23
DPTP-2270: Make CVP produce OO_ params again as this will lessen the migration burden for ci workflows.

emilvberglind:
This is the first phase of moving these over where we need to duplicate the params so we don't break existing CVP.


stevekuznetsov (approved):
:(

PR #2100, opened 2021-06-23
POC: Basic yaml comment preserving

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-2252

A basic POC on how to do yaml unmarshal -> Modify -> Marshal without losing comments.

Implications of doing this:
* The comments are state that needs to be preserved, so all our tooling that alters config needs updating (but could be commodized e.G. in `OperateOnCIOperatorConfigDir`)
* We will need yaml tags, because the lib doesn't want to support json tags (https://github.com/go-yaml/yaml/issues/424) and the approach taken in sigs.k8s.io/yaml of json.Marshaling, then converting to yaml can not work here, as json doesn't support comments

Edit: Another possibly better interface would be to just expose an `WriteYaml(path string, data interface{}) error` that would then read the file and if it exists preserve the comments


petr-muller:
> We will need yaml tags, because the lib doesn't want to support json tags (Please allow the struct tag to be changed to "json" go-yaml/yaml#424) and the approach taken in sigs.k8s.io/yaml of json.Marshaling, then converting to yaml can not work here, as json doesn't support comments

This could cause trouble in cases where we reuse types from test-infra b/c they only have `json` :(

alvaroaleman:
> This could cause trouble in cases where we reuse types from test-infra b/c they only have json :(

Yeah, that is a good point. OTOH we might be able to convince upstream to add `yaml` tags as well? If we write some kind of script/tool that can add yaml tags identical to the json tags, this should cause minimal friction.

petr-muller:
SGTM

PR #2101, opened 2021-06-24
pod-scaler: add an e2e test

stevekuznetsov:
pod-scaler: add e2e tests that use a real Prometheus

In these tests (and the local running mode for the pod-scaler in
general), we start by filling a Prometheus server with some fake data
using the new backfill command from `promtool`. Then, we run the
pod-scaler producer and assert something about the data we've fetched.
We cannot sensibly assert about the rates, so for now we just assert
about memory usage, but this should suffice as the content of a query
does not change how we get the data from Prometheus and store it. In the
time it took to write this e2e test, the following issues were resolved:

 - the promtool utility is extremely slow, we use the latest possible
   version now to take advantages of new patches I wrote to increase
   the speed it runs by 100x
 - there were two off-by-one errors in the way that we queried
   Prometheus which would lead to double-counted data points in rare
   cases

### DIFF
--- a/cmd/pod-scaler/producer.go
+++ b/cmd/pod-scaler/producer.go
@@ -217,27 +217,12 @@ func (q *querier) execute(ctx context.Context, c *clusterMetadata, until time.Ti
 	})
 	logger.Info("Initiating queries to Prometheus.")
 	uncovered := q.uncoveredRanges(c.name, rangeFrom(r))
-	for _, uncoveredRange := range uncovered {
-		// Prometheus has practical limits for how much data we can ask for in any one request,
-		// so we take each uncovered range and split it into chunks we can ask for.
-		start := uncoveredRange.Start
-		stop := uncoveredRange.End
-		c.lock.RLock()
-		numSteps := c.maxSize - 1
-		c.lock.RUnlock()
-		for {
-			if start.Equal(uncoveredRange.End) {
-				break
-			}
-			if int64(stop.Sub(start)/r.Step) > numSteps {
-				stop = start.Add(time.Duration(numSteps) * r.Step)
-			}
-			c.wg.Add(1)
-			// subtracting 1s from the end will ensure we don't double-count samples on the edge, as ranges are inclusive
-			go q.executeOverRange(ctx, c, prometheusapi.Range{Start: start, End: stop.Add(-1 * time.Second), Step: r.Step})
-			start = stop
-			stop = uncoveredRange.End
-		}
+	c.lock.RLock()
+	numSteps := c.maxSize - 1
+	c.lock.RUnlock()
+	for _, r := range divideRange(uncovered, r.Step, numSteps) {
+		c.wg.Add(1)
+		go q.executeOverRange(ctx, c, r)
 	}
 	c.wg.Wait()
 	q.lock.RLock()
@@ -255,6 +240,39 @@ func (q *querier) uncoveredRanges(cluster string, r pod_scaler.TimeRange) []pod_
 	q.lock.RLock()
 	defer q.lock.RUnlock()
 	return pod_scaler.UncoveredRanges(r, q.data.RangesByCluster[cluster])
+}
+
+// divideRange divides a range into smaller ranges based on how many samples we think is reasonable
+// to ask for from Prometheus in one query
+func divideRange(uncovered []pod_scaler.TimeRange, step time.Duration, numSteps int64) []prometheusapi.Range {
+	var divided []prometheusapi.Range
+	for _, uncoveredRange := range uncovered {
+		// Prometheus has practical limits for how much data we can ask for in any one request,
+		// so we take each uncovered range and split it into chunks we can ask for.
+		start := uncoveredRange.Start
+		stop := uncoveredRange.End
+		for {
+			if start.After(uncoveredRange.End) {
+				break
+			}
+			steps := int64(stop.Sub(start) / step)
+			if steps <= 1 {
+				// this range is likely too small to contain novel data and asking for this in a query leads
+				// to weird behavior - if we ignore it for now, we won't call it covered and will subsume it
+				// into a larger range in the future
+				break
+			} else if steps > numSteps {
+				stop = start.Add(time.Duration(numSteps) * step)
+			}
+			divided = append(divided, prometheusapi.Range{Start: start, End: stop, Step: step})
+			// adding a step to the start will ensure we don't double-count samples on the edge, as ranges are inclusive
+			// and the query is evaulated at start, start+step, start+2*step, etc - if we started less than end+step, we
+			// would get the same value we got at the end of the previous query
+			start = stop.Add(step)
+			stop = uncoveredRange.End
+		}
+	}
+	return divided
 }
 
 func (q *querier) executeOverRange(ctx context.Context, c *clusterMetadata, r prometheusapi.Range) {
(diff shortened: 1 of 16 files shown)


stevekuznetsov:
Part of DPTP-2349

stevekuznetsov:
I need to get the block duration flag in to `promtool` (should merge in the morning) and clean up some of the command handling logic in here but this is ready for review for the overall flow.

test/e2e/pod-scaler/prometheus/backfill.go
```diff
@@ -0,0 +1,392 @@
...
+		file = fmt.Sprintf("pods/%s.json", metric)
+	}
+	if _, ok := d.ByOffset[offset].ByFile[file]; !ok {
+		d.ByOffset[offset].ByFile[file] = map[pod_scaler.FullMetadata][]*circonusllhist.HistogramWithoutLookups{}
+	}
+
+	// round-tripping ensures we have a slice of buckets in the histogram that's as small as possible,
+	// which ends up being necessary for equality - the natural growth of the slice will lead to some
+	// unused capacity in the hist structure which trips the equality function up
+	withoutLookups := circonusllhist.NewHistogramWithoutLookups(hist)
```
emilvberglind, line 76: Is that "by design" with the `circonusllhist` library? It almost sounds like a bug if it's failing equality checks due to unused capacity and nothing else.

stevekuznetsov: Good point - I will go upstream to fix it as well

test/e2e/pod-scaler/local/main.go
```diff
@@ -0,0 +1,79 @@
...
+
+	"k8s.io/test-infra/prow/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/testhelper"
+	"github.com/openshift/ci-tools/test/e2e/pod-scaler/kubernetes"
+	"github.com/openshift/ci-tools/test/e2e/pod-scaler/prometheus"
+	"github.com/openshift/ci-tools/test/e2e/pod-scaler/run"
+)
+
+func main() {
```
alvaroaleman, line 22: Why is this a binary instead of a lib?

stevekuznetsov: Because you run `make local-pod-scaler` and have the thing running locally to e.g. interact with the API, look at the UI, whatever - it's the local dev setup

stevekuznetsov:
OK, I think this is now ready ...

test/e2e/pod-scaler/prometheus/backfill.go
stevekuznetsov: (as a follow up)

test/e2e/pod-scaler/prometheus/backfill.go
```diff
@@ -0,0 +1,392 @@
...
+
+						// record a value in the kube_pod_labels metric, which we join on during queries
+						kubePodLabels.Metric = append(kubePodLabels.Metric, &prometheus_client.Metric{
+							Label:       labelPairs,
+							TimestampMs: ts,
+							Counter:     &prometheus_client.Counter{Value: pointer.Float64Ptr(1)},
+						})
+					}
+
+					// add a stale marker to the end of the series to ensure that Prometheus knows this data is done
```
stevekuznetsov, line 180: https://www.robustperception.io/staleness-and-promql

cmd/pod-scaler/producer.go
```diff
@@ -257,6 +242,39 @@ func (q *querier) uncoveredRanges(cluster string, r pod_scaler.TimeRange) []pod_
...
+				// into a larger range in the future
+				break
+			} else if steps > numSteps {
+				stop = start.Add(time.Duration(numSteps) * step)
+			}
+			divided = append(divided, prometheusapi.Range{Start: start, End: stop, Step: step})
+			// adding a step to the start will ensure we don't double-count samples on the edge, as ranges are inclusive
+			// and the query is evaulated at start, start+step, start+2*step, etc - if we started less than end+step, we
+			// would get the same value we got at the end of the previous query
+			start = stop.Add(step)
```
stevekuznetsov, line 271: https://www.robustperception.io/step-and-query_range

stevekuznetsov (commented):
added some docs on why i did some stuff

test/e2e/pod-scaler/e2e_test.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		if diff := cmp.Diff(actualIdentifiers, expectedIdentifiers); diff != "" {
+			t.Errorf("%s: did not get correct identifiers: %v", filename, diff)
+		}
+		for identifier := range c.DataByMetaData {
+			if actual, expected := len(c.DataByMetaData[identifier]), len(data[identifier]); actual != expected {
+				t.Errorf("%s: %s: expected %d histograms but got %d", filename, identifier.String(), expected, actual)
+			}
+			if strings.Contains(filename, "container_cpu_usage_seconds_total") {
+				// no sensible way to assert about histograms containing data using rate()
```
stevekuznetsov, line 86: note: this would require reverse-engineering how `rate()` works and that is entirely non-trivial ... I looked ...

stevekuznetsov (commented):
oh and *this* e2e is kind of important but the ones we really care about are the ones that test the admission server suggesting values for Pod requests/limits...

test/e2e/pod-scaler/local/main.go
alvaroaleman: If its a local test setup, why is it under the e2e tests?

stevekuznetsov: I just put everything in that new directory tree. Was more focused on the functionality than the placement - where should it go?

alvaroaleman: Probably just under cmd as `fake-pod-scaler` or sth?

stevekuznetsov: I'm ambivalent and mostly just want the `make` target to work since there's dependencies that need to be around to make it functional - I'll move it over in the next round of review.

test/e2e/pod-scaler/prometheus/backfill.go
```diff
@@ -0,0 +1,378 @@
...
+		d.ByOffset[offset].ByFile[file] = map[pod_scaler.FullMetadata][]*circonusllhist.HistogramWithoutLookups{}
+	}
+	d.ByOffset[offset].ByFile[file][identifier] = append(d.ByOffset[offset].ByFile[file][identifier], circonusllhist.NewHistogramWithoutLookups(hist))
+
+	return nil
+}
+
+// Backfill generates data and backfills it into Prometheus, returning the data written by stage.
+func Backfill(t testhelper.TestingTInterface, prometheusDir string, retentionPeriod time.Duration) *DataInStages {
+	rand.Seed(time.Now().UnixNano())
```
petr-muller, line 78: Would it be useful to write out the seed / allow to inject it, to allow reproducible runs?

petr-muller (approved):
/hold

ok, this was a hell of a review ;) The current form is really good, there is a lot of tricky places, but I never managed to propose something that much better. In a previous iteration I got scared of the `Backfill` function and it is now broken down in manageable pieces.

The one thing I'm worried about is how much could this suffer from random Prom implementation detail changes (aren't we relying too much on implementation details? hard for me to judge from the PR).

Holding to allow response / teammates reviews, but feel free to merge if needed.

test/e2e/pod-scaler/prometheus/backfill.go
stevekuznetsov: Yeah, good point

stevekuznetsov:
> The one thing I'm worried about is how much could this suffer from random Prom implementation detail changes 

@petr-muller it's kind of a chicken-and-egg problem - we do want to be fairly sure we're reading data out of Prometheus correctly, and we can only do so by catering to those details. The staleness of series and the way that range queries sample using instant queries are tricky but fundamental to how Prometheus works, so I don't think they will change.

stevekuznetsov:
@petr-muller @alvaroaleman addressed your review comments

PR #2105, opened 2021-06-24
pod-scaler: factor out types for reuse

stevekuznetsov:
We need these in tests.


petr-muller (approved):
This is just moving code around, no big logic changes, correct? I just skimmed it. If there are any fishy spots that were changed more, point at them and I can take a better look if needed.

PR #2109, opened 2021-06-24
pod-scaler: don't double count data on query edges

stevekuznetsov:
Previously, we would double-count data that happened to be on an edge
between two queries. We can offset a query by 1s (smaller than the
Prometheus sampling interval) to remedy this.

Part of [DPTP-2349](https://issues.redhat.com/browse/DPTP-2349)


stevekuznetsov:
(the forthcoming e2e tests for this, it is simply not feasible to test this at a unit level since it depends on Prometheus query API behavior)

alvaroaleman (approved):

PR #2122, opened 2021-06-25
Containertest: Add setting to clone

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-2263

This change extends the `podStep` with a setting to make it clone through `clonerefs`. It is off by default and can currently only enabled for `container` tests. There, it will be enabled by default if `from` is a `base_image`.


test/e2e/simple/config.yaml
```diff
@@ -18,6 +18,10 @@ tests:
   commands: exit 1
   container:
     from: os
+- as: cloning
```
stevekuznetsov, line 21: Maybe call this implicit cloning since it's from a base image, and add another test for explicit cloning from some other image? Can we test for the content of the repo?

alvaroaleman: Updated to have an implicit cloning test and one where we disable it.

pkg/api/types.go
```diff
@@ -977,10 +977,13 @@ type MemoryBackedVolume struct {
 type ContainerTestConfiguration struct {
 	// From is the image stream tag in the pipeline to run this
 	// command in.
-	From PipelineImageStreamTagReference `json:"from"`
+	From PipelineImageStreamTagReference `json:"from,omitempty"`
```
stevekuznetsov, line 980: This isn't optional, though?

pkg/defaults/defaults.go
```diff
@@ -769,6 +770,13 @@ func stepConfigsForBuild(
 			if test.Secret != nil {
 				test.Secrets = append(test.Secrets, test.Secret)
 			}
+			if test.ContainerTestConfiguration != nil && test.ContainerTestConfiguration.Clone == nil {
+				if config.IsBaseImage(string(test.ContainerTestConfiguration.From)) {
```
stevekuznetsov, line 774: nit:

```go
test.ContainerTestConfiguration.Clone = utilpointer.BoolPtr(config.IsBaseImage(string(test.ContainerTestConfiguration.From)))
```

pkg/steps/artifacts.go
```diff
@@ -413,6 +415,51 @@ func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.
 
 	pod.Spec.Volumes = append(pod.Spec.Volumes, logVolume, toolsVolume)
 	pod.Spec.Volumes = append(pod.Spec.Volumes, blobStorageVolumes...)
+
+	if clone {
+		codeMount, codeVolume := decorate.CodeMountAndVolume()
```
stevekuznetsov, line 420: Is there going to be anything messy with having code and $HOME on different devices that we wouldn't already have hit with the current setup?

pkg/steps/source.go
```diff
@@ -733,3 +710,30 @@ func istObjectReference(ctx context.Context, client ctrlruntimeclient.Client, re
 	}
 	return corev1.ObjectReference{Kind: "DockerImage", Name: fmt.Sprintf("%s@%s", repo, ist.Image.Name)}, nil
 }
+
+func cloneRefsOptions(jobSpec *api.JobSpec, cloneAuthConfig *CloneAuthConfig) clonerefs.Options {
```
stevekuznetsov, line 714: Is there not something upstream to do this for us?

stevekuznetsov (commented):
Mostly LGTM - couple questions

petr-muller:
LGTM

pkg/api/types.go
alvaroaleman: remainder from a previous approach, fixed

pkg/steps/source.go
alvaroaleman: For this particular bit here no, for what the pod step does kind-of. The upstream code requires a prowjob which we don't have, but we have the required attributes for said prowjob. I just constructed one for the pod step, WDYT?

pkg/steps/artifacts.go
alvaroaleman: I don't think so

pkg/steps/source.go
stevekuznetsov: Ah, gotcha - this approach sounds good

PR #2127, opened 2021-06-28
DPTP-2244, DPTP-2345: Registry-replacer should prune unused base images.

emilvberglind:
/hold

Round 2 of the registry replacer changes. I _think_ this should now account for all of the scenarios, including the ones that were previously missed. Let me know if I missed anything.

### DIFF
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -33,6 +33,8 @@ import (
 	"github.com/openshift/ci-tools/pkg/api/ocpbuilddata"
 	"github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/github"
+	"github.com/openshift/ci-tools/pkg/load"
+	"github.com/openshift/ci-tools/pkg/registry"
 	"github.com/openshift/ci-tools/pkg/steps/release"
 )
 
@@ -47,7 +49,10 @@ type options struct {
 	currentRelease                               ocpbuilddata.MajorMinor
 	pruneUnusedReplacements                      bool
 	pruneOCPBuilderReplacements                  bool
+	pruneUnusedBaseImages                        bool
+	applyReplacements                            bool
 	ensureCorrectPromotionDockerfileIngoredRepos *flagutil.Strings
+	registryPath                                 string
 	flagutil.GitHubOptions
 }
 
@@ -63,8 +68,11 @@ func gatherOptions() (*options, error) {
 	flag.IntVar(&o.maxConcurrency, "concurrency", 500, "Maximum number of concurrent in-flight goroutines to handle files.")
 	flag.StringVar(&o.ocpBuildDataRepoDir, "ocp-build-data-repo-dir", "../ocp-build-data", "The directory in which the ocp-build-data repository is")
 	flag.StringVar(&o.currentRelease.Minor, "current-release-minor", "6", "The minor version of the current release that is getting forwarded to from the master branch")
-	flag.BoolVar(&o.pruneUnusedReplacements, "prune-unused-replacements", false, "If replacements that match nothing should get pruned from the config")
+	flag.BoolVar(&o.pruneUnusedReplacements, "prune-unused-replacements", false, "If replacements that match nothing should get pruned from the config. Note that if --apply-replacements is set to false pruning will not take place.")
+	flag.BoolVar(&o.pruneUnusedBaseImages, "prune-unused-base-images", false, "If base images that match nothing should get pruned from the config")
+	flag.BoolVar(&o.applyReplacements, "apply-replacements", true, "If we should apply Dockerfile image replacements. You will probably always leave this as the default, and it's mostly used by tests that validate that base image pruning doesn't botch things. Note: If not applying replacements we will also skip unused replacement pruning.")
 	flag.BoolVar(&o.pruneOCPBuilderReplacements, "prune-ocp-builder-replacements", false, "If all replacements that target the ocp/builder imagestream should be removed")
+	flag.StringVar(&o.registryPath, "registry", "", "Path to the step registry directory")
 	flag.Parse()
 
 	var errs []error
@@ -133,6 +141,11 @@ func main() {
 		}
 	}
 
+	resolver, err := loadResolver(opts.registryPath)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load resolver")
+	}
+
 	var errs []error
 	errLock := &sync.Mutex{}
 	sem := semaphore.NewWeighted(int64(opts.maxConcurrency))
@@ -152,11 +165,16 @@ func main() {
 					},
 					opts.pruneUnusedReplacements,
 					opts.pruneOCPBuilderReplacements,
+					opts.pruneUnusedBaseImages,
+					opts.applyReplacements,
 					opts.ensureCorrectPromotionDockerfile,
 					sets.NewString(opts.ensureCorrectPromotionDockerfileIngoredRepos.Strings()...),
 					promotionTargetToDockerfileMapping,
 					opts.currentRelease,
 					credentials,
+					func(config api.ReleaseBuildConfiguration) (api.ReleaseBuildConfiguration, error) {
+						return registry.ResolveConfig(resolver, config)
+					},
 				)(config, info); err != nil {
 					errLock.Lock()
 					errs = append(errs, err)
@@ -184,6 +202,17 @@ func main() {
 	}
 }
 
+func loadResolver(path string) (registry.Resolver, error) {
+	if path == "" {
+		return nil, nil
+	}
+	refs, chains, workflows, _, _, observers, err := load.Registry(path, false)
+	if err != nil {
+		return nil, err
+	}
+	return registry.NewResolver(refs, chains, workflows, observers), nil
+}
+
 type usernameToken struct {
 	username string
 	token    string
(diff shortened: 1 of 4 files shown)


cmd/registry-replacer/main.go
```diff
@@ -606,6 +627,117 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+		}
+	}
+
+	return nil
+}
+
+func getOperatorImages(config *api.ReleaseBuildConfiguration, usedBaseImages sets.String) {
+	if config.Operator != nil {
+		for _, substitution := range config.Operator.Substitutions {
+			usedBaseImages.Insert(substitution.With)
```
bbguimaraes, line 696: We need to pass values through `DependencyParts` whenever that is used to resolve the image names.  In this case:

https://github.com/openshift/ci-tools/blob/a5b7efc5af0350059cdb2fa7a94d4c5c7e68a9ff/pkg/steps/bundle_source.go#L103

cmd/registry-replacer/main.go
```diff
@@ -606,6 +627,117 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+}
+
+func getBundleSourceImages(images *sets.String, step *api.BundleSourceStepConfiguration) {
+	for _, substitution := range step.Substitutions {
+		images.Insert(substitution.With)
+	}
+}
+
+func getTestStepImages(config *api.ReleaseBuildConfiguration, images *sets.String, step *api.TestStepConfiguration) {
+	if step.MultiStageTestConfigurationLiteral == nil {
```
bbguimaraes, line 711: Container tests can also use an input image:

https://github.com/openshift/ci-tools/blob/a5b7efc5af0350059cdb2fa7a94d4c5c7e68a9ff/pkg/steps/pod.go#L162
https://github.com/openshift/ci-tools/blob/a5b7efc5af0350059cdb2fa7a94d4c5c7e68a9ff/pkg/steps/pod.go#L132

cmd/registry-replacer/main.go
```diff
@@ -606,6 +627,117 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+	for _, image := range resolvedConfig.Images {
+		usedBaseImages.Insert(string(image.From))
+		for input := range image.Inputs {
+			usedBaseImages.Insert(input)
+		}
+	}
+
+	for sourceImage := range config.BaseImages {
+		var keep bool
+		for _, candidate := range usedBaseImages.List() {
```
bbguimaraes, line 671: This is the same as iterating over the keys, it doesn't need to construct a temporary slice.

cmd/registry-replacer/main.go
```diff
@@ -606,6 +627,117 @@ func pruneReplacements(config *api.ReleaseBuildConfiguration, filter asDirective
...
+		getTestStepImages(resolvedConfig, &usedBaseImages, &test)
+	}
+
+	for _, image := range resolvedConfig.Images {
+		usedBaseImages.Insert(string(image.From))
+		for input := range image.Inputs {
+			usedBaseImages.Insert(input)
+		}
+	}
+
```
bbguimaraes, line 668: Missing here:

- `BaseRPMImages`

bbguimaraes (commented):
Posting a partial review, I'll resume next week.

petr-muller:
LOL I reviewed this and I'm pretty sure I would never notice most of what Bruno found. Good job

cmd/registry-replacer/main.go
emilvberglind: Sorry, I'm not totally following this one. Do you mean for the Operator bundles that we'd need something like:

```suggestion
func getOperatorImages(config *api.ReleaseBuildConfiguration, usedBaseImages sets.String) {
	if config.Operator != nil {
		for _, substitution := range config.Operator.Substitutions {
			usedBaseImages.Insert(substitution.With)
		}
		for _, bundle := range config.Operator.Bundles {
			_, name, _ := config.DependencyParts(api.StepDependency{Name: bundle.As}, nil)
			usedBaseImages.Insert(name)
		}
	}
}
```

or am I totally missing it?

bbguimaraes: Yes, and the same for the substitutions, and anywhere else `DependencyParts` is used in the steps.

bbguimaraes:
@emilvberglind: given the impact of this change and the amount of time it would take to thoroughly review and anticipate every usage, how about we:

- add an option to only perform base image pruning (if that doesn't already exist)
- add a test that applies it to `openshift/release` and validates that the resulting configuration is valid

We may be able to just add it to the beginning of https://github.com/openshift/ci-tools/blob/master/test/validate-checkconfig.sh.

This way we can be certain that we'll at least not cause any disruption in `release` with this PR or any future modifications and will have a comprehensive automatic test suite without having to enumerate all possible cases.

emilvberglind:
> @emilvberglind: given the impact of this change and the amount of time it would take to thoroughly review and anticipate every usage, how about we:
> 
> * add an option to only perform base image pruning (if that doesn't already exist)
> * add a test that applies it to `openshift/release` and validates that the resulting configuration is valid
> 
> We may be able to just add it to the beginning of https://github.com/openshift/ci-tools/blob/master/test/validate-checkconfig.sh.
> 
> This way we can be certain that we'll at least not cause any disruption in `release` with this PR or any future modifications and will have a comprehensive automatic test suite without having to enumerate all possible cases.



> @emilvberglind: given the impact of this change and the amount of time it would take to thoroughly review and anticipate every usage, how about we:
> 
> * add an option to only perform base image pruning (if that doesn't already exist)
> * add a test that applies it to `openshift/release` and validates that the resulting configuration is valid
> 
> We may be able to just add it to the beginning of https://github.com/openshift/ci-tools/blob/master/test/validate-checkconfig.sh.
> 
> This way we can be certain that we'll at least not cause any disruption in `release` with this PR or any future modifications and will have a comprehensive automatic test suite without having to enumerate all possible cases.

Good call, makes sense to me.

test/validate-checkconfig.sh
```diff
@@ -18,6 +18,15 @@ for org in openshift redhat-operator-ecosystem; do
   # We need to enter the git directory and run git commands from there, our git
   # is too old to know the `-C` option.
   pushd "${clonedir}/${org}"
+
+  # First we'll run registry-replacer to prune unused images.
+  registry-replacer \
```
bbguimaraes, line 23: Does this only perform the image pruning?  It seems to have added a non-trivial amount of time to the test:

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/2127/pull-ci-openshift-ci-tools-master-checkconfig/1413572707332657152#1:build-log.txt%3A11

vs.

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/2138/pull-ci-openshift-ci-tools-master-checkconfig/1410710606075400192#1:build-log.txt%3A11

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/2164/pull-ci-openshift-ci-tools-master-checkconfig/1414979492593537024#1:build-log.txt%3A11

emilvberglind: It also does some image tag replacement stuff. Part of it is reading the Dockerfile from the GH repo, which would be network IO, but that is currently a pre-requisite to doing the image pruning as well.

bbguimaraes: > that is currently a pre-requisite to doing the image pruning

In what way?  The only case we'd miss by pruning without first replacing is if the code happened to remove a base image added by the replacement process (and if that was a unique case that was not already present in `release`), correct?

I think we can get enough signal by just pruning `release` as is without performing the expensive replacement in this test.

emilvberglind: I changed it around so that the base image pruning is not dependent upon the dockerfile image replacement stuff. Also, I added a flag to allow bypassing the image replacement that can be used in this test. Let me know if that makes sense to you.

cmd/registry-replacer/main.go
```diff
@@ -227,65 +259,69 @@ func replacer(
...
+				}
+				allReplacementCandidates.Insert(replacementCandidates.UnsortedList()...)
 			}
-			allReplacementCandidates.Insert(replacementCandidates.UnsortedList()...)
-		}
 
-		if pruneUnusedReplacementsEnabled && hasNonEmptyDockerfile {
-			if err := pruneUnusedReplacements(config, allReplacementCandidates); err != nil {
-				return fmt.Errorf("failed to prune unused replacements: %w", err)
+			if pruneUnusedReplacementsEnabled && hasNonEmptyDockerfile {
```
bbguimaraes, line 316:
```suggestion
			if hasNonEmptyDockerfile {
```

cmd/registry-replacer/main.go
```diff
@@ -64,7 +69,10 @@ func gatherOptions() (*options, error) {
 	flag.StringVar(&o.ocpBuildDataRepoDir, "ocp-build-data-repo-dir", "../ocp-build-data", "The directory in which the ocp-build-data repository is")
 	flag.StringVar(&o.currentRelease.Minor, "current-release-minor", "6", "The minor version of the current release that is getting forwarded to from the master branch")
 	flag.BoolVar(&o.pruneUnusedReplacements, "prune-unused-replacements", false, "If replacements that match nothing should get pruned from the config")
+	flag.BoolVar(&o.pruneUnusedBaseImages, "prune-unused-base-images", false, "If base images that match nothing should get pruned from the config")
+	flag.BoolVar(&o.skipApplyReplacements, "skip-apply-replacements", false, "If we should skip applying Dockerfile image replacements. You will probably not use this, and it's mostly used by tests that validate that base image pruning doesn't botch things.")
```
bbguimaraes, line 73: All other flags are positive: "perform this step".  This one looks out of place.

We should also mention in `prune-unused-replacements` that it will not be performed if replacements are disabled.

bbguimaraes (commented):
Looks good, only a couple of minor comments.

cmd/registry-replacer/main.go
emilvberglind: Yeah, I had trouble coming up with a good name for this flag. I can't really think of a good "positive" name for it. Definitely open to suggestions :)

bbguimaraes: `--apply-replacements=false --prune-unused-base-images`?

I'm not the greatest namer of things either, but that sounds better.

emilvberglind: Sure, I can change it to that.

bbguimaraes:
Very good.

/lgtm
/hold

Cancel the hold whenever convenient.

PR #2130, opened 2021-06-30
Add a new cli: dptp-pools-cm

hongkailiu:
https://issues.redhat.com/browse/DPTP-2353

### DIFF
--- a/cmd/dptp-pools-cm/main.go
+++ b/cmd/dptp-pools-cm/main.go
@@ -1,0 +1,147 @@
+package main
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify.v1"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/rest"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/logrusutil"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	poolspullsecretprovider "github.com/openshift/ci-tools/pkg/controller/cluster_pools_pull_secret_provider"
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const (
+	hiveContextName = string(api.HiveCluster)
+)
+
+var allControllers = sets.NewString(
+	poolspullsecretprovider.ControllerName,
+)
+
+type options struct {
+	leaderElectionNamespace        string
+	kubeconfig                     string
+	leaderElectionSuffix           string
+	enabledControllers             flagutil.Strings
+	enabledControllersSet          sets.String
+	dryRun                         bool
+	poolsPullSecretProviderOptions poolsPullSecretProviderOptions
+}
+
+func (o *options) addDefaults() {
+	o.enabledControllers = flagutil.NewStrings(poolspullsecretprovider.ControllerName)
+}
+
+type poolsPullSecretProviderOptions struct {
+	sourcePullSecretNamespace string
+	sourcePullSecretName      string
+}
+
+func newOpts() (*options, error) {
+	opts := &options{}
+	opts.addDefaults()
+	flag.StringVar(&opts.leaderElectionNamespace, "leader-election-namespace", "ci", "The namespace to use for leaderelection")
+	// Controller-Runtimes root package imports the package that sets this flag
+	kubeconfigFlagDescription := "The kubeconfig to use. All contexts in it will be considered a build cluster. If it does not have a context named 'hive', loading in-cluster config will be attempted."
+	if f := flag.Lookup("kubeconfig"); f != nil {
+		f.Usage = kubeconfigFlagDescription
+		// https://i.kym-cdn.com/entries/icons/original/000/018/012/this_is_fine.jpeg
+		defer func() { opts.kubeconfig = f.Value.String() }()
+	} else {
+		flag.StringVar(&opts.kubeconfig, "kubeconfig", "", kubeconfigFlagDescription)
+	}
+	flag.StringVar(&opts.leaderElectionSuffix, "leader-election-suffix", "", "Suffix for the leader election lock. Useful for local testing. If set, --dry-run must be set as well")
+	flag.Var(&opts.enabledControllers, "enable-controller", fmt.Sprintf("Enabled controllers. Available controllers are: %v. Can be specified multiple times. Defaults to %v", allControllers.List(), opts.enabledControllers.Strings()))
+	flag.StringVar(&opts.poolsPullSecretProviderOptions.sourcePullSecretNamespace, "poolsPullSecretProviderOptions.sourcePullSecretNamespace", "ci-cluster-pool", "The namespace where the source pull secret is")
+	flag.StringVar(&opts.poolsPullSecretProviderOptions.sourcePullSecretName, "poolsPullSecretProviderOptions.sourcePullSecretName", "pull-secret", "The name of the source pull secret")
+	flag.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	flag.Parse()
+
+	var errs []error
+	if opts.leaderElectionNamespace == "" {
+		errs = append(errs, errors.New("--leader-election-namespace must be set"))
+	}
+	if vals := opts.enabledControllers.Strings(); len(vals) > 0 {
+		opts.enabledControllersSet = sets.NewString(vals...)
+		if diff := opts.enabledControllersSet.Difference(allControllers); len(diff.UnsortedList()) > 0 {
+			errs = append(errs, fmt.Errorf("the following controllers are unknown but were disabled via --disable-controller: %v", diff.List()))
+		}
+	}
+
+	return opts, utilerrors.NewAggregate(errs)
+}
+
+func main() {
+	logrusutil.ComponentInit()
+
+	opts, err := newOpts()
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to get options")
+	}
+
+	ctx := controllerruntime.SetupSignalHandler()
+	ctx, cancel := context.WithCancel(ctx)
+
+	kubeconfigChangedCallBack := func(e fsnotify.Event) {
+		logrus.WithField("event", e.String()).Info("Kubeconfig changed, exiting to get restarted by Kubelet and pick up the changes")
+		cancel()
+	}
+
+	kubeconfigs, _, err := util.LoadKubeConfigs(opts.kubeconfig, kubeconfigChangedCallBack)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load kubeconfigs")
+	}
+	hiveConfig, hasHive := kubeconfigs[hiveContextName]
+	if !hasHive {
+		kubeconfigs[hiveContextName], err = rest.InClusterConfig()
+		if err != nil {
+			logrus.WithError(err).Fatalf("--kubeconfig had no context for '%s' and loading InClusterConfig failed", hiveContextName)
+		}
+		logrus.Infof("Loaded %q context from in-cluster config", hiveContextName)
+		if err := util.WatchFiles([]string{"/var/run/secrets/kubernetes.io/serviceaccount/token"}, kubeconfigChangedCallBack); err != nil {
+			logrus.WithError(err).Fatal("failed to watch in-cluster token")
+		}
+	}
+
+	mgr, err := controllerruntime.NewManager(hiveConfig, controllerruntime.Options{
+		DryRunClient:                  opts.dryRun,
+		Logger:                        ctrlruntimelog.NullLogger{},
+		LeaderElection:                true,
+		LeaderElectionReleaseOnCancel: true,
+		LeaderElectionNamespace:       opts.leaderElectionNamespace,
+		LeaderElectionID:              fmt.Sprintf("dptp-pools-cm%s", opts.leaderElectionSuffix),
+	})
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to construct manager for the hive cluster")
+	}
+
+	if err := hivev1.AddToScheme(mgr.GetScheme()); err != nil {
+		logrus.WithError(err).Fatal("Failed to add hivev1 to scheme")
+	}
+
+	if opts.enabledControllersSet.Has(poolspullsecretprovider.ControllerName) {
+		if err := poolspullsecretprovider.AddToManager(mgr, opts.poolsPullSecretProviderOptions.sourcePullSecretNamespace, opts.poolsPullSecretProviderOptions.sourcePullSecretName); err != nil {
+			logrus.WithError(err).Fatal("Failed to construct the testimagestreamimportcleaner controller")
+		}
+	}
+
+	if err := mgr.Start(ctx); err != nil {
+		logrus.WithError(err).Fatal("Manager ended with error")
+	}
+
+	logrus.Info("Process ended gracefully")
+}
--- a/hack/run-dptp-pools-cm.sh
+++ b/hack/run-dptp-pools-cm.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd $(dirname $0)/..
+
+kubeconfig="$(mktemp)"
+
+trap 'rm -f $kubeconfig' EXIT
+kubectl config view --raw>$kubeconfig
+IFS=$'\n'
+for additional_context in $(kubectl --kubeconfig $kubeconfig config get-contexts -o name|egrep -v 'hive'); do
+  kubectl --kubeconfig $kubeconfig config delete-context "$additional_context"
+done
+
+# Make sure user env var wont overrule
+unset KUBECONFIG
+
+go run  ./cmd/dptp-pools-cm \
+  --leader-election-namespace=ci \
+  --leader-election-suffix="$USER" \
+  --kubeconfig=$kubeconfig \
+  --dry-run=true
--- a/pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
+++ b/pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
@@ -1,0 +1,191 @@
+package cluster_pools_pull_secret_provider
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/sirupsen/logrus"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const ControllerName = "cluster_pools_pull_secret_provider"
+
+func AddToManager(manager manager.Manager,
+	sourcePullSecretNamespace, sourcePullSecretName string) error {
+	log := logrus.WithField("controller", ControllerName)
+
+	client := manager.GetClient()
+	r := &reconciler{
+		log:                       log,
+		client:                    client,
+		sourcePullSecretNamespace: sourcePullSecretNamespace,
+		sourcePullSecretName:      sourcePullSecretName,
+	}
+	c, err := controller.New(ControllerName, manager, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	if err := c.Watch(
+		source.NewKindWithCache(&hivev1.ClusterPool{}, manager.GetCache()),
+		clusterPoolHandler(sourcePullSecretNamespace),
+	); err != nil {
+		return fmt.Errorf("failed to create watch for clusterpools: %w", err)
+	}
+
+	if err := c.Watch(
+		source.NewKindWithCache(&corev1.Secret{}, manager.GetCache()),
+		secretHandler(sourcePullSecretNamespace, sourcePullSecretName, client),
+	); err != nil {
+		return fmt.Errorf("failed to create watch for secrets: %w", err)
+	}
+
+	r.log.Info("Successfully added reconciler to manager")
+	return nil
+}
+
+func clusterPoolHandler(sourcePullSecretNamespace string) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+		clusterPool, ok := o.(*hivev1.ClusterPool)
+		if !ok {
+			logrus.WithField("type", fmt.Sprintf("%T", o)).Error("Got object that was not a ClusterPool")
+			return nil
+		}
+		if clusterPool.Namespace == sourcePullSecretNamespace {
+			return nil
+		}
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{
+			Namespace: clusterPool.Namespace,
+			Name:      clusterPool.Name,
+		}}}
+	})
+}
+
+func requestsFactoryForSecretEvent(namespace string, client ctrlruntimeclient.Client) []reconcile.Request {
+	clusterPools := &hivev1.ClusterPoolList{}
+	if err := client.List(context.TODO(), clusterPools, ctrlruntimeclient.InNamespace(namespace)); err != nil {
+		logrus.WithField("namespace", namespace).WithError(err).Error("Failed to list the cluster pools")
+		return nil
+	}
+	var requests []reconcile.Request
+	for _, pool := range clusterPools.Items {
+		logrus.WithField("namespace", pool.Namespace).WithField("name", pool.Name).Info("Found cluster pool")
+		requests = append(requests, reconcile.Request{NamespacedName: types.NamespacedName{
+			Namespace: pool.Namespace,
+			Name:      pool.Name,
+		}})
+	}
+	return requests
+}
+
+func secretHandler(
+	sourcePullSecretNamespace string,
+	sourcePullSecretName string,
+	client ctrlruntimeclient.Client,
+) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+		secret, ok := o.(*corev1.Secret)
+		if !ok {
+			logrus.WithField("type", fmt.Sprintf("%T", o)).Error("Got object that was not a Secret")
+			return nil
+		}
+		if secret.Name != sourcePullSecretName {
+			return nil
+		}
+		var namespace string
+		if secret.Namespace != sourcePullSecretNamespace {
+			logrus.Info("The pull secret in the target namespace is changed")
+			namespace = secret.Namespace
+		} else {
+			logrus.Info("The source pull secret is changed")
+		}
+		return requestsFactoryForSecretEvent(namespace, client)
+	})
+}
+
+type reconciler struct {
+	log                       *logrus.Entry
+	client                    ctrlruntimeclient.Client
+	sourcePullSecretNamespace string
+	sourcePullSecretName      string
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithField("request", req.String())
+	err := r.reconcile(ctx, req, log)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		log.WithError(err).Error("Reconciliation failed")
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *logrus.Entry) error {
+
+	*log = *log.WithField("namespace", req.Namespace).WithField("name", req.Name)
+	log.Info("Starting reconciliation")
+
+	clusterPool := &hivev1.ClusterPool{}
+	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: req.Namespace, Name: req.Name}, clusterPool); err != nil {
+		if apierrors.IsNotFound(err) {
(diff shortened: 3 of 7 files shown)


pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+	err := r.reconcile(ctx, req, log)
+	if err != nil && !apierrors.IsAlreadyExists(err) {
+		log.WithError(err).Error("Reconciliation failed")
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+const (
```
stevekuznetsov, line 87: Let's make this config, not constant

pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+
+	target := getSecret(req.Namespace, secret)
+	if _, err := util.UpsertImmutableSecret(ctx, r.client, target); err != nil {
+		return fmt.Errorf("failed to upsert the secret %s in namespace %s: %w", target.Namespace, target.Namespace, err)
+	}
+	log.Debug("The pull secret is upserted successfully")
+	return nil
+}
+
+func getSecret(namespace string, src *corev1.Secret) *corev1.Secret {
```
stevekuznetsov, line 119: This does not "get" a secret - it clones one (why not using DeepCopy?) then replaces the namespace and adds a label.

cmd/dptp-controller-manager/main.go
```diff
@@ -271,6 +274,10 @@ func main() {
 
 	var errs []error
 	for cluster, cfg := range kubeconfigs {
+		// no jobs run on the hive cluster and hive is not considered a build cluster
```
stevekuznetsov, line 277: This smells. If we have a controller that operates on the state of one cluster, it should not be in the core manager and then `if`-checked out of running the normal way. We should just deploy a the controller we want on `hive`.

stevekuznetsov: (that furthermore allows us to to simplify the deployment significantly by e.g. using in-cluster config and SA tokens without involving the entire cluster->vault->cluster sync)

stevekuznetsov (commented):
I think you need to carefully consider the edges here. If the central secret changes, you must reconcile derivative secrets in ClusterPool namespaces to match.

pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+	r := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:   namespace,
+			Name:        src.Name,
```
stevekuznetsov, line 123: This needs to match `clusterPool.spec.pullSecretRef.name`

cmd/dptp-controller-manager/main.go
```diff
@@ -27,6 +27,7 @@ import (
 
 	imagev1 "github.com/openshift/api/image/v1"
 
+	clusterpoolreconciler "github.com/openshift/ci-tools/pkg/controller/cluster_pool_reconciler"
```
petr-muller, line 30: Hm, what's the plan with this reconciler? I noticed that this actually does not reconcile cluster pools, but secrets (based on a presence of cluster pools). Do you plan this to be something larger? At first I actually expected that this will implement the other clusterpool story, the one that would reconcile relesae images in `imageclustersets`.

Can we figure out a better name? `pools_pull_secret_provider` perhaps? (I've tried hard to avoid the word `reconciler` because pretty much every controller is a reconciler)

cmd/dptp-controller-manager/main.go
petr-muller: I second this: it looks like we'll eventually write several other pool-related controllers, so having a separate binary just for that purpose (`dptp-pools-cm`?) sounds like a good idea.

pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *logrus.Entry) error {
+
+	*log = *log.WithField("namespace", req.Namespace).WithField("name", req.Namespace)
+	log.Info("Starting reconciliation")
+
+	clusterPool := &hivev1.ClusterPool{}
+	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: req.Namespace, Name: req.Name}, clusterPool); err != nil {
+		if apierrors.IsNotFound(err) {
+			log.Debug("The cluster pool is deleted")
```
petr-muller, line 100: We may want to delete a pull secret from the namespace if this was a last pool in that namespace.

petr-muller: ofc not 100% necessary, just a suggesting

cmd/dptp-controller-manager/main.go
hongkailiu: I will do a new cli: `dptp-pools-cm`.

pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
hongkailiu: Updated.

pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
hongkailiu: Updated.

cmd/dptp-controller-manager/main.go

pkg/controller/cluster_pool_reconciler/cluster_pool_reconciler.go
hongkailiu: Let us keep it simple for now.
We could add it when it becomes necessary in the future.

pkg/util/secret.go
```diff
@@ -54,7 +54,7 @@ func UpsertImmutableSecret(ctx context.Context, client ctrlruntimeclient.Client,
 		return false, err
 	}
 	existing := &coreapi.Secret{}
-	if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: secret.Name}, existing); err != nil {
+	if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: secret.Name, Namespace: secret.Namespace}, existing); err != nil {
```
hongkailiu, line 57: This one surprised me.
How does it work without `Namespace`?

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
```diff
@@ -0,0 +1,137 @@
...
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: r.sourcePullSecretNamespace, Name: r.sourcePullSecretName}, secret); err != nil {
+		return fmt.Errorf("failed to get the secret %s in namespace %s: %w", r.sourcePullSecretName, r.sourcePullSecretNamespace, err)
+	}
+
+	target := secret.DeepCopy()
+	target.ObjectMeta.SetResourceVersion("")
```
alvaroaleman, line 122: That works? I would have expected at least the UID and selflink to clash.How about re-creating ObjectMeta after the DeepCopy?

pkg/util/secret.go
alvaroaleman: Through a namespacing client

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider_test.go
```diff
@@ -0,0 +1,165 @@
...
+					Immutable: &im,
+				}
+				if diff := cmp.Diff(expected, actual, testhelper.RuntimeObjectIgnoreRvTypeMeta); diff != "" {
+					return fmt.Errorf("actual does not match expected, diff: %s", diff)
+				}
+				return nil
+			},
+		},
+		{
+			name:   "the pool is deleted",
```
alvaroaleman, line 102:
```suggestion
			name:   "the pool doesn't exist",
```

It doesn't matter how the pool became to be absent. A name like this suggests that you want to do post-delete cleanup which is not the case

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider_test.go
```diff
@@ -0,0 +1,165 @@
...
+				return nil
+			},
+		},
+		{
+			name:   "the pool does not use the pull secret",
+			nn:     types.NamespacedName{Namespace: "ns", Name: "pool"},
+			client: fakeclient.NewFakeClient(srcSecret.DeepCopy(), pool_with_another_secret.DeepCopy()),
+			verify: func(client ctrlruntimeclient.Client) error {
+				actual := &corev1.Secret{}
+				if err := client.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: "ns", Name: "pull-secret"}, actual); !kerrors.IsNotFound(err) {
```
alvaroaleman, line 132: Nit: Can be simplified as
```suggestion
				if err := client.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: "ns", Name: "pull-secret"}, &corev1.Secret{}); !kerrors.IsNotFound(err) {
```

cmd/dptp-pools-cm/main.go
```diff
@@ -0,0 +1,147 @@
+package main
```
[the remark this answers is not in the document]
hongkailiu: A manager with one controller, ATM. :)
We probably will have more for the things that Hive does not want to do.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
```diff
@@ -0,0 +1,137 @@
...
+		sourcePullSecretName:      sourcePullSecretName,
+	}
+	c, err := controller.New(ControllerName, hiveManager, controller.Options{
+		Reconciler: r,
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	if err := c.Watch(
```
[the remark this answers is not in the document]
hongkailiu: if any pull secrets get modified, we can reconcile them all after restarting the tool.
Is it good enough?

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider_test.go
```diff
@@ -0,0 +1,165 @@
...
+			Namespace: "ns",
+			Name:      "pool",
+		},
+		Spec: hivev1.ClusterPoolSpec{
+			PullSecretRef: &corev1.LocalObjectReference{
+				Name: "pull-secret",
+			},
+		},
+	}
+	pool_with_another_secret := &hivev1.ClusterPool{
```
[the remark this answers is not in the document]
hongkailiu: Updated.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider_test.go
hongkailiu: Updated.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider_test.go
hongkailiu: Updated.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
```diff
@@ -0,0 +1,137 @@
...
+			Infof("The cluster pool does not refer secret %s", r.sourcePullSecretName)
+		return nil
+	}
+
+	secret := &corev1.Secret{}
+	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: r.sourcePullSecretNamespace, Name: r.sourcePullSecretName}, secret); err != nil {
+		return fmt.Errorf("failed to get the secret %s in namespace %s: %w", r.sourcePullSecretName, r.sourcePullSecretNamespace, err)
+	}
+
+	target := secret.DeepCopy()
```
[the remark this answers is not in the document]
hongkailiu: Sorry. I do not understand your comment.
It is copied from the source secret, but the target secret.
Maybe you meant something else?

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
```diff
@@ -0,0 +1,137 @@
...
+	hivev1 "github.com/openshift/hive/apis/hive/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const ControllerName = "pools_pull_secret_provider"
+
+func AddToManager(hiveManager manager.Manager,
```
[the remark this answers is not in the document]
hongkailiu: Done.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
```diff
@@ -0,0 +1,137 @@
+package pools_pull_secret_provider
```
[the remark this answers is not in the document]
hongkailiu: Use `cluster_pools_pull_secret_provider` instead.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
hongkailiu: Updated.

pkg/util/secret_test.go
```diff
@@ -0,0 +1,153 @@
...
+					Immutable: &im,
+				}
+				if diff := cmp.Diff(expected, actual, testhelper.RuntimeObjectIgnoreRvTypeMeta); diff != "" {
+					return fmt.Errorf("actual does not match expected, diff: %s", diff)
+				}
+				return nil
+			},
+		},
+		{
+			name:   "labels are ignored",
```
hongkailiu, line 108: Interesting.
https://github.com/openshift/ci-tools/blob/b3a4dbe39adc4d0461054019701cb41d9d6788ef/pkg/util/secret.go#L60

I cannot justify with the upstream's source: we wont enforce the labels.

hack/run-dptp-pools-cm.sh
```diff
@@ -0,0 +1,24 @@
...
+kubectl config view --raw>$kubeconfig
+IFS=$'\n'
+for additional_context in $(kubectl --kubeconfig $kubeconfig config get-contexts -o name|egrep -v 'hive'); do
+  kubectl --kubeconfig $kubeconfig config delete-context "$additional_context"
+done
+
+# Make sure user env var wont overrule
+unset KUBECONFIG
+
+go build  -v -o /tmp/dptp-pools-cm ./cmd/dptp-pools-cm
```
droslean, line 19: can we just `go run` here?

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,142 @@
...
+type reconciler struct {
+	log                       *logrus.Entry
+	client                    ctrlruntimeclient.Client
+	sourcePullSecretNamespace string
+	sourcePullSecretName      string
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
+	log := r.log.WithField("request", req.String())
+	err := r.reconcile(ctx, req, log)
```
droslean, line 83:
```suggestion
	if err := r.reconcile(ctx, req, log); err != nil && !apierrors.IsAlreadyExists(err) {
		log.WithError(err).Error("Reconciliation failed")
	} else {
		log.Info("Finished reconciliation")
	}
```
Also that means that we will log `Finished reconciliation` if the errors is `Already Exists`. Do we care?

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,142 @@
...
+		log.WithError(err).Error("Reconciliation failed")
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *logrus.Entry) error {
+
+	*log = *log.WithField("namespace", req.Namespace).WithField("name", req.Name)
```
droslean, line 94: why this has a pointer?

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,142 @@
...
+	target := secret.DeepCopy()
+	target.ObjectMeta = metav1.ObjectMeta{
+		Name:        r.sourcePullSecretName,
+		Namespace:   req.Namespace,
+		Labels:      target.Labels,
+		Annotations: target.Annotations,
+	}
+
+	if target.Labels == nil {
+		target.Labels = map[string]string{}
```
droslean, line 131: IMHO I think its safer to `make(map[string]string)`

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
hongkailiu: `err` is used outside the `if` sentence.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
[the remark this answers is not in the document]
alvaroaleman: That feels like kicking the can down the road IMHO. It shouldn't be difficult, you just have to watch them and use a predicate to filter out secrets with the wrong name, I can show you over Slack if you need help here

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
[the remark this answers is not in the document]
alvaroaleman: The reason for doing a `DeepCopy` is if you need to keep the unmodified original, for example because you do not want to manipulate your cache or because you want to construct a Patch. You don't do any of that here though, so this is useless, you can just modify and use `secret`

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
alvaroaleman: It is the exact same thing. `make` allows you to also specify a size which can make it more performant, but if you don't do that anyways, its the same.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
alvaroaleman: (And it is extremely important that that doesn't get changed, as it is responsible for the control flow (if err != nil means it will get retried after a backoff))

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
alvaroaleman: We manipulate the value behind the existing pointer we get, rather than constructing a new pointer. This way, the `log` that is used in `Reconcile` gets updated, so if we log an error there, it has all the fields we set.

hongkailiu: Thanks @alvaroaleman 
I got the impression somewhere but do not remember where.
Probably you explained to me before.

hack/run-dptp-pools-cm.sh
hongkailiu: Updated.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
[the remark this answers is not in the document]
hongkailiu: let me give it a try.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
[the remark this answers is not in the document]
hongkailiu: Updated.

pkg/controller/pools_pull_secret_provider/pools_pull_secret_provider.go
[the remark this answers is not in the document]
hongkailiu: Updated.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{
+			Namespace: clusterPool.Namespace,
+			Name:      clusterPool.Name,
+		}}}
+	})
+}
+
+func requestsFactoryForSecretEvent(namespace string, client ctrlruntimeclient.Client) []reconcile.Request {
+	if namespace != "" {
```
alvaroaleman, line 84: You can remove this and deduplicate the code, if `ctrlruntimeclient.InNamespace` gets an empty namespace, it should act like it isn't present

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,210 @@
...
+	client ctrlruntimeclient.Client,
+	requestsFactory func(namespace string, client ctrlruntimeclient.Client) []reconcile.Request,
+) handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+		secret, ok := o.(*corev1.Secret)
+		if !ok {
+			logrus.WithField("type", fmt.Sprintf("%T", o)).Error("Got object that was not a Secret")
+			return nil
+		}
+		if secret.Name != sourcePullSecretName {
```
alvaroaleman, line 131: See the other comment re namespace, this can just be
```suggestion
		if secret.Name != sourcePullSecretName || secret.Namespace == sourcePullSecret Namespace {
```

allowing you to simplify the code below and only call `requestsFactory` once

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,210 @@
...
+		logrus.WithField("namespace", namespace).WithField("name", clusterPools.Items[0].Name).Info("Found cluster pool")
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{
+			Namespace: namespace,
+			Name:      clusterPools.Items[0].Name,
+		}}}
+	}
+	var requests []reconcile.Request
+	clusterPools := &hivev1.ClusterPoolList{}
+	if err := client.List(context.TODO(), clusterPools); err != nil {
+		logrus.WithError(err).Error("Failed to list the cluster pools")
```
alvaroaleman, line 102:
```suggestion
		logrus.WithError(err).Error("Failed to list the cluster pools")
		return nil
```

cmd/dptp-pools-cm/main.go
```diff
@@ -0,0 +1,147 @@
...
+		LeaderElectionReleaseOnCancel: true,
+		LeaderElectionNamespace:       opts.leaderElectionNamespace,
+		LeaderElectionID:              fmt.Sprintf("dptp-pools-cm%s", opts.leaderElectionSuffix),
+	})
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to construct manager for the hive cluster")
+	}
+
+	if err := hivev1.AddToScheme(mgr.GetScheme()); err != nil {
+		logrus.WithError(err).Fatal("Failed to add prowv1 to scheme")
```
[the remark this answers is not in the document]
hongkailiu: Updated.

hack/run-dptp-pools-cm.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd $(dirname $0)/..
+
+kubeconfig="$(mktemp)"
+
+#trap 'rm -f $kubeconfig' EXIT
```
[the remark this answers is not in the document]
hongkailiu: Updated

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,210 @@
...
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+const ControllerName = "cluster_pools_pull_secret_provider"
+
+func AddToManager(hiveManager manager.Manager,
+	sourcePullSecretNamespace, sourcePullSecretName string) error {
+	log := logrus.WithField("controller", ControllerName)
+
+	hiveClient := hiveManager.GetClient()
```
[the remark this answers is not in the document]
hongkailiu: Updated

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,210 @@
...
+		}})
+	}
+	return requests
+}
+
+func secretHandler(
+	sourcePullSecretNamespace string,
+	sourcePullSecretName string,
+	client ctrlruntimeclient.Client,
+	requestsFactory func(namespace string, client ctrlruntimeclient.Client) []reconcile.Request,
```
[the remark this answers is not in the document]
hongkailiu: Updated

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,210 @@
...
+		clusterPools := &hivev1.ClusterPoolList{}
+		if err := client.List(context.TODO(), clusterPools, ctrlruntimeclient.InNamespace(namespace)); err != nil {
+			logrus.WithField("namespace", namespace).WithError(err).Error("Failed to list the cluster pools in namespace")
+		}
+		if len(clusterPools.Items) == 0 {
+			logrus.WithField("namespace", namespace).Info("No cluster pools are in namespace")
+			return nil
+		}
+		logrus.WithField("namespace", namespace).WithField("name", clusterPools.Items[0].Name).Info("Found cluster pool")
+		return []reconcile.Request{{NamespacedName: types.NamespacedName{
```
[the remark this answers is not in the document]
hongkailiu: I did it another way around.
Generate request only if needed.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
hongkailiu: Updated.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
hongkailiu: Updated.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
hongkailiu: Still need to call it twice because I still look at pool in reconcile.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,217 @@
...
+func requestsFactoryForSecretEvent(namespace string, client ctrlruntimeclient.Client, sourcePullSecretName string) []reconcile.Request {
+	clusterPools := &hivev1.ClusterPoolList{}
+	if err := client.List(context.TODO(), clusterPools, ctrlruntimeclient.InNamespace(namespace)); err != nil {
+		logrus.WithField("namespace", namespace).WithError(err).Error("Failed to list the cluster pools")
+		return nil
+	}
+
+	if namespace != "" {
+		for _, pool := range clusterPools.Items {
+			if referPullSecret(pool, sourcePullSecretName) {
```
alvaroaleman, line 102: Why duplicate this logic? Fetching some data from cache and checking that in `Reconcile` is extremely cheap. Simply enqueue all clusterPools here. Trying to reduce the number of requests here is an absolute micro optimization that does not warrant the big potential for hard to debug issues it introduces

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
```diff
@@ -0,0 +1,217 @@
...
+					Name:      pool.Name,
+				}}}
+			}
+		}
+		return nil
+	}
+	var requests []reconcile.Request
+	namespacesWithClusterPool := sets.NewString()
+	for _, pool := range clusterPools.Items {
+		if namespacesWithClusterPool.Has(pool.Namespace) || !referPullSecret(pool, sourcePullSecretName) {
```
alvaroaleman, line 114: Same as above, please just enqueue all cluster pools and don't try to be too smart

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider_test.go
```diff
@@ -0,0 +1,241 @@
...
+			Name:      "another-pool",
+		},
+		Spec: hivev1.ClusterPoolSpec{
+			PullSecretRef: &corev1.LocalObjectReference{
+				Name: "pull-secret",
+			},
+		},
+	}
+
+	pool_without_pull_secret = &hivev1.ClusterPool{
```
alvaroaleman, line 69: Nit: Please stick to the canonical camelCase

hongkailiu: Why did I keep doing this? Stupid me.

hongkailiu: Updated.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
hongkailiu: Updated.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider.go
hongkailiu: Updated.

pkg/controller/cluster_pools_pull_secret_provider/cluster_pools_pull_secret_provider_test.go
```diff
@@ -0,0 +1,249 @@
...
+			log := logrus.NewEntry(logrus.StandardLogger())
+			logrus.SetLevel(logrus.TraceLevel)
+			r := &reconciler{
+				log:                       log,
+				client:                    tc.client,
+				sourcePullSecretNamespace: "ci-cluster-pool",
+				sourcePullSecretName:      "pull-secret",
+			}
+			request := reconcile.Request{NamespacedName: tc.nn}
+			actual := r.reconcile(context.Background(), request, r.log)
```
alvaroaleman, line 185: Nit: Mind calling `Reconcile` instead of `reconcile` here to make this fail if someone optimizes the error return away there as previously suggested?

hongkailiu: Updated.

alvaroaleman (approved):

hongkailiu:
/hold

e2e is broken because hive is broken.

PR #2133, opened 2021-06-30
DPTP-2347: do not mirror postsubmits and periodics

AlexNPavel:
This PR updates ci-operator to ignore cloning postsubmits and periodics
in the config mirror command.


petr-muller:
LGTM, just needs to update the integration test output (the result looks fine to me: the configs need to be present in the output too, just not in the *private* repo)

PR #2136, opened 2021-07-01
prowgen: generate stricter `branch:` stanzas for presubmits

petr-muller:
In the past, we considered any `$branch$suffix` a feature branch of
"base" CI branch (e.g. PRs to `master-next` would receive presubmits
from `master` config). We need to make this convention stricter to not
break on `4.10`, because the current convention means that
`release-4.10` is a feature branch for `release-4.1`.

All generated presubmits now have two items in their `branches:` stanza.
One item that matches exactly the branch (`^branch$`), just like
postsubmits have, and one item that matches branches with `$base-`
prefix. Practically this means that we enforce exactly the `-` separator
for a branch to be considered a feature branch.

### DIFF
--- a/cmd/ci-operator-prowgen/testdata/zz_fixture_presubmit_TestFromCIOperatorConfigToProwYaml_Input_is_YAML_and_it_is_correctly_processed.yaml
+++ b/cmd/ci-operator-prowgen/testdata/zz_fixture_presubmit_TestFromCIOperatorConfigToProwYaml_Input_is_YAML_and_it_is_correctly_processed.yaml
@@ -3,7 +3,8 @@ presubmits:
   - agent: kubernetes
     always_run: true
     branches:
-    - branch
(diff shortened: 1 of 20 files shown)


petr-muller:
/hold
Needs a o/release followup after merge

stevekuznetsov (commented):
Can we get a unit test for this?

petr-muller:
> Can we get a unit test for this?
@stevekuznetsov hm, do you have a specific one in mind? `GeneratePresubmitForTest` is already unit tested (`TestGeneratePresubmitForTest`) and I just needed to change its fixtures to reflect the changed expectations about its output. Do you think there's any specific case that we would need to test?

stevekuznetsov (approved):
I was thinking something at a lower level than the full fixture-based thing but whatever

/lgtm
/hold cancel

PR #2137, opened 2021-07-01
[DPTP-2280] Add options to determine and operate only on changed files

droslean:

### DIFF
--- a/cmd/ci-operator-checkconfig/main.go
+++ b/cmd/ci-operator-checkconfig/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"os"
 	"runtime"
 	"strings"
 
@@ -25,7 +26,7 @@ type promotedTag struct {
 }
 
 type options struct {
-	configDir      string
+	config.Options
 	maxConcurrency uint
 
 	resolver registry.Resolver
@@ -33,15 +34,26 @@ type options struct {
 
 func (o *options) parse() error {
 	var registryDir string
-	flag.StringVar(&o.configDir, "config-dir", "", "The directory containing configuration files.")
-	flag.StringVar(&registryDir, "registry", "", "Path to the step registry directory")
-	flag.UintVar(&o.maxConcurrency, "concurrency", uint(runtime.GOMAXPROCS(0)), "Maximum number of concurrent in-flight goroutines.")
-	flag.Parse()
-	if o.configDir == "" {
-		return errors.New("The --config-dir flag is required but was not provided")
+
+	fs := flag.NewFlagSet("", flag.ExitOnError)
+
+	fs.StringVar(&registryDir, "registry", "", "Path to the step registry directory")
+	fs.UintVar(&o.maxConcurrency, "concurrency", uint(runtime.GOMAXPROCS(0)), "Maximum number of concurrent in-flight goroutines.")
+
+	o.Options.Bind(fs)
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		return fmt.Errorf("failed to parse flags: %w", err)
 	}
+
 	if err := o.loadResolver(registryDir); err != nil {
 		return fmt.Errorf("failed to load registry: %w", err)
+	}
+	if err := o.Options.Validate(); err != nil {
+		return fmt.Errorf("failed to validate config options: %w", err)
+	}
+	if err := o.Options.Complete(); err != nil {
+		return fmt.Errorf("failed to complete config options: %w", err)
 	}
 	return nil
 }
@@ -78,7 +90,7 @@ func (o *options) validate() (ret []error) {
 			ret = append(ret, err)
 		}
 	}()
-	if err := config.OperateOnCIOperatorConfigDir(o.configDir, func(configuration *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
+	if err := o.OperateOnCIOperatorConfigDir(o.ConfigDir, func(configuration *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
 		workCh <- workItem{configuration, repoInfo}
 		return nil
 	}); err != nil {
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -20,6 +20,8 @@ import (
 )
 
 type options struct {
+	config.Options
+
 	fromDir         string
 	fromReleaseRepo bool
 
@@ -39,6 +41,8 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.BoolVar(&opt.toReleaseRepo, "to-release-repo", false, "If set, it behaves like --to-dir=$GOPATH/src/github.com/openshift/release/ci-operator/jobs")
 
 	flag.BoolVar(&opt.help, "h", false, "Show help for ci-operator-prowgen")
+
+	opt.Options.Bind(flag)
 
 	return opt
 }
@@ -64,6 +68,15 @@ func (o *options) process() error {
 
 	if o.toDir == "" {
 		return fmt.Errorf("ci-operator-prowgen needs exactly one of `--to-{dir,release-repo}` options")
+	}
+
+	// TODO: deprecate --from-dir
+	o.ConfigDir = o.fromDir
+	if err := o.Options.Validate(); err != nil {
+		return fmt.Errorf("failed to validate config options: %w", err)
+	}
+	if err := o.Options.Complete(); err != nil {
+		return fmt.Errorf("failed to complete config options: %w", err)
 	}
 
 	return nil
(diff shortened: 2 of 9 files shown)


pkg/config/options.go
```diff
@@ -4,12 +4,20 @@ import (
...
 
 	"github.com/sirupsen/logrus"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
 )
 
+const (
+	mainBranch = "master"
```
stevekuznetsov, line 18: Let's get `clonerefs` to set configure the upstream tracking branch and use:

```
git rev-parse --abbrev-ref --symbolic-full-name @{upstream}
```

To figure this out dynamically.

pkg/config/options.go
```diff
@@ -31,6 +42,13 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("invalid --log-level: %w", err)
 	}
 	logrus.SetLevel(level)
+
+	if o.determineChanges {
```
stevekuznetsov, line 46: Why is this in `Validate()` ?

pkg/config/options.go
```diff
@@ -39,6 +57,8 @@ func (o *Options) Bind(fs *flag.FlagSet) {
 	fs.StringVar(&o.LogLevel, "log-level", "info", "Level at which to log output.")
 	fs.StringVar(&o.Org, "org", "", "Limit repos affected to those in this org.")
 	fs.StringVar(&o.Repo, "repo", "", "Limit repos affected to this repo.")
+
+	fs.BoolVar(&o.determineChanges, "determine-changes", false, "If true, compare changes with the main branch")
```
stevekuznetsov, line 61: nit: maybe `--only-process-changes` or something? This does a lot more than just determine them

pkg/config/options.go
droslean: all of the tools already run this method, so the easiest way was just to put it here.

pkg/config/options.go
droslean: This doesn't work in my environment though.

pkg/config/options.go
droslean: I moved it to a `Complete` method.

pkg/config/options.go
droslean: Btw, most of the tools won't work in a different repo than `release`. So the value will always be `master` until we change it.

stevekuznetsov: What does that do in your environment? Have you configured an upstream tracking branch?

I don't want us to hard-code this since we're going to get bitten by the cost of dealing with it once `master` changes to `main`

droslean: Updated, but I still have one concern. When a user will `make jobs` we want to operate on a subset of the changes, but this won't be possible since the user has to configure an @{upstream} in his local git release repo, right?

pkg/config/options.go
```diff
@@ -34,11 +42,51 @@ func (o *Options) Validate() error {
...
+		// D       path/of/deleted/file.yaml
+		// R100    path/of/renamed/old-name.yaml   path/of/renamed/new-name.yaml
+		o.modifiedFiles = sets.NewString()
+		for _, f := range changedFilesWithStatus {
+			statusAndFile := strings.Fields(f)
+			if statusAndFile[0] == "M" {
+				o.modifiedFiles.Insert(statusAndFile[1])
+				continue
+			}
+			o.shouldProcessChanges = false
```
stevekuznetsov, line 73: Why do we care to track two booleans? In what case do we want to know that we were asked to only process changes but will be doing everything?

pkg/config/options.go
```diff
@@ -34,11 +42,51 @@ func (o *Options) Validate() error {
...
+		// make shouldProcessChanges = false, in order to process all the files instead of a subset. For now we
+		// care of processing on modified files when there are ONLY modified files.
+		// Example output of `git diff --name-status master`
+		// M       path/of/modified/file.yaml
+		// D       path/of/deleted/file.yaml
+		// R100    path/of/renamed/old-name.yaml   path/of/renamed/new-name.yaml
+		o.modifiedFiles = sets.NewString()
+		for _, f := range changedFilesWithStatus {
+			statusAndFile := strings.Fields(f)
+			if statusAndFile[0] == "M" {
```
stevekuznetsov, line 69: A length check here would help us not panic

pkg/config/options.go
```diff
@@ -77,3 +130,17 @@ func (o *ConfirmableOptions) Bind(fs *flag.FlagSet) {
 	o.Options.Bind(fs)
 	fs.BoolVar(&o.Confirm, "confirm", false, "Create the branched configuration files.")
 }
+
+func getUpstreamBranch() string {
+	cmd := exec.Command("git", "rev-parse", "--abbrev-ref", "--symbolic-full-name", "@{upstream}")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
```
stevekuznetsov, line 137: Probably better to check for an exact error - what if this fails because there's no `git` in the system?

cmd/ci-operator-prowgen/main.go
```diff
@@ -175,18 +188,15 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to process arguments")
 	}
 
-	args := flagSet.Args()
```
stevekuznetsov, line 178: I'm not sure we want to remove this functionality in this PR - sure, it might be useful in the future to remove it, but by not having it in this specific PR you don't block yourself on migrating users, etc

pkg/config/options.go
droslean: well, the user can ask to process only the changes, but if we find that there are non `M` changes we should skip it.

pkg/config/options.go
droslean: yea if it fails we use `master` as a value. Also if there is no `git` in the system, we will fail later on either way. But I don't think that the actual error is relevant here. We try to get the upstream branch, if not, use `master`.

droslean: I added a check to look for git command in the system [here](https://github.com/openshift/ci-tools/pull/2137/commits/e7454527ca93589780fa84854d686a705d705fed#diff-6731fb6f709392ce3e37d3b0c42074cddbce566dad2bab86af24ba7585eeb57cR56-R60)

stevekuznetsov: I am making a comment about this error check - you want to fall back to `master` in the very specific case that there is no upstream tracking branch configured. Check for that error instead of swallowing everything.

pkg/config/options.go
stevekuznetsov: I understand that, but why do we need anything more than a single boolean to track whether we will or will not process everything? Set it to true when the user asks for it, set it to false when we detect non-`M` changes.

pkg/config/options.go
```diff
@@ -34,11 +42,57 @@ func (o *Options) Validate() error {
...
+		// make shouldProcessChanges = false, in order to process all the files instead of a subset. For now we
+		// care of processing on modified files when there are ONLY modified files.
+		// Example output of `git diff --name-status master`
+		// M       path/of/modified/file.yaml
+		// D       path/of/deleted/file.yaml
+		// R100    path/of/renamed/old-name.yaml   path/of/renamed/new-name.yaml
+		o.modifiedFiles = sets.NewString()
+		for _, f := range changedFilesWithStatus {
+			statusAndFile := strings.Fields(f)
+			if len(statusAndFile) > 0 && statusAndFile[0] == "M" {
```
stevekuznetsov, line 75: This `len` check can still make the `[1]` below panic

droslean: :man_facepalming:

pkg/config/options.go
droslean: addressed

pkg/config/options.go
droslean: addressed

pkg/config/options.go
```diff
@@ -34,11 +41,58 @@ func (o *Options) Validate() error {
 	return nil
 }
 
+func (o *Options) Complete() error {
+	if o.onlyProcessChanges {
```
stevekuznetsov, line 45:
```go
if !o.onlyProcessChanges {
    return nil
}
```

pkg/config/options.go
```diff
@@ -34,11 +41,58 @@ func (o *Options) Validate() error {
...
+				o.modifiedFiles.Insert(statusAndFile[1])
+				continue
+			}
+			o.onlyProcessChanges = false
+		}
+	}
+	return nil
+}
+
+func (o *Options) IsOnlyProcessChanges() bool {
```
stevekuznetsov, line 85:
```suggestion
func (o *Options) ProcessAll() bool {
```

Then your callsite becomes

```go
if opt.ProcessAll()
```

cmd/ci-operator-checkconfig/main.go
```diff
@@ -25,24 +25,29 @@ type promotedTag struct {
...
 	resolver registry.Resolver
 }
 
 func (o *options) parse() error {
 	var registryDir string
-	flag.StringVar(&o.configDir, "config-dir", "", "The directory containing configuration files.")
 	flag.StringVar(&registryDir, "registry", "", "Path to the step registry directory")
 	flag.UintVar(&o.maxConcurrency, "concurrency", uint(runtime.GOMAXPROCS(0)), "Maximum number of concurrent in-flight goroutines.")
+
+	o.Options.Bind(flag.CommandLine)
```
stevekuznetsov, line 39:
```go
flagSet := flag.NewFlagSet("", flag.ExitOnError)
```

We do this for a reason. Default flag names often collide.

cmd/ci-operator-prowgen/main.go
```diff
@@ -181,13 +194,16 @@ func main() {
...
-			fields := logrus.Fields{"target": opt.toDir, "source": opt.fromDir, "subdir": subDir}
+		if err := opt.OperateOnCIOperatorConfigDir(filepath.Join(opt.fromDir, subDir), genJobs); err != nil {
+			fields := logrus.Fields{"target": opt.toDir, "source": opt.fromDir}
 			logrus.WithError(err).WithFields(fields).Fatal("Failed to generate jobs")
 		}
-		if err := pruneStaleJobs(opt.toDir, subDir); err != nil {
-			fields := logrus.Fields{"target": opt.toDir, "source": opt.fromDir, "subdir": subDir}
-			logrus.WithError(err).WithFields(fields).Fatal("Failed to prune stale generated jobs")
+
+		if opt.ProcessAll() {
```
stevekuznetsov, line 202: Are we sure this is always going to do the right thing? There's no case where the user won't have pruning run but needs to?

stevekuznetsov:
Is there a simple way to test this?

cmd/ci-operator-prowgen/main.go
droslean: The problem is that the logic that we do the pruning doesn't fit with the current changes. For now, I would focus on pruning only when we process all the files. In the future, I can refactor the whole thing.

droslean:
> Is there a simple way to test this?

I will try to find a way.

stevekuznetsov (approved):

PR #2144, opened 2021-07-06
Release: Fix importing for non-amd64 architectures

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-2265

Makes us use a harcoded y-stream version of the `oc` image when importing a non-amd64 release as we can not resolve the correct one because we can not run the `cluster-version-operator` as it will be for the wrong arch

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -897,9 +897,9 @@ func resolveCLIOverrideImage(architecture api.ReleaseArchitecture, version strin
 	if version == "" {
 		return nil, errors.New("non-amd64 releases require a version to be configured")
 	}
-	// Should never happen™ but better than a NPD below when stripping everything after the Y stream
-	if len(version) < 3 {
-		return nil, fmt.Errorf("version %q has less than three digits", version)
+	majorMinor, err := official.ExtractMajorMinor(version)
+	if err != nil {
+		return nil, err
 	}
-	return &coreapi.ObjectReference{Kind: "ImageStreamTag", Namespace: "ocp", Name: version[:3]}, nil
+	return &coreapi.ObjectReference{Kind: "ImageStreamTag", Namespace: "ocp", Name: majorMinor}, nil
 }
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -901,5 +901,5 @@ func resolveCLIOverrideImage(architecture api.ReleaseArchitecture, version strin
 	if err != nil {
 		return nil, err
 	}
-	return &coreapi.ObjectReference{Kind: "ImageStreamTag", Namespace: "ocp", Name: majorMinor}, nil
+	return &coreapi.ObjectReference{Kind: "ImageStreamTag", Namespace: "ocp", Name: majorMinor + ":cli"}, nil
 }
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -16,6 +16,7 @@ import (
 	kapierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	coreclientset "k8s.io/client-go/kubernetes/typed/core/v1"
@@ -167,6 +168,8 @@ func fromConfig(
 			// factor release steps into something more reusable
 			hasReleaseStep = true
 			var value string
+			var overrideCLIReleaseExtractImage *coreapi.ObjectReference
+			var overrideCLIResolveErr error
 			if env := utils.ReleaseImageEnv(resolveConfig.Name); params.HasInput(env) {
 				value, err = params.Get(env)
 				if err != nil {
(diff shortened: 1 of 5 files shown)


stevekuznetsov:
Unit test flaked.

/retest
/lgtm
/hold

stevekuznetsov:
Hrm - one thing. We should import this to some *other* tag, so that the CLI tag in that release stream has no chance of being our hard-coded thing and continues to be the one that matches version with the rest of the components. Other than that, I think this works.

/lgtm cancel

smarterclayton:
This change means that PR jobs don't test and run the version of `oc` being tested, so it's possible to deliver OC regressions that break old releases (a 4.9 oc may not build a payload 4.6 can use, at worst).  It's also possible to merge oc changes (after this) that mean every other release breaks.  We also lose the verification that every payload can self identify (oc works) on that version.

In general I'd prefer to keep this property of releases (that the CLI part of the payload is used to extract the payload) and look for a different multi-arch solution.

pkg/steps/release/import_release.go
```diff
@@ -188,14 +160,32 @@ func (s *importReleaseStep) run(ctx context.Context) error {
 					Type: imagev1.LocalTagReferencePolicy,
 				},
 				From: &coreapi.ObjectReference{
-					Kind: "DockerImage",
-					Name: cliImage,
+					Kind: "ImageStreamTag",
+					// Hardcode the image to an x86 one, so this works for multi-arch.
+					Name:      "4.9:cli-artifacts",
+					Namespace: "ocp",
```
smarterclayton, line 166: This would be incorrect in some jobs - you can't make `stable:cli` not equal to the version of the release (a 4.6 cluster can't use a 4.9 cli).

If the release image is not the current platform arch, I'd suggest importing the version that matches that release to a separate tag and trying to use that (if it exists) in other ci-operator code.

smarterclayton:
/hold

Because I think the invariants here shouldn't be broken

pkg/steps/release/import_release.go
stevekuznetsov: Yeah, this is my comment above - we should be able to change it trivially to use some other intermediate target.

stevekuznetsov:
@smarterclayton all of that testing should happen in a presubmit on the `openshift/oc` repo, not implicitly in how this tool happens to run ... and besides, this is for *extracting* a release that's already been created elsewhere (and we're pulling in via cinci or release-controller) so what utility is there in having those "checks" happen on unrelated repos out-of-band with the release payload being created in the first place?

smarterclayton:
Should have added my comments from slack here.

ci-operator is a part of the release process, so you're right that presubmit blockers on `oc` (where ci-operator uses cli image to extract cli, and oc adm release new is generally coming from that version) catch the bulk of the invariant checking (that you can take an image, extract the cli, then use the cli to get info from the payload).  This code was depending on that invariant being in place, because the API contract of a payload doesn't completely guarantee we may not have to make breaking changes in the future (unlikely now, not out of bounds).

So when we are building / extracting / using payloads from version `4.8`, we prefer to use `4.8` cli.  The easiest way to get that is from the current payload.  If you break that assumption by hardcoding, we then will have to come back later and undo it if we do change (and I don't want to sort that out then).

This change for multi-arch appears to be because the `4.8` ppc image doesn't have `4.8` x86 binaries, which is possibly wrong (it was a path we went down without fully considering all the implications, and this is an implication).  If we decide to live with it, we still need the closest possible 4.8 `oc` because 4.9 `oc` may not be able to handle the extraction (although in practice there have been no breaks in 3-4 release).  And my preferred approach there would be have ci-operator note that it did that and use a different source tag for cli images it creates when execution platform != running platform (which isolates the problem to multi-arch) by getting as close as possible (or just hardcoding, and dealing with the compatibility break only on multi-arch if it happens).  But normal path x86 can still use the cli in the payload.

alvaroaleman:
@stevekuznetsov @smarterclayton updated to only use a different CLI image for alternate archs. WDYT?

pkg/defaults/defaults.go
```diff
@@ -883,3 +889,17 @@ func getClusterPoolPullSpec(ctx context.Context, claim *api.ClusterClaim, hiveCl
...
+
+func resolveCLIOverrideImage(architecture api.ReleaseArchitecture, version string) (*coreapi.ObjectReference, error) {
+	if architecture == "" || architecture == api.ReleaseArchitectureAMD64 {
+		return nil, nil
+	}
+	if version == "" {
+		return nil, errors.New("non-amd64 releases require a version to be configured")
+	}
+	// Should never happen™ but better than a NPD below when stripping everything after the Y stream
+	if len(version) < 3 {
```
stevekuznetsov, line 901: We probably want to re-use the major/minor parsing from `processVersionChannel()` in the official release code - IIRC we can't parse e.g. `4.3` as full SemVer so that's why the regex-based stuff exists. In any case, `[:3]` breaks on e.g. `4.10`

pkg/steps/release/import_release.go
```diff
@@ -429,16 +385,117 @@ func (s *importReleaseStep) Objects() []ctrlruntimeclient.Object {
...
+			return true, nil
+
+		}); err != nil {
+			return nil, fmt.Errorf("failed to import override CLI image into %s:latest: %w", overrideCLIStreamName, err)
+		}
+		return &api.ImageStreamTagReference{Name: overrideCLIStreamName, Tag: "latest"}, nil
+	}
+
+	targetCLI := fmt.Sprintf("%s-cli", target)
+	if _, err := steps.RunPod(context.TODO(), s.client, &coreapi.Pod{
```
stevekuznetsov, line 451: nit: use the `ctx` you pass in to `getCLIImage()`

pkg/steps/release/import_release.go
```diff
@@ -429,16 +385,117 @@ func (s *importReleaseStep) Objects() []ctrlruntimeclient.Object {
...
+					Image:   fmt.Sprintf("release:%s", s.name), // the cluster will resolve this relative ref for us when we create Pods with it
+					Command: []string{"/bin/sh", "-c", "cluster-version-operator image cli > /dev/termination-log"},
+				},
+			},
+		},
+	}); err != nil {
+		return nil, fmt.Errorf("unable to find the 'cli' image in the provided release image: %w", err)
+	}
+	pod := &coreapi.Pod{}
+	if err := s.client.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: s.jobSpec.Namespace(), Name: targetCLI}, pod); err != nil {
```
stevekuznetsov, line 471: nit: use the passed `ctx`

stevekuznetsov (commented):
Mostly LGTM - we should be able to add an e2e that imports a non-amd64 release to test this

stevekuznetsov (approved):

PR #2145, opened 2021-07-06
sprint-automation: digest issues ready for acceptance

stevekuznetsov:
Currently looks like:

Screenshot from 2021-07-06 12-13-01

SVG based images for avatars hosted on Jira are not possible to display in Slack so we can't see what is a sub-task/bug/etc and users with icons anywhere but gravatar don't display, but /shrug

### DIFF
--- a/cmd/sprint-automation/main.go
+++ b/cmd/sprint-automation/main.go
@@ -1,0 +1,195 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	jiraapi "github.com/andygrunwald/go-jira"
+	"github.com/sirupsen/logrus"
+	"github.com/slack-go/slack"
+
+	"k8s.io/test-infra/pkg/flagutil"
+	"k8s.io/test-infra/prow/config/secret"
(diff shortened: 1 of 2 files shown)


stevekuznetsov:
xref https://dptp-robot-testing.slack.com/archives/C027534HBPF/p1625598728006100

cmd/sprint-automation/main.go
```diff
@@ -0,0 +1,195 @@
...
+						},
+						&slack.TextBlockObject{
+							Type: slack.MarkdownType,
+							Text: issue.Fields.Assignee.DisplayName,
+						},
+					},
+				},
+			}
+		}
+		// we really don't want these things to line wrap, so truncate the summary
```
droslean, line 140: Why? Does slack has any limit or we don't want to create big messages?

droslean (approved):

cmd/sprint-automation/main.go
stevekuznetsov: Slack has a limit to the width of an app's post, so if you use the full title it just line-wraps and makes the thing hard to read

PR #2148, opened 2021-07-07
ci-operator: don't use the root cache when building new caches

stevekuznetsov:
We label the output images we build with the versions of the inputs that
were used to build them, in order to be able to detect when an input has
changed so that we do not use a cached image in error. When we build on
top of a cache, however, we end up recording the cache as the input,
since we're built on top of it. However, we only know how to check for
input changes from the actual root. By not building on top of caches
when we're about to promote new caches, we side-step this issue.


bbguimaraes:
Port conflict:
```
         Error initializing listener of type tcp: listen tcp 127.0.0.1:40735: bind: address already in use
```
/retest

PR #2155, opened 2021-07-08
sprint-automation: post information about rotating roles

stevekuznetsov:

Example: https://dptp-robot-testing.slack.com/archives/C027534HBPF/p1625760282009600

Screenshot from 2021-07-08 09-06-42

Caveats: technically we can have really flexible schedules with people handing off stuff more than once a day (how often this runs) but in practice it's likely not an issue. I'll follow up with automation to assign people to the user groups like @dptp-helpdesk automatically and also send the intake person their digest.


stevekuznetsov:
Added a commit to message the intake person, example: 
Screenshot from 2021-07-08 10-18-35

stevekuznetsov:
Also implements https://issues.redhat.com/browse/DPTP-2356

PR #2162, opened 2021-07-12
Testimagedistributor: Log error and ignore requestst w/o cluster in app.ci

alvaroaleman:
If there is a testimagestreamtagimport on app.ci that doesn't have the
cluster set, we will currently fail with `no client for cluster  available`
which is impossible to understand without looking at the code.

Ref https://issues.redhat.com/browse/DPTP-2373

### DIFF
--- a/pkg/controller/test-images-distributor/test_images_distributor.go
+++ b/pkg/controller/test-images-distributor/test_images_distributor.go
@@ -93,7 +93,7 @@ func AddToManager(mgr manager.Manager,
 	// TODO: Watch buildCluster ImageStreams as well. For now we assume no one will tamper with them.
 	if err := c.Watch(
 		source.NewKindWithCache(&testimagestreamtagimportv1.TestImageStreamTagImport{}, mgr.GetCache()),
-		testImageStreamTagImportHandler(),
+		testImageStreamTagImportHandler(log),
 	); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -196,13 +196,17 @@ func testImageStreamTagImportHandlerForNamedCluster(clusterName string) handler.
...
-func testImageStreamTagImportHandler() handler.EventHandler {
+func testImageStreamTagImportHandler(l *logrus.Entry) handler.EventHandler {
 	return handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
 		testimagestreamtagimport, ok := o.(*testimagestreamtagimportv1.TestImageStreamTagImport)
 		if !ok {
 			logrus.WithField("type", fmt.Sprintf("%T", o)).Error("Got object that was not an ImageStream")
 			return nil
 		}
+		if testimagestreamtagimport.Spec.ClusterName == "" {
+			l.WithField("name", testimagestreamtagimport.Namespace+"/"+testimagestreamtagimport.Name).Error("found testimagestreamtagimport on app.ci that doesn't have .spec.cluster set, can not infer what cluster it is for, ignoring.")
```
hongkailiu, line 207: The card expects no errors.
Maybe Debug or Info?

PR #2163, opened 2021-07-12
Add a new tool: promoted-image-governor

hongkailiu:
https://issues.redhat.com/browse/DPTP-1766

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -298,3 +298,12 @@ $(TMPDIR)/promtool:
 	mv $(TMPDIR)/image/promtool $(TMPDIR)/promtool
 	chmod +x $(TMPDIR)/promtool
 	rm -rf $(TMPDIR)/image
+
+$(TMPDIR)/.promoted-image-governor-kubeconfig:
+	oc --context app.ci --as system:admin --namespace ci serviceaccounts create-kubeconfig promoted-image-governor > $(TMPDIR)/.promoted-image-governor-kubeconfig
+
+release_folder := $$PWD/../release
+
+promoted-image-governor: $(TMPDIR)/.promoted-image-governor-kubeconfig
+	go run  ./cmd/promoted-image-governor --kubeconfig=$(TMPDIR)/.promoted-image-governor-kubeconfig --ci-operator-config-path=$(release_folder)/ci-operator/config --release-controller-mirror-config-dir=$(release_folder)/core-services/release-controller/_releases --ignored-image-stream-tags='^ocp\S*/\S+:machine-os-content$$' --ignored-image-stream-tags='^openshift/origin-v3.11:' --dry-run=true
+.PHONY: promoted-image-governor
--- a/cmd/promoted-image-governor/main.go
+++ b/cmd/promoted-image-governor/main.go
@@ -1,0 +1,274 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/json"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/logrusutil"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+
+	imagev1 "github.com/openshift/api/image/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/steps/release"
+)
+
+type options struct {
+	ciOperatorconfigPath             string
+	kubeconfig                       string
+	dryRun                           bool
+	ignoredImageStreamTagsRaw        flagutil.Strings
+	ignoredImageStreamTags           []*regexp.Regexp
+	releaseControllerMirrorConfigDir string
+}
+
+func parseOptions() *options {
+	opts := &options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	// Controller-Runtimes root package imports the package that sets this flag
+	kubeconfigFlagDescription := "Path to the kubeconfig file to use for CLI requests."
+	if f := fs.Lookup("kubeconfig"); f != nil {
+		f.Usage = kubeconfigFlagDescription
+		// https://i.kym-cdn.com/entries/icons/original/000/018/012/this_is_fine.jpeg
+		defer func() { opts.kubeconfig = f.Value.String() }()
+	} else {
+		fs.StringVar(&opts.kubeconfig, "kubeconfig", "", kubeconfigFlagDescription)
+	}
+	fs.StringVar(&opts.ciOperatorconfigPath, "ci-operator-config-path", "", "Path to the ci operator config")
+	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.Var(&opts.ignoredImageStreamTagsRaw, "ignored-image-stream-tags", "A regex to match tag in the form of namespace/name:tag format. Can be passed multiple times.")
+	fs.StringVar(&opts.releaseControllerMirrorConfigDir, "release-controller-mirror-config-dir", "", "Path to the release controller mirror config directory")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatal("could not parse input")
+	}
+	return opts
+}
+
+func (o *options) validate() error {
+	if o.ciOperatorconfigPath == "" {
+		return fmt.Errorf("--ci-operator-config-path must be set")
+	}
+	if o.releaseControllerMirrorConfigDir == "" {
+		return fmt.Errorf("--release-controller-mirror-config-dir must be set")
+	}
+	for _, s := range o.ignoredImageStreamTagsRaw.Strings() {
+		re, err := regexp.Compile(s)
+		if err != nil {
+			return fmt.Errorf("failed to compile regex from %q: %w", s, err)
+		}
+		o.ignoredImageStreamTags = append(o.ignoredImageStreamTags, re)
+	}
+	return nil
+}
+
+func addSchemes() error {
+	if err := imagev1.AddToScheme(scheme.Scheme); err != nil {
+		return fmt.Errorf("failed to add imagev1 to scheme: %w", err)
+	}
+	return nil
+}
+
+func tagsToDelete(ctx context.Context, client ctrlruntimeclient.Client, promotedTags []api.ImageStreamTagReference, toIgnore []*regexp.Regexp, imageStreamRefs []ImageStreamRef) (map[api.ImageStreamTagReference]interface{}, error) {
+	imageStreamsWithPromotedTags := map[ctrlruntimeclient.ObjectKey]interface{}{}
+	for _, promotedTag := range promotedTags {
+		imageStreamsWithPromotedTags[ctrlruntimeclient.ObjectKey{Namespace: promotedTag.Namespace, Name: promotedTag.Name}] = nil
+	}
+
+	tagsToCheck := map[api.ImageStreamTagReference]interface{}{}
+	var errs []error
+	for objectKey := range imageStreamsWithPromotedTags {
+		imageStream := &imagev1.ImageStream{}
+		if err := client.Get(ctx, objectKey, imageStream); err != nil {
+			if !kerrors.IsNotFound(err) {
+				errs = append(errs, fmt.Errorf("could not get image stream %s in namespace %s", objectKey.Name, objectKey.Namespace))
+			} else {
+				logrus.WithField("objectKey", objectKey).Warn("image stream not found")
+			}
+			continue
+		}
+		for _, tag := range imageStream.Status.Tags {
+			tagsToCheck[api.ImageStreamTagReference{Namespace: imageStream.Namespace, Name: imageStream.Name, Tag: tag.Tag}] = nil
+		}
+	}
+
+	for _, promotedTag := range promotedTags {
+		delete(tagsToCheck, promotedTag)
+	}
+	for tag := range tagsToCheck {
+		for _, re := range toIgnore {
+			if re.MatchString(tag.ISTagName()) {
+				delete(tagsToCheck, tag)
+			}
+		}
+	}
+	mirroredTags, err := mirroredTagsByReleaseController(ctx, client, imageStreamRefs)
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to get the mirrored tags by release controller: %w", err))
+	}
+	for _, tag := range mirroredTags {
+		delete(tagsToCheck, tag)
+	}
+	if len(errs) > 0 {
+		return nil, utilerrors.NewAggregate(errs)
+	}
+	return tagsToCheck, nil
+}
+
+func mirroredTagsByReleaseController(ctx context.Context, client ctrlruntimeclient.Client, refs []ImageStreamRef) ([]api.ImageStreamTagReference, error) {
+	var ret []api.ImageStreamTagReference
+	for _, ref := range refs {
+		imageStream := &imagev1.ImageStream{}
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: "ocp", Name: ref.Name}, imageStream); err != nil {
+			return nil, fmt.Errorf("could not get image stream %s in namespace ocp", ref.Name)
+		}
+		excludedTags := sets.NewString(ref.ExcludeTags...)
+		for _, tag := range imageStream.Status.Tags {
+			if !excludedTags.Has(tag.Tag) {
+				ret = append(ret, api.ImageStreamTagReference{
+					Namespace: ref.Namespace,
+					Name:      ref.Name,
+					Tag:       tag.Tag,
+				})
+			}
+		}
+	}
+
+	return ret, nil
+}
+
+type ReleaseControllerMirrorConfig struct {
+	Publish Publish `json:"publish"`
+}
+
+type Publish struct {
+	MirrorToOrigin MirrorToOrigin `json:"mirror-to-origin"`
+}
+
+type MirrorToOrigin struct {
+	ImageStreamRef ImageStreamRef `json:"imageStreamRef"`
+}
+
+type ImageStreamRef struct {
+	Namespace   string   `json:"namespace"`
+	Name        string   `json:"name"`
+	ExcludeTags []string `json:"excludeTags,omitempty"`
+}
+
+func main() {
+	logrusutil.ComponentInit()
+
+	opts := parseOptions()
+
+	if err := opts.validate(); err != nil {
+		logrus.WithError(err).Fatal("failed to validate the option")
+	}
+
+	if err := addSchemes(); err != nil {
+		logrus.WithError(err).Fatal("failed to add schemes")
+	}
+
+	abs, err := filepath.Abs(opts.releaseControllerMirrorConfigDir)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to determine absolute release controller mirror config path")
+	}
+
+	var imageStreamRefs []ImageStreamRef
+	if err := filepath.Walk(abs,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				logrus.WithField("source-file", path).WithError(err).Error("Failed to walk release controller mirror config dir")
+				return err
+			}
+			if strings.HasSuffix(path, ".json") {
+				data, err := ioutil.ReadFile(path)
+				if err != nil {
+					logrus.WithField("source-file", path).WithError(err).Error("Failed to read file")
+					return err
+				}
+				c := &ReleaseControllerMirrorConfig{}
+				if err := json.Unmarshal(data, c); err != nil {
+					logrus.WithField("source-file", path).WithError(err).Error("Failed to unmarshal ReleaseControllerMirrorConfig")
+					return err
+				}
+				ref := c.Publish.MirrorToOrigin.ImageStreamRef
+				if ref.Namespace == "" {
+					logrus.WithField("source-file", path).Warn("publish.mirror-to-origin.imageStreamRef.namespace is empty")
+				}
+				if ref.Name == "" {
+					logrus.WithField("source-file", path).Warn("publish.mirror-to-origin.imageStreamRef.name is empty")
+				}
+				if ref.Namespace != "" && ref.Name != "" {
+					imageStreamRefs = append(imageStreamRefs, ref)
+				}
+			}
+			return nil
+		}); err != nil {
+		logrus.WithError(err).Fatal("Failed to walk release controller mirror config dir")
+	}
+
+	logrus.WithField("imageStreamRefs", imageStreamRefs).Info("Found imageStreamRefs from release controller config directory")
+
+	abs, err = filepath.Abs(opts.ciOperatorconfigPath)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to determine absolute CI Operator configuration path")
+	}
+	var promotedTags []api.ImageStreamTagReference
+	if err := config.OperateOnCIOperatorConfigDir(abs, func(cfg *api.ReleaseBuildConfiguration, metadata *config.Info) error {
+		promotedTags = append(promotedTags, release.PromotedTags(cfg)...)
+		return nil
+	}); err != nil {
+		logrus.WithField("path", abs).Fatal("failed to operate on CI Operator's config directory")
+	}
+
+	kubeConfig, err := clientcmd.BuildConfigFromFlags("", opts.kubeconfig)
+	if err != nil {
+		logrus.WithError(err).Fatalf("could not load kube config from path %s", opts.kubeconfig)
+	}
+
+	client, err := ctrlruntimeclient.New(kubeConfig, ctrlruntimeclient.Options{})
+	if err != nil {
+		logrus.WithError(err).Fatalf("could not create client")
+	}
+
+	ctx := interrupts.Context()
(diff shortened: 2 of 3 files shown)


stevekuznetsov:
Is this implementing a card, or fixing a bug, or ... ?

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+	if len(kubeConfigs) != 1 {
+		logrus.Fatalf("found %d contexts in kube config %s: must be %s only", len(kubeConfigs), opts.kubeconfig, string(api.ClusterAPPCI))
+	}
+
+	client, err := ctrlruntimeclient.New(kubeConfigs[string(api.ClusterAPPCI)], ctrlruntimeclient.Options{})
+	if err != nil {
+		logrus.WithError(err).Fatalf("could not create client")
+	}
+
+	ctx := context.TODO()
```
stevekuznetsov, line 117: :(

hack/run-promoted-image-governor.sh
```diff
@@ -0,0 +1,24 @@
...
+set -euo pipefail
+
+cd $(dirname $0)/..
+
+kubeconfig="$(mktemp)"
+
+trap 'rm -f $kubeconfig' EXIT
+kubectl config view --raw>$kubeconfig
+IFS=$'\n'
+for additional_context in $(kubectl --kubeconfig $kubeconfig config get-contexts -o name|egrep -v 'app.ci'); do
```
stevekuznetsov, line 12: Prefer to follow the examples in the `Makefile` for e.g. the `pod-scaler`. Notably:

1. use `oc` to fetch the credentials you need from the cluster, run as the service account
2. pass through flags as needed

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+	if err := addSchemes(); err != nil {
+		logrus.WithError(err).Fatal("failed to add schemes")
+	}
+
+	abs, err := filepath.Abs(opts.ciOperatorconfigPath)
+	if err != nil {
+		logrus.WithError(err).WithField("opts.ciOperatorconfigPath", opts.ciOperatorconfigPath).Fatal("failed to determine absolute filepath")
+	}
+	promotedTags := sets.NewString()
+	if err := config.OperateOnCIOperatorConfigDir(abs, func(cfg *api.ReleaseBuildConfiguration, metadata *config.Info) error {
```
stevekuznetsov, line 97: This needs to be a dynamic load, not static at process start.

hongkailiu: I thought we want to run this cli as a periodic job, not as a long running deployment.

stevekuznetsov (commented):
This should not be a controller. From my comment on Jira:

> a controller is overkill here since the important edges/source of truth we want to react to are the `openshift/release` promotion configurations, and we can authoritatively derive intended state from that repo. I do not think a useful solution to this problem has to react to `ImageStreamTag` events in the clusters, since we do not have rogue authors changing cluster state. It would be sufficient to write a postsubmit and periodic to `openshift/release` to react to config changes.

cmd/promoted-image-governor/main.go
stevekuznetsov: Ah ack, sorry - thought this was using controller-runtime to build a controller

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+}
+
+func (o *options) validate() error {
+	if o.ciOperatorconfigPath == "" {
+		return fmt.Errorf("--ci-operator-config-path must be set")
+	}
+	return nil
+}
+
+func tagsToPromote(cfg *api.ReleaseBuildConfiguration) sets.String {
```
stevekuznetsov, line 61: If a function is used only once, it should not be a function unless there's a lot to be gained by giving it a name and hiding it

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+	tags := sets.NewString()
+	for _, tag := range release.PromotedTags(cfg) {
+		tags.Insert(fmt.Sprintf("%s/%s:%s", tag.Namespace, tag.Name, tag.Tag))
```
stevekuznetsov, line 64: Use `tas.ISTagName()`

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+	if f := fs.Lookup("kubeconfig"); f != nil {
+		f.Usage = kubeconfigFlagDescription
+		// https://i.kym-cdn.com/entries/icons/original/000/018/012/this_is_fine.jpeg
+		defer func() { opts.kubeconfig = f.Value.String() }()
+	} else {
+		fs.StringVar(&opts.kubeconfig, "kubeconfig", "", kubeconfigFlagDescription)
+	}
+	fs.StringVar(&opts.ciOperatorconfigPath, "ci-operator-config-path", "", "Path to the ci operator config")
+	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.Var(&opts.ignoredImageStreamTagsRaw, "ignored-image-stream-tags", "An imagestreamtag to ignore while checking. It must be in namespace/name:tag format (e.G `ci/clonerefs:latest`). Can be passed multiple times.")
```
stevekuznetsov, line 47: Why?

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+		logrus.WithError(err).Fatal("failed to validate the option")
+	}
+
+	if err := addSchemes(); err != nil {
+		logrus.WithError(err).Fatal("failed to add schemes")
+	}
+
+	abs, err := filepath.Abs(opts.ciOperatorconfigPath)
+	if err != nil {
+		logrus.WithError(err).WithField("opts.ciOperatorconfigPath", opts.ciOperatorconfigPath).Fatal("failed to determine absolute filepath")
```
stevekuznetsov, line 94: I do not think we should bring code (`opts.ciOperatorconfigPath` into logs). A succinct message will more than suffice:

```go
logrus.WithError(err).Fatal("failed to determine absolute CI Operator configuration path")
```

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+
+	kubeConfigs, _, err := util.LoadKubeConfigs(opts.kubeconfig, nil)
+	if err != nil {
+		logrus.Fatalf("could not load kube config from path %s: %w", opts.kubeconfig, err)
+	}
+	if len(kubeConfigs) != 1 {
+		logrus.Fatalf("found %d contexts in kube config %s: must be %s only", len(kubeConfigs), opts.kubeconfig, string(api.ClusterAPPCI))
+	}
+
+	client, err := ctrlruntimeclient.New(kubeConfigs[string(api.ClusterAPPCI)], ctrlruntimeclient.Options{})
```
stevekuznetsov, line 112: We should not pass this tool a kubeconfig with multiple contexts, and we should not load multiple contexts only to pick out one specific one. There's no need for the context to have a specific name.

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+	for _, bc := range buildConfigs.Items {
+		if bc.Spec.Output.To != nil {
+			buildConfigOutputTags.Insert(fmt.Sprintf("%s/%s", bc.Spec.Output.To.Namespace, bc.Spec.Output.To.Name))
+		}
+	}
+	tagsToCheck := sets.NewString()
+	var errs []error
+	imageStreamsWithPromotedTags := sets.NewString()
+	for _, promotedTag := range promotedTags.List() {
+		namespacedImageStreamAndTagName := strings.Split(promotedTag, ":")
```
stevekuznetsov, line 132: Why do we collect `ImageStreamTagReference` objects, which have fields, only to string-format them into dimensionless strings, only to split them back into the fields we started with? Just use a set of `ImageStreamTagReference`s. Or a list, since we have other validation that ensures no two configurations promote the same tag.

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+
+	ctx := context.TODO()
+	buildConfigs := &buildv1.BuildConfigList{}
+	if err := client.List(ctx, buildConfigs); err != nil {
+		logrus.WithError(err).Fatalf("could not list buildConfigs")
+	}
+	buildConfigOutputTags := sets.NewString()
+	for _, bc := range buildConfigs.Items {
+		if bc.Spec.Output.To != nil {
+			buildConfigOutputTags.Insert(fmt.Sprintf("%s/%s", bc.Spec.Output.To.Namespace, bc.Spec.Output.To.Name))
```
stevekuznetsov, line 125: What is this?

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	intersection := promotedTags.Intersection(buildConfigOutputTags)
+	if intersection.Len() > 0 {
+		errs = append(errs, fmt.Errorf("found build config overwrite promoted tags: %s", intersection.List()))
+	}
+
+	tagsToDelete := sets.NewString()
+	for _, tag := range tagsToCheck.List() {
+		if buildConfigOutputTags.Has(tag) {
```
stevekuznetsov, line 167: This does not seem like a feature we need.

cmd/promoted-image-governor/main.go
hongkailiu: Have not used it yet.
It came from the last AC in https://issues.redhat.com/browse/DPTP-2344

cmd/promoted-image-governor/main.go
hongkailiu: I need a context for client calls but I do not know what to use.
I did remember you complained of the debt it creates but I do not know what was the solution.

hack/run-promoted-image-governor.sh
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: Removed

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: Use a list. Also added a function to avoid promotion conflict.

cmd/promoted-image-governor/main.go
hongkailiu: Removed.

cmd/promoted-image-governor/main.go
hongkailiu: Removed.

Originally it came from https://issues.redhat.com/browse/DPTP-2344?focusedCommentId=16366770&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16366770

```
 If they want an image to exist, it must have a BuildConfig or ImageStream in the openshift/release repo, or be built and promoted by a job.
```

hongkailiu:
https://github.com/openshift/release/pull/20188
shows the result.
Only very few tags are promoted.

cmd/promoted-image-governor/main.go
stevekuznetsov: `interrupts.Context()` works, or `context.Background()` if you're ok with the process getting SIGKILL

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+		defer func() { opts.kubeconfig = f.Value.String() }()
+	} else {
+		fs.StringVar(&opts.kubeconfig, "kubeconfig", "", kubeconfigFlagDescription)
+	}
+	fs.StringVar(&opts.ciOperatorconfigPath, "ci-operator-config-path", "", "Path to the ci operator config")
+	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.Var(&opts.ignoredImageStreamTagsRaw, "ignored-image-stream-tags", "An imagestreamtag to ignore while checking. It must be in namespace/name:tag format (e.G `ci/clonerefs:latest`). Can be passed multiple times.")
+	fs.StringVar(&opts.openshiftMappingDir, "openshift-mapping-dir", "", "Path to the openshift mapping directory")
+	fs.StringVar(&opts.openshiftMappingConfigPath, "openshift-mapping-config", "", "Path to the openshift mapping config file")
+	if err := fs.Parse(os.Args[1:]); err != nil {
```
stevekuznetsov, line 81: parsing, loading, etc, does not belong in the constructor for the options

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+		if _, ok := tagCount[tag]; ok {
+			tagCount[tag] = tagCount[tag] + 1
+		} else {
+			tagCount[tag] = 1
+		}
+	}
+	dups := sets.NewString()
+	for tag, c := range tagCount {
+		if c > 1 {
+			dups.Insert(fmt.Sprintf("%s/%s:%s", tag.Namespace, tag.Name, tag.Tag))
```
stevekuznetsov, line 124: Use the method for this, instead of formatting it yourself.

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+func validatePromotedTags(tags []api.ImageStreamTagReference) error {
+	tagCount := map[api.ImageStreamTagReference]int{}
+	for _, tag := range tags {
+		if _, ok := tagCount[tag]; ok {
+			tagCount[tag] = tagCount[tag] + 1
+		} else {
+			tagCount[tag] = 1
+		}
+	}
+	dups := sets.NewString()
```
stevekuznetsov, line 121: Tags are keys in a map - they must be unique. No need for a set.

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+	for version := range mappingConfig.Images {
+		mappings[fmt.Sprintf("mapping_origin_%s", strings.ReplaceAll(version, ".", "_"))] = map[string][]string{}
+	}
+	for _, tag := range promotedTags {
+		if tag.Namespace == mappingConfig.SourceNamespace {
+			if mappingConfig.Images != nil {
+				if targetTags, ok := mappingConfig.Images[tag.Name]; ok {
+					for _, targetTag := range targetTags {
+						filename := fmt.Sprintf("mapping_origin_%s", strings.ReplaceAll(tag.Name, ".", "_"))
+						src := fmt.Sprintf("%s/%s/%s:%s", mappingConfig.SourceRegistry, mappingConfig.SourceNamespace, tag.Name, tag.Tag)
```
stevekuznetsov, line 245: again, use the tag's method to format it

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+	if dups.Len() > 0 {
+		return fmt.Errorf("found tags promoted more than once: %s", dups.List())
+	}
+	return nil
+}
+
+func tagsToDelete(ctx context.Context, client ctrlruntimeclient.Client, promotedTags []api.ImageStreamTagReference, toIgnore sets.String) (sets.String, error) {
+	tagsToCheck := sets.NewString()
```
stevekuznetsov, line 134: We do not want/need to stringify everything. A set is a map where we ignore keys, so a set of tags is

```go
map[api.ImageStreamTagReference]interface{}
```

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+		imageStreamsWithPromotedTags.Insert(namespacedImageStream)
+		imagestream := &imagev1.ImageStream{}
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: promotedTag.Namespace, Name: promotedTag.Name}, imagestream); err != nil {
+			if !kerrors.IsNotFound(err) {
+				errs = append(errs, fmt.Errorf("could not get image steam %s in namespace %s", promotedTag.Name, promotedTag.Namespace))
+			}
+			continue
+		}
+		for _, tag := range imagestream.Status.Tags {
+			tagsToCheck.Insert(fmt.Sprintf("%s/%s:%s", promotedTag.Namespace, promotedTag.Name, tag.Tag))
```
stevekuznetsov, line 152: for sanity, use the imagestream's name and namespace for populating this

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+	}
+
+	for _, promotedTag := range promotedTags {
```
stevekuznetsov, line 156: the only convenience the sets would give you is e.g. Difference() - not used :)

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+	client, err := ctrlruntimeclient.New(kubeConfig, ctrlruntimeclient.Options{})
+	if err != nil {
+		logrus.WithError(err).Fatalf("could not create client")
+	}
+
+	if err := validatePromotedTags(promotedTags); err != nil {
+		logrus.WithError(err).Fatalf("could not validate promoted tags")
+	}
+
+	toDelete, err := tagsToDelete(context.TODO(), client, promotedTags, opts.ignoredImageStreamTags)
```
stevekuznetsov, line 207: think about what you would do with `toDelete` when you actually implement deletion - you will need to split it up into name/namespace/tag anyway, so let's use the struct instead of stringifying it

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+}
+
+func generateMappings(promotedTags []api.ImageStreamTagReference, mappingConfig *OpenshiftMappingConfig) map[string]map[string][]string {
+	mappings := map[string]map[string][]string{}
+	for version := range mappingConfig.Images {
+		mappings[fmt.Sprintf("mapping_origin_%s", strings.ReplaceAll(version, ".", "_"))] = map[string][]string{}
+	}
+	for _, tag := range promotedTags {
+		if tag.Namespace == mappingConfig.SourceNamespace {
```
stevekuznetsov, line 240: This is what causes your tool to delete all the mapping configs. Let's focus on the ImageStreamTags for now.

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	return nil
+}
+
+func tagsToDelete(ctx context.Context, client ctrlruntimeclient.Client, promotedTags []api.ImageStreamTagReference, toIgnore sets.String) (sets.String, error) {
+	tagsToCheck := sets.NewString()
+	var errs []error
+	imageStreamsWithPromotedTags := sets.NewString()
+
+	for _, promotedTag := range promotedTags {
```
stevekuznetsov, line 138: I think you want a slightly different approach:

1. filter the set of promoted tags to a set of imagestreams into which at least one tag is promoted
2. get all of those from the cluster, record their tags
3. take the difference between promoted tags and tags on streams

cmd/promoted-image-governor/main.go
hongkailiu: Renamed the function name. Move the validation logic to its own function.

cmd/promoted-image-governor/main.go
hongkailiu: Sorry I forgot it again.

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: Done.

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: > promotedTags []api.ImageStreamTagReference
it is a slice now.

cmd/promoted-image-governor/main.go
hongkailiu: Removed the logic of the mappings.

cmd/promoted-image-governor/main.go
hongkailiu: Also ignore the tags mirrored from release controller configs.

hongkailiu:
With current ignore
```
  --ignored-image-stream-tags='^ocp\S*/\S+:machine-os-content$' \
  --ignored-image-stream-tags='^openshift/origin-v3.11:' \
```

stats on deleted tags.
[istag.to.delete.json.zip](https://github.com/openshift/ci-tools/files/6850332/istag.to.delete.json.zip)

```
cat istag.to.delete.json| jq -r '.toDelete[]|.namespace' | sort  | uniq -c | sort -k1 -n
   2 kiegroup
   2 ocp-kni
   4 ci
   5 cincinnati-ci
  12 build-cache
  17 open-cluster-management
 132 ocp-private
 196 ocp
 406 origin
 538 openshift
```

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,314 @@
...
+}
+
+func addSchemes() error {
+	if err := imagev1.AddToScheme(scheme.Scheme); err != nil {
+		return fmt.Errorf("failed to add imagev1 to scheme: %w", err)
+	}
+	return nil
+}
+
+func validatePromotedTags(tags []api.ImageStreamTagReference) error {
```
stevekuznetsov, line 90: You could simplify this:

```go
tags := map[api.ImageStreamTagReference]interface{}{}
var duplicates []string
for _, tag := tags {
    if _, exists := tags[tag]; exists {
        duplicates = append(duplicates, tag.ISTagName())
    } else {
        tags[tag] = nil
    }
}
if len(duplicates) > 0 {
    return fmt.Errorf("found tags promoted more than once: %s", dups)
}
return nil

Comment (stevekuznetsov) [reply cmd/promoted-image-governor/main.go:90]: However, we already do this validation in `ci-operator-checkconfig` and I don't know that we need to duplicate it here ...

File: cmd/promoted-image-governor/main.go [hunk 23]
```diff
@@ -0,0 +1,314 @@
...
+	for _, promotedTag := range promotedTags {
+		imageStreamsWithPromotedTags[ctrlruntimeclient.ObjectKey{Namespace: promotedTag.Namespace, Name: promotedTag.Name}] = nil
+	}
+
+	tagsToCheck := map[api.ImageStreamTagReference]interface{}{}
+	var errs []error
+	for objectKey := range imageStreamsWithPromotedTags {
+		imageStream := &imagev1.ImageStream{}
+		if err := client.Get(ctx, objectKey, imageStream); err != nil {
+			if !kerrors.IsNotFound(err) {
```
stevekuznetsov, line 122: Why do we want to special-case this? Streams that get promoted to should exist

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,314 @@
...
+	if err != nil {
+		errs = append(errs, fmt.Errorf("failed to get the mirrored tags by release controller: %w", err))
+	}
+	for _, tag := range mirroredTags {
+		delete(tagsToCheck, tag)
+	}
+	if len(errs) > 0 {
+		return nil, utilerrors.NewAggregate(errs)
+	}
+	var tagsToDelete []api.ImageStreamTagReference
```
stevekuznetsov, line 152: What is the point of copying the data from `tagsToCheck` into this when they are equal?

cmd/promoted-image-governor/main.go
```diff
@@ -0,0 +1,314 @@
...
+	}
+
+	var imageStreamRefs []ImageStreamRef
+	if err := filepath.Walk(abs,
+		func(path string, info os.FileInfo, err error) error {
+			if err != nil {
+				logrus.WithField("source-file", path).WithError(err).Error("Failed to walk release controller mirror config dir")
+				return err
+			}
+			if strings.HasSuffix(path, "-ci.json") {
```
stevekuznetsov, line 236: Why only CI configs? Do the others have mirroring options? Would it be wrong to read all mirroring options?

cmd/promoted-image-governor/main.go
hongkailiu: Removed.

cmd/promoted-image-governor/main.go
hongkailiu: We do have those cases when I run the tool locally.
Added logs for the case.

cmd/promoted-image-governor/main.go
hongkailiu: Removed.
Returned the map instead.

cmd/promoted-image-governor/main.go
hongkailiu: I checked. All the mirroring options are in the `*-ci.json` files.
But yeah, changed to 

```
if strings.HasSuffix(path, ".json") {
```

hack/run-promoted-image-governor.sh
```diff
@@ -0,0 +1,21 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+cd $(dirname $0)/..
+
+kubeconfig="$(mktemp)"
+
+trap 'rm -f $kubeconfig' EXIT
+oc --context app.ci --as system:admin --namespace ci serviceaccounts create-kubeconfig promoted-image-governor >$kubeconfig
```
stevekuznetsov, line 10: IMO we can have this as a make target like the other ones for e.g. local-e2e. no need to make people re-download this every time

hongkailiu: Updated.

Makefile
```diff
@@ -297,4 +297,15 @@ $(TMPDIR)/promtool:
...
 	chmod +x $(TMPDIR)/promtool
-	rm -rf $(TMPDIR)/image
\ No newline at end of file
+	rm -rf $(TMPDIR)/image
+
+$(TMPDIR)/.promoted-image-governor-kubeconfig:
+	oc --context app.ci --as system:admin --namespace ci serviceaccounts create-kubeconfig promoted-image-governor > $(TMPDIR)/.promoted-image-governor-kubeconfig
+
+go_path := $(shell go env GOPATH)
+release_folder := $(go_path)/src/github.com/openshift/release
```
stevekuznetsov, line 306: I don't think we need to know this? You can just make a path that's relative to this Makefile - look at the `local-pod-scaler` target

hongkailiu: Updated.

stevekuznetsov (approved):

PR #2164, opened 2021-07-13
CI Operator: Create namespaces that are promoted to

alvaroaleman:
Currently we do not have anything that ensures namespaces we promote to
actually exist. This change adds a controller that does that.

The controller approach was taken over extending ci-operator, because
doing that would require to plug through an app.ci kubeconfig and give
it access to create namespaces.

Ref https://issues.redhat.com/browse/DPTP-2359


stevekuznetsov:
Doesn't `ci-operator` already have the app.ci credentials to do the push? I feel like an API call and a change to RBAC is a little easier to reason about as compared to a (disconnected, harder to discover) controller, no?

alvaroaleman:
> Doesn't ci-operator already have the app.ci credentials to do the push? I feel like an API call and a change to RBAC is a little easier to reason about as compared to a (disconnected, harder to discover) controller, no?

It has a dockerconfig, it doesn't have a kubeconfig: https://github.com/openshift/ci-tools/blob/e7971e7190c9daf859fbac2df62b1a9c5622f348/pkg/steps/release/promote.go#L139

Thus we'd need to:
* Generate and propagate the kubeconfig
* Make ci-operator propagete it into the test namespaces
* Mount it into the promotion pod and use it

stevekuznetsov:
Sure, but a `dockercfg` is identical to the `kubeconfig` - not only are they literally the same token but if we have the `kubeconfig` we can trivially get our own `dockercfg` dynamically as well, so one input would allow us to do everything we need to do for promotion in one spot

alvaroaleman:
> Sure, but a dockercfg is identical to the kubeconfig - not only are they literally the same token but if we have the kubeconfig we can trivially get our own dockercfg dynamically as well, so one input would allow us to do everything we need to do for promotion in one spot

Not quite, the token might be identical, but the format is not.

stevekuznetsov:
Right, but if we have the token we can get the `dockercfg` - idk. I think the mental overhead of another disjoint controller is pretty high. We already have 100 components, why do we think it's worth it to add to that? Imagine debugging this - instead of a log for a job we would now need access to app.ci, the logs for a specific controller, etc.

alvaroaleman:
> Right, but if we have the token we can get the dockercfg - idk. I think the mental overhead of another disjoint controller is pretty high. We already have 100 components, why do we think it's worth it to add to that? Imagine debugging this - instead of a log for a job we would now need access to app.ci, the logs for a specific controller, etc.

Sure, I just think that is something that is not going to happen like that. The controller will attempt to create the namespaces way before any job ever uses them (as soon as they appear in the config, today there are a bunch already that do not exist), so as long as we regularly check on the error log, we will notice issues here before users do. Furthermore your suggestion implies to add a bunch of additional api calls into the success path of all promotion jobs, which is another step that can fail instead of simply making sure that the env is as the jobs need it, which is what this PR does.

petr-muller:
My opinion: I would probably prefer Steve's solution if code wouldn't be already written (mostly because of the mental overhead any single additional specialized component adds when reasoning about the whole system), but I dislike asking people to throw away things they already built (unless they are really broken or really wrong, which is not this case, I think Alvaro just chose a legit different approach), so if I came here first, I'd probably just review the code, perhaps ask why it's a controller instead of ci-op-side code, then be satisfied with pretty much anything that's a sound reason ("credentials" is one to me).

Maybe if we go the controller way, we can simplify the system by putting whole promotion responsiblity to the controller, `ci-op` would just somehow signal "I'm done, pull the images".

pkg/controller/promotion_namespace_reconciler/promotion_namespace_reconciler.go
```diff
@@ -0,0 +1,115 @@
...
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, err
+}
+
+func (r *reconciler) reconcile(ctx context.Context, l *logrus.Entry, req reconcile.Request) error {
+	l.Debug("Starting reconciliation")
+	var ns corev1.Namespace
+	if err := r.client.Get(ctx, req.NamespacedName, &ns); err != nil {
```
stevekuznetsov, line 99: Why not just create and ignore AlreadyExists, making 50% fewer calls?

stevekuznetsov (commented):
Petr's argument works for me. One comment.

pkg/controller/promotion_namespace_reconciler/promotion_namespace_reconciler.go
alvaroaleman: Because the Get is not an api call, it is just copying data from cache, so this reduces the number of api calls we do drastically

pkg/steps/release/promote_test.go
```diff
@@ -578,8 +578,8 @@ func TestGetImageMirror(t *testing.T) {
 
 	for _, testCase := range testCases {
 		t.Run(testCase.name, func(t *testing.T) {
-			if actual, expected := getImageMirrorTarget(testCase.tags, testCase.pipeline, "registry.ci.openshift.org"), testCase.expected; !reflect.DeepEqual(actual, expected) {
-				t.Errorf("%s: got incorrect ImageMirror mapping: %v", testCase.name, diff.ObjectDiff(actual, expected))
+			if actual, _ := getImageMirrorTarget(testCase.tags, testCase.pipeline, "registry.ci.openshift.org"); !reflect.DeepEqual(actual, testCase.expected) {
```
stevekuznetsov, line 581: Why?

pkg/steps/release/promote.go
```diff
@@ -64,18 +70,58 @@ func (s *promotionStep) run(ctx context.Context) error {
...
+	if err := json.Unmarshal(s.pushSecret.Data[coreapi.DockerConfigJsonKey], &dockercfg); err != nil {
+		return fmt.Errorf("failed to deserialize push secret: %w", err)
+	}
+
+	appCIDockercfg, hasAppCIDockercfg := dockercfg.Auths[api.ServiceDomainAPPCIRegistry]
+	if !hasAppCIDockercfg {
+		return fmt.Errorf("push secret has no entry for %s", api.ServiceDomainAPPCIRegistry)
+	}
+
+	appCIKubeconfig := &rest.Config{Host: "https://api.ci.l2s4.p1.openshiftapps.com:6443", BearerToken: appCIDockercfg.Password}
```
stevekuznetsov, line 100: Do we have this as a constant somewhere?

pkg/steps/release/promote_test.go
alvaroaleman: Bcause `getImageMirrorTarget` now has two return values, and golang doesn't allow sth like

```
if a, b, c := funcWithTworeturnvals, c; {...
```

pkg/steps/release/promote.go
alvaroaleman: Nope

pkg/steps/release/promote.go
```diff
@@ -64,18 +70,58 @@ func (s *promotionStep) run(ctx context.Context) error {
...
 
-	imageMirrorTarget := getImageMirrorTarget(tags, pipeline, registryDomain(s.configuration.PromotionConfiguration))
+	imageMirrorTarget, namespaces := getImageMirrorTarget(tags, pipeline, registryDomain(s.configuration.PromotionConfiguration))
 	if len(imageMirrorTarget) == 0 {
 		logrus.Info("Nothing to promote, skipping...")
 		return nil
 	}
 
+	if err := s.ensureNamespaces(ctx, namespaces); err != nil {
+		return fmt.Errorf("failed to ensure namespaces to promote to in central rewgistry: %w", err)
```
emilvberglind, line 80: nit: registry

alvaroaleman:
PTAL

pkg/steps/release/promote_test.go
stevekuznetsov: Sure, but what was wrong about the previous code ... ?

pkg/steps/release/promote.go
stevekuznetsov: Can we add one?

pkg/steps/release/promote_test.go
alvaroaleman: It doesn't compile because `getImageMirrorTarget` now has two return values

pkg/steps/release/promote.go
alvaroaleman: Done

PR #2171, opened 2021-07-14
ci-operator: expose hive conditions on failure

stevekuznetsov:
We can't DRY out this logic since the conditions are unique structs with
the same fields instead of any construct that shares the definition.

/assign @hongkailiu

### DIFF
--- a/pkg/steps/cluster_claim.go
+++ b/pkg/steps/cluster_claim.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
@@ -97,13 +98,13 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
 		// always attempt to delete claim if one exists
 		var releaseErr error
 		if clusterClaim != nil {
-			releaseErr = results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim))
+			releaseErr = results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim, true))
 		}
 		return aggregateWrappedErrorAndReleaseError(acquireErr, releaseErr)
 	}
 
 	wrappedErr := results.ForReason("executing_test").ForError(s.wrapped.Run(ctx))
-	releaseErr := results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim))
+	releaseErr := results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim, false))
 
 	return aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr)
 }
@@ -190,9 +191,9 @@ func getHiveSecret(src *corev1.Secret, name, namespace, testName string) (*corev
 	}, nil
 }
 
-func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hivev1.ClusterClaim) error {
+func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hivev1.ClusterClaim, printConditions bool) error {
 	logger := logrus.WithField("clusterClaim.Namespace", clusterClaim.Namespace).WithField("clusterClaim.Name", clusterClaim.Name)
-	if err := s.saveArtifacts(ctx, clusterClaim.Namespace, clusterClaim.Name); err != nil {
+	if err := s.saveArtifacts(ctx, clusterClaim.Namespace, clusterClaim.Name, printConditions); err != nil {
 		// logging the error without failing the test
 		logger.Error("Failed to save artifacts before releasing the claimed cluster")
 	}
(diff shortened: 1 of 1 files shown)


pkg/steps/cluster_claim.go
```diff
@@ -212,31 +213,60 @@ func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hiv
...
-			&hivev1.ClusterDeployment{},
+			&clusterDeployment,
 			"cluster deployment",
 			filepath.Join(namespaceDir, "clusterDeployment.json"),
 		); err != nil {
 			errs = append(errs, err)
 		}
+		if printConditions {
+			builder := &strings.Builder{}
+			_, _ = builder.WriteString(fmt.Sprintf("Found %d conditions for ClusterDeployment:", len(claim.Status.Conditions)))
```
petr-muller, line 258:
```suggestion
			_, _ = builder.WriteString(fmt.Sprintf("Found %d conditions for ClusterDeployment:", len(clusterDeployment.Status.Conditions)))
```

(noticed via tests)

pkg/steps/cluster_claim.go
```diff
@@ -212,31 +213,60 @@ func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hiv
...
+		builder := &strings.Builder{}
+		_, _ = builder.WriteString(fmt.Sprintf("Found %d conditions for ClusterClaim:", len(claim.Status.Conditions)))
+		sort.Slice(claim.Status.Conditions, func(i, j int) bool {
+			return claim.Status.Conditions[i].LastTransitionTime.Before(&claim.Status.Conditions[j].LastTransitionTime)
+		})
+		for _, condition := range claim.Status.Conditions {
+			if condition.Status == corev1.ConditionUnknown || condition.Reason == "Initialized" {
+				continue
+			}
+			_, _ = builder.WriteString(fmt.Sprintf("\n[%s]%s: %s", condition.LastTransitionTime.Format(time.RFC3339), condition.Reason, condition.Message))
```
petr-muller, line 240:
```suggestion
			_, _ = builder.WriteString(fmt.Sprintf("\n[%s] %s: %s", condition.LastTransitionTime.Format(time.RFC3339), condition.Reason, condition.Message))
```

add whitespace for readability

petr-muller:
Also e2e tests need to be adjusted for new output

stevekuznetsov:
I think the e2e was a flake?

stevekuznetsov:
@petr-muller sorry about that - should be good now

pkg/steps/cluster_claim.go
```diff
@@ -212,31 +213,60 @@ func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hiv
...
+			&clusterDeployment,
 			"cluster deployment",
 			filepath.Join(namespaceDir, "clusterDeployment.json"),
 		); err != nil {
 			errs = append(errs, err)
 		}
+		if printConditions {
+			builder := &strings.Builder{}
+			_, _ = builder.WriteString(fmt.Sprintf("Found %d conditions for ClusterDeployment:", len(clusterDeployment.Status.Conditions)))
+			sort.Slice(clusterDeployment.Status.Conditions, func(i, j int) bool {
```
hongkailiu, line 259: NIT: SliceStable. Not sure if the origin order (when 2 conditions have the same transition time) has some meaning though.
```suggestion
			sort.SliceStable(clusterDeployment.Status.Conditions, func(i, j int) bool {
```

pkg/steps/cluster_claim.go
```diff
@@ -212,31 +213,60 @@ func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hiv
...
-			&hivev1.ClusterDeployment{},
+			&clusterDeployment,
 			"cluster deployment",
 			filepath.Join(namespaceDir, "clusterDeployment.json"),
 		); err != nil {
 			errs = append(errs, err)
 		}
+		if printConditions {
+			builder := &strings.Builder{}
+			_, _ = builder.WriteString(fmt.Sprintf("Found %d conditions for ClusterDeployment:", len(clusterDeployment.Status.Conditions)))
```
hongkailiu, line 258: Why do we hide the error there?
I have seen those in other saveArtifacts related functions. Like, 
https://github.com/openshift/ci-tools/blob/55b9360b568894384c9fe705379bd90a528c5b78/cmd/ci-operator/main.go#L832

hongkailiu (commented):
We have saved the 2 objects as artifacts, failure or success.
Why do we want to expose their conditions in the log?
Better UX than accessing them in the artifact folder?

/lgtm

/hold
in case you want to deal with the NIT.

pkg/steps/cluster_claim.go
stevekuznetsov: We're just printing, it does not really matter that this is stable

pkg/steps/cluster_claim.go
stevekuznetsov: This is just best-effort

stevekuznetsov:
We don't need a stable sort for human-readable output, we are not asserting on the output format.

/hold cancel

PR #2097, opened 2021-06-23
*: Add ClusterProfileAWSArm64

LorbusChris:
/cc @jcpowermac

PR #2175, opened 2021-07-19
azurestack: Add azurestack cluster profile for CI tests

rna-afk:
Adding azurestack cluster profile for e2e-azurestack tests in CI.

### DIFF
--- a/cmd/repo-init/main.go
+++ b/cmd/repo-init/main.go
@@ -640,6 +640,8 @@ func determineWorkflowFromClusterPorfile(clusterProfile api.ClusterProfile) *str
 		ret = "ipi-aws"
 	case api.ClusterProfileAzure:
 		ret = "ipi-azure"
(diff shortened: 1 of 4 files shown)


patrickdillon:
I'm not very familiar with the cluster profiles but cargo culting off of #2097 it looks like we also need an entry here:
https://github.com/openshift/ci-tools/blob/master/cmd/repo-init/main.go#L634

And that would tie in with the steps configuration

patrickdillon:
For context, Azure Stack is an on-prem variant of Azure. So for CI we will have an azure stack environment separate from our public azure environment.

petr-muller:
@rna-afk looks liike you amended your changes to a merge commit, can you please clean it up?

rna-afk:
My bad. Fixed it thanks.

petr-muller (approved):

PR #2180, opened 2021-07-20
Vault proxy: Validate key conflicts

alvaroaleman:
This change makes the vault proxy check for conflicts and error if one is found.

Sample in the UI:

vault-kv-conflict

Ref https://issues.redhat.com/browse/DPTP-2206

### DIFF
--- a/cmd/vault-subpath-proxy/kv_update_transport.go
+++ b/cmd/vault-subpath-proxy/kv_update_transport.go
@@ -9,12 +9,17 @@ import (
 	"net/http"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/sync/semaphore"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/client-go/util/retry"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -22,6 +27,7 @@ import (
 
 	"github.com/openshift/ci-tools/pkg/api/vault"
 	"github.com/openshift/ci-tools/pkg/steps"
+	"github.com/openshift/ci-tools/pkg/vaultclient"
 )
 
 // https://github.com/openshift/ci-tools/blob/7af2e075f381ecae1562d1406bad2c86a23e72a3/vendor/k8s.io/api/core/v1/types.go#L5748-L5749
@@ -36,6 +42,16 @@ type kvUpdateTransport struct {
 	// If enabled, the roundtripper will wait for secret
 	// sync to complete. Should only be enabled in tests.
 	synchronousSecretSync bool
+
+	privilegedVaultClient                 *vaultclient.VaultClient
+	existingSecretKeysByNamespaceName     map[types.NamespacedName]sets.String
+	existingSecretKeysByNamespaceNameLock sync.Mutex
+	existingSecretKeysByVaultSecretName   map[string][]namespacedNameKey
+}
+
+type namespacedNameKey struct {
+	name types.NamespacedName
+	key  string
 }
 
 func (k *kvUpdateTransport) RoundTrip(r *http.Request) (*http.Response, error) {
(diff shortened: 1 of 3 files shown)


cmd/vault-subpath-proxy/kv_update_transport.go
```diff
@@ -36,6 +42,16 @@ type kvUpdateTransport struct {
 	// If enabled, the roundtripper will wait for secret
 	// sync to complete. Should only be enabled in tests.
 	synchronousSecretSync bool
+
+	privilegedVaultClient                 *vaultclient.VaultClient
+	existingSecretKeysByNamespaceName     map[types.NamespacedName]sets.String
+	existingSecretKeysByNamespaceNameLock sync.Mutex
+	existingSecretKeysByVaultSecretName   map[string][]namespacedNameKey
```
alvaroaleman, line 49: The duplicate data exists because:

* For validation of requests we need to index by Kubernetes secret
* But on update/delete we need to know all values by vault path - We could fetch it but that adds another api call and makes the code more complicated, because we have to handle the case that the user request fails and we have to revert our cache


Hypothetically we could store this in one data structure, but that would require to iterate everything for either validation or cache update.

cmd/vault-subpath-proxy/kv_update_transport.go
```diff
@@ -96,9 +119,124 @@ func (k *kvUpdateTransport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+	}
+}
+
+func (k *kvUpdateTransport) validateKeysDontConflict(ctx context.Context, data map[string]string) (validationErrs []string, err error) {
+	if k.privilegedVaultClient == nil {
+		return nil, nil
+	}
+	k.existingSecretKeysByNamespaceNameLock.Lock()
+	defer k.existingSecretKeysByNamespaceNameLock.Unlock()
+	if k.existingSecretKeysByNamespaceName == nil {
```
bbguimaraes, line 184: This whole `if` looks like a good delimitation for an independent function.

bbguimaraes: Do we have an idea of how long this initialization takes?  Can it make the first request time out?  Would it make sense to do it (synchronously or not) at startup?

cmd/vault-subpath-proxy/kv_update_transport.go
```diff
@@ -96,9 +119,124 @@ func (k *kvUpdateTransport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+				}
+			}()
+		}
+		wg.Wait()
+		if err := utilerrors.NewAggregate(fetchErrs); err != nil {
+			return nil, fmt.Errorf("failed to fetch secrets from vault: %w", err)
+		}
+	}
+
+	name := types.NamespacedName{Namespace: data[vault.SecretSyncTargetNamepaceKey], Name: data[vault.SecretSyncTargetNameKey]}
```
bbguimaraes, line 250: We could use a read-only lock from here on.

cmd/vault-subpath-proxy/kv_update_transport.go
```diff
@@ -96,9 +119,124 @@ func (k *kvUpdateTransport) RoundTrip(r *http.Request) (*http.Response, error) {
...
+					fetchErrLock.Unlock()
+					return
+				}
+
+				existingSecretKeysByNamespaceNameWriteLock.Lock()
+				defer existingSecretKeysByNamespaceNameWriteLock.Unlock()
+
+				name := types.NamespacedName{Namespace: item.Data[vault.SecretSyncTargetNamepaceKey], Name: item.Data[vault.SecretSyncTargetNameKey]}
+				if k.existingSecretKeysByNamespaceName[name] == nil {
+					k.existingSecretKeysByNamespaceName[name] = make(sets.String, len(item.Data)-2)
```
bbguimaraes, line 233: We could `delete` the two keys we don't care about and remove the special case here and below.

cmd/vault-subpath-proxy/kv_update_transport.go
bbguimaraes: It's worth adding comments indicating where each are used and that `ByVaultSecretName` is used as an index into `ByNamespaceName`.

alvaroaleman: Done

cmd/vault-subpath-proxy/kv_update_transport.go
alvaroaleman: Factored it out and made it run during startup. Not sure how long it takes and depends on the amount of data in Vault. But its definitely nicer to run this outside of user requests.

cmd/vault-subpath-proxy/kv_update_transport.go
alvaroaleman: Done

cmd/vault-subpath-proxy/kv_update_transport.go
alvaroaleman: Done

PR #2184, opened 2021-07-22
Abuse determinize-prow-config to implement 4.9 FF (no merge)

petr-muller:
Several notes:

- OCP repos are identified by a presence of a `group-lead-approved`-guarded `release-4.8` branch query
- OCP repos exempt from feature freeze process are allowlisted, and follow what was manually set up for openshift/builder repository: overlapping queries that merge either on valid bugs (one query) or three approvals (second query)
- Non-exempt repositories simply add valid-bug requirement for their master/main branch

The resulting configuration is unfornunately not in the canonical ordering (as all the sorting etc is done *before* the config is sharded, but we modify it on shard writes. You need to run `make prow-config` again after running this.

### DIFF
--- a/cmd/determinize-prow-config/main.go
+++ b/cmd/determinize-prow-config/main.go
@@ -11,10 +11,13 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"reflect"
+	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


droslean:
merged with https://github.com/openshift/ci-tools/pull/2188

PR #2185, opened 2021-07-22
pod-scaler: de-race startup

stevekuznetsov:
We were not correctly managing concurrency when loading data. This
resulted in incomplete loaded state for the first reload tick when very
small (short) data loads occurred, as in end-to-end tests.


stevekuznetsov:
Both of those are (different, unrelated) flakes

/retest

PR #2187, opened 2021-07-23
testhelper: use a file to share port state

stevekuznetsov:
We need to use a global marker of state to share the intent to use ports
as the in-memory lock we had in the past could not handle the case where
tests were run in parallel in separate processes.

This should fix a number of flakes in e2e

### DIFF
--- a/pkg/testhelper/accessory.go
+++ b/pkg/testhelper/accessory.go
@@ -13,7 +13,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
-	"sync"
 	"testing"
 	"time"
 
@@ -250,12 +249,6 @@ func (a *Accessory) ClientFlags() []string {
 	return a.clientFlags(a.port, a.healthPort)
 }
 
-var ports sync.Map
-
-func init() {
-	ports = sync.Map{}
-}
-
 // GetFreePort asks the kernel for a free open port that is ready to use.
 func GetFreePort(t TestingTInterface) string {
 	for {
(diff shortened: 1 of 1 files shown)


pkg/testhelper/accessory.go
```diff
@@ -268,14 +261,37 @@ func GetFreePort(t TestingTInterface) string {
...
 			t.Fatalf("could not listen on free port: %v", err)
 		}
-		defer func() {
-			if err := l.Close(); err != nil {
+		defer func(c io.Closer) {
+			if err := c.Close(); err != nil {
 				t.Errorf("could not close listener: %v", err)
 			}
-		}()
+		}(l)
```
stevekuznetsov, line 268: This is technically unrelated but it could leak unclosed connections as written today.

pkg/testhelper/accessory.go
```diff
@@ -268,16 +261,45 @@ func GetFreePort(t TestingTInterface) string {
...
+		// - process B calls net.ListenTCP() and resolves the same port
+		// - process A attempts to use the port, fails as it is in use
+		// Therefore, holding the listener open is our kernel-based lock for this process, and while
+		// we hold it open we must record our intent to disk.
+		lockDir := filepath.Join(os.TempDir(), "ci-tools-e2e-ports")
+		if err := os.MkdirAll(lockDir, 0777); err != nil {
+			t.Fatalf("could not create port lockfile dir: %v", err)
+		}
+		lockFile := filepath.Join(lockDir, port)
+		if _, err := os.Stat("/path/to/whatever"); os.IsNotExist(err) {
```
alvaroaleman, line 284: ?

pkg/testhelper/accessory.go
```diff
@@ -268,16 +261,45 @@ func GetFreePort(t TestingTInterface) string {
...
 			// we've never seen this before, we can use it
+			f, err := os.Create(lockFile)
```
alvaroaleman, line 286: This doesn't error if the file already exists, you need to use `os.OpenFile` with `O_EXCL`, ref https://man7.org/linux/man-pages/man2/open.2.html

pkg/testhelper/accessory.go
```diff
@@ -268,16 +261,45 @@ func GetFreePort(t TestingTInterface) string {
...
+			if err != nil {
+				t.Fatalf("could not record port lockfile: %v", err)
```
alvaroaleman, line 288: This is expected to fail in the case of a race, needs to be `continue`

pkg/testhelper/accessory.go
```diff
@@ -268,16 +261,45 @@ func GetFreePort(t TestingTInterface) string {
...
 				t.Errorf("could not close listener: %v", err)
 			}
-		}()
+		}(l)
 		port := strconv.Itoa(l.Addr().(*net.TCPAddr).Port)
-		if _, previouslyAllocated := ports.LoadOrStore(port, nil); !previouslyAllocated {
+		// Tests run in -parallel will run in separate processes, so we must use the file-system
+		// for sharing state and locking across them to coordinate who gets which port. Without
+		// some mechanism for sharing state, the following race is possible:
+		// - process A calls net.ListenTCP() to resolve a new port
```
alvaroaleman, line 273: calling `ListenTCP` here is now pointless

pkg/testhelper/accessory.go
stevekuznetsov: No, it is not. We have a lock on the port because we have an open listener.

pkg/testhelper/accessory.go
stevekuznetsov: We only ever hit this if the file does not exist, as we hold the kernel lock on the port.

pkg/testhelper/accessory.go
stevekuznetsov: What do you mean ?

pkg/testhelper/accessory.go
stevekuznetsov: .... why?

pkg/testhelper/accessory.go
emilvberglind: Should this be stat-ing the lockFile path instead of `/path/to/whatever` ?

pkg/testhelper/accessory.go
alvaroaleman: Ah oh wow. That is pretty complicated to follow along. Why not simply get a port, OpenFile with O_EXCL and retry if that fails? Then we have exactly one lock mechanism (the file) rather than two interleaved ones

pkg/testhelper/accessory.go
alvaroaleman: Yeah, that is what I meant

pkg/testhelper/accessory.go
stevekuznetsov: Because ~anything~ other than the kernel port lock is racy?

alvaroaleman: Sorry, what exactly  is racy when you try to open a file with `O_CREATE|O_EXCL`? It will fail if the file already exist, there is no race

PR #2193, opened 2021-07-28
Add method to inject a test from one ci-op config to another

petr-muller:
`WithPresubmitFrom()` returns a new configuration, where a selected test
from the source configuration is injected into the base configuration,
together with all elements from the source configuration that are
potentially necessary to allow that test to function in the context of
the base configuration. The intended use case is to inject the test
definition of a "release job" (informing/blocking) into a component
ci-operator config to allow dynamically executing any such job on any
component PR, without the need to clutter each individual component
ci-operator config.

The actual config merge looks a bit arbirary, but it is based on my experiments
in https://github.com/openshift/release/pull/20463

This commit is a part of an experimental effort and the code is not used
anywhere in the production path, so it intentionally does not deal with
all corner cases stemming from different ci-op config shapes on both
sides, conflicts etc - handling all cases is currently not the goal, handling
at least *some* cases is.

My eventual goal is to allow running `ci-operator` somewhat like this example
in a presubmit configured for all OCP repositories:

```console
$ ci-operator ... --target-injected-test=openshift/release@master__ci-4.9:e2e-aws-upgrade-single-node` 
```

The above would ask ci-operator-configresolver to resolve and return a
ci-operator config where the component config is a "base" (like now,
component config would be determined from `JOB_SPEC`) with the
`e2e-aws-upgrade-single-node` test from [o-release-master__ci-4.9 ](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/release/openshift-release-master__ci-4.9.yaml#L202-L208) config
injected into it (and then executed).

/cc @deads2k @stevekuznetsov 


pkg/api/config_test.go
```diff
@@ -0,0 +1,177 @@
...
+			expected: &ReleaseBuildConfiguration{
+				InputConfiguration: InputConfiguration{Releases: map[string]UnresolvedRelease{
+					"base-release":   {Release: &Release{Version: "4.9.base"}},
+					"source-release": {Release: &Release{Version: "4.9.source"}}},
+				},
+			},
+			defaultTests: true,
+		},
+		{
+			name:         "release from source is overwrites the one with same name from base",
```
emilvberglind, line 135:
```suggestion
			name:         "release from source overwrites the one with same name from base",
```

pkg/api/config_test.go
```diff
@@ -0,0 +1,177 @@
...
+			base: &ReleaseBuildConfiguration{InputConfiguration: InputConfiguration{BuildRootImage: &BuildRootImageConfiguration{FromRepository: true}}},
+			source: &ReleaseBuildConfiguration{InputConfiguration: InputConfiguration{BuildRootImage: &BuildRootImageConfiguration{
+				ImageStreamTagReference: &ImageStreamTagReference{Namespace: "source", Name: "source-is", Tag: "source-tag"},
+				UseBuildCache:           true,
+			}}},
+			expected:     &ReleaseBuildConfiguration{InputConfiguration: InputConfiguration{BuildRootImage: &BuildRootImageConfiguration{FromRepository: true}}},
+			defaultTests: true,
+		},
+		{
+			name:   "base_images is an union of both configs",
```
emilvberglind, line 101:
```suggestion
			name:   "base_images is a union of both configs",
```

pkg/api/config.go
```diff
@@ -1,9 +1,12 @@
 package api
 
 import (
+	"fmt"
 	"strings"
 	"time"
 
+	"github.com/getlantern/deepcopy"
```
alvaroaleman, line 8: Nit: We could use kubes DeepCopy codegeneration here by extending https://github.com/openshift/ci-tools/blob/eb21567cf6c99a76bb707bdf263b4b3fb751cd51/hack/update-codegen.sh#L5 and enabling it on package level like here: https://github.com/openshift/ci-tools/blob/eb21567cf6c99a76bb707bdf263b4b3fb751cd51/pkg/api/testimagestreamtagimport/v1/doc.go#L1-L4

Just mentioning that because I had a couple of situations in the past where I thought that would be useful for this package

pkg/api/config.go
```diff
@@ -103,3 +106,55 @@ func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependen
...
+// the source configuration that are potentially necessary to allow that test to function
+// in the context of the base configuration. The intended use case is to inject the test
+// definition of a "release job" (informing/blocking) into a component ci-operator config
+// to allow dynamically executing any such job on any component PR, without the need to
+// clutter each individual component ci-operator config.
+//
+// WARNING: This code is currently experimental and should not be used outside of the
+// "release jobs on PRs" effort
+// TODO: handle the presubmit/periodic better, extract code etc.
+func (config *ReleaseBuildConfiguration) WithPresubmitFrom(source *ReleaseBuildConfiguration, test string) (*ReleaseBuildConfiguration, error) {
```
deads2k, line 121: is `ReleaseBuildConfiguration` user input?  How does this get called?

petr-muller: 1. ci-operator asks ci-operator-config resolver for component config with a test injected from a different config
2. ci-operator-config resolver returns the result of `WithPresubmitFrom`, where a component config is base and different config is source

`ReleaseBuildConfiguration` is parsed ci-operator config (like https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oc/openshift-oc-master.yaml )

pkg/api/config.go
petr-muller: Sure, let's give that a shot but I don't have much experience with it - if it just works then I'll happily include it here.

petr-muller: Done. Had to disable for some types though.

pkg/api/config.go
petr-muller: Hopefully resolved in Slack convo

emilvberglind:
Makes sense to me, but someone else should probably slap the label on here.

pkg/api/config.go
```diff
@@ -103,3 +104,53 @@ func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependen
...
+		if name == "latest" && result.ReleaseTagConfiguration != nil {
+			continue
+		}
+		if result.Releases == nil {
+			result.Releases = map[string]UnresolvedRelease{}
+		}
+		result.Releases[name] = release
+	}
+
+	// TODO: handle resources, likely needs to be union, with max(config, source) on conflicts
```
stevekuznetsov, line 141: just let `pod-scaler` do it /shrug

pkg/api/config.go
```diff
@@ -103,3 +104,53 @@ func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependen
...
+
+	for i := range source.Tests {
+		if source.Tests[i].As == test {
+			// TODO: deal with name conflicts
```
stevekuznetsov, line 145: Why not just nuke all tests except for the one we need?

pkg/api/config.go
```diff
@@ -103,3 +104,53 @@ func (config *ReleaseBuildConfiguration) DependencyParts(dependency StepDependen
 	}
 	return stream, name, explicit
 }
+
+// WithPresubmitFrom returns a new configuration, where a selected test from the source
```
stevekuznetsov, line 108: It might almost make sense to think of it the other way - what do we want the root config to provide? Is it *just* the images? It's potentially hard to know "everything" that a test may need from the source config, could it be easier to just take source config, add images in, and call it a day? At least that is pretty bounded (base_images, images, raw_steps?)

pkg/api/config.go
petr-muller: ideally!

pkg/api/config.go
petr-muller: I think that's a good point. I think that it's easier to reason about the API in the direction I wrote it here (inject a test into a config), and the implementation may potentially differ if we find some use for it. But I gave it a mental experiment:

I think that transitive dependencies of a test (images, releases) are ultimately similar to transitive dependencies of images (images, releases), so you don't really save any work or complexity. I think that the source (root in your sentence) config mainly provides everything needed to build the images from source, so you need `images`, images likely need `src` (=`binary_build_command`), src needs build_root, you potentially need `operators` (although operators testing are beyong what I attempt to do here). You need base_images union. Maybe raw steps, but likely not.

I don't think anything like this will ever work (without a significant change in ci-operator) if the injected test depends on an image that's actually built in the source config - that implies you need access to the source code of source config's repo, and you don't have that in base config's repo presubmit.

So I'm leaning towards keeping it this way.

stevekuznetsov (approved):

PR #2198, opened 2021-07-30
configresolver: add an endpoint to inject a test into a config

petr-muller:
Builds over https://github.com/openshift/ci-tools/pull/2193 and adds a new `/configWithInjectedTest` endpoint that uses serves configs created by code https://github.com/openshift/ci-tools/pull/2193 . I was deciding between adding a new endpoint and just adding new set of query inputs to the existing `/config` one, but I think I will want to add more injection behavior in the future ("inject test with workflow X from the registry") so I think separate endpoint is better.

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -143,28 +143,21 @@ func injectTestFromQuery(w http.ResponseWriter, r *http.Request) (api.Metadata, 
 		}
 		return metadata, test, err
 	}
-	metadata.Org = r.URL.Query().Get(injectFromOrgQuery)
-	if metadata.Org == "" {
-		webreg.MissingQuery(w, injectFromOrgQuery)
-		return metadata, test, fmt.Errorf("missing query %s", injectFromOrgQuery)
-	}
-	metadata.Repo = r.URL.Query().Get(injectFromRepoQuery)
-	if metadata.Repo == "" {
-		webreg.MissingQuery(w, injectFromRepoQuery)
-		return metadata, "", fmt.Errorf("missing query %s", injectFromRepoQuery)
-	}
-	metadata.Branch = r.URL.Query().Get(injectFromBranchQuery)
-	if metadata.Branch == "" {
-		webreg.MissingQuery(w, injectFromBranchQuery)
-		return metadata, test, fmt.Errorf("missing query %s", injectFromVariantQuery)
+
+	for query, field := range map[string]*string{
+		injectFromOrgQuery:    &metadata.Org,
+		injectFromRepoQuery:   &metadata.Repo,
+		injectFromBranchQuery: &metadata.Branch,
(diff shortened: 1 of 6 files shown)


petr-muller:
I guess this needs a test

cmd/ci-operator-configresolver/main.go
```diff
@@ -119,6 +119,99 @@ func resolveConfig(configAgent agents.ConfigAgent, registryAgent agents.Registry
...
+			metrics.RecordError("config not found", configresolverMetrics.ErrorRate)
+			w.WriteHeader(http.StatusNotFound)
+			fmt.Fprintf(w, "failed to get config: %v", err)
+			logger.WithError(err).Warning("failed to get config")
+			return
+		}
+
+		injectFromMetadata, test, err := injectTestFromQuery(w, r)
+		if err != nil {
+			metrics.RecordError("invalid query", configresolverMetrics.ErrorRate)
```
emilvberglind, line 192: Do we still want to continue with the function if an error happens here?

petr-muller: No. Wow, sloppy. Thanks!

cmd/ci-operator-configresolver/main.go
```diff
@@ -119,6 +119,99 @@ func resolveConfig(configAgent agents.ConfigAgent, registryAgent agents.Registry
...
+
+	if r.Method != "GET" {
+		w.WriteHeader(http.StatusNotImplemented)
+		err := fmt.Errorf("expected GET, got %s", r.Method)
+		if _, errWrite := w.Write([]byte(http.StatusText(http.StatusNotImplemented))); errWrite != nil {
+			err = fmt.Errorf("%s and writing the response body failed with %w", err.Error(), errWrite)
+		}
+		return metadata, test, err
+	}
+	metadata.Org = r.URL.Query().Get(injectFromOrgQuery)
```
stevekuznetsov, line 142:
```go
var meta api.Metadata
for query, field := range map[string]*string{
    injectFromOrgQuery: &meta.Org,
    injectFromRepoQuery: &meta.Repo,
// ...
} {
    value := r.URL.Query().Get(query)
    if value == "" {
        webreg.MissingQuery(w, query)
        return metadata, test, fmt.Errorf("missing query %s", query)
    }
    *field = value
}

stevekuznetsov (commented):
one additional comment otherwise lgtm

cmd/ci-operator-configresolver/main.go
petr-muller: I just copied the code from the other method, so I added a commit that addresses this comment in all places.

cmd/ci-operator-configresolver/main.go
petr-muller: I based this method on a copy of an existing method, so I addressed both places in a separate commit. Thanks!

PR #2200, opened 2021-07-30
pod-scaler: add created-by-ci label to builds

stevekuznetsov:
The query we use to find pods for which we record data includes this
label, so the previous admission controller actually did nothing as it
never added enough labels to Build Pods such that they were recorded.

/assign @openshift/openshift-team-developer-productivity-platform


cmd/pod-scaler/frontend/node_modules/.bin/acorn
```diff
@@ -0,0 +1 @@
+../acorn/bin/acorn
```
alvaroaleman, line 1: What is this?

stevekuznetsov: WTF, should have been omitted - it's part of the frontend

stevekuznetsov:
@alvaroaleman fixed, sorry

alvaroaleman (approved):

alvaroaleman (approved):

PR #2201, opened 2021-07-30
pod-scaler: mutate resource requests by default

stevekuznetsov:
The server is already configured to only override resource requests when
the suggested value is *larger* than that present in the Pod, so we risk
nothing by turning on resource request suggestion. Limits, on the other
hand, can cause Pod eviction and require more thought.


cmd/pod-scaler/testdata/zz_fixture_TestMutatePodResources_resources_to_add__limits_disabled.diff
```diff
@@ -0,0 +1,48 @@
...
+  		Volumes:        nil,
+  		InitContainers: nil,
+  		Containers: []v1.Container{
+  			{Name: "large", Resources: {Limits: {s"cpu": {i: {value: 400}, Format: "DecimalSI"}, s"memory": {i: {value: 40000000000}, Format: "BinarySI"}}, Requests: {s"cpu": {i: {value: 200}, Format: "DecimalSI"}, s"memory": {i: {value: 30000000000}, Format: "BinarySI"}}}},
+  			{
+  				... // 6 identical fields
+  				EnvFrom: nil,
+  				Env:     nil,
+  				Resources: v1.ResourceRequirements{
+  					Limits: {},
```
stevekuznetsov, line 14: Limits are left untouched :)

stevekuznetsov: (Smaller limits, as well)

PR #2205, opened 2021-08-02
pkg/defaults: move defaulting code to `pkg/api`

bbguimaraes:
Breaking these minor ones from another change set as they are fairly
independent.


bbguimaraes:
Hmm, so much for that.  That's a strange special case, this approach won't work.

/close

PR #2212, opened 2021-08-03
Rehearse the corresponding tests only when a resource configuration changes

droslean:

https://issues.redhat.com/browse/DPTP-1265

### DIFF
--- a/pkg/diffs/diffs.go
+++ b/pkg/diffs/diffs.go
@@ -45,17 +45,18 @@ func GetChangedCiopConfigs(masterConfig, prConfig config.DataByFilename, logger 
 			continue
 		}
 
-		withoutTests := func(in cioperatorapi.ReleaseBuildConfiguration) cioperatorapi.ReleaseBuildConfiguration {
+		withoutTestsAndResources := func(in cioperatorapi.ReleaseBuildConfiguration) cioperatorapi.ReleaseBuildConfiguration {
 			var out cioperatorapi.ReleaseBuildConfiguration
 			if err := deepcopy.Copy(&out, &in); err != nil {
 				logrus.WithError(err).Warn("Could not deep copy configuration.") // this is a programming error
 				return out
 			}
 			out.Tests = nil
+			out.Resources = nil
 			return out
 		}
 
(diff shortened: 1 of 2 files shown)


pkg/diffs/diffs.go
```diff
@@ -72,6 +73,25 @@ func GetChangedCiopConfigs(masterConfig, prConfig config.DataByFilename, logger
 			}
 		}
 
+		oldResources := oldConfig.Configuration.Resources
+		newResources := newConfig.Configuration.Resources
+
+		for as, resource := range newResources {
+			if as == "*" && !equality.Semantic.DeepEqual(oldResources["*"], resource) {
+				for testName := range newTests {
+					jobs.Insert(testName)
```
stevekuznetsov, line 82: Changing `"*"` will only affect tests that otherwise do not have their own specific config, so this is still greedy

pkg/diffs/diffs.go
```diff
@@ -72,6 +73,25 @@ func GetChangedCiopConfigs(masterConfig, prConfig config.DataByFilename, logger
...
+					ret[filename] = newConfig
+				}
+				break
+			}
+
+			if !equality.Semantic.DeepEqual(oldResources[as], resource) {
+				logger.WithField(logCiopConfig, filename).Info(changedCiopConfigMsg)
+				ret[filename] = newConfig
+				jobs.Insert(as)
```
stevekuznetsov, line 91: Why can you insert the resource name identifier (could be a container name, build name, etc) as a job>

stevekuznetsov (commented):
Do we even care to re-trigger on changed resources? Setting them manually should be obsolete soon, pod-scaler just does it for you

pkg/diffs/diffs.go
droslean: Wait, I thought that the resource name is the same as the test name.

stevekuznetsov: Not necessarily - the resource name could be `*` or a build name, etc

stevekuznetsov: (we have logic to manage the matching)

droslean:
/close

Closing this since this change is not planned to impact the multi-stage jobs and this design will be deprecated in the future.

PR #2211, opened 2021-08-03
pkg/defaults: do not modify configuration

bbguimaraes:
Second attempt, after https://github.com/openshift/ci-tools/pull/2205.

Changes made more localized since defaulting affected other tools.

PR #2213, opened 2021-08-03
pkg/validation: validate targets are unique

bbguimaraes:
One unfortunate aspect of the current static validation is that it is done
on the unprocessed input configuration.  This results in significant
duplication between validation and parsing code.  This change attempts to
move to a more tenable position by performing validation on the graph
configuration structures, which are much closer to the final execution
graph.

https://issues.redhat.com/browse/DPTP-2388 is implemented in this change as
well as a PoC.

---

Use the naming methods and graph configuration data added in the previous
commits to validate that all targets in a `ci-operator` configuration are
unique.

Add this validation to the configuration loading process in
`ci-operator` and make `checkconfig` build and validate the static portion
of the graph.

/hold

built on top of / includes https://github.com/openshift/ci-tools/pull/2211.

### DIFF
--- a/cmd/ci-operator-checkconfig/main.go
+++ b/cmd/ci-operator-checkconfig/main.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/defaults"
 	"github.com/openshift/ci-tools/pkg/load"
 	"github.com/openshift/ci-tools/pkg/registry"
 	"github.com/openshift/ci-tools/pkg/steps/release"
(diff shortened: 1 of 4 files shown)


bbguimaraes:
/retest

`unit` failure is https://issues.redhat.com/browse/DPTP-2351.

pkg/defaults/defaults.go
```diff
@@ -787,8 +763,66 @@ func stepConfigsForBuild(
...
+		}
+		if target != nil {
+			istTagRef := &target.InputImage.BaseImage
+			if root.FromRepository {
+				var err error
+				istTagRef, err = buildRootImageStreamFromRepository(readFile)
+				if err != nil {
+					return nil, fmt.Errorf("failed to read buildRootImageStream from repository: %w", err)
+				}
+				ensureImageStreamTag(ctx, client, istTagRef, second)
```
stevekuznetsov, line 798: Is this a valid refactor? Looks like we're ensuring the input image before doing the build cache override?

stevekuznetsov (commented):
Mostly LGTM but this is a big one - @emilvberglind can you reivew?

pkg/defaults/defaults.go
bbguimaraes: Technically it is, because it maintains what we had before.

But the sequence does seem suspicious, I'll take a look.

bbguimaraes:
First couple of commits moved to https://github.com/openshift/ci-tools/pull/2239.

pkg/defaults/defaults.go
bbguimaraes: https://github.com/openshift/ci-tools/pull/2242

emilvberglind:
Makes sense to me!

/lgtm

PR #2215, opened 2021-08-03
ci-operator: ask resolver for config with additional test inject

petr-muller:
/hold

Builds over https://github.com/openshift/ci-tools/pull/2198

Code is a bit hacky, but I wanted to have something working. Probably needs tests and maybe an integration test. I did not want to change `load.Config` signature, so I chose to use functional option pattern. An annoying part of the whole thing is that `config.Load` is able to read the config from million different sources (files envvar, resolved, unresolved etc), but the injection only works with one of them, which may be unintuitive. /shrug


pkg/load/load.go
```diff
@@ -121,10 +122,57 @@ func fromPath(path string) (filenameToConfig, error) {
 	return configs, utilerrors.NewAggregate([]error{err, errGroup.Wait()})
 }
 
-func Config(path, unresolvedPath, registryPath string, info *ResolverInfo) (*api.ReleaseBuildConfiguration, error) {
+var testPathRegex = regexp.MustCompile(`(?P<org>[^/]+)/(?P<repo>[^@]+)@(?P<branch>[^:]+):(?P<test>.+)`)
+
+func parseCiopTestPath(path string) (ciopTestPath, error) {
```
stevekuznetsov, line 127: Can we just pass JSON please to the flag

pkg/load/load.go
```diff
@@ -186,22 +234,42 @@ func Config(path, unresolvedPath, registryPath string, info *ResolverInfo) (*api
 	return &configSpec, nil
 }
 
-func configFromResolver(info *ResolverInfo) (*api.ReleaseBuildConfiguration, error) {
+func configFromResolver(info *ResolverInfo, options configLoadOpt) (*api.ReleaseBuildConfiguration, error) {
```
stevekuznetsov, line 237: `ResolverInfo` is already a structure to hold the data we use to query the resolver - why not extend it?

pkg/load/load.go
```diff
@@ -186,22 +234,42 @@ func Config(path, unresolvedPath, registryPath string, info *ResolverInfo) (*api
...
+	queryParams := map[string]string{
+		"org":    info.Org,
```
stevekuznetsov, line 239: surely we have constants for those query parameters?

pkg/load/load.go
```diff
@@ -186,22 +234,42 @@ func Config(path, unresolvedPath, registryPath string, info *ResolverInfo) (*api
...
+		"repo":   info.Repo,
+		"branch": info.Branch,
+	}
 	identifier := fmt.Sprintf("%s/%s@%s", info.Org, info.Repo, info.Branch)
 	if info.Variant != "" {
 		identifier = fmt.Sprintf("%s [%s]", identifier, info.Variant)
```
stevekuznetsov, line 245: Isn't this logic (and the version on the injected one) already a method on `api.Metadata` ? Why doesn't the `ResolverInfo` embed one?

pkg/load/load.go
petr-muller: you mean `--inject-test='{"org": "o","repo": "r","branch": "b","test": "t"}`? That's super ugly and terrible to use...

I know you hate parsing information from strings, but string inputs are comfortable and readable, and regex parsing is not even that ugly.

pkg/load/load.go
petr-muller: My intent is to explore the dynamic injection concept further in the future, I want to try things like "inject a workflow test". That leads to different resolver endpoints with different sets of parameters, and I don't think we want to have a single struct that's basially an union of all different parameters?

I'll try to invest a bit more effort into refactoring the client code, let's see if I can try to come up with something better.

pkg/load/load.go
petr-muller: I guess, but the code here did not use them.

petr-muller:
@stevekuznetsov given that your feedback is more about the existing code and not my changes, let's hold this and I'll try to refactor the code to something more sensible.

pkg/load/load.go
petr-muller: But it doesn't seem this logic is a method on api.Metadata

petr-muller:
Extracted the existing code refactors to separate PRs:

- https://github.com/openshift/ci-tools/pull/2245
- https://github.com/openshift/ci-tools/pull/2246

stevekuznetsov:
Will review after rebase

pkg/load/load.go
petr-muller: I don't expect humans to pass this, but I expect humans (mostly DPTP) to occasionally eyeball the calls with the params, and I think it is useful if they are readable.

I don't find regex parsing here that complex or fragile - it's not using any complicated patterns that basically only deals with separators) and it could be even read by some `scanf` (using regex with named groups instead of scanf is probably my Python-influenced preference). Of course I'd add tests.

How strong are you on this? I agree that parsing JSON will make the code simpler, but at the expense of human observability, which is IMO rarely a good trade.

petr-muller:
@stevekuznetsov PTAL, took a slightly different approach than previous version (`ResolverClient` and using its new method directly from ci-op, instead of making `load.Config` more complicated). 

For input feedback, I gave some arguments in https://github.com/openshift/ci-tools/pull/2215#discussion_r693500683 and wrote some tests to deal with potential fragility, but it's not an issue I'd like to get blocked on, so if you feel strongly about it, let me know and I'll just let it parse JSON.

pkg/api/metadata.go
```diff
@@ -134,3 +134,28 @@ func BuildCacheFor(metadata Metadata) ImageStreamTagReference {
...
+	var ret MetadataWithTest
+	match := testPathRegex.FindStringSubmatch(input)
+	if match == nil || len(match) != 5 {
+		return &ret, fmt.Errorf("test path not in org/repo@branch:test or org/repo@branch__variant:test format: %s", input)
+	}
+	ret.Org = match[1]
+	ret.Repo = match[2]
+	ret.Test = match[4]
+
+	branchAndVariant := strings.Split(match[3], "__")
```
stevekuznetsov, line 150: Branches can contain `__` - not sure it matters to handle it, but also why IMO if you're going to stringify some data it's better to do it to a spec.

pkg/registry/server/client.go
```diff
@@ -0,0 +1,118 @@
+package server
```
stevekuznetsov, line 1: `client.go` in `package server`? lol

pkg/registry/server/client.go
```diff
@@ -0,0 +1,118 @@
...
+}
+
+func (r *resolverClient) ConfigWithTest(base *api.Metadata, testSource *api.MetadataWithTest) (*api.ReleaseBuildConfiguration, error) {
+	logrus.Infof("Loading configuration from %s for %s", r.Address, base.AsString())
+	req, err := http.NewRequest("GET", fmt.Sprintf("%s/configWithInjectedTest", r.Address), nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create request for configresolver: %w", err)
+	}
+	query := req.URL.Query()
+	query.Add(InjectTestQuery, testSource.Test)
```
stevekuznetsov, line 55:
```go
for query, value := range map[string]string {
    InjectTestQuery: testSource.Test,
// ...
} {
    query.add(query, value)
}
```

?

pkg/registry/server/client_test.go
```diff
@@ -0,0 +1,326 @@
...
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"k8s.io/utils/pointer"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+var parsedConfig = &api.ReleaseBuildConfiguration{
```
stevekuznetsov, line 16: Why is this a global (and not a `testdata/` entry?)

pkg/registry/server/client.go
petr-muller: Honestly our package game is :hankey: :rofl: 

Any suggestions? I think I previously had `pkg/registry/client/client.go` but then decided to merge them together b/c the client is coupled with the server via param names etc.

pkg/api/metadata.go
petr-muller: It's the same specification we use in ci-op naming convention though, so afaik we'll break there already for branches that contain `__`.

pkg/registry/server/client.go
petr-muller: I considered that when I wrote it, but then I thought it has no advantages in this specific case. It just makes code more indirect. You'd still need code lines to make that map.

pkg/registry/server/client_test.go
petr-muller: This existed before, I just moved it. I guess I can modernize it.

pkg/registry/server/client.go
```diff
@@ -0,0 +1,118 @@
...
+	query.Add(OrgQuery, base.Org)
+	query.Add(InjectTestQuery, testSource.Org)
```
petr-muller, line 57: bug

pkg/registry/server/client_test.go
petr-muller: Addressed

pkg/registry/server/client.go
petr-muller: Addressed

petr-muller:
/test images
Registry flake: `Error fetching blob: invalid status code from registry 503 (Service Unavailable)`

@stevekuznetsov addressed all feedback

pkg/load/load.go
stevekuznetsov: Humans should not read the flags we generate on the fly inside of a ProwJob YAML either - let's make the `ci-operator` log output clear when it's doing this on resolution?

petr-muller:
Thanks for all the input here!

PR #2218, opened 2021-08-03
Autoconfigbrancher: Don't create empty PRs

alvaroaleman:
If one command does a change that then gets reverted by another one,
the autoconfigbrancher will create empty PRs.

This change detecs that and aborts the PR creation if it happens.

Ref https://issues.redhat.com/browse/DPTP-2394

/assign @bbguimaraes

### DIFF
--- a/cmd/autoconfigbrancher/main.go
+++ b/cmd/autoconfigbrancher/main.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
 	"time"
@@ -114,6 +115,11 @@ func runAndCommitIfNeeded(stdout, stderr io.Writer, author, cmd string, args []s
 	return true, nil
 }
 
+type step struct {
+	command   string
+	arguments []string
+}
+
 func main() {
 	o := parseOptions()
 	if err := validateOptions(o); err != nil {
@@ -135,10 +141,7 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to change to root dir")
 	}
 
-	steps := []struct {
-		command   string
-		arguments []string
-	}{
+	steps := []step{
 		{
 			// Will check if repos that are part of the OCP payload
 			// have an up-to-date .ci-operator.yaml and if yes, update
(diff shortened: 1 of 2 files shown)


cmd/autoconfigbrancher/main.go
```diff
@@ -265,3 +259,39 @@ func main() {
...
 }
+
+func runSteps(steps []step, author string, stdout, stderr io.Writer) (needsPushing bool, err error) {
+	startCommitOut, err := exec.Command("git", "rev-parse", "HEAD").CombinedOutput()
+	if err != nil {
+		return false, fmt.Errorf("failed to execute `git rev-parse HEAD`: %w\noutput:%s\n", err, string(startCommitOut))
+	}
+	startCommitSHA := strings.TrimSpace(string(startCommitOut))
+
+	var commitsCounter int
```
bbguimaraes, line 270: All this tracking code seems redundant now since the inexpensive `git` call at the end will give us the same information.

cmd/autoconfigbrancher/main.go
```diff
@@ -265,3 +259,39 @@ func main() {
...
+			commitsCounter++
+		}
+	}
+
+	if commitsCounter == 0 {
+		logrus.Info("No new commits")
+		return false, nil
+	}
+
+	overallDiff, err := exec.Command("git", "diff", startCommitSHA).CombinedOutput()
```
bbguimaraes, line 287: `git diff` has a `--quiet` option for this, or we could compare tree hashes directly (the Jira issue has examples of the latter).

cmd/autoconfigbrancher/main.go
alvaroaleman: Execing out is all but inexpensive :P

cmd/autoconfigbrancher/main.go
alvaroaleman: I don't see anything with `--quiet` in https://issues.redhat.com/browse/DPTP-2394 ?

The reason I chose to compare the output and not the exit code is that I want to be able to differentiate "Executing the command failed" from "There is a diff".

cmd/autoconfigbrancher/main.go
bbguimaraes: "Inexpensive" with respect to the operation itself, which is essentially a hash comparison.

I don't see the need to have an extra return value for the function and a counter (which doesn't need to be a counter to begin with) if the result will be the same as letting `git` handle it with negligible cost, all in a single code path.

cmd/autoconfigbrancher/main.go
bbguimaraes: "The latter" referred to comparing tree hashes (i.e. `--format=%T`).

There are distinct exit codes for errors:

```
$ git diff --quiet; echo $?
1
$ git diff --quiet 404 --; echo $?
fatal: bad revision '404'
128
$ git -C /404 diff --quiet; echo $?
fatal: cannot change to '/404': No such file or directory
128
```

```
$ git stash push --quiet
$ git diff --quiet; echo $?
0
$ git diff --quiet 404 --; echo $?
fatal: bad revision '404'
128
$ git -C 404 diff --quiet; echo $?
fatal: cannot change to '404': No such file or directory
128
```

cmd/autoconfigbrancher/main.go
alvaroaleman: Because being able to differentiate those two cases might be interesting, ideally the second case would never happen.

cmd/autoconfigbrancher/main.go
alvaroaleman: Yeah, but that makes for both more complicated code as I have to assert the returned error to get the error code and makes it harder to understand, because most ppl do not know about the different exit codes of `git diff` while OTOH it provides no advantage (we still have to have an output buffer in case the command fails, otherwise debugging is impossible).

PR #2220, opened 2021-08-04
promoted-image-governor: init k8s client when needed

hongkailiu:

### DIFF
--- a/cmd/promoted-image-governor/main.go
+++ b/cmd/promoted-image-governor/main.go
@@ -328,11 +328,6 @@ func main() {
 		logrus.WithField("path", abs).Fatal("failed to operate on CI Operator's config directory")
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/promoted-image-governor/main.go
```diff
@@ -355,6 +350,11 @@ func main() {
 		return
 	}
 
+	kubeConfig, err := clientcmd.BuildConfigFromFlags("", opts.kubeconfig)
```
alvaroaleman, line 353: Why is this needed? It always has a kubeconfig or not?

hongkailiu: It loads from inCluster config if we do not provide it from flag.

alvaroaleman (approved):

PR #2233, opened 2021-08-06
pkg: Allow alternative reference-step command file extensions

wking:
2a31056baf (#1776) pivoted to directly executing command files allowing non-shell via shebangs.  But writing Python in a file with a `*.sh` extension isn't all that intuitive, and also confuses some editors.  This commit softens things to allow any extension.

Ideally we'd load the reference YAML and [pull commands out of there][1].  This would require some rewrites, since a bunch of these existing handlers are file-based, with `GetChangedRegistrySteps` logic like:

1. Get a set of changed files.
2. Iterate over the files to see if they seem to match a step.
3. Return any identified steps.

Instead we'd want:

1. Get a set of changed files.
2. Iterate over the nodes in the registry.
3. For each node, get the set of files consumed (e.g. both the ref YAML and command file for a reference step).
4. Mark the node as changed if there is any overlap between the changed-files set and the node-consumed-files sets.
5. Return any identified steps.

But that's a bigger pivot, so punting for now.

[1]: https://github.com/openshift/release/blob/66ee37c6805b900b1dc1d537126956b352286cbb/ci-operator/step-registry/ipi/install/install/ipi-install-install-ref.yaml#L5

### DIFF
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -186,7 +186,6 @@ func GetChangedTemplates(path, baseRev string) ([]string, error) {
 
 func loadRegistryStep(filename string, graph registry.NodeByName) (registry.Node, error) {
 	// if a commands script changed, mark reference as changed
-	filename = strings.ReplaceAll(filename, load.CommandsSuffix, load.RefSuffix)
 	var node registry.Node
 	var ok bool
 	switch {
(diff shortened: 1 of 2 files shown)


pkg/config/release.go
```diff
@@ -213,7 +215,7 @@ func GetChangedRegistrySteps(path, baseRev string, graph registry.NodeByName) ([
 		return changes, err
 	}
 	for _, c := range revChanges {
-		if filepath.Ext(c) == ".yaml" || strings.HasSuffix(c, load.CommandsSuffix) {
+		if filepath.Ext(c) == ".yaml" || strings.Contains(c, load.CommandsSuffix) {
```
AlexNPavel, line 218: This is too non-specific IMO and could result in an error if a step contains a `-commands` string in it since this might then try to load the metadata JSON for the step. I think something like `strings.HasSuffix(c, fmt.Sprintf("%s%s", load.CommandsSuffix, filepath.Ext(c)))` would be better.

AlexNPavel (changes requested):

pkg/config/release.go
wking: Picked up your suggested alternative in c75863a18 -> c47694dbb.

PR #2235, opened 2021-08-06
Check if the promoted tags are built from the latest commit

hongkailiu:
https://issues.redhat.com/browse/DPTP-1899

/cc @alvaroaleman

### DIFF
--- a/cmd/promoted-image-governor/main.go
+++ b/cmd/promoted-image-governor/main.go
@@ -21,9 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/tools/clientcmd"
-	"k8s.io/test-infra/prow/config/secret"
 	"k8s.io/test-infra/prow/flagutil"
-	git "k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/interrupts"
 	"k8s.io/test-infra/prow/logrusutil"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -35,6 +33,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/controller/promotionreconciler"
 	"github.com/openshift/ci-tools/pkg/steps/release"
+	"github.com/openshift/ci-tools/pkg/util"
 )
 
 type options struct {
@@ -52,7 +51,6 @@ type options struct {
 	logLevel string
 
 	ensurePromotedImagesUpToDate bool
-	flagutil.GitHubOptions
 }
 
 func parseOptions() *options {
@@ -75,7 +73,6 @@ func parseOptions() *options {
 	fs.StringVar(&opts.openshiftMappingDir, "openshift-mapping-dir", "", "Path to the openshift mapping directory")
 	fs.StringVar(&opts.openshiftMappingConfigPath, "openshift-mapping-config", "", "Path to the openshift mapping config file")
 	fs.BoolVar(&opts.ensurePromotedImagesUpToDate, "ensure-promoted-images-up-to-date", false, "Whether to check if promoted images are up to date")
-	opts.GitHubOptions.AddFlags(fs)
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("could not parse input")
 	}
@@ -111,9 +108,6 @@ func (o *options) validate() error {
 		}
 		logrus.WithField("re", re.String()).Info("Ignore tags as required by flag")
 		o.ignoredImageStreamTags = append(o.ignoredImageStreamTags, re)
-	}
-	if o.ensurePromotedImagesUpToDate {
-		return o.GitHubOptions.Validate(o.dryRun)
 	}
 	return nil
 }
@@ -286,24 +280,10 @@ type refGetter interface {
 }
 
 type gitRefGetter struct {
-	clone func(org, repo string) (git.RepoClient, error)
 }
 
 func (g *gitRefGetter) GetRef(org, repo, ref string) (ret string, retError error) {
-	r, err := g.clone(org, repo)
-	if err != nil {
-		return "", fmt.Errorf("failed to clone %s/%s: %w", org, repo, err)
-	}
-	defer func() {
-		if err := r.Clean(); err != nil {
-			retError = err
-		}
-	}()
-	ret, err = r.ShowRef(ref)
-	if err != nil {
-		return "", fmt.Errorf("failed to show ref %s of %s/%s: %w", ref, org, repo, err)
-	}
-	return ret, retError
+	return util.GetRemoteBranchCommitSha(org, repo, ref)
 }
 
 // checkImageStreamTags checks if promotedTags are built with the current head of org/repo/branch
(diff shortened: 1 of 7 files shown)


pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -195,24 +195,26 @@ func (r *reconciler) promotionConfig(ist *imagev1.ImageStreamTag) (*cioperatorap
...
+// CommitForIST returns the commit for an isTag
+func CommitForIST(ist *imagev1.ImageStreamTag) (string, error) {
 	metadata := &docker10.DockerImage{}
 	if err := json.Unmarshal(ist.Image.DockerImageMetadata.Raw, metadata); err != nil {
 		return "", fmt.Errorf("failed to unmarshal imagestream.image.dockerImageMetadata: %w", err)
 	}
 
 	commit := metadata.Config.Labels["io.openshift.build.commit.id"]
 	if commit == "" {
-		return "", controllerutil.TerminalError(errors.New("ImageStreamTag has no `io.openshift.build.commit.id` label, can't find out source commit"))
```
hongkailiu, line 206: We do not need `controllerutil.TerminalError` because its caller has it already.

cmd/promoted-image-governor/main.go
```diff
@@ -96,6 +103,9 @@ func (o *options) validate() error {
 		logrus.WithField("re", re.String()).Info("Ignore tags as required by flag")
 		o.ignoredImageStreamTags = append(o.ignoredImageStreamTags, re)
 	}
+	if o.ensurePromotedImagesUpToDate {
+		return o.GitHubOptions.Validate(false)
```
stevekuznetsov, line 107: pass the dry-run param

cmd/promoted-image-governor/main.go
```diff
@@ -262,6 +272,47 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
 	return false
 }
 
+type githubClient interface {
+	GetRef(org, repo, ref string) (string, error)
```
stevekuznetsov, line 276: Use Git please, this will consume way too many tokens.

cmd/promoted-image-governor/main.go
```diff
@@ -262,6 +272,47 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+// checkImageStreamTags checks if promotedTags are built with the current head of org/repo/branch
+func checkImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client, promotedTags map[api.ImageStreamTagReference]api.Metadata, gitHubClient githubClient) []error {
+	var errs []error
+	for promotedTag, metadata := range promotedTags {
+		ist := &imagev1.ImageStreamTag{}
+		istName := fmt.Sprintf("%s:%s", promotedTag.Name, promotedTag.Tag)
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: promotedTag.Namespace, Name: istName}, ist); err != nil {
+			if kerrors.IsNotFound(err) {
+				// The tag has not been promoted yet
+				logrus.WithError(err).Warnf("failed to get promoted isTag %s in namespace %s", istName, promotedTag.Namespace)
```
stevekuznetsov, line 288: put the comment into the error message, and consider logging nothing here since this is an expected state

cmd/promoted-image-governor/main.go
```diff
@@ -262,6 +272,47 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+		if err != nil {
+			errs = append(errs, fmt.Errorf("failed to get current git head for isTag %s in namespace %s: %w", istName, promotedTag.Namespace, err))
+			continue
+		}
+		if !found {
+			errs = append(errs, fmt.Errorf("got 404 for %s/%s/%s from github for isTag %s in namespace %s, this likely means the repo or branch got deleted or we are not allowed to access it", metadata.Org, metadata.Repo, metadata.Branch, istName, promotedTag.Namespace))
+			continue
+		}
+		if currentHEAD != istCommit {
+			errs = append(errs, fmt.Errorf("the isTag %s in namespace %s is not built from the head %s of %s/%s/%s", istName, promotedTag.Namespace, currentHEAD, metadata.Org, metadata.Repo, metadata.Branch))
```
stevekuznetsov, line 310: We should just trigger the job instead of logging something

stevekuznetsov: I'm also pretty sure I wrote a postsubmit for `openshift/release` that did this in the past, but it gor reverted due to some issues - can you look that up?

cmd/promoted-image-governor/main.go
hongkailiu: Done.

cmd/promoted-image-governor/main.go
hongkailiu: Done.

cmd/promoted-image-governor/main.go
hongkailiu: Removed.

cmd/promoted-image-governor/main.go
hongkailiu: Probably the postsubmit has been replaced by https://github.com/openshift/ci-tools/tree/master/pkg/controller/promotionreconciler

Back in time, promotionreconciler stopped working for some reason.
The current PR for DPTP-1899 will detect it and fire alerts (since it is an infra job).

cmd/promoted-image-governor/main.go
```diff
@@ -270,6 +281,67 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+type refGetter interface {
+	GetRef(org, repo, ref string) (string, error)
+}
+
+type gitRefGetter struct {
+	clone func(org, repo string) (git.RepoClient, error)
+}
+
+func (g *gitRefGetter) GetRef(org, repo, ref string) (ret string, retError error) {
+	r, err := g.clone(org, repo)
```
hongkailiu, line 293: @stevekuznetsov 
It will git-clone all repos if it builds promoted images, right?
Seems it is lots of work.

hongkailiu: Or you mean this?

```
$ git ls-remote https://github.com/openshift/images.git release-4.0
2f60da39a9d2e5cc00293b8ec7ad559fcd32446a	refs/heads/release-4.0
```

stevekuznetsov: `ls-remote` is the more efficient way, yes - we should already be doing this in promotion reconciler and have utilities for it upstream.

hongkailiu: promotion reconciler uses github client (that is where my first draft came from).
https://github.com/openshift/ci-tools/blob/73781b2ee20eee321a86d1de080e8b879d882737/pkg/controller/promotionreconciler/reconciler.go#L212-L215

We could switch that later if this one goes well.

I cannot find the API upstream.
The closest thing is 
https://github.com/kubernetes/test-infra/blob/2e2328ea3fa8bd91b11122b848b539a74292cb60/prow/git/v2/interactor.go#L151

stevekuznetsov:
This was the commit I was talking about: https://github.com/openshift/ci-tools/commit/28118640c7d0f70b526d95a1b4da27f20eb93dcf#diff-f6f0bb8e8bd20b0d74f1b2c9a5a847be762c883dcd0b7c64d20f440d0d570e44 - let's look into why it was reverted, IIRC there was some hot-loop triggering behavior or something. In any case, we will need to fix that before we deploy this alert, or we will get a lot of alerts. The current controller only ensures you get correct images within a ~day

cmd/promoted-image-governor/main.go
```diff
@@ -270,6 +275,53 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+	return util.GetRemoteBranchCommitSha(org, repo, ref)
+}
+
+// checkImageStreamTags checks if promotedTags are built with the current head of org/repo/branch
+func checkImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client, promotedTags map[api.ImageStreamTagReference]api.Metadata, refGetter refGetter) []error {
+	var errs []error
+	for promotedTag, metadata := range promotedTags {
+		ist := &imagev1.ImageStreamTag{}
+		istName := fmt.Sprintf("%s:%s", promotedTag.Name, promotedTag.Tag)
+		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: promotedTag.Namespace, Name: istName}, ist); err != nil {
```
alvaroaleman, line 295: Not sure if its a great idea to put this into a CLI, this will do # of promoted imagestreamtag api calls, which I believe is a high 4 digit/low 5 digit number?

It might be simpler to add this as a metric to the controller, it has the data cached anyways.

PR #2240, opened 2021-08-10
ci-operator: (fix) do not create TestISTagImport when running on app.ci

hongkailiu:
Follow up https://github.com/openshift/ci-tools/pull/2227

/cc @bbguimaraes @petr-muller

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -788,11 +788,8 @@ func (o *options) Run() []error {
 	if o.leaseServer != "" && o.leaseServerCredentialsFile != "" {
 		leaseClient = &o.leaseClient
 	}
-
-	o.resolveConsoleHost()
-
 	// load the graph from the configuration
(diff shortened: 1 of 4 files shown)


bbguimaraes:
For both review and posterity, can we base the change on the original code?

```
$ git revert 2e38dc07cda3782fa102c1f6f007d0eb4cbf5d0a
$ git cherry-pick cebbd0a4eb7a6915bb710c4ff2850ce74708d356
```

should do it (with a minor conflict fix).

hongkailiu:
> For both review and posterity, can we base the change on the original code?
> 
> ```
> $ git revert 2e38dc07cda3782fa102c1f6f007d0eb4cbf5d0a
> $ git cherry-pick cebbd0a4eb7a6915bb710c4ff2850ce74708d356
> ```
> 
> should do it (with a minor conflict fix).

Just tried. The fix is much closer to the current head.
If revert and then cherry-pick, I would have to copy over the missing part in the difference of the 2 RPs.

I did those instead:

```
git revert <commit in the old pr>
git cherry-pick <commit in the old pr>
git cherry-pick <commit in the new pr>
```
Then squashed the last 2 commits into 1.

PR #2241, opened 2021-08-10
[WIP] pkg/validation: ensure valid container test images

bbguimaraes:
Based on https://github.com/openshift/ci-tools/pull/2213.


pkg/validation/graph.go
```diff
@@ -9,14 +9,21 @@ import (
...
 // IsValidGraphConfiguration verifies the intermediary configuration is correct.
 // This is the ideal place for validations since the graph configuration is the
 // intermediary structure between the input configuration and the final
 // execution graph.  Validations done here take advantage of all the parsing
 // logic that has already been applied.
 func IsValidGraphConfiguration(rawSteps []api.StepConfiguration) error {
 	var ret []error
+	var containerTests []*api.TestStepConfiguration
 	names := sets.NewString()
+	pipelineImages := pipelineImageSet{
```
stevekuznetsov, line 23: What guarantees that the keys in this are unique?

bbguimaraes: Sorry, I thought there was a Jira issue for this already but there isn't, I've [added](https://issues.redhat.com/browse/DPTP-2445) one.

That verification was done in https://github.com/openshift/ci-tools/pull/2062, still done the old way, using the input configuration.  We can eventually migrate those, but this PR does not.

Here, we are only verifying that all images used by container tests are valid pipeline images, for which it is only necessary to know what names are valid, duplicates (which should already have been rejected by earlier validations) are not relevant.

stevekuznetsov (approved):

PR #2244, opened 2021-08-12
openstack: Rename massopencloud cluster profile

pierreprinetti:
While there are no jobs running on MOC due to a downtime, rename its
cluster profile to something less confusing.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1055,7 +1055,7 @@ const (
 	ClusterProfileIBMCloud              ClusterProfile = "ibmcloud"
 	ClusterProfileLibvirtPpc64le        ClusterProfile = "libvirt-ppc64le"
 	ClusterProfileLibvirtS390x          ClusterProfile = "libvirt-s390x"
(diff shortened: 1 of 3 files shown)


EmilienM:
I don't think this will work, you need to add a new profile and then switch jobs to use it, and then remove this one here, I think.

pierreprinetti:
> then switch jobs to use it

there are currently exactly zero jobs using it, that’s why I thought it was a good moment

EmilienM:
> > then switch jobs to use it
> 
> there are currently exactly zero jobs using it, that’s why I thought it was a good moment

oh right, since they are on maintenance. Good call!

EmilienM:
/lgtm
:+1:

pierreprinetti:
The window of opportunity is closed. Next time.
/close

EmilienM:
> The window of opportunity is closed. Next time.

I don't understand why, I think we can still do it.

/reopen

pierreprinetti:
deferred to after MOC's migration
/close

PR #2245, opened 2021-08-12
Embed `api.Metadata` in `ResolverInfo`

petr-muller:
`ResolverInfo` contains the same data with the same semantics, so using `api.Metadata` will allow using its helpders when working with resolver info

### DIFF
--- a/pkg/api/metadata.go
+++ b/pkg/api/metadata.go
@@ -36,6 +36,16 @@ func (m *Metadata) IsComplete() error {
 	return nil
 }
 
+// AsString returns a string representation of the metadata suitable for using
+// in human-oriented output
+func (m *Metadata) AsString() string {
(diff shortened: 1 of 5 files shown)


pkg/api/metadata.go
```diff
@@ -36,6 +36,16 @@ func (m *Metadata) IsComplete() error {
 	return nil
 }
 
+// AsString returns a string representation of the metadata suitable for using
+// in human-oriented output
+func (m *Metadata) AsString() string {
```
bbguimaraes, line 41: We could implement [`Stringer`](https://pkg.go.dev/fmt#Stringer) instead and drop the method calls.

stevekuznetsov: Good as a follow-up

stevekuznetsov (approved):

PR #2253, opened 2021-08-16
Add ClusterProfileVSphereDiscon

jinyunma:
vsphere use network segment as lease resource, in order to test disconnected ocp installation on vsphere (require machines in network segment unable to access to network), add new disconnected network segments as lease resources


petr-muller (approved):
/hold

petr-muller:
> /hold
I only held for Joseph to have a chance, ofc unhold at will

PR #2254, opened 2021-08-16
prowgen: generate conditionally triggered jobs correctly

petr-muller:
Conditionally triggered jobs (ones with `run_if_changed` and `skipIfOnlyChanged`) need to be generated as `always_run: false` to beconsistent.

Followup to https://github.com/openshift/ci-tools/pull/2221


petr-muller:
/hold
incomplete

PR #2257, opened 2021-08-16
lifecycle-oracle: implement a webserver

stevekuznetsov:
This simple server reads the data it's given and serves it to users when
they ask for it.

Implements [DPTP-2446](https://issues.redhat.com/browse/DPTP-2446)
/assign @petr-muller 
/cc @openshift/test-platform

### DIFF
--- a/cmd/lifecycle-oracle/README.md
+++ b/cmd/lifecycle-oracle/README.md
@@ -1,0 +1,25 @@
+# `lifecycle-oracle`
+
+The release lifecycle oracle serves data about phases for versions of products. Currently one API is supported: `/api/phases/`. This will return a document of all phases for all versions for all products, e.g.:
+
+```yaml
+ocp: # a product
+  "4.7": # a version
+    previous: # all previous phases
+      - event: "open" # the event
+        when: "2020-05-17T14:47:58Z" # when it occurred, optional
+    next: # the next event
+      event: "feature-freeze"
+  "4.6":
+    previous:
+    - event: "generally-available"
+      when: "2020-04-17T14:47:58Z"
+    - event: "code-freeze"
+      when: "2020-03-17T14:47:58Z"
+    - event: "feature-freeze"
+      when: "2020-02-17T14:47:58Z"
+    - event: "open"
+      when: "2020-01-17T14:47:58Z"
+    next:
+      event: "end-of-life"
+```
--- a/cmd/lifecycle-oracle/main.go
+++ b/cmd/lifecycle-oracle/main.go
@@ -1,0 +1,78 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/logrusutil"
+	"k8s.io/test-infra/prow/pjutil"
+	pprofutil "k8s.io/test-infra/prow/pjutil/pprof"
+	"k8s.io/test-infra/prow/version"
+)
+
+type options struct {
+	port                   int
+	instrumentationOptions prowflagutil.InstrumentationOptions
+	dataFile               string
+}
+
+func (o *options) validate() error {
+	if o.port == 0 {
+		return errors.New("--port is required")
+	}
+	if o.dataFile == "" {
+		return errors.New("--data-file is required")
+	}
+	return nil
+}
+
+func bindOptions(fs *flag.FlagSet) *options {
+	o := options{}
+	o.instrumentationOptions.AddFlags(fs)
+	fs.IntVar(&o.port, "port", 0, "Port to serve admission webhooks on.")
+	fs.StringVar(&o.dataFile, "data-file", "", "Path to file containing lifecycle phase data to serve.")
+	return &o
+}
+
+func main() {
+	flagSet := flag.NewFlagSet("", flag.ExitOnError)
+	opts := bindOptions(flagSet)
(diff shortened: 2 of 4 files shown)


pkg/lifecycle/api/types.go
```diff
@@ -0,0 +1,51 @@
...
+// VersionLifecyclePhases holds the lifecycle phases for all versions of a product.
+type VersionLifecyclePhases struct {
+	// ByVersion maps version to the lifecycle phases
+	ByVersion map[string]LifecyclePhases `json:"by_version"`
+}
+
+// LifecyclePhases holds the lifecycle phases for a version of a product.
+type LifecyclePhases struct {
+	// Previous lists all lifecycle phases that have already occurred. Optional.
+	Previous []LifecyclePhase `json:"previous,omitempty"`
```
stevekuznetsov, line 20: Note that this differs from the spec - but I believe we need the full list of previous events unless we start inferring that all releases all have the same steps in the same order (which I think is already close to not being the case, e.g. see how we're changing the handling of feature freeze), so this is much safer.

If we don't, then e.g. a config manager which never passes during a phase or which is written after a phase is over may not know that it needs to do something.

cmd/lifecycle-oracle/main.go
```diff
@@ -0,0 +1,78 @@
...
+
+func bindOptions(fs *flag.FlagSet) *options {
+	o := options{}
+	o.instrumentationOptions.AddFlags(fs)
+	fs.IntVar(&o.port, "port", 0, "Port to serve admission webhooks on.")
+	fs.StringVar(&o.dataFile, "data-file", "", "Path to file containing lifecycle phase data to serve.")
+	return &o
+}
+
+func main() {
```
alvaroaleman, line 46: What advantage does this provide over just serving the file through a static webserver like nginx?

cmd/lifecycle-oracle/main.go
```diff
@@ -0,0 +1,78 @@
...
+	if o.dataFile == "" {
+		return errors.New("--data-file is required")
+	}
+	return nil
+}
+
+func bindOptions(fs *flag.FlagSet) *options {
+	o := options{}
+	o.instrumentationOptions.AddFlags(fs)
+	fs.IntVar(&o.port, "port", 0, "Port to serve admission webhooks on.")
```
emilvberglind, line 41: This looks like a copy/paste from the pod-scaler.

stevekuznetsov: guilty

cmd/lifecycle-oracle/main.go
stevekuznetsov: It's a seed that's intended to grow but yes in this state it is no better than static

cmd/lifecycle-oracle/main.go
```diff
@@ -0,0 +1,78 @@
...
+	logrus.Infof("%s version %s", version.Name, version.Version)
+
+	pprofutil.Instrument(opts.instrumentationOptions)
+	health := pjutil.NewHealthOnPort(opts.instrumentationOptions.HealthPort)
+
+	mux := http.NewServeMux()
+	mux.Handle("/api/phases/", http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+		data, err := os.Open(opts.dataFile)
+		if err != nil {
+			http.Error(writer, fmt.Sprintf("Could not open data file: %v", err), 500)
```
alvaroaleman, line 65: You need to return here and in the other error

pkg/lifecycle/api/types.go
```diff
@@ -0,0 +1,51 @@
+package api
+
+import v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
alvaroaleman, line 3: Lets please alias this as `metav1`, otherwise this is really annoying to read as you have to guess what `v1` package the code refers to

pkg/lifecycle/api/types.go
```diff
@@ -0,0 +1,51 @@
+package api
```
alvaroaleman, line 1: Why not `pkg/api/lifecycle` like for everything else?

cmd/lifecycle-oracle/README.md
```diff
@@ -0,0 +1,25 @@
+# `lifecycle-oracle`
+
+The release lifecycle oracle serves data about phases for versions of products. Currently one API is supported: `/api/phases/`. This will return a document of all phases for all versions for all products, e.g.:
+
+```yaml
+ocp: # a product
+  "4.7": # a version
+    previous: # all previous phases
+      - event: "open" # the event
+        when: "2020-05-17T14:47:58Z" # when it occurred, optional
```
petr-muller, line 10: Are *past* event dates optional though? Not sure.

cmd/lifecycle-oracle/README.md
```diff
@@ -0,0 +1,25 @@
...
+  "4.6":
+    previous:
+    - event: "generally-available"
+      when: "2020-04-17T14:47:58Z"
+    - event: "code-freeze"
+      when: "2020-03-17T14:47:58Z"
+    - event: "feature-freeze"
+      when: "2020-02-17T14:47:58Z"
+    - event: "open"
+      when: "2020-01-17T14:47:58Z"
```
petr-muller, line 22: In the related PR for a config manager, Alvaro asked if we can assume this array to be sorted (and hence the first one  in the list is always the last event). We may want to document the assumtion about ordering (or the lack of it, if needed) here.

petr-muller (commented):
LOL, so... one nit about the name. I called it "oracle" during design in a math-y sense ("entity capable of solving some problem") because I wanted to abstract the actual problem of determining the lifecycle data (which at that time seemed hard). "Let's assume we build a thing that know how to do it". I'm not sure if we need to call the real thing an "oracle". Maybe simply "lifecycle-server"?

PR #2260, opened 2021-08-17
DPTP-1989: Make sure that we add boskos client args/mounts if any test step has a lease.

emilvberglind:
[DPTP-1989](https://issues.redhat.com/browse/DPTP-1989): Make sure that we add boskos client args/mounts if any test step has a lease.

### DIFF
--- a/pkg/api/leases_test.go
+++ b/pkg/api/leases_test.go
@@ -1,0 +1,49 @@
+package api
+
+import (
+	"testing"
+
+	"k8s.io/utils/diff"
+)
+
+func TestLeasesForTest(t *testing.T) {
+	for _, tc := range []struct {
+		name     string
+		tests    MultiStageTestConfigurationLiteral
+		expected []StepLease
+	}{{
+		name:  "no configuration or cluster profile, no lease",
+		tests: MultiStageTestConfigurationLiteral{},
+	}, {
+		name: "cluster profile, lease",
+		tests: MultiStageTestConfigurationLiteral{
+			ClusterProfile: ClusterProfileAWS,
+		},
+		expected: []StepLease{{
+			ResourceType: "aws-quota-slice",
+			Env:          DefaultLeaseEnv,
+			Count:        1,
+		}},
+	}, {
+		name: "explicit configuration, lease",
+		tests: MultiStageTestConfigurationLiteral{
+			Leases: []StepLease{{ResourceType: "aws-quota-slice"}},
+		},
+		expected: []StepLease{{ResourceType: "aws-quota-slice"}},
+	}, {
+		name: "explicit configuration in step, lease",
+		tests: MultiStageTestConfigurationLiteral{
+			Test: []LiteralTestStep{
+				{Leases: []StepLease{{ResourceType: "aws-quota-slice"}}},
+			},
+		},
+		expected: []StepLease{{ResourceType: "aws-quota-slice"}},
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			ret := LeasesForTest(&tc.tests)
+			if diff := diff.ObjectReflectDiff(tc.expected, ret); diff != "<no diffs>" {
+				t.Errorf("incorrect leases: %s", diff)
+			}
+		})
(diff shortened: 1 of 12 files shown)


pkg/prowgen/prowgen.go
```diff
@@ -333,18 +333,39 @@ func generatePodSpecMultiStage(info *ProwgenInfo, test *cioperatorapi.TestStepCo
...
+		container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{Name: "cluster-profile", MountPath: clusterProfilePath})
 	}
-	podSpec.Volumes = append(podSpec.Volumes, generateClusterProfileVolume(profile, profile.ClusterType()))
-	clusterProfilePath := fmt.Sprintf("/usr/local/%s-cluster-profile", test.As)
-	container := &podSpec.Containers[0]
-	container.Args = append(container.Args, fmt.Sprintf("--secret-dir=%s", clusterProfilePath))
-	container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{Name: "cluster-profile", MountPath: clusterProfilePath})
-	addLeaseClient(podSpec)
+
+	if testContainsLease(test) || profile != "" {
```
bbguimaraes, line 345: Reverse the predicates since the second is much less expensive to compute and can likely be combined with the previous `if`.

pkg/prowgen/prowgen.go
```diff
@@ -333,18 +333,39 @@ func generatePodSpecMultiStage(info *ProwgenInfo, test *cioperatorapi.TestStepCo
...
+func testContainsLease(test *cioperatorapi.TestStepConfiguration) bool {
+	// this is predicated upon the config being fully resolved at this time.
+	if test.MultiStageTestConfigurationLiteral == nil {
+		return false
+	}
+
+	hasLeases := false
+	for _, test := range test.MultiStageTestConfigurationLiteral.Test {
+		if len(test.Leases) > 0 {
+			hasLeases = true
```
bbguimaraes, line 361: We can just `return` and remove the local variable.

pkg/prowgen/prowgen.go
```diff
@@ -333,18 +333,39 @@ func generatePodSpecMultiStage(info *ProwgenInfo, test *cioperatorapi.TestStepCo
...
 }
 
+func testContainsLease(test *cioperatorapi.TestStepConfiguration) bool {
+	// this is predicated upon the config being fully resolved at this time.
+	if test.MultiStageTestConfigurationLiteral == nil {
+		return false
+	}
+
+	hasLeases := false
+	for _, test := range test.MultiStageTestConfigurationLiteral.Test {
```
bbguimaraes, line 359: This also needs to look at leases declared in steps, see for example:

https://github.com/openshift/ci-tools/blob/0521ae628982b19092d6d38bb9cb92d945474cc1/pkg/defaults/defaults.go#L484-L500

bbguimaraes: We could also take the opportunity to create some proper functions in the `api` package for these now that they're needed elsewhere.

bbguimaraes (changes requested):

pkg/prowgen/prowgen.go
emilvberglind: Yeah, good call.

pkg/api/leases_test.go
```diff
@@ -0,0 +1,49 @@
...
+		tests: MultiStageTestConfigurationLiteral{
+			Test: []LiteralTestStep{
+				{Leases: []StepLease{{ResourceType: "aws-quota-slice"}}},
+			},
+		},
+		expected: []StepLease{{ResourceType: "aws-quota-slice"}},
+	}} {
+		t.Run(tc.name, func(t *testing.T) {
+			ret := LeasesForTest(&tc.tests)
+			if diff := diff.ObjectReflectDiff(tc.expected, ret); diff != "<no diffs>" {
```
stevekuznetsov, line 44: nit: prefer cmp.Diff

PR #2268, opened 2021-08-19
[WIP] ci-operator: add support for (optional) CLUSTER_OWNER to determine lease type

andfasano:
This PR adds the possibility to use, optionally, the env var `CLUSTER_OWNER` to determine the effective lease type requested to Boskos.
The original request came from https://github.com/openshift/release/pull/21071 where (specifically for the Packet profile) there was the need to specify which Equinix project the instance should be accounted for (but still the cluster profile remains the same).
The idea is to allow each team (or owner) to specify its own quota based on its own budget in the [Boskos configuration](https://github.com/openshift/release/blob/0017d437faeb1a00496c318c59a9f5a1b6ba1f66/core-services/prow/02_config/_boskos.yaml#L725-L728)

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1267,10 +1267,14 @@ func (p ClusterProfile) LeaseType() string {
 }
 
 // LeaseTypeFromClusterType maps cluster types to lease types
-func LeaseTypeFromClusterType(t string) (string, error) {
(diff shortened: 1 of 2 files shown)


andfasano:
Not required anymore

PR #2269, opened 2021-08-19
[WIP]DPTP-2412: Automate the applyconfig jobs and the secret configs for a new cluster in build-farm

smg247:


cmd/cluster-init/main.go
```diff
@@ -0,0 +1,115 @@
...
+	err := validateOptions(o)
+	check(err, "Invalid arguments.")
+
+	//TODO: probably a good idea to validate that this cluster doesn't exist
+	// i think we can use the presence of a build dir
+
+	updateInfraPeriodics(o)
+	updatePostsubmits(o)
+	updatePresubmits(o)
+	//TODO: is the following good enough? it is hard to modify MD programmatically
```
smg247, line 84: I think I am going to accept user input for what the entry about the new cluster should be in the MD

cmd/cluster-init/config_io.go
```diff
@@ -0,0 +1,22 @@
+package main
+
+import (
+	"io/ioutil"
+	"sigs.k8s.io/yaml"
+)
+
+func loadConfig(filename string, o interface{}) {
```
stevekuznetsov, line 8: nit: Go leans very heavily into being verbose. While it may look DRY to write a function like this, it is not, as the inputs to all of the calls are parameterized. Furthermore, `o interface{}` is the Go equivalent of `void*` or `Object` and circumvents the type system, which is not desirable.

cmd/cluster-init/generate_job.go
```diff
@@ -0,0 +1,231 @@
...
+		Context: fmt.Sprintf("%s/%s/%s-%s", Ci, BuildFarm, clusterName, Dry),
+	}
+}
+
+func generateJobBase(ps *v1.PodSpec, uc prowconfig.UtilityConfig, name string) prowconfig.JobBase {
+	return prowconfig.JobBase{
+		Name:            name,
+		Agent:           Kubernetes,
+		Cluster:         AppDotCi,
+		Namespace:       nil,
```
stevekuznetsov, line 131: Fields that are not provided will take on their zero values, which for well-formed Go structs will be valid values that do what you want. So, idiomatic Go would do the following instead of writing `generateJobBase`:

```go
base := prowconfig.JobBase{
    Name: name,
    Agent: prowconfig.KubernetesAgent, // find the upstream constant for this
    Spec: ps,
    UtilityConfig: uc,
}
```

cmd/cluster-init/generate_job.go
```diff
@@ -0,0 +1,231 @@
...
+	mounts := generateVolumeMounts()
+	mounts = append(mounts, v1.VolumeMount{
+		Name:      Tmp,
+		MountPath: fmt.Sprintf("/%s", Tmp),
+	})
+	container := generateContainer(image, generateArgs(buildFarmDir), mounts)
+	secretVolume := generateSecretVolume(clusterName)
+	emptyVolume := generateEmptyVolume(Tmp)
+	podSpec := generatePodSpec(container, []v1.Volume{secretVolume, emptyVolume})
+	name := fmt.Sprintf("%s-%s-%s-%s-%s-%s-%s", Pull, Ci, Openshift, Release, Master, clusterName, Dry)
```
stevekuznetsov, line 98: You want `api.Metadata.JobName()`

cmd/cluster-init/generate_job.go
```diff
@@ -0,0 +1,231 @@
...
+
+func generatePodSpec(c v1.Container, v []v1.Volume) *v1.PodSpec {
+	return &v1.PodSpec{
+		Volumes:            v,
+		Containers:         []v1.Container{c},
+		ServiceAccountName: ConfigUpdater,
+	}
+}
+
+func generateReleaseRef() prowapi.Refs {
```
stevekuznetsov, line 152: Functions which have only one call-site should be inlined

cmd/cluster-init/generate_job.go
```diff
@@ -0,0 +1,231 @@
...
+	name := fmt.Sprintf("%s-%s-%s-%s-%s-%s-%s", Pull, Ci, Openshift, Release, Master, clusterName, Dry)
+	jobBase := generateJobBase(podSpec, utilityConfig, name)
+	jobBase.Labels = map[string]string{
+		fmt.Sprintf("%s.%s.%s/%s", PjRehearse, Openshift, Io, CanBeRehearsed): "true",
+	}
+	brancher := prowconfig.Brancher{
+		Branches: []string{"master"},
+	}
+	rerun := fmt.Sprintf("/%s %s-%s", Test, clusterName, Dry)
+	trigger := fmt.Sprintf("(?m)^/%s( | .* )%s-%s,?($|\\s.*)", Test, clusterName, Dry)
```
stevekuznetsov, line 107: You want `prowconfig.DefaultRerunCommandFor` and `prowconfig.DefaultTriggerFor` - check out `prowgen.go#L460`

cmd/cluster-init/generate_job.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+	return prowconfig.Postsubmit{
+		JobBase:  jobBase,
+		Brancher: brancher,
+	}
+}
+
+func generatePresubmit(clusterName string, buildFarmDir string) prowconfig.Presubmit {
+	utilityConfig := generateUtilityConfig()
+	image := fmt.Sprintf("%s.%s.%s.%s/%s/%s:%s", Registry, Ci, Openshift, Org, Ci, ApplyConfig, Latest)
```
stevekuznetsov, line 88: You likely want the image itself to be the constant - will any individual field in this ever change?

cmd/cluster-init/generate_job.go
```diff
@@ -0,0 +1,231 @@
...
+	mounts := generateVolumeMounts()
+	mounts = append(mounts, v1.VolumeMount{
+		Name:      Tmp,
+		MountPath: fmt.Sprintf("/%s", Tmp),
+	})
+	container := generateContainer(image, generateArgs(buildFarmDir), mounts)
+	secretVolume := generateSecretVolume(clusterName)
+	emptyVolume := generateEmptyVolume(Tmp)
+	podSpec := generatePodSpec(container, []v1.Volume{secretVolume, emptyVolume})
```
stevekuznetsov, line 97: You don't need to be able to generate any possible spec, but just the two that we need here. Let's find a way to factor over just the differences (or just inline the struct literals if there's not much shared between the two)

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,115 @@
...
+
+func validateOptions(o options) error {
+	if o.clusterName == "" {
+		return fmt.Errorf("--cluster-name must be provided")
+	}
+	if o.releaseRepo == "" {
+		return fmt.Errorf("--release-repo must be provided")
+	}
+	if o.buildFarmDir == "" {
+		return fmt.Errorf("--build-farm-dir must be provided")
```
stevekuznetsov, line 47: Is this not deterministic from the name of the new cluster, under the release repo ?

cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -0,0 +1,326 @@
...
+		},
+	}
+	return sc
+}
+
+func generatePushPullSecretFrom(clusterName string, items []secretbootstrap.DockerConfigJSONData) secretbootstrap.ItemContext {
+	itemContext := secretbootstrap.ItemContext{
+		DockerConfigJSONData: []secretbootstrap.DockerConfigJSONData{
+			{
+				AuthField:   fmt.Sprintf("token_image-puller_%s_reg_auth_value.txt", clusterName),
```
stevekuznetsov, line 134: Seems like we want a `func registryPullTokenField(clusterName string) string`

stevekuznetsov (commented):
Looking good! Most of the comments are structural - we want to think about control flow and error returns bottom-up, and we can get away with fewer small functions that don't get re-used. I'm sure Robert Martin would disagree but IMHO unless the function is doing something non-obvious or complex where naming it would increase readability, or if you need to test some minor utility alone, in-lining logic that is used once has no cost and makes program flow easier to read.

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,115 @@
...
+	createSymlink(fmt.Sprintf("../%s", CommonExceptAppCi), commonExceptAppLink)
+}
+
+func createSymlink(target string, link string) {
+	fmt.Printf("Creating symlink from: %s to: %s\n", link, target)
+	err := os.Symlink(target, link)
+	check(err, "cannot create symlink.")
+}
+
+func check(err error, args ...interface{}) {
```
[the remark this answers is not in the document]
smg247: Gotcha, I saw this being used on https://gobyexample.com/, but its good to know that this isn't generally how this is done.

cmd/cluster-init/generate_job.go
smg247: Cool, that makes sense!

cmd/cluster-init/generate_job.go
smg247: That makes sense, I increasingly was disliking the way I was handling the constants as they grew larger.

cmd/cluster-init/main.go
smg247: It could be, I thought allowing the option to be provided would make the tool more "future proof" but for now we can just derive it from the cluster name pretty easily.

cmd/cluster-init/update_jobs.go
```diff
@@ -0,0 +1,141 @@
...
+	configPath := fmt.Sprintf("%s.%s.%s.%s", Sa, ConfigUpdater, clusterName, Config)
+	path := v1.KeyToPath{
+		Key:  configPath,
+		Path: configPath,
+	}
+	v.Secret.Items = append(v.Secret.Items, path)
+}
+
+func findPeriodic(ip *InfraPeriodics, name string) (*prowconfig.Periodic, error) {
+	idx := -1
```
droslean, line 90: IMHO It should be easier to create a map with the job names, and then just check if the key exists.

cmd/cluster-init/update_jobs.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+	}
+	if idx != -1 {
+		return &c.Env[idx], nil
+	}
+	return &v1.EnvVar{}, fmt.Errorf("couldn't find Env with name: %s", name)
+}
+
+func findVolume(ps *v1.PodSpec, name string) (*v1.Volume, error) {
+	idx := -1
```
droslean, line 130: same for volumes.

cmd/cluster-init/config_io.go
smg247: This is listed as a "nit" but I really would like to learn the go way to do this. The `yaml.Unmarshal` function takes `o interface{}` as a parameter so I was mimicking the signature there. Do you suggest that this all be inlined for _each_ config that I am utilizing?

cmd/cluster-init/main.go
smg247: So in the `build03` cluster case we have the `buildFarmDir` of `03_cluster`. We _could_ deterministically generate this from the cluster name (strip of alpha chars, use the numeric as a prefix, append `_cluster`), but I am not sure that that is _better_ than just accepting it as a parameter to the tool. What do you think?

cmd/cluster-init/main.go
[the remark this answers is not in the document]
stevekuznetsov: They get it right here: https://gobyexample.com/errors - I think in other areas they might be "cheating" for simplicity

cmd/cluster-init/config_io.go
stevekuznetsov: Yep! That's the verbosity cost of Go. Seems crazy, but the general thought is - so what? This code is written once. Hiding *how* you serialize and write a file only makes the callsite harder to read, and there's no real cost after write-time for `ioutil.ReadFile()` followed by `yaml.Unmarshal`, so the benefit to read is where the language heads.

cmd/cluster-init/main.go
stevekuznetsov: Let's change the directory structure to be 1:1 with cluster names - there's nothing we gain by having `03_cluster` instead of `build03`

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,115 @@
...
+	check(err)
+
+	commonLink := filepath.Join(o.releaseRepo, Clusters, BuildClusters, o.buildFarmDir, Common)
+	createSymlink(fmt.Sprintf("../%s", Common), commonLink)
+
+	commonExceptAppLink := filepath.Join(o.releaseRepo, Clusters, BuildClusters, o.buildFarmDir, CommonExceptAppCi)
+	createSymlink(fmt.Sprintf("../%s", CommonExceptAppCi), commonExceptAppLink)
+}
+
+func createSymlink(target string, link string) {
```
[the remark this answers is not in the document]
smg247: That makes sense, especially considering that we want to bubble up the error after all.

cmd/cluster-init/main.go
smg247: Should we modify the existing clusters (`build01`, and `build02`) to use this also? or just do this going forward? I am a bit concerned with missing something if I do the former, but I could do an exhaustive full-text search and take care of it if that is desired (possibly in a new PR).

cmd/cluster-init/main.go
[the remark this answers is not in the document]
smg247: ah yeah, you're right!

PR #2272, opened 2021-08-19
Add a bugzilla config manager

alvaroaleman:
Ref https://issues.redhat.com/browse/DPTP-2441

/assign @petr-muller 

Can you sanity check if the implementation is generally correct? Just want to make sure I understood the jira issue properly before I invest time in testing this more granular.


cmd/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,145 @@
...
+	"strconv"
+	"strings"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/test-infra/prow/plugins"
+	utilpointer "k8s.io/utils/pointer"
+)
+
+func main() {
+	// TODO: Fetch oracle, run(), createPR
```
petr-muller, line 16: Managers themselves will not maintain their PRs - just commits. This is to allow non-Go managers to participate.

DPTP-2443 is the story for the helper to manager PRs. I'm not super-strong on whether the manager should make commits or not, but I'm leaning towards yes, because it knows what changes it makes.

cmd/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,145 @@
...
+		if ocpProduct != "ocp" {
+			continue
+		}
+
+		for productVersion, config := range productConfig {
+			if len(config.Previous) < 1 {
+				continue
+			}
+
+			// TODO: We assume the first event is the last - Is that reasonable?
```
petr-muller, line 45: IMO it is, but it needs to be a contract with the lifecycle server

cmd/bugzilla-config-manager/testdata/zz_fixture_TestRun.yaml
```diff
@@ -0,0 +1,59 @@
+default:
+  main:
+    target_release: "4.70"
```
petr-muller, line 3: `4.70` looks fishy?

cmd/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,145 @@
...
+		config.TargetRelease = utilpointer.String(developmentVersion.String() + "0")
+
+		cfg.Default[developmentBranchName] = config
+	}
+
+	// developmentVersion/developmentVersion+1: Set config for {release/openshift}-$VERSION:
+	// * dependentBugStatus: ["MODIFIED", "ON_QA", "VERIFIED"]
+	// * DependentBugTargetReleases: "$VERSION+1.0"
+	// * ValidateByDefault: true
+	for _, developmentVersion := range []MajorMinor{developmentVersion, developmentVersion.WithIncrementedMinor(1)} {
```
petr-muller, line 110: I'm not really sure we should assume anything about the X.Y+1 one: if that's a version, it should be in a lifecycle with an "open" even some time in the future, and if we need special logic for that phase we should just have it instead of piggy backing on the development version.

petr-muller (commented):
Overall, looks good to me, with just one comment about the logic itself (inline)

In general, one thing that's generally hard for me is reasoning over the full config that includes all versions, so I'd be really happy if we could have a set of small tests that each executes one lifecycle event on a single OCP version. Basically, to see "after CC, this is exactly what happens".

alvaroaleman:
> In general, one thing that's generally hard for me is reasoning over the full config that includes all versions, so I'd be really happy if we could have a set of small tests that each executes one lifecycle event on a single OCP version. Basically, to see "after CC, this is exactly what happens".

We can build tests like that, however the logic for "latest GA" depends on multiple events, as the info if a given GA is the latest doesn't exist explicitly, just implicitly.

cmd/bugzilla-config-manager/main.go
alvaroaleman: No strong feelings on that one way or another, it is just what I found in the existing config

cmd/bugzilla-config-manager/main.go
alvaroaleman: It might be easier to test if it doesn't though and the helper tool could just use "Ran $configmanager name\n\n$OUTPUT" for the commit message I think?

cmd/bugzilla-config-manager/main.go
alvaroaleman: Note though that there can only be one version where `open` is the latest event, as that is the basis for configuring master/main so what is currently in our BZ config can't be modeled without adding X-1 logic to the tool.

alvaroaleman:
> Basically, to see "after CC, this is exactly what happens".

I've added more granular testcases for the relevant scenarios.

cmd/bugzilla-config-manager/main.go
petr-muller: Certainly, it's a tradeoff for the opportunity to have better commit messages, because the managers know what they are actually enforcing - but I'm not sure if that's valuable enough. I don't have strong opinion about it. Let's start with the easier route (yours).

cmd/bugzilla-config-manager/main.go
petr-muller: Yeah, just a single branch with `open` as latest past event, but it should be possible to have ones with `open` as a *future* event?

alvaroaleman: So from the description in the card, future events are irrelevant for this configmanager, did I get that wrong?

alvaroaleman:
/hold cancel
@petr-muller anything missing?

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,197 @@
...
+		if ocpProduct != "ocp" {
+			continue
+		}
+
+		for productVersion, config := range productConfig {
+			if len(config.Previous) < 1 {
+				continue
+			}
+
+			// TODO: We assume the first event is the last - Is that reasonable?
```
stevekuznetsov, line 115: Yes, I this should be documented as the spec.

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,197 @@
...
+			}
+		}
+	}
+
+	// GA versions:
+	// * DependentBugTargetReleases: "$VERSION+1.0"
+	// * if not latestGA: add $Version+1.z to DependentBugTargetReleases
+	// * TargetRelease: $VERSION.z
+	// * ValidateByDefault: true
+	sort.Slice(gaVersions, func(i, j int) bool { return gaVersions[i].String() < gaVersions[j].String() })
```
stevekuznetsov, line 177: Does `4.11` sort correctly w.r.t `4.1` ?

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,197 @@
...
+			// TODO: We assume the first event is the last - Is that reasonable?
+			switch config.Previous[0].Event {
+			case ocplifecycle.LifecycleEventGenerallyAvailable:
+				parsedVersion, err := ocplifecycle.ParseMajorMinor(productVersion)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to parse %s as majorMinor version: %w", productVersion, err))
+					continue
+				}
+				gaVersions = append(gaVersions, *parsedVersion)
+			case ocplifecycle.LifecycleEventOpen:
```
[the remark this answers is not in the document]
alvaroaleman: The latest event describes the expected current state. If we go back in the events we will find an `GA,Open` for all released versions, but that is irrelevant. In general, I think the config managers should be able to reconcile based on the current desired state and not depend on having reconciled a prior desired state, that would make everything super fragile.

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
alvaroaleman: Yeah seems so, I've added an explicit testcase for that, the generated `/testdata/zz_fixture_TestRun_lifecycle_multiple_ga_411_biger_41.yaml.yaml` has the `z` `dependent_bug_target_release` for `4.1` and not `4.11`.

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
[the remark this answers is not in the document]
stevekuznetsov: I think I disagree - we know that after open but before GA, a certain set of rules apply. If we, like you say, only look at the latest level we need to know all possible levels between GA an open for a verison, in order to know what to do, right ... ?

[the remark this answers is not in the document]
alvaroaleman: > I think I disagree - we know that after open but before GA, a certain set of rules apply. If we, like you say, only look at the latest level we need to know all possible levels between GA an open for a verison, in order to know what to do, right ... ?

Could you then please describe these rules in https://issues.redhat.com/browse/DPTP-2441? As of written, everything can be inferred from the current phase without knowledge of history and phases are linear so we always know what came before and what will come next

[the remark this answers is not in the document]
stevekuznetsov: Is that not already exactly what the card says?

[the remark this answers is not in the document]
stevekuznetsov: > Before the version X.Y is open for development, ...

> After version X.Y is open for development ...

> after version X.Y GA, ...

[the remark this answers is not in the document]
alvaroaleman: > Before the version X.Y is open for development, ...

That case doesn't actually have a phase and @petr-muller said its best to ignore it: https://github.com/openshift/ci-tools/pull/2272#discussion_r693001776

> After version X.Y is open for development ...
> after version X.Y GA, ...

Yeah those two cases have rules but they are derived from "last state is {open, ga}" and they don't care about what the state before that one was. Can you please point out concretely in which case we care about the history and can not derive the current configuration from the last state?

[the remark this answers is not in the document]
petr-muller: I had a conversation with Steve about this in the past, and in a sense he is right, it's a tradeoff. I personally think that a manager should simply either know exactly what phase we are in and what should hold  in that phase, or do nothing (if we are in a phase where we do not know what should hold, do nothing). 

Steve pointed out that this makes you vulnerable to changes: if a program adds a phase, we need to make all managers aware of that phase, even if they need need to enforce the criteria of the previous phase. So what we could do is, if you find yourself in a phase where you have no separate criteria, you'd work backwards until you find a phase that you know, and enforce that.  The expense is that the observable logic is more complicated and the simple idea of "in a phase x, enforce y" ceases to hold.

> That case doesn't actually have a phase and @petr-muller said its best to ignore it: #2272 (comment)

I think that's a misunderstanding - that definitely is a "phase". You'd see this as a release with no past event and an `open` event in the future: that's certainly actionable. What I meant by that feedback is that we should not infer the state of 4.X+1 from the lifecycle information of 4.X. The 4.X+1 release should have its own lifecycle record.

petr-muller:
So one more bit of feedback, seeing this in action - I think my original design of the lifecycle data format sucks, because it *assumes* a point in time where it holds: if the dates do not move at all, but we cross an event, the lifecycle data will change (an event moves from future to the past). This makes one very useful use case unnecessarily complicated - date injection. I was missing an easy way to do `--date YYYY-MM-DD` option when testing (and it would be useful even in production, to preview "future" changes). With the current format, we would need to have some recalculation code.

How about we ditch that crap and simply track a sorted list of events for each release. It would make the format simpler, date injection easier (the manager would quite simply need to find the position between two events based on the current day). We would use today as a default day and allowed to override it. WDYT?

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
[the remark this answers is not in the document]
stevekuznetsov: Also idk what if we deploy this manager after feature freeze, so open is not the next phase? It doesn't start "doing anything" until GA? The logic of this particular manager iterating through the list to figure out the last applicable thing is not that complex and makes it resilient to when it starts running or if it somehow misses a level ...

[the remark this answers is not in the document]
alvaroaleman: > Steve pointed out that this makes you vulnerable to changes: if a program adds a phase, we need to make all managers aware of that phase, even if they need need to enforce the criteria of the previous phase. So what we could do is, if you find yourself in a phase where you have no separate criteria, you'd work backwards until you find a phase that you know, and enforce that

But this is a potentially dangerous assumption, isn't it? We assume "unknown phase" means "enforce most recent known phase" but we can not say today if that is correct or not. So doing nothing or failing seem better options to me. If we know that phase X means "skip it and realize phase X-N" then we can explicitly add code for that.

> . What I meant by that feedback is that we should not infer the state of 4.X+1 from the lifecycle information of 4.X. The 4.X+1 release should have its own lifecycle record

Ah ok. So should that handling exist? The jira issue doesn't mention it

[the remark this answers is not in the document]
petr-muller: > But this is a potentially dangerous assumption, isn't it? 

Yeah, that was exactly what I was thinking. But at the same time, I realize that I have actually described the acceptance criteria in https://issues.redhat.com/browse/DPTP-2441 exactly like Steve proposes:

> Before the version X.Y is open for development, ensure that branches {openshift,release}-X.Y are validated by default with the following criteria:

^^^ This actually needs to hold before the branch is open, *after* the branch is open, durihng feature freeze and also during code freeze. But I did not bother to describe that explicitly. So I think Steve has a point.

> Ah ok. So should that handling exist? The jira issue doesn't mention it
I think it's okay to not have it for now.  There is very little config that's actually *needed* to exist before we even open a branch for development. Our current docs make sure we prepare future config in advance to make future switches easier, and I described the acceptance criteria based on that. But in a world where future switches are automated, preparing config "for the future" may be useless.

alvaroaleman:
> How about we ditch that crap and simply track a sorted list of events for each release. It would make the format simpler, date injection easier (the manager would quite simply need to find the position between two events based on the current day). We would use today as a default day and allowed to override it. WDYT?

Sure. @stevekuznetsov wdyt?

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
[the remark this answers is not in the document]
stevekuznetsov: It just seems like if we're going to enforce this "know exactly what to do in every phase" logic it's fragile because:

1. most of our config doesn't actually work like this, we latch at decision points and hold until new decision points - most config managers are not doing something different in every event
2. the release team has talked about increase the number of feature freeze events - one for teams that depend on the rebase, one for the rebase, one for unrelated things .... why should we need to recompile every manager every time?
3. if managers do nothing when they don't know the exact phase, it means we a) cannot deploy new managers in that phase and b) cannot add new repos in that phase

IMHO if we're going to be writing managers at random points in time they need to know how to start enforcing state at the moment they are deployed, not the next time in the next release they have something to do. Similarly, if a new repo is added that matches the criteria for the manager, it needs to make that repo conformant immediately, not in the next phase it happens to know something about.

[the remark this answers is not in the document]
stevekuznetsov: Like I honestly don't see the alternative - re-compile every manager every time we change the event list? Really?

stevekuznetsov:
> a sorted list of events for each release

Where future events have no date attached? So just get rid of the previous/future distinction?

petr-muller:
Right, optional dates... that's why I did that. Dammit. But I still think the logic is doable: given any input date, events without date are assumed to be in the future (I think there's no reason to have past events without dates), unless they are *provably* in the past (there is a past event *after* them). The events are expected to be sorted timewise.

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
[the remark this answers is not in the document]
petr-muller: > So for all states other than GA or end of life?
Yes.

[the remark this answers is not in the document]
petr-muller: I think I am convinced

stevekuznetsov:
Sounds good to me

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,236 @@
...
+			config.TargetRelease = utilpointer.String(developmentVersion.String() + ".0")
+
+			cfg.Default[developmentBranchName] = config
+		}
+
+		// nonEOLNonGAVerion: Set config for {release/openshift}-$VERSION:
+		// * dependentBugStatus: ["MODIFIED", "ON_QA", "VERIFIED"]
+		// * DependentBugTargetReleases: "$VERSION+1.0"
+		// * ValidateByDefault: true
+		for _, nonEOLNonGAVerion := range nonEOLNonGAVersions {
```
stevekuznetsov, line 199: nit: why is this under `if developmentVersion != nil` ?

stevekuznetsov (commented):
One question, otherwise LGTM

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
alvaroaleman: Good catch, fixed

cmd/branchingconfigmanagers/bugzilla-config-manager/main.go
```diff
@@ -0,0 +1,236 @@
...
+
+				if event.Event == ocplifecycle.LifecycleEventGenerallyAvailable {
+					gaVersionsSet.Insert(productVersion)
+					break
+				}
+
+				if event.Event == ocplifecycle.LifecycleEventOpen {
+					if developmentVersionFound {
+						errs = append(errs, fmt.Errorf("found multiple development versions: %s and %s", developmentVersion, productVersion))
+					}
```
petr-muller, line 151: I think we'll falsely hit this error when we enforce the same criteria for versions in code freeze and the one that is open (like 4.9 and 4.10 are now). Timewise, the "older" release is not open, it just enforces the same criteria because last event that had any effect was "open".

alvaroaleman: So much about being able to automatically do the correct thing by just falling through until a known condition is hit. Updated and added a testcase.

petr-muller (approved):
LGTM

I tested this on our current config and several future transitions, and it does the right thing.

@alvaroaleman thanks a lot for your help and for your patience!

PR #2277, opened 2021-08-20
DPTP-2460: Allow Round-Tripping Secret Bootstrapper Configuration

smg247:
In order to implement this I opted to retrain the cluster_group info in `secretbootstrap#resolve` in order to more easily retain the structure of the config file when Marshaling.


cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -440,7 +440,7 @@ func TestCompleteOptions(t *testing.T) {
 				ClusterGroups: map[string][]string{"group-a": {"default"}},
 				Secrets: []secretbootstrap.SecretConfig{{
 					From: map[string]secretbootstrap.ItemContext{"key-name-1": {Item: "item-name-1", Field: "field-name-1"}},
-					To:   []secretbootstrap.SecretContext{{Cluster: "default", Namespace: "ns", Name: "name"}},
+					To:   []secretbootstrap.SecretContext{{ClusterGroups: []string{"group-a"}, Cluster: "default", Namespace: "ns", Name: "name"}},
```
stevekuznetsov, line 443: Why did this need to change?

stevekuznetsov: (Was the previous test wrong? Do we need an additional test?)

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -19,7 +21,7 @@ const (
 )
 
 type ItemContext struct {
-	Item                 string                 `json:"item"`
+	Item                 string                 `json:"item,omitempty"`
```
stevekuznetsov, line 25: This is technically correct why?

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -45,7 +47,7 @@ type DockerAuth struct {
 
 type SecretContext struct {
 	// A cluster to target. Mutually exclusive with 'ClusterGroups'
-	Cluster string `json:"cluster"`
+	Cluster string `json:"cluster,omitempty"`
```
stevekuznetsov, line 51: This is correct since a group may be set? Do we have validation to ensure that `cluster = ""` ensures a group is set?

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -91,6 +102,64 @@ func (c *Config) UnmarshalJSON(d []byte) error {
 	return c.resolve()
 }
 
+func (c *Config) MarshalJSON() ([]byte, error) {
+	target := &configWithoutUnmarshaler{
+		VaultDPTPPrefix:           c.VaultDPTPPrefix,
+		ClusterGroups:             c.ClusterGroups,
```
stevekuznetsov, line 109: Technically this is being copied by reference, so this marshal will change the source `Config` - not sure if we care?

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -91,6 +102,64 @@ func (c *Config) UnmarshalJSON(d []byte) error {
...
+		}
+	}
+
+	s.To = secrets
+}
+
+func (c *Config) stripVaultPrefix(s *SecretConfig) {
+	pre := c.VaultDPTPPrefix + "/"
+	for key, from := range s.From {
+		from.Item = strings.Replace(from.Item, pre, "", 1)
```
stevekuznetsov, line 156:
```go
strings.TrimPrefix()
```
?

pkg/api/secretbootstrap/secretboostrap_test.go
```diff
@@ -209,6 +217,41 @@ func TestLoadConfigFromFile(t *testing.T) {
...
+				t.Errorf("error loading config file: %v", err)
+			}
+
+			outFile := filepath.Join("testdata", fmt.Sprintf("%s_out.yaml", t.Name()))
+			err = SaveConfigToFile(outFile, c)
+			if err != nil {
+				t.Errorf("error saving config file: %v", err)
+			}
+
+			in, _ := ioutil.ReadFile(inFile)
```
stevekuznetsov, line 244: `t.Fatal()` on these errors :)

pkg/api/secretbootstrap/secretboostrap_test.go
```diff
@@ -209,6 +217,41 @@ func TestLoadConfigFromFile(t *testing.T) {
...
+			out, _ := ioutil.ReadFile(outFile)
+			if !bytes.Equal(in, out) {
```
stevekuznetsov, line 246:
```go
if diff := cmp.Diff(in, out); diff != "" {
    t.Errorf("input and output configs are not equal: %s", diff)
}
```

pkg/api/secretbootstrap/secretboostrap_test.go
```diff
@@ -209,6 +217,41 @@ func TestLoadConfigFromFile(t *testing.T) {
...
+				t.Errorf("input and output configs are not equal. See output file: %s", outFile)
+			} else {
+				_ = os.Remove(outFile)
```
stevekuznetsov, line 249: Register this on `t.Cleanup()` and handle the error

pkg/api/secretbootstrap/secretboostrap_test.go
```diff
@@ -209,6 +217,41 @@ func TestLoadConfigFromFile(t *testing.T) {
...
+	}{
+		{
+			name: "basic base",
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			c := &Config{}
+			inFile := filepath.Join("testdata", fmt.Sprintf("%s.yaml", t.Name()))
```
stevekuznetsov, line 232: Use `testhelper.ReadFromFixture()`

pkg/api/secretbootstrap/testdata/TestRoundtripConfig/basic_base.yaml
```diff
@@ -0,0 +1,58 @@
+cluster_groups:
```
stevekuznetsov, line 1: So notably we do not have overlapping groups here, right? If we want to make a statement that there should never be an item that contains the same cluster from more than one group, we should enforce that via validation

cmd/ci-secret-bootstrap/main_test.go
smg247: we are  intentionally retaining cluster_group info now in the `SecretContext`. This is so we can determine if it was originally part of a group or not. In the latter case we don't want to group it in the output. I don't _think_ retaining this info will cause any issues, but the test's expectation did need to change in order to specify that we do expect it to be there.

pkg/api/secretbootstrap/secretboostrap.go
smg247: When `DockerConfigJSONData` is present there will not be an `Item`. Before I made this change it was outputting `null` in the Marshal process.

pkg/api/secretbootstrap/secretboostrap.go
smg247: I don't believe so, I will research and add that if needed.

pkg/api/secretbootstrap/secretboostrap.go
smg247: I don't think we care currently. But, it would probably be better if it didn't do that...

pkg/api/secretbootstrap/secretboostrap.go
smg247: Thanks!

pkg/api/secretbootstrap/testdata/TestRoundtripConfig/basic_base.yaml
smg247: I'm not sure I understand exactly what you mean here. I made the test config intentionally contain cases that are possible, but not currently within the actual data. I am fine with adding that validation and removing the case from the test config, but wasn't sure if we would _never_ want that.

stevekuznetsov: I don't see a config entry that contains both `build_farm` and `cg_2` or `build_farm` and `cg_3` - what entry currently overlaps?

pkg/api/secretbootstrap/secretboostrap.go
stevekuznetsov: OK, cool - that's another good validation. One of the two must be set.

pkg/api/secretbootstrap/secretboostrap_test.go
smg247: Ah thats useful!

pkg/api/secretbootstrap/testdata/TestRoundtripConfig/basic_base.yaml
smg247: Oh I get what you are saying now! I'll try that out and, assuming it doesn't function properly, add some validation, thanks

pkg/api/secretbootstrap/secretboostrap_test.go
smg247: Correct me if I'm wrong, but I think I actually want to use the `secretbootstrap.LoadConfigFromFile` function here to get all the custom Unmarshaling. I could get the same result using `testhelper.ReadFromFixture()` and following that up with `yaml.Unmarshal()` but that seems like more effort

pkg/api/secretbootstrap/testdata/TestRoundtripConfig/basic_base.yaml
smg247: I modified the cluster_groups to create an overlap in `cg_2` and `cg_3`, and the test still passes so it seems like we are good here

pkg/api/secretbootstrap/secretboostrap.go
smg247: Evidently a corresponding validation for this exists in `cmd/ci-secret-bootstrap/main.go`. I am wondering if I should pull this validation into this file and remove the existing one?

pkg/api/secretbootstrap/secretboostrap.go
smg247: This validation also currently exists in `cmd/ci-secret-bootstrap/main.go`.

smg247:
@stevekuznetsov the requested validations are present in `cmd/ci-secret-bootstrap/main.go`. I think it's probably okay to just leave them there, but I can move them up to this file if that is preferred. If I try to do them in both places I will have to ensure that the error message is identical or else will have to modify the unit tests in `cmd/ci-secret-bootstrap/main_test.go` to expect the error from this file. That seems hacky, and I would rather not do that.

smg247:
Please excuse the commit history...I thought I would be able to clean it up by rebasing off of `ci-tools/master`, but that didn't work out as planned...

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -91,6 +99,67 @@ func (c *Config) UnmarshalJSON(d []byte) error {
...
+
+	s.To = secrets
+}
+
+func (c *Config) stripVaultPrefix(s *SecretConfig) {
+	pre := c.VaultDPTPPrefix + "/"
+	for key, from := range s.From {
+		from.Item = strings.TrimPrefix(from.Item, pre)
+		for i, dcj := range from.DockerConfigJSONData {
+			from.DockerConfigJSONData[i].Item = strings.Replace(dcj.Item, pre, "", 1)
```
stevekuznetsov, line 157: Is this not also `strings.TrimPrefix` ?

pkg/api/secretbootstrap/secretboostrap_test.go
stevekuznetsov: What you get for free with `testhelper.ReadFromFixture()` is the test data file name management - it has utility in being central and common. Calling `yaml.Unmarshal` is the expected cost of Go's verbosity.

pkg/api/secretbootstrap/secretboostrap_test.go
stevekuznetsov: This didn't get addressed

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -91,6 +99,67 @@ func (c *Config) UnmarshalJSON(d []byte) error {
...
+		c.stripVaultPrefix(&secret)
+		secret.groupClusters()
+		secrets = append(secrets, secret)
+	}
+
+	target.Secrets = secrets
+	return json.Marshal(target)
+}
+
+func (s *SecretConfig) groupClusters() {
```
stevekuznetsov, line 123: Could we get some unit tests for this?

pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -91,6 +99,67 @@ func (c *Config) UnmarshalJSON(d []byte) error {
...
+		} else {
+			// This cluster was not from a group, so nothing special needs to be done
+			secrets = append(secrets, to)
+		}
+	}
+
+	s.To = secrets
+}
+
+func (c *Config) stripVaultPrefix(s *SecretConfig) {
```
stevekuznetsov, line 152: The `c *Config` already contains many `*SecretConfig` - why are they getting passed in here? This data flow is unexpected.

pkg/api/secretbootstrap/secretboostrap_test.go
smg247: Looking into what happened here, I have it changed to `Fatal` locally...

pkg/api/secretbootstrap/secretboostrap.go
smg247: Yeah this is a bit of a weird one...I was more interested in making the function have a receiver to match all the others in the file than actually making it make any sense. Let me clean that up.

pkg/api/secretbootstrap/secretboostrap_test.go
```diff
@@ -209,6 +218,167 @@ func TestLoadConfigFromFile(t *testing.T) {
...
+				Name:          "a",
+			}},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			sc := tc.input
+			sc.groupClusters()
+			if !reflect.DeepEqual(sc.To, tc.expected) {
```
petr-muller, line 375: can we use the `diff := cmp.Diff(in, out); diff != "` pattern here too? I think all our new tests use `cmp` and not `reflect`.

smg247: ahhh didn't think about using that here as well, thanks for the info!

petr-muller (approved):
/hold
LGTM, holding for Steve

smg247:
@stevekuznetsov, one last look?

PR #2278, opened 2021-08-23
prowgen: disallow cluster_claim on non-multistage tests

petr-muller:
IMO there's no reason why to allow this, it will just cause confusion if
anyone uses it this way. Tests that want a cluster should be using our
multi-stage features, otherwise users will likely hit various
limitations where they cannot use built images, custom credentials etc.


stevekuznetsov (commented):
Can we get a unit test for this?

hongkailiu:
@petr-muller Do you want me to remove the code in `ci-operator` to handle the non-prowgen jobs as well?

pkg/validation/test.go
```diff
@@ -486,40 +489,33 @@ func (v *Validator) validateTestConfigurationType(
 	var needsReleaseRpms bool
 	if testConfig := test.OpenshiftAnsibleClusterTestConfiguration; testConfig != nil {
 		typeCount++
-		clusterCount++
```
petr-muller, line 489: These were previously used to compute "count of installed clusters", but the new validation targets the problematic state directly.

petr-muller:
> Do you want me to remove the code in ci-operator to handle the non-prowgen jobs as well?

I assume you mean non-multistage tests. I guess yes? I thought this is just an omission in validation, but if ci-op can handle it, I guess the support was intentional? Do we see value for it?

hongkailiu:
> I assume you mean non-multistage tests.
I misunderstood the range of tests this PR disallows.
I thought non-multistage tests == container tests.
From the added unit tests, we have more than container tests.
Yes, that means we have a bug on the validation.
 

> I guess the support was intentional? Do we see value for it?

The intension was we-could-do-it-why-not without seeing the limits.
I do not have a case where we have to use container test, not even mentioning a container test with a claim.
I will remove the code for container tests as a follow up of this one, just to keep our code base clean.
Also make it clear in our doc.

stevekuznetsov (approved):

PR #2283, opened 2021-08-23
Load kubeconfig files from a directory

hongkailiu:

/hold
wait for https://github.com/openshift/ci-tools/pull/2282

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -49,6 +49,7 @@ type options struct {
 	validateItemsUsage bool
 
 	kubeConfigPath      string
+	kubeConfigDir       string
 	configPath          string
 	generatorConfigPath string
 	cluster             string
@@ -79,6 +80,7 @@ func parseOptions(censor *secrets.DynamicCensor) (options, error) {
 	fs.BoolVar(&o.validateItemsUsage, "validate-bitwarden-items-usage", false, fmt.Sprintf("If set, the tool only validates if all fields that exist in Vault and were last modified before %d days ago are being used in the given config.", allowUnusedDays))
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
 	fs.StringVar(&o.kubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
+	fs.StringVar(&o.kubeConfigDir, "kubeconfig-dir", "", "Path to the directory containing kubeconfig files to use for CLI requests.")
 	fs.StringVar(&o.configPath, "config", "", "Path to the config file to use for this tool.")
 	fs.StringVar(&o.generatorConfigPath, "generator-config", "", "Path to the secret-generator config file.")
 	fs.StringVar(&o.cluster, "cluster", "", "If set, only provision secrets for this cluster")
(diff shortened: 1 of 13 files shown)


alvaroaleman:
/retest
@hongkailiu is this supposed to be marked as WIP?

hongkailiu:
> /retest
> @hongkailiu is this supposed to be marked as WIP?

I am thinking adding some unit tests for load function from dir.
Should be ready soon.

pkg/util/client.go
```diff
@@ -48,35 +49,43 @@ func LoadKubeConfig(path string) (*rest.Config, error) {
 
 // LoadKubeConfigs loads kubeconfigs. If the kubeconfigChangedCallBack is non-nil, it will watch all kubeconfigs it loaded
 // and call the callback once they change.
-func LoadKubeConfigs(kubeconfig string, kubeconfigChangedCallBack func(fsnotify.Event)) (map[string]*rest.Config, string, error) {
-	loader := clientcmd.NewDefaultClientConfigLoadingRules()
-	loader.ExplicitPath = kubeconfig
-	cfg, err := loader.Load()
+func LoadKubeConfigs(kubeconfig, kubeconfigDir string, kubeconfigChangedCallBack func(fsnotify.Event)) (map[string]*rest.Config, error) {
+	var errs []error
+	configs, err := kube.LoadClusterConfigs(kube.NewConfig(kube.ConfigFile(kubeconfig),
```
alvaroaleman, line 54: Do we still load the configs from env if `KUBECONFIG` is set?

hongkailiu: If both `kubeconfig` and `kubeconfigDir` are empty, it will loads `$KUBECONFIG`.

https://github.com/kubernetes/test-infra/blob/93ddf012c30d14168b62a9e9dcffd8f546b1d7f6/prow/kube/config.go#L117

But if we found file, for example, from `kubeconfigDir`, then `$KUBECONFIG` is ignored.
Comparing to previous implementation, this does not change.

hongkailiu: In our use case, we either set $KUBECONFIG or use kubeconfigDir, not both, right?

cmd/cluster-display/main.go
```diff
@@ -169,7 +169,7 @@ func main() {
 		interrupts.Terminate()
 	}
 
-	kubeConfigs, _, err := util.LoadKubeConfigs(o.hiveKubeconfigPath, kubeconfigChangedCallBack)
+	kubeConfigs, err := util.LoadKubeConfigs(o.hiveKubeconfigPath, "", kubeconfigChangedCallBack)
```
stevekuznetsov, line 172: Didn't you write the utility upstream with the express design of having composable options, so nobody ever passes `""` like this?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -79,6 +80,7 @@ func parseOptions(censor *secrets.DynamicCensor) (options, error) {
 	fs.BoolVar(&o.validateItemsUsage, "validate-bitwarden-items-usage", false, fmt.Sprintf("If set, the tool only validates if all fields that exist in Vault and were last modified before %d days ago are being used in the given config.", allowUnusedDays))
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to actually create the secrets with oc command")
 	fs.StringVar(&o.kubeConfigPath, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
+	fs.StringVar(&o.kubeConfigDir, "kubeconfig-dir", "", "Path to the directory containing kubeconfig files to use for CLI requests.")
```
stevekuznetsov, line 83: Can we use upstream k8s options?

cmd/dptp-controller-manager/main.go
```diff
@@ -107,6 +108,7 @@ func newOpts() (*options, error) {
 	} else {
 		flag.StringVar(&opts.kubeconfig, "kubeconfig", "", kubeconfigFlagDescription)
 	}
+	flag.StringVar(&opts.kubeconfigDir, "kubeconfig-dir", "", "Path to the directory containing kubeconfig files. All contexts in it will be considered a build cluster. If it does not have a context named 'app.ci', loading in-cluster config will be attempted.")
```
stevekuznetsov, line 111: Can we use upstream k8s options?

stevekuznetsov (commented):
Two major comments:

1. can we use the upstream flag utilities
2. can we make use of the options pattern you wrote

PR #2284, opened 2021-08-23
WIP: Prowgen refactor for reusability

petr-muller:
Early WIP. So far refactored `PodSpec`-building code into a generator (code is not polished yet, but IMO is now better structured, and individual PodSpec building blocks are now coded only once, are (mostly -- except templates some template) independent on each other, and IMO better testable.

It is supposed to be a refactoring, hence it should not fail `breaking-changes`.


petr-muller:
@bbguimaraes @stevekuznetsov I believe this is now ready for review. The change looks scary, but most of the content are test fixtures.

It also notably does not fail `breaking-changes`, which serves as a proof that this is a pure refactoring, without any functional
changes.

petr-muller:
/test integration
r-c outage

pkg/prowgen/podspec.go
```diff
@@ -0,0 +1,412 @@
...
+	templateTestIST    string
+
+	targets sets.String
+	secrets []cioperatorapi.Secret
+}
+
+// CiOperatorPodSpecGenerator is a builder-pattern for building PodSpecs that
+// run ci-operator for various purposes, ensuring the consistency of volumes,
+// mounts, parameters and other content.
+type CiOperatorPodSpecGenerator interface {
```
stevekuznetsov, line 93: This is backwards from the normal builder pattern, right? Each new functionality must be added to this mega-interface and as fields to the data struct. Instead, couldn't we do something like

```go
type mutator func(*v1.PodSpec) error

type ciOperatorPodSpecGenerator struct {
    meta cioperatorapi.Metadata
    mutators []mutator
}

func NewCIOperatorPodSpecGenerator(meta cioperatorapi.Metadata, mutators ...mutator) *ciOperatorPodSpecGenerator {
    // ...
}

func CIPullSecret() mutator {
    return func(spec *v1.PodSpec) error {
        handleSecret(spec, cioperatorapi.Secret{Name: "ci-pull-credentials"})
        return nil
    }
}
```

Then the application is encapsulated to any individual thing, as long as they're independent?

stevekuznetsov: (you already have a good encapsulation of intent in the `handle*` functions, but unless we need complex multi-option dependencies we should be able to get away without a mega-struct of options?)

petr-muller: Hmm...  Interesting, I know *this* to be a normal builder (I agree it's OOPy) - the job level plays the role of a director, you accumulate internal state and eventually build the whole thing.

I like your pattern and the suggestion to get rid of the interface and internal state, but I think some of the things needed here would be hard if we simply iterated mutators over shared `PodSpec` state. There ** are** some dependencies between individual mutators (mostly related to templates: templates need cluster profiles, must have one target, some need RPMs, but RPMs do not make sense outside of templates...), plus accumulating simple intents makes it easier to be idempotent (e.g. to avoid mistakenly add one secret or target twice). It's also easier to do some sanity checking to error out on the final `PodSpec` - this is easier to do over intents than over the final product.

I originally had something like you propose - the builder operated over an internal PodSpec and eventually just returned it, but it was really annoying to work with internally.

petr-muller: I wonder if I could simplify the thing at least partially though... Parts of the internal state could be replaced with easy non-state mutators like you suggest. The interface could be generalized to accept something abstract...

stevekuznetsov: OK, if we have complicated inter-mutator dependencies (ew) then we can't do anything super simple like I had in mind. Let me know if you decide to change anything or if you want to just review as-is

petr-muller: They are not that complicated, but there are deps. I have a partial change to make this PR simpler, I hope to update the PR today.

pkg/prowgen/podspec_test.go
```diff
@@ -0,0 +1,414 @@
...
+			t.Parallel()
+			podspec, err := NewCiOperatorPodSpecGenerator(meta).ClusterProfile(tc).Targets("test-name").Build()
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+			testhelper.CompareWithFixture(t, podspec)
+		})
+	}
+
+	errorTests := []struct {
```
stevekuznetsov, line 326: nit: which errors in this file cannot be described with this three-member struct? What's the utility of copy-pasting the test setup and parallel execution logic in so many individual tests instead of just using subtests in one func?

pkg/prowgen/testdata/zz_fixture_TestClusterProfile_gcp_logging_journald.yaml
```diff
@@ -0,0 +1,42 @@
...
+- name: result-aggregator
+  secret:
+    secretName: result-aggregator
+- name: cluster-profile
+  projected:
+    sources:
+    - secret:
+        name: cluster-secrets-gcp
+    - configMap:
+        name: cluster-profile-gcp-logging-journald
```
stevekuznetsov, line 42: What do we gain by testing every single possible cluster profile?

pkg/prowgen/testdata/zz_fixture_TestReleaseRpms_envvar_additional_envvar_generated_for_template.yaml
```diff
@@ -3,20 +3,21 @@ containers:
   - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
   - --gcs-upload-secret=/secrets/gcs/service-account.json
   - --report-credentials-file=/etc/report/credentials
-  - --target=test
-  - --secret-dir=/usr/local/test-cluster-profile
-  - --template=/usr/local/test
+  - --target=tgt
```
stevekuznetsov, line 6: if possible, a zero-diff would be good where the functionality did not actually change

stevekuznetsov (commented):
I guess the breaking changes presubmit would be the best test here (and we may want to roll out some medium-term change to the existing logic which sorts fields to help that pass) but some of the diffs in the test fixtures are hard to parse here.

petr-muller:
Yeah sorry, the changes pushed here are just partial, I use the PR to sync code across devices. Sorry for not marking it as such (I just marked the new commit). Some tests and fixtures are wrong. As you say, `breaking-changes` pass is the ultimate judge and goal, and I will try to minimize the fixture changes. I think we *do* have sorting on save, the failures here are genuine (because of WIP, not sorting).

pkg/prowgen/testdata/zz_fixture_TestClusterProfile_gcp_logging_journald.yaml
petr-muller: Neverending annoyance TBH. Originally it was just a shortcut to get all relevant behaviors, but maintaining the PR taught me that lesson.

petr-muller:
I've added a separate PR to sort the fixtures in advance, so that the diff here is smaller: https://github.com/openshift/ci-tools/pull/2424

Once that PR is merged, I will rebase this PR and it is IMO reviewable. I'll post some notes to help with reviewing, too.

pkg/prowgen/testdata/zz_fixture_TestClaims_secret_is_added.yaml
```diff
@@ -3,8 +3,8 @@ containers:
   - --image-import-pull-secret=/etc/pull-secret/.dockerconfigjson
   - --gcs-upload-secret=/secrets/gcs/service-account.json
   - --report-credentials-file=/etc/report/credentials
+  - --variant=variant
   - --hive-kubeconfig=/secrets/hive-hive-credentials/kubeconfig
-  - --target=test
```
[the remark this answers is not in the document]
petr-muller: Looks like Git identified some of the fixtures as renames of previous fixtures, which leads to these weird pseudochanges. The actual fixture looks correct to me.

pkg/prowgen/podspec.go
```diff
@@ -0,0 +1,502 @@
...
+
+// NewCiOperatorPodSpecGenerator returns a new CiOperatorPodSpecGenerator instance
+func NewCiOperatorPodSpecGenerator() CiOperatorPodSpecGenerator {
+	return &ciOperatorPodSpecGenerator{}
+}
+
+func aggregateMutator(mutators ...PodSpecMutator) PodSpecMutator {
+	return func(spec *corev1.PodSpec) error {
+		var errs []error
+		for _, f := range mutators {
```
droslean, line 97: There is a chance here that one mutator can revert changes the happened through another mutator. Do we care?

pkg/prowgen/podspec.go
```diff
@@ -0,0 +1,502 @@
...
+			errs = append(errs, f(spec))
+		}
+		return kerrors.NewAggregate(errs)
+	}
+}
+
+func (c *ciOperatorPodSpecGenerator) Add(f ...PodSpecMutator) CiOperatorPodSpecGenerator {
+	c.mutators = append(c.mutators, f...)
```
droslean, line 105: I would prefer to lock the object before appending here.

pkg/prowgen/podspec.go
```diff
@@ -0,0 +1,502 @@
...
+		container.Env = append(container.Env, want)
+	}
+	return nil
+}
+
+func addUniqueParameter(container *corev1.Container, wantArg string) {
+	var foundArg bool
+	for i := range container.Args {
+		if container.Args[i] == wantArg {
+			foundArg = true
```
droslean, line 130: we can `break` the loop here since the arg was found, or even better we can just `return` and the if-not-found check below won't be needed.

pkg/prowgen/podspec.go
```diff
@@ -0,0 +1,502 @@
...
+			return nil
+		}
+		container := &spec.Containers[0]
+		url := cioperatorapi.URLForService(cioperatorapi.ServiceRPMs)
+		var repoPath = fmt.Sprintf("%s/openshift-origin-v%s/", url, version)
+		if strings.HasPrefix(version, "origin-v") {
+			repoPath = fmt.Sprintf("%s/openshift-%s/", url, version)
+		}
+		return addEnvVar(container, corev1.EnvVar{
+			Name:  "RPM_REPO_OPENSHIFT_ORIGIN",
```
droslean, line 217: let's const all the env vars

PR #2285, opened 2021-08-23
prowgen: parallelize pruning

petr-muller:
I plagiarized the parallelization from `sanitize` so if it is wrong it
probably needs to fixed there too ;)

Before:

```console
$ time ci-operator-prowgen --from-dir ./ci-operator/config --to-dir ./ci-operator/jobs
...
________________________________________________________
Executed in   51.33 secs    fish           external
   usr time   63.30 secs  282.00 micros   63.30 secs
   sys time    3.01 secs   55.00 micros    3.01 secs
```

After:

```console
$ time ci-operator-prowgen --from-dir ./ci-operator/config --to-dir ./ci-operator/jobs

...
________________________________________________________
Executed in   41.72 secs    fish           external
   usr time   64.63 secs  250.00 micros   64.63 secs
   sys time    2.84 secs   49.00 micros    2.84 secs
```

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -7,10 +7,12 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"sync"
 
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	prowconfig "k8s.io/test-infra/prow/config"
 
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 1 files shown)


cmd/ci-operator-prowgen/main.go
```diff
@@ -170,25 +172,46 @@ func getReleaseRepoDir(directory string) (string, error) {
...
-		if len(pruned.PresubmitsStatic) == 0 && len(pruned.PostsubmitsStatic) == 0 && len(pruned.Periodics) == 0 {
-			if err := os.Remove(info.Filename); err != nil && !os.IsNotExist(err) {
-				return err
-			}
-		} else {
-			if err := jc.WriteToFile(info.Filename, pruned); err != nil {
-				return err
-			}
+	errReadingDone := make(chan struct{})
+	go func() {
```
droslean, line 179: This will run only to close the channel. Instead what we need here is a `for{ select{ ` and grab the case of `errChan` and then we can just append to `errs`. We can also send a `stop` channel to close the goroutin as well.

petr-muller: WDYM by "this will run only to close the channel"?

droslean: Oh, Go seems to wait when it ranges over a channel, so it generally makes sense to design it this way.  Sorry for the noise either way.

petr-muller: Not noise, made me recheck at least :) Good thing.

stevekuznetsov (approved):

PR #2286, opened 2021-08-24
Script to easily run e2e tests, and a make target for debug

smg247:
When running e2e tests on a mac there are a couple of things that need to be set up. This new script performs those, and is also useful for simply running an e2e test with verbose output.

Also included a new make target that is useful for building the code for debugging.


hack/e2e-test.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
petr-muller, line 1: I think the `local-e2e` target was supposed to serve this purpose - can we merge the two if the existing target does not work?

smg247: This script utilizes the `local-e2e` target, but you are right it is probably preferable to have that target call this script for setup instead. I was thinking of this as more of a _helper_ that wasn't required to run the e2e tests locally, but just did some setup and extra logging. Do you think that this would be useful every time, and thus should be merged with the `local-e2e` target?

PR #2273, opened 2021-08-20
DPTP-2452 [ci-secret-bootstrap] Delete the stale keys from the existing secrets

droslean:
/cc @openshift/test-platform 

https://issues.redhat.com/browse/DPTP-2452

Signed-off-by: Nikolaos Moraitis <nmoraiti@redhat.com>

PR #2287, opened 2021-08-24
[ci-secret-bootstrap] Change the dryrun design by passing the options to the actual clients

droslean:
/cc @openshift/test-platform 

Before this change, the tool was only generating some YAML files containing the secrets that should be created/updated. This is not an efficient way to test the behavior, and also it is not being used anywhere. Since the `metav1.{Create,Update,Delete}Options` support running the client with dry-run, we can use that instead, which can be used as a presubmit to test any breaking changes as well.


alvaroaleman:
IMHO the purpose of the dry run is to be able to manually inspect what the tool would do without actually doing it. I can not think of a case where running the tool with a dry run kube client would help us debug something.

droslean:
For the record, I tested it. `--dry-run=true` runs the tool exactly as it runs without dry-run, but does not mutate any k8s object.

droslean:
> IMHO the purpose of the dry run is to be able to manually inspect what the tool would do without actually doing it. I can not think of a case where running the tool with a dry run kube client would help us debug something.

Perhaps we can keep the old `dry-run` logic, and add this kind of behavior in another flag maybe?

alvaroaleman:
> Perhaps we can keep the old dry-run logic, and add this kind of behavior in another flag maybe?

What is the motivation for wanting to run this with a dry run kube client?

droslean:
> What is the motivation for wanting to run this with a dry run kube client?

Currently, there is no way to test changes like: https://github.com/openshift/ci-tools/pull/2273

Also, the concept of dry-run should be like: run as you should run but don't mutate anything. right?

alvaroaleman:
> Currently, there is no way to test changes like: #2273

How does this help with testing those changes? I don't understand. What we want to test is is the logic is correct, i.E. we are deleting the right keys, not if our api calls pass server side dry run (which they will even if they delete all keys all the time).

droslean:
> How does this help with testing those changes? I don't understand. What we want to test is is the logic is correct, i.E. we are deleting the right keys, not if our api calls pass server side dry run (which they will even if they delete all keys all the time).

Yea it seems correct, but it would be nice to have at least a dry-run of that behavior to actually see the list of the stale keys that we will delete.

alvaroaleman:
> Yea it seems correct, but it would be nice to have at least a dry-run of that behavior to actually see the list of the stale keys that we will delete.

Ok but for that you just need to log the keys you are going to delete and not do the api call? That would be faster

droslean:
> Ok but for that you just need to log the keys you are going to delete and not do the api call? That would be faster

The current dry-run design doesn't even reach that code. We only construct the secrets and write them to files. There is no way to detect old keys.

PR #1510, opened 2020-12-09
prowgen: provide the GCS secret to tests

stevekuznetsov:
We need to forward the GCS upload secret to test namespaces to allow us
to mount it into sidecar processes in test Pods. ci-operator already
knows how to do this if it's provided via flag; this change just sets
the flag and provides the data to the owning ci-operator process.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

/assign @alvaroaleman @petr-muller

PR #2289, opened 2021-08-24
Add a flag with the gcs credentials secret name for the private jobs

droslean:
Private jobs are using a different secret name to update the artifacts. This PR:

- Adds a new flag in `ci-operator` to specify the name of the secret, but keeping the default `gce-sa-credentials-gcs-publisher`
- Make `prowgen` to append the extra argument in private jobs only.

Described in https://issues.redhat.com/browse/DPTP-2468

/cc @openshift/test-platform 


droslean:
This should be a follow-up of https://github.com/openshift/ci-tools/pull/1510 and https://github.com/openshift/ci-tools/pull/1480

droslean:
/hold

Holding until I have a follow-up PR to openshift/release

stevekuznetsov (commented):
Why does this add a new flag instead of setting the `--gcs-upload-secret` flag? Also, we will need to make sure that multi-stage tests use the correct secret in each test, as they use pod-utils as well.

droslean:
> Why does this add a new flag instead of setting the --gcs-upload-secret flag? Also, we will need to make sure that multi-stage tests use the correct secret in each test, as they use pod-utils as well.

Before we just hardcoded the secret name, and we ended up passing the wrong secret to the private jobs. Adding a new flag to specify the name of the secret is the simplest solution I thought of. The private jobs have specified the GCS secret in the decoration config, check this [pjyaml](https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/prowjob?prowjob=aa531592-0517-11ec-b53b-0a580a812243)

stevekuznetsov (approved):

PR #2293, opened 2021-08-25
Disallow `secrets` in non-container tests, improve validation tests

petr-muller:
- Validate `secret/secrets` usage on non-container tests
- ci-op validation: improve tests and fix nits

### DIFF
--- a/pkg/validation/test_test.go
+++ b/pkg/validation/test_test.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/google/go-cmp/cmp"
+
 	"k8s.io/apimachinery/pkg/util/sets"
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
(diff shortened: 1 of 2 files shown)


pkg/validation/test_test.go
```diff
@@ -237,43 +235,45 @@ func TestValidateTests(t *testing.T) {
...
+					Commands:                   "commands",
+					ContainerTestConfiguration: &api.ContainerTestConfiguration{From: "src"},
 					Secret: &api.Secret{
 						Name:      "secret",
 						MountPath: "/path/to/secret:exec",
 					},
 				},
 			},
-			expectedValid: false,
+			// TODO: The code actually just checks `filepath.IsAbs()`
```
bbguimaraes, line 253: Amazing.  Let's just make the test case value a relative path, I don't think anyone cares.

bbguimaraes (commented):
LGTM

PR #2294, opened 2021-08-25
prowgen: support overriding prowjob timeout

vrutkovs:
Add "Timeout" param to the test, which would reset `decoration_config.timeout` so that some tests could run longer than 4 hours.

Example job: https://github.com/openshift/release/pull/21355

TODO:
* [ ] Add a limit to the timeouts - perhaps `24h` would be sufficient for now?

Ref: https://issues.redhat.com/browse/DPTP-889


vrutkovs:
Limited job time to 8hrs

pkg/prowgen/prowgen.go
```diff
@@ -618,6 +619,10 @@ func generateJobBase(name, prefix string, info *ProwgenInfo, podSpec *corev1.Pod
 	} else if !skipCloning && info.Config.Private {
 		decorationConfig = &prowv1.DecorationConfig{OauthTokenSecret: &prowv1.OauthTokenSecret{Key: api.OauthTokenSecretKey, Name: api.OauthTokenSecretName}}
 	}
+	maxCustomDuration := time.Hour * 8
+	if timeout != nil && timeout.Duration <= maxCustomDuration {
+		decorationConfig.Timeout = timeout
+	}
```
petr-muller, line 625: I think we should actively refuse this in ci-op config validation (https://github.com/openshift/ci-tools/blob/master/pkg/validation/test.go#L116) with some sensible error message instead of capping it silently.

petr-muller (commented):
I think we'll also need a small doc followup for https://docs.ci.openshift.org/docs/architecture/timeouts/ (does not block merge)

pkg/prowgen/prowgen.go
vrutkovs: good idea, implemented

petr-muller (approved):

wking:
I've filed openshift/ci-docs#196 to doc this new setting.

PR #2295, opened 2021-08-25
ci-operator: include images into release explicitly

stevekuznetsov:
Previously, we always did this for integration releases with the
`latest` name. This is hard(er) to document and annoying to reason
about, so we should expose this as an option.

/assign @bbguimaraes @petr-muller @emilvberglind

### DIFF
--- a/pkg/validation/release.go
+++ b/pkg/validation/release.go
@@ -45,7 +45,7 @@ func validateReleases(fieldRoot string, releases map[string]api.UnresolvedReleas
 		} else if set == 0 {
 			validationErrors = append(validationErrors, fmt.Errorf("%s.%s: must set integration, candidate, prerelease or release", fieldRoot, name))
 		} else if release.Integration != nil {
-			validationErrors = append(validationErrors, validateIntegration(fmt.Sprintf("%s.%s", fieldRoot, name), *release.Integration)...)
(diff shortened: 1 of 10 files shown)


pkg/validation/release.go
```diff
@@ -57,14 +57,17 @@ func validateReleases(fieldRoot string, releases map[string]api.UnresolvedReleas
...
+func validateIntegration(fieldRoot, name string, integration api.Integration) []error {
 	var validationErrors []error
 	if integration.Name == "" {
 		validationErrors = append(validationErrors, fmt.Errorf("%s.name: must be set", fieldRoot))
 	}
 	if integration.Namespace == "" {
 		validationErrors = append(validationErrors, fmt.Errorf("%s.namespace: must be set", fieldRoot))
 	}
+	if integration.IncludeBuiltImages && name != api.LatestReleaseName {
+		validationErrors = append(validationErrors, fmt.Errorf("%s: only the `latest` release can set `include_built_images`", fieldRoot))
```
stevekuznetsov, line 69: We would be able to relax this in the future if we stop using this shenanigans around `stable(-foo)?` streams and do something rigorous.

stevekuznetsov:
Part of DPTP-2431

PR #2299, opened 2021-08-26
Make private jobs to use the oauth volume defined in decoration config

droslean:
/cc @openshift/test-platform 

### DIFF
--- a/pkg/prowgen/prowgen_test.go
+++ b/pkg/prowgen/prowgen_test.go
@@ -23,17 +23,20 @@ func TestGeneratePodSpec(t *testing.T) {
 		secrets        []*ciop.Secret
 		targets        []string
 		additionalArgs []string
+		skipCloning    bool
 	}{
 		{
 			description: "standard use case",
 			info:        &ProwgenInfo{Metadata: ciop.Metadata{Org: "org", Repo: "repo", Branch: "branch"}},
 			targets:     []string{"target"},
+			skipCloning: true,
 		},
 		{
 			description:    "additional args are included in podspec",
 			info:           &ProwgenInfo{Metadata: ciop.Metadata{Org: "org", Repo: "repo", Branch: "branch"}},
 			targets:        []string{"target"},
 			additionalArgs: []string{"--promote", "--some=thing"},
+			skipCloning:    true,
 		},
 		{
 			description:    "additional args and secret are included in podspec",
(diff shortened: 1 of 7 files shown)


droslean:
follow-up https://github.com/openshift/release/pull/21394

droslean:
/hold

This change needs 
- Bump prow to latest which allows this kind of validation
- Follow-up in o/release to regenerate the private jobs.

droslean:
@stevekuznetsov This PR is ready for review.

pkg/prowgen/prowgen_test.go
```diff
@@ -22,6 +22,7 @@ func TestGeneratePodSpec(t *testing.T) {
 		secrets        []*ciop.Secret
 		targets        []string
 		additionalArgs []string
+		shouldClone    bool
```
stevekuznetsov, line 25: invert this to make it easier to read - we always consider "skipCloning"

pkg/prowgen/prowgen_test.go
```diff
@@ -54,11 +55,20 @@ func TestGeneratePodSpec(t *testing.T) {
 			},
 			targets: []string{"target"},
 		},
+		{
+			description: "private job with skip_cloning false",
```
stevekuznetsov, line 59: test case for the opposite?

pkg/prowgen/prowgen_test.go
```diff
@@ -726,6 +736,17 @@ func TestGenerateJobBase(t *testing.T) {
 			podSpec: &corev1.PodSpec{Containers: []corev1.Container{{Name: "test"}}},
 			clone:   true,
 		},
+		{
+			testName: "private jobs without skipcloning should contain oauth_token_secret config",
```
stevekuznetsov, line 740: don't use double negatives, just harder to read

```
private jobs that clone should ...
```

test/integration/ci-operator-prowgen/output/jobs/private-org/super/private-org-super-master-presubmits.yaml
```diff
@@ -0,0 +1,57 @@
+presubmits:
```
stevekuznetsov, line 1: test for skip cloning?

droslean: We should already have those tests. see [here](https://github.com/openshift/ci-tools/blob/master/test/integration/ci-operator-prowgen/output/jobs/private-org/duper/private-org-duper-master-presubmits.yaml#L1-L58)

pkg/prowgen/prowgen_test.go
droslean: The `private job` test case is the opposite.

PR #2301, opened 2021-08-26
pkg/config: load in parallel

bbguimaraes:
Uses the same diamond fork-join pattern as `ci-operator-checkconfig`.

See https://github.com/openshift/ci-tools/pull/2304.


bbguimaraes:
/hold

Now includes #2307.

stevekuznetsov (approved):

PR #2302, opened 2021-08-26
pkg/jobconfig: load in parallel

bbguimaraes:
Uses the same diamond fork-join pattern as `ci-operator-checkconfig`.

See https://github.com/openshift/ci-tools/pull/2304.


bbguimaraes:
After some experimentation, I found an option that I think is clearer, at the cost of one goroutine that is blocked for most of its lifetime.

stevekuznetsov (approved):

PR #2308, opened 2021-08-27
(DPTP: 2412) Cluster init: sanitize prow jobs and basic framework

smg247:
This represents _only_ the basic framework (including creating the build farm dir) and the sanitize prow jobs config updates for the cluster-init task. This was created in response to https://github.com/openshift/ci-tools/pull/2269 getting too large. The rest of the files will be added in subsequent PRs.

/cc openshit/test-platform


cmd/cluster-init/main.go
```diff
@@ -0,0 +1,118 @@
...
+		logrus.WithError(err).Fatal("cannot parse args: ", os.Args[1:])
+	}
+	return o
+}
+
+func validateOptions(o options) []error {
+	var errs []error
+	if o.clusterName == "" {
+		errs = append(errs, fmt.Errorf("--cluster-name must be provided"))
+	} else if strings.ContainsAny(o.clusterName, " \t\n") {
```
stevekuznetsov, line 42: You might try 

```go
else if trimmed := strings.TrimFunc(o.clusterName, unicode.IsSpace); trimmed != o.clusterName {
```

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,118 @@
...
+		initClusterBuildFarmDir,
+		updateSanitizeProwJobs,
+	} {
+		if err := step(o); err != nil {
+			logrus.WithError(err)
+			errorCount++
+		}
+	}
+	if errorCount > 0 {
+		logrus.Infof("Due to the %d error(s) encountered a PR will not be generated. The resulting files can be PR'd manually", errorCount)
```
stevekuznetsov, line 77: Fatal?

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+}
+
+func updateClustersReadme(o options) error {
+	reader := bufio.NewReader(os.Stdin)
+	clustersReadmeFile := o.releaseRepo + "/clusters/README.md"
+	fmt.Printf("Would you like to add information about the '%s' cluster to %s? [y,n]: ",
```
stevekuznetsov, line 84: Why do we want this to be interactive? In what cases would we *not* update this?

cmd/cluster-init/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+					expectedDest := filepath.Join("..", item)
+					dest, err := os.Readlink(filepath.Join(buildDir, item))
+					if err != nil || dest != expectedDest {
+						t.Fatalf("item: %s was not symlinked to: %s", item, expectedDest)
+					}
+				}
+			}
+		})
+
+		t.Cleanup(func() {
```
stevekuznetsov, line 63: nit: register your cleanup first, or you this won't run when the test fails early

cmd/cluster-init/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+	for _, tc := range testCases {
+		buildDir := filepath.Join(testdata, "clusters", "build-clusters", tc.clusterName)
+		t.Run(tc.name, func(t *testing.T) {
+			err := initClusterBuildFarmDir(tc.options)
+			if diff := cmp.Diff(tc.expectedError, err, testhelper.EquateErrorMessage); diff != "" {
+				t.Fatalf("error does not match expectedError, diff: %s", diff)
+			}
+
+			if tc.expectedError == nil {
```
stevekuznetsov, line 49: we have in `test-infra` the package `testutil.CompareWithFixtureDir` - can we use that as a blanket check for the output dir?

cmd/cluster-init/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+				releaseRepo: testdata,
+			},
+		},
+		{
+			name: "missing cluster name",
+			options: options{
+				clusterName: "",
+				releaseRepo: testdata,
+			},
+			expectedErrors: []error{fmt.Errorf("--cluster-name must be provided")},
```
stevekuznetsov, line 99: The linter might nab you - this does not use formatting, so it should be `errors.New("--cluster...`

cmd/cluster-init/update_sanitize_prow_jobs.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/sirupsen/logrus"
+
+	"sigs.k8s.io/yaml"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/dispatcher"
+)
+
+func updateSanitizeProwJobs(o options) error {
+	logrus.Println("Updating sanitize-prow-jobs config")
```
stevekuznetsov, line 17: nit: use the `Info/Error` etc instead of a bare print in order to allow users to choose their level of logging output

cmd/cluster-init/update_sanitize_prow_jobs.go
```diff
@@ -0,0 +1,46 @@
...
+	filename := filepath.Join(o.releaseRepo, "core-services", "sanitize-prow-jobs", "_config.yaml")
+	data, err := ioutil.ReadFile(filename)
+	if err != nil {
+		return err
+	}
+	c := &dispatcher.Config{}
```
stevekuznetsov, line 23: Go-ism: Initialize intentionally empty values using `var`

```go
var c dispatcher.Config
```

cmd/cluster-init/update_sanitize_prow_jobs.go
```diff
@@ -0,0 +1,46 @@
...
+	if err = yaml.Unmarshal(data, c); err != nil {
+		return err
+	}
+	updateConfig(c, o.clusterName)
+	y, err := yaml.Marshal(c)
```
stevekuznetsov, line 28: What is `y`? Raw data? :)

cmd/cluster-init/update_sanitize_prow_jobs.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile(filename, y, 0644)
+}
+
+func updateConfig(c *dispatcher.Config, clusterName string) {
+	appGroup := c.Groups[api.ClusterAPPCI]
+	jobs := appGroup.Jobs
+	jobs = append(jobs, fmt.Sprintf("pull-ci-openshift-release-master-%s-dry", clusterName))
```
stevekuznetsov, line 38: I think you want `api.Metadata.JobName()` - make sure to use the `jobconfig.PresubmitPrefix` and associated constants for the prefix

cmd/cluster-init/update_sanitize_prow_jobs.go
```diff
@@ -0,0 +1,46 @@
...
+	jobs = append(jobs, fmt.Sprintf("branch-ci-openshift-release-master-%s-apply", clusterName))
+	jobs = append(jobs, fmt.Sprintf("periodic-openshift-release-master-%s-apply", clusterName))
+	c.Groups[api.ClusterAPPCI] = dispatcher.Group{
```
stevekuznetsov, line 41: I believe you should be able to do

```go
c.Groups[api.ClusterAPPCI].Jobs = append(c.Groups[api.ClusterAPPCI].Jobs, ....)
```

cmd/cluster-init/update_sanitize_prow_jobs_test.go
```diff
@@ -0,0 +1,111 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/dispatcher"
+)
+
+func TestUpdateSanitizeProwJobs(t *testing.T) {
```
stevekuznetsov, line 16: I think we get the best balance of unit and integration tests by having one large integration test for "the whole tool" that runs the entrypoint and asserts about the state of the entire output directory. Think about a flow that:

1. copies some static testdata input to a temporary directory
2. runs the whole tool you're working on, on this new directory
3. uses `testutil.CompareWithFixtureDir` to compare the modified directory with a static testdata output directory

cmd/cluster-init/main.go
smg247: It appears that only trims out the leading and trailing spaces. We also need to exclude interior spaces.

cmd/cluster-init/main.go
smg247: good point!

cmd/cluster-init/main.go
smg247: I wasn't sure if that was something we would want in the future. I can't think of any specific cases, so I'll remove that.

cmd/cluster-init/main_test.go
smg247: I'm going to look into using this for one **big** integration test for everything

cmd/cluster-init/main_test.go
smg247: It hasn't! but thats a good point, thanks

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: Meant to do that already...this one was missed, good eye

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: it's yaml...obviously ;)

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: That was my initial thought as well, but go doesn't allow that. It actually throws a compiler error. Apparently you can't append to a slice that is contained in a map. I had read an SO article on this at some point, but I lost it now. LMK if you want me to look for it.

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: How do you find all these!?

cmd/cluster-init/update_sanitize_prow_jobs_test.go
smg247: Completely agree with you here. The output checking unit tests along with a big integration test would be overkill so I'm going to remove those in favor of the integration test.

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,109 @@
...
+	return fmt.Sprintf("%#v", o)
+}
+
+func parseOptions() (options, error) {
+	var o options
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.clusterName, "cluster-name", "", "The name of the new cluster.")
+	fs.StringVar(&o.releaseRepo, "release-repo", "", "Path to the root of the openshift/release repository.")
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
```
stevekuznetsov, line 30: nit: you can 

```go
return o, fs.Parse(os.Args[1:])
```

cmd/cluster-init/main.go
stevekuznetsov: OK, still please use `unicode.IsSpace` in your check - there's a whole lot more to whitespace than your three chars :)

cmd/cluster-init/main_test.go
stevekuznetsov: This seems unresolved?

cmd/cluster-init/update_sanitize_prow_jobs.go
stevekuznetsov: Let's do this, then, so the you do not have to manage each individual field of the struct in the copy:

```go
appGroup := c.Groups[api.ClusterAPPCI]
appGroup.Jobs = append(appGroup.Jobs, ...)
c.Groups[api.ClusterAPPCI] = appGroup
```

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: I tried this, but thought the `yaml#Unmarshal` was throwing an error. Turns out it must have been something else, so I will make this change after all.

stevekuznetsov: Outstanding

cmd/cluster-init/main_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/openshift/ci-tools/pkg/testhelper"
+)
+
+func TestInitClusterBuildFarmDir(t *testing.T) {
```
stevekuznetsov, line 15: Why do we still need this if we are doing the bash-based directory integration test?

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,109 @@
...
+			logrus.WithError(err)
+			errorCount++
+		}
+	}
+	if errorCount > 0 {
+		logrus.Fatalf("Due to the %d error(s) encountered a PR will not be generated. The resulting files can be PR'd manually", errorCount)
+	}
+}
+
+func updateClustersReadme(o options) error {
```
stevekuznetsov, line 82: Let's let the user provide a static input (using a flag to pass in text or a file) instead of anything interactive.

smg247: As discussed we are just killing this feature for now. We can come back later and programmatically update the README if we think that actually provides value.

cmd/cluster-init/main.go
smg247: Yes, I figured this was a very naive approach for checking for whitespace. I'll work the `unicode.IsSpace` in there. Thanks!

cmd/cluster-init/main_test.go
smg247: Discussed in slack. Removing this test.

cmd/cluster-init/main_test.go
smg247: I thought I had made this update. updating now.

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: That does work, and is a bit cleaner, thanks

cmd/cluster-init/main.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+	if errorCount > 0 {
+		logrus.Fatalf("Due to the %d error(s) encountered a PR will not be generated. The resulting files can be PR'd manually", errorCount)
+	}
+}
+
+func initClusterBuildFarmDir(o options) error {
+	buildDir := buildFarmDirFor(o.releaseRepo, o.clusterName)
+	logrus.Infof("Creating build dir: %s\n", buildDir)
```
stevekuznetsov, line 83: nit: you don't need the newline in these

cmd/cluster-init/update_sanitize_prow_jobs.go
```diff
@@ -0,0 +1,48 @@
...
+func updateConfig(c *dispatcher.Config, clusterName string) {
+	appGroup := c.Groups[api.ClusterAPPCI]
+	metadata := api.Metadata{
+		Org:    "openshift",
+		Repo:   "release",
+		Branch: "master",
+	}
+	appGroup.Jobs = append(appGroup.Jobs, metadata.JobName(jobconfig.PresubmitPrefix, clusterName+"-dry"))
+	appGroup.Jobs = append(appGroup.Jobs, metadata.JobName(jobconfig.PostsubmitPrefix, clusterName+"-apply"))
+	appGroup.Jobs = append(appGroup.Jobs, fmt.Sprintf("%s-%s-%s-%s-%s-apply",
```
stevekuznetsov, line 45: Why is this not using the `JobName()` helper? Can we add a periodic helper if the format is different?

cmd/cluster-init/main.go
smg247: ah, didn't realize that, will update.

cmd/cluster-init/update_sanitize_prow_jobs.go
smg247: The helper adds "ci" in there, which we don't want for periodic. I can add a new helper though.

PR #2313, opened 2021-08-31
Update test-infra vendors

droslean:
/cc @openshift/test-platform 


petr-muller:
/hold

We needed to revert recent Prow deployment bump that included this code, so it is unlikely that we want to bump test-infra with this code now.

PR #2319, opened 2021-09-01
(DPTP-2412): Cluster init secret generator

smg247:
The second PR from the cluster-init series. This represents updates to the secret-generator config and corresponding unit tests.

/cc @openshift/test-platform


cmd/cluster-init/update_secret_generator.go
```diff
@@ -0,0 +1,100 @@
...
+
+func findSecretItem(itemName string, name string, likeCluster string, c SecretGenConfig) (*secretgenerator.SecretItem, error) {
+	idx := -1
+	for i, si := range c {
+		if itemName == si.ItemName {
+			containsName := false
+			containsCluster := false
+			for _, fj := range si.Fields {
+				if name == fj.Name {
+					containsName = true
```
emilvberglind, line 78: Do you want to break out of the loop here?

cmd/cluster-init/update_secret_generator.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+			}
+			for _, cluster := range si.Params["cluster"] {
+				if likeCluster == cluster {
+					containsCluster = true
```
emilvberglind, line 83: Do you want to break out of the loop here?

cmd/cluster-init/update_secret_generator.go
```diff
@@ -0,0 +1,100 @@
...
+			if containsName && containsCluster {
+				idx = i
+				break
+			}
+		}
+	}
+	if idx != -1 {
+		return &c[idx], nil
+	}
+	return &secretgenerator.SecretItem{},
```
emilvberglind, line 95: Could you just return a nil pointer here since there isn't an actual `SecretItem` found?

smg247: Good point, that was an artifact of the pattern I was using prior to bubbling up the errors. Updated

cmd/cluster-init/main.go
```diff
@@ -97,3 +103,7 @@ func initClusterBuildFarmDir(o options) error {
 func buildFarmDirFor(releaseRepo string, clusterName string) string {
 	return filepath.Join(releaseRepo, "clusters", "build-clusters", clusterName)
 }
+
+func serviceAccountKubeconfigPath(serviceAccount, clusterName string) string {
+	return fmt.Sprintf("sa.%s.%s.config", serviceAccount, clusterName)
```
droslean, line 108: This name format depends on a file that exists in Vault. @hongkailiu do we have constants for these files anywhere, to avoid hardcoding them here?

cmd/cluster-init/update_secret_generator.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	logrus.Infof("Appending to secret item: {itemName: %s, name: %s, likeCluster: %s}\n", itemName, name, string(api.ClusterBuild01))
+	si.Params["cluster"] = append(si.Params["cluster"], o.clusterName)
+	return nil
+}
+
+func findSecretItem(itemName string, name string, likeCluster string, c SecretGenConfig) (*secretgenerator.SecretItem, error) {
+	idx := -1
+	for i, si := range c {
+		if itemName == si.ItemName {
```
droslean, line 73: Perhaps you can expose https://github.com/openshift/ci-tools/blob/master/pkg/api/secretgenerator/secretgenerator.go#L51-L59 and use it to achieve the same here?

stevekuznetsov (approved):

cmd/cluster-init/main.go
stevekuznetsov: I think this *is* the authority now

smg247: With that in mind, is there a better place to put this?

cmd/cluster-init/update_secret_generator.go
smg247: In order to utilize that I would actually have to fork  it to use `SecretGenConfig` as the receiver.

PR #2320, opened 2021-09-01
Generalize image revert script

petr-muller:
- Use to revert any image, not just ci-operator
- Execute the command directly instead of just printing it


stevekuznetsov (commented):
This is amazing - can we cross-link it into the SOP in release or the on-call doc or somewhere?

PR #2311, opened 2021-08-30
Use flagutil.KubernetesOptions to set kube flags

hongkailiu:
Will rebase after https://github.com/openshift/ci-tools/pull/2306 and https://github.com/openshift/ci-tools/pull/2310

PR #2329, opened 2021-09-02
dptp-pools-cm: use inClusterConfig

hongkailiu:

/cc @stevekuznetsov

### DIFF
--- a/cmd/dptp-pools-cm/main.go
+++ b/cmd/dptp-pools-cm/main.go
@@ -1,14 +1,12 @@
 package main
 
 import (
-	"context"
 	"errors"
 	"flag"
 	"fmt"
 	"os"
 
 	"github.com/sirupsen/logrus"
-	"gopkg.in/fsnotify.v1"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -20,13 +18,7 @@ import (
 
 	hivev1 "github.com/openshift/hive/apis/hive/v1"
 
-	"github.com/openshift/ci-tools/pkg/api"
 	poolspullsecretprovider "github.com/openshift/ci-tools/pkg/controller/cluster_pools_pull_secret_provider"
-	"github.com/openshift/ci-tools/pkg/util"
-)
-
-const (
-	hiveContextName = string(api.HiveCluster)
 )
 
 var allControllers = sets.NewString(
@@ -35,8 +27,6 @@ var allControllers = sets.NewString(
 
 type options struct {
 	leaderElectionNamespace        string
-	kubeconfig                     string
-	kubeconfigDir                  string
 	leaderElectionSuffix           string
 	enabledControllers             flagutil.Strings
 	enabledControllersSet          sets.String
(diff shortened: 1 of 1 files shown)


pkg/util/client.go
```diff
@@ -117,3 +118,48 @@ func WatchFiles(candidates []string, callback func(fsnotify.Event)) error {
...
+	for k := range configs {
+		v := configs[k]
+		ret[k] = &v
+	}
+	return ret, nil
+}
+
+// WatchFilesV2 watches the passed files if they exist and calls callback for all events
+// except Chmod, as Openshift seems to generate frequent Chmod events
+func WatchFilesV2(candidates []string, callback func()) error {
```
stevekuznetsov, line 139: Doesn't the upstream code do this for us?

pkg/util/client.go
```diff
@@ -117,3 +118,48 @@ func WatchFiles(candidates []string, callback func(fsnotify.Event)) error {
...
 }
+
+// LoadKubeConfigsV2 loads kubeconfigs. If the kubeconfigChangedCallBack is non-nil, it will watch all kubeconfigs it loaded
+// and call the callback once they change.
+func LoadKubeConfigsV2(opts flagutil.KubernetesOptions, kubeconfigChangedCallBack func()) (map[string]*rest.Config, error) {
+	configs, err := opts.LoadClusterConfigs(kubeconfigChangedCallBack)
+	if err != nil {
+		return nil, fmt.Errorf("failed to load cluster configs: %w", err)
+	}
+	ret := map[string]*rest.Config{}
```
stevekuznetsov, line 129: Is this just a mapping from `rest.Config` to `*rest.Config`? Can we just make our downstream code expect `rest.Config` and get rid of `LoadKubeConfigsV2`, allowing us to simply use `opts.LoadClusterConfigs` ?

cmd/dptp-pools-cm/main.go
```diff
@@ -95,28 +92,29 @@ func main() {
...
+		inClusterConfig, err := rest.InClusterConfig()
+		if err != nil {
+			logrus.WithError(err).Fatal("failed to load in-cluster config")
+		}
+		kubeconfigs[hiveContextName] = *inClusterConfig
 		if err != nil {
 			logrus.WithError(err).Fatalf("--kubeconfig had no context for '%s' and loading InClusterConfig failed", hiveContextName)
 		}
 		logrus.Infof("Loaded %q context from in-cluster config", hiveContextName)
-		if err := util.WatchFiles([]string{"/var/run/secrets/kubernetes.io/serviceaccount/token"}, kubeconfigChangedCallBack); err != nil {
```
hongkailiu, line 114: This is the default path for SA's token.
If we still need this, we will fix it upstream.

pkg/util/client.go
hongkailiu: Removed.

pkg/util/client.go
hongkailiu: Updated.

cmd/dptp-pools-cm/main.go
```diff
@@ -54,13 +51,11 @@ type poolsPullSecretProviderOptions struct {
 }
 
 func newOpts() (*options, error) {
-	opts := &options{}
+	opts := &options{kubernetesOptions: flagutil.KubernetesOptions{NOInClusterConfigDefault: true}}
```
stevekuznetsov, line 54: The logic here seems wrong. What does this tool expect? Why do we allow *either* a `hive` context _or_ in-cluster config? What's the utility of this? Can we collapse to using just one? If not, why not configure the options here to load in-cluster, then in the code below simply use it when no `hive` config is passed? No need to load twice.

hongkailiu: NOInClusterConfigDefault skips this function at upstream.

https://github.com/kubernetes/test-infra/blob/665b28a81b9e9c97420ca03b5937a518938527f0/prow/kube/config.go#L51

It generates 2 more keys in the kubeconfig map: "" and "default".
The kubeconfig for "" is the inClusterConfig and "default" is the cluster Prow runs.
I feel those things are only for Prow and our tools should not include the logic.

If `opts := &options{kubernetesOptions: flagutil.KubernetesOptions{}}`, then I have to those 2 keys for build-clusters.

stevekuznetsov: You check later to see if the hive context exists, and if not, load the in-cluster config anyway. Why?

hongkailiu: Copied code from dptp-cm to support both the env. var and inClusterConfig. Actually, we can just use the inClusterConfig for this tool since it works only with `hive` and we use the SA for the pod.

https://github.com/openshift/release/blob/b809b4e30b517d13a1a6db73e1a989900636ac8a/clusters/hive/dptp-pools-cm.yaml#L114

stevekuznetsov (approved):

PR #2330, opened 2021-09-02
validate: remove wrong test and improve message

petr-muller:
>  The only requirement for mount paths is they should be absolute. That is tested correctly for credentials (validateCredentials in the same file) and is what should be done here too, IMO.

There's a test for absolute-ness of the mount path even for secrets (valid secret with invalid path), so the test just needs to be deleted, then.

Followup to https://github.com/openshift/ci-tools/pull/2293

/cc @bbguimaraes @stevekuznetsov

### DIFF
--- a/pkg/validation/test.go
+++ b/pkg/validation/test.go
@@ -208,7 +208,7 @@ func (v *Validator) validateTestStepConfiguration(
 			// validate path only if name is passed
 			if secret.MountPath != "" {
 				if ok := filepath.IsAbs(secret.MountPath); !ok {
-					validationErrors = append(validationErrors, fmt.Errorf("%s.path: '%s' secret mount path is not valid value, should be ^((\\/*)\\w+)+", fieldRootN, secret.MountPath))
+					validationErrors = append(validationErrors, fmt.Errorf("%s.path: '%s' secret mount path must be an absolute path", fieldRootN, secret.MountPath))
 				}
 			}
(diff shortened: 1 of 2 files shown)


petr-muller:
LOL wut how exactly did I test this

stevekuznetsov (commented):
Why is this a correct change? What part of `/path/to/secret:exec` is invalid? Why is `filepath.IsAbs()` wrong?

petr-muller:
> Why is this a correct change? What part of /path/to/secret:exec is invalid? Why is filepath.IsAbs() wrong?

I'm only doing this followup to make the code conform to an existing testcase that we have for a long time. Previous to https://github.com/openshift/ci-tools/pull/2293 the tests were only testing  validation fail/pass, and the testcase was mistakenly passing because the it failed validation for an unrelated reason. I do not know why the testcase was added, but the original error message (which suggests a regex instead of saying that the path should be absolute) indicates that this was intentional

bbguimaraes:
I don't see the existing implementation/test as authoritative.  The only requirement for mount paths is they should be absolute.  That is tested correctly for credentials (`validateCredentials` in the same file) and is what should be done here too, IMO.

petr-muller:
There's a test for absolute-ness of the mount path even for secrets (`valid secret with invalid path`), so the test just needs to be deleted, then.

PR #2337, opened 2021-09-03
ci-secret-bootstrap: Use flagutil.KubernetesOptions

hongkailiu:

/cc @stevekuznetsov

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -38,7 +38,6 @@ import (
 	vaultapi "github.com/openshift/ci-tools/pkg/api/vault"
 	"github.com/openshift/ci-tools/pkg/kubernetes/pkg/credentialprovider"
 	"github.com/openshift/ci-tools/pkg/secrets"
-	"github.com/openshift/ci-tools/pkg/util"
 )
 
 type options struct {
@@ -49,8 +48,7 @@ type options struct {
 	validateItemsUsage bool
 	confirm            bool
 
-	kubeConfigPath      string
-	kubeConfigDir       string
+	kubernetesOptions   flagutil.KubernetesOptions
 	configPath          string
 	generatorConfigPath string
 	cluster             string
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -358,6 +318,17 @@ var (
 	}
 )
 
+func getKubernetesOptions(kubeconfigPath string) (flagutil.KubernetesOptions, error) {
```
stevekuznetsov, line 321: Why do we need this? What are we actually testing? To my eyes it looks like this test is asserting something about the output of the upstream kubernetes options parsing, which is not useful - if we expect some subset of the upstream API surface, we need to assert on it in the runtime code. If we simply use the upstream code, we do not need to write our own unit tests for it here.

hongkailiu: The targeting function of the test is `completeOptions` which loads build clusters' kubeconfigs and uses them to verify the tools config.

https://github.com/openshift/ci-tools/pull/2337/files#diff-aff03726ea1d77880402ef9c410d101a01b0581c2449237a8d739fa21c09387bR134

I need a `flagutil.KubernetesOptions` to load kubeconfigs from a file.
The only way that I can do it is from `args`.

It is not testing the upstream code, at least not only for that.
It also tests if the loaded kubeconfigs can be used for the tools config.

stevekuznetsov: Is that right? We have test cases in here that pass kubeconfigs with typos, etc - and we expect to see an error. What are we testing? Nothing, other than that we actually *use* the underlying upstream options. It's not a useful test. It was useful when we had our own flag parsing logic, but we don't anymore. Maybe factor out the logic that takes as input the map of cluster aliases to clients, then test the downstream code?

hongkailiu: Updated.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -794,7 +791,7 @@ func main() {
 	if err := o.validateOptions(); err != nil {
 		logrus.WithError(err).Fatal("Invalid arguments.")
 	}
-	if err := o.completeOptions(&censor); err != nil {
+	if err := o.completeOptions(&censor, o.kubernetesOptions.LoadClusterConfigs); err != nil {
```
stevekuznetsov, line 794: Why pass the function instead of calling it and passing the results ... ?

hongkailiu: Because of my broken brain.

Updated.

hongkailiu:
The broken bootstrapper validation will be fixed by https://github.com/kubernetes/test-infra/pull/23513

stevekuznetsov (approved):

PR #2338, opened 2021-09-03
cluster-display: Use flagutil.KubernetesOptions

hongkailiu:

/cc @stevekuznetsov

### DIFF
--- a/cmd/cluster-display/main.go
+++ b/cmd/cluster-display/main.go
@@ -10,11 +10,12 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
-	"gopkg.in/fsnotify.v1"
 
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	"k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/interrupts"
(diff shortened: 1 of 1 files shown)


cmd/cluster-display/main.go
```diff
@@ -164,19 +164,35 @@ func main() {
...
 
-	kubeConfigs, err := util.LoadKubeConfigs(o.hiveKubeconfigPath, "", kubeconfigChangedCallBack)
-	if err != nil {
-		logrus.WithError(err).WithField("o.hiveKubeconfigPath", o.hiveKubeconfigPath).Fatal("could not load Hive kube config")
-	}
-	hiveConfig, ok := kubeConfigs[string(api.HiveCluster)]
-	if len(kubeConfigs) != 1 || !ok {
-		logrus.WithError(err).WithField("o.hiveKubeconfigPath", o.hiveKubeconfigPath).Fatalf("found %d contexts in Hive kube config and it must be %s", len(kubeConfigs), string(api.HiveCluster))
+	var hiveConfig *rest.Config
+	if o.hiveKubeconfigPath == "" {
```
hongkailiu, line 173: Tried to "use --hive-kubeconfig to set a flag value for --kubeconfig to support both" but
there is no setter for `o.kubernetesOptions.kubeconfig`. I have to play with the unparsed `args` to make this work, right?

I feel that the current solution might be easier.

stevekuznetsov (approved):

PR #2340, opened 2021-09-04
Use the same key name in oauth secret with upstream

droslean:
When `decoration_config.oauth_token_secret` is specified, the upstream code creates a volume using a specific name for the key. In order to use it we need to generate the same key name.

This change needs:

- Change the secret's key name in Vault
- Regenerate the jobs

Eventually, we can change the upstream code to respect the key name value from `decoration_config.oauth_token_secret.key`

/cc @openshift/test-platform 

### DIFF
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	oauthTokenPath = "/usr/local/github-credentials"
-	oauthKey       = "oauth"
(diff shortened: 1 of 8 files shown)


droslean:
Or we can just add a new key named `oauth-token` to that private secret

droslean:
@stevekuznetsov I think https://github.com/kubernetes/test-infra/pull/23481 should be better approach.
/close

PR #2341, opened 2021-09-06
Add a specification for optional jobs in ci-operator configuration

droslean:
/cc @openshift/test-platform 

https://issues.redhat.com/browse/DPTP-2501

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -650,6 +650,9 @@ type TestStepConfiguration struct {
 	// RunIfChanged is a regex that will result in the test only running if something that matches it was changed.
 	RunIfChanged string `json:"run_if_changed,omitempty"`
 
+	// Optional indicates that the job's status context, that is generated from the corresponding test, should not be required for merge.
+	Optional bool `json:"optional,omitempty"`
+
 	// SkipIfOnlyChanged is a regex that will result in the test being skipped if all changed files match that regex.
 	SkipIfOnlyChanged string `json:"skip_if_only_changed,omitempty"`
 
--- a/pkg/validation/test.go
+++ b/pkg/validation/test.go
@@ -151,6 +151,9 @@ func (v *Validator) validateTestStepConfiguration(
 		if test.Postsubmit && test.Interval != nil {
 			validationErrors = append(validationErrors, fmt.Errorf("%s: `interval` and `postsubmit` are mututally exclusive", fieldRootN))
 		}
+		if test.Postsubmit && test.Optional {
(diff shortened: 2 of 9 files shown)


droslean:
/hold 

This should break the current config, since there are jobs that have `optional: true` already added by the users manually. Holding this to investigate the safest way to proceed with those changes.

droslean:
/hold cancel

Since we already allow users to edit their `optional` field without breaking the job generation, this should be safe to merge.

pkg/api/types.go
```diff
@@ -650,6 +650,9 @@ type TestStepConfiguration struct {
 	// RunIfChanged is a regex that will result in the test only running if something that matches it was changed.
 	RunIfChanged string `json:"run_if_changed,omitempty"`
 
+	// Optional defines if the job will be marked as optional or required.
```
petr-muller, line 653: Such Godoc really has no added value over the name of the field, please make it more precise (good sentences will probably mention something about merging)

petr-muller (commented):
This is missing some important bits:

1) Please add validation that this is only used on presubmits
2) Please add a treatment like https://github.com/openshift/ci-tools/blob/752705aa4b09d516b21ae614bf28890a552a809d/pkg/jobconfig/files.go#L476-L483 to ensure that if a ci-op config sets the field, it cannot be overriden on the job level.

pkg/validation/test.go
```diff
@@ -161,8 +161,8 @@ func (v *Validator) validateTestStepConfiguration(
 		if test.Interval != nil && test.ReleaseController {
 			validationErrors = append(validationErrors, fmt.Errorf("%s: `interval` cannot be set for release controller jobs", fieldRootN))
 		}
-		if (test.Cron != nil || test.Interval != nil) && (test.RunIfChanged != "" || test.SkipIfOnlyChanged != "") {
-			validationErrors = append(validationErrors, fmt.Errorf("%s: `cron` and `interval` are mutually exclusive with `run_if_changed`/`skip_if_only_changed`", fieldRootN))
+		if (test.Cron != nil || test.Interval != nil) && (test.RunIfChanged != "" || test.SkipIfOnlyChanged != "" || test.Optional) {
+			validationErrors = append(validationErrors, fmt.Errorf("%s: `cron` and `interval` are mutually exclusive with `run_if_changed`/`skip_if_only_changed`/`optional`", fieldRootN))
```
petr-muller, line 165: optional is also exclusive with postsubmit

droslean: addressed. Test case also added.

petr-muller (approved):

PR #2342, opened 2021-09-06
WIP: Add a Prow plugin for triggering release jobs

petr-muller:
Just a skeleton for now


cmd/payload-testing/prow-plugin/server.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if !ocpPayloadTestsPattern.MatchString(ic.Comment.Body) {
+		return
+	}
+
+	meta := api.Metadata{
+		Org:    ic.Repo.Owner.Login,
+		Repo:   ic.Repo.Name,
+		Branch: "", // TODO(muller): How to get branch from Issue
```
droslean, line 57: I think you have to

```go
pr, err := ghc.GetPullRequest(ic.Repo.Owner.Login, ic.Repo.Name, ic.Issue.Number)
```
And then you should have the branch info from `pr.Bare.Ref`

cmd/payload-testing/prow-plugin/main.go
```diff
@@ -0,0 +1,85 @@
...
+		ghc:                githubClient,
+		ciopResolverClient: resolverClient,
+	}
+
+	eventServer := githubeventserver.New(o.githubEventServerOptions, getWebhookHMAC, logger)
+	eventServer.RegisterHandleIssueCommentEvent(serv.handleIssueComment)
+	eventServer.RegisterHelpProvider(helpProvider, logger)
+
+	interrupts.OnInterrupt(func() {
+		eventServer.GracefulShutdown()
```
stevekuznetsov, line 77: Not your code but why doesn't it just take a context and honor cancel?

cmd/payload-testing/prow-plugin/server.go
```diff
@@ -0,0 +1,85 @@
...
+
+func helpProvider(_ []prowconfig.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	// TODO(muller): Better descriptions, better help
+	pluginHelp := &pluginhelp.PluginHelp{
+		Description: `The ocp-payload-testing plugin spawns a battery of OCP release qualification testsuite jobs`,
+	}
+	pluginHelp.AddCommand(pluginhelp.Command{
+		// TODO(muller): What commands do we actually need?
+		// We will likely want to specify which jobs: blocking | informing | analysis | periodics | all
+		// We will also likely want to specify which release: ci | nightly | arm64 | s390x | ppc64
```
stevekuznetsov, line 31: We can't build those for PRs so not yet

cmd/payload-testing/prow-plugin/server.go
```diff
@@ -0,0 +1,85 @@
...
+var ocpPayloadTestsPattern = regexp.MustCompile(`(?mi)^/ocp-payload-tests\s*$`)
+
+func helpProvider(_ []prowconfig.OrgRepo) (*pluginhelp.PluginHelp, error) {
+	// TODO(muller): Better descriptions, better help
+	pluginHelp := &pluginhelp.PluginHelp{
+		Description: `The ocp-payload-testing plugin spawns a battery of OCP release qualification testsuite jobs`,
+	}
+	pluginHelp.AddCommand(pluginhelp.Command{
+		// TODO(muller): What commands do we actually need?
+		// We will likely want to specify which jobs: blocking | informing | analysis | periodics | all
```
deads2k, line 30: this list is good.

cmd/payload-testing/prow-plugin/server.go
deads2k: > We can't build those for PRs so not yet

ci vs nightly (in this list) is important. The set of jobs run on nightlys is different than the set of jobs for ci

cmd/payload-testing/prow-plugin/main.go
droslean: Well, this doesn't take wait for any context at all. `eventserver.Shutdown(ctx)` does, but this is being handled by interrupts. So we just need to pass a context to `interupts` instead.

petr-muller: I think that what Steve means is that the eventServer should be changed to accept a context and honor cancel by gracefully shutting down. THen on interrupt you'd just cancel that context.

cmd/payload-testing/prow-plugin/server.go
petr-muller: Yeah, but it somehow needs to be systematic: from the outside `arm64` is no different than`nightly`, but we actually need to treat them differntly (because we can do our thing with `nishglty `but not with `arm64`)

cmd/payload-testing/prow-plugin/main.go
droslean: Yes, this is not a blocker for now. The event server is already honoring a context in `Shutdown(ctx)`. The GracefullShutdown does something differently. It just waits for all the spawned goroutines to finish. Either way you can change this code to 
```go
interrupts.OnInterrupt(func() {
    eventServer.GracefulShutdown()
    if err := eventServer.Shutdown(myctx);err !=nil{.....}
}
```

PR #2343, opened 2021-09-06
WIP: Scaffold a CRD for PR release jobs

petr-muller:


pkg/api/pullrequestpayloadtest/v1/ci.openshift.io_pullrequestpayloadtests.yaml
```diff
@@ -0,0 +1,80 @@
+
+---
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.3.0
+  creationTimestamp: null
+  name: pullrequestpayloadtests.ci.openshift.io
```
stevekuznetsov, line 9: Should this live in the release controller?

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+type ReleaseJobSpec struct {
+	Org     string `json:"org"`
+	Repo    string `json:"repo"`
+	Branch  string `json:"branch"`
+	Variant string `json:"variant,omitempty"`
+	Test    string `json:"test"`
+}
+
+type PullRequestPayloadJobSpec struct {
+	ReleaseJobSpec `json:",inline"`
```
deads2k, line 28: I'd prefer this under a sub-element instead of inline

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+// payload validating jobs
+type PullRequestPayloadTest struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata"`
+
+	Spec   PullRequestPayloadTestSpec   `json:"spec"`
+	Status PullRequestPayloadTestStatus `json:"status"`
+}
+
+type ReleaseJobSpec struct {
```
deads2k, line 19: I need some help with godoc here

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+	Org     string `json:"org"`
+	Repo    string `json:"repo"`
+	Branch  string `json:"branch"`
+	Variant string `json:"variant,omitempty"`
```
deads2k, line 23: An optional coordinate seems odd.  What is a variant?

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+	Test    string `json:"test"`
+}
+
+type PullRequestPayloadJobSpec struct {
+	ReleaseJobSpec `json:",inline"`
+	ReleaseJobName string `json:"jobName,omitempty"`
```
deads2k, line 29: Does the name have to follow an particular format?  Does it match another system?

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PullRequestPayloadTest represents the intent to run a battery of OCP release
+// payload validating jobs
+type PullRequestPayloadTest struct {
```
deads2k, line 11: Can we think of a name other than Test? `PullRequestPayloadQualificationRun`?

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+	ReleaseJobSpec `json:",inline"`
+	ReleaseJobName string `json:"jobName,omitempty"`
+}
+
+type PullRequestPayloadTestSpec struct {
+	Jobs []PullRequestPayloadJobSpec `json:"jobs"`
+}
+
+type PullRequestPayloadTestStatus struct {
+	// TODO(muller): List of spawned jobs
```
deads2k, line 37: start with `conditions []metav1.Condition` at least

pkg/api/pullrequestpayloadtest/v1/ci.openshift.io_pullrequestpayloadtests.yaml
deads2k: This resource looks more directly related to CI at the moment.

pkg/api/pullrequestpayloadtest/v1/types.go
```diff
@@ -0,0 +1,48 @@
...
+
+type ReleaseJobSpec struct {
+	Org     string `json:"org"`
+	Repo    string `json:"repo"`
+	Branch  string `json:"branch"`
+	Variant string `json:"variant,omitempty"`
+	Test    string `json:"test"`
+}
+
+type PullRequestPayloadJobSpec struct {
```
deads2k, line 27: I expected PR coordinates.

pkg/api/pullrequestpayloadtest/v1/types.go
petr-muller: The coordinates address a test in a specific file and variant is an optional coordinate:

See
 https://github.com/openshift/release/blob/master/ci-operator/config/openshift/release/openshift-release-master__ci-4.9.yaml#L301-L305 and compare to e.g.
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/metallb-operator/openshift-metallb-operator-main.yaml#L67-L70

Also: https://docs.ci.openshift.org/docs/how-tos/contributing-openshift-release/#variants

I think that at the moment, release jobs happen to *always* be present in a variant file, but that's just an arbitrary state, not something to rely on.

pkg/api/pullrequestpayloadtest/v1/types.go
petr-muller: It's the same name like r-c config references now, and if set, I expect it to match the `ReleaseJobSpec`. I'm not yet sure if I'll even use/need this name.

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+	metav1.ObjectMeta `json:"metadata"`
+
+	Spec   PullRequestPayloadTestSpec   `json:"spec"`
+	Status PullRequestPayloadTestStatus `json:"status"`
+}
+
+// PullRequestPayloadTestSpec specifies for which PR the payload qualification run was requested
+// and the list of individual jobs that should be executed
+type PullRequestPayloadTestSpec struct {
+	PullRequest PullRequestUnderTest        `json:"pullRequest"`
```
deads2k, line 24: doc please.  mutable, immutable, required?

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+	Jobs        []PullRequestPayloadJobSpec `json:"jobs"`
+}
+
+// PullRequestUnderTest describes the state of the repo that will be under test
+// This is a combination of the PR revision and base ref revision. Tested code
+// is the specific revision of the PR merged into the base branch with
+// a specific branch as a HEAD
+type PullRequestUnderTest struct {
+	// Org is something like kubernetes or k8s.io
```
deads2k, line 33: can you explain this in terms of github.com/foo/bar or is it related to the golang package?

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+	Org string `json:"org"`
+	// Repo is something like test-infra
```
deads2k, line 35: Same as above

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+
+	Spec   PullRequestPayloadTestSpec   `json:"spec"`
+	Status PullRequestPayloadTestStatus `json:"status"`
+}
+
+// PullRequestPayloadTestSpec specifies for which PR the payload qualification run was requested
+// and the list of individual jobs that should be executed
+type PullRequestPayloadTestSpec struct {
+	PullRequest PullRequestUnderTest        `json:"pullRequest"`
+	Jobs        []PullRequestPayloadJobSpec `json:"jobs"`
```
deads2k, line 25: is this set all at once or added to over time?

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// PullRequestPayloadQualificationRun represents the intent to run a battery of OCP release
+// payload validating jobs
+type PullRequestPayloadQualificationRun struct {
+	metav1.TypeMeta   `json:",inline"`
```
deads2k, line 14: Given a PR, how do I find the instance of this?  How many instances can there be?

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+type PullRequestPayloadJobSpec struct {
+	ReleaseJobSpec `json:"releaseJobSpec"`
+}
+
+// ReleaseJobSpec identifies the release payload one qualification test to execute. In this context,
+// "test" means one item in the specified ci-operator configuration file.
+type ReleaseJobSpec struct {
+	// CiOperatorConfig identifies the ci-operator configuration with the test
+	CiOperatorConfig api.Metadata `json:"ciOperatorConfig"`
+	// Test is the name of the test in the ci-operator configuration
```
deads2k, line 63: is this a Job?

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+// PullRequestPayloadTestStatus provides runtime data, such as references to submitted ProwJobs,
+// whether all jobs are submitted, finished, etc.
+type PullRequestPayloadTestStatus struct {
+	Conditions []metav1.Condition            `json:"conditions"`
+	Jobs       []PullRequestPayloadJobStatus `json:"jobs"`
+}
+
+// PullRequestPayloadJobStatus is a reference to a Prowjob submitted for a single item
+// from the list of jobs to be submitted
+type PullRequestPayloadJobStatus struct {
```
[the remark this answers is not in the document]
deads2k: there can be multiple jobruns for the same job.

pkg/api/pullrequestpayloadqualification/v1/types.go
petr-muller: immutable, required

pretty much everything should be immutable I guess - the requestor should just create the CR and then it makes no sense to change it (similar to Prowjobs)

pkg/api/pullrequestpayloadqualification/v1/types.go
petr-muller: in github.com/foo/bar repo would be `bar`

pkg/api/pullrequestpayloadqualification/v1/types.go
petr-muller: All at once, fully specified by the requestor when they create the CR

pkg/api/pullrequestpayloadqualification/v1/types.go
petr-muller: Not alone. Whole `ReleaseJobSpec` instance corresponds to one Prowjob configuration.

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -0,0 +1,93 @@
...
+	// ReleaseJobName is a name of the job that corresponds to the name corresponding to the
+	// ReleaseJobSpec tuple. This name is inferred from ReleaseJobSpec data and corresponds to
+	// the name which the user would see in e.g. release-controller
+	ReleaseJobName string `json:"jobName"`
+	// ProwJob is a name of the submitted ProwJob resource
+	ProwJob string `json:"prowJob"`
```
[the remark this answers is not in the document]
petr-muller: My current intent is to run this system in the same cluster/namespace where Prow runs, IMO no need to make these coordinates more complicated. Cluster and namespace are constants.

pkg/api/pullrequestpayloadqualification/v1/types.go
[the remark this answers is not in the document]
petr-muller: That's a property of the `Prowjob`  resource. Prowjob resource is a unique execution of the job (==jobrun, I guess?):

```console
$ appci get -n ci prowjob 078ea460-26b8-11ec-b60a-0a580a804747 -o json | jq .status.build_id -r
1445769306611650560
```

So `ProwJob` here would say `078ea460-26b8-11ec-b60a-0a580a804747` and the UI would get the build ID from that.

pkg/api/pullrequestpayloadqualification/v1/types.go
petr-muller: N instances: corresponding to multiple `/payload informing 4.7` comments over time, and they could run over different tested state, and they can (but not must) run different sets of jobs

 Given a PR, you would need to search for resources with matching `PullRequestUnderTest`, I guess we could label the resource for easier search?

petr-muller:
Thanks! Responded to individual comments, will address the actionable ones in the code.

pkg/api/pullrequestpayloadqualification/v1/types.go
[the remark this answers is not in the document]
petr-muller: I do not think it is worth listing these in every custom resource. Both cluster and namespace are configuration-time constants of a Prow system. At this time, I do not think there is any benefit from adding it.

PR #2347, opened 2021-09-08
Tolerate neither of `skip_if/run_if` changes in jobs if one is set in ci-op config

petr-muller:
If either of these is set on ci-operator level, *both* ones must not be manually changed in the job (otherwise we may end up with a job having both).


droslean (approved):

petr-muller:
/hold
I want to see the breaking-changes test pass (I don't expect it to fail)

PR #2350, opened 2021-09-08
branching: add a schedule generator utility

stevekuznetsov:
This will run on the schedules provided by the PM team to generate the
data our tools expect to operate on, committing it to build farms in a
config map for easy loading.

### DIFF
--- a/cmd/branchingconfigmanagers/schedule-generator/main.go
+++ b/cmd/branchingconfigmanagers/schedule-generator/main.go
@@ -1,0 +1,143 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/fs"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/logrusutil"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	crcontrollerutil "sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/yaml"
+
+	"github.com/openshift/ci-tools/pkg/api/ocplifecycle"
+)
+
+type options struct {
+	logLevel          string
+	scheduleDir       string
+	dryRun            bool
+	kubernetesOptions flagutil.KubernetesOptions
+}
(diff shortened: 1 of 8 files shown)


cmd/branchingconfigmanagers/schedule-generator/main.go
```diff
@@ -0,0 +1,138 @@
...
+func main() {
+	logrusutil.ComponentInit()
+	o, err := gatherOptions()
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to gather options")
+	}
+	if err := validateOptions(o); err != nil {
+		logrus.WithError(err).Fatalf("Invalid options")
+	}
+	level, _ := logrus.ParseLevel(o.logLevel)
```
petr-muller, line 62: lol was that error check too much typing :trollface:

cmd/branchingconfigmanagers/schedule-generator/main.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		logrus.WithError(err).Fatal("Could not find marshal schedules")
+	}
+
+	for _, config := range kubeConfigs {
+		client, err := ctrlruntimeclient.New(&config, ctrlruntimeclient.Options{})
+		if err != nil {
+			logrus.WithError(err).Fatal("Could not get client for kube config")
+		}
+		if err := upsertConfigMap(string(raw), client); err != nil {
```
petr-muller, line 85: So one ugly thing that could happen is CMs going out of sync and then we could act on the wrong config... I think that at least we should continue to the rest of the clusters on errors (we could have already succeeded on some some so aborting does not save much). 

If we want to be robust, we could perhaps mark the CM (via a separate key) as stale before we update it, then update it (and clean the staleness). Config managers would then only need to act on non-stale configs.

cmd/branchingconfigmanagers/schedule-generator/main.go
stevekuznetsov: It's checked in the validation

cmd/branchingconfigmanagers/schedule-generator/main.go
stevekuznetsov: Interesting idea, but IMHO limited in utility - we would only leave the thing in a stale state if and only if we are able to update it once but not twice.

stevekuznetsov:
@petr-muller updated to be best-effort across clusters

petr-muller (approved):

cmd/branchingconfigmanagers/schedule-generator/main.go
petr-muller: Yeah, we would have only proceed with actual update if we managed to stale CMs everywhere. But I guess operating over stale (but consistent) config everywhere is not good situation either.

PR #2354, opened 2021-09-08
dptp-cm: Use flagutil.KubernetesOptions

hongkailiu:

/cc @stevekuznetsov

### DIFF
--- a/cmd/dptp-controller-manager/main.go
+++ b/cmd/dptp-controller-manager/main.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/bombsimon/logrusr"
 	"github.com/sirupsen/logrus"
-	"gopkg.in/fsnotify.v1"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -35,7 +34,6 @@ import (
 	"github.com/openshift/ci-tools/pkg/controller/testimagestreamimportcleaner"
 	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
 	"github.com/openshift/ci-tools/pkg/load/agents"
-	"github.com/openshift/ci-tools/pkg/util"
 )
 
 const (
@@ -54,8 +52,7 @@ type options struct {
 	ciOperatorconfigPath                 string
 	stepConfigPath                       string
 	prowconfig                           configflagutil.ConfigOptions
-	kubeconfig                           string
-	kubeconfigDir                        string
+	kubernetesOptions                    flagutil.KubernetesOptions
 	leaderElectionSuffix                 string
 	enabledControllers                   flagutil.Strings
 	enabledControllersSet                sets.String
@@ -96,15 +93,14 @@ type serviceAccountSecretRefresherOptions struct {
 }
 
 func newOpts() (*options, error) {
-	opts := &options{GitHubOptions: &flagutil.GitHubOptions{}}
+	opts := &options{GitHubOptions: &flagutil.GitHubOptions{}, kubernetesOptions: flagutil.KubernetesOptions{NOInClusterConfigDefault: true}}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	opts.prowconfig.AddFlags(fs)
 	opts.addDefaults()
 	opts.GitHubOptions.AddFlags(fs)
 	opts.GitHubOptions.AllowAnonymous = true
 	fs.StringVar(&opts.leaderElectionNamespace, "leader-election-namespace", "ci", "The namespace to use for leaderelection")
-	fs.StringVar(&opts.kubeconfig, "kubeconfig", "", "The kubeconfig to use. All contexts in it will be considered a build cluster. If it does not have a context named 'app.ci', loading in-cluster config will be attempted.")
-	fs.StringVar(&opts.kubeconfigDir, "kubeconfig-dir", "", "Path to the directory containing kubeconfig files. All contexts in it will be considered a build cluster. If it does not have a context named 'app.ci', loading in-cluster config will be attempted.")
+	opts.kubernetesOptions.AddFlags(fs)
 	fs.StringVar(&opts.ciOperatorconfigPath, "ci-operator-config-path", "", "Path to the ci operator config")
 	fs.StringVar(&opts.stepConfigPath, "step-config-path", "", "Path to the registries step configuration")
 	fs.StringVar(&opts.leaderElectionSuffix, "leader-election-suffix", "", "Suffix for the leader election lock. Useful for local testing. If set, --dry-run must be set as well")
@@ -170,7 +166,9 @@ func newOpts() (*options, error) {
 	if err := opts.GitHubOptions.Validate(opts.dryRun); err != nil {
 		errs = append(errs, err)
 	}
-
+	if err := opts.kubernetesOptions.Validate(false); err != nil {
+		errs = append(errs, err)
+	}
 	return opts, utilerrors.NewAggregate(errs)
 }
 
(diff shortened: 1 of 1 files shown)


cmd/dptp-controller-manager/main.go
```diff
@@ -236,24 +234,21 @@ func main() {
...
+		logrus.Info("Kubeconfig changed, exiting to get restarted by Kubelet and pick up the changes")
 		cancel()
 	}
 
-	kubeconfigs, err := util.LoadKubeConfigs(opts.kubeconfig, opts.kubeconfigDir, kubeconfigChangedCallBack)
+	kubeconfigs, err := opts.kubernetesOptions.LoadClusterConfigs(kubeconfigChangedCallBack)
 	if err != nil {
 		logrus.WithError(err).Fatal("failed to load kubeconfigs")
 	}
 	if _, hasAppCi := kubeconfigs[appCIContextName]; !hasAppCi {
```
stevekuznetsov, line 246: Do we need this branch? What do we do in the deployment? I don't understand why our loading has to be so complex. Either always expect it to be in-cluster, or always expect it to be present in the configs ...

hongkailiu: It turns out we need both:

development: load from $KUBECONFIG
https://github.com/openshift/ci-tools/blob/12fad0a042c90eda3cddf44fb8dd84a1ea6ea9e6/hack/run-dptp-controller-manager.sh#L11

production: load from inClusterConfig because the secret does not include the one for `app.ci`:

https://github.com/openshift/release/blob/fa66cd849fb407495cd50fab336d3f0e88fcb622/core-services/ci-secret-bootstrap/_config.yaml#L1948-L1961

Maybe add it into the secret and then remove the code for inClusterConfig?

stevekuznetsov: Change development. We do not write code to handle development situations, we write it to handle production. Developing on different code-paths to production is a recipe for disaster.

stevekuznetsov: The k8s loader also allows you to provide it a loading strategy, so that it perfers one to the other. @alvaroaleman is there any simpler way to do this? I really dislike how damn complicated we've made loading. The library is complex as all get-out, then each caller does stuff like this to make it even more impossible.

alvaroaleman: This uses the in-cluster config for app.ci if there is no app.ci config. I don't really see how anything in client-go would be able to do this, as it doesn't know what name the in-cluster config is for.

I also don't really see how any change in the deployment would improve things here. Sure, we could add a flag for `--default-in-cluster-config-for-cluster-name` but I don't anticipate this changing so it looks more like busywork than like something that improves things to me.

stevekuznetsov: I just hate that we have 35 places that load config, and their call-sites are all over the place. Half of them do something like this, half don't, half have custom handling for different things, half don't ... next time anyone has to do a broad change it's going to be as much of a nightmare as it has been for Hongkai. Surely we can do better?

stevekuznetsov: Maybe load in-cluster if possible, add it to a well-known key in the output mapping that the upstream library provides, then the caller just moves map keys around?

hongkailiu: Updated.
Not sure if it is getting better or worse though.

stevekuznetsov: Do we put the in-cluster config into a known key? Can we fail over to that? @alvaroaleman any thoughts?

alvaroaleman: Yeah we do put it into a known key, it is handled above

cmd/dptp-controller-manager/main.go
```diff
@@ -236,24 +235,30 @@ func main() {
...
-			logrus.WithError(err).Fatalf("--kubeconfig had no context for '%s' and loading InClusterConfig failed", appCIContextName)
-		}
-		logrus.Infof("Loaded %q context from in-cluster config", appCIContextName)
-		if err := util.WatchFiles([]string{"/var/run/secrets/kubernetes.io/serviceaccount/token"}, kubeconfigChangedCallBack); err != nil {
-			logrus.WithError(err).Fatal("faild to watch in-cluster token")
+		if !hasInClusterConfig {
+			logrus.WithError(err).Fatalf("had no context for '%s' and loading InClusterConfig failed", appCIContextName)
+		} else {
+			// https://github.com/kubernetes/test-infra/blob/5bce0be55da8a04f62c77d2292a81739491544df/prow/kube/config.go#L56-L62
+			if inClusterConfig.Host != api.APPCIKubeAPIURL {
```
alvaroaleman, line 257: That is never going to pass, the Host will be `$KUBERNETES_SERVICE_HOST + KUBERNETES_SERVICE_PORT" which is an ip address: port (and probably the same among all openshift clusters. I would just remove the `else` clause

alvaroaleman: /hold

cmd/dptp-controller-manager/main.go
```diff
@@ -295,7 +300,7 @@ func main() {
 			syncPeriod := 24 * time.Hour
 			options.SyncPeriod = &syncPeriod
 		}
-		mgr, err := controllerruntime.NewManager(cfg, options)
+		mgr, err := controllerruntime.NewManager(&cfg, options)
```
alvaroaleman, line 303: You can not pass a pointer to the loop var

hongkailiu: Sorry. One more thing that I need to remember.

cmd/dptp-controller-manager/main.go
hongkailiu: Updated. Removing the `else` may break the dev script.

cmd/dptp-controller-manager/main.go
```diff
@@ -236,24 +235,33 @@ func main() {
...
-		logrus.Infof("Loaded %q context from in-cluster config", appCIContextName)
-		if err := util.WatchFiles([]string{"/var/run/secrets/kubernetes.io/serviceaccount/token"}, kubeconfigChangedCallBack); err != nil {
-			logrus.WithError(err).Fatal("faild to watch in-cluster token")
+		if !hasInClusterConfig {
+			logrus.WithError(err).Fatalf("had no context for '%s' and loading InClusterConfig failed", appCIContextName)
+		} else {
+			// There is a corner case that the inClusterConfig is the current context of a loaded kubeconfig file
+			// In dev, it has to be app.ci's public URL or in a pod, it ${KUBERNETES_SERVICE_HOST}:${KUBERNETES_SERVICE_PORT}
+			// https://github.com/kubernetes/test-infra/blob/5bce0be55da8a04f62c77d2292a81739491544df/prow/kube/config.go#L56-L62
+			inClusterConfigHost := fmt.Sprintf("%s:%s", os.Getenv("KUBERNETES_SERVICE_HOST"), os.Getenv("KUBERNETES_SERVICE_PORT"))
```
alvaroaleman, line 259: But why? This code shouldn't reimplement part of the client-go logic (and break if client-go should be changed). So far, we trusted that a context named `app.ci` is for app.ci and that the in-cluster config is for that as well. What is the reason not to trust that anymore?

hongkailiu: For example, if we accidentally removed this line

https://github.com/openshift/ci-tools/blob/cc37aa0f41d5e5d71739d87123433e6d34ec7d9c/hack/run-dptp-controller-manager.sh#L11

then we would use another cluster as `app.ci`.

I just want to be safer.

hongkailiu: Previously, we load `inCluster` without the logic from upstream and now the upstream logic could give me a wrong result.

alvaroaleman: Why would in the example give a wrong result?

hongkailiu: It could come from `currentContext` of the last kubeconfig file it loads.

https://github.com/kubernetes/test-infra/blob/c3f3e307169f26b3790a132060c4bb2f3c76a66a/prow/kube/config.go#L58

and

https://github.com/kubernetes/test-infra/blob/c3f3e307169f26b3790a132060c4bb2f3c76a66a/prow/kube/config.go#L121-L135

alvaroaleman: Ok, I am just uneasy about relying on implementation details here. The thing needs to be prefixed with `https://` in the in-cluster case and joined with net.JoinHostPorT: https://github.com/openshift/release-controller/blob/efd4b6d1c92ff73804e397524de974d160a8dc31/vendor/k8s.io/client-go/rest/config.go#L508

alvaroaleman (approved):

hongkailiu:
/hold cancel

I will be watching ...

hongkailiu:
```
oc get pod -n ci dptp-controller-manager-9fc689774-th5hw
NAME                                      READY   STATUS    RESTARTS   AGE
dptp-controller-manager-9fc689774-th5hw   1/1     Running   0          20m
```
seems working

PR #2356, opened 2021-09-08
serviceaccount-secret-rotation-trigger: Use flagutil.KubernetesOptions

hongkailiu:

/cc @stevekuznetsov

### DIFF
--- a/cmd/serviceaccount-secret-rotation-trigger/main.go
+++ b/cmd/serviceaccount-secret-rotation-trigger/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"os"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


cmd/serviceaccount-secret-rotation-trigger/main.go
```diff
@@ -64,7 +71,7 @@ func main() {
 	for clusterName, config := range kubeconfigs {
 		clusterName, config := clusterName, config
 		eg.Go(func() error {
-			client, err := ctrlruntimeclient.New(config, ctrlruntimeclient.Options{})
+			client, err := ctrlruntimeclient.New(&config, ctrlruntimeclient.Options{})
```
stevekuznetsov, line 74: This takes the address of the loop variable `config` which gets reset at every iteration. Pass it as a parameter to the anonymous function given to `eg.Go()`

alvaroaleman: No, it gets shadowed in line 72

stevekuznetsov: ffs. I really dislike that pattern, it's super hard to see and read :eyeroll:

stevekuznetsov (approved):

PR #2362, opened 2021-09-10
add lease type for the azure-2 cluster profile and its missing references

droslean:
/cc @openshift/test-platform 

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1130,6 +1130,7 @@ func ClusterProfiles() []ClusterProfile {
 		ClusterProfileAWSGovCloud,
 		ClusterProfileAWSGluster,
 		ClusterProfileAlibaba,
+		ClusterProfileAzure2,
 		ClusterProfileAzure4,
 		ClusterProfileAzureArc,
 		ClusterProfileAzureStack,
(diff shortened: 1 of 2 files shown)


pkg/api/types.go
```diff
@@ -1188,6 +1189,8 @@ func (p ClusterProfile) ClusterType() string {
 		return "aws-china"
 	case ClusterProfileAWSGovCloud:
 		return "aws-usgov"
+	case ClusterProfileAzure2:
+		return "azure-2"
```
petr-muller, line 1193: This likely belongs to the case below, this is not a differnt cluster type, just a new account for an existing cluster type

petr-muller (commented):
Cna you make sure all places form this example PR are handled?

https://github.com/openshift/ci-tools/commit/b89a00a9a39acd29d68f7490f49cf93b50cc0d21#

pkg/api/types.go
droslean: I was planning to make the tests to use it as `azure-2` cluster profile. see https://github.com/openshift/release/blob/master/ci-operator/step-registry/ipi/install/install/ipi-install-install-commands.sh#L64

So that wasn't necessary, right?

petr-muller: Yes, but cluster profile is something differnt than cluster type

petr-muller: I don't see why we would have a differnt clsuter type for 2nd auzre account? We did not have different cluster type for 2nd gcp or aws accounts. This just creates a pitfall for step authors.

droslean: Ok I will remove this, and its reference from the steps.

droslean: https://github.com/openshift/release/pull/21745

stevekuznetsov (approved):

PR #2364, opened 2021-09-10
test-images-distributor: stop generating requests for disabled clusters

hongkailiu:
We have handled it in the reconcile function:

https://github.com/openshift/ci-tools/blob/f96d4cad19030ad3fb827b55966869a78dca0154/pkg/controller/test-images-distributor/test_images_distributor.go#L292-L295

However, the requests should not even be generated in the first place.

/cc @alvaroaleman

### DIFF
--- a/pkg/controller/test-images-distributor/test_images_distributor.go
+++ b/pkg/controller/test-images-distributor/test_images_distributor.go
@@ -98,7 +98,7 @@ func AddToManager(mgr manager.Manager,
 	// TODO: Watch buildCluster ImageStreams as well. For now we assume no one will tamper with them.
 	if err := c.Watch(
 		source.NewKindWithCache(&testimagestreamtagimportv1.TestImageStreamTagImport{}, mgr.GetCache()),
-		testImageStreamTagImportHandler(log),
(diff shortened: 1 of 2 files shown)


pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -89,7 +89,7 @@ func AddToManager(mgr manager.Manager,
 		}
 		if err := c.Watch(
 			source.NewKindWithCache(&testimagestreamtagimportv1.TestImageStreamTagImport{}, buildClusterManager.GetCache()),
-			testImageStreamTagImportHandlerForNamedCluster(buildClusterName),
+			testImageStreamTagImportHandlerForNamedCluster(buildClusterName, ignoreClusterNames),
```
alvaroaleman, line 92: This is incorrect, we don't even get here if the cluster is in `ignoreClusterNames`.

What needs adjustment is `testImageStreamTagImportHandler`, which handles `testImageStreamTags` on app ci by creating a request using the cluster that is in them. If that cluster is in `ignoreClusterNames` we need to do nothing there

alvaroaleman (approved):

alvaroaleman (approved):

PR #2366, opened 2021-09-10
Add a README for config manager authors

petr-muller:
This document attempts to describe the contract/conventions for all
config managers, so individual cards do not need to repeat the contract.

Not sure if this is the right place, I'll gladly move it somewhere else if needed.

[DPTP-2440](https://issues.redhat.com/browse/DPTP-2440)

### DIFF
--- a/cmd/branchingconfigmanagers/README.md
+++ b/cmd/branchingconfigmanagers/README.md
@@ -1,0 +1,102 @@
+# Branching Config Managers
+
+This directory contains individual components of the Automated Branch Cuts project, specifically the so-called "Config
+Managers". Config Managers are programs that reconcile CI configuration based on the current state of the OCP lifecycle.
+This document describes the contract and conventions for individual Config Managers, so they fit into the wider
+Automated Branch Cuts system. In other words, Config Manager authors should refer to this document for guidance about
+what are Config Managers expected to do.
+
+## Config Manager Contract
+
+Given the current state of the CI configuration (openshift/release repository working copy) and product lifecycle data,
+update the CI configuration to match the state expected by the product policy at given time.
+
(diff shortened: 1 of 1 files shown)


alvaroaleman (approved):
/hold
Looks good modulo the configmap details. Maybe also mention it is a configmap because there might be non-public information in it?

cmd/branchingconfigmanagers/README.md
```diff
@@ -0,0 +1,102 @@
...
+Any new config manager should be deployed to run in a periodic Prow job which mounts a ConfigMap `TBD` that contains the
+YAML file with [OCP lifecycle data](#product-lifecycle-data). The Prowjob should first run the Config Manager, then if
+it ended with zero exit code, it should use the `prcreator` tool to commit the changes and submit a PR to propagate the
+changes.
+
+TODO: What period
+
+### ConfigMap with OCP Lifecycle Data
+
+TODO: What ConfigMap
```
petr-muller, line 102: @stevekuznetsov do we know this information already?

petr-muller:
@alvaroaleman thanks, I added a item to the `Config Managers Must` section about the confidentiality. I'm not sure if the details about ConfigMap placement is finalized already, pinged Steve about it.

cmd/branchingconfigmanagers/README.md
stevekuznetsov: No, this is TODO

petr-muller: I'll add a note about why this is a TODO and when it needs to be filled in, then I think we can merge, WDYT?. It's useful for CM authors even without discussing how to deploy.

PR #2370, opened 2021-09-13
promoted-image-governor: prune images on build farm

hongkailiu:
https://issues.redhat.com/browse/DPTP-2480

### DIFF
--- a/cmd/promoted-image-governor/main.go
+++ b/cmd/promoted-image-governor/main.go
@@ -21,9 +21,9 @@ import (
 	"k8s.io/apimachinery/pkg/util/json"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/kube"
 	"k8s.io/test-infra/prow/logrusutil"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/yaml"
@@ -37,7 +37,7 @@ import (
 
 type options struct {
 	ciOperatorconfigPath             string
-	kubeconfig                       string
+	kubernetesOptions                flagutil.KubernetesOptions
 	dryRun                           bool
 	ignoredImageStreamTagsRaw        flagutil.Strings
 	ignoredImageStreamTags           []*regexp.Regexp
@@ -56,7 +56,7 @@ type options struct {
 func parseOptions() *options {
 	opts := &options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
-	fs.StringVar(&opts.kubeconfig, "kubeconfig", "", "Path to the kubeconfig file to use for CLI requests.")
+	opts.kubernetesOptions.AddFlags(fs)
 	fs.StringVar(&opts.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
 	fs.StringVar(&opts.ciOperatorconfigPath, "ci-operator-config-path", "", "Path to the ci operator config")
 	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
@@ -106,6 +106,10 @@ func (o *options) validate() error {
 		return fmt.Errorf("--openshift-mapping-config and --explain cannot be set together")
 	}
 
+	if err := o.kubernetesOptions.Validate(o.dryRun); err != nil {
+		return err
+	}
+
 	o.explains = map[api.ImageStreamTagReference]string{}
 	var errs []error
 	for _, val := range o.explainsRaw.Strings() {
@@ -130,6 +134,7 @@ func (o *options) validate() error {
 
 const (
 	explanationUnknown = "unknown"
+	appCIContextName   = string(api.ClusterAPPCI)
 )
 
 func loadMappingConfig(path string) (*OpenshiftMappingConfig, error) {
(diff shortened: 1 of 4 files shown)


cmd/promoted-image-governor/main.go
```diff
@@ -399,16 +454,37 @@ func main() {
...
 	}
 
-	kubeConfig, err := clientcmd.BuildConfigFromFlags("", opts.kubeconfig)
+	kubeconfigs, err := opts.kubernetesOptions.LoadClusterConfigs()
 	if err != nil {
-		logrus.WithError(err).Fatalf("could not load kube config from path %s", opts.kubeconfig)
+		logrus.WithError(err).Fatal("failed to load kubeconfigs")
+	}
+	if _, hasAppCi := kubeconfigs[appCIContextName]; !hasAppCi {
+		InClusterConfig, err := rest.InClusterConfig()
```
hongkailiu, line 462: See first how this goes.
https://github.com/openshift/ci-tools/pull/2354#discussion_r704762206

stevekuznetsov (approved):

hongkailiu:
[output.txt](https://github.com/openshift/ci-tools/files/7171427/output.txt)

```
cat prom.log | grep "deleting image stream tag on build farm" | jq --slurp '.' | jq -r '.[] | .cluster + "|" + .tagKey' | sort
```

@stevekuznetsov @droslean 

The above is generated from the log under dry-run mode. When this PR is merged, those images will be deleted on build-farm.

hongkailiu:
https://issues.redhat.com/browse/DPTP-2480

cmd/promoted-image-governor/main.go
```diff
@@ -303,6 +309,62 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+				} else {
+					// the isTag should be deleted on the build farm since it is not on app.ci
+					logrus.WithField("cluster", appCIContextName).WithField("tagKey", tagKey).Debug("image stream tag not found")
+				}
+				logrus.WithField("cluster", cluster).WithField("tagKey", tagKey.String()).Info("deleting image stream tag on build farm")
+				if dryRun {
+					continue
+				}
+				isTagOnBuildFarm := &imagev1.ImageStreamTag{}
+				if err := client.Get(ctx, tagKey, isTagOnBuildFarm); err != nil {
```
stevekuznetsov, line 345: Why do you need to GET to DELETE? Could we just DELETE and ignore NotFound?

cmd/promoted-image-governor/main.go
```diff
@@ -303,6 +309,62 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+			imageStream := &imagev1.ImageStream{}
+			if err := client.Get(ctx, streamKey, imageStream); err != nil {
+				if !kerrors.IsNotFound(err) {
+					errs = append(errs, fmt.Errorf("could not get image stream %s in namespace %s on cluster %s: %w", streamKey.Name, streamKey.Namespace, cluster, err))
+				} else {
+					logrus.WithField("cluster", cluster).WithField("streamKey", streamKey).Debug("image stream not found")
+				}
+				continue
+			}
+			for _, tag := range imageStream.Status.Tags {
```
stevekuznetsov, line 325: Instead of fanning out and doing N API calls per tag, why don't we just fetch the stream, then process the list of tags client-side?

cmd/promoted-image-governor/main.go
hongkailiu: Thought about this too during implementation.
Attempt to avoid delete actions if i can.
I will change accordingly.

cmd/promoted-image-governor/main.go
hongkailiu: This is why it took so much time on each run.
I will change.

cmd/promoted-image-governor/main.go
```diff
@@ -303,6 +309,80 @@ func isMirroredFromOCP(tag api.ImageStreamTagReference, refs []ImageStreamRef) b
...
+				}
+				continue
+			}
+
+			tags := sets.NewString()
+			for _, tag := range imageStream.Status.Tags {
+				tags.Insert(tag.Tag)
+			}
+
+			for _, tag := range appCIImageStream.Status.Tags {
```
stevekuznetsov, line 353: nit: this might be a little more legibile if you create two sets and do a `Difference()`

cmd/promoted-image-governor/main.go
```diff
@@ -399,16 +479,51 @@ func main() {
...
+
+	inClusterConfig, hasInClusterConfig := kubeconfigs[kube.InClusterContext]
+	delete(kubeconfigs, kube.InClusterContext)
+	delete(kubeconfigs, kube.DefaultClusterAlias)
+
+	if _, hasAppCi := kubeconfigs[appCIContextName]; !hasAppCi {
+		if !hasInClusterConfig {
+			logrus.WithError(err).Fatalf("had no context for '%s' and loading InClusterConfig failed", appCIContextName)
+		} else {
+			// There is a corner case that the inClusterConfig is the current context of a loaded kubeconfig file
```
stevekuznetsov, line 495: Change dev. We should never, ever write code in production to handle dev cases. Make dev look like prod.

cmd/promoted-image-governor/main.go
hongkailiu: Updated.

cmd/promoted-image-governor/main.go
hongkailiu: Simplified it a bit.
The `else-branch` here is to ensure if the inClusterConfig is correct because it could be the one for the current context of the last loaded kubeconfig file.

If we do not consider the Dev env at all, the code would be

```
         inClusterConfig := kubeconfigs[kube.InClusterContext]
	delete(kubeconfigs, kube.InClusterContext)
	delete(kubeconfigs, kube.DefaultClusterAlias)
	kubeconfigs[appCIContextName] = inClusterConfig
```

However, I cannot figure out a way to make the above code work for dev because dev does not run in pod, and have no inclusterconfig.

stevekuznetsov: Why do we need it? In dev, pass a `$KUBECONFIG` with the `app.ci` context explicitly and do not use in-cluster.

stevekuznetsov (approved):

stevekuznetsov (approved):

PR #2371, opened 2021-09-13
test/e2e: remove resource limits in ci-op configs

AlexNPavel:
Resource limits should not be set in ci-op configs. This PR removes the
resources limit from all ci-op configs used in the e2e tests.

/cc @stevekuznetsov


stevekuznetsov (approved):
*CPU* limits should not be set, but effectively yes what this commit does is right.

PR #2375, opened 2021-09-14
Jobconfig prune with explicit labels

smg247:
`jobconfig.WriteToDir` currently prunes **only** based on the `ci-operator.openshift.io/prowgen-controlled` label. With this change it will allow pruning with any set of labels. This is needed for the cluster-init tool to maintain current prowgen created periodics when adding its own periodics to the config.

In the future `cluster-init` will be updated to allow for build cluster maintenance. This is the reason for the generation placing 2 labels on each job: one to mark it as being "owned" by cluster-init, and another to mark it as a part of a specific build cluster. With this we will be able to regenerate **all** cluster-init controlled jobs at once.


pkg/jobconfig/files.go
```diff
@@ -610,32 +611,45 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
 }
 
-// IsGenerated returns true if the job was generated using prowgen
-func IsGenerated(job prowconfig.JobBase) bool {
-	_, generated := job.Labels[LabelGenerated]
-	return generated
+// IsGenerated returns true if the job was generated according to ANY of the generatedLabels
+func IsGenerated(job prowconfig.JobBase, generatedLabels ...string) bool {
+	for _, generatedLabel := range generatedLabels {
+		_, generated := job.Labels[generatedLabel]
```
emilvberglind, line 617: `if _, generated := job.Labels[generatedLabel]; generated {
	return true
}`

pkg/jobconfig/files.go
```diff
@@ -610,32 +611,45 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
+	}
+	return false
 }
 
-func isStale(job prowconfig.JobBase) bool {
-	genLabel, generated := job.Labels[LabelGenerated]
-	return generated && genLabel != string(NewlyGenerated)
+func isStale(job prowconfig.JobBase, generatedLabels ...string) bool {
+	for _, generatedLabel := range generatedLabels {
+		value, generated := job.Labels[generatedLabel]
```
emilvberglind, line 627: `if value, generated := job.Labels[generatedLabel]; generated && value != string(NewlyGenerated) {
			return true
		}`

pkg/jobconfig/files.go
smg247: cool, I didn't realize you could use that syntactic sugar if the function returned more than one value.

emilvberglind: I think there are a couple other places in your changes where you're doing similar checks that could be updated with this.

cmd/cluster-init/update_jobs.go
```diff
@@ -60,6 +65,8 @@ func generatePeriodic(clusterName string) prowconfig.Periodic {
 			},
 			Labels: map[string]string{
 				labelRole: jobRoleInfra,
+				clusterInitControlledLabelFor(clusterName): string(jobconfig.NewlyGenerated),
+				jobconfig.LabelClusterInitGenerated:        "true",
```
stevekuznetsov, line 69: Why do we need this?

pkg/jobconfig/files.go
```diff
@@ -271,11 +272,11 @@ func readFromFile(path string) (*prowconfig.JobConfig, error) {
...
+// WriteToDir takes a JobConfig and a target directory, and writes the Prow job configuration
 // into files in that directory. Jobs are sharded by branch and by type. If
 // target files already exist and contain Prow job configuration, the jobs will
-// be merged.
-func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig) error {
+// be merged. Jobs will be pruned based on the provided generatedLabels
+func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, generatedLabels ...string) error {
 	allJobs := sets.String{}
 	files := map[string]*prowconfig.JobConfig{}
 	key := fmt.Sprintf("%s/%s", org, repo)
```
stevekuznetsov, line 282: Was it not possible to *add* the newly-generated labels to the jobs in the following loops, instead of requiring the caller to remember to do so in every case?

pkg/jobconfig/files_test.go
```diff
@@ -808,8 +816,216 @@ func TestInfo_ConfigMapName(t *testing.T) {
...
+
+func TestPrune(t *testing.T) {
+	testCases := []struct {
+		name            string
+		jobconfig       *prowconfig.JobConfig
+		expectedPruned  bool
+		generatedLabels []string
+	}{
+		{
+			name: "stale generated presubmit is pruned",
```
stevekuznetsov, line 834: nit: test periodics?

pkg/jobconfig/files_test.go
```diff
@@ -808,8 +816,216 @@ func TestInfo_ConfigMapName(t *testing.T) {
...
+					Labels: map[string]string{LabelClusterInitGenerated: "true"}}}},
+			},
+			expectedPruned:  true,
+			generatedLabels: []string{LabelClusterInitGenerated, "some-other-label"},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Expect either unchanged or empty JobConfig
```
stevekuznetsov, line 927: Not the biggest fan of this type of test code - it can hide bugs and adds complexity. Is it hard to have literal expected data?

test/integration/cluster-init/expected/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
```diff
@@ -8,6 +8,7 @@ periodics:
     repo: release
   interval: 12h
   labels:
+    ci-operator.openshift.io/prowgen-controlled: "true"
```
stevekuznetsov, line 11: Is this correct?

test/integration/cluster-init/expected/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
```diff
@@ -46,6 +47,8 @@ periodics:
     repo: release
   interval: 12h
   labels:
+    ci-operator.openshift.io/cluster-init-controlled: "true"
+    ci-operator.openshift.io/cluster-init-controlled-newCluster: "true"
```
stevekuznetsov, line 51: nit: labels are a mapping. Why don't we do something like

```yaml
  labels:
    ci-operator.openshift.io/cluster-init-controlled: "true"
    ci-operator.openshift.io/cluster-init: "newCluster"
```

We can keep the `ci-operator.openshift.io/cluster-init-controlled` as the one that switches around between `newly-generated` etc and the other can just be an identifying label for selection.

pkg/jobconfig/files.go
```diff
@@ -610,32 +611,45 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
 	return label
 }
 
-// IsGenerated returns true if the job was generated using prowgen
-func IsGenerated(job prowconfig.JobBase) bool {
-	_, generated := job.Labels[LabelGenerated]
-	return generated
+// IsGenerated returns true if the job was generated according to ANY of the generatedLabels
+func IsGenerated(job prowconfig.JobBase, generatedLabels ...string) bool {
```
stevekuznetsov, line 615: Why do we want OR instead of AND?

stevekuznetsov (commented):
I think we'd be a little cleaner if we used the full space of labels - both keys and values - and a label selector instead of custom matching logic.

test/integration/cluster-init/expected/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml
smg247: Yeah, I want to verify that that job doesn't disappear when `cluster-init` is run.

pkg/jobconfig/files_test.go
smg247: I just moved this test over from `prowgen_test.go`. Seems it would be trivial to make that update though.

pkg/jobconfig/files.go
bbguimaraes: Note that this is probably true of `prowgen` as well.  I mentioned in my other PRs that we can rethink pruning now that there isn't a round trip to the file system as there used to be and we have old and new jobs conveniently in memory.

I was making a large enough change at the time, so I left it out, but we can consider that if we're making changes here.

pkg/jobconfig/files_test.go
smg247: If you mean that I should also test periodics, I have one toward the bottom now.

pkg/jobconfig/files.go
```diff
@@ -610,34 +627,56 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
+// Prune removes all generated jobs of the supplied generator with values that are NOT newly-generated.
 // Prune() returns the resulting job config (which may even be completely empty).
-func Prune(jobConfig *prowconfig.JobConfig) *prowconfig.JobConfig {
+func Prune(jobConfig *prowconfig.JobConfig, generator generator, pruneLabels labels.Set) (*prowconfig.JobConfig, error) {
 	var pruned prowconfig.JobConfig
 
 	for repo, jobs := range jobConfig.PresubmitsStatic {
 		for _, job := range jobs {
-			if isStale(job.JobBase) {
+			stale, err := isStale(job.JobBase, generator, pruneLabels)
```
bbguimaraes, line 666: This creates the same selector with the same input values for every job.

Create it once and use it here.

pkg/jobconfig/files.go
```diff
@@ -610,34 +627,56 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
+			if err != nil {
+				return nil, err
+			}
+			if stale {
 				continue
 			}
-
-			if IsGenerated(job.JobBase) {
-				job.Labels[LabelGenerated] = string(Generated)
+			generated, err := IsGenerated(job.JobBase, generator)
```
bbguimaraes, line 673: Same here and other cases below.

pkg/jobconfig/files.go
```diff
@@ -25,21 +27,24 @@ import (
...
+	ProwJobLabelVariant              = "ci-operator.openshift.io/variant"
+	ReleaseControllerLabel           = "ci-operator.openshift.io/release-controller"
+	LabelCluster                     = "ci.openshift.io/cluster"
+	LabelGenerator                   = "ci.openshift.io/generator"
+	ReleaseControllerValue           = "true"
+	JobReleaseKey                    = "job-release"
+	PresubmitPrefix                  = "pull"
+	PostsubmitPrefix                 = "branch"
+	PeriodicPrefix                   = "periodic"
+	NewlyGenerated         label     = "newly-generated"
```
bbguimaraes, line 45: This can be private to the module now.

pkg/jobconfig/files.go
```diff
@@ -25,21 +27,24 @@ import (
...
+	Prowgen                generator = "prowgen"
+	ClusterInit            generator = "cluster-init"
```
bbguimaraes, line 47: Nothing in here uses these constants, so they can be moved to their respective modules (`pkg/prowgen`, `cmd/cluster-init`).

smg247: I was thinking that, but the `generator` type is defined here, so I wasn't sure what the best practice was with that. I'll move them.

pkg/jobconfig/files.go
smg247: Good call!

droslean:
@smg247 What are all those `Merge branch 'openshift:master' into master` commits?  Those changes are not related to this PR. I guess they were accidentally injected here from rebasing?

smg247:
> @smg247 What are all those `Merge branch 'openshift:master' into master` commits? Those changes are not related to this PR. I guess they were accidentally injected here from rebasing?

Correct. I was not used to working with forks when I started out, and made some silly mistakes toward the beginning that made my fork have a messed up history. I am going to blow all that away as soon as I finish up all the PRs for DPTP-2412, but I don't want to lose my branches until then.

droslean:
> Correct. I was not used to working with forks when I started out, and made some silly mistakes toward the beginning that made my fork have a messed up history. I am going to blow all that away as soon as I finish up all the PRs for DPTP-2412, but I don't want to lose my branches until then.

No worries. I just wanted to make sure that we won't merge something that it shouldn't.

pkg/jobconfig/files.go
```diff
@@ -610,34 +625,64 @@ func MakeRegexFilenameLabel(possibleRegex string) string {
...
-// remaining prowgen-generated jobs will be labeled as simply "generated" and
+// Prune removes all generated jobs of the supplied Generator with values that are NOT newly-generated.
 // Prune() returns the resulting job config (which may even be completely empty).
-func Prune(jobConfig *prowconfig.JobConfig) *prowconfig.JobConfig {
+func Prune(jobConfig *prowconfig.JobConfig, generator Generator, pruneLabels labels.Set) (*prowconfig.JobConfig, error) {
 	var pruned prowconfig.JobConfig
+	staleSelector, err := staleSelectorFor(generator, pruneLabels)
+	if err != nil {
+		return nil, err
+	}
```
stevekuznetsov, line 672: nit: for review it may be easier to write a function here so that there's no diff in the following code:

```go
isGenerated := func(job JobBase) bool {
    return generatedSelector.Matches(labels.Set(job.Labels))
}

isStale := func(job JobBase) bool {
    return staleSelector.Matches(labels.Set(job.Labels))
}
```

If it's not a hassle, can we try this and see the diff in the algorithm go to 0?

pkg/prowgen/testdata/zz_fixture_TestGenerateJobBase_config_variant.yaml
```diff
@@ -3,7 +3,6 @@ decorate: true
 decoration_config:
   skip_cloning: true
 labels:
-  ci-operator.openshift.io/prowgen-controlled: newly-generated
```
stevekuznetsov, line 6: Why are these removed and not changed to the new label?

smg247: Generating the job base no longer needs to add the old label or the new `ci.openshift.io/generator` label as that is done in the `jc.WriteToDir` function.

pkg/jobconfig/files.go
smg247: Its not going to be no diff as the `isGenerated` case body needs to delete a temporary label now, rather than change it, but I can do that.

stevekuznetsov (approved):

PR #2378, opened 2021-09-16
Validate existing build-farm configs

hongkailiu:
https://issues.redhat.com/browse/DPTP-2521

/cc @smg247 @stevekuznetsov

PR #2379, opened 2021-09-17
(DPTP-2412) Allow cluster-init tool to update the clusters that it generates jobs

smg247:


smg247:
Merge: https://github.com/openshift/release/pull/21998 alongside

cmd/cluster-init/main.go
```diff
@@ -141,13 +152,17 @@ func main() {
...
-		updateJobs,
-		initClusterBuildFarmDir,
-		updateCiSecretBootstrap,
-		updateSecretGenerator,
-		updateSanitizeProwJobs,
-	} {
+	steps := []func(options) error{updateJobs}
+	if !o.update {
+		steps = append(steps,
+			initClusterBuildFarmDir,
```
stevekuznetsov, line 158: I'd rather see `os.MkDirAll()` (for example)

cmd/cluster-init/main.go
```diff
@@ -141,13 +152,17 @@ func main() {
...
+			updateCiSecretBootstrap,
```
stevekuznetsov, line 159: What would it take to make this work on update? What if we change the definition of what the secrets should look like? It's useful to update everything.

smg247: I thought about that as well...Since we don't have the equivalent to a `jc.WriteToDir` in place for `secretbootstrap` I think it would result in a fair bit of effort. We would either need to go through every stanza in the file checking for the current cluster's configurations and remove them prior to doing a regenerate (this may not end up being too crazy, but there are a lot of configs in there), or we would need to write something like `prune` to prune the old configs when new ones are generated. Either way, I came to the conclusion that this might be more work than it would end up saving. Thoughts?

cmd/cluster-init/main.go
```diff
@@ -56,7 +58,7 @@ func parseOptions() (options, error) {
 
 func validateOptions(o options) []error {
 	var errs []error
-	if o.clusterName == "" {
+	if !o.update && o.clusterName == "" {
```
hongkailiu, line 61: What do we do when `o.update && o.clusterName==""`?

cmd/cluster-init/main.go
```diff
@@ -92,10 +94,19 @@ func validateOptions(o options) []error {
 			}
 		}
 
-		if o.clusterName != "" {
-			buildDir := buildFarmDirFor(o.releaseRepo, o.clusterName)
-			if _, err := os.Stat(buildDir); !os.IsNotExist(err) {
-				errs = append(errs, fmt.Errorf("build farm directory: %s already exists", o.clusterName))
+		if o.update {
+			if o.clusterName != "" {
+				buildDir := buildFarmDirFor(o.releaseRepo, o.clusterName)
```
hongkailiu, line 99: https://github.com/openshift/ci-tools/pull/2378/files#diff-5b92b4f3254b078d3a1a836f7055b073b60ddc08f024494639503b0052188d8dR247

We could do more detailed checks like ^.

cmd/cluster-init/main.go
smg247: yes, all build clusters in the new `clusters/build-clusters/cluster-init.yaml` file get updated to current spec.

cmd/cluster-init/update_jobs.go
```diff
@@ -21,24 +25,75 @@ const (
...
 
+type BuildClusters struct {
+	Managed []string `json:"managed,omitempty"`
+}
+
 func updateJobs(o options) error {
+	var clusters []string
+	if o.clusterName == "" {
+		// Updating ALL cluster-init managed clusters
+		buildClusters, err := loadBuildClusters(o)
```
hongkailiu, line 36: Could we derive this from the bootstrap's config? like https://github.com/openshift/ci-tools/pull/2378/files#diff-5b92b4f3254b078d3a1a836f7055b073b60ddc08f024494639503b0052188d8dR190

smg247:
@stevekuznetsov per our slack convo: let's get this merged in, and I will focus on the updates to the other pieces on a new PR.

stevekuznetsov (commented):
I think an approach where the top-level tool (in `main` or whatever) does the mapping of "`--update` -> all clusters from file" and then loops over them, calling the code that already exists, would be easier to extend, as it would require that no individual step ever be aware of if it was generating or updating. In the future we might want to think about a more formal way for steps to opt out of updates, if e.g. they require one-time admin access that's useful on generate only (set up DNS via google cloud cli)

cmd/cluster-init/update_jobs.go
```diff
@@ -21,24 +25,75 @@ const (
...
-	return jobconfig.WriteToDir(jobsDir,
-		metadata.Org,
-		metadata.Repo,
-		&config,
-		generator,
-		map[string]string{jobconfig.LabelCluster: o.clusterName})
+	var matchLabels map[string]string
+	if o.clusterName != "" {
+		// We are only updating a single cluster, so we must only match on that
+		matchLabels = map[string]string{jobconfig.LabelCluster: o.clusterName}
```
hongkailiu, line 61: What is the meaning of this label?

cmd/cluster-init/main.go
stevekuznetsov: Again, instead of *validating* the content, just generate the correct content and merge.

cmd/cluster-init/update_jobs.go
smg247: I thought about that as well, but I thought that a first-class config file would be less hacky here.

cmd/cluster-init/update_jobs.go
smg247: It is used so that the pruning doesn't remove jobs for other clusters that are managed by cluster-init, but that we are not updating here.

test/integration/cluster-init.sh
```diff
@@ -12,12 +12,18 @@ suite_dir="${OS_ROOT}/test/integration/cluster-init/"
...
-os::integration::compare "${actual}" "${expected}"
+# test the create scenario
+actual_create="${tempdir}/create/input"
+expected_create="${suite_dir}/create/expected"
+os::cmd::expect_success "cluster-init -cluster-name=newCluster -release-repo=${actual_create} -create-pr=false"
+os::integration::compare "${actual_create}" "${expected_create}"
+# test the update scenario
+actual_update="${tempdir}/update/input"
+expected_update="${suite_dir}/update/expected"
+os::cmd::expect_success "cluster-init -cluster-name=existingCluster -release-repo=${actual_update} -update=true -create-pr=false"
```
stevekuznetsov, line 26: Why does this require a separate input and output? When would we expect it to actually change something from the above?

cmd/cluster-init/main.go
smg247: Build farm directory updating will come in the next PR. I will add the relevant ones with that.

test/integration/cluster-init.sh
smg247: I actually went in and made some fake changes to the new input (for example adding extra args to the commands in the jobs) to verify that the update actually changed it back to the expected spec. I didn't want the create scenario to have those silly changes in there, so I opted to keep the directories separate. 

As for the output, it will be different in the update scenario because we aren't adding a new cluster, but only updating the existing one.

I thought about just chaining these tests, so do a create, and then do an update to the output of the create. However, I wanted the results to be independent. If I chained them then the update would always fail when the create did.

stevekuznetsov:
One last thing - can we look at implementing in e.g. breaking-changes presubmit a run of this tool to ensure that we know in this repo when changes we're making will affect the `openshfit/release` repository? That is only a requirement once we enforce something in `openshift/release` so perhaps not technically a blocker for this particular PR but should be something we think about doing soon

smg247:
> One last thing - can we look at implementing in e.g. breaking-changes presubmit a run of this tool to ensure that we know in this repo when changes we're making will affect the `openshfit/release` repository? That is only a requirement once we enforce something in `openshift/release` so perhaps not technically a blocker for this particular PR but should be something we think about doing soon

Good call! I will add that to my radar.

stevekuznetsov (approved):

PR #2380, opened 2021-09-17
Regulate the existing applyconfig jobs

hongkailiu:
- handle build01, build02, vsphere
- fix the diff

### DIFF
--- a/cmd/cluster-init/update_jobs.go
+++ b/cmd/cluster-init/update_jobs.go
@@ -54,7 +54,7 @@ func generatePeriodic(clusterName string) prowconfig.Periodic {
 					generateContainer("applyconfig:latest",
 						clusterName,
 						[]string{"--confirm=true"},
-						nil)},
+						nil, nil)},
 				ServiceAccountName: configUpdater,
 			},
 			UtilityConfig: prowconfig.UtilityConfig{
@@ -84,7 +84,7 @@ func generatePostsubmit(clusterName string) prowconfig.Postsubmit {
 			Spec: &v1.PodSpec{
 				Volumes: []v1.Volume{generateSecretVolume(clusterName)},
 				Containers: []v1.Container{
-					generateContainer(latestImage, clusterName, []string{"--confirm=true"}, nil)},
+					generateContainer(latestImage, clusterName, []string{"--confirm=true"}, nil, nil)},
 				ServiceAccountName: configUpdater,
 			},
 			UtilityConfig: prowconfig.UtilityConfig{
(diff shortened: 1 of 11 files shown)


cmd/cluster-init/update_jobs.go
```diff
@@ -160,16 +164,71 @@ func generateSecretVolume(clusterName string) v1.Volume {
 	}
 }
 
-func generateContainer(image, clusterName string, extraArgs []string, extraVolumeMounts []v1.VolumeMount) v1.Container {
+func generateContainer(image, clusterName string, extraArgs []string, extraVolumeMounts []v1.VolumeMount, home bool) v1.Container {
```
smg247, line 167: rather than the `home` arg can we do a `extraEnvVars []v1.EnvVar` similar to the rest of the signature?

cmd/cluster-init/main.go
```diff
@@ -236,7 +236,18 @@ func initClusterBuildFarmDir(o options) error {
 }
 
 func buildFarmDirFor(releaseRepo string, clusterName string) string {
-	return filepath.Join(releaseRepo, "clusters", "build-clusters", clusterName)
+	return filepath.Join(releaseRepo, "clusters", "build-clusters", folder(clusterName))
+}
+
+func folder(name string) string {
```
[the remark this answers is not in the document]
stevekuznetsov: or just move the two directories so we don't need this? :)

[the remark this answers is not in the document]
stevekuznetsov: or even symlink them to minimize disruption

cmd/cluster-init/update_jobs.go
hongkailiu: Updated.

cmd/cluster-init/main.go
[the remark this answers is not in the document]
hongkailiu: Removed.

Need https://github.com/openshift/release/pull/22027

cmd/cluster-init/update_jobs.go
```diff
@@ -160,16 +164,69 @@ func generateSecretVolume(clusterName string) v1.Volume {
...
+						Key: "url",
+						LocalObjectReference: v1.LocalObjectReference{
+							Name: "ci-slack-api-url",
+						},
+					},
+				},
+			},
+		}...)
+	}
+	if clusterName == string(api.ClusterVSphere) {
```
stevekuznetsov, line 207: or build01 or build02, don't dupe this block

hongkailiu: I thought the ordering in the list mattered.
Updated.

stevekuznetsov (approved):

PR #2382, opened 2021-09-20
Cluster-init: update secret bootstrap, secret generator, sanitize prow jobs, and build dirs for existing clusters

smg247:
/cc @openshift/test-platform 


cmd/cluster-init/main.go
```diff
@@ -220,9 +220,16 @@ func submitPR(o options) error {
 	return exec.Command("git", "checkout", master).Run()
 }
 
-func initClusterBuildFarmDir(o options) error {
+func updateClusterBuildFarmDir(o options) error {
 	buildDir := buildFarmDirFor(o.releaseRepo, o.clusterName)
-	logrus.Infof("Creating build dir: %s", buildDir)
+	if o.update {
+		logrus.Infof("Updating build dir: %s", buildDir)
+		if err := os.RemoveAll(buildDir); err != nil {
```
hongkailiu, line 227: If we remove all, it will remove all the contents that are manually added, right? Something like, 

```
init build03
add some.yaml into build03 folder manually
update build03: this will remove some.yaml
```

How are we going to recover it?

smg247: yes, I was under the impression that we wouldn't want to be adding anything manually there. I guess we could just remove the 2 symlinks and leave the rest in tact. thoughts?

hongkailiu: Removing (and recreating later) the symlinks sounds good to me.

I think in general: if our tool expects
- a file (completed controlled by our tool) or a symlink, we remove and create
- a dir, we do nothing

cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -274,8 +280,14 @@ func generateRegistryPullCredentialsAllSecrets(c *secretbootstrap.Config, o opti
 			generateDockerConfigJsonSecretConfigTo(regPullCredsAll, testCredentials, o.clusterName),
 		},
 	}
-	logrus.Infof("Creating new secret with 'to' of: %v", sc.To)
-	c.Secrets = append(c.Secrets, sc)
+	idx, _, _ := findSecretConfig(regPullCredsAll, o.clusterName, c.Secrets)
```
stevekuznetsov, line 283: Why is this not using `updateSecret()` ?

cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -257,7 +263,7 @@ func generateRegistryPullCredentialsAllSecrets(c *secretbootstrap.Config, o opti
 		},
 	}
 	for _, cluster := range c.UserSecretsTargetClusters {
-		if cluster != string(api.ClusterHive) {
+		if cluster != string(api.ClusterHive) && cluster != o.clusterName {
```
stevekuznetsov, line 266: Why?

cmd/cluster-init/main.go
```diff
@@ -235,10 +242,19 @@ func initClusterBuildFarmDir(o options) error {
...
-func buildFarmDirFor(releaseRepo string, clusterName string) string {
+func buildFarmDirFor(releaseRepo, clusterName string) string {
 	return filepath.Join(releaseRepo, "clusters", "build-clusters", clusterName)
 }
 
 func serviceAccountKubeconfigPath(serviceAccount, clusterName string) string {
 	return fmt.Sprintf("sa.%s.%s.config", serviceAccount, clusterName)
 }
+
+func appendIfNotContains(slice []string, item string) []string {
```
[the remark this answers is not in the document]
smg247: I had considered converting it to a set, but wasn't sure what the best practice there was. If it's standard to just use `sets.NewString` I will do so.

cmd/cluster-init/main.go
```diff
@@ -220,25 +220,44 @@ func submitPR(o options) error {
...
+			return fmt.Errorf("failed to create base directory for cluster: %w", err)
+		}
 	}
-
 	for _, item := range []string{"common", "common_except_app.ci"} {
-		if err := os.Symlink(fmt.Sprintf("../%s", item), filepath.Join(buildDir, item)); err != nil {
+		target := fmt.Sprintf("../%s", item)
+		source := filepath.Join(buildDir, item)
+		if o.update {
+			if err := os.RemoveAll(source); err != nil {
```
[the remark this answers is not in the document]
smg247: This is _only_ removing the symlinks that we know to generate. I made that change after Hongkai was concerned about me blowing away the entire build dir.

cmd/cluster-init/update_bootstrap_secrets.go
smg247: we also generate that exact `SecretItem` in `generatePushPullSecretFrom`. I only noticed it when running in update mode with the existing build clusters.

cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -67,11 +67,17 @@ func updateCiSecretBootstrapConfig(o options, c *secretbootstrap.Config) error {
 	return nil
 }
 
-func appendSecret(secretGenerator func(options) secretbootstrap.SecretConfig) func(c *secretbootstrap.Config, o options) error {
+func updateSecret(secretGenerator func(options) secretbootstrap.SecretConfig) func(c *secretbootstrap.Config, o options) error {
```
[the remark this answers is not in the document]
smg247: No, but now that it is replacing the secret in place its probably worth doing.

cmd/cluster-init/update_bootstrap_secrets.go
smg247: `updateSecret`'s contract takes a `generatorFunction` that _only_ takes `options` as an argument. Here, we also need either the `secretbootstrap.Config` (or simply the list of `UserSecretsTargetClusters`) in order to properly generate the items.

cmd/cluster-init/main.go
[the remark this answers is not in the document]
smg247: Actually, doing so results in a new ordering of all the configs we touch. It is deterministic, so we could just run in `update` mode and commit the new ordering. Thoughts?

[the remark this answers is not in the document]
stevekuznetsov: I see - the default behavior of `.List()` is to sort. Likely useful to have these sorted so that new additions are deterministic (they would be order-dependent today).

cmd/cluster-init/update_bootstrap_secrets.go
stevekuznetsov: Could you not adapt it, like:

```go
func wrapped(c *secretbootstrap.Config) func(o options) error {
    return func(o options) error {
        for _, cluster := range c.UserSecretsTargetClusters {
            //...
        }
    }
}
```

cmd/cluster-init/update_bootstrap_secrets.go
stevekuznetsov: Ok, this might use a comment - it's not obvious from the logic itself why the code exists

cmd/cluster-init/update_bootstrap_secrets.go
smg247: yeah that works!

PR #2390, opened 2021-09-23
stop collecting disruption from jobs that don't have disruption data

deads2k:
To use this, I delete all content from Jobs and then run the prime jobs command.

It takes a lot of time to pull every gcs bucket for jobs that have no disruption content


pkg/jobrunaggregator/jobtableprimer/migration_test.go
```diff
@@ -1,238 +0,0 @@
-package jobtableprimer
-
-import (
-	"reflect"
-	"testing"
-
-	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
-)
-
-func TestMigrationOfJobAttributes(t *testing.T) {
```
deads2k, line 10: I only created this to be sure that I got the same content I had manually entered.  it's no longer needed since we create via command now.

dgoodwin (commented):
/lgtm

PR #2398, opened 2021-09-28
Make m(github_login)=kerberos_id as ci-op-xxxx's admin

hongkailiu:

/cc @openshift/test-platform

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -1202,23 +1202,26 @@ func (o *options) initializeNamespace() error {
 		return errors.New("timed out waiting for image pull secrets")
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/ci-operator/main.go
```diff
@@ -1316,6 +1321,30 @@ func (o *options) initializeNamespace() error {
 	return nil
 }
 
+func generateAuthorAccessRoleBinding(namespace string, authors []string, mappingLoadFunc func() map[string]string) *rbacapi.RoleBinding {
```
emilvberglind, line 1324: Do we get anything out of passing in the `mappingLoadFunc` vs just passing in the mappings?

hongkailiu: Updated.

PR #2399, opened 2021-09-28
ignore jobs we're aggregating that time out

deads2k:
if job runs have timed out, we can ignore them for the purposes of aggregation.  This only becomes a practical problem if the jobs start running very very long.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
@@ -10,8 +10,9 @@ import (
 	"strings"
 	"time"
 
+	"gopkg.in/yaml.v2"
+
 	"k8s.io/apimachinery/pkg/util/clock"
-	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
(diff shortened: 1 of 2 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
```diff
@@ -66,8 +65,9 @@ func (o *JobRunAggregatorAnalyzerOptions) getRelatedJobs(ctx context.Context) ([
 func (o *JobRunAggregatorAnalyzerOptions) Run(ctx context.Context) error {
 	// if it hasn't been more than hour since the jobRuns started, the list isn't complete.
 	readyAt := o.jobRunStartEstimate.Add(1 * time.Hour)
+	timeToStopWaiting := o.jobRunStartEstimate.Add(3*time.Hour + 10*time.Minute)
```
dgoodwin, line 68: Eyeballing sippy and prox: I see some 4.10 Azure ovn upgrade jobs passing after 3:20 pretty consistently, so might need a little more time here.
AWS around this range, 2:45 -> 3:05, 
GCP a little faster around 2:35 often.

deads2k: > Eyeballing sippy and prow: I see some 4.10 Azure ovn upgrade jobs passing after 3:20 pretty consistently, so might

there is no more time.  this is pressing up against the max ci-operator time

wking: In #22289, I'm talking about this a bit, and I think we might need to talk the test-platform folks into raising the Plank/Prow timeout above its current 4h, or giving us a way to do that for particular generated jobs (there's [already a way][1] to raise it for non-generated jobs via `decoration_config.timeout`).

[1]: https://docs.ci.openshift.org/docs/architecture/timeouts/#prow-test-process-timeouts

dgoodwin: And IIRC we're not planning on aggregating ovn-azure anyhow?

deads2k: > And IIRC we're not planning on aggregating ovn-azure anyhow?

we will

dgoodwin (commented):
Looks like we've got something to search on to see how much we're hitting this.

/lgtm

PR #2401, opened 2021-09-28
prowgen: mount configmap/github-ldap-mapping

hongkailiu:

/cc @openshift/test-platform 

/hold
will bring the fix of breaking on release repo.

### DIFF
--- a/cmd/ci-operator-prowgen/testdata/zz_fixture_postsubmit_TestFromCIOperatorConfigToProwYaml_Custom_test_timeout.yaml
+++ b/cmd/ci-operator-prowgen/testdata/zz_fixture_postsubmit_TestFromCIOperatorConfigToProwYaml_Custom_test_timeout.yaml
@@ -51,6 +51,9 @@ postsubmits:
         - mountPath: /secrets/gcs
           name: gcs-credentials
           readOnly: true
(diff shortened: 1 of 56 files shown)


droslean:
To be honest, I am not convinced if this is the best way we can come up with, but for now, we can live with that :). I would prefer to have a small tool running periodically (reading the mapping file), that creates user groups in app.ci cluster named as GH name, and make ci-operator to create RBAC for the group instead. With this way everybody is happy, and we don't have to mount extra stuff in the jobs.

petr-muller:
> To be honest, I am not convinced if this is the best way we can come up with, but for now, we can live with that :). I would prefer to have a small tool running periodically (reading the mapping file), that creates user groups in app.ci cluster named as GH name, and make ci-operator to create RBAC for the group instead (see https://github.com/openshift/ci-tools/blob/master/cmd/ci-operator/main.go#L1325). With this way everybody is happy, and we don't have to mount extra stuff in the jobs.

That's smart, I like that. You'd need to maintain the groups not only on app.ci but on all clusters, but that should not be that hard? The tool would need to take care of removing stale groups, too.

hongkailiu:
/hold

Group sounds a better strategy.

PR #2403, opened 2021-09-28
update required passcounts

deads2k:
like https://github.com/openshift/ci-tools/pull/2402 but with 98% assurance the test got worse.  this might be good given the total number of tests

/hold


deads2k:
I think I like 95 better

PR #2405, opened 2021-09-29
a new tool: github-ldap-user-group-creator

hongkailiu:

### DIFF
--- a/cmd/github-ldap-user-group-creator/main.go
+++ b/cmd/github-ldap-user-group-creator/main.go
@@ -1,0 +1,215 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/kube"
+	"k8s.io/test-infra/prow/logrusutil"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
+
+	userv1 "github.com/openshift/api/user/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+type options struct {
+	kubernetesOptions flagutil.KubernetesOptions
+
+	mappingFile string
+	logLevel    string
+	dryRun      bool
+}
+
+func parseOptions() *options {
+	opts := &options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	opts.kubernetesOptions.AddFlags(fs)
+	fs.StringVar(&opts.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
+	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.StringVar(&opts.mappingFile, "mapping-file", "", "File to the mapping results of m(github_login)=kerberos_id.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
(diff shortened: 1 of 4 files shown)


cmd/github-ldap-user-group-creator/main.go
```diff
@@ -0,0 +1,215 @@
...
+	dryRun      bool
+}
+
+func parseOptions() *options {
+	opts := &options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	opts.kubernetesOptions.AddFlags(fs)
+	fs.StringVar(&opts.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
+	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.StringVar(&opts.mappingFile, "mapping-file", "", "File to the mapping results of m(github_login)=kerberos_id.")
```
droslean, line 44: Why this tool can't generate the mapping itself?

hongkailiu: The tool which generates the mapping has to run inside RH network (the info to generate the mapping is RH LDAP server, which is inside the RH network).
The mapping itself is stored in our cluster though.
We want it as simple as possible: Just generate the mapping, no other logic is there.

petr-muller:
Code looks good on a skim!

/cc

cmd/github-ldap-user-group-creator/main.go
droslean: Does the tool that generates the mapping, that runs inside RH network, can reach the clusters?

hongkailiu: Yes as long as the machine hosting the tool has Internet.

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -0,0 +1,215 @@
...
+
+const (
+	appCIContextName = string(api.ClusterAPPCI)
+	toolName         = "github-ldap-user-group-creator"
+	groupSuffix      = "-group"
+)
+
+func addSchemes() error {
+	if err := userv1.AddToScheme(scheme.Scheme); err != nil {
+		return fmt.Errorf("failed to add imagev1 to scheme: %w", err)
```
droslean, line 72: The error here is wrong.

hongkailiu: Updated.

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -0,0 +1,215 @@
...
+		}
+		var mapping map[string]string
+		if err := yaml.Unmarshal(bytes, &mapping); err != nil {
+			logrus.WithField("string(bytes)", string(bytes)).WithError(err).Debug("Failed to unmarshal")
+			return nil
+		}
+		return mapping
+	}(opts.mappingFile)
+
+	if err := ensureGroups(ctx, clients, mapping, opts.dryRun); err != nil {
```
droslean, line 145: since the tool will run once, it doesn't have to hold the mapping in an closure, since its going to marshal it once, right?

droslean: No, I mean that you can just pass the `map[string]string` directly.

hongkailiu:
```
func(path string, mapping *map[string]string)  {}
```
Do you want me to do something like ^?

droslean: I mean that the `ensureGroups()` should get a `map[string]string` instead of a closure.

droslean: You can just do 

```go

		logrus.WithField("path", opts.mappingFile).Debug("Loading the mapping file ...")
		bytes, err := ioutil.ReadFile(opts.mappingFile)
		if err != nil {
			logrus.WithField("path", opts.mappingFile).WithError(err).Debug("Failed to read file")
			return nil
		}
		var mapping map[string]string
		if err := yaml.Unmarshal(bytes, &mapping); err != nil {
			logrus.WithField("string(bytes)", string(bytes)).WithError(err).Debug("Failed to unmarshal")
			return nil
		}

	if err := ensureGroups(ctx, clients, mapping, opts.dryRun); err != nil {
```

hongkailiu: What about the `return`s?

hongkailiu: As we discussed in slack, I would just leave the code as it is.
I think it is just a different taste of code style, not correctness or performance.

droslean (approved):

PR #2407, opened 2021-09-30
write html status for aggregator

deads2k:
Creates an html summary we'll be able to add to spyglass

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/spyglass_summary.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/spyglass_summary.go
@@ -1,0 +1,222 @@
+package jobrunaggregatoranalyzer
+
+import (
+	"context"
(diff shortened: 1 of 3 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/spyglass_summary.go
```diff
@@ -0,0 +1,222 @@
...
+%v
+%v
+`,
+		status,
+		strings.Join(parents, "    "),
+		testCase.Name,
+		currDetails.Summary,
+		failureHTML,
+		flakeHTML)
+	html += "</li>\n<br/>\n"
```
dgoodwin, line 219: Is it useful to include tests that ran and had no flakes/failures, seems unnecessary to have all 3000 tests in here.

dgoodwin:
/lgtm

Thanks!

PR #2408, opened 2021-09-30
DPTP-2523: Clear out lastLoaded CachedQuery once used

smg247:
The lastLoaded CachedQuery is very large. Once it is digested there is no reason for it to hang around. This significantly reduces the memory use of the pod-scaler. See pprof results for running the `local-pod-scaler-ui` make target before and after this change:
before_change
 
after_change

I also added the ability to specify the amount of seconds to wait for an accessory service to be ready. 90 seconds was not enough for me to run the `local-pod-scaler-ui` target locally, it actually took around 180 seconds each time.

### DIFF
--- a/cmd/pod-scaler/consumer.go
+++ b/cmd/pod-scaler/consumer.go
@@ -87,10 +87,13 @@ func (c *cacheReloader) reload() {
 	logger.Debug("Newer update loaded.")
 }
 
+// data returns the lastLoaded CachedQuery and clears it to conserve memory
 func (c *cacheReloader) data() *pod_scaler.CachedQuery {
 	c.lock.RLock()
 	defer c.lock.RUnlock()
-	return c.lastLoaded
+	lastLoaded := c.lastLoaded
+	c.lastLoaded = nil
+	return lastLoaded
 }
 
 func digestAll(data map[string][]*cacheReloader, digesters map[string]digester, health *pjutil.Health, logger *logrus.Entry) {
--- a/cmd/pod-scaler/frontend.go
+++ b/cmd/pod-scaler/frontend.go
@@ -470,7 +470,7 @@ func (s *frontendServer) digestCPU(data *pod_scaler.CachedQuery) {
 }
 
 func (s *frontendServer) digestMemory(data *pod_scaler.CachedQuery) {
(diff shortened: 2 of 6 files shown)


petr-muller (approved):
Code LGTM but I'm not certain I can fully vet the "Once it is digested there is no reason for it to hang around" claim with my current knowledge of pod-scaler. But looks right from what I was able to check. Feel free to unhold, or make someone (Steve?) confirm if needed.

/hold

bbguimaraes:
A few things:
- Separate the `testhelper` changes into another commit.
- The production servers are at a constant minimum of 5GB of RSS.  We should take at least one `pprof` memory dump there to make sure this is the source of the memory usage.  
- Clearing the data on first access does not really work with the current API.  When signaled, the first subscriber to lock the mutex will get the latest data and all others will miss it:
https://github.com/openshift/ci-tools/blob/5f4c96d0f5492dbaca3a8f2ab337a3a87e94bcb8/cmd/pod-scaler/consumer.go#L80-L86
https://github.com/openshift/ci-tools/blob/5f4c96d0f5492dbaca3a8f2ab337a3a87e94bcb8/cmd/pod-scaler/consumer.go#L137-L157

smg247:
@bbguimaraes to your points:

- I can separate those out if desired, but it seemed minor enough that the overhead wasn't worth it there.
- Getting a pprof dump from app.ci was the first thing I did here, I just hadn't shared it on this PR: 
production
- From looking through the code I had thought _each_ subscriber created its own `cacheReloader`, but now I see where there is just a separate one created for the UI and the admission server. This will need to be rethought to make sure every subscriber has seen it prior to clearing it out.

smg247:
@bbguimaraes , I've updated this to track the number of subscribers that have seen the data, and only clear once they all have. Interestingly enough none of the e2e tests (including the `make local-pod-scaler-ui`) create a scenario where there are multiple subscribers for any one piece of data. I tried to modify them so that scenario would come out, but admitting multiple pods and using multiple admission servers did not result in any one `cacheReloader` having more than 1 subscriber. I added some extra logging (and an error scenario) so we can determine if this ever happens. If you know of some scenario I am missing let me know and I can write a test for it.

PR #2416, opened 2021-10-04
ignore aggregation result for generic upgrade test run

deads2k:
This test is a side-effect of other tests.  For the purpose of aggregation, we can have each individual job run fail this test, but the aggregated output can be successful.


deads2k:
https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/aggregated-azure-sdn-upgrade-4.10-minor-release-openshift-release-analysis-aggregator/1444969182415294464 is a good example.

PR #2417, opened 2021-10-04
[WIP] Aggregate disruption tests separate, and determine pass/fail based on historical mean

stbenjam:
- Gets the weekly historical average from 10 days ago
- Aggregates disruption tests separately from other tests
- Fails if > 110% of historical average


pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
```diff
@@ -187,14 +190,22 @@ func aggregateTestCase(combined *junit.TestCase, jobName, toAddJobRunID string,
...
 	case toAdd.FailureOutput != nil:
 		humanURL := jobrunaggregatorapi.GetHumanURL(jobName, toAddJobRunID)
+		testCaseFailure := TestCaseFailure{
+			JobRunID:       toAddJobRunID,
+			HumanURL:       humanURL,
+			GCSArtifactURL: jobrunaggregatorapi.GetGCSArtifactURL(jobName, toAddJobRunID),
+		}
+		if jobrunaggregatorlib.IsDisruptionTest(toAdd.Name) {
+			disruptionSeconds, err := jobrunaggregatorlib.GetOutageSecondsFromMessage(toAdd.FailureOutput.Message)
+			if err == nil {
```
deads2k, line 200: nit: I prefer the `!=` because that's the dominant way to see it.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -47,25 +50,39 @@ func assignPassFailForTestSuite(ctx context.Context, parentTestSuites []string,
...
 
-		if !failed {
-			continue
-		}
-		currTestCase.FailureOutput = &junit.FailureOutput{
-			Message: message,
-			Output:  currTestCase.SystemOut,
+		if jobrunaggregatorlib.IsDisruptionTest(currTestCase.Name) {
+			failed, message := baselinePassFail.CheckDisruption(ctx, currDetails)
+			currDetails.Summary = message
```
deads2k, line 56: need to set `currTestCase.SystemOut`, right?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -246,7 +321,11 @@ func testShouldAlwaysPass(name string) bool {
 		case strings.HasSuffix(name, "-openshift-e2e-test container test"):
 			return true
 		}
+	} else if jobrunaggregatorlib.IsDisruptionTest(name) {
+		// For disruption tests, we'll calculate test/fail later based on actual durations.
+		return true
```
deads2k, line 326: this should never be called right?  If it should never be called, let's return some crazy message about never being here and force a failure so we notice.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -137,6 +140,47 @@ LIMIT 1
 	return lastJobRun, nil
 }
 
+func (c *ciDataClient) GetMeanDisruptionByJob(ctx context.Context, jobName string) ([]jobrunaggregatorapi.BackendDisruptionMeanRow, error) {
+	rows := make([]jobrunaggregatorapi.BackendDisruptionMeanRow, 0)
+
+	queryString := c.dataCoordinates.SubstituteDataSetLocation(`
+SELECT
```
deads2k, line 147: this query looks correct to me.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"fmt"
+	"math"
+	"regexp"
+	"strings"
+	"time"
+)
+
+var (
+	UpgradeBackendNameToTestSubstring = map[string]string{
```
deads2k, line 12: straight move?

stbenjam: Yea direct move to lib, since it's being used in the analyzer now too

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
stbenjam: Leftover from a previous approach, removed.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
```diff
@@ -187,14 +190,22 @@ func aggregateTestCase(combined *junit.TestCase, jobName, toAddJobRunID string,
 	switch {
 	case toAdd.FailureOutput != nil:
 		humanURL := jobrunaggregatorapi.GetHumanURL(jobName, toAddJobRunID)
+		testCaseFailure := TestCaseFailure{
+			JobRunID:       toAddJobRunID,
+			HumanURL:       humanURL,
+			GCSArtifactURL: jobrunaggregatorapi.GetGCSArtifactURL(jobName, toAddJobRunID),
+		}
+		if jobrunaggregatorlib.IsDisruptionTest(toAdd.Name) {
+			disruptionSeconds, err := jobrunaggregatorlib.GetOutageSecondsFromMessage(toAdd.FailureOutput.Message)
```
deads2k, line 199: this should be updated now that we have separate durations provided.  Possible followup if this is close enough to merge this afternoon.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
```diff
@@ -187,14 +190,22 @@ func aggregateTestCase(combined *junit.TestCase, jobName, toAddJobRunID string,
...
+			if err != nil {
+				fmt.Fprintf(os.Stderr, "Failed to get disruption seconds from %q\n", toAdd.FailureOutput.Message)
+			} else {
+				testCaseFailure.DisruptionSeconds = disruptionSeconds
```
deads2k, line 203: we can succeed and have more than zero disruptionseconds.  I think that needs to be reflected here as well.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -141,6 +144,84 @@ LIMIT 1
...
+                    JobRuns.JobName = @JobName
+            )
+            GROUP BY
+                BackendName
+      ) p95
+LEFT JOIN
+    (
+        SELECT
+            BackendName,
+            AVG(BackendDisruption.DisruptionSeconds) as Mean
```
deads2k, line 182: Can't do this in the same select as the percentile?

pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
```diff
@@ -17,8 +14,46 @@ import (
...
+	backendDisruptionInserter jobrunaggregatorlib.BigQueryInserter
+}
+
+func newDisruptionUploader(backendDisruptionInserter jobrunaggregatorlib.BigQueryInserter) uploader {
+	return &disruptionUploader{
+		backendDisruptionInserter: backendDisruptionInserter,
+	}
+}
+
+func (o *disruptionUploader) uploadContent(ctx context.Context, jobRun jobrunaggregatorapi.JobRunInfo, prowJob *prowv1.ProwJob) error {
```
deads2k, line 31: a move or logic change?

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
stbenjam: I couldn't get it to work

pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
stbenjam: What do you mean by separate durations?

stbenjam:
/retitle [wip] Aggregate disruption tests and determine pass/fail based on historical mean

Working on refactoring this

pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
```diff
@@ -18,168 +14,6 @@ import (
...
-	if len(matches) < 2 {
-		return 0, fmt.Errorf("not the expected format: %v", message)
-	}
-	outageDuration, err := time.ParseDuration(matches[1])
-	if err != nil {
-		return 0, err
-	}
-	return int(math.Ceil(outageDuration.Seconds())), nil
-}
-
```
stbenjam, line 182: Direct move to lib since it's shared between uploader + analyzer now.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -0,0 +1,196 @@
...
+
+var (
+	upgradeBackendNameToTestSubstring = map[string]string{
+		"kube-api-new-connections":                          "Kubernetes APIs remain available for new connections",
+		"kube-api-reused-connections":                       "Kubernetes APIs remain available with reused connections",
+		"openshift-api-new-connections":                     "OpenShift APIs remain available for new connections",
+		"openshift-api-reused-connections":                  "OpenShift APIs remain available with reused connections",
+		"oauth-api-new-connections":                         "OAuth APIs remain available for new connections",
+		"oauth-api-reused-connections":                      "OAuth APIs remain available with reused connections",
+		"service-load-balancer-with-pdb-reused-connections": "Application behind service load balancer with PDB is not disrupted",
```
stbenjam, line 24: There's something wrong with this one.

Reused and new I think are part of the same test -- https://github.com/openshift/origin/blob/master/test/e2e/upgrade/service/service.go#L42

https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-gcp-ovn-upgrade/1454701479095963648 failed on new connections, but it was persisted to bigquery as "service-load-balancer-with-pdb-reused-connections".

I think this needs to be renamed to `service-load-balancer-with-pdb` and the bigquery rows updated? @deads2k WDYT?

deads2k:
> * Allows 1 outlier > P95

excludes 1 outlier > P95 ?

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
deads2k: The disruption is calculated in two steps.  Reused connections are here https://github.com/openshift/origin/blob/master/test/e2e/upgrade/service/service.go#L198-L230

new connections are here https://github.com/openshift/origin/blob/master/test/e2e/upgrade/service/service.go#L246-L278

both produce uniquely identifiable monitor events, which are then added individually by the code writing the backenddisruption json file.

PR #2418, opened 2021-10-04
ci-operator: use lowercase login for setting up author RBAC

petr-muller:
Looks like our group data source is feeding us lowecase variants of GH usernames, our groups are all lowecase. See e.g.:

```console
$ b01 get group alexnpavel-group
NAME               USERS
alexnpavel-group   alexnpavel, apavel
```

Therefore we need to use lowecase in ci-operator when setting up author RBAC.


hongkailiu (approved):

hongkailiu:
/hold

Let me ask @petr-muller over slack.

petr-muller:
/close

@hongkailiu is right, this needs to be fixed in the producer, not here. The actual logins (and groups) should respect the case.

PR #2422, opened 2021-10-05
ci-op: fix rolebinding/ci-op-author-access creating

hongkailiu:
- remove the loop on authors: this bug should have been there before the group support feature. This should help https://coreos.slack.com/archives/CEKNRGF25/p1633445908185500
- handle the duplicated authors. https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/release-openshift-origin-installer-launch-gcp/1445412322951565312/artifacts/initial/ci-operator.log Not sure where the authors came from though.

/cc @openshift/test-platform

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -1205,12 +1205,11 @@ func (o *options) initializeNamespace() error {
 	if o.givePrAuthorAccessToNamespace && len(o.authors) > 0 {
 		roleBinding := generateAuthorAccessRoleBinding(o.namespace, o.authors)
 		// Generate rolebinding for all the PR Authors.
-		for _, author := range o.authors {
-			logrus.Debugf("Creating rolebinding for user %s in namespace %s", author, o.namespace)
-			if err := client.Create(ctx, roleBinding); err != nil && !kerrors.IsAlreadyExists(err) {
-				return fmt.Errorf("could not create role binding for: %w", err)
-			}
+		logrus.Debugf("Creating ci-op-author-access rolebinding in namespace %s", o.namespace)
+		if err := client.Create(ctx, roleBinding); err != nil && !kerrors.IsAlreadyExists(err) {
+			return fmt.Errorf("could not create role binding for: %w", err)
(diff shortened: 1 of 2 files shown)


cmd/ci-operator/main.go
```diff
@@ -1309,7 +1308,12 @@ func (o *options) initializeNamespace() error {
 
 func generateAuthorAccessRoleBinding(namespace string, authors []string) *rbacapi.RoleBinding {
 	var subjects []rbacapi.Subject
+	authorSet := sets.NewString()
```
petr-muller, line 1311:
```
authorSet := sets.NewString(authors...)
for author := range authorSet {
  subjects = append(subjects, rbacapi.Subject{Kind: "Group", Name: author + api.GroupSuffix})
}

**Review by petr-muller — COMMENTED** [review]
Hmm... Looks correct but I don't see how this could cause the 

```
could not initialize namespace: could not create role binding for: resourceVersion should not be set on objects to be created
```

issue...:(

Otherwise, just a single code simplification proposal.

File: cmd/ci-operator/main.go (hunk 1 above)
Comment (hongkailiu) [reply cmd/ci-operator/main.go]: Updated.

**Comment by hongkailiu** [comment]
> Hmm... Looks correct but I don't see how this could cause the
> 
> ```
> could not initialize namespace: could not create role binding for: resourceVersion should not be set on objects to be created
> ```

The idea is that there were two authors in the list.
```
{"level":"debug","msg":"Creating rolebinding for user system:serviceaccount:ci:ci-chat-bot in namespace ci-ln-7rks2wt","time":"2021-10-05T15:36:50Z"}
{"level":"debug","msg":"Creating rolebinding for user system:serviceaccount:ci:ci-chat-bot in namespace ci-ln-7rks2wt","time":"2021-10-05T15:36:50Z"}
```

Without the PR, the rolebinding got created twice and `resourceVersion` got set after the first creation.

droslean (approved):

PR #2428, opened 2021-10-13
makefile: fix kubeconfig creation target

petr-muller:


petr-muller:
/hold

Looks genuinely broken

PR #2430, opened 2021-10-13
pod-scaler; add a hardcoded CPU request cap of 10

petr-muller:
A quick bandaid for [DPTP-2525](https://issues.redhat.com/browse/DPTP-2525). The proper solution is likely
configurable and smarter, but this will do the job for now, preventing
CPU-intensive builds/jobs from increasing its requests until
unschedulable.

I do not want to spend time fully solving [DPTP-2525](https://issues.redhat.com/browse/DPTP-2525)
but I want to eliminate the failure case fast, because it
causes helpdesk requests that are hard to resolve.

10 CPU cap is a ballpark number, our workers have 16 and we saw unschedulable pods around 14.


petr-muller:
/test e2e
r-c outage...

droslean (approved):

PR #2433, opened 2021-10-18
[schedule-generator] add the ability to validate the schedule configuration

droslean:
/cc @openshift/test-platform 

This PR adds adds the ability in the tool to validate the schedule configuration. There is also some refactoring that removes the duplicated types and the mapping and uses the general types in ocplifestyle instead.

For better reviewing, enable `Hide whitespace` option.

### DIFF
--- a/cmd/branchingconfigmanagers/schedule-generator/main.go
+++ b/cmd/branchingconfigmanagers/schedule-generator/main.go
@@ -27,6 +27,7 @@ type options struct {
 	logLevel          string
 	scheduleDir       string
 	dryRun            bool
+	validateOnly      bool
 	kubernetesOptions flagutil.KubernetesOptions
 }
 
@@ -37,6 +38,7 @@ func gatherOptions() (*options, error) {
 	fs.StringVar(&o.scheduleDir, "schedule-dir", "", "Directory holding schedules.")
 	o.kubernetesOptions.AddFlags(fs)
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Do not mutate cluster state.")
+	fs.BoolVar(&o.validateOnly, "validate-only", false, "Whether to only validate the schedule files.")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		return o, fmt.Errorf("failed to parse flags: %w", err)
 	}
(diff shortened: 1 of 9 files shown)


cmd/branchingconfigmanagers/schedule-generator/testdata/schedules/4.1.yaml
```diff
@@ -6,7 +6,7 @@ events:
     date: "2020-05-05T16:00:00Z"
   - name: "end-of-full-support"
     date: "2019-11-16T16:00:00Z"
-  - name: "general-availability"
+  - name: "generally-available"
```
petr-muller, line 9: @stevekuznetsov was there a good reason for why these names should be different in the source and target schedule files?

cmd/branchingconfigmanagers/schedule-generator/testdata/zz_fixture_TestReadSchedules.yaml
```diff
@@ -1,7 +1,9 @@
 ocp:
   "4.1":
-  - event: end-of-life
+  - event: end-of-maintenance-support
     when: "2020-05-05T16:00:00Z"
```
petr-muller, line 4: Having two events instead of one looks good here :+1:

cmd/branchingconfigmanagers/schedule-generator/main.go
```diff
@@ -48,9 +50,29 @@ func validateOptions(o *options) error {
...
+		return fmt.Errorf("--schedule-dir must be specified")
+	}
 	return o.kubernetesOptions.Validate(o.dryRun)
 }
 
+func validateSchedules(schedules ocplifecycle.Config) error {
+	var errs []error
+	for _, lifecyclePhaseByVersions := range schedules {
+		for version, lifecyclePhases := range lifecyclePhaseByVersions {
+			if _, err := ocplifecycle.ParseMajorMinor(version); err != nil {
```
petr-muller, line 63: One thing that I'd like to have validated is that the events are sorted timewise (descending)

cmd/branchingconfigmanagers/schedule-generator/main.go
```diff
@@ -48,9 +50,38 @@ func validateOptions(o *options) error {
...
+				errs = append(errs, fmt.Errorf("invalid version: %s", version))
+			}
+			for i, lifecyclePhase := range lifecyclePhases {
+				if err := lifecyclePhase.Event.Validate(); err != nil {
+					errs = append(errs, fmt.Errorf("unknown event: %s", lifecyclePhase.Event))
+				}
+
+				date := lifecyclePhase.When
```
smg247, line 71: this initialization can go after the proceeding if statement

petr-muller (approved):

PR #2448, opened 2021-10-25
payload testing: create PullRequestPayloadQualificationRun

hongkailiu:
https://issues.redhat.com/browse/DPTP-2540

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -1,17 +1,23 @@
 package main
 
 import (
+	"context"
 	"fmt"
 	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/sirupsen/logrus"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/kube"
 	"k8s.io/test-infra/prow/pluginhelp"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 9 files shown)


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -50,15 +61,60 @@ type jobSetSpecification struct {
 	jobs        jobType
 }
 
-func triggerReleaseJobs(specification jobSetSpecification, meta api.Metadata, tests []string) error {
-	// TODO(DPTP-2540): Translate the list of job names to list of (org, repo, branch, variant, test) tuples
-	// TODO(DPTP-2540): Create a CR for the controller to pick up
+func triggerReleaseJobs(params *releaseJobParams, jobTuples []api.MetadataWithTest) error {
```
petr-muller, line 64: While reading this, I think we can still improve the code structure here for readability and testability:

- Call `releaseJobParams` something like `PRPQRBuilder` and do not include the kubeclient and context bits in it
- Make this method a `func (b *PRPQRBuilder) build(jobs []api.MetadataWithTest) *prpqv1.PullRequestPayloadQualificationRun` method and just make it return a struct 

That way, it will be better testable, because you will not need to test it via a mock client, it will be an input->output function with just a minimal side effect (the counter).

At callsite, you can then simply do:

```golang
run := builder.build(jobs)
if err := client.Create(ctx, run); err != nil {
	return fmt.Errorf("failed to create PullRequestPayloadQualificationRun: %w", err)
}
```

That way, 1) structs are simpler 2) methods are simpler and follow the single-responsibitliy principle.

hongkailiu: Updated.

petr-muller (approved):

petr-muller:
Thanks!

PR #2452, opened 2021-10-27
Payload testing UI

bbguimaraes:
[preview](https://payload-testing-ui-bbguimaraes.apps.ci.l2s4.p1.openshiftapps.com/runs)

https://issues.redhat.com/browse/DPTP-2542

### DIFF
--- a/pkg/api/pullrequestpayloadqualification/v1/types.go
+++ b/pkg/api/pullrequestpayloadqualification/v1/types.go
@@ -5,6 +5,7 @@ import (
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 
 	"github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/jobconfig"
 )
 
 const (
@@ -122,4 +123,10 @@ type PullRequestPayloadQualificationRunList struct {
 	metav1.ListMeta `json:"metadata"`
 
 	Items []PullRequestPayloadQualificationRun `json:"items"`
+}
+
+// JobName maps the name in the spec to the corresponding Prow job name.
+// It matches the `ReleaseJobName` value in the status.
+func (s ReleaseJobSpec) JobName() string {
+	return s.CIOperatorConfig.JobName(jobconfig.PeriodicPrefix, s.Test)
 }
--- a/cmd/payload-testing-ui/server.go
+++ b/cmd/payload-testing-ui/server.go
@@ -10,10 +10,12 @@ import (
 	"github.com/sirupsen/logrus"
 
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	prpqv1 "github.com/openshift/ci-tools/pkg/api/pullrequestpayloadqualification/v1"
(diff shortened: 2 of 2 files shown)


cmd/payload-testing-ui/main.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	return nil
+}
+
+const (
+	appCIContextName = string(api.ClusterAPPCI)
+)
+
+var (
+	metrics = prowmetrics.NewMetrics("configresolver")
```
petr-muller, line 59: not configresolver ;)

cmd/payload-testing-ui/main.go
```diff
@@ -0,0 +1,117 @@
...
+)
+
+func l(fragment string, children ...simplifypath.Node) simplifypath.Node {
+	return simplifypath.L(fragment, children...)
+}
+
+func main() {
+	logrusutil.ComponentInit()
+	logger := logrus.WithField("plugin", "payload-testing")
```
petr-muller, line 68: Probably not a plugin?

cmd/payload-testing-ui/main.go
```diff
@@ -0,0 +1,117 @@
...
+		if !hasInClusterConfig {
+			logrus.WithError(err).Fatalf("had no context for '%s' and loading InClusterConfig failed", appCIContextName)
+		}
+		logrus.Infof("use InClusterConfig for %s", appCIContextName)
+		kubeconfigs[appCIContextName] = inClusterConfig
+	}
+	kubeConfig := kubeconfigs[appCIContextName]
+	kubeClient, err := ctrlruntimeclient.New(&kubeConfig, ctrlruntimeclient.Options{})
+	if err != nil {
+		logrus.WithError(err).Fatal("could not get Hive client for Hive kube config")
```
petr-muller, line 96: Not hive ;)

petr-muller (commented):
Mostly LGTM up to several copy-paste artifacts. I think it would be reasonable to merge something like this soon-ish and implement in increments, WDYT?

bbguimaraes:
Ready for review.  Changes from the previous version:

- I reverted to a simpler dump of `spec`/`status` for now until the final format is determined and the other components are populating them
- `main.go` is now a proper file

Ready for deployment in: https://github.com/openshift/release/pull/23358.

bbguimaraes:
Job mapping re-re-verted, now conforming to the internal discussion.

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -123,3 +124,9 @@ type PullRequestPayloadQualificationRunList struct {
 
 	Items []PullRequestPayloadQualificationRun `json:"items"`
 }
+
+// JobName maps the name in the spec to the corresponding Prow job name.
+// It matches the `ReleaseJobName` value in the status.
+func (s ReleaseJobSpec) JobName() string {
+	return s.CIOperatorConfig.SimpleJobName(jobconfig.PeriodicPrefix, s.Test)
```
petr-muller, line 131: Don't we need `JobName`?

(why do we enven have `SimpleJobName` / `JobName`?

bbguimaraes: You're right, that was based on a previous (erroneous) idea about what the job names were, I didn't notice when I reinstated it.  Corrected.

(I also found the number of random combinations of `Metadata -> string` functions amusing)

petr-muller (approved):

PR #2454, opened 2021-10-29
[DPTP-2541] Initial implementation of the job-trigger-controller-manager

droslean:
/cc @openshift/test-platform 

The tool includes two controllers. The `prpqr_reconciler` is responsible to watch create events for `prpqr` objects. It generates the corresponding `prowjobs` and updates the Status of the `prpqr` object.

The `prowjob_status_syncer` controller is responsible to watch `prowjobs` update events that hold the `pullrequestpayloadqualificationruns.ci.openshift.io` label, and updating the status in the corresponding `prpqr` object.

### DIFF
--- a/cmd/job-trigger-controller-manager/main.go
+++ b/cmd/job-trigger-controller-manager/main.go
@@ -1,0 +1,80 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/bombsimon/logrusr"
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/client-go/rest"
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	"k8s.io/test-infra/prow/logrusutil"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	ctrlruntimelog "sigs.k8s.io/controller-runtime/pkg/log"
+
+	prpqv1 "github.com/openshift/ci-tools/pkg/api/pullrequestpayloadqualification/v1"
+	"github.com/openshift/ci-tools/pkg/controller/prpqr_reconciler"
+)
+
+type options struct {
+	namespace string
+	dryRun    bool
+}
+
+func gatherOptions() (*options, error) {
+	o := &options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.StringVar(&o.namespace, "namespace", "ci", "In which namespace the operation will take place")
+
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		return o, fmt.Errorf("failed to parse flags: %w", err)
+	}
+	return o, nil
+}
+
+func main() {
+	logrusutil.ComponentInit()
+	controllerruntime.SetLogger(logrusr.NewLogger(logrus.StandardLogger()))
+
+	o, err := gatherOptions()
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to get options")
+	}
+
+	ctx := controllerruntime.SetupSignalHandler()
+
+	cfg, err := rest.InClusterConfig()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load in-cluster config")
+	}
+
+	mgr, err := controllerruntime.NewManager(cfg, controllerruntime.Options{
+		DryRunClient: o.dryRun,
+		Logger:       ctrlruntimelog.NullLogger{},
+	})
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to construct manager")
+	}
+
+	if err := prowv1.AddToScheme(mgr.GetScheme()); err != nil {
+		logrus.WithError(err).Fatal("Failed to add prowv1 to scheme")
+	}
+
+	if err := prpqv1.AddToScheme(mgr.GetScheme()); err != nil {
+		logrus.WithError(err).Fatal("Failed to add prpqv1 to scheme")
+	}
+
+	if err := prpqr_reconciler.AddToManager(mgr, o.namespace); err != nil {
+		logrus.WithError(err).Fatal("Failed to add prpqr_reconciler to manager")
+	}
+
+	if err := mgr.Start(ctx); err != nil {
+		logrus.WithError(err).Fatal("Manager ended with error")
+	}
+
+	logrus.Info("Process ended gracefully")
+}
--- a/pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
+++ b/pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
@@ -1,0 +1,123 @@
+package pjstatussyncer
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/types"
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
+	v1 "github.com/openshift/ci-tools/pkg/api/pullrequestpayloadqualification/v1"
+	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
+)
+
+const (
+	controllerName = "prowjob_status_syncer"
+)
+
+func AddToManager(mgr controllerruntime.Manager, ns string) error {
+	ctrl, err := controller.New(controllerName, mgr, controller.Options{
+		MaxConcurrentReconciles: 1,
+		Reconciler: &reconciler{
+			logger: logrus.WithField("controller", controllerName),
+
+			client: mgr.GetClient(),
+		},
+	})
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	// Watch only on updates
+	predicateFuncs := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return false },
+		DeleteFunc: func(event.DeleteEvent) bool { return false },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if _, ok := e.ObjectNew.GetLabels()[v1.PullRequestPayloadQualificationRunLabel]; !ok {
+				return false
+			}
+
+			if e.ObjectNew.GetNamespace() != ns {
+				return false
+			}
+
+			return true
+		},
+		GenericFunc: func(event.GenericEvent) bool { return false },
+	}
+
+	if err := ctrl.Watch(source.NewKindWithCache(&prowv1.ProwJob{}, mgr.GetCache()), pjHandler(), predicateFuncs); err != nil {
+		return fmt.Errorf("failed to create watch: %w", err)
+	}
+
+	return nil
+}
+
+func pjHandler() handler.EventHandler {
+	return handler.EnqueueRequestsFromMapFunc(func(o ctrlruntimeclient.Object) []reconcile.Request {
+		pj, ok := o.(*prowv1.ProwJob)
+		if !ok {
+			logrus.WithField("type", fmt.Sprintf("%T", o)).Error("Got object that was not a ProwJob")
+			return nil
+		}
+
+		return []reconcile.Request{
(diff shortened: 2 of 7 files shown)


cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,113 @@
+package main
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+
+	"github.com/bombsimon/logrusr"
```
petr-muller, line 9: Why do we need this? It does not look like the other our controllers use this, what makes this one special?

cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,113 @@
...
+		logrus.Info("Kubeconfig changed, exiting to get restarted by Kubelet and pick up the changes")
+		cancel()
+	}
+
+	kubeconfigs, err := o.kubernetesOptions.LoadClusterConfigs(kubeconfigChangedCallBack)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load kubeconfigs")
+	}
+
+	cfg := kubeconfigs[appCIContextName]
```
petr-muller, line 87: We need to error-check for the presence here

cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+type options struct {
+	kubernetesOptions flagutil.KubernetesOptions
+
+	namespace string
+	dryRun    bool
+}
+
+func gatherOptions() (*options, error) {
+	o := &options{kubernetesOptions: flagutil.KubernetesOptions{NOInClusterConfigDefault: true}}
```
petr-muller, line 35: Is there any reason why we shouldn't use in-cluster config? This controller will only interact with a single cluster (the one where it is running, so I'd guess we could even support in-cluster config exclusively (like the pools-cm does https://github.com/openshift/ci-tools/blob/c3f8a3cdfc64de1120d0c56b8badd690e8c9a6eb/cmd/dptp-pools-cm/main.go#L83-L86)

cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+	kubeconfigChangedCallBack := func() {
+		logrus.Info("Kubeconfig changed, exiting to get restarted by Kubelet and pick up the changes")
+		cancel()
+	}
+
+	kubeconfigs, err := o.kubernetesOptions.LoadClusterConfigs(kubeconfigChangedCallBack)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to load kubeconfigs")
+	}
```
petr-muller, line 85: I'd expect code like we do in other places:

https://github.com/openshift/ci-tools/blob/c3f8a3cdfc64de1120d0c56b8badd690e8c9a6eb/cmd/dptp-controller-manager/main.go#L247-L257

cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,113 @@
...
+
+	cfg := kubeconfigs[appCIContextName]
+	mgr, err := controllerruntime.NewManager(&cfg, controllerruntime.Options{
+		DryRunClient: o.dryRun,
+		Logger:       ctrlruntimelog.NullLogger{},
+	})
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to construct manager: %w", err)
```
petr-muller, line 93: no need to inline the err, it's in `WithError` already

pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to construct controller: %w", err)
+	}
+
+	// Watch only on updates
+	predicateFuncs := predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool { return false },
+		DeleteFunc: func(event.DeleteEvent) bool { return false },
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			if _, ok := e.ObjectNew.GetLabels()["pullrequestpayloadqualificationruns.ci.openshift.io"]; !ok {
```
petr-muller, line 46: extract the label name to a const

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -0,0 +1,190 @@
...
+	prowjobs := generateProwjobs(prpqr.Spec.PullRequest.Org, prpqr.Spec.PullRequest.Repo, prpqr.Spec.PullRequest.BaseRef, req.Name, req.Namespace)
+	for _, pj := range prowjobs {
+		logger = logger.WithFields(logrus.Fields{"name": pj.Name, "namespace": req.Namespace})
+
+		err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: pj.Namespace, Name: pj.Name}, &prowv1.ProwJob{})
+		if err != nil && !kerrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get the PullRequestPayloadQualificationRun: %s in namespace %s: %w", req.Name, req.Namespace, err)
+		}
+
+		if kerrors.IsNotFound(err) {
```
petr-muller, line 119: Maybe invert this to a guard clause (`if !IsNotFound { continue }`) and shift the code left?

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -0,0 +1,190 @@
...
+					}
+					return false, fmt.Errorf("getting prowJob failed: %w", err)
+				}
+				return true, nil
+			}); err != nil {
+				return fmt.Errorf("failed to wait for created ProwJob to appear in cache: %w", err)
+			}
+
+			prpqr.Status.Jobs = append(prpqr.Status.Jobs, prpqv1.PullRequestPayloadJobStatus{
+				ReleaseJobName: pj.Spec.Job, // TODO: Is this the correct name?
```
petr-muller, line 141: ad todo; yes

pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
droslean: Yea this most probably will be shared with the other components.

cmd/job-trigger-controller-manager/main.go
droslean: This is the magic moment, where I accidentally add an extra `r` in `logrus` and vscode finds a new library. This is hilarious

droslean: Ah no.. my bad.. We are already using this library. Check the other controllers. See for example: https://github.com/openshift/ci-tools/blob/master/cmd/ci-operator/main.go#L26 or https://github.com/openshift/ci-tools/blob/master/cmd/dptp-controller-manager/main.go#L13

cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,102 @@
...
+	namespace string
+	dryRun    bool
+}
+
+func gatherOptions() (*options, error) {
+	o := &options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
+	fs.StringVar(&o.namespace, "namespace", "", "In which namespace the operation will take place")
```
hongkailiu, line 35: `ci` as default value?

pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer_test.go
```diff
@@ -0,0 +1,104 @@
...
+				t.Fatal(err)
+			}
+		}
+
+		var actualPrpqr v1.PullRequestPayloadQualificationRunList
+		if err := r.client.List(context.Background(), &actualPrpqr); err != nil {
+			t.Fatal(err)
+		}
+
+		if diff := cmp.Diff(actualPrpqr.Items, tc.expected, cmpopts.IgnoreFields(prowv1.ProwJobStatus{}, "StartTime"), cmpopts.IgnoreFields(v1.PullRequestPayloadQualificationRun{}, "TypeMeta.Kind", "TypeMeta.APIVersion", "ResourceVersion")); diff != "" {
```
hongkailiu, line 100: The 2nd ignore looks like
https://github.com/openshift/ci-tools/blob/1a3791f2c57e8d36c940b3de95a0143bb0764e02/pkg/testhelper/testhelper.go#L165-L172

Could we reuse?

pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
```diff
@@ -0,0 +1,122 @@
...
+	client ctrlruntimeclient.Client
+}
+
+func (r *reconciler) Reconcile(ctx context.Context, request controllerruntime.Request) (controllerruntime.Result, error) {
+	log := r.logger.WithField("request", request.String())
+	err := r.reconcile(ctx, log, request)
+	if err != nil {
+		log.WithError(err).Error("Reconciliation failed")
+	}
+	return reconcile.Result{}, err
```
hongkailiu, line 93: Why does it differ from the other controller?

```
return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -0,0 +1,193 @@
...
+			return fmt.Errorf("failed to wait for created ProwJob to appear in cache: %w", err)
+		}
+
+		prpqr.Status.Jobs = append(prpqr.Status.Jobs, v1.PullRequestPayloadJobStatus{
+			ReleaseJobName: pj.Spec.Job,
+			ProwJob:        pj.Name,
+			Status:         pj.Status,
+		})
+
+		logger.Info("Updating PullRequestPayloadQualificationRun...")
```
hongkailiu, line 149: Could we do this via the another controller? (if the another one watch the create events of prowjob)

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -0,0 +1,193 @@
...
+
+	}
+	return nil
+}
+
+// TODO: Currently we create a single dummy prowjob just for testing. The actual implementation
+// will be introduced in https://issues.redhat.com/browse/DPTP-2577
+func generateProwjobs(org, repo, branch, prpqrName, prpqrNamespace string) []prowv1.ProwJob {
+	labels := map[string]string{
+		v1.PullRequestPayloadQualificationRunLabel: prpqrName,
```
hongkailiu, line 162: label value has some restrictions and we have some lib on it.
Should we ensure the restrictions here?

droslean: We could, but this is a constant. I can still write a test to ensure if we change the value, we won't break it.

cmd/job-trigger-controller-manager/main.go
```diff
@@ -0,0 +1,102 @@
...
+		logrus.WithError(err).Fatal("Failed to get options")
+	}
+
+	if err := o.validate(); err != nil {
+		logrus.WithError(err).Fatal("Failed to validate options")
+	}
+
+	ctx := controllerruntime.SetupSignalHandler()
+
+	cfg, err := rest.InClusterConfig()
```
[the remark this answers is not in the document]
hongkailiu: If we use only the InClusterConfig (i think this is the case for the tool), we could just remove `o.kubernetesOptions` completely.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
droslean: The other controller watches only the updates. I could have made it watch the create events as well, and update the `prpqr` but this way we lock the behavior of each controller.

cmd/job-trigger-controller-manager/main.go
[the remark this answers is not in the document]
droslean: Yea, that would be better.

pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer_test.go
droslean: It seems not, since the Comparer is for single objects, and we are checking a list of them.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
hongkailiu: Controller1 watches PullRequestPayloadQualificationRun and creates ProwJobs if needed.
Controller2 watches ProwJobs and updates PullRequestPayloadQualificationRun's status accordingly.

How they get locked? I do not see the infinite loop either.

droslean: I mean that each controller's behavior is related to each action. The first controller is responsible to watch only `create` events and creates the corresponding pj and updates the prpqr status. The second controller is watching only `update` events for prowjobs that hold the label, and then updates the prpqr job statuses.

deads2k:
> [openshift-merge-robot](/openshift-merge-robot) merged commit [`25cae57`](/openshift/ci-tools/commit/25cae57afa9ba6968eb2b546ebc74b0ad2f054ef) into openshift:master [4 hours ago](https://github.com/openshift/ci-tools/pull/2454#event-5569014408)

Yay!  Great to see progress on this front.

PR #2455, opened 2021-11-01
Add unfinished jobs to list when we fail to get prow job, too

stbenjam:


deads2k:
/lgtm

but CI is quite red.

PR #2456, opened 2021-11-01
test injection: respect `releases.latest` in the base

petr-muller:
`releases.latest` serves as an installation or upgrade target, so we must respect its definition in the base in order to correctly build the ephemeral release including the images built in the tested PR

/cc @openshift/test-platform


droslean (approved):

petr-muller:
`the pod ci-op-s1csm96g/validate-vendor was deleted without completing after 2m0s` 
b01 upgrade interference perhaps?

PR #2466, opened 2021-11-03
Allow for configurable shm-size

smg247:
Allowing for the increase in shm-size to facilitate passing Selenium tests for the QE effort.

Following the implementation details from: https://docs.google.com/document/d/1jfgIBqND5DtuzJY9nJG9G9FeHoDqqqqDl8H3ZU2QT7k/edit#heading=h.66y4kqbj468a

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -474,7 +474,7 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			continue
 		}
 		allResources := &resources
-		if !resources.Limits.Name(api.ShmResource, resource.BinarySI).IsZero() {
+		if !resources.Requests.Name(api.ShmResource, resource.BinarySI).IsZero() {
 			// If shm is in Limits it must also be in Requests
 			allResources = resources.DeepCopy()
 			logrus.Info("removing shm from resources for container")
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -13,6 +13,7 @@ import (
 	coreapi "k8s.io/api/core/v1"
 	rbacapi "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -472,6 +473,14 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			errs = append(errs, err)
 			continue
 		}
+		allResources := &resources
+		if !resources.Limits.Name(api.ShmResource, resource.BinarySI).IsZero() {
+			// If shm is in Limits it must also be in Requests
+			allResources = resources.DeepCopy()
+			logrus.Info("removing shm from resources for container")
+			delete(resources.Requests, api.ShmResource)
+			delete(resources.Limits, api.ShmResource)
+		}
 		if step.BestEffort != nil && *step.BestEffort {
 			bestEffort.Insert(name)
 		}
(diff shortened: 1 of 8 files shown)


pkg/validation/config.go
```diff
@@ -460,8 +460,9 @@ func validateResourceList(fieldRoot string, list api.ResourceList) []error {
 	var numInvalid int
 	for key := range list {
 		switch key {
-		case "cpu", "memory":
-			if quantity, err := resource.ParseQuantity(list[key]); err != nil {
+		case "cpu", "memory", api.ShmResource:
```
petr-muller, line 463: Does it makes sense to allow `ShmResource` in `limits`? Maybe it would make sense to validate it is only present in `requests` (or at least that is how I understand our intent).

pkg/steps/multi_stage.go
```diff
@@ -472,6 +473,14 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			errs = append(errs, err)
 			continue
 		}
+		allResources := &resources
+		if !resources.Limits.Name(api.ShmResource, resource.BinarySI).IsZero() {
```
petr-muller, line 477: I'm not sure I entirely follow - I think we should check requests here and not limits?

petr-muller (commented):
All my comments are related to a single thing, maybe my misunderstanding of whether the users are supposed to set `requests` or `limits` (https://docs.google.com/document/d/1jfgIBqND5DtuzJY9nJG9G9FeHoDqqqqDl8H3ZU2QT7k/edit#heading=h.3hjicn8midx5 says `requests` but code seems like both is allowed and prefers `limits`)

pkg/validation/config.go
smg247: There is a k8s validation that makes us set the `Limit` as well on these custom resources. I wasn't aware of that when I wrote the doc up.

`could not run steps: step shm-increase failed: "shm-increase" test steps failed: failed to create or restart shm-increase-step-with-increased-shm pod: unable to create pod: Pod "shm-increase-step-with-increased-shm" is invalid: spec.containers[0].resources.limits: Required value: Limit must be set for non overcommitable resources`

In practice we should just have the config list the `Request` and the `Limit` the same.

pkg/steps/multi_stage.go
smg247: Since we have to set them both to pass k8s validations we could check for the presence of either here, but I agree that checking `Requests` makes more sense and will update that.

pkg/steps/testdata/zz_fixture_TestGeneratePods.yaml
```diff
@@ -78,7 +78,9 @@
         value: /var/run/secrets/ci.openshift.io/multi-stage
       image: pipeline:src
       name: test
-      resources: {}
+      resources:
+        requests:
+          ci-operator.openshift.io/shm: 2G
```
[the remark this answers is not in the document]
smg247: ah I see what happened...its due to checking `Limits` as opposed to `Requests`. I am updating that and it will be pruned now.

pkg/steps/multi_stage.go
```diff
@@ -472,6 +473,14 @@ func (s *multiStageTestStep) generatePods(steps []api.LiteralTestStep, env []cor
 			errs = append(errs, err)
 			continue
 		}
+		allResources := &resources
+		if !resources.Requests.Name(api.ShmResource, resource.BinarySI).IsZero() {
+			// If shm is in Limits it must also be in Requests
+			allResources = resources.DeepCopy()
+			logrus.Info("removing shm from resources for container")
+			delete(resources.Requests, api.ShmResource)
+			delete(resources.Limits, api.ShmResource)
```
hongkailiu, line 482: Why do we delete them here? Do we want this on the container?

```
      resources:
        limits:
          ci-operator.openshift.io/shm: 2G
        requests:
          ci-operator.openshift.io/shm: 2G
```

smg247: No, we can't have it in the container unless we actually "advertise" it as an extended resource. See this slack thread for a deeper explanation: https://coreos.slack.com/archives/GB7NB0CUC/p1635958978160200.

PR #2468, opened 2021-11-04
Collect additional info for release tables

stbenjam:
This adds support for multi-arch (s390x and arm64), as well as
collecting the X.Y release in addition to the tag (for easier
use with Sippy).

Usage:

```
$ job-run-aggregator upload-releases --bigquery-dataset ci_data --google-service-account-credential-file <creds.json> --releases 4.10.0-0.nightly --releases 4.10.0-0.ci --releases 4.9.0-0.nightly --releases 4.9.0-0.ci --releases 4.8.0-0.nightly --releases 4.8.0-0.ci --architectures amd64 --architectures arm64 --architectures s390x
```

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
@@ -8,6 +8,12 @@ type ReleaseRow struct {
 	// the release after it's "fully baked."
 	Phase string `bigquery:"phase"`
 
+	// Release contains the X.Y version of the payload, e.g. 4.8
+	Release string `bigquery:"release"`
+
+	// Architecture contains the architecture for the release, e.g. amd64
+	Architecture string `bigquery:"architecture"`
+
 	// ReleaseTag contains the release version, e.g. 4.8.0-0.nightly-2021-10-28-013428.
 	ReleaseTag string `bigquery:"releaseTag"`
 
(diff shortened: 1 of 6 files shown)


pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
```diff
@@ -8,6 +8,12 @@ type ReleaseRow struct {
 	// the release after it's "fully baked."
 	Phase string `bigquery:"phase"`
 
+	// Release contains the X.Y version of the payload, e.g. 4.8
+	Release string `bigquery:"release"`
+
+	// Architecture contains the architecture for the release, e.g. amd64
+	Architecture string `bigquery:"architecture"`
+
 	// ReleaseTag contains the release version, e.g. 4.8.0-0.nightly-2021-10-28-013428.
```
deads2k, line 17: does the release tag change for each architecture or is the logical key of this table changing to releasetag and architecture?

stbenjam: The release tag includes the arch, except for amd64. e.g., 4.10.0-0.nightly-arm64-2021-11-04-001554

So there'd no collision if the timestamp happened to be the same (but in practice isn't the case, there isn't an 11-04-001554 for each arch).

deads2k: Oh, I see.  I'm only watching part of payloads today.  I didn't even think about that.

PR #2471, opened 2021-11-04
move disruption job runs to its own table

deads2k:
This allows us to treat the jobrun entry as authoritative during upload so that we avoid duplicate rows for job runs and become much more efficient.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go
@@ -1,0 +1,65 @@
+package jobrunaggregatorapi
+
+import "time"
+
+const (
+	LegacyJobRunTableName     = "JobRuns"
+	DisruptionJobRunTableName = "BackendDisruption_JobRuns"
+
+	jobRunSchema = `
+[
+  {
+    "name": "Name",
+    "description": "name of the jobrun (the long number)",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "JobName",
+    "description": "name of the job from CI",
+    "type": "STRING",
+    "mode": "REQUIRED"
+  },
+  {
+    "name": "Status",
+    "description": "error, failure, success",
+    "type": "STRING",
+    "mode": "NULLABLE"
+  },
+  {
+    "name": "StartTime",
+    "description": "time the jobrun started",
+    "type": "TIMESTAMP",
+    "mode": "NULLABLE"
+  },
+  {
+    "name": "EndTime",
+    "description": "time the jobrun started",
+    "type": "TIMESTAMP",
+    "mode": "NULLABLE"
+  },
+  {
+    "name": "ReleaseTag",
+    "description": "",
+    "type": "STRING",
+    "mode": "NULLABLE"
+  },
+  {
+    "name": "Cluster",
+    "description": "the build farm cluster that the CI job ran on: build01, build02, build03, vsphere, etc",
+    "type": "STRING",
+    "mode": "NULLABLE"
+  }
+]
+`
+)
+
+type JobRunRow struct {
+	Name       string
+	JobName    string
+	Status     string
+	StartTime  time.Time
+	EndTime    time.Time
+	ReleaseTag string
+	Cluster    string
+}
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
@@ -14,38 +14,66 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 )
 
-type CIDataClient interface {
-	ListAllJobs(ctx context.Context) ([]jobrunaggregatorapi.JobRow, error)
-
+// AggregationJobClient client view used by the aggregation job
+type AggregationJobClient interface {
 	// GetJobRunForJobNameBeforeTime returns the jobRun closest to, but BEFORE, the time provided.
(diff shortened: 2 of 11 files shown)


pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go
```diff
@@ -0,0 +1,65 @@
+package jobrunaggregatorapi
+
+import "time"
+
+const (
+	LegacyJobRunTableName     = "JobRuns"
+	DisruptionJobRunTableName = "BackendDisruption_JobRuns"
+
+	jobRunSchema = `
```
deads2k, line 9: this matches our existing schema

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go
```diff
@@ -0,0 +1,65 @@
...
+    "name": "Cluster",
+    "description": "the build farm cluster that the CI job ran on: build01, build02, build03, vsphere, etc",
+    "type": "STRING",
+    "mode": "NULLABLE"
+  }
+]
+`
+)
+
+type JobRunRow struct {
```
deads2k, line 57: straight move

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -14,38 +14,66 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 )
 
-type CIDataClient interface {
```
deads2k, line 17: separated this client into the bits used by specific pieces so that I could figure out how broad the scope of change was for various jobrun tables.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -112,11 +140,13 @@ LIMIT 1
 func (c *ciDataClient) GetLastJobRunWithDisruptionDataForJobName(ctx context.Context, jobName string) (*jobrunaggregatorapi.JobRunRow, error) {
 	// the JobRun.Name is always increasing, so we can sort by that name.  The starttime is based on the prowjob
 	// time and I don't think that is coordinated.
+	// the disruption jobrun table is now distinct, so we will use the disruption jobrun table as authoritative for what data should and should not
```
deads2k, line 143: This created a huge efficiency gain.  Apparently we were trying to gather backend disruption data for many job types that don't have it, so we effectively scanned the entire GCS bucket over and over.

pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
```diff
@@ -29,6 +32,8 @@ func NewBigQueryTestRunUploadFlags() *BigQueryTestRunUploadFlags {
 func (f *BigQueryTestRunUploadFlags) BindFlags(fs *pflag.FlagSet) {
 	f.DataCoordinates.BindFlags(fs)
 	f.Authentication.BindFlags(fs)
+
+	fs.BoolVar(&f.DryRun, "dry-run", f.DryRun, "Run the command, but don't mutate data.")
```
deads2k, line 36: I added this so that I could have confidence in running this command locally to test.

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
```diff
@@ -15,34 +20,55 @@ type BigQueryInserter interface {
...
+func NewDryRunInserter(out io.Writer, table string) BigQueryInserter {
+	return dryRunInserter{
+		table: table,
+		out:   out,
 	}
-
 }
 
-var _ bigquery.ValueSaver = &jobRunRow{}
+func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
```
deads2k, line 35: it won't win any awards, but it prints what you're trying to insert.

deads2k:
3x speed improvement *and* for the first time in a while we didn't hit a ratelimit problem.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go
```diff
@@ -0,0 +1,65 @@
+package jobrunaggregatorapi
+
+import "time"
+
+const (
+	LegacyJobRunTableName     = "JobRuns"
```
dgoodwin, line 6: Am I following that LegacyJobRunTableName contains all job runs, and DisruptionJobRunTableName contains only those with disruption data?

If so, given this is a permanent table we are not trying to phase out, it doesn't feel like it's Legacy, rather AllJobRunsTableName or equivalent.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go
```diff
@@ -0,0 +1,65 @@
...
+	DisruptionJobRunTableName = "BackendDisruption_JobRuns"
```
dgoodwin, line 7: Please add a comment on both of these table names explaining what goes into each.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -112,11 +140,13 @@ LIMIT 1
...
+	// be uploaded rather than using the absence of backenddisruption itself.  Some jobs don't include this data so we end up
+	// inserting many duplicated entries
```
dgoodwin, line 145: Can you clarify the last sentence. "Some jobs do not include disruption data, those that do will be duplicated in both the backend disruption job run table, as well as the main job run table."?

PR #2475, opened 2021-11-08
generate prowjobs for each of the prpqr realease job spec items and improve tests

droslean:

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -27,6 +27,7 @@ import (
 	v1 "github.com/openshift/ci-tools/pkg/api/pullrequestpayloadqualification/v1"
 	"github.com/openshift/ci-tools/pkg/controller/prpqr_reconciler/pjstatussyncer"
 	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
+	"github.com/openshift/ci-tools/pkg/jobconfig"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -154,37 +154,42 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
-		JobBase: base,
-		Cron:    "@yearly",
-	}
+		periodicJob := prowconfig.Periodic{
+			JobBase: base,
+			Cron:    "@yearly",
+		}
+
+		pj := pjutil.NewProwJob(pjutil.PeriodicSpec(periodicJob), labels, nil)
+		pj.Name = fmt.Sprintf("dummy-pj-for-%s", prpqrName)
```
petr-muller, line 189: This does not look right - this could result in creating multiple ProwJobs with the same name. I think using the UUIDs (just like for normal ProwJobs is enough here) should be fine here.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -154,37 +154,42 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 
 // TODO: Currently we create a single dummy prowjob just for testing. The actual implementation
 // will be introduced in https://issues.redhat.com/browse/DPTP-2577
-func generateProwjobs(org, repo, branch, prpqrName, prpqrNamespace string) []prowv1.ProwJob {
-	labels := map[string]string{
-		v1.PullRequestPayloadQualificationRunLabel: prpqrName,
-	}
+func generateProwjobs(org, repo, branch, prpqrName, prpqrNamespace string, releaseJobSpec []v1.ReleaseJobSpec) []prowv1.ProwJob {
+	var ret []prowv1.ProwJob
+	for _, releaseJob := range releaseJobSpec {
```
petr-muller, line 159: So one thing we need to record somewhere is the "immitated job name", which is inferred from the spec with `JobName`: 

https://github.com/openshift/ci-tools/blob/ac7aa8f84ac5086728cf8318d0aa58cae2841977/pkg/api/metadata.go#L66-L68

This value then needs to go to the `Status.Jobs.ReleaseJobName` field of the status. I think it would be appropriate if this method returned a `map[string]Prowv1.ProwJob` instead, and this name would be the key. I think it could also be sensible to put that name into a label.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -154,37 +154,40 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+		}
+
+		pj := pjutil.NewProwJob(pjutil.PeriodicSpec(periodicJob), labels, nil)
+		pj.Namespace = prpqrNamespace
 
-	pj := pjutil.NewProwJob(pjutil.PeriodicSpec(periodicJob), labels, nil)
-	pj.Name = fmt.Sprintf("dummy-pj-for-%s", prpqrName)
-	pj.Namespace = prpqrNamespace
-	return []prowv1.ProwJob{pj}
+		ret[releaseJob.Test] = pj
```
petr-muller, line 190: Not just test, but `releaseJob.JobName(PeriodicPrefix, releaseJob.Test)`

droslean: Thanks! Addressed.

petr-muller (approved):
LGTM

/cc @bbguimaraes 
Bruno could you take a look if this meets your expecations?

petr-muller:
oops I wanted to hold :grimacing:

PR #2476, opened 2021-11-08
payload-testing-prow-plugin: only collaborator can trigger runs

hongkailiu:
https://issues.redhat.com/browse/DPTP-2606

/cc @openshift/test-platform 

/hold
need prepare the arg in the deployment.

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -14,6 +14,7 @@ import (
 	"k8s.io/test-infra/prow/github"
 	"k8s.io/test-infra/prow/kube"
 	"k8s.io/test-infra/prow/pluginhelp"
+	"k8s.io/test-infra/prow/plugins/trigger"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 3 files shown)


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -44,13 +46,32 @@ func helpProvider(_ []prowconfig.OrgRepo) (*pluginhelp.PluginHelp, error) {
...
+	trustedPullRequest(author, org, repo string, num int) (bool, error)
+}
+
+type githubTrustedChecker struct {
+	githubClient github.Client
+	pluginAgent  *plugins.ConfigAgent
+}
+
+func (c *githubTrustedChecker) trustedPullRequest(author, org, repo string, num int) (bool, error) {
+	_, trusted, err := trigger.TrustedPullRequest(
```
petr-muller, line 59: Hmm... Why this and not `TrudstedUser`:

https://github.com/kubernetes/test-infra/blob/ee1e7c845bb25177944bb5de37545f007f529059/prow/plugins/trigger/trigger.go#L221-L223

IMO we just need to validate who does the `/payload`, we do not care about the actual PR.

hongkailiu: Updated.

cmd/payload-testing-prow-plugin/server.go
```diff
@@ -44,13 +46,32 @@ func helpProvider(_ []prowconfig.OrgRepo) (*pluginhelp.PluginHelp, error) {
...
+type trustedChecker interface {
+	trustedUser(author, org, repo string, num int) (bool, error)
+}
+
+type githubTrustedChecker struct {
+	githubClient github.Client
+	pluginAgent  *plugins.ConfigAgent
+}
+
+func (c *githubTrustedChecker) trustedUser(author, org, repo string, num int) (bool, error) {
```
droslean, line 58: quick Q: Why do we need to use trigger config for that? Also, do we even have a configuration for the trigger plugin?

hongkailiu: The trigger's config is used in the function.
https://github.com/kubernetes/test-infra/blob/ee1e7c845bb25177944bb5de37545f007f529059/prow/plugins/trigger/trigger.go#L221-L223

IIUC, it allows more than collaborators.

We do have config for trigger:
https://github.com/kubernetes/test-infra/blob/0d86dc3aed9603385d3e91ad76f367cfce5dd513/prow/plugins/config.go#L90

droslean: About the config, I mean that we don't have any `triggers` configuration in our prow config.

droslean: By adding a `trigers` configuration to the plugin config, is there any chance that it will affect anything else?

petr-muller: I think Nikos has a great point here - it looks like we are only using two fields from `trigger` config, but our config does not actually utilize them at all, anywhere. So we should be fine with using the basic org/repo collaborator check you get when `onlyOrgMembers=false` and `trustedOrg=""`, and that way we do not need `trigger`s config at all.

hongkailiu: I see.
Updating...

hongkailiu: Done.

petr-muller (approved):

PR #2479, opened 2021-11-09
Collect some additional data in our bigquery release tables

stbenjam:
- Add more comments
- Collect stream ("nightly", or "ci"), and timestamp for a release tag
- Remove bigquery types all together and use native go types for time and int

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
@@ -1,6 +1,8 @@
 package jobrunaggregatorapi
 
-import "cloud.google.com/go/bigquery"
+import (
+	"time"
+)
 
 type ReleaseRow struct {
 	// Phase contains the overall status of a payload: e.g. Ready, Accepted,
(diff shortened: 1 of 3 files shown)


pkg/jobrunaggregator/releasebigqueryloader/release_uploader.go
```diff
@@ -206,6 +207,23 @@ func releaseDetailsToBigQuery(architecture string, tag ReleaseTag, details Relea
 		release.Release = strings.Join(parts[:2], ".")
 	}
 
+	// Get "nightly" or "ci" from the string
```
deads2k, line 210: I don't know when and I don't know how, but this will break someday.  Try to remember it's optional in your consuming code.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
```diff
@@ -80,23 +90,42 @@ type ReleasePullRequestRow struct {
...
+
+	// State holds the overall status of the job, such as Failed.
+	State string `bigquery:"state"`
+
+	// URL contains a link to Prow.
+	URL string `bigquery:"url"`
+
+	// Transition time contains the transition time from the release
+	// controller.
+	TransitionTime time.Time `bigquery:"transitionTime"`
```
deads2k, line 115: yay!

PR #2484, opened 2021-11-10
Close the blocker issue when there is no blocked branch

jmguzik:
Related to the changes done in https://github.com/openshift/release/pull/23503

### DIFF
--- a/cmd/blocking-issue-creator/main.go
+++ b/cmd/blocking-issue-creator/main.go
@@ -46,7 +46,7 @@ func gatherOptions() options {
 	o := options{}
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 
-	fs.BoolVar(&o.dryRun, "dry-run", false, "Dry run for testing. Uses API tokens but does not mutate.")
+	fs.BoolVar(&o.dryRun, "dry-run", true, "Dry run for testing. Uses API tokens but does not mutate.")
 
 	o.github.AddFlags(fs)
 	o.FutureOptions.Bind(fs)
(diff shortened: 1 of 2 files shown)


cmd/blocking-issue-creator/main.go
```diff
@@ -136,6 +132,17 @@ func manageIssues(client githubClient, githubLogin string, repoInfo *config.Info
 		return err
 	}
 
+	if len(branches) == 0 {
```
petr-muller, line 135: I think it would be useful to have a info-level log here about the decision, something like "Repository does not have any blocked branches, deleting blocker issues", maybe conditionally on `len(issues) > 0`

cmd/blocking-issue-creator/main.go
```diff
@@ -136,6 +132,17 @@ func manageIssues(client githubClient, githubLogin string, repoInfo *config.Info
...
+		for _, issue := range issues {
+			if err := client.CloseIssue(repoInfo.Org, repoInfo.Repo, issue.Number); err != nil {
+				logger.WithError(err).Error("Failed to close issue.")
+				return err
+			}
+			logger.WithField("number", issue.Number).Info("Closed issue.")
+		}
```
petr-muller, line 142: suggestion: maybe extract this into a `closeIssues(issues []Issue, repoInfo, logger) error` and use it both here and in the `len(issues) > 1` block below.

Could be even an anonymous funciton inline to simplify the signature :

```go
closeIssues := func(issues []Issue) error {
}
```

jmguzik: Actually, I thought about this redundancy for a little while and came to the conclusion it does not make much sense in this situation, but I can adjust this.

jmguzik: My rationale was: logs are different and we are acting on different sets: issues from 0 vs issues from 1

petr-muller: Yeah I thought it would take a slice as an input, and then it could be called as `closeIssues(issues)` from one location and `closeIssues(issues[1:])` from the other. But not a big deal.

petr-muller (approved):

PR #2485, opened 2021-11-10
payload-testing-prow-plugin: add some debugging logs

hongkailiu:
/cc @openshift/test-platform

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -6,6 +6,7 @@ import (
 	"regexp"
 	"strconv"
 	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
@@ -116,6 +117,7 @@ func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 }
 
 func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
+	start := time.Now()
 	org := ic.Repo.Owner.Login
 	repo := ic.Repo.Name
 	prNumber := ic.Issue.Number
@@ -126,6 +128,7 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 		github.OrgLogField:  org,
 		github.RepoLogField: repo,
 		github.PrLogField:   prNumber,
+		github.EventGUID:    guid,
 	})
 
 	// only reacts on comments on PRs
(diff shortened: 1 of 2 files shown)


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -176,11 +184,16 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 		builder.spec = spec
 		var jobNames []string
 		var jobTuples []api.MetadataWithTest
+		specLogger.Debug("resolving jobs ...")
+		startResolveJobs := time.Now()
 		jobs, err := s.jobResolver.resolve(spec.ocp, spec.releaseType, spec.jobs)
+		specLogger.WithField("duration", time.Since(startResolveJobs)).Debug("resolving jobs completed")
```
petr-muller, line 190: Could we log how many jobs do we have received?

cmd/payload-testing-prow-plugin/server.go
```diff
@@ -199,16 +212,23 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 				jobTuples = append(jobTuples, jobTuple)
 			}
 		}
+		specLogger.WithField("duration", time.Since(startResolveTests)).Debug("resolving tests completed")
```
petr-muller, line 215: Could we log how many jobs did we manage to resolve?

cmd/payload-testing-prow-plugin/server.go
```diff
@@ -199,16 +212,23 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
...
 				specLogger.WithError(err).Error("could not create PullRequestPayloadQualificationRun")
 				return fmt.Sprintf("could not create PullRequestPayloadQualificationRun: %v", err)
 			}
+			specLogger.WithField("duration", time.Since(startCreateRuns)).WithField("runNames", runNames).Debug("creating PullRequestPayloadQualificationRuns completed")
 		} else {
 			specLogger.Warn("found no resolved tests")
 		}
 		messages = append(messages, message(spec, jobNames))
 	}
+	logger.WithField("duration", time.Since(start)).Debug("handle completed")
```
petr-muller, line 231: Would it make sense to publish all these durations as metrics?

hongkailiu: I think yes if we have more profiling to come.
In that case, we just need to switch those duration to a histogram and expose it to prow-monitoring.

cmd/payload-testing-prow-plugin/server.go
hongkailiu: Done.

cmd/payload-testing-prow-plugin/server.go
hongkailiu: Done.

petr-muller (approved):

PR #2486, opened 2021-11-10
fix a race when trying to update the prpqr object

droslean:

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -11,6 +11,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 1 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -138,6 +138,12 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 			return fmt.Errorf("failed to wait for created ProwJob to appear in cache: %w", err)
 		}
 
+		// Before updating, re-get the object in case it was modified
+		prpqr := &v1.PullRequestPayloadQualificationRun{}
+		if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: req.Namespace, Name: req.Name}, prpqr); err != nil {
+			return fmt.Errorf("failed to get the PullRequestPayloadQualificationRun: %s in namespace %s: %w", req.Name, req.Namespace, err)
+		}
+
```
petr-muller, line 146: This will not fix the race,there is still a window between the get and the update. In this case, I think we need to act on the conflict error after update, perhaps in a backoff. 
- Try to update
- Check for errors, if conflict, get and retry
- If errr but not a conflict, a real error
- If too many retries, a real error

petr-muller (approved):

PR #2487, opened 2021-11-10
prevent from creating prowjob when it is already exists based on labels

droslean:

Before, the code was checking if a prowjob exists based on its name. This is incorrect because the `pj.Name` will always be different. 

This PR adds a label, when we create a `prowjob`, with information about the release job name. Example:
```yaml
pullrequestpayloadqualificationruns.ci.openshift.io: prpqr-test,
releaseJobName: periodic-ci-test-org-test-repo-test-branch-test-name
``` 

With the `releaseJobName` label we can make sure that a `prowjob` exists for a specific release job for a specific `prpqr`

Also test cases added.

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -29,10 +29,12 @@ import (
 	"github.com/openshift/ci-tools/pkg/controller/prpqr_reconciler/pjstatussyncer"
 	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
 	"github.com/openshift/ci-tools/pkg/jobconfig"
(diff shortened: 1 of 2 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -193,7 +211,7 @@ func generateProwjobs(org, repo, branch, prpqrName, prpqrNamespace string, relea
 		pj := pjutil.NewProwJob(pjutil.PeriodicSpec(periodicJob), labels, nil)
```
hongkailiu, line 211: I guess this is the place which gives us the name differently each time?

hongkailiu:
/lgtm

/hold
whenever you are ready for rolling it out.

hongkailiu (approved):

PR #2492, opened 2021-11-11
payload-testing-ui: check presence of status URL

bbguimaraes:
The URL in the status is filled by `plank` after the creation of the
`ProwJob`, so it may be empty while the test is scheduled.


bbguimaraes:
https://payload-testing-ui-bbguimaraes.apps.ci.l2s4.p1.openshiftapps.com/runs/bbguimaraes/test
vs.
https://payload-testing-ui-bbguimaraes.apps.ci.l2s4.p1.openshiftapps.com/runs/ci/686a2450-4256-11ec-9b44-50db765473cc-0

droslean (approved):

PR #2493, opened 2021-11-11
Include a message with the UI link when commenting back to the PR

droslean:
/cc @openshift/test-platform 

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -23,6 +23,10 @@ import (
 	"github.com/openshift/ci-tools/pkg/release/config"
 )
 
+const (
+	prPayloadTestsUIURL = "https://pr-payload-tests.ci.openshift.org/runs"
+)
+
 type githubClient interface {
 	CreateComment(owner, repo string, number int, comment string) error
 	GetPullRequest(org, repo string, number int) (*github.PullRequest, error)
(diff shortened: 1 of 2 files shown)


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -229,6 +236,9 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 		}
 		messages = append(messages, message(spec, jobNames))
 	}
+
+	messages = append(messages, fmt.Sprintf("See details on %s/%s/%s\n", prPayloadTestsUIURL, builder.namespace, prpqrName))
```
bbguimaraes, line 240: This has to be done for each `spec`.

bbguimaraes (changes requested):

cmd/payload-testing-prow-plugin/server.go
droslean: Why? The prpqr name will be the same. In the link the user can see all the jobs then.

bbguimaraes: `specs` can have multiple items if the comment being processed has more than one `/payload` command.

petr-muller: yes, and `prpqr` names will be different (they must be, they are kube resources), the builder is adding counter-based suffixes `-0` to the names

droslean: Yea got it! Updated.

I left the `trigger 0 jobs ` message untouched for now

PR #2497, opened 2021-11-12
read backend disruption for aggregator from the raw data instead of the junit

deads2k:
This lets us add disruption tests in origin and have them reported without change the code in the aggregator.

I left the old code in to gather data from pre-4.10 releases until we ship 4.10.  Then we can probably remove them

### DIFF
--- a/pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
+++ b/pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
@@ -2,12 +2,14 @@ package jobrunbigqueryloader
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"math"
 	"regexp"
 	"strings"
 	"time"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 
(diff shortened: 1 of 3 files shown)


pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
```diff
@@ -151,17 +192,56 @@ func newDisruptionUploader(backendDisruptionInserter jobrunaggregatorlib.BigQuer
...
+	releaseHasDisruptionData := strings.Contains(jobRun.GetJobName(), "4.10") ||
+		strings.Contains(jobRun.GetJobName(), "4.11") ||
+		strings.Contains(jobRun.GetJobName(), "4.12") ||
+		strings.Contains(jobRun.GetJobName(), "4.13") ||
+		strings.Contains(jobRun.GetJobName(), "4.14") ||
+		strings.Contains(jobRun.GetJobName(), "4.15") ||
+		strings.Contains(jobRun.GetJobName(), "4.16") ||
+		strings.Contains(jobRun.GetJobName(), "4.17") ||
+		strings.Contains(jobRun.GetJobName(), "4.17") ||
+		strings.Contains(jobRun.GetJobName(), "4.19")
```
stbenjam, line 217: :eyes:

deads2k: Yeah, it's not all beautiful.

stbenjam:
Looks good to me.  The static list of versions is going to ruin some future TRT persons day, though.

/lgtm

PR #2500, opened 2021-11-16
Change http client to http client with retry options

jmguzik:
Due to rare protocol errors, a standard HTTP client is exchanged with a retriable HTTP client. To keep compatibility, after creating a retriable client, it is converted to the standard client with a custom transport.

An alternative approach to this problem would be to inject a custom transport to the standard HTTP client (which is exactly what a retriable HTTP client does).
Logger is nulled similarly to pkg/github/github.go

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -6,10 +6,10 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"net/http"
 	"strings"
 	"time"
 
+	"github.com/hashicorp/go-retryablehttp"
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


jmguzik:
Not sure how many retries you want. I left default which is 4.

pkg/defaults/defaults.go
```diff
@@ -104,8 +104,10 @@ func FromConfig(
 			return nil, nil, fmt.Errorf("could not get Hive client for Hive kube config: %w", err)
 		}
 	}
+	httpClient := retryablehttp.NewClient()
+	httpClient.Logger = nil
 
-	return fromConfig(ctx, config, graphConf, jobSpec, templates, paramFile, promote, client, buildClient, templateClient, podClient, leaseClient, hiveClient, &http.Client{}, requiredTargets, cloneAuthConfig, pullSecret, pushSecret, api.NewDeferredParameters(nil), censor, consoleHost)
+	return fromConfig(ctx, config, graphConf, jobSpec, templates, paramFile, promote, client, buildClient, templateClient, podClient, leaseClient, hiveClient, httpClient.StandardClient(), requiredTargets, cloneAuthConfig, pullSecret, pushSecret, api.NewDeferredParameters(nil), censor, consoleHost)
```
droslean, line 110: here we still pass an http.Client. Is this expected?

droslean:
Can we also have a test case somewhere which ensures that the client will retry?

pkg/defaults/defaults.go
jmguzik: Yes, it's fine. The client is working the same way as before, it has just modified transport functionality which includes all the benefits of retryable-http.

jmguzik:
> Can we also have a test case somewhere which ensures that the client will retry?

That would need to be a time measurement test and it's very hard to recreate protocol error scenario.

droslean (approved):

PR #2502, opened 2021-11-16
retry on query quota rate exceeded

deads2k:
we see errors like
 
> time="2021-11-16T04:00:44Z" level=fatal msg="Command failed" error="failed to query aggregation table with \"\\nSELECT *\\nFROM openshift-ci-data-analysis.ci_data.BackendDisruption_JobRuns as JobRuns\\nWHERE JobRuns.JobName = @JobName\\nORDER BY JobRuns.Name DESC\\nLIMIT 1\\n\": googleapi: Error 400: Job exceeded rate limits: Your project_and_region exceeded quota for concurrent queries. For more information, see https://cloud.google.com/bigquery/docs/troubleshoot-quotas, jobRateLimitExceeded"

in our log, which cause retries.  This wraps our client to retry on a backoff.


stbenjam:
/hold

Prevent retesting codegen

PR #2508, opened 2021-11-18
Include conditions when reconciling the prpqr resources

droslean:
/cc @openshift/test-platform 

part of https://issues.redhat.com/browse/DPTP-2541

This PR adds the following types of conditions:

1. `AllJobsStarted` when all jobs started without errors
2. `WithErrors` when not all jobs have been created. The errors are included.

/cc @bbguimaraes 

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -9,6 +9,7 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
@@ -35,6 +36,9 @@ import (
 const (
 	controllerName      = "prpqr_reconciler"
 	releaseJobNameLabel = "releaseJobName"
+
+	conditionAllJobsTriggered = "AllJobsTriggered"
+	conditionWithErrors       = "WithErrors"
 )
 
 func AddToManager(mgr manager.Manager, ns string) error {
@@ -103,6 +107,9 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 	logger = logger.WithField("namespace", req.Namespace).WithField("prpqr_name", req.Name)
 	logger.Info("Starting reconciliation")
 
+	var prpqrMutations []func(prpqr *v1.PullRequestPayloadQualificationRun)
+	createdJobs := make(map[string]v1.PullRequestPayloadJobStatus)
+
 	prpqr := &v1.PullRequestPayloadQualificationRun{}
 	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: req.Namespace, Name: req.Name}, prpqr); err != nil {
 		return fmt.Errorf("failed to get the PullRequestPayloadQualificationRun: %s in namespace %s: %w", req.Name, req.Namespace, err)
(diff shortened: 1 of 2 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -103,69 +112,155 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+	// missing then we add the AllJobsFinished condition
+	if prpqr.Status.Jobs != nil && hasAllJobsFinished(prpqr.Status.Jobs, prowjobs) {
+		prpqrMutations = append(prpqrMutations, func(prpqr *v1.PullRequestPayloadQualificationRun) {
+			prpqr.Status.Conditions = append(prpqr.Status.Conditions, metav1.Condition{
+				Status:             metav1.ConditionTrue,
+				LastTransitionTime: metav1.Time{Time: time.Now()},
+				Reason:             conditionAllJobsFinished,
+				Message:            "All jobs are finished.",
+			})
+		})
```
petr-muller, line 135: I don't understand why we expect this condition to happen here. This is the PRPQR reconciler which acts on new PRPQRs being created, and it reconciles by spawning jobs. So I would expect this controller to only ever set the `AllJobsStarted` condition (with or without error).

IMO the other controller (which watches jobs) is the only part that can reasonably maintain the "All Jobs Finished" condition, because is the only controller that knows that fact.

petr-muller (commented):
A comment inline

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
droslean: I will make this controller to watch also the updates (which I forgot to do). The other controller watches specific jobs and updates the `prpqr` status once the job is finished. If we add to also enumerate the jobs and apply the condition we can end up in an infinitive loop, if the first controller will fail to create one of the jobs. That is why, with these changes, I make the `prpqr_reconciler` controller more intelligent by managing also the failed jobs in a second reconciliation. Also, it makes sure to add an extra condition if we fail to create anything.

bbguimaraes: Under what conditions do we reconcile the same CR here to retry failed job creations?

(How) do we distinguish temporary from permanent failures so we don't retry indefinitely?

droslean: > Why would there be a loop if this controller only processes creation events?

I changed the controller to reconcile also in update events as well. 

> Under what conditions do we reconcile the same CR here to retry failed job creations

We should reconcile the object on every update in order to apply a condition and to create the missing jobs as well.

> (How) do we distinguish temporary from permanent failures so we don't retry indefinitely?

Currently, we don't distinguish temporary from permanent failures. The only failures that we manage with the conditions are the prowjob creating failure. It is really hard to have a permanent failure there. Any ideas?

bbguimaraes: How will it make progress if the CR is never updated (it fails to create all jobs, jobs are not scheduled or not updated for whatever reason, etc.)?

We just had a permanent failure where a job had a label that was too long.  A situation like this would make the controller try to create it endlessly.

petr-muller: > I changed the controller to reconcile also in update events as well.

That IMO does not make sense. The job of one controller is to spawn jobs, and it should therefore control the "AllJobsTriggered" condition. The job of the other one is to watch jobs to completion, and therefore it should own the `AllJobsFinished` condition. I think separating concerns between the two controllers is important - each of them should have a clear responsibility and ownership, and they should not mix, otherwise things become confusing.

Maybe it would make sense to break this PR into something smaller, looks like we are discussing too many problems at once. How about make this PR handle just the `AllJobsTriggered` condition (and leave the "all jobs finished" for another PR), and while doing that, let's solve the problem of handling failures. 

We definitely need to handle permanent failures on creating jobs (the integration PR brings more failure modes there). I would suggest to go with treating all failures as permanent right now. I think we will need to invent some data structure to record this case in the status.

So I would suggest the following:

- The controller reacts on create
- Attempts to spawn all jobs
- For jobs that fail to spawn (for whatever reason), updates their status to record that it failed (and won't be retried)

Which sounds to me that we will need two conditions for the "Triggered" idea:
- `FullyProcessed` (the controller processed the PRPQR, attempted to trigger all jobs, and they either all triggered, or some of them failed to trigger and the failure is permanent). This means that no further attempts will be made. Basically the triggering controller is fully done with this PRPQR.
- `AllJobsSuccessfullyTriggered` is a subset of `FullyProcessed` and means that all jobs were successfully triggered, there were no failures.

If we keep things simple, we start with all failures as permanent: just try once, and if it fails, record the fact and give up. We can then build retries afterwards.

droslean: OK, I will try to make this PR simpler and also treat the errors as permanent. One concern only is that for the "failed to create jobs" we will need to customize the Prowjob's status. Something like:
```
jobs:
- jobName: periodic-blalabla
  prowJob: ""
  status: 
    state: Error
    description: "failed to create prowjob error"
```

petr-muller: Yeah we'll need to represent the permanent failure anyway, the actual integration will have them. Enhancing status is not a problem - just make it generic enough so that we can communicate why it failed (your example seems fine!)

petr-muller (approved):

PR #2511, opened 2021-11-18
Upload alerts

deads2k:
Scans the alert durations we include from openshift-tests and writes them into bigquery for later aggregation.


pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
```diff
@@ -123,114 +123,6 @@ func (f *BigQueryTestRunUploadFlags) ToOptions(ctx context.Context) (*allJobsLoa
 	}, nil
 }
 
-type BigQueryDisruptionUploadFlags struct {
```
deads2k, line 126: this moved to disruption.go

PR #2512, opened 2021-11-19
dptp-cm: fix forbidden registry handling

petr-muller:
`Union` returns a new set, it does not change the input sets. Therefore
the `registriesExceptAppCI` did not really propagate anywhere.

Noticed while looking into [DPTP-2626](https://issues.redhat.com/browse/DPTP-2626)
@hongkailiu , it looks like we were running *without* the forbidden registries for few days. What is the impact?


hongkailiu (approved):
Thanks. @petr-muller

PR #2515, opened 2021-11-22
Aggregated disruption

deads2k:
tweak on https://github.com/openshift/ci-tools/pull/2417 to build the suite independent of junit results so we can start having the disruption tests on individual job runs only fail on passing the p95.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
@@ -6,9 +6,9 @@ import (
 	"os"
 
 	"gopkg.in/yaml.v2"
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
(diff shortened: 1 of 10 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
```diff
@@ -0,0 +1,174 @@
...
+				currDetails.Failures = append(currDetails.Failures, TestCaseFailure{
+					JobRunID:       jobRunID,
+					HumanURL:       humanURL,
+					GCSArtifactURL: gcsArtifactURL,
+				})
+				currDetails.Passes = append(currDetails.Passes, TestCasePass{
+					JobRunID:       jobRunID,
+					HumanURL:       humanURL,
+					GCSArtifactURL: gcsArtifactURL,
+				})
```
stbenjam, line 92: I'm a little confused by this -- having a pass and a failure makes a flake? Why not just create a TestCaseFlake?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
```diff
@@ -0,0 +1,174 @@
...
+
+	default:
+		// ignore the errors if we have at least three results
+		fmt.Fprintf(os.Stderr, "Could not fetch backend disruption data for all runs %v\n", err)
+	}
+
+	allBackends := getAllDisruptionBackendNames(jobRunIDToBackendNameToAvailabilityResult)
+	for _, backendName := range allBackends.List() {
+		jobRunIDToAvailabilityResultForBackend := getDisruptionForBackend(jobRunIDToBackendNameToAvailabilityResult, backendName)
+		historicalStats, failed, message, err := o.passFailCalculator.CheckDisruption(ctx, jobRunIDToAvailabilityResultForBackend, backendName)
```
stbenjam, line 54: This shouldn't be called CheckDisruption anymore

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
deads2k: > I'm a little confused by this -- having a pass and a failure makes a flake? Why not just create a TestCaseFlake?

This matches the content that we consume from the jobs we're aggregating today, where a pass and a fail indicate a flake that we've allowed to pass.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
```diff
@@ -0,0 +1,174 @@
...
+			overP95 := float64(currAvailabilityStat.SecondsUnavailable) > historicalStats.p95
+			switch {
+			case overP95:
+				currDetails.Failures = append(currDetails.Failures, TestCaseFailure{
+					JobRunID:       jobRunID,
+					HumanURL:       humanURL,
+					GCSArtifactURL: gcsArtifactURL,
+				})
+
+			case overMean: // this will mark it as a flake in higher layers
```
stbenjam, line 82: Only time we'll fail is if the mean for the job run is over the historical p95?

I was marking any backend that was more than 110% of the mean, excluding 1 outlier, as a failure.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
stbenjam: Maybe I'm wrong, I'm a little confused what's going on here compared to the for loop below. I need to look with fresh eyes tomorrow

deads2k: > Maybe I'm wrong, I'm a little confused what's going on here compared to the for loop below. I need to look with fresh eyes tomorrow

I wanted to keep the diff small. I'm open to renaming this or even moving it if you like.  The fail, message, and err from this method are honored.  the historicalstats allow synthetic junitest creation.

deads2k:
merged in original PR

PR #2516, opened 2021-11-23
ci-secret-bootstrap: --secret-names

hongkailiu:
https://issues.redhat.com/browse/DPTP-2585

/cc @openshift/test-platform

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -47,12 +47,12 @@ type options struct {
 	force              bool
 	validateItemsUsage bool
 	confirm            bool
-	configUpdaterOnly  bool
 
 	kubernetesOptions   flagutil.KubernetesOptions
 	configPath          string
 	generatorConfigPath string
 	cluster             string
+	secretNamesRaw      flagutil.Strings
 	logLevel            string
 	impersonateUser     string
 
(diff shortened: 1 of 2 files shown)


smg247:
Do you think there would be any future value to making this argument more generic? Something like, `secretsNamed` and then pass a value (in this case `config-updater`). That way we could use this for other cases of only working with specifically named secrets.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -120,6 +122,12 @@ func (o *options) completeOptions(censor *secrets.DynamicCensor, kubeConfigs map
 		return err
 	}
 
+	if o.configUpdaterOnly {
+		logrus.WithField("o.configUpdaterOnly", o.configUpdaterOnly).Info("Removing secrets whose name is not config-updater ...")
```
smg247, line 126: The field on this log line will only ever be true. I think a regular old `logrus.Info` would suffice here.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -120,6 +122,12 @@ func (o *options) completeOptions(censor *secrets.DynamicCensor, kubeConfigs map
...
+		configUpdaterOnly(&o.config)
+		logrus.WithField("o.config.Secrets", o.config.Secrets).WithField("o.configUpdaterOnly", o.configUpdaterOnly).Info("Removed secrets whose name is not config-updater")
```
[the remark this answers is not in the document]

cmd/ci-secret-bootstrap/main.go
```diff
@@ -120,6 +122,13 @@ func (o *options) completeOptions(censor *secrets.DynamicCensor, kubeConfigs map
 		return err
 	}
 
+	if vals := o.secretNamesRaw.Strings(); len(vals) > 0 {
+		secretNames := sets.NewString(vals...)
+		logrus.Info("Removing secrets whose name is not config-updater ...")
+		configUpdaterOnly(&o.config, secretNames)
```
smg247, line 128: This method and preceeding log call should also be genericized

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -2830,3 +2830,59 @@ func TestIntegration(t *testing.T) {
...
+					},
+				},
+			},
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			configUpdaterOnly(tc.given, sets.NewString("config-updater"))
+			if diff := cmp.Diff(tc.given, tc.expected); diff != "" {
+				t.Errorf("%s: actual differs from expected after casin update e: %s", tc.name, diff)
```
smg247, line 2884: "casin"? typo?

hongkailiu: Leftover from copying.
Removed.

cmd/ci-secret-bootstrap/main.go

smg247 (approved):

PR #2524, opened 2021-11-24
Apply a condition when all jobs are finished in the pjstatussyncer controller

droslean:
/cc @openshift/test-platform 

This change adds the `AllJobsFinished` condition if all jobs are finished.

### DIFF
--- a/pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
+++ b/pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
@@ -4,9 +4,12 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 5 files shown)


pkg/controller/prpqr_reconciler/pjstatussyncer/pjstatussyncer.go
```diff
@@ -111,23 +117,49 @@ func (r *reconciler) reconcile(ctx context.Context, log *logrus.Entry, req contr
...
+
+		if hasAllJobsFinished(prpqr.Status.Jobs) {
+			prpqr.Status.Conditions = append(prpqr.Status.Conditions, metav1.Condition{
+				Type:               conditionAllJobsFinished,
+				Status:             metav1.ConditionTrue,
+				LastTransitionTime: metav1.Time{Time: time.Now()},
+				Reason:             conditionAllJobsFinished,
+				Message:            "All jobs have finished.",
+			})
+		}
```
petr-muller, line 144: The condition needs to go there even if not all jobs finished, just with `Status` set to `false` instead of true` and a different message ("X jobs are still running")

The logic is:
- Compute the condition that should be there (just type, status, reason, message)
- Compare with the condition in the status of the matching type (all members)
- If the condition is not there at all, or if the condition differs, update the condition

petr-muller (approved):

petr-muller:
This looks nice!

PR #2526, opened 2021-11-24
Allow build04 for a cluster name

hongkailiu:
/cc @openshift/test-platform

### DIFF
--- a/pkg/api/constant.go
+++ b/pkg/api/constant.go
@@ -58,7 +58,7 @@ const (
 )
 
 var (
-	ValidClusterNames = sets.NewString(
+	clusterNames = sets.NewString(
 		string(ClusterAPPCI),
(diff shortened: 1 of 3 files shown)


hongkailiu:
/hold

Let think if there is a better solution.

smg247 (approved):

pkg/api/constant.go
```diff
@@ -58,11 +58,16 @@ const (
 )
 
 var (
-	ValidClusterNames = sets.NewString(
+	clusterNames = sets.NewString(
 		string(ClusterAPPCI),
 		string(ClusterBuild01),
 		string(ClusterBuild02),
 		string(ClusterBuild03),
```
petr-muller, line 65: do we need a line for 04 too?

PR #2527, opened 2021-11-25
prpqr: integrate the dynamic job creation

petr-muller:
This PR should do the final trick. The PR is ready for review, but I feel like we are missing some tests, so I will add some.


petr-muller:
/hold
Rebased over https://github.com/openshift/ci-tools/pull/2528 to improve tests and their readability

petr-muller:
No need to retest, this is genuine

cmd/job-trigger-controller-manager/main.go
```diff
@@ -29,6 +39,8 @@ func gatherOptions() (*options, error) {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
 	fs.StringVar(&o.namespace, "namespace", "ci", "In which namespace the operation will take place")
+	fs.StringVar(&o.prowConfigPath, "prow-config-path", "", "Path to the Prow configuration directory.")
```
droslean, line 42: Can you just use `k8s.io/test-infra/prow/flagutil/config` instead?

petr-muller: Yes, you are right, we could do that instead :+1:

petr-muller:
@droslean Addressed. Also added a fix for variant presence in the job name

droslean (approved):

petr-muller:
DPTP-2577

PR #2529, opened 2021-11-25
prowgen: do not use projected secrets unless needed

petr-muller:
Projected secrets are useful when we need multiple sources melded into a
single volume. In most cases, our cluster profiles are just a single
secret, so using secrets directly as a volume source saves a bunch of
YAML lines (and simplifies the config).

Expected to fail breaking-changes check, needs a followup PR

/hold


bbguimaraes:
Nice.

/lgtm

PR #2543, opened 2021-12-01
job-trigger-cm: code structure refactor

petr-muller:
Previous code structure assumed `generateProwjobs` is a cheap operation
(it is not, each Prowjob is a call to ci-operator-configresolver) so it
first generated all prowjobs and then it processed them.

This PR changes the reconciliation algorithm to straightforward
iteration over all jobs that *should* exist and hence should be present
in the status. First the code checks whether there is evidence that they
already exists (or existed in the past): either by presence in the
status, or by presence in the cluster. If the job is not present, it is
created. For all jobs, we track our observation of what their status
should be in our PRPQR resource.

Finally, the status reconciling takes into account that the updated
PRPQR can have *better* information than what *we* have, and we only
update the status when ours is more precise (for example, we will not
overwrite a status for an existing job with our error status).

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -5,6 +5,7 @@ import (
 	"crypto/md5"
 	"encoding/hex"
 	"fmt"
+	"reflect"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -134,8 +135,7 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 	logger = logger.WithField("namespace", req.Namespace).WithField("prpqr_name", req.Name)
 	logger.Info("Starting reconciliation")
 
-	var prpqrMutations []func(prpqr *v1.PullRequestPayloadQualificationRun)
-	createdJobs := make(map[string]v1.PullRequestPayloadJobStatus)
+	statuses := make(map[string]*v1.PullRequestPayloadJobStatus)
 
 	prpqr := &v1.PullRequestPayloadQualificationRun{}
 	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: req.Namespace, Name: req.Name}, prpqr); err != nil {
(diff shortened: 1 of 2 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -254,103 +345,94 @@ func jobNameHash(name string) string {
 	return hex.EncodeToString(hasher.Sum(nil))
 }
 
-func generateProwjobs(rc injectingResolverClient, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, releaseJobSpec []v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest) (map[string]prowv1.ProwJob, map[string]error) {
+func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, spec *v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest) (*prowv1.ProwJob, error) {
```
petr-muller, line 348: This ugly part is really just all code shifted left after the for cycle inside was removed. I suggest to review with hidden whitespace changes.

pkg/controller/prpqr_reconciler/testdata/zz_fixture_prpqr_TestReconcile_basic_case__prowjob_already_exists__no_updates.yaml
```diff
@@ -34,3 +34,9 @@
       reason: AllJobsTriggered
       status: "True"
       type: AllJobsTriggered
+    jobs:
+    - jobName: periodic-ci-test-org-test-repo-test-branch-test-name
+      prowJob: some-uuid
+      status:
+        startTime: "1970-01-01T00:00:00Z"
+        state: triggered
```
petr-muller, line 42: This change shows one of the impacts of the change - the job status was reconciled because it was previously missing in the resource.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -225,12 +258,70 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+	for _, spec := range theirs.Spec.Jobs.Jobs {
+		jobName := spec.JobName(jobconfig.PeriodicPrefix)
+		our := ourStatuses[jobName]
+		their := statusByJobName[jobName]
+		theirs.Status.Jobs = append(theirs.Status.Jobs, reconcileJobStatus(jobName, their, our))
+	}
+}
+
+func reconcileJobStatus(name string, their, our *v1.PullRequestPayloadJobStatus) v1.PullRequestPayloadJobStatus {
+	if their == nil && our == nil {
```
droslean, line 292: The only way for `our == nil` is to fail to create a prowjob before, which in this case we return error either way. So this if statement will never be true.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -151,26 +151,50 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
 
-	baseMetadata := metadataFromPullRequestUnderTest(prpqr.Spec.PullRequest)
-	prowjobs, errs := generateProwjobs(r.configResolverClient, r.prowConfigGetter.Config(), baseMetadata, req.Name, req.Namespace, prpqr.Spec.Jobs.Jobs, &prpqr.Spec.PullRequest)
-	for job, err := range errs {
-		createdJobs[job] = v1.PullRequestPayloadJobStatus{ReleaseJobName: job, Status: prowv1.ProwJobStatus{
-			State:       prowv1.ErrorState,
-			Description: err.Error(),
-		}}
+	statusByJobName := map[string]*v1.PullRequestPayloadJobStatus{}
+	for i := range prpqr.Status.Jobs {
```
droslean, line 155: This will always be empty, so there is no need for this loop.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -225,12 +258,70 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+	theirs.Status.Jobs = []v1.PullRequestPayloadJobStatus{}
+	for _, spec := range theirs.Spec.Jobs.Jobs {
+		jobName := spec.JobName(jobconfig.PeriodicPrefix)
+		our := ourStatuses[jobName]
+		their := statusByJobName[jobName]
+		theirs.Status.Jobs = append(theirs.Status.Jobs, reconcileJobStatus(jobName, their, our))
+	}
+}
+
+func reconcileJobStatus(name string, their, our *v1.PullRequestPayloadJobStatus) v1.PullRequestPayloadJobStatus {
```
droslean, line 291: I don't think that we need to reconcile the job statuses, since they will always be empty when we will reach this code.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
petr-muller: That's just defensive coding, preventing from NPA on bugs.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
petr-muller: That's not true. Status will have normal content e.g. when you restart the controller and that will make you re-reconcile all PRPQRs in the cluster.

petr-muller: But overall, I think that the contrroller (any controller basically) should have as few assumptions about is the incoming state of the resource as possible. That makes it robust against changes. If you do not assume *anything* and always act only on what you actually observe, you are robust against any external change.

PR #2553, opened 2021-12-02
[poc] [wip] example of wiring to real kube wait for waitforpod

deads2k:
This is a POC to show how the native kubernetes constructs for list/watch can be used to avoid races with deleted pods and avoid polling GET requests from every job against the kube-apiserver.

I did not take the plumbing all the way out, but provided an example in one chain

/hold

### DIFF
--- a/pkg/steps/template.go
+++ b/pkg/steps/template.go
@@ -12,17 +12,27 @@ import (
 	"strings"
 	"time"
 
+	"k8s.io/apimachinery/pkg/runtime/schema"
+
+	corev1 "k8s.io/client-go/kubernetes/typed/core/v1"
+
+	"k8s.io/client-go/tools/cache"
+
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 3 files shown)


pkg/steps/template.go
```diff
@@ -519,84 +531,101 @@ func waitForPodCompletion(ctx context.Context, podClient PodClient, namespace, n
...
 }
 
-func waitForPodCompletionOrTimeout(ctx context.Context, podClient PodClient, namespace, name string, completed map[string]time.Time, notifier ContainerNotifier, skipLogs bool) (*coreapi.Pod, error) {
-	// Warning: this is extremely fragile, inherited legacy code.  Please be
-	// careful and test thoroughly when making changes, as they very frequently
-	// lead to systemic production failures.  Some guidance:
-	// - There is a complex interaction between this code and the container
-	//   notifier.  Updates to the state of the pod are received via the watch
-	//   and communicated to the notifier.  Even in case of interruption (i.e.
-	//   cancellation of `ctx`) and/or failure, events should continue to be
```
deads2k, line 529: the existing code doesn't match this comment.  When the context is cancelled, the for loop hits  branch that insta-returns.

PR #2554, opened 2021-12-02
resolver: base images only conflict when different

petr-muller:
Previously we errored out on the sole presence of a base image name in both files. We can easily tolerate that unless the imagestream references are actually different.

/cc @openshift/test-platform @deads2k


hongkailiu:
/lgtm

/hold

Give David a chance to take a look.

petr-muller:
/hold cancel

IMO not necessary, I cced David mostly for FYI if interested

PR #2558, opened 2021-12-07
Renaming cluster_profile for Alibaba to alibabacloud

kwoodson:
To be consistent with naming I would like to rename the cluster_profile field to `alibabacloud`. This matches the installer and other OpenShift components.

Corresponding rename: https://github.com/openshift/release/pull/24347


kwoodson:
This is the related PR to rename the quotas https://github.com/openshift/release/pull/24347

PR #2562, opened 2021-12-08
force all disruption aggregation to skipped status while we re-gather…

deads2k:
… unbugged data

/hold

holding until we're ready.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
@@ -70,6 +70,9 @@ func (o *JobRunAggregatorAnalyzerOptions) CalculateDisruptionTestSuite(ctx conte
 				return nil, err
 			}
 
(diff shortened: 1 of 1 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_disruption.go
```diff
@@ -70,6 +70,9 @@ func (o *JobRunAggregatorAnalyzerOptions) CalculateDisruptionTestSuite(ctx conte
 				return nil, err
 			}
 
+			// we are correcting our disruption numbers, so we are forcing everything to skipped until 12/15
+			status = testCaseSkipped
```
dgoodwin, line 74: status can be reset to testCaseFailed just below here, should this be right before we return?

dgoodwin: n/m I misread == below

dgoodwin:
/close

Replaced by #2566

PR #2563, opened 2021-12-08
cluster-init: updateClusterDisplay

hongkailiu:


cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -219,6 +220,14 @@ func updatePromotedImageGovernor(c *secretbootstrap.Config, o options) error {
 	})
 }
 
+func updateClusterDisplay(c *secretbootstrap.Config, o options) error {
```
petr-muller, line 223: Err... what is cluster display and why didn't we need this yet? Seems like the `breaking-changes` check shows this added for all four `build0X` clusters.

petr-muller (approved):
/hold

Could be worth a comment about what is "cluster display" because I don't know that :P

cmd/cluster-init/update_bootstrap_secrets.go
hongkailiu: > what is cluster display?

It is a web service to provide information about clusters in CI.
https://github.com/openshift/release/blob/5db2fb70172ce07c7f11c9f4e5c989ffb87bb50e/clusters/app.ci/assets/docs.yaml#L44

It is used by our our doc site to provide information about pools:
https://github.com/openshift/ci-docs/blob/b1084e5d654c90a18a195a0bf66e12e80b6a11f9/layouts/partials/head.html#L61

> why didn't we need this yet?

Pools are in the hive cluster. It needs only its kubeconfig.
But it is going to be changed because it provides information about registries and consoles for all clusters.

> Seems like the breaking-changes check shows this added for all four build0X clusters.

The breaking change should be fixed by https://github.com/openshift/release/pull/24392

hongkailiu: Our doc site: hugo + docs theme
It is mainly about static pages.

With cluster-display and some hacks into docs theme, we can do dynamic contents on the doc site.
I thought we could do it for the payload UI but it did not go that way. ^_^

hongkailiu:
breaking has been fixed as expected.

/hold cancel

PR #2564, opened 2021-12-08
[payload-testing-prow-plugin] Include aggregated job count information in PRPQR creation

droslean:
/cc @openshift/test-platform 

in the effort of https://issues.redhat.com/browse/DPTP-2598

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -186,7 +186,6 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 		})
 		builder.spec = spec
 		var jobNames []string
-		var jobTuples []api.MetadataWithTest
 		specLogger.Debug("resolving jobs ...")
 		startResolveJobs := time.Now()
 		jobs, err := s.jobResolver.resolve(spec.ocp, spec.releaseType, spec.jobs)
@@ -198,12 +197,20 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 		}
 		specLogger.Debug("resolving tests ...")
 		startResolveTests := time.Now()
+
+		var releaseJobSpecs []prpqv1.ReleaseJobSpec
 		for _, job := range jobs {
 			if job.Test != "" {
 				jobNames = append(jobNames, job.Name)
-				jobTuples = append(jobTuples, api.MetadataWithTest{
-					Metadata: job.Metadata,
-					Test:     job.Test,
+				releaseJobSpecs = append(releaseJobSpecs, prpqv1.ReleaseJobSpec{
+					CIOperatorConfig: prpqv1.CIOperatorMetadata{
+						Org:     job.Metadata.Org,
+						Repo:    job.Metadata.Repo,
+						Branch:  job.Metadata.Branch,
+						Variant: job.Metadata.Variant,
+					},
+					Test:            job.Test,
+					AggregatedCount: job.AggregatedCount,
 				})
 			} else {
 				jobTuple, err := s.testResolver.resolve(job.Name)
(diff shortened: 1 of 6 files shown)


pkg/release/config/client.go
```diff
@@ -22,12 +22,19 @@ func endpoint(c api.Candidate) string {
 type Job struct {
 	Name                 string `json:"name"`
 	api.MetadataWithTest `json:",inline"`
+
+	AggregatedCount int `json:"aggregatedCount,omitempty"`
```
hongkailiu, line 26:
```suggestion
	AggregatedCount int `json:"-"`
```
This field should be ignored in marshal/unmarshal, right?

pkg/release/config/client.go
```diff
@@ -22,12 +22,19 @@ func endpoint(c api.Candidate) string {
...
+type AggregatedJob struct {
+	AnalysisJobCount int `json:"analysisJobCount"`
 }
 
 type Verify map[string]VerifyItem
 
 type VerifyItem struct {
-	ProwJob Job `json:"prowJob"`
+	ProwJob           Job           `json:"prowJob"`
+	AggregatedProwJob AggregatedJob `json:"aggregatedProwJob,omitempty"`
```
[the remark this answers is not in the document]
droslean: This can actually be empty. This struct is being used to marshal the release controller's response to an object. The response can contain jobs that don't have the `AggregatedProwJob` field.

pkg/release/config/client.go
droslean: Well, we never marshal/unmarshal this struct. To be honest I added the tag because I was unmarshaling it while debugging.

cmd/payload-testing-prow-plugin/server.go
```diff
@@ -213,15 +219,25 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
 					continue
 				}
 				jobNames = append(jobNames, job.Name)
-				jobTuples = append(jobTuples, jobTuple)
+				releaseJobSpecs = append(releaseJobSpecs, prpqv1.ReleaseJobSpec{
```
hongkailiu, line 222: Good improvement!

cmd/payload-testing-prow-plugin/server.go
```diff
@@ -213,15 +219,25 @@ func (s *server) handle(l *logrus.Entry, ic github.IssueCommentEvent) string {
...
+						Repo:    jobTuple.Metadata.Repo,
+						Branch:  jobTuple.Metadata.Branch,
+						Variant: jobTuple.Metadata.Variant,
+					},
+					Test:            jobTuple.Test,
+					AggregatedCount: job.AggregatedCount,
+				})
 			}
 		}
+		startResolveTests := time.Now()
```
hongkailiu, line 234: Could we keep this line as before?

pkg/release/config/client.go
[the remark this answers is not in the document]
droslean: I updated the code, and added also a testcase where this field is empty.

pkg/api/pullrequestpayloadqualification/v1/types.go
```diff
@@ -101,6 +101,8 @@ type ReleaseJobSpec struct {
 	CIOperatorConfig CIOperatorMetadata `json:"ciOperatorConfig"`
 	// Test is the name of the test in the ci-operator configuration
 	Test string `json:"test"`
+	// AggregatedCount is a number that specifies how many instances of the job will run in parallel
```
petr-muller, line 104: Let's clarify the corner cases of 0 and 1. It seems that `0` means that the job is not run as aggregated, and `1` would mean an aggregated job with a single execution? If so let's put it into the comment.

droslean: updated.

pkg/release/config/client.go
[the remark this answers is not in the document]
hongkailiu: This is what i meant.
https://www.sohamkamani.com/golang/omitempty/#values-that-cannot-be-omitted

pkg/release/config/client.go
hongkailiu: We did. `Job` is part of `Verify`.
https://github.com/openshift/ci-tools/blob/26d2a699900570e2eff559da8a8543d8b298b493/pkg/release/config/client.go#L73

hongkailiu: We did. `Job` is part of `Verify`.
https://github.com/openshift/ci-tools/blob/26d2a699900570e2eff559da8a8543d8b298b493/pkg/release/config/client.go#L73

pkg/release/config/client.go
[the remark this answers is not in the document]
droslean: updated. Thanks!

hongkailiu (approved):

PR #2565, opened 2021-12-08
prowgen: hash-spread @daily periodics

petr-muller:
With `@daily`, all periodics hit exactly at midnight which makes jobs race for boskos leases, hit cloud platform APIs at the same time and stress build farm clusters. We could spread these with Prowgen so they hit deterministically between 22 and 04 UTC.


petr-muller:
//hold 

Expect breaking changes to fail, needs o/release followup

petr-muller:
/hold

I had a typo above

PR #2566, opened 2021-12-09
skip disrupt tests

dgoodwin:
- force all disruption aggregation to skipped status while we re-gather unbugged data
- Fix lint issue.


dgoodwin:
Replaces #2562

PR #2567, opened 2021-12-10
Support runIfChanged and skipIfOnlyChanged for postsubmit

clyang82:
We have a case to define a publish the image job with `skip_if_only_changed` or `run_if_changed`. we want to configure it in ci config, but ci-operator cannot generate `skip_if_only_changed` or `run_if_changed` in jobs.

Signed-off-by: clyang82 <chuyang@redhat.com>


petr-muller (changes requested):
Please fix integration tests.

/hold
This will need o/release followup so holding

petr-muller (approved):

PR #2573, opened 2021-12-18
handle missing history for aggregated disruption

deads2k:
Fixes an out of range problem.  This has to merge until we get history built up.  I'm also planning to inject history manually if possible.


petr-muller:
/hold

obey the linter :trollface:

deads2k:
manually labelling so that we'll match reality

PR #2575, opened 2021-12-20
Rename mis-spelled variable, add comments

DennisPeriquet:


pkg/jobrunaggregator/jobrunaggregatorapi/types.go
```diff
@@ -34,6 +38,7 @@ type BackendDisruptionRow struct {
 	DisruptionSeconds int
 }
 
+// There does not seem to be a corresponding table for this
```
deads2k, line 41: This one has a view I think.  It may be built in-memory

pkg/jobrunaggregator/jobrunaggregatorapi/types.go
```diff
@@ -2,6 +2,9 @@ package jobrunaggregatorapi
 
 import "time"
 
+// For all structs that end in "Row", they correspond to a Big Query row in a similarly
+// named table.  See https://console.cloud.google.com/bigquery?project=<theProject>
+// and pick a dataset in use to see the table.
```
dgoodwin, line 7: This is probably technically an invalid godoc comment, I believe they're supposed to start with the name of the type your comment is attached to. (i.e. UnifiedTestRunRow corresponds to...) Some linters might reject this, and perhaps you could work around that with a newline before the typedef starts, or moving up to package level documentation. See: https://go.dev/blog/godoc

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -80,9 +80,12 @@ func NewCIDataClient(dataCoordinates BigQueryDataCoordinates, client *bigquery.C
 	}
 }
 
+// Get every job in the Jobs table ordered by name and return it as one big
```
dgoodwin, line 83: ListAllJobs gets every job in the Jobs ...

pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
```diff
@@ -98,11 +98,14 @@ func (o *ciGCSClient) ListJobRunNamesOlderThanFourHours(ctx context.Context, job
 	return jobRunProcessingCh, errorCh, nil
 }
 
+// This is where we read the GCS bucket to get:
```
dgoodwin, line 101: ReadJobRunFromGCP reads the GCS...

pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
```diff
@@ -115,6 +118,10 @@ func (o *ciGCSClient) ReadJobRunFromGCS(ctx context.Context, jobGCSRootLocation,
 	if err := query.SetAttrSelection([]string{"Name", "Created", "Generation"}); err != nil {
 		return nil, err
 	}
+
+	// NOTE: StartOffset and EndOffset are strings that the GCS api uses to limit where you
+	// start and end reading; here, we limit it by the jobRunID which is a 19 digit integer
```
dgoodwin, line 123: I doubt it's fixed at 19 digits, probably best to just say very large.

pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
```diff
@@ -16,6 +16,7 @@ type BigQueryTestRunUploadFlags struct {
 	DataCoordinates *jobrunaggregatorlib.BigQueryDataCoordinates
 	Authentication  *jobrunaggregatorlib.GoogleAuthenticationFlags
 
+	// Indicates if we actually upload to Big Query or just output to stdout (for debugging purposes)
```
dgoodwin, line 19: DryRun indicates if...

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
```diff
@@ -45,8 +49,12 @@ func newTestRunRow(jobRun jobrunaggregatorapi.JobRunInfo, prowJob *prowv1.ProwJo
 
 }
 
+// Ensure (at compile time) that testRunRow implements the bigquery.ValueSaver interface
 var _ bigquery.ValueSaver = &testRunRow{}
 
+// The "Put" method uploads a row to Big Query.
```
dgoodwin, line 55: Should also start with "Save ..."

dgoodwin (commented):
Just nits to fixup the godoc slightly so it will pass the linters that check that sort of thing.

pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
```diff
@@ -115,6 +118,10 @@ func (o *ciGCSClient) ReadJobRunFromGCS(ctx context.Context, jobGCSRootLocation,
 	if err := query.SetAttrSelection([]string{"Name", "Created", "Generation"}); err != nil {
 		return nil, err
 	}
+
+	// NOTE: StartOffset and EndOffset are strings that the GCS api uses to limit where you
```
deads2k, line 122: nit, isn't the explanation of these simply an IDE hover for their godoc?  Those who don't use modern IDEs are punished accordingly ;)

pkg/jobrunaggregator/jobrunaggregatorlib/retrying_ci_data_client.go
```diff
@@ -16,6 +16,7 @@ type retryingCIDataClient struct {
 	delegate CIDataClient
 }
 
+// Ensure (at compile time) that retryingCIDataClient implements the CIDataClient interface
```
deads2k, line 19: nit: this is canonical. documenting it looks weird to me.

pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
```diff
@@ -16,6 +16,8 @@ type BigQueryTestRunUploadFlags struct {
 	DataCoordinates *jobrunaggregatorlib.BigQueryDataCoordinates
 	Authentication  *jobrunaggregatorlib.GoogleAuthenticationFlags
 
+	// DryRun indicates if we are actually going to upload to Big Query or just output to
```
deads2k, line 19: I prefer the explicity. 

```
// DryRun controls if we are actually going to upload to Big Query.  If it is true, then no actual mutation happens.  If it is false, then data is inserted into bigquery
```

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
```diff
@@ -29,8 +29,12 @@ func newJobRunRow(jobRun jobrunaggregatorapi.JobRunInfo, prowJob *prowv1.ProwJob
 }
 
 type testRunRow struct {
-	prowJob    *prowv1.ProwJob
-	jobRun     jobrunaggregatorapi.JobRunInfo
+	prowJob *prowv1.ProwJob
+	jobRun  jobrunaggregatorapi.JobRunInfo
+
+	// ci-tools/pkg/junit/types.go mentions that a TestSuite can itself hold
+	// TestSuites, hence declared as a slice below; but we only use a single
```
deads2k, line 36: do we store anything here at all?

pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
DennisPeriquet: Good point; if I over on StartOffset, it takes me to pkg.go.dev and I see what you mean.

DennisPeriquet:
too old.

PR #2576, opened 2021-12-20
Add aggregating job results to testgrid

xueqzhan:
Aggregating jobs are added to blocking dashboard of the corresponding release.
Release is determined by the aggregated job prow config.

### DIFF
--- a/cmd/testgrid-config-generator/main.go
+++ b/cmd/testgrid-config-generator/main.go
@@ -206,10 +206,14 @@ func addDashboardTab(p prowConfig.Periodic,
 	configuredJobs map[string]string,
 	allowList map[string]string,
 	aggregateJobName *string) {
-	name := p.Name
+	prowName := p.Name
+	jobName := p.Name
 	var dashboardType string
 
-	label, ok := allowList[name]
+	if aggregateJobName != nil {
+		jobName = *aggregateJobName
+	}
+	label, ok := allowList[jobName]
 	if len(label) == 0 && ok {
 		// if the allow list has an empty label for the type, exclude it from dashboards
 		return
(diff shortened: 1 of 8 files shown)


petr-muller:

LGTM (I'm not familiar with the code but I'm as good reviewer as any)  but has conflicts

pkg/release/config/config.go
```diff
@@ -34,12 +34,13 @@ type Job struct {
 }
 
 type AggregatedJob struct {
-	AnalysisJobCount int `json:"analysisJobCount"`
+	ProwJob          *Job `json:"prowJob,omitempty"`
```
hongkailiu, line 37: Do we have an example of this in https://github.com/openshift/release/tree/master/core-services/release-controller/_releases?

pkg/release/config/config.go
```diff
@@ -34,12 +34,13 @@ type Job struct {
...
 
 type VerifyItem struct {
-	Optional          bool          `json:"optional"`
-	Upgrade           bool          `json:"upgrade"`
-	ProwJob           Job           `json:"prowJob"`
-	AggregatedProwJob AggregatedJob `json:"aggregatedProwJob"`
+	Optional          bool           `json:"optional"`
+	Upgrade           bool           `json:"upgrade"`
+	ProwJob           Job            `json:"prowJob"`
+	AggregatedProwJob *AggregatedJob `json:"aggregatedProwJob,omitempty"`
```
hongkailiu, line 45: Changing it to a pointer might break our existing tool.
/cc @droslean

hongkailiu (commented):
Not familiar with the `generator`.
Just checked on the change on the RC's configuration.

pkg/release/config/config.go
hongkailiu: I am surprised that unit tests passed.

pkg/release/config/config.go
xueqzhan: Not at this moment. The codes support it. So if anyone adds the config, this should be support by my change.

pkg/release/config/config.go
xueqzhan: I added the check in the client code for nil pointer. Is there anywhere else we are using this?

pkg/release/config/config.go
hongkailiu: >Not at this moment. The codes support it.

Do you mean the code of release controllers? Could you please link?

xueqzhan: Yes, release controller. Here is the line: https://github.com/openshift/release-controller/blob/e6e5923afde7a61bd763c3f1b3e0e3bef0ab4e9a/cmd/release-controller/sync_verify_prow.go#L73

pkg/release/config/config.go
hongkailiu: At least, this one.
https://github.com/openshift/ci-tools/blob/98048e212a80054c036fdf38a4a31ddce9c5ad9a/pkg/release/config/client.go#L74

xueqzhan: This one has been addressed. Maybe you are not seeing all my changes? Yes, I did miss that one in my very first commit. :)

hongkailiu: ohoh. I see.
All good to me for the config's change.

xueqzhan:
@petr-muller Need a lgtm from you again after the merge. Thanks.

petr-muller (approved):

PR #2577, opened 2021-12-20
[DPTP-2598] Support spawing aggregated jobs when its needed

droslean:
https://issues.redhat.com/browse/DPTP-2598

These changes allow the controller to check if the job is aggregated by reading the [AggregatedCount](https://github.com/openshift/ci-tools/blob/master/pkg/api/pullrequestpayloadqualification/v1/types.go#L107) field.
If that number is more than 0, then the controller spawns the same prowjob X times including the `release.openshift.io/aggregation-id` label with the same UID. Then it spawns the aggregator analysis job.

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -178,7 +178,20 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 			continue
 		}
 
-		prowjob, err := generateProwjob(r.configResolverClient, r.prowConfigGetter.Config(), baseMetadata, req.Name, req.Namespace, &jobSpec, &prpqr.Spec.PullRequest)
+		ciopConfig, err := resolveCiopConfig(r.configResolverClient, baseMetadata, jobSpec)
+		if err != nil {
+			logger.WithError(err).Error("Failed to resolve the ci-operator configuration")
+			statuses[mimickedJob] = &v1.PullRequestPayloadJobStatus{
+				ReleaseJobName: mimickedJob,
+				Status: prowv1.ProwJobStatus{
+					State:       prowv1.ErrorState,
+					Description: err.Error(),
+				},
+			}
+			continue
+		}
+
+		prowjob, err := generateProwjob(ciopConfig, r.prowConfigGetter.Config(), baseMetadata, req.Name, req.Namespace, &jobSpec, &prpqr.Spec.PullRequest)
 		if err != nil {
 			logger.WithError(err).Error("Failed to create a payload prowjob")
 			statuses[mimickedJob] = &v1.PullRequestPayloadJobStatus{
@@ -345,7 +358,26 @@ func jobNameHash(name string) string {
 	return hex.EncodeToString(hasher.Sum(nil))
 }
 
-func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, spec *v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest) (*prowv1.ProwJob, error) {
+func resolveCiopConfig(rc injectingResolverClient, baseCiop *api.Metadata, spec v1.ReleaseJobSpec) (*api.ReleaseBuildConfiguration, error) {
+	inject := &api.MetadataWithTest{
+		Metadata: api.Metadata{
+			Org:     spec.CIOperatorConfig.Org,
+			Repo:    spec.CIOperatorConfig.Repo,
+			Branch:  spec.CIOperatorConfig.Branch,
+			Variant: spec.CIOperatorConfig.Variant,
+		},
+		Test: spec.Test,
+	}
+
+	ciopConfig, err := rc.ConfigWithTest(baseCiop, inject)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get config from resolver: %w", err)
+	}
+
+	return ciopConfig, nil
+}
+
+func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, spec *v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest) (*prowv1.ProwJob, error) {
 	fakeProwgenInfo := &prowgen.ProwgenInfo{Metadata: *baseCiop}
 
 	mimickedJob := spec.JobName(jobconfig.PeriodicPrefix)
@@ -365,11 +397,6 @@ func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, ba
 			Variant: spec.CIOperatorConfig.Variant,
 		},
 		Test: spec.Test,
-	}
-
-	ciopConfig, err := rc.ConfigWithTest(baseCiop, inject)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get config from resolver: %w", err)
 	}
 
 	var periodic *prowconfig.Periodic
@@ -425,14 +452,14 @@ func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, ba
 	}
 	periodic.ExtraRefs = []prowv1.Refs{extraRefs}
 
-	if err = defaulter.DefaultPeriodic(periodic); err != nil {
+	if err := defaulter.DefaultPeriodic(periodic); err != nil {
 		return nil, fmt.Errorf("failed to default the ProwJob: %w", err)
 	}
 
 	pj := pjutil.NewProwJob(pjutil.PeriodicSpec(*periodic), labels, annotations)
 	pj.Namespace = prpqrNamespace
 
-	return &pj, err
+	return &pj, nil
 }
 
 func metadataFromPullRequestUnderTest(pr v1.PullRequestUnderTest) *api.Metadata {
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -10,7 +10,9 @@ import (
 
 	"github.com/sirupsen/logrus"
 
+	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -42,6 +44,8 @@ const (
 
 	conditionAllJobsTriggered = "AllJobsTriggered"
 	conditionWithErrors       = "WithErrors"
+
+	aggregationIDLabel = "release.openshift.io/aggregation-id"
 )
 
 type injectingResolverClient interface {
@@ -159,7 +163,13 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 
 	baseMetadata := metadataFromPullRequestUnderTest(prpqr.Spec.PullRequest)
 	for _, jobSpec := range prpqr.Spec.Jobs.Jobs {
+		var prowjobsToCreate []*prowv1.ProwJob
 		mimickedJob := jobSpec.JobName(jobconfig.PeriodicPrefix)
+		if jobSpec.AggregatedCount > 0 {
+			// We treat the aggregator job as the mimicked job, and we assume if this job exists then
+			// all the aggregated jobs exist too.
+			mimickedJob = fmt.Sprintf("aggregator-%s", jobSpec.JobName(jobconfig.PeriodicPrefix))
+		}
 		logger = logger.WithFields(logrus.Fields{"want-job": mimickedJob})
 
 		if status, exists := statusByJobName[mimickedJob]; exists {
@@ -178,7 +188,17 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 			continue
 		}
 
-		ciopConfig, err := resolveCiopConfig(r.configResolverClient, baseMetadata, jobSpec)
+		inject := &api.MetadataWithTest{
+			Metadata: api.Metadata{
+				Org:     jobSpec.CIOperatorConfig.Org,
+				Repo:    jobSpec.CIOperatorConfig.Repo,
+				Branch:  jobSpec.CIOperatorConfig.Branch,
+				Variant: jobSpec.CIOperatorConfig.Variant,
+			},
+			Test: jobSpec.Test,
+		}
+
+		ciopConfig, err := resolveCiopConfig(r.configResolverClient, baseMetadata, inject)
 		if err != nil {
 			logger.WithError(err).Error("Failed to resolve the ci-operator configuration")
 			statuses[mimickedJob] = &v1.PullRequestPayloadJobStatus{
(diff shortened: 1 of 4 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -438,3 +529,69 @@ func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, ba
...
+			{
+				Name: "pull-secret",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{SecretName: "registry-pull-credentials"},
+				},
+			},
+			{
+				Name: "gcs-credentials",
+				VolumeSource: corev1.VolumeSource{
+					Secret: &corev1.SecretVolumeSource{SecretName: "gce-sa-credentials-gcs-publisher"},
```
droslean, line 546: I guess we will need to have another secret just for this case. 
@deads2k Can you let us know the required permissions for the service account?

petr-muller:
Can you add a description of the high-level logic that the controller is using to manage aggregated jobs?

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -345,10 +384,56 @@ func jobNameHash(name string) string {
...
+			logger.WithField("prowjob", job.Name).Debug("Prowjob already exists")
+			statuses[jobName] = &v1.PullRequestPayloadJobStatus{
+				ReleaseJobName: jobName,
+				ProwJob:        job.Name,
+				Status:         job.Status,
+			}
+			continue
+		}
+
+		pj, err := generateProwjob(rc, defaulter, baseCiop, prpqrName, prpqrNamespace, spec, pr, jobName)
```
petr-muller, line 410: This needs to be optimized: right now `generateProwjob` is an expensive operation: injecting the test involves hitting the configresolver. For aggregated jobs we should only need to do it once, and generate the prowjobs from the structure what we obtain.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -157,44 +165,88 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+	var prowjobs []*prowv1.ProwJob
+
 	baseMetadata := metadataFromPullRequestUnderTest(prpqr.Spec.PullRequest)
 	for _, jobSpec := range prpqr.Spec.Jobs.Jobs {
 		mimickedJob := jobSpec.JobName(jobconfig.PeriodicPrefix)
-		logger = logger.WithFields(logrus.Fields{"want-job": mimickedJob})
+		jobSpecWithUID := jobSpecWithUID{ReleaseJobSpec: &jobSpec}
+
+		if jobSpec.AggregatedCount > 0 {
+			jobSpecWithUID.UID = uuid.NewV1().String()
```
petr-muller, line 176: Could we make the aggregation UID by hashing together the name of the PRPQR and the name of the aggreated job? That way it would be deterministic for the given prpqr+job and still unique.

petr-muller (commented):
I'm afraid we'll need to spend more time and effort on this one - I expected some discussion to happen before coding this about how we will represent the aggregate jobs in the status. 

I don't think including individual "subjobs" (components of the aggregated job) need to be present directly in the status - that will allow to keep the controller much simpler. Including the subjobs in the status on the same level like we do for "normal" job will just clutter the UI - the individual jobs are not important, and even release-controller does not show them, see:

https://amd64.ocp.releases.ci.openshift.org/releasestream/4.10.0-0.nightly/release/4.10.0-0.nightly-2021-12-18-034942

So I think the overall structure of the existing loop should not change so much like it is changed here. The loop over the jobs in the spec should look something like this:

- determine the name of the "mimicked job" (could be an aggregated job or a normal one, we dont' care)
- if we already have that job in status, do nothing 
- if we find that job in the cluster, update our status
^^^ these three are identical like here 

Now the differnet part:
- If the job is not aggregated, do the same thing like now
- If the job IS aggregated
-- Spawn the X subjobs: here we'll need to optimize the existing code, because the existing code couples the test injection (which is an API call to config resolver) with generating the prowjobs. We need to do the injection once and then generate X Prowjobs from it.
-- If we fail to spawn the X prowjobs, set the status of the high-level aggregator job to "failed to spawn aggregated jobs" or something similar
-- If we succeed to spawn X prowjobs, spawn the aggregator job and handle it the same way like we handle the current non-aggregated job: if we fail to create it, set its status to failure, if it fails to appear in the cache, set its status to failure etc...

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +199,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 			continue
 		}
 
-		logger.Info("Creating prowjob...")
-		if err := r.client.Create(ctx, prowjob); err != nil {
+		if jobSpec.AggregatedCount > 0 {
+			uid := jobNameHash(req.Name + mimickedJob)
+			aggregatedProwjobs, err := generateAggregatedProwjobs(uid, ciopConfig, r.prowConfigGetter.Config(), baseMetadata, req.Name, req.Namespace, &jobSpec, &prpqr.Spec.PullRequest)
+			if err != nil {
+				logger.WithError(err).Error("Failed to create the aggregated prowjobs")
```
hongkailiu, line 206:
```suggestion
				logger.WithError(err).Error("Failed to generate the aggregated prowjobs")
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +199,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+					},
+				}
+				continue
+			}
+			prowjobsToCreate = append(prowjobsToCreate, aggregatedProwjobs...)
+
+			// Generagete the aggregator job
+			aggregatorJob, err := createJobRunAggregatorJob(uid, mimickedJob, "test", req.Name, req.Namespace, r.prowConfigGetter.Config(), time.Now())
+			if err != nil {
+				logger.WithError(err).Error("Failed to create a payload prowjob")
```
hongkailiu, line 221:
```suggestion
				logger.WithError(err).Error("Failed to generate an aggregator prowjob")
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +199,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+						State:       prowv1.ErrorState,
+						Description: err.Error(),
+					},
+				}
+				continue
+			}
+			prowjobsToCreate = append(prowjobsToCreate, aggregatedProwjobs...)
+
+			// Generagete the aggregator job
+			aggregatorJob, err := createJobRunAggregatorJob(uid, mimickedJob, "test", req.Name, req.Namespace, r.prowConfigGetter.Config(), time.Now())
```
hongkailiu, line 219:
```suggestion
			aggregatorJob, err := generateAggregatorJob(uid, mimickedJob, "test", req.Name, req.Namespace, r.prowConfigGetter.Config(), time.Now())
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +199,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+			if err := r.client.Create(ctx, prowjob); err != nil {
+				statuses[mimickedJob] = &v1.PullRequestPayloadJobStatus{
+					ReleaseJobName: mimickedJob,
+					Status: prowv1.ProwJobStatus{
+						State:       prowv1.ErrorState,
+						Description: fmt.Errorf("failed to create prowjob: %w", err).Error(),
+					},
 				}
-				return false, fmt.Errorf("getting prowJob failed: %w", err)
+				continue
```
hongkailiu, line 264:
```suggestion
                                 logger.withField("job", prowjob.Job).withError(err).error("Creating prowjob...")
				continue
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +199,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
 
-		// There is some delay until it gets back to our cache, so block until we can retrieve
-		// it successfully.
-		key := ctrlruntimeclient.ObjectKey{Namespace: prowjob.Namespace, Name: prowjob.Name}
-		if err := wait.Poll(100*time.Millisecond, 5*time.Second, func() (bool, error) {
-			if err := r.client.Get(ctx, key, &prowv1.ProwJob{}); err != nil {
-				if kerrors.IsNotFound(err) {
-					return false, nil
+		for _, prowjob := range prowjobsToCreate {
+			logger.Info("Creating prowjob...")
```
hongkailiu, line 255:
```suggestion
			logger.withField("job", prowjob.Job).Info("Creating prowjob...")
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -345,10 +407,32 @@ func jobNameHash(name string) string {
...
+	}
+
+	return ciopConfig, nil
+}
+
+type options struct {
+	labels map[string]string
+}
+
+func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, spec *v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest, mimickedJob string, extraOptions ...options) (*prowv1.ProwJob, error) {
```
hongkailiu, line 433:
```suggestion
func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, spec *v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest, mimickedJob string, labels map[string]string) (*prowv1.ProwJob, error) {
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -425,16 +510,104 @@ func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, ba
...
 }
 
 func metadataFromPullRequestUnderTest(pr v1.PullRequestUnderTest) *api.Metadata {
 	return &api.Metadata{Org: pr.Org, Repo: pr.Repo, Branch: pr.BaseRef}
 }
+
+func generateAggregatedProwjobs(uid string, ciopConfig *api.ReleaseBuildConfiguration, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, spec *v1.ReleaseJobSpec, pr *v1.PullRequestUnderTest) ([]*prowv1.ProwJob, error) {
+	var ret []*prowv1.ProwJob
+
+	opts := options{labels: map[string]string{"release.openshift.io/aggregation-id": uid}}
```
hongkailiu, line 530: `release.openshift.io/aggregation-id` shows up multi-times: make a constant?

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -425,16 +510,104 @@ func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, ba
...
+				},
+				Command: []string{"job-run-aggregator"},
+				Args: []string{
+					"--analyze-job-runs",
+					"--google-service-account-credential-file=/secrets/gcs/service-account.json",
+					fmt.Sprintf("--job=%s", jobName),
+					fmt.Sprintf("--aggregation-id=%s", uid),
+					// TODO: determine how to get the correct gcs prefix for a job
+					// Now this value should be wrong.
+					fmt.Sprintf("--explicit-gcs-prefix=logs/%s", jobName),
```
petr-muller, line 579: Yep the `--job=` and `--explicit-gcs-prefix` will be different. `--job` is the original name of the job that we mimic, the value in `--explicit-gcs-prefix` is the name of th job that we submit, on L481

```go
periodic.Name = fmt.Sprintf("%s-%s-%d%s-%s", baseCiop.Org, baseCiop.Repo, pr.PullRequest.Number, variant, inject.Test)
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +199,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
 		}
 
-		// There is some delay until it gets back to our cache, so block until we can retrieve
-		// it successfully.
-		key := ctrlruntimeclient.ObjectKey{Namespace: prowjob.Namespace, Name: prowjob.Name}
-		if err := wait.Poll(100*time.Millisecond, 5*time.Second, func() (bool, error) {
-			if err := r.client.Get(ctx, key, &prowv1.ProwJob{}); err != nil {
-				if kerrors.IsNotFound(err) {
-					return false, nil
+		for _, prowjob := range prowjobsToCreate {
```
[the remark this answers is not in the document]
petr-muller: Also, I'm not entirely sure what to think about using the slice here: using the slice will allow you to reuse the code for submitting prowjobs (good) but also erases the difference between the aggregated and aggregator jobs: at the very least, if something fails, the description should inform whether we failed to submit an aggregator or aggregated job.

Maybe we could use the iteration index for that? If we failed to submit the last job in the slice, it was the "final" job - either it was the normal (non-aggregation) job, or it was the aggregator one. If it was anything else, it was an aggregated job.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
petr-muller, line 530: the label name should be a constant somewhere

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -367,11 +457,6 @@ func generateProwjob(rc injectingResolverClient, defaulter periodicDefaulter, ba
 		Test: spec.Test,
 	}
```
petr-muller, line 458: This is repeated here and in `resolveCiopConfig` - perhaps this needs to be also extracted and passed to both places as a parameter.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
petr-muller, line 433: One thing that must change in this method is this part on L468 - the individual aggregated jobs must not share a namespace, each one of them must have an  additional (different) custom job hash input

```
// Avoid sharing when we run the same job multiple times.
// PRPQR name should be safe to use as a discriminating input, because
// there should never be more than one execution of a specific job per
// PRPQR (until aggregated jobs, but for them we'll have a sequence index)
jobBaseGen.PodSpec.Add(prowgen.CustomHashInput(prpqrName))
```

pkg/controller/prpqr_reconciler/testdata/zz_fixture_prowjobs_TestReconcile_basic_aggregated_case.yaml
```diff
@@ -0,0 +1,215 @@
...
+    labels:
+      created-by-prow: "true"
+      prow.k8s.io/context: ""
+      prow.k8s.io/job: test-org-test-repo-100-test-name
+      prow.k8s.io/refs.base_ref: test-branch
+      prow.k8s.io/refs.org: test-org
+      prow.k8s.io/refs.pull: "100"
+      prow.k8s.io/refs.repo: test-repo
+      prow.k8s.io/type: periodic
+      pullrequestpayloadqualificationruns.ci.openshift.io: prpqr-test
```
petr-muller, line 154: So the aggregated jobs should **not** have this - only the top-level jobs should have it. Individual aggregated jobs' status will never need to be propagated to PRPQR status and we never need to track them back.

pkg/controller/prpqr_reconciler/testdata/zz_fixture_prowjobs_TestReconcile_basic_aggregated_case.yaml
```diff
@@ -0,0 +1,215 @@
...
+  status:
+    startTime: "1970-01-01T00:00:00Z"
+    state: triggered
+- apiVersion: prow.k8s.io/v1
+  kind: ProwJob
+  metadata:
+    annotations:
+      prow.k8s.io/context: ""
+      prow.k8s.io/job: test-org-test-repo-100-test-name
+      releaseJobName: periodic-ci-test-org-test-repo-test-branch-test-name-1
```
petr-muller, line 143: I think the value here should not have the `-1` suffix

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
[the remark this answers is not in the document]
droslean: I will fix the resetting of the slice (nice catch). After that, that slice will hold either aggregator/aggregated jobs or a single one. For the first case, the naming of the jobs already includes an index, so we can distinguish the job type while logging the error.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
droslean: I added the `extraOptions` here in order to avoid passing nil maps for the case where there are no extra labels needed for the job, and also passing a struct here allows to be extended in every future need.

petr-muller:
I don't have mental power to review code today anymore, will look tomorrow!

petr-muller:
Did not get to this because of the AWS incident :(

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -191,44 +211,98 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
-				return false, fmt.Errorf("getting prowJob failed: %w", err)
+				continue
 			}
-			return true, nil
-		}); err != nil {
+
+			// There is some delay until it gets back to our cache, so block until we can retrieve
+			// it successfully.
+			key := ctrlruntimeclient.ObjectKey{Namespace: prowjob.Namespace, Name: prowjob.Name}
+			if err := wait.Poll(100*time.Millisecond, 5*time.Second, func() (bool, error) {
```
jmguzik, line 282: I do not like "some delay" without actual backing the rationale (eg why 5 secs not 10, or not 15). On the other hand, one can spot this on several occasions coming across test-infra and ci-tools repos. So, it's just polemics.

jmguzik:
It's my second round with this PR. This time I'll take some responsibility. Nicely crafted!
/lgtm

PR #2580, opened 2021-12-21
prpqr: add a PR deploy makefile target

petr-muller:
Simplify hacking on PRPQR UI

https://payload-testing-ui-ci-tools-2580.apps.ci.l2s4.p1.openshiftapps.com/runs/


images/payload-testing-ui/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM centos:8
+FROM quay.io/centos/centos:stream8
```
petr-muller, line 1: I'm changing this because with vanilla `centos:8` we are hitting the docker hub rate limit

PR #2582, opened 2021-12-21
Create the Jobs table used by job-run-aggregation

DennisPeriquet:
The `Jobs` and `JobRuns` tables were manually created and are required for job-run-aggregation to run.

This PR adds code to create those two tables in case someone wants to try to run job-run-aggregation using a different dataset vs. the one populated with real data and in use by CI.

If this PR looks good, we'll augment it to add more tables as needed.


pkg/jobrunaggregator/jobruntablecreator/cmd.go
```diff
@@ -0,0 +1,93 @@
...
+}
+
+func NewBigQueryJobRunTableCreateFlagsCommand() *cobra.Command {
+	f := NewBigQueryJobRunTableCreateFlags()
+
+	cmd := &cobra.Command{
+		Use:          "create-tables",
+		Short:        "Create Jobs table in bigquery",
+		Long:         "Create Jobs table in bigquery",
+		SilenceUsage: false,
```
DennisPeriquet, line 37: Looking for feedback on if this should be silent or not.

dgoodwin: Looking at the docs on that flag it looks largely irrelevant, but I would say false is what we want. False is also the zero (or default) value for a bool in go so you can just technically just omit this field. (but not a required change)

pkg/jobrunaggregator/jobruntablecreator/table_creator.go
```diff
@@ -0,0 +1,37 @@
...
+	ciDataSet    *bigquery.Dataset
+}
+
+func (r *allJobRunTableCreatorOptions) Run(ctx context.Context) error {
+
+	// Create JobRunTable
+	jobRunTable := r.ciDataSet.Table(jobrunaggregatorlib.JobsTableName)
+	_, err := jobRunTable.Metadata(ctx)
+	if err != nil {
+		schema, err := bigquery.InferSchema(jobrunaggregatorapi.JobRow{})
```
dgoodwin, line 25: Have you been able to successfully run this on your own data set, and did the schema compare to David's actual table properly?

pkg/jobrunaggregator/jobruntablecreator/table_creator.go
```diff
@@ -0,0 +1,37 @@
...
+		if err != nil {
+			return err
+		}
+		if err := jobRunTable.Create(ctx, &bigquery.TableMetadata{Schema: schema}); err != nil {
+			return err
+		}
+	} else {
+		fmt.Fprintf(os.Stderr, "table already exists: %s\n", jobrunaggregatorlib.JobRunTableName)
```
dgoodwin, line 33: I feel like this should not be an Stderr as virtually all executions would print this line. If that was in Stephen's module as well it should probably be changed as well, as this is not really an error just expected behavior.

dgoodwin (commented):
Couple little questions but looking good.

pkg/jobrunaggregator/jobruntablecreator/cmd.go
DennisPeriquet: When set to `false`, the help screen will show the `create-tables` command; I guess there's no reason to hide it.

pkg/jobrunaggregator/jobruntablecreator/table_creator.go
DennisPeriquet: yes.  I also added commands I used in the README.md.  Also, added a comment about inferring the schema (without field comments) where the schema/struct is defined.

cmd/job-run-aggregator/README.md
```diff
@@ -5,4 +5,37 @@ The analysis allows failures within (ideally) a standard deviation of the norm f
...
+```
+cd ci-tools
+go build -gcflags='-N -l' `grep "module " go.mod |awk '{print $2}'`/cmd/job-run-aggregator
+```
+
+Example command lines:
+
+```
+# Run in dry run mode with read credential
+dlv exec ./job-run-aggregator -- upload-test-runs --dry-run --bigquery-dataset ci_data --google-service-account-credential-file ~/project-reader.json
```
dgoodwin, line 28: And I learned something today. :) Never really used delve.

pkg/jobrunaggregator/releasebigqueryloader/table_creator.go
```diff
@@ -29,7 +29,7 @@ func (r *allReleaseTableCreatorOptions) Run(ctx context.Context) error {
 			return err
 		}
 	} else {
-		fmt.Fprintf(os.Stderr, "table already exists: %s\n", jobrunaggregatorlib.ReleaseTableName)
+		fmt.Fprintf(os.Stdout, "table already exists: %s\n", jobrunaggregatorlib.ReleaseTableName)
```
dgoodwin, line 32: I'm less sure about my advice now, I thought this was like an idempotent thing that was run every time we used the main commands. If it's a separate sub-command, stderr might make some sense. In any case if nobody objects stdout is still probably good. Apologies for my confusion.

pkg/jobrunaggregator/jobruntablecreator/table_creator.go
```diff
@@ -0,0 +1,37 @@
...
+
+type allJobRunTableCreatorOptions struct {
+	ciDataClient jobrunaggregatorlib.CIDataClient
+	ciDataSet    *bigquery.Dataset
+}
+
+func (r *allJobRunTableCreatorOptions) Run(ctx context.Context) error {
+
+	// Create JobRunTable
+	jobRunTable := r.ciDataSet.Table(jobrunaggregatorlib.JobsTableName)
```
dgoodwin, line 22: Are we making the jobRunTable or the jobsTable? Some confusion in this PR referencing jobRuns when I think you're making jobs.

pkg/jobrunaggregator/jobruntablecreator/cmd.go
```diff
@@ -0,0 +1,93 @@
+package jobruntablecreator
```
dgoodwin, line 1: Per previous comment, jobtablecreator is probably more accurate.

pkg/jobrunaggregator/jobruntablecreator/table_creator.go
DennisPeriquet: hm... you're right.  Let me think about how to change that to be clearer because I'm hoping to also create the `JobRuns` table soon.

pkg/jobrunaggregator/releasebigqueryloader/table_creator.go
DennisPeriquet: no worries.  I figure since these commands are run once to build up tables, stdout is fine.

pkg/jobrunaggregator/jobruntablecreator/table_creator.go
DennisPeriquet: I renamed things to reflect that we're generating tables (now JobRuns along with Jobs).

dgoodwin:
/lgtm

Nice work!

deads2k:
the approach seems ok, I haven't dug into every line

/approve

rules
1. don't auto-create tables if they are missing
2. table creation must include descriptions for every field

suggestions
1. we create many copies of several of the tables, I suggest making a structure to map multiple table names to one schema
2. write some code to diff the existing table against the desired definition

DennisPeriquet:
I created these two issues to track the suggestions above: https://issues.redhat.com/browse/TRT-75, https://issues.redhat.com/browse/TRT-76

PR #2588, opened 2022-01-04
Add creation timestamps to records

stbenjam:
This adds timestamps to row in bigquery. The method of relying on generated ID's from bigquery doesn't seem reliable when exporting to postgres, so I want to base it off creation time stamps instead.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
@@ -48,6 +48,9 @@ type ReleaseRow struct {
 
 	// OSDiffURL is a link to the release page diffing the two OS versions.
 	OSDiffURL string `bigquery:"osDiffURL"`
+
+	// CreatedAt contains a timestamp for when this record was created in BigQuery.
+	CreatedAt time.Time `bigquery:"createdAt" json:"-"`
 }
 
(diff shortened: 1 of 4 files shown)


stbenjam:
Note: If we're ok with this change, I'll have to update the schemas manually before this merges.
/hold

stbenjam:
/hold cancel
/retest-required

This should be ready, the fields are in the schema now

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_release.go
```diff
@@ -48,6 +48,9 @@ type ReleaseRow struct {
 
 	// OSDiffURL is a link to the release page diffing the two OS versions.
 	OSDiffURL string `bigquery:"osDiffURL"`
+
+	// CreatedAt contains a timestamp for when this record was created in BigQuery.
+	CreatedAt time.Time `bigquery:"createdAt" json:"-"`
```
deads2k, line 53: `insertedAt` instead of `createdAt` so that even without reading docs, it's clear that this is about the insertion of the row, not the creation of a thing.

Also, the field documentation needs to warn consumers that the value isn't unique and multiple rows can be inserted with the same `insertedAt`.  In general, uploads for other tables run in parallel, so given that this is not late binding you're likely to end up with moving back in time too.

PR #2591, opened 2022-01-05
Add command to create "TestRuns" table with TestSuite; insert TestSuite as new column

DennisPeriquet:
Before merging this PR, we need to manually add a new column to "TestRuns" (or there will be a lot of errors due to missing field).

We need the TestSuite field to help distinguish test cases that have identical names.  For now, we'll collect the TestSuite name.  There should only be 1 TestSuite name but if there are more, we'll concatenate them to make a unique TestSuite.

I added some hints for debugging to facilitate testing of the code (on a smaller scale).  I confirmed there is an error (as expected) if we try to run this with a TestRuns table with no TestSuite column.  I also confirmed that if you manually add the column, and rerun, the error will go away and the TestSuite value will be populated.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
@@ -1,0 +1,52 @@
+package jobrunaggregatorapi
+
+const (
+	TestRunTableName = "TestRuns"
+
+	// The TestRunsSchema below is used to build the "TestRuns" table.
+	//
+	TestRunsSchema = `
+[
+  {
+    "mode": "REQUIRED",
+    "name": "Name",
+    "description" : "Name of the test run",
+    "type": "STRING"
+  },
+  {
+    "mode": "REQUIRED",
+    "name": "JobRunName",
+    "description" : "Name of the JobRun (big number) that ran this test (e.g., 1389486541524439040)",
+    "type": "STRING"
+  },
+  {
+    "mode": "REQUIRED",
+    "name": "JobName",
+    "description" : "Name of the Job that as this test in it",
+    "type": "STRING"
+  },
+  {
+    "mode": "REQUIRED",
+    "name": "Status",
+    "description" : "Status of the test (e.g., pass, fail)",
+    "type": "STRING"
+  },
+  {
+    "mode": "NULLABLE",
+    "name": "TestSuite",
+    "description" : "Testsuite that this test belongs to",
+    "type": "STRING"
+  }
+]
+`
+)
+
+// Move here from jobrunbigqueryloader/types.go
+//
+type TestRunRow struct {
+	Name       string
+	JobRunName string
+	JobName    string
+	Status     string
+	TestSuite  string
+}
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
@@ -9,6 +9,8 @@ import (
 	"strings"
 
 	"github.com/spf13/pflag"
+
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 )
 
 const (
@@ -92,8 +94,25 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "BULK INSERT into %v\n", d.table)
 	for i := 0; i < srcVal.Len(); i++ {
-		s := srcVal.Index(i).Interface()
-		fmt.Fprintf(buf, "\tINSERT into %v: %v\n", d.table, s)
+
+		switch s := srcVal.Index(i).Interface().(type) {
+		case *jobrunaggregatorapi.TestRunRow:
+			fmt.Fprintf(buf, "\tINSERT into %v: %#v\n", d.table, s)
+
+		case *jobrunaggregatorapi.JobRunRow:
+			fmt.Fprintf(buf, "\tINSERT into %v: name=%v, jobname=%v, status=%v\n", d.table, s.Name, s.JobName, s.Status)
+
+		case *jobrunaggregatorapi.BackendDisruptionRow:
+			fmt.Fprintf(buf, "\tINSERT into %v: %#v\n", d.table, s)
+
+		case jobrunaggregatorapi.JobRow:
+			fmt.Fprintf(buf, "\tINSERT into %v: JobName=%v\n", d.table, s.JobName)
+
+		default:
+
+			// If we don't know the type, output something generic.
+			fmt.Fprintf(buf, "\tINSERT into %v: %#v\n", d.table, s)
+		}
 	}
 	fmt.Fprint(d.out, buf.String())
 
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
@@ -42,6 +42,9 @@ func (o *ciGCSClient) ListJobRunNamesOlderThanFourHours(ctx context.Context, job
 	if err := query.SetAttrSelection([]string{"Name", "Created"}); err != nil {
 		return nil, nil, err
 	}
+
+	// When debugging, you can set the starting ID to a number such that you
+	// will process a relatively small number of jobsRuns.
 	query.StartOffset = fmt.Sprintf("logs/%s/%s", jobName, startingID)
 	fmt.Printf("  starting from %v\n", query.StartOffset)
 
--- a/pkg/jobrunaggregator/jobrunbigqueryloader/test_run_uploader.go
+++ b/pkg/jobrunaggregator/jobrunbigqueryloader/test_run_uploader.go
@@ -3,8 +3,7 @@ package jobrunbigqueryloader
 import (
 	"context"
 	"fmt"
-
-	"cloud.google.com/go/bigquery"
+	"strings"
 
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 
@@ -43,7 +42,7 @@ func (o *testRunUploader) uploadTestSuites(ctx context.Context, jobRun jobrunagg
 	return nil
 }
 
-func (o *testRunUploader) uploadTestSuite(ctx context.Context, jobRun jobrunaggregatorapi.JobRunInfo, prowJob *prowv1.ProwJob, parentSuites []string, suite *junit.TestSuite) error {
+func (o *testRunUploader) uploadTestSuite(ctx context.Context, jobRun jobrunaggregatorapi.JobRunInfo, prowJob *prowv1.ProwJob, parentSuites []string, suite *junit.TestSuite) error { //nolint
 	currSuites := append(parentSuites, suite.Name)
 	for _, testSuite := range suite.Children {
 		if err := o.uploadTestSuite(ctx, jobRun, prowJob, currSuites, testSuite); err != nil {
@@ -51,13 +50,25 @@ func (o *testRunUploader) uploadTestSuite(ctx context.Context, jobRun jobrunaggr
 		}
 	}
 
-	toInsert := []bigquery.ValueSaver{}
+	toInsert := []*jobrunaggregatorapi.TestRunRow{}
 	for i := range suite.TestCases {
 		testCase := suite.TestCases[i]
 		if testCase.SkipMessage != nil {
 			continue
 		}
-		toInsert = append(toInsert, newTestRunRow(jobRun, prowJob, currSuites, testCase))
+
+		var status string
+		switch {
+		case testCase.FailureOutput != nil:
+			status = "Failed"
+		case testCase.SkipMessage != nil:
+			status = "Skipped"
+		default:
+			status = "Passed"
+		}
+
+		testSuiteStr := strings.Join(currSuites, "|||")
+		toInsert = append(toInsert, newTestRunRow(jobRun, status, testSuiteStr, testCase))
 	}
 	if err := o.testRunInserter.Put(ctx, toInsert); err != nil {
 		return err
--- a/pkg/jobrunaggregator/jobrunbigqueryloader/types.go
+++ b/pkg/jobrunaggregator/jobrunbigqueryloader/types.go
@@ -3,8 +3,6 @@ package jobrunbigqueryloader
 import (
 	"time"
 
-	"cloud.google.com/go/bigquery"
-
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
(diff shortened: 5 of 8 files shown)


pkg/jobrunaggregator/jobrunbigqueryloader/types.go
```diff
@@ -46,8 +47,12 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "BULK INSERT into %v\n", d.table)
 	for i := 0; i < srcVal.Len(); i++ {
-		s := srcVal.Index(i).Interface()
-		fmt.Fprintf(buf, "\tINSERT into %v: %v\n", d.table, s)
+		s := srcVal.Index(i).Interface().(*testRunRow)
```
deads2k, line 50: I would rather see this structured as a registry of types that can added to the dryrunInserter instance.

deads2k, line 50: actually, does this even work if its not a testRunRow?

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
```diff
@@ -105,11 +110,20 @@ func (v *testRunRow) Save() (map[string]bigquery.Value, string, error) {
 		status = "Passed"
 	}
 
+	var testSuite_str string
```
deads2k, line 113: This was the first inserter I ever made.  Waht do you think about making this a struct like the other "normal" ones now?

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
```diff
@@ -105,11 +110,20 @@ func (v *testRunRow) Save() (map[string]bigquery.Value, string, error) {
...
+	if len(v.testSuites) == 1 {
+		testSuite_str = v.testSuites[0]
+	} else {
+		// There is generally a single test suite.  But if there are more, we will
+		// concat them to make a unique test suite name.
+		testSuite_str = strings.Join(v.testSuites, "|")
```
deads2k, line 119: I'm ok starting with this. I'm also ok choosing more than one character as the limiter (I would probably choose at least two characters, like `|||`).

pkg/jobrunaggregator/tablescreator/table_creator.go
```diff
@@ -48,5 +48,19 @@ func (r *allJobsTableCreatorOptions) Run(ctx context.Context) error {
 		fmt.Fprintf(os.Stdout, "table already exists: %s\n", jobrunaggregatorlib.JobRunTableName)
 	}
 
+	// Create TestRunsTable
+	testRunsTable := r.ciDataSet.Table(jobrunaggregatorlib.TestRunTableName)
```
deads2k, line 52: We now have several views on top of this table.  How would you like to handle those?  Would you consider an approach that has a map of table name to schema definition and have this command create from that?

DennisPeriquet: Let me see if I follow: make `map[string]string` (first string is tableName, second string is Schema represented in Json string), modify the function to create all the tables using `range` over the map keys? -- does that sound right?

I think existing views on top of `TestRuns` table would remain intact and unaffected (because today, they don't depend on the new `TestSuite` field).

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
DennisPeriquet: ah, you mean like how you did [JobRunRow](https://github.com/openshift/ci-tools/blob/master/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_jobrun.go#L57)?

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -92,8 +95,36 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
...
+		case jobrunaggregatorapi.JobsTableName:
+			s := srcVal.Index(i).Interface().(jobrunaggregatorapi.JobRow)
+			fmt.Fprintf(buf, "\tINSERT into %v: JobName=%v\n", d.table, s.JobName)
+
+		default:
+
+			// Today, NewDryRunInserter is created for the above types.  If there are more,
+			// abort so that we know to add another case block above.
+			fmt.Println("Unknown table in dry-run mode")
+			os.Exit(1)
```
deads2k, line 126: don't exit from library code.  Given the intent of this code, simply writing the error will do.

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -92,8 +95,36 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "BULK INSERT into %v\n", d.table)
 	for i := 0; i < srcVal.Len(); i++ {
-		s := srcVal.Index(i).Interface()
-		fmt.Fprintf(buf, "\tINSERT into %v: %v\n", d.table, s)
+
+		// Infer the struct type from the table name
+		switch d.table {
+		case jobrunaggregatorapi.TestRunTableName:
+			s := srcVal.Index(i).Interface().(*jobrunaggregatorapi.TestRunRow)
```
deads2k, line 102: can you type switch on the .Interface()?

pkg/jobrunaggregator/jobrunbigqueryloader/types.go
DennisPeriquet: this is addressed by changing to using type switch

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
```diff
@@ -0,0 +1,97 @@
...
+	"strings"
+
+	"cloud.google.com/go/bigquery"
+
+	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
+
+	"github.com/openshift/ci-tools/pkg/junit"
+)
+
+// The TestRunSchema below is used to build the "TestRuns" table.
```
dgoodwin, line 13: Per other PR this should move down to right above the TestRunSchema variable, and read "TestRunSchema is used to build the TestRuns table."

dgoodwin: Suggest renaming to match the table, pluralizing Run(s)

pkg/jobrunaggregator/jobrunaggregatorlib/gcs_client.go
```diff
@@ -42,6 +42,9 @@ func (o *ciGCSClient) ListJobRunNamesOlderThanFourHours(ctx context.Context, job
 	if err := query.SetAttrSelection([]string{"Name", "Created"}); err != nil {
 		return nil, nil, err
 	}
+
+	// When debugging, you can set the starting ID to a number such that you
+	// will process a relatively small number of jobsRuns.
```
dgoodwin, line 47: Nice idea adding these tips. If we could get these wired up to CLI options or env vars in the future that would be great.

pkg/jobrunaggregator/tablescreator/table_creator.go
dgoodwin: My understanding of David's comment is more specific to views, which as far as I know, are not created by code today, only by hand. I think he was asking about what we can do to change that, probably in followup PRs.

dgoodwin (commented):
Looks just about ready, just a godoc nits.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
```diff
@@ -0,0 +1,97 @@
...
+		status = "Failed"
+	case v.TestCase.SkipMessage != nil:
+		status = "Skipped"
+	default:
+		status = "Passed"
+	}
+
+	var testSuite_str string
+	if len(v.TestSuites) == 1 {
+		testSuite_str = v.TestSuites[0]
```
dgoodwin, line 82: Sorry, couple more, go vars should be testSuiteStr. See https://go.dev/doc/effective_go#mixed-caps

That whole document is pretty good to read.

DennisPeriquet:
TestSuite field is added: 

TestRuns_TestSuite_column

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -92,8 +94,34 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
...
+
+		case *jobrunaggregatorapi.JobRunRow:
+			s := srcVal.Index(i).Interface().(*jobrunaggregatorapi.JobRunRow)
+			fmt.Fprintf(buf, "\tINSERT into %v: name=%v, jobname=%v, status=%v\n", d.table, s.Name, s.JobName, s.Status)
+
+		case *jobrunaggregatorapi.BackendDisruptionRow:
+			s := srcVal.Index(i).Interface().(*jobrunaggregatorapi.BackendDisruptionRow)
+			fmt.Fprintf(buf, "\tINSERT into %v: BEName=%v, JobRunName=%v, DisSecs=%v\n", d.table, s.BackendName, s.JobRunName, s.DisruptionSeconds)
+
+		case jobrunaggregatorapi.JobRow:
```
dgoodwin, line 115: Hmm, should this be a pointer like the others?

DennisPeriquet: Good question. I found that one is actually not a pointer.  [This](https://github.com/openshift/ci-tools/blob/541d29e664365b43a93aa46f8c0c916cb460f5f4/pkg/jobrunaggregator/jobtableprimer/create_jobs.go#L44) shows that for Jobs, it's called with a slice of JobRows and not slice of pointers like the others.

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -92,8 +94,34 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "BULK INSERT into %v\n", d.table)
 	for i := 0; i < srcVal.Len(); i++ {
-		s := srcVal.Index(i).Interface()
-		fmt.Fprintf(buf, "\tINSERT into %v: %v\n", d.table, s)
+
+		switch srcVal.Index(i).Interface().(type) {
```
deads2k, line 98: you should be able to assign it up here so you don't have to type assert in every case below.

`switch t := foo` I think.

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
deads2k: I think that's an indication that you want to handle both the pointer and non-pointer cases.

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -92,8 +94,34 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
...
+
+		case jobrunaggregatorapi.JobRow:
+			s := srcVal.Index(i).Interface().(jobrunaggregatorapi.JobRow)
+			fmt.Fprintf(buf, "\tINSERT into %v: JobName=%v\n", d.table, s.JobName)
+
+		default:
+
+			// If we don't know the type, output something generic.
+			s := srcVal.Index(i).Interface()
+			fmt.Fprintf(buf, "\tINSERT into %v: %v\n", d.table, s)
```
deads2k, line 123: is `%#v` better for this use-case?

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
```diff
@@ -0,0 +1,98 @@
...
+    "type": "STRING"
+  }
+]
+`
+)
+
+// Move here from jobrunbigqueryloader/types.go
+//
+type TestRunRow struct {
+	ProwJob    *prowv1.ProwJob
```
deads2k, line 57: `*Row` types should include only the types that get stored.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
```diff
@@ -0,0 +1,98 @@
...
+	JobRun     JobRunInfo
+	TestSuites []string
+	TestCase   *junit.TestCase
+}
+
+// Ensure (at compile time) that testRunRow implements the bigquery.ValueSaver interface
+var _ bigquery.ValueSaver = &TestRunRow{}
+
+func (v *TestRunRow) Save() (map[string]bigquery.Value, string, error) {
```
deads2k, line 66: Avoid using our API types as method receivers. It seems convenient, but it causes deeper package dependencies for APIs we want to rely on elsewhere and historically has caused diamond dependency resolution problems.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_testrun.go
deads2k: > *Row types should include only the types that get stored.

If you adjust this to match, then you can remove the Save method below and use it for reading as well.

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -92,8 +94,34 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
...
+				fmt.Println("Found a testSuite list greater than one")
+			}
+
+		case *jobrunaggregatorapi.JobRunRow:
+			s := srcVal.Index(i).Interface().(*jobrunaggregatorapi.JobRunRow)
+			fmt.Fprintf(buf, "\tINSERT into %v: name=%v, jobname=%v, status=%v\n", d.table, s.Name, s.JobName, s.Status)
+
+		case *jobrunaggregatorapi.BackendDisruptionRow:
+			s := srcVal.Index(i).Interface().(*jobrunaggregatorapi.BackendDisruptionRow)
+			fmt.Fprintf(buf, "\tINSERT into %v: BEName=%v, JobRunName=%v, DisSecs=%v\n", d.table, s.BackendName, s.JobRunName, s.DisruptionSeconds)
```
deads2k, line 113: does `%#v` not do what you want?

pkg/jobrunaggregator/tablescreator/table_creator.go
deads2k, line 52: Consider a structure like

```go
TableNamesToSchemas := map[string]string{
  jobrunaggregatorlib.TestRunTableName: jobrunaggregatorapi.TestRunsSchema,
  jobrunaggregatorlib.JobRunTableName: jobrunaggregatorapi.JobRunSchema
}

```

And then this command iterates over it creating each table.

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
DennisPeriquet: ah, it does -- thx

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
DennisPeriquet: ah nice -- thanks!

pkg/jobrunaggregator/tablescreator/table_creator.go
DennisPeriquet: that works out really well -- thanks!

pkg/jobrunaggregator/jobrunbigqueryloader/test_run_uploader.go
```diff
@@ -51,13 +49,34 @@ func (o *testRunUploader) uploadTestSuite(ctx context.Context, jobRun jobrunaggr
...
+		case testCase.FailureOutput != nil:
+			status = "Failed"
+		case testCase.SkipMessage != nil:
+			status = "Skipped"
+		default:
+			status = "Passed"
+		}
+
+		var testSuiteStr string
+		testSuiteStr = suite.Name
```
deads2k, line 70: it's the test suite logically

`testSuiteStr := strings.Join(currSuites, "|||")` ?

This looks like it would assign the current test case the value of all the child suites.

deads2k (changes requested):

pkg/jobrunaggregator/jobrunbigqueryloader/test_run_uploader.go
```diff
@@ -51,13 +50,33 @@ func (o *testRunUploader) uploadTestSuite(ctx context.Context, jobRun jobrunaggr
...
+			status = "Skipped"
+		default:
+			status = "Passed"
+		}
+
+		testSuiteStr := strings.Join(currSuites, "|||")
+		if suite.Children != nil {
+			// There is generally a single test suite.  But if there are more, we will
+			// concat them to make a unique test suite name.
+			for _, childTestSuite := range suite.Children {
```
deads2k, line 74: why do we add the children of this test suite to the test case.  it logically looks like

```yaml
suiteName: foo
- childSuiteName: bar
  - testCaseName: the-test
  - childSuiteName: trip
    - testCaseName:  smaller-test
  - childSuiteName: fourth
    - testCaseName:  now-a-test
```

we want
1. foo|||bar the-test
2. foo|||bar|||trip smaller-test
3. foo|||bar|||four now-a-test

but this code gives me
1. foo||bar||trip||four the-test
2. foo|||bar|||trip smaller-test
3. foo|||bar|||four now-a-test

deads2k:
/hold

I'm hearing reports of a failure.

PR #2595, opened 2022-01-11
Add a new tool: sync-rover-groups

hongkailiu:
/cc @openshift/test-platform


petr-muller:
Let me know when you want reviews on this one :+1:

hongkailiu:
> Let me know when you want reviews on this one 👍

Yes Sir.
Currently this is my own PoC for writing up the doc and I will let you know when it is ready for review.

hongkailiu:
@petr-muller 
this one is ready for review.

/hold

I will make a PR for the changes of the deployment

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -6,14 +6,14 @@ import (
...
 	"os"
-	"strings"
 
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
```
petr-muller, line 16: I am surprised that this PR also changes the `github-ldap-user-group-creator` and I don't know why. Could you accompany the PR with the description about what is the new tool supposed to do, and why is the PR accompanied with these seemingly unrelated changes?

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -42,6 +43,7 @@ func parseOptions() *options {
 	fs.StringVar(&opts.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
 	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
 	fs.StringVar(&opts.mappingFile, "mapping-file", "", "File to the mapping results of m(github_login)=kerberos_id.")
+	fs.StringVar(&opts.groupsFile, "groups-file", "", "The file to store the groups")
```
petr-muller, line 46: Is this an output of the tool? An input? What is the format of the file?

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -142,12 +148,46 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to load the mapping")
 	}
 
-	if err := ensureGroups(ctx, clients, mapping, opts.dryRun); err != nil {
+	data, err := ioutil.ReadFile(opts.groupsFile)
+	if err != nil {
+		logrus.WithError(err).Fatal("Failed to read the group file")
+	}
+	roverGroups := map[string][]string{}
```
petr-muller, line 155: Using a list of strings is prone to e.g. containing duplicate names and other mistakes (empty string is a string too). I'd add at least basic validation of the input:

- duplicate items
- both group names and member names should be checked to be valid names

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -142,12 +148,46 @@ func main() {
...
+			Users:      userv1.OptionalNames{githubLogin, kerberosId},
+		})
+	}
+	for k, v := range roverGroups {
+		groupsNamesToCreate.Insert(k)
+		groupsToCreate = append(groupsToCreate, userv1.Group{
+			ObjectMeta: metav1.ObjectMeta{Name: k, Labels: map[string]string{api.DPTPRequesterLabel: toolName}},
+			Users:      v,
+		})
+	}
```
petr-muller, line 189: I think that this method ends up doing too much: figuring out which groups should exist from two separate sources, and then creating them.

How about splitting the three actions into three separate methods, each with much simpler interface, called straight from the `main()` level? Somehting like:

```go
func makeGitHubMappingGroups(mapping) map[string]*userv1.Group {...}
func makeRoverGroups(roverGroups) map[string]*userv1.Group {...}

func ensureGroups(ctx, clients, dryRun, groups map[string]*userv1.Group) error {
  groupNames := sets.NewString()
  for name := groups {
    groupNames.Insert(name)
  }
  ...
}

func main() {
...
  createGroups := map[string]*userv1.Group
  // iterate over sources of groups to create
  for groups := range []map[string]*userv1.Group { makeGitHubMappingGroups(mappings), makeRoverGroups(roverGroups) } {
    for name, group := range groups {
      if _, exists := createGroups[name]; exists { 
        // error out, duplicate group in rover and mapping
      }
      createGroups[name] = group 
    }
  }
}
```

petr-muller (commented):
Do you think you could give both `github-ldap-user-group-creator` and `sync-rover-groups` a README that follows the format I recently started writing for some of our other tools:

- What it does
- Why it exists
- How it works
- How is it deployed

See https://github.com/openshift/ci-tools/blob/master/cmd/autoperibolossync/README.md for an example. The intended audience is for example a new hire after a month. My intent is for each of our tools to have an unified, very brief record with useful links to other resources.

petr-muller:
@hongkailiu I was not able to look at the actual `sync-rover-groups` code yet, but I had some items to address even after walking thought the ldap-creator changes, so I submitted them so that you have at least some feedback. I'll look at the new tool tomorrow.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: These questions have been addressed in https://docs.google.com/document/d/1Q003WYYHpzKj5y232QLW-BgA5pTrDPXRjWShCesWlc8/edit#

I will update the readme after our arch call.

petr-muller: The fact that these questions were addressed somewhere else is not entirely useful. I find it reasonable to require large PRs like this to carry one or two paragraph description of the change. For you it may seem like repeating the content from elsewhere, but PR descriptions is the thing that the reviewers will read, not the google document.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: Done.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: Done.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: I mixed up two things in the comments: PR description and ReadMe for the tools.
I missed the former (my mistake) and wanted to added the latter after arch meeting (in case something is changed there).

Let me add both now to make the review easier.

cmd/sync-rover-groups/README.md
```diff
@@ -0,0 +1,22 @@
...
+`sync-rover-groups` collects the groups in the manifests and resolves their members
+and save resolved groups in a file.
+
+## How it works
+
+`sync-rover-groups` resolve the group by querying the Red Hat LDAP server.
+
+## How is it deployed
+
+The cronjob [sync-rover-groups-update](https://console-openshift-console.apps.ocp-c1.prod.psi.redhat.com/k8s/ns/ocp-test-platform/batch~v1~CronJob/sync-rover-groups-update) ([definition](http://TODO))
```
hongkailiu, line 21: The URL will be updated after https://github.com/openshift/release/pull/25357 is merged.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: Updated.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: Done.

cmd/sync-rover-groups/ldap.go
```diff
@@ -0,0 +1,80 @@
...
+
+func (e *notFoundErr) Error() string {
+	return fmt.Sprintf("failed to find group on the ldap server: %s", e.group)
+}
+
+func (r *ldapGroupResolver) resolve(name string) (*Group, error) {
+	filter := fmt.Sprintf("(&(objectClass=rhatRoverGroup)(cn=%s))", ldapv3.EscapeFilter(name))
+	searchReq := ldapv3.NewSearchRequest(baseDN, ldapv3.ScopeWholeSubtree, 0, 0, 0, false, filter, []string{"uniqueMember"}, []ldapv3.Control{})
+
+	result, err := r.conn.Search(searchReq)
```
petr-muller, line 46: Can `r.conn` be nil? If yes we should check to avoid nil poitner dereference. Even if this is checked in another place, it is better to be defensive and panic with an explicit error ("connection to LDAP is nil") rather than NPD. You'd guard against a potential bug.

cmd/sync-rover-groups/ldap.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if len(result.Entries) == 0 {
+		return nil, &notFoundErr{group: name}
+	}
+
+	if l := len(result.Entries); l > 1 {
+		// this should never happen
+		return nil, fmt.Errorf("found %d group with the name: %s", l, name)
+	}
```
petr-muller, line 58: these both deal with len() so we could do:

```go
switch items := len(); {
case items == 0:
  return something
case items > 1:
  return something else
}
```

cmd/sync-rover-groups/main.go
```diff
@@ -0,0 +1,164 @@
...
+	ldapServer       string
+	validateSubjects bool
+	groupsFile       string
+}
+
+func parseOptions() *options {
+	opts := &options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&opts.logLevel, "log-level", "info", fmt.Sprintf("Log level is one of %v.", logrus.AllLevels))
+	fs.Var(&opts.manifestDirRaw, "manifest-dir", "directory containing Kubernetes manifests. Can be specified multiple times.")
```
petr-muller, line 39: Iterating over manifests to figure out the groups that should exist is smart, I like how it avoids a necessity to maintain a separate config file or something similar. Two concerns:

- Do we expect the need to support groups that are supposed to have different names in Rover and in Kubernetes (are the two naming rules fully compatible)?
- Do we expect the need to support certain groups to only exist on certain clusters? I imagine this is a legit request for e.g. the Hive cluster which will need to provide access to groups to their pools, but these groups are not necessary on normal build farm clsuters (also, app.ci is probably unique too).

In my head, I was expecting a config file to drive these use cases, but maybe they are not possible.

---

As an entirely wild thought: maybe this should be a controller? Imagine the following:
- We would have a CRD called `RoverGroup`. Creating such resource in a cluster would represent the intent of creating a `Group` of the same name, and maintaining its members to be equivalent to the Rover Group
- We would have a controller running in the PSI cluster that would watch `RoverGroup` events on our clusters and would maintain a matching `Group` in the same cluster (by reaching to LDAP, getting the members list, etc). If the `RoverGroup` would be deleted, then it would delete the matching `Group` 
- The controller would also periodically poll LDAP for members of the groups it maintains on the cluster and updated the menbers in clusters accordingly. 

The above ^^^ is just a wild thought. Please do not immediately jump and implement it unless you really like it. It also does not solve the groups for the GitHub<->RH pairing.

cmd/sync-rover-groups/main.go
```diff
@@ -0,0 +1,164 @@
...
+
+	templatev1 "github.com/openshift/api/template/v1"
+	userv1 "github.com/openshift/api/user/v1"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+type options struct {
+	logLevel         string
+	manifestDirRaw   flagutil.Strings
```
petr-muller, line 28: if it's a set then it should be called `manifestDirsRaw` (plural)

cmd/sync-rover-groups/main.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+		groupNames.Insert(names.List()...)
+	}
+
+	groupNames.Insert(api.CIAdminsGroupName)
+
+	groups := map[string][]string{}
+	for _, name := range groupNames.List() {
+		logrus.WithField("group", name).Debug("resolving group ...")
+		g, err := groupResolver.resolve(name)
```
petr-muller, line 143: Could we reasonably parallelize? I imagine this could get slow on I/O if we have more groups.

cmd/sync-rover-groups/yamlgroupcollector.go
```diff
@@ -0,0 +1,206 @@
...
+
+const yamlSeparator = "\n---"
+
+func (c *yamlGroupCollector) collect(dir string) (sets.String, error) {
+	groups := sets.NewString()
+	abs, err := filepath.Abs(dir)
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine absolute path for dir %s: %w", dir, err)
+	}
+	if err := filepath.Walk(abs,
```
petr-muller, line 37: This whole thing below is IMO a bit hard to read, too long and too many abstraction levels (indents and nested cycles). Some proposals inline about bits that could perhaps be extracted into separate methods.

cmd/sync-rover-groups/yamlgroupcollector.go
```diff
@@ -0,0 +1,206 @@
...
+			if skip, err := fileFilter(info, path); skip || err != nil {
+				return err
+			}
+
+			data, err := ioutil.ReadFile(path)
+			if err != nil {
+				logrus.WithField("source-file", path).WithError(err).Error("Failed to read file")
+				return err
+			}
+			for _, yamlDoc := range bytes.Split(data, []byte(yamlSeparator)) {
```
petr-muller, line 53: Could this block be extracted into something like `processYamlDoc` method to avoid deep nesting?

cmd/sync-rover-groups/yamlgroupcollector.go
```diff
@@ -0,0 +1,206 @@
...
+				if err != nil {
+					if runtime.IsNotRegisteredError(err) && (!c.validateSubjects ||
+						!strings.HasPrefix(err.Error(), "no kind \"Group\" is registered for version \"v1\"")) {
+						continue
+					}
+					logrus.WithField("source-file", path).WithError(err).Error("Failed to decode yaml file")
+					return err
+				}
+				switch o := obj.(type) {
+				case *v1.List:
```
petr-muller, line 67: Could this block be extracted to soething like `processList`?

cmd/sync-rover-groups/yamlgroupcollector.go
```diff
@@ -0,0 +1,206 @@
...
+							return err
+						}
+						more, err := primitive(subObj, c.validateSubjects)
+						if err != nil {
+							logrus.WithField("source-file", path).WithError(err).Error("Failed to get groups")
+							return err
+						}
+						groups = groups.Union(more)
+					}
+				case *templatev1.Template:
```
petr-muller, line 85: Could this block be extracted to soething like `processTemplate`?

cmd/sync-rover-groups/yamlgroupcollector.go
```diff
@@ -0,0 +1,206 @@
...
+					for _, object := range o.Objects {
+						subObj, _, err := c.decoder.Decode(object.Raw, nil, nil)
+						if err != nil {
+							if runtime.IsNotRegisteredError(err) && (!c.validateSubjects ||
+								!strings.HasPrefix(err.Error(), "no kind \"Group\" is registered for version \"v1\"")) {
+								continue
+							}
+							// Trying to avoid reinventing oc-process
```
petr-muller, line 93: This is fine. We do not need to be super-proof. Best reasonable effort is fine.

cmd/sync-rover-groups/yamlgroupcollector_test.go
```diff
@@ -0,0 +1,84 @@
...
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			r := newYamlGroupCollector(tc.validateSubjects)
+			actual, actualErr := r.collect(tc.dir)
+			if diff := cmp.Diff(tc.expected, actual, testhelper.RuntimeObjectIgnoreRvTypeMeta); diff != "" {
+				t.Errorf("%s differs from expected:\n%s", tc.name, diff)
+			}
+			if diff := cmp.Diff(tc.expectedErr, actualErr, testhelper.EquateErrorMessage); diff != "" {
```
petr-muller, line 79: I'd say that errors should be checked first and when they differ, they should cause to *not* check the actual outputs? When the errors are different (especially when we expect nil but some error is returned) then comparing actual output values is IMO just noise in the output.

cmd/sync-rover-groups/yamlgroupcollector.go
```diff
@@ -0,0 +1,206 @@
...
+			}
+			if s.Kind == "Group" {
+				if strings.HasPrefix(s.Name, "system:") {
+					continue
+				}
+				if strings.Contains(s.Name, "${") {
+					return nil, fmt.Errorf("cannot use ${ in a subject of ClusterRoleBinding: %s", o.Name)
+				}
+				groups.Insert(s.Name)
+			}
```
petr-muller, line 171: Can these two blocks share code? Looks like the only difference is the `{Cluster}RoleBinding` bit in the errors...

cmd/github-ldap-user-group-creator/README.md
```diff
@@ -1,18 +1,29 @@
...
+  in the release repo.
+
+## How it works
+
+`github-ldap-user-group-creator` reads 
+- the mapping files generated by [another tool](../ldap-users-from-github-owners-files) `ldap-users-from-github-owners-files`
+that stores the mapping from `github-id` to its Red Hat `kerberos-id` and for each `github-id`, creates a group `github-id-group`
+on each cluster.
+
+- the groups file generated by [another tool](../syn-rover-groups) that stores the group names and their members from
```
smg247, line 22:
```suggestion
- the groups file generated by the [sync-rover-groups](../sync-rover-groups) tool that stores the group names and their members from 
```

cmd/github-ldap-user-group-creator/README.md
```diff
@@ -1,18 +1,29 @@
...
+that stores the mapping from `github-id` to its Red Hat `kerberos-id` and for each `github-id`, creates a group `github-id-group`
+on each cluster.
+
+- the groups file generated by [another tool](../syn-rover-groups) that stores the group names and their members from 
+the Red Hat LDAP server and for each group creates a group on each cluster.
+
+## How is it deployed
+
+The periodic
+job [periodic-github-ldap-user-group-creator](https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/?job=periodic-github-ldap-user-group-creator) ([definition](https://github.com/openshift/release/blob/d5ac4fc821a8c366a02ce47cd47e3d774c98497c/ci-operator/jobs/infra-periodics.yaml#L2157-L2199))
```
smg247, line 28: I don't know that there is much value in adding the exact lines of the job definition. That seems likely to change a ton and get immediately out of date. Maybe just a link to the `infra-periodics.yaml` file saying it lives in there?

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -142,12 +148,60 @@ func main() {
...
+	if err := ensureGroups(ctx, clients, groups, opts.dryRun); err != nil {
 		logrus.WithError(err).Fatal("could not ensure groups")
 	}
 }
 
-func ensureGroups(ctx context.Context, clients map[string]ctrlruntimeclient.Client, mapping map[string]string, dryRun bool) error {
+func makeGitHubMappingGroups(mapping map[string]string) map[string]*userv1.Group {
+	groups := map[string]*userv1.Group{}
+	for githubLogin, kerberosId := range mapping {
+		groupName := fmt.Sprintf("%s%s", githubLogin, api.GroupSuffix)
```
smg247, line 185: this contradicts with your README. I would expect it to be 
```suggestion
		groupName := fmt.Sprintf("%s-%s", githubLogin, api.GroupSuffix)
```

cmd/sync-rover-groups/main.go
hongkailiu: The 2 cases are valid. The config file is a supplement of what cannot be covered by the auto-detection. What about this?

```yaml
some-rover-group:
  clusters: # default to all clusters
  - app.ci
  - build01
  name: # default to the one at rover 
```

About the controller: Functionality wise, it is the same as the current implementation. It is more efficient in the sense that it is realtime: Do we need realtime? The groups should not change that often. My concern is the maintenance of such a tool: we do not GitOps on psi and secrets there are out of control of ci-secret-bootstrap. The effort on implementing a controller watching multiple clusters is not trivial either.

I am not worrying about the groups for the GitHub<->RH pairing: The current implementation (https://github.com/openshift/ci-tools/tree/master/cmd/ldap-users-from-github-owners-files) can easily included in the sync-rover-groups (no matter to it is a simple binary or a controller).

cmd/sync-rover-groups/README.md
smg247: This ties into my previous comment, I don't think we need to provide exact lines in a file that changes often.

cmd/sync-rover-groups/main.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+	return opts
+}
+
+func (o *options) validate() error {
+	level, err := logrus.ParseLevel(o.logLevel)
+	if err != nil {
+		return fmt.Errorf("invalid log level specified: %w", err)
+	}
+	logrus.SetLevel(level)
```
smg247, line 54: I see that this is done elsewhere as well, but it's a bit misleading to set stuff in the `validate` method. I would rather the setting happen elsewhere and pure validations exist here.

cmd/sync-rover-groups/main.go
```diff
@@ -0,0 +1,164 @@
...
+
+	values := o.manifestDirRaw.Strings()
+	if len(values) == 0 {
+		return fmt.Errorf("--manifest-dir must be set")
+	}
+	o.manifestDirs = sets.NewString()
+	o.manifestDirs.Insert(values...)
```
smg247, line 61: see previous comment

cmd/sync-rover-groups/yamlgroupcollector_test.go
hongkailiu: I feel this way is easier to read than including the logic between output and error: treat them as a tuple, and we compare every element no matter what.

cmd/github-ldap-user-group-creator/README.md
hongkailiu: It is a permalink which is not going to be out of date.
https://docs.github.com/en/repositories/working-with-files/using-files/getting-permanent-links-to-files
The content of the job could be tough.

The link to infra job file could be changed as well (i know it is more unlikely).
I do not have a strong opinion on either of them: just we have got many of those permalinks already. :)

cmd/github-ldap-user-group-creator/main.go
hongkailiu: https://github.com/openshift/ci-tools/blob/02d9d4fac19ef0184c322ff9a261521f5a643782/pkg/api/constant.go#L57

Actually, the hyphen is a part of `api.GroupSuffix`.
Let me remove the confusion.

cmd/sync-rover-groups/README.md
```diff
@@ -0,0 +1,22 @@
+# sync-rover-groups
+
+## What it does
+
+`sync-rover-groups` is a tool to resolve the groups in [the manifests](https://github.com/openshift/release/tree/master/clusters) of CI clusters
+in the release repo.
```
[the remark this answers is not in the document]
hongkailiu: Updated.

cmd/sync-rover-groups/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+
+## Why it exists
+
+Only [rover-groups](https://rover.redhat.com/groups/) are allowed in the manifests.
+`sync-rover-groups` collects the groups in the manifests and resolves their members
+and saves the resolved groups in a file.
```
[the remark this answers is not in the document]
hongkailiu: Updated.

cmd/sync-rover-groups/ldap.go
hongkailiu: Updated.

cmd/sync-rover-groups/ldap.go
hongkailiu: Updated.

cmd/sync-rover-groups/yamlgroupcollector.go
hongkailiu: The current implementation is more succinct.
Maybe a new function is not that essential?

cmd/sync-rover-groups/yamlgroupcollector.go
hongkailiu: Same as `processList`.

cmd/sync-rover-groups/yamlgroupcollector.go
hongkailiu: Updated.

cmd/github-ldap-user-group-creator/README.md
hongkailiu: Done.

cmd/github-ldap-user-group-creator/main.go
hongkailiu: Updated.

cmd/sync-rover-groups/main.go
hongkailiu: Done.

cmd/sync-rover-groups/main.go
hongkailiu: Done.

cmd/sync-rover-groups/yamlgroupcollector_test.go
petr-muller: Yeah, but "no matter what" is not a useful behavior when you are a user of these tests. Readability is one thing, but the logic needs to be correct first, and only then readability is a concern.

- Actual value does not have a contract when the error is expected: Checking it in that case prevents the implementation to legitimately change (the method can legitimately return whatever it wants when there is an error, but the tests would prevent it)
- When error is not expected but is returned, then that is a main problem. Comparing actual values in that case only adds noise to the output.

cmd/sync-rover-groups/main.go
petr-muller: Yeah, it looks like advanced behavior could be added with a config file like that in the future when it is needed. No need to add that to this PR, it's already large.

cmd/sync-rover-groups/yamlgroupcollector.go
petr-muller: Yes, looks good now :+1: Thanks!

cmd/sync-rover-groups/yamlgroupcollector.go
petr-muller: Yes, looks good now +1 Thanks!

cmd/sync-rover-groups/main.go
hongkailiu: I will do the config file with a follow up PR. This one is big enough.

cmd/sync-rover-groups/main.go
hongkailiu: I missed this one. Strange.
On it now.

cmd/sync-rover-groups/main.go
hongkailiu: It is quick enough for the current amount of groups.
We could improve this later in the future if necessary.

cmd/sync-rover-groups/main.go
hongkailiu: https://github.com/kubernetes/test-infra/blob/681f3531eceb7bc827d6f99de567cd9087bcbcfc/prow/flagutil/github.go#L44

Maybe singular is also fine?

cmd/sync-rover-groups/yamlgroupcollector_test.go
hongkailiu: Updated.

I was trying to grep some reference on "the method can legitimately return whatever it wants when there is an error" but did not find any.
In case you have it handy, I would like to see it.

cmd/github-ldap-user-group-creator/README.md
hongkailiu: Updated.

hongkailiu:
https://github.com/openshift/ci-tools/pull/2595#discussion_r788197914

I added the support for the config file into this PR. I promise I wont add more into this PR.

cmd/sync-rover-groups/README.md
hongkailiu: Updated.

PR #2596, opened 2022-01-13
Skip certain tests that require history rewrite

DennisPeriquet:


pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator.go
```diff
@@ -126,7 +130,12 @@ func (a *analysisJobAggregator) FindRelatedJobs(ctx context.Context) ([]jobrunag
 
 			if a.prowJobMatcher(prowJob) {
 				relatedJobRuns = append(relatedJobRuns, jobRunInfo)
-				break
+
+				// We run aggregated jobs 10 times, so once we find them all
```
deads2k, line 134: this is a configuration knob.  We cannot hardcode ten

DennisPeriquet: do you have a constant defined somewhere already? I can us it if so.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -490,6 +490,69 @@ var testsRequiringHistoryRewrite = map[testCoordinates]string{
...
 		testName: "[sig-api-machinery] Kubernetes APIs remain available with reused connections",
 	}: "history correction on improperly loose disruption criteria, expires 2022-01-15",
+
+	{
+		jobName:  "aggregated-azure-ovn-upgrade-4.10-micro-release-openshift-release-analysis-aggregator",
+		testName: "[sig-arch][bz-kube-apiserver][Late] Alerts alert/KubeAPIErrorBudgetBurn should not at or above info [Suite:openshift/conformance/parallel]",
+	}: "history correction on improperly loose disruption criteria, expires 2022-01-24",
+	{
+		jobName:  "aggregated-azure-ovn-upgrade-4.10-micro-release-openshift-release-analysis-aggregator",
+		testName: "[bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not at or above info",
```
dgoodwin, line 500: deads just changed this to "should not be at" in https://github.com/openshift/origin/pull/26743

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -490,6 +490,69 @@ var testsRequiringHistoryRewrite = map[testCoordinates]string{
 		jobName:  "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
 		testName: "[sig-api-machinery] Kubernetes APIs remain available with reused connections",
 	}: "history correction on improperly loose disruption criteria, expires 2022-01-15",
+
+	{
+		jobName:  "aggregated-azure-ovn-upgrade-4.10-micro-release-openshift-release-analysis-aggregator",
```
deads2k, line 495: you need to wire against the periodic that is the basis for this aggregated job.

deads2k: > you need to wire against the periodic that is the basis for this aggregated job.

you can click on one of the sub-job-runs to get that name.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -490,6 +490,69 @@ var testsRequiringHistoryRewrite = map[testCoordinates]string{
...
+		testName: "[sig-arch][bz-kube-apiserver][Late] Alerts alert/KubeAPIErrorBudgetBurn should not at or above info [Suite:openshift/conformance/parallel]",
```
deads2k, line 496: name is getting fixed in https://github.com/openshift/origin/pull/26743/files

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
DennisPeriquet: ok, I'm taking the actual periodic jobName from the name of the 10 jobs in `aggregation-jobrun-summary.html/FinshedJobs` output -- hopefully this is correct.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -497,6 +497,30 @@ type testCoordinates struct {
...
+		}
+		tc2 := testCoordinates{
+			jobName:  jobName,
+			testName: "bz-kube-apiserver][invariant] alert/KubeAPIErrorBudgetBurn should not be at or above info",
+		}
+		testsRequiringHistoryRewrite[tc1] = "history correction on improperly loose disruption criteria, expires 2022-01-24"
+		testsRequiringHistoryRewrite[tc2] = "history correction on improperly loose disruption criteria, expires 2022-01-24"
+	}
+}
+
```
dgoodwin, line 523: Using the init here is a clever find and does cut down on some boilerplate. It's a little concerning to use a package variable and then an init block to modify it, so the definition is split in two places. I'm not 100% sure what that would do looking at it, but it's definitely unusual. I'd suggest either adding the 7 manually to the map without the loop, or going to a getTestsRequiringHistoryRewrite func if we're at a point where we'd like programmatic logic to build the list. Both will be a little simpler for anyone modifying this in the future.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -497,6 +497,30 @@ type testCoordinates struct {
...
+func init() {
+
+	for _, jobName := range []string{
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-azure-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade",
+		"periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-gcp-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-gcp-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-aws-ovn-upgrade",
```
dgoodwin, line 509: Job names look good.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -481,22 +481,46 @@ func (a *weeklyAverageFromTenDays) CheckFailed(ctx context.Context, jobName stri
...
+var testsRequiringHistoryRewrite map[testCoordinates]string
 
 type testCoordinates struct {
 	jobName  string
 	testName string
 }
 
+func init() {
+
+	testsRequiringHistoryRewrite[testCoordinates{
```
deads2k, line 493: panics on nil map.

DennisPeriquet: bah, sorry about that -- I didn't get to unit test it yet.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -481,22 +481,46 @@ func (a *weeklyAverageFromTenDays) CheckFailed(ctx context.Context, jobName stri
...
-	{
-		jobName:  "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
-		testName: "[sig-api-machinery] OpenShift APIs remain available with reused connections",
-	}: "history correction on improperly loose disruption criteria, expires 2022-01-15",
-	{
-		jobName:  "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
-		testName: "[sig-api-machinery] Kubernetes APIs remain available with reused connections",
-	}: "history correction on improperly loose disruption criteria, expires 2022-01-15",
-}
+var testsRequiringHistoryRewrite map[testCoordinates]string
```
deads2k, line 484: var testsRequiringHistoryRewrite = map[testCoordinates]string{}

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
DennisPeriquet: fixed

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -481,22 +481,48 @@ func (a *weeklyAverageFromTenDays) CheckFailed(ctx context.Context, jobName stri
...
 
 type testCoordinates struct {
 	jobName  string
 	testName string
 }
 
+func init() {
+
+	testsRequiringHistoryRewrite = make(map[testCoordinates]string)
```
deads2k, line 493: don't overwrite here.  if you  have multiple init blocks, you'll overwriting multiple times, right?

PR #2597, opened 2022-01-13
Rename filename cluster-init.yaml

hongkailiu:
The yaml files in the folder are usually manifests for the clusters and are applied by `applyconfig`.
This file is a config file for the tool.
Let us ignore it by `_` prefix in the name.

/cc @openshift/test-platform


hongkailiu:
/hold

need to go together with https://github.com/openshift/release/pull/25221

petr-muller (approved):

hongkailiu:
> Huh does this mean that `applyconfig` currently tries to `oc apply` it? Why isn't it failing?

No. Its parent folder is not used as an arg for `applyconfig`. We use its subfolders `build0n` instead.
But for sync-rover-groups, I will use its parent folder.
https://github.com/openshift/ci-tools/pull/2595

PR #2598, opened 2022-01-13
Clusterclaim: Make sure cleanup always happens

alvaroaleman:
* Use the cleanupCtx for cleanup, as the default ctx gets cancelled
* Start a goroutine that does the cleanup, so it will happen even if the
  wrapped step blocks

Ref https://issues.redhat.com/browse/DPTP-2706

### DIFF
--- a/pkg/steps/cluster_claim.go
+++ b/pkg/steps/cluster_claim.go
@@ -13,6 +13,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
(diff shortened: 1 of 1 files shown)


pkg/steps/cluster_claim.go
```diff
@@ -93,18 +94,28 @@ func (s *clusterClaimStep) run(ctx context.Context) error {
...
+			}
+		}()
+	}
 	if err != nil {
 		acquireErr := results.ForReason("acquiring_cluster_claim").ForError(err)
 		// always attempt to delete claim if one exists
 		var releaseErr error
 		if clusterClaim != nil {
-			releaseErr = results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(ctx, clusterClaim, true))
+			releaseErr = results.ForReason("releasing_cluster_claim").ForError(s.releaseCluster(cleanupCtx, clusterClaim, true))
```
hongkailiu, line 112: TIL: Nice.
I was looking for such a thing the last time you told me the issue.
Also wondering if we should the same for Boskos accquire: what happened to an acquired resource from Boskos if the test is aborted? I guess the current implementation also does the clean up based on some timeout.



https://github.com/openshift/ci-tools/blob/83e16a4b6b391f438d57de976689624617a7d418/pkg/steps/pod.go#L33-L36

pkg/steps/cluster_claim.go
```diff
@@ -199,13 +210,13 @@ func (s *clusterClaimStep) releaseCluster(ctx context.Context, clusterClaim *hiv
 	logger := logrus.WithField("clusterClaim.Namespace", clusterClaim.Namespace).WithField("clusterClaim.Name", clusterClaim.Name)
 	if err := s.saveArtifacts(ctx, clusterClaim.Namespace, clusterClaim.Name, printConditions); err != nil {
 		// logging the error without failing the test
-		logger.Error("Failed to save artifacts before releasing the claimed cluster")
+		logger.WithError(err).Error("Failed to save artifacts before releasing the claimed cluster")
```
hongkailiu, line 213: Oops. My bad.

hongkailiu (commented):
/lgtm

PR #2600, opened 2022-01-14
handle history correction for service load balancer disruption on azure

deads2k:
see https://coreos.slack.com/archives/C01CQA76KMX/p1642174183197200


dgoodwin:
/lgtm

cc @DennisPeriquet

PR #2603, opened 2022-01-18
[WIP] pkg/util: add `openat(2)`-based walk function

bbguimaraes:
Replace `filepath.WalkDir` in the configuration and step registry loading
functions with a reimplementation that always opens files relative to their
parent directory.

This should make sure the traversal is consistent even if the root
directory is moved, e.g. when it is a Kubernetes mount that is updated
while the directory tree is being traversed.

The implementation strives to stay as close to the stdlib version as
possible:

https://go.googlesource.com/go.git/+/refs/tags/go1.17.6/src/io/fs/walk.go

---

Outstanding work still required:

- [x] Merge the subset in https://github.com/openshift/ci-tools/pull/2604 before this PR.
- [x] This has bumped the open file descriptor count at startup beyond my machine's limit (1024).  The [unbounded parallelism in our configuration loading](https://github.com/openshift/ci-tools/blob/c0bdd517188f9592a6677296894a1ee48292f9f5/pkg/load/load.go#L96) is probably counterproductive anyway, it'll need to be synchronized.
  - [ ] https://github.com/openshift/ci-tools/pull/2627
  - [ ] https://github.com/openshift/ci-tools/pull/2628
- [ ] Extensive testing with Kubernetes mounts in a pod.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -183,7 +183,7 @@ require (
 	golang.org/x/lint v0.0.0-20210508222113-6edffad5e616 // indirect
 	golang.org/x/mod v0.4.2 // indirect
 	golang.org/x/net v0.0.0-20211020060615-d418f374d309 // indirect
-	golang.org/x/sys v0.0.0-20210817190340-bfb29a6856f2 // indirect
+	golang.org/x/sys v0.0.0-20210817190340-bfb29a6856f2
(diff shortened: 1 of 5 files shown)


bbguimaraes:
Upon further examination, this is unlikely to be the (sole/as-is) solution used.  Closing to remove from the PR list, will reopen at a later point.

/close

PR #2605, opened 2022-01-18
update the wiring to handle testsuites

deads2k:
This is needed to lookup data correctly.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
@@ -151,11 +151,16 @@ type weeklyAverageFromTenDays struct {
 
 	queryTestRunsOnce        sync.Once
 	queryTestRunsErr         error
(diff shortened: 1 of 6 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -531,11 +549,15 @@ func init() {
...
+func testShouldAlwaysPass(jobName, testName, testSuiteName string) string {
 	coordinates := testCoordinates{
-		jobName:  jobName,
-		testName: testName,
+		jobName:       jobName,
+		testName:      testName,
+		testSuiteName: testSuiteName,
 	}
+	// TODO remove this, but for now we have it to match our current hardcoded constants
+	coordinates.testSuiteName = ""
```
deads2k, line 559: This is the bit to remove once you fix the hardcoded lookups from above.

DennisPeriquet:
makes enough sense to me to
/lgtm

PR #2606, opened 2022-01-20
Skip [sig-network] test that was failing for a while but unnoticed

DennisPeriquet:
Regarding: https://bugzilla.redhat.com/show_bug.cgi?id=2042956

For now we'll skip since it's been failing for a while but unnoticed (as it was mistakenly classified as a flake before [a separate testSuite name was added to distinguish them](https://github.com/openshift/origin/pull/26755/files)).

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
@@ -509,13 +509,15 @@ type testCoordinates struct {
 func init() {
 
 	testsRequiringHistoryRewrite[testCoordinates{
-		jobName:  "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
-		testName: "[sig-api-machinery] OpenShift APIs remain available with reused connections",
+		jobName:       "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
+		testName:      "[sig-api-machinery] OpenShift APIs remain available with reused connections",
+		testSuiteName: "Cluster upgrade",
 	}] = "history correction on improperly loose disruption criteria, expires 2022-01-15"
 
 	testsRequiringHistoryRewrite[testCoordinates{
-		jobName:  "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
-		testName: "[sig-api-machinery] Kubernetes APIs remain available with reused connections",
+		jobName:       "periodic-ci-openshift-release-master-nightly-4.10-e2e-aws-upgrade",
+		testName:      "[sig-api-machinery] Kubernetes APIs remain available with reused connections",
+		testSuiteName: "Cluster upgrade",
 	}] = "history correction on improperly loose disruption criteria, expires 2022-01-15"
 
 	for _, jobName := range []string{
@@ -523,8 +525,9 @@ func init() {
 		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade",
 	} {
 		testsRequiringHistoryRewrite[testCoordinates{
-			jobName:  jobName,
-			testName: "[sig-network-edge] Application behind service load balancer with PDB remains available using new connections",
+			jobName:       jobName,
+			testName:      "[sig-network-edge] Application behind service load balancer with PDB remains available using new connections",
+			testSuiteName: "Cluster upgrade",
 		}] = "history correction on kube update, expires 2022-01-24 - https://bugzilla.redhat.com/show_bug.cgi?id=2040715"
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -547,6 +547,22 @@ func init() {
...
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-aws-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-azure-upgrade-single-node",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade",
+		"periodic-ci-openshift-release-master-nightly-4.10-upgrade-from-stable-4.9-e2e-metal-ipi-upgrade",
+	} {
+		testsRequiringHistoryRewrite[testCoordinates{
+			jobName:       jobName,
+			testName:      "[sig-network] pods should successfully create sandboxes by other",
+			testSuiteName: "openshift-upgrade-test",
```
dgoodwin, line 563: openshift-tests-upgrade

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -555,8 +571,6 @@ func testShouldAlwaysPass(jobName, testName, testSuiteName string) string {
 		testName:      testName,
 		testSuiteName: testSuiteName,
 	}
-	// TODO remove this, but for now we have it to match our current hardcoded constants
-	coordinates.testSuiteName = ""
```
dgoodwin, line 559: Nervous about this but it does seem correct. We don't normally define a testSuite before this, now we do and this value should come through. I'm unclear why we were setting empty string prior to this patch though, it should be that by default.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -547,6 +547,22 @@ func init() {
...
 	}
+
+	for _, jobName := range []string{
+		"periodic-ci-openshift-multiarch-master-nightly-4.10-upgrade-from-nightly-4.9-ocp-remote-libvirt-s390x",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-aws-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-e2e-azure-upgrade-single-node",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-ovn-upgrade",
+		"periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-azure-upgrade",
+		"periodic-ci-openshift-release-master-nightly-4.10-upgrade-from-stable-4.9-e2e-metal-ipi-upgrade",
```
dgoodwin, line 558: Some of these are not jobs we aggregate on but we didn't exactly make that clear this morning, I didn't even realize. The multiarch one for sure can go. Probably the single-node, and the metal job, none of those should be aggregated, but may want to wait to see what David thinks.

dgoodwin (commented):
Looks good but we'll need your input and /approve @deads2k

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
deads2k, line 563: this value needs to be set in all the skips above too, right?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
deads2k: Bring up the results in the test comparison view.  You only need to force skip on ones below about 50% pass rate.  And be sure that blocker bugs are open and don't let them be closed on, "no longer failing in CI".

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
dgoodwin: I wonder which value needs to be set though, openshift-tests or openshift-tests-upgrade. Possible some have moved.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
dgoodwin: Test comparison view: https://datastudio.google.com/reporting/519bfc5e-6996-46bf-9269-0e2ca6d95eb4/page/qKmaC

Filter on test name contains "sandboxes by other".

Here's the blocker for at least aws-own: https://bugzilla.redhat.com/show_bug.cgi?id=2038481

The only one below 50% in that chart is aws ovn as far as I can tell from this.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
deads2k: The "Test Comparison" page can tell you.

dgoodwin: It's snipping the end of the name, so I can't quite see it unless I'm missing something.

dgoodwin: Can use the export link in top right of the form to get a CSV that will show you the full suite name.

dgoodwin: Or just hover :)

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
dgoodwin: I think it's just periodic-ci-openshift-release-master-ci-4.10-upgrade-from-stable-4.9-e2e-aws-ovn-upgrade

We failed a gcp ovn but it was only off by one pass, 8 when we required 9, so that's not the same issue at all.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
dgoodwin: @DennisPeriquet so to be clear, check each of the tests above, look for them in the Test Comparison chart in data studio, find the ones not passing well (thus were added to ignore), verify their suite name, and add to these maps so each has a suite name.

DennisPeriquet: oh man, can't believe I got that wrong.  Thx.

DennisPeriquet: @deads2k when I put the testSuiteName in there, do I need to skip both the version that has the `"openshift-tests"`, `"openshift-tests-upgrade"` and `null` (I'm assuming `null` renders to `""`)?

DennisPeriquet: actually, the suite may also be something other than `openshift-tests` and `openshift-tests-upgrade`.

dgoodwin: i think you need to lookup which one is failing in the Test Comparison chart.

DennisPeriquet: yup, the "34.63% Working" one has Suite=openshift-tests-upgrade

PR #2607, opened 2022-01-20
add equinix-ocp-metal cluster profile

andfasano:
As part of the reorganization of the Equinix accounts per team, this new cluster profile will be used by the Metal team for their CI jobs, instead of the `packet` cluster profile.
Once all the jobs will have been migrated to the new profile, the `packet` profile could be safely deleted

cc @derekhiggins @ardaguclu


derekhiggins:
lgtm

PR #2625, opened 2022-01-26
Improve logging in ldap-user-group-creator

petr-muller:
- Log whether we ended up modifying cluster state
- Log what operation failed when something fails

### DIFF
--- a/cmd/github-ldap-user-group-creator/main.go
+++ b/cmd/github-ldap-user-group-creator/main.go
@@ -297,10 +297,14 @@ func ensureGroups(ctx context.Context, clients map[string]ctrlruntimeclient.Clie
 		if dryRun {
 			return nil
 		}
-		if _, err := UpsertGroup(ctx, client, group); err != nil {
+		if modified, err := upsertGroup(ctx, client, group); err != nil {
 			return fmt.Errorf("failed to upsert group %s on cluster %s: %w", group.Name, cluster, err)
+		} else if modified {
+			logger.Info("Upserted group (created or modified on the cluster")
+		} else {
+			logger.Info("Group with expected members already present in the cluster")
 		}
-		logger.Info("Upserted group")
+
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/github-ldap-user-group-creator/main.go
```diff
@@ -359,24 +363,27 @@ func validate(group *userv1.Group) error {
 	return nil
 }
 
-func UpsertGroup(ctx context.Context, client ctrlruntimeclient.Client, group *userv1.Group) (created bool, err error) {
```
petr-muller, line 362: I've repurposed `created` because it was ignored anyway. The method is not used elsewhere, so I made it private, too.

cmd/github-ldap-user-group-creator/main.go
```diff
@@ -359,24 +363,27 @@ func validate(group *userv1.Group) error {
...
 	if err := client.Delete(ctx, existing); err != nil {
-		return false, fmt.Errorf("delete failed: %w", err)
+		return false, fmt.Errorf("[3] delete failed: %w", err)
 	}
 	// Recreate counts as "Update"
-	return false, client.Create(ctx, group)
+	if err := client.Create(ctx, group); err != nil {
+		return false, fmt.Errorf("[4] create failed: %w", err)
+	}
+	return true, nil
```
hongkailiu, line 388: This changes the semantics though: from "true only when creating" to "true when creating or modifying".

hongkailiu (approved):

cmd/github-ldap-user-group-creator/main.go
hongkailiu: It was copied from the existing work. 😉

cmd/github-ldap-user-group-creator/main.go
petr-muller: Yeah, but there's no consumer of that value until now, so we can change semantics like we please :smiling_imp:

PR #2627, opened 2022-01-26
Unify `ci-operator` configuration loading

bbguimaraes:
Two completely independent but equivalent versions have evolved in `pkg/config`
and `pkg/load`.  This change removes the latter and corrects all users of the
former (the opposite would result in a circular dependency).


stevekuznetsov (approved):

bbguimaraes:
/hold cancel

Ready for review.  A few notes:

- As usual, each commit is independent and passes tests.
- The final line count is massive mostly because code is shuffled around, and most of the changes are in tests:
```
$ git diff --shortstat master -- ':^*test*'
 5 files changed, 120 insertions(+), 160 deletions(-)
```
- The rough outline is:
  - improve existing tests (3 commits)
  - remove one of the duplicated functions (1 commit)
  - ensure loading from a `ConfigMap` is exercised in our unit tests (2 commits, this was the original issue caught by the E2E tests)
  - move single-caller function out of the shared packages (1 commit)

pkg/config/options_test.go
```diff
@@ -266,68 +267,70 @@ func TestOperateOnCIOperatorConfigDir(t *testing.T) {
...
 		t.Run(tc.id, func(t *testing.T) {
 			var errs []error
-			processed := sets.NewString()
-
+			var processed []string
 			if err := tc.options.OperateOnCIOperatorConfigDir(testConfigDir, func(configuration *cioperatorapi.ReleaseBuildConfiguration, info *Info) error {
 				filename := filepath.Base(info.Filename)
-				if !tc.expectedProcessedFiles.Has(filename) {
-					errs = append(errs, fmt.Errorf("file %s wasn't expected to be processed", filename))
+				for _, x := range tc.expectedProcessedFiles {
```
stevekuznetsov, line 348: TBH the set stuff was easier to read - why change it?

pkg/config/options_test.go
```diff
@@ -266,68 +267,70 @@ func TestOperateOnCIOperatorConfigDir(t *testing.T) {
...
+					if x == filename {
+						processed = append(processed, filename)
+						return nil
+					}
 				}
-				processed.Insert(filename)
+				errs = append(errs, fmt.Errorf("file %s wasn't expected to be processed", filename))
```
stevekuznetsov, line 354: Hm, I see the dupe logic now - but why is this necessary? If you record all of the processed names, you should be able to see extras, missings and dupes in the cmp.Diff call

pkg/config/load.go
```diff
@@ -120,6 +120,10 @@ func isConfigFile(info fs.DirEntry) bool {
 	return !info.IsDir() && (extension == ".yaml" || extension == ".yml")
 }
 
+func isMountSpecialFile(path string) bool {
```
stevekuznetsov, line 123: We probably want to copy the comment that explains k8s configmap behavior to  this func?

pkg/config/options_test.go
bbguimaraes: mental stack overflow

pkg/config/options_test.go
bbguimaraes: Reimplemented in a less intrusive way, let me know if that does not address the comment.

pkg/config/load.go
bbguimaraes: Added a comment with the general structure of a mount.

bbguimaraes: (this can be handled better, but… one problem at a time)

pkg/config/options_test.go
```diff
@@ -288,57 +290,74 @@ func TestOperateOnCIOperatorConfigDir(t *testing.T) {
...
-		}
-
-		if len(errs) > 0 {
-			t.Fatal("unexpected errors: %w", errs)
-		}
-
-		if diff := cmp.Diff(processed, tc.expectedProcessedFiles); diff != "" {
-			t.Fatal(diff)
-		}
+			if diff := cmp.Diff(processed, tc.expectedProcessedFiles); diff != "" {
```
stevekuznetsov, line 357: Super-nit: you have sets, so you can just .Difference() and check on the length of it

stevekuznetsov (approved):
/hold

If anyone else wants a look, holding

bbguimaraes:
/hold cancel

The original `lgtm` was removed with the latest change, make it easier to see this as the current base in the PR list.

petr-muller (approved):
It was intimidating at first glance, but when walking commit-by-commit everything made perfect sense, really well broken down :+1:

PR #2628, opened 2022-01-26
pkg/config: use generic pipeline function

bbguimaraes:
This pattern is everywhere, implemented slightly differently each time.  Some
versions do not bound the maximum number of Go routines, causing file descriptor
exhaustion.

This change creates a small library that is as generic as I managed to make it
in Go and reimplements `ci-operator` configuration loading using it.  More to
come.

/hold

Based on https://github.com/openshift/ci-tools/pull/2627.

https://issues.redhat.com/browse/DPTP-2531

---

See also:

- [ ] https://github.com/openshift/ci-tools/pull/2629
- [ ] https://github.com/openshift/ci-tools/pull/2630
- [ ] https://github.com/openshift/ci-tools/pull/2631
- [ ] https://github.com/openshift/ci-tools/pull/2632

### DIFF
--- a/pkg/util/sync_test.go
+++ b/pkg/util/sync_test.go
@@ -1,0 +1,44 @@
+package util
+
+import (
+	"fmt"
+)
+
+func ExampleProduceMapReduce() {
+	input := []int{0, 1, 2, 3}
+	inputCh := make(chan int)
+	errCh := make(chan error)
+	produce := func() error {
+		defer close(inputCh)
+		for _, x := range input {
+			inputCh <- x
+		}
+		return nil
+	}
+	outputCh := make(chan int)
+	map_ := func() error {
+		for x := range inputCh {
+			if x == 2 {
+				errCh <- fmt.Errorf("%d", x)
+			} else {
+				outputCh <- x
+			}
+		}
+		return nil
+	}
+	done := func() { close(outputCh) }
+	reduce := func() error {
+		r := 0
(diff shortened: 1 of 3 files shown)


pkg/util/sync_test.go
```diff
@@ -0,0 +1,44 @@
+package util
+
+import (
+	"fmt"
+)
+
+func ExampleProduceMapReduce() {
```
petr-muller, line 7: Can we make this a real test?

bbguimaraes: It is a real test:

```
$ go test -v ./pkg/util | grep Example
=== RUN   ExampleProduceMapReduce
--- PASS: ExampleProduceMapReduce (0.00s)
```

and is [supposed to be how examples are embedded in package documentation](https://go.dev/blog/examples), which is why I decided to try to write it this way.

bbguimaraes: After the usual fight (I hate Go modules), this is how it looks like rendered:

<details>

Screen Shot 2022-02-09 at 12 50 29

</details>

Apparently, there is no way to format `code`, ` `` ` is not respected and I can't find examples in the stdlib.

petr-muller: TIL

petr-muller: Pretty awesome

petr-muller (approved):

PR #2633, opened 2022-01-26
Refactor the aggregator job to be generated by an unresolved ci-op config

droslean:
/cc @openshift/test-platform 

Reviewer notes:
- Example of an aggregator periodic: [openshift/release/openshift-release-infra-periodics.yaml#L2270-L2356](https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml#L2270-L2356)
- Step [prpqr-aggregator/openshift-release-analysis-prpqr-aggregator](https://github.com/openshift/release/blob/master/ci-operator/step-registry/openshift/release/analysis/prpqr-aggregator/openshift-release-analysis-prpqr-aggregator-ref.yaml)

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -9,11 +9,11 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -46,7 +46,8 @@ const (
 	conditionAllJobsTriggered = "AllJobsTriggered"
 	conditionWithErrors       = "WithErrors"
 
-	aggregationIDLabel = "release.openshift.io/aggregation-id"
+	aggregationIDLabel          = "release.openshift.io/aggregation-id"
+	defaultAggregatorJobTimeout = 6 * time.Hour
 )
 
 type injectingResolverClient interface {
@@ -229,7 +230,7 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 			prowjobsToCreate = append(prowjobsToCreate, aggregatedProwjobs...)
 
 			submitted := generateJobNameToSubmit(baseMetadata, inject, &prpqr.Spec.PullRequest)
-			aggregatorJob, err := generateAggregatorJob(uid, mimickedJob, jobSpec.JobName(jobconfig.PeriodicPrefix), req.Name, req.Namespace, r.prowConfigGetter.Config(), time.Now(), submitted)
+			aggregatorJob, err := generateAggregatorJob(baseMetadata, uid, mimickedJob, jobSpec.JobName(jobconfig.PeriodicPrefix), req.Name, req.Namespace, r.prowConfigGetter.Config(), time.Now(), submitted)
 			if err != nil {
 				logger.WithError(err).Error("Failed to generate an aggregator prowjob")
 				statuses[mimickedJob] = &v1.PullRequestPayloadJobStatus{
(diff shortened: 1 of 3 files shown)


pkg/controller/prpqr_reconciler/prpqr_reconciler_test.go
```diff
@@ -204,10 +205,26 @@ func prunePRPQRForTests(items []v1.PullRequestPayloadQualificationRun) {
...
+			if _, ok := c.Tests[0].MultiStageTestConfiguration.Environment["JOB_START_TIME"]; ok {
+				c.Tests[0].MultiStageTestConfiguration.Environment["JOB_START_TIME"] = "1970-01-01T01:00:00+01:00"
+			}
+
+			unresolvedConfigRaw, err := yaml.Marshal(c)
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			items[i].Spec.PodSpec.Containers[0].Env[0].Value = string(unresolvedConfigRaw)
```
jmguzik, line 227: Demeter cries :)

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -556,80 +556,62 @@ func generateAggregatedProwjobs(uid string, ciopConfig *api.ReleaseBuildConfigur
...
-		Name:      aggregatorJobName,
-		Namespace: &[]string{prpqrNamespace}[0],
-		Spec:      spec,
-	}}
+	jobBaseGen := prowgen.NewProwJobBaseBuilderForTest(ciopConfig, &prowgen.ProwgenInfo{}, prowgen.NewCiOperatorPodSpecGenerator(), ciopConfig.Tests[0])
+
+	periodic := prowgen.GeneratePeriodicForTest(jobBaseGen, &prowgen.ProwgenInfo{}, "@yearly", "", false, ciopConfig.CanonicalGoRepository)
+	periodic.Name = aggregatorJobName
+
+	periodic.DecorationConfig.Timeout = &prowv1.Duration{Duration: 6 * time.Hour}
```
jmguzik, line 600: Why 6? Can we describe this somehow?

jmguzik:
I think I have a problem with those arbitrary 6 hours. It's related to periodic probably but for newbies it's not really visible why 6 not 5, 4 etc.
Otherwise LGTM.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
droslean: I was advised on how the other aggregator job is configured. See https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/release/openshift-release-infra-periodics.yaml#L2274-L2275. However, the step already has a 6h timeout, but shrug.

pkg/controller/prpqr_reconciler/prpqr_reconciler_test.go
droslean: The problem here is that the unresolved config is inlined into the env var. We have to prune the job time to allow tests to run, so this is an unfortunate way to do this.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
jmguzik: I am saying if there is no way to configure this, at least it can be explained by giving 6 a proper name so that the next person would know what 6 means actually.

pkg/controller/prpqr_reconciler/prpqr_reconciler_test.go
jmguzik: Yes. Not a big issue, just spotted and commented :) I see it happens quite a lot in Go.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
droslean: Do you mean to add a comment to explain why I put 6hours?

jmguzik: Could be a comment but better: a descriptive name for this time period. It's basically a const.

droslean: But the `periodic.DecorationConfig.Timeout` variable is already descriptive enough.

droslean: I will try to add a comment to explain why we need to configure a 6-hour timeout.  But to be honest I don't even know why :D. Also, I believe that we could just remove that completely. The step has already a 6-hour timeout. I was hoping that won't affect the whole pod timeout. @bbguimaraes ??

stevekuznetsov: If you don't specify a timeout here, the Prow default will kick in and I think that's 2h.

pkg/controller/prpqr_reconciler/prpqr_reconciler_test.go
```diff
@@ -204,10 +205,26 @@ func prunePRPQRForTests(items []v1.PullRequestPayloadQualificationRun) {
...
-			items[i].Spec.PodSpec.Containers[0].Args[5] = "--job-start-time=some-time"
+			unResolvedConfig := items[i].Spec.PodSpec.Containers[0].Env[0].Value
+
+			c := &api.ReleaseBuildConfiguration{}
+			if err := yaml.Unmarshal([]byte(unResolvedConfig), c); err != nil {
+				t.Fatal(err)
+			}
+
+			if _, ok := c.Tests[0].MultiStageTestConfiguration.Environment["JOB_START_TIME"]; ok {
+				c.Tests[0].MultiStageTestConfiguration.Environment["JOB_START_TIME"] = "1970-01-01T01:00:00+01:00"
```
stevekuznetsov, line 219: Why is this necessary? Surely you could either make the assertion logic later smarter, or use an injectable fixture for `time` so the test can set it?

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
petr-muller: @droslean using named constants instead of magic values is a good practice though. You could call `Duration: 6 * time.Hour` something like `defaultAggregatorJobTimeout`, give it a comment like "Aggregator jobs need more time to finish than the jobs they are aggregating (default job timeout is set to 4h in OpenShift CI)" and the result is that whoever comes up here later knows that the value is set intentionally, not as a ballpark number and exactly knows why the number was chosen.

See: https://refactoring.guru/replace-magic-number-with-symbolic-constant

droslean: addressed

PR #2638, opened 2022-02-02
[WIP] omitempty on bool fields causes genyaml to omit them

smg247:


petr-muller:
Hm... do we want this though? Not using `omitempty` has a price, we'll generate a million of explicit `something: false` everywhere ;(

smg247:
> Hm... do we want this though? Not using `omitempty` has a price, we'll generate a million of explicit `something: false` everywhere ;(

Yes, I hate every solution to this problem including this one. I have actually explored other options as well. Here are my findings so far:

- Remove omitempty from all boolean fields. It requires a lot of changes to tests and outstanding changes to integration tests (that for some reason pass locally) and changes things cosmetically.
- Reflectively populate the bool fields as true. That kind of sucks because it's misleading. I don’t think this will fly.
- Reflectively change the type of the bool field to *bool. This isn't actually possible in go.
- Reflectively change the tag to remove omitempty. Not sure how I can do this https://go.dev/play/p/QNArOeqy94 exists, but I need to figure out how to apply that to populate_struct which is proving difficult.
- Change all of the bool fields to *bool. This will change the semantics of the configs, and perhaps create bugs.
- And finally, not fix this bug ;)

PR #2640, opened 2022-02-03
fix logging for aggregation

deads2k:


droslean (approved):

petr-muller:
That's our 5pm GH rate limiting spike ^^^

PR #2641, opened 2022-02-03
promotion: treat promoting to `origin` as promoting official images

petr-muller:
The motivation for this change is to include OKD in various release automation. From how the code looked like in the past, it seems that it was always the intent, but the code was not maintained and kept updated with how OKD configuration is structured: previously, the OKD check was `namespace == okdPromotionNamespace && name == okd40Imagestream`, but none of the existing ci-operator configs use these names.

Because the usage of `*OfficialImages` proliferated over time, some of the callsites expect the check to include OKD, and some callsites do not (notably, the `openshift-priv` tools are not needed to handle OKD and things actually break if they do). Therefore, I have added an explicit parameter to all these methods to control whether the caller wants to work also with OKD or not.

We also need to evaluate the impact on all code that uses this check, because most of the callsites are probably not aware of OKD at all.

- `RefersToOfficialImages` callsites: `config-brancher`, `BuildsOfficialImages`
- `BuildsOfficialImages` callsites: `autopublicizeconfig`, `ci-operator-config-mirror`, `private-org-peribolos-sync`, `private-org-sync`, `private-prow-configs-mirror`, `PromotesOfficialImages`
- `PromotesOfficialImages` callsites: `payload-testing-prow-plugin`, `Prowgen`, `blocking-issue-creator`, `repo-brancher`

---

- [x] config-brancher: looks good in https://github.com/openshift/release/pull/26703: creates and carries modifications to branches' OKD config
- [x] `autopublicizeconfig`: should exclude OKD, otherwise includes OKD-only repositories in openshift-priv machinery
- [x] `ci-operator-config-mirror`: should exclude OKD, otherwise includes OKD-only repositories in openshift-priv machinery
- [x] `private-org-peribolos-config`: should exclude OKD, otherwise includes OKD-only repositories in openshift-priv machinery
- [x] `private-org-sync`: should exclude OKD, otherwise includes OKD-only repositories in openshift-priv machinery
- [x] `private-prow-configs-mirror`: should exclude OKD, otherwise includes OKD-only repositories in openshift-priv machinery
- [x] `payload-testing-prow-plugin`: does not matter too much, can include OKD
- [x] `blocking-issues-creator`: should exclude OKD - branches on OKD repositories are not fast forwarded, so no need to block merges
- [x] `repo-brancher`: should exclude OKD - branches on OKD were never fast-forwarded and nobody ever asked them to be
- [x] `prowgen`: should include OKD, will make OKD-promoting image jobs check the feasibility of building a release, this is desirable

### DIFF
--- a/cmd/config-brancher/main.go
+++ b/cmd/config-brancher/main.go
@@ -69,7 +69,7 @@ func main() {
 	}
 
 	var toCommit []config.DataWithInfo
-	if err := o.OperateOnCIOperatorConfigDir(o.ConfigDir, func(configuration *api.ReleaseBuildConfiguration, info *config.Info) error {
+	if err := o.OperateOnCIOperatorConfigDir(o.ConfigDir, promotion.WithOKD, func(configuration *api.ReleaseBuildConfiguration, info *config.Info) error {
 		for _, output := range generateBranchedConfigs(o.CurrentRelease, o.BumpRelease, o.FutureReleases.Strings(), config.DataWithInfo{Configuration: *configuration, Info: *info}) {
 			if !o.Confirm {
 				output.Logger().Info("Would commit new file.")
@@ -169,7 +169,7 @@ func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string) 
 func updateImages(config *api.ReleaseBuildConfiguration, currentRelease, futureRelease string) {
 	for name := range config.InputConfiguration.BaseImages {
 		image := config.InputConfiguration.BaseImages[name]
-		if promotion.RefersToOfficialImage(image.Name, image.Namespace) && image.Name == currentRelease {
+		if promotion.RefersToOfficialImage(image.Namespace, promotion.WithOKD) && image.Name == currentRelease {
 			image.Name = futureRelease
 		}
 		config.InputConfiguration.BaseImages[name] = image
@@ -177,7 +177,7 @@ func updateImages(config *api.ReleaseBuildConfiguration, currentRelease, futureR
 
 	for i := range config.InputConfiguration.BaseRPMImages {
 		image := config.InputConfiguration.BaseRPMImages[i]
-		if promotion.RefersToOfficialImage(image.Name, image.Namespace) && image.Name == currentRelease {
+		if promotion.RefersToOfficialImage(image.Namespace, promotion.WithOKD) && image.Name == currentRelease {
 			image.Name = futureRelease
 		}
 		config.InputConfiguration.BaseRPMImages[i] = image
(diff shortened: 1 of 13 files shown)


petr-muller:
The followup to address breaking-changes: https://github.com/openshift/release/pull/27827

petr-muller:
Note for reviewers: first commit is the actual change in `promote` package, the remaining commits adapt individual tools to the change.

cmd/config-brancher/main.go
```diff
@@ -169,23 +169,23 @@ func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string)
 func updateImages(config *api.ReleaseBuildConfiguration, currentRelease, futureRelease string) {
 	for name := range config.InputConfiguration.BaseImages {
 		image := config.InputConfiguration.BaseImages[name]
-		if promotion.RefersToOfficialImage(image.Name, image.Namespace) && image.Name == currentRelease {
+		if promotion.RefersToOfficialImage(image.Namespace, true) && image.Name == currentRelease {
```
hongkailiu, line 172: true ---> promotion.WithOKD ?

cmd/config-brancher/main.go
```diff
@@ -169,23 +169,23 @@ func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string)
...
 			image.Name = futureRelease
 		}
 		config.InputConfiguration.BaseImages[name] = image
 	}
 
 	for i := range config.InputConfiguration.BaseRPMImages {
 		image := config.InputConfiguration.BaseRPMImages[i]
-		if promotion.RefersToOfficialImage(image.Name, image.Namespace) && image.Name == currentRelease {
+		if promotion.RefersToOfficialImage(image.Namespace, true) && image.Name == currentRelease {
```
hongkailiu, line 180: true ---> promotion.WithOKD ?

cmd/config-brancher/main.go
```diff
@@ -169,23 +169,23 @@ func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string)
...
 			image.Name = futureRelease
 		}
 		config.InputConfiguration.BaseRPMImages[i] = image
 	}
 
 	if config.InputConfiguration.BuildRootImage != nil {
 		image := config.InputConfiguration.BuildRootImage.ImageStreamTagReference
-		if image != nil && promotion.RefersToOfficialImage(image.Name, image.Namespace) && image.Name == currentRelease {
+		if image != nil && promotion.RefersToOfficialImage(image.Namespace, true) && image.Name == currentRelease {
```
hongkailiu, line 188: true ---> promotion.WithOKD ?

cmd/config-brancher/main.go
petr-muller: Good catch!

hongkailiu:
/lgtm

/hold

wait for the change to fix breaking to be ready.

petr-muller:
It is, that's https://github.com/openshift/release/pull/27827

/hold cancel

PR #2643, opened 2022-02-04
autoowners: include source and date feched in header comment

petr-muller:
Currently we put a [header like this](https://github.com/openshift/release/blob/08e400216f5d0fd121a517bc97512d7928f44c8f/ci-operator/config/openshift/ci-tools/OWNERS#L1-L2) to OWNERS files:

```
# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools.
# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
```

We can provide more information in there:

```
# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools.
# Fetched from https://github.com/openshift/ci-tools root OWNERS on 2022-02-04T17:14:58Z
# If the repo had OWNERS_ALIASES then the aliases were expanded
# Logins who are not members of 'openshift' organization were filtered out
# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md
```

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -27,8 +27,8 @@ import (
 )
 
 const (
-	doNotEdit     = "# DO NOT EDIT; this file is auto-generated using https://github.com/openshift/ci-tools."
-	ownersComment = "# See the OWNERS docs: https://git.k8s.io/community/contributors/guide/owners.md"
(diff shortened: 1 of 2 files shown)


hongkailiu (commented):
I love the timestamp.
I would love more if we do not repeat the same content in each file but that is not the issue introduced by this PR which only makes it more visible. 😉

hongkailiu (approved):

PR #2645, opened 2022-02-04
sync-rover-groups: generate github/ldap mapping

hongkailiu:
This replaces the implementation [here](https://github.com/openshift/release/blob/678af225364bd492cf796f57e5cb91e6b1fc794f/clusters/psi/github-ldap-mapping-update_cronjob.yaml#L22-L23):

```
$ ldapsearch -LLL -x -h ldap.corp.redhat.com -b ou=users,dc=redhat,dc=com '(rhatSocialURL=GitHub*)' rhatSocialURL uid 2>&1 | tee /tmp/out
$ ldap-users-from-github-owners-files --ldap-file /tmp/out --mapping-file /tmp/mapping.yaml
```

With this PR, `sync-rover-groups` generates the mapping file directly with ldap GoLang lib instead of calling `ldapsearch`.

https://issues.redhat.com/browse/DPTP-2755

/cc @openshift/test-platform


cmd/sync-rover-groups/ldap.go
```diff
@@ -81,3 +78,57 @@ func (r *ldapGroupResolver) resolve(name string) (*Group, error) {
...
+			logrus.WithField("kerberosID", kerberosID).WithField("entry", entry).Warn("failed to parse GitHub ID")
+			continue
+		}
+		ret[gitHubID] = kerberosID
+	}
+
+	return ret, nil
+}
+
+func getGitHubID(value string) string {
```
smg247, line 124: Can you put a comment on this function of what the values of `value` could look like? Its a bit confusing that the result is either the last element or the second to last element. I do see the 2 example from your test case, but having it here as well would make this easier to read.

hongkailiu: Updated.

PR #2648, opened 2022-02-07
ci-operator: retry infra-failed builds immediately

petr-muller:
`ci-operator` was already able to recognize infrastructure-failed builds
from previous runs and retry them. This is an attempt to reuse that code
to retry such failed builds immediately, with two attempts in an
exponential backoff. The backoff has an intentionally long starting
delay of 1 minute to give the infrastructure problem a chance to go
away. The way the code is structured makes it less optimal for the case
where we are retrying infra failures from the previous executions: it
will eat one of the backoff iterations, but such cases should be rare
because ci-op runs should not result in failures caused by
infrastructure failures anymore (because they are retried immediately).

/cc @bbguimaraes @jupierce @openshift/test-platform

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -369,46 +369,57 @@ func isBuildPhaseTerminated(phase buildapi.BuildPhase) bool {
 }
 
 func handleBuild(ctx context.Context, buildClient BuildClient, build *buildapi.Build) error {
-	if err := buildClient.Create(ctx, build); err != nil {
-		if !kerrors.IsAlreadyExists(err) {
-			return fmt.Errorf("could not create build %s: %w", build.Name, err)
+	var buildErr error
+	attempts := 5
+	if boErr := wait.ExponentialBackoff(wait.Backoff{Duration: time.Minute, Factor: 1.5, Steps: attempts}, func() (bool, error) {
+		if err := buildClient.Create(ctx, build); err != nil && !kerrors.IsAlreadyExists(err) {
+			return false, fmt.Errorf("could not create build %s: %w", build.Name, err)
 		}
+
+		buildErr = waitForBuildOrTimeout(ctx, buildClient, build.Namespace, build.Name)
+		if buildErr == nil {
+			if err := gatherSuccessfulBuildLog(buildClient, build.Namespace, build.Name); err != nil {
+				// log error but do not fail successful build
+				logrus.WithError(err).Warnf("Failed gathering successful build %s logs into artifacts.", build.Name)
+			}
+			return true, nil
+		}
+
 		b := &buildapi.Build{}
 		if err := buildClient.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: build.Namespace, Name: build.Name}, b); err != nil {
-			return fmt.Errorf("could not get build %s: %w", build.Name, err)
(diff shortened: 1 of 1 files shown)


pkg/steps/source.go
```diff
@@ -369,46 +369,57 @@ func isBuildPhaseTerminated(phase buildapi.BuildPhase) bool {
 }
 
 func handleBuild(ctx context.Context, buildClient BuildClient, build *buildapi.Build) error {
-	if err := buildClient.Create(ctx, build); err != nil {
-		if !kerrors.IsAlreadyExists(err) {
-			return fmt.Errorf("could not create build %s: %w", build.Name, err)
+	var buildErr error
+	attempts := 3
+	if boErr := wait.ExponentialBackoff(wait.Backoff{Duration: 1 * time.Minute, Factor: 2, Steps: attempts}, func() (bool, error) {
```
jupierce, line 374: Seems like it takes about 5 minutes for a new node to come online, this backoff gives us a chance at one new node arriving in time. Given the low cost of the kubelet rejecting us, I'd consider attempts=5 and factor=1.5 which would give us a few changes for new nodes joining the back.

jupierce (commented):
One thought. Otherwise, looks great!

pkg/steps/source.go
petr-muller: Sounds good!

bbguimaraes: I've always found the `apimachinery` parameters confusing, but note that these values will give us:

```lua
n, d, f, t = 5, 1, 1.5, 0
print("i", "t", "d")
for i = 0, n - 1 do
    print(i, t, d)
    t, d = t + d, d * f
end
```
```
i	t	d
0	0	1
1	1	1.5
2	2.5	2.25
3	4.75	3.375
4	8.125	5.0625
```

Which will give us a single retry (the last) after a new node arrives at `t ~ 5`, unless the combined retries take more than 1.875s (I don't know how long they take and could not find an example).

pkg/steps/source.go
```diff
@@ -369,46 +369,57 @@ func isBuildPhaseTerminated(phase buildapi.BuildPhase) bool {
...
+		if err := buildClient.Delete(ctx, build, &ctrlruntimeclient.DeleteOptions{Raw: &opts}); err != nil && !kerrors.IsNotFound(err) && !kerrors.IsConflict(err) {
+			return false, fmt.Errorf("could not delete build %s: %w", build.Name, err)
+		}
+		if err := waitForBuildDeletion(ctx, buildClient, build.Namespace, build.Name); err != nil {
+			return false, fmt.Errorf("could not wait for build %s to be deleted: %w", build.Name, err)
+		}
+		return false, nil
+	}); boErr != nil {
+		if boErr == wait.ErrWaitTimeout {
+			return fmt.Errorf("build not successful after %d attempts, last error: %w", attempts, buildErr)
```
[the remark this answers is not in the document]
petr-muller: Good point, will address

pkg/steps/source.go
petr-muller: I think that we do not necessarily need to retry until there is a new node - that's scheduler's job. We just need to give the scheduler some chance to retry the bad placement that results in `OutOfCpu`. It is fine if the retried Pod ends up unschedulable, that makes it wait for the node.

PR #2649, opened 2022-02-07
Adding Multiarch platforms

deepsm007:
Adding the architecture to support multiarch. 

/cc @deads2k

### DIFF
--- a/pkg/jobrunaggregator/jobtableprimer/job_typer.go
+++ b/pkg/jobrunaggregator/jobtableprimer/job_typer.go
@@ -29,6 +29,18 @@ func newJob(name string) *jobRowBuilder {
 		platform = openstack
 	}
 
+	architecture := ""
+	switch {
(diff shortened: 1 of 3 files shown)


pkg/jobrunaggregator/jobtableprimer/job_typer.go
```diff
@@ -15,6 +15,8 @@ func newJob(name string) *jobRowBuilder {
 	switch {
 	case strings.Contains(name, "gcp"):
 		platform = gcp
+	case strings.Contains(name, "arm64"):
```
deads2k, line 18: Are these actually platforms, not a new field type call architecture?

deepsm007: In theory there are architecture, thats where my original confusion came along as we want it to be part of separate field

PR #2650, opened 2022-02-08
sync-rover-groups: exclude gitlab users

hongkailiu:
/cc @smg247


cmd/sync-rover-groups/ldap.go
```diff
@@ -126,6 +126,9 @@ func (r *ldapGroupResolver) getGitHubUserKerberosIDMapping() (map[string]string,
 // A user can have multiple "rhatSocialURL"s for GitHub, Twitter, LinkedIn etc, even multiple for GitHub.
 // Example of such values: "Github->https://github.com/tom", "Github->https://github.com/tom/", or "Twitter->https://twitter.com/tom"
 func getGitHubID(value string) string {
+	if strings.Contains(value, "https://gitlab.") {
```
smg247, line 129: should we just verify that it does contain "https://github." instead? I am not sure if there are any valid ids that wouldn't

hongkailiu: Updated.
All the test cases are real (username replaced).

smg247 (approved):

PR #2655, opened 2022-02-08
ci-operator: record aggregate error instead last one

petr-muller:
Addresses https://github.com/openshift/ci-tools/pull/2648#discussion_r801591825

/cc @bbguimaraes @openshift/test-platform


petr-muller:
/hold

Will address in a re-revert of https://github.com/openshift/ci-tools/pull/2656

PR #2659, opened 2022-02-09
ci-operator: retry infra-failed builds immediately

petr-muller:
Previous version caused `resourceVersion should not be set on objects to be created ` errors and needed to be reverted. This PR addresses that bug in a new commit. I also included the error handling change from https://github.com/openshift/ci-tools/pull/2655.

ci-operator was already able to recognize infrastructure-failed builds
from previous runs and retry them. This is an attempt to reuse that code
to retry such failed builds immediately, with two attempts in an
exponential backoff. The backoff has an intentionally long starting
delay of 1 minute to give the infrastructure problem a chance to go
away. The way the code is structured makes it less optimal for the case
where we are retrying infra failures from the previous executions: it
will eat one of the backoff iterations, but such cases should be rare
because ci-op runs should not result in failures caused by
infrastructure failures anymore (because they are retried immediately).

/cc @openshift/test-platform @bbguimaraes @jupierce 

### DIFF
--- a/pkg/steps/bundle_source.go
+++ b/pkg/steps/bundle_source.go
@@ -75,7 +75,7 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
 		s.pullSecret,
 		nil,
 	)
-	return handleBuild(ctx, s.client, build)
+	return handleBuild(ctx, s.client, *build)
 }
 
 func replaceCommand(pullSpec, with string) string {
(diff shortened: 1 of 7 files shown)


bbguimaraes:
It's very sad that the existing code

- discards intermediate results
- uses polling
- has no tests

But none of those are new or need to block this PR.

/lgtm

bbguimaraes:
Maybe someday Go will also advance to 1980s C/++ state of the art and adopt `const` pointers.

petr-muller:
/override ci/prow/e2e-oo 

This is not a new failure (I just happened to trip on a perma-broken conditionally triggered job (which we run as a postsubmit and we ignore it: https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/branch-ci-openshift-ci-tools-master-e2e-oo-post)

PR #2661, opened 2022-02-09
Update cluster type to packet-edge for Assisted and SNO profiles

lranjbar:
Originally all the cluster profiles associated with Equinix were expecting a cluster type to be `packet`. We are now deprecating the old `packet` cluster profile and cluster type. There was a requirement for the two cluster_profiles `packet-assisted` and `packet-sno` to use the same qouta. This updates the cluster type for those profiles to be `packet-edge`

See: https://github.com/openshift/release/pull/25952 where the jobs under `packet-assisted` and `packet-sno` are now being blocked from running.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1260,10 +1260,12 @@ func (p ClusterProfile) ClusterType() string {
 	case ClusterProfileOvirt:
 		return "ovirt"
 	case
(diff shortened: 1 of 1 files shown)


osherdp:
/test integration
same commit has passed this test previously, so I'm assuming it's a flake

PR #2663, opened 2022-02-10
override job_url_prefix for rehearsals

smg247:
In order to allow for the correct Details link to be shared on the gh status for rehearsals for qe-private-deck repos (https://issues.redhat.com/browse/DPTP-2742) we can utilize the `DecorationConfig` where an existing override is in place: https://github.com/kubernetes/test-infra/blob/b800243aebae95801cec2b9f16950623c6a42497/prow/config/config.go#L820

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
 	"strings"
@@ -37,6 +38,7 @@ import (
 	apihelper "github.com/openshift/ci-tools/pkg/api/helper"
 	testimagestreamtagimportv1 "github.com/openshift/ci-tools/pkg/api/testimagestreamtagimport/v1"
 	"github.com/openshift/ci-tools/pkg/config"
+	ciopconfig "github.com/openshift/ci-tools/pkg/config"
 	"github.com/openshift/ci-tools/pkg/diffs"
 	"github.com/openshift/ci-tools/pkg/load"
 	"github.com/openshift/ci-tools/pkg/registry"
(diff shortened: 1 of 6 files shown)


pkg/rehearse/jobs.go
```diff
@@ -1015,3 +1024,15 @@ func (e *Executor) submitPresubmit(job *prowconfig.Presubmit) (*pjapi.ProwJob, e
 	e.loggers.Job.WithFields(pjutil.ProwJobFields(&prowJob)).Info("Submitting a new prowjob.")
 	return &prowJob, e.pjclient.Create(context.Background(), &prowJob)
 }
+
+func determineJobURLPrefix(plank prowconfig.Plank, org string, repo string) string {
+	jobURLPrefix := plank.JobURLPrefixConfig["*"]
```
hongkailiu, line 1029: Do we need a nil check on `JobURLPrefixConfig`?

hongkailiu (commented):
Only a minor. Otherwise,  LGTM.

pkg/rehearse/jobs.go
smg247: Technically no, as it's defined now, and removing it would break a lot of functionality. I toyed with the idea of adding one here, but if it were to be removed in the future (due to deprecation) we would want to blow up here so we could restore functionality.

hongkailiu: Sounds good enough to me.

petr-muller:
Eek came here late, pj-rehearse already has Prow config (also, in pj-rehearse you need to take care whether you are using baseline or pr version of the config)

cmd/pj-rehearse/main.go
```diff
@@ -306,7 +318,11 @@ func rehearseMain() error {
 
 	resolver := registry.NewResolver(refs, chains, workflows, observers)
 	jobConfigurer := rehearse.NewJobConfigurer(prConfig.CiOperator, resolver, prNumber, loggers, rehearsalTemplates.Names, rehearsalClusterProfiles.Names, jobSpec.Refs)
```
petr-muller, line 320: Note that `prConfig` also has `Prow` member, so you do not need to load it separately by `loadProwConfig` and therefore you don't even need that method. The convention in `pj-rehearse` is that we have complicated structs (that carry the "context" of the rehearsals)  that have simpler methods (that operate moslty over just jobs to be rehearsed).

So ideally `NewJobConfigurer` should either receive `prConfig.Prow` (I assume you want to use the PR version of the config, not the baseline) and save it into the struct, and `ConfigurePresubmitRehearsals` should continue to just take the jobs as an input.

smg247:
> Eek came here late, pj-rehearse already has Prow config (also, in pj-rehearse you need to take care whether you are using baseline or pr version of the config)

@petr-muller I'll rework this with that in mind. Do you think I need to revert this, or just a new PR will do?

PR #2673, opened 2022-02-14
Better retester: make a controller

petr-muller:
Some notable changes:
- pwned the fan-out in querying: retest will not need to query in parallel and I immediately hit rate limiting when testing
- I removed the `!orgs.Has(org)` which I do not understand why it was added
- Added some logging
- Added a `status:failure` into the query to only search failing PRs (only ever makes sense to retest failing PRs)

### DIFF
--- a/cmd/better-retester/main.go
+++ b/cmd/better-retester/main.go
@@ -3,22 +3,15 @@ package main
 import (
 	"context"
 	"flag"
-	"fmt"
 	"os"
-	"sync"
 	"time"
 
-	githubql "github.com/shurcooL/githubv4"
 	"github.com/sirupsen/logrus"
 
-	utilerrors "k8s.io/apimachinery/pkg/util/errors"
-	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/pkg/flagutil"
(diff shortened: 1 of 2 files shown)


cmd/better-retester/main.go
```diff
@@ -29,10 +22,10 @@ type options struct {
 	config configflagutil.ConfigOptions
 	github prowflagutil.GitHubOptions
 
-	dryRun bool
+	runOnce bool
+	dryRun  bool
 
-	orgRaw prowflagutil.Strings
-	orgs   sets.String
```
petr-muller, line 35: I ended removing this because I needed to remove the code that used it. There was a 

```
if !orgs.Has(org) {
  continue
}
```

guard clause in the `query` method which made the code skip all queries unless the organization was explicitly allowlisted by this parameter, and I did not understand its purpose.

hongkailiu: It was to mimik the current tester:

https://github.com/openshift/release/blob/1c62bf4564daeda670edae0f89730104583640ca/ci-operator/jobs/infra-periodics.yaml#L222-L223

I was thinking if we want to enable the retester for all repos or for no repos by default.

hongkailiu (approved):

PR #2675, opened 2022-02-15
better-retester: filter out PRs with failed required presubmits

hongkailiu:
/cc @petr-muller 

/hold

Need to make a kubeconfig for the SA which is used by the deployment.


petr-muller:
/hold

A lot of this code is not needed, and it causes useless dependencies. I strongly believe we do not need Prowjob client, and if we do not need Prowjob client, we do not need kube client, and then we do not need new kubeconfig...

petr-muller:
Check out https://github.com/openshift/ci-tools/pull/2677 that removed a lot of unneeded (at least for us) Tide code

PR #2672, opened 2022-02-14
Modify repo-init-apiserver logging to provide finer-grained info

smg247:
I need some more easily searchable logs to follow the steps and fix issues here.

PR #2676, opened 2022-02-15
Add committer info to repo-init-ui commit

smg247:
Without this info the commit fails.

I have also refactored `api.go` based on @droslean's suggestion in https://github.com/openshift/ci-tools/pull/2672.

### DIFF
--- a/cmd/repo-init/api.go
+++ b/cmd/repo-init/api.go
@@ -29,22 +29,21 @@ import (
 	"github.com/openshift/ci-tools/pkg/validation"
 )
 
-// l keeps the tree legible
-func l(fragment string, children ...simplifypath.Node) simplifypath.Node {
-	return simplifypath.L(fragment, children...)
-}
-
-var (
-	apiMetrics = metrics.NewMetrics("repo_init_api")
-
-	configTypes = []serverConfigType{GitHubClientId, GitHubClientSecret, GitHubRedirectUri}
-	// NOTE: this map should not be altered outside of the loadServerConfig function.
-	serverConfig = make(map[serverConfigType]string)
+type server struct {
+	// NOTE: this map should not be altered outside the loadServerConfig function.
+	serverConfig map[serverConfigType]string
 
 	githubOptions flagutil.GitHubOptions
 	disableCors   bool
 	rm            *repoManager
-)
+
+	logger *logrus.Entry
+}
+
+// l keeps the tree legible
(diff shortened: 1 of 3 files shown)


cmd/repo-init/api.go
```diff
@@ -228,10 +234,10 @@ func authHandler(log *logrus.Entry) http.HandlerFunc {
 	}
 }
 
-func clusterProfileHandler(log *logrus.Entry) http.HandlerFunc {
+func (s *server) clusterProfileHandler() http.HandlerFunc {
```
hongkailiu, line 237: I feel handler funcs should not be attached to the server.

cmd/repo-init/repo_manager.go
```diff
@@ -200,10 +199,19 @@ func commitChanges(message, email, name string) error {
 	if err := bumper.Call(os.Stdout, os.Stderr, "git", "add", "-A"); err != nil {
 		return fmt.Errorf("git add: %w", err)
 	}
-	commitArgs := []string{"commit", "-m", message}
-	if name != "" && email != "" {
-		commitArgs = append(commitArgs, "--author", fmt.Sprintf("%s <%s>", name, email))
+
+	if err := bumper.Call(os.Stdout, os.Stderr, "git", "config", "--local", "user.email", email); err != nil {
```
hongkailiu, line 203: Is this what this PR is about?
Can we set this up as a git globe setting, more properly in the image that this tool runs?
We probably did this already in our commonly used os image.

We have many tools run the git-push after git-commit.
I do not see anyone need git-config.
Do you notice the failure in production or locally?

smg247: Yes, this is the actual change in this PR. I thought about setting it in the image, but this way allows for the authorized user themselves to appear as the committer. I actually found the pattern in https://github.com/openshift/ci-tools/blob/master/pkg/github/prcreation/prcreation.go#L139-L148

cmd/repo-init/api.go
smg247: I may be missing something, but I think that is exactly what @droslean was suggesting I do in https://github.com/openshift/ci-tools/pull/2672#pullrequestreview-882975470.

hongkailiu: I get the logger part, but not the handler part: the functions do not use anything from the server's objects.
I guess no big deal except personal taste.

PR #2681, opened 2022-02-17
Update test-infra to 9703a3111217

alvaroaleman:
In order to pick up https://github.com/kubernetes/test-infra/pull/25305
and https://github.com/kubernetes/test-infra/pull/25301 so we wont break
when https://github.com/kubernetes/test-infra/pull/25281 merges.

/cc @petr-muller


petr-muller:
Wow, well spotted, Alvaro. I was not paying too much attention to that PR until now, and it would be quite spectacular dumpster fire if it merged. So thanks!

I will read up of course, but because you are already involved upstream, can you advise on the way how we can minimize risk in this whole process? We seem to be using pod-utils even in other, non-Prow places - afaik ci-operator uses `clonerefs` internally (and I remember we even have some derived images to deal with some binary path differences) to clone the code in `src`, and iirc we also use `sidecar` and `entrypoint` in pods spawned by ci-operator... Maybe other places.

I think we need to do something like:

1. Merge test-infra bump to chatbot and release-controller rpeos (to make the `validate-test-infra` check happy)
2. Merge this PR and check if everything works
3. Disable automated Prow bump (so that we have control over how and when we bump to new images)
4. Merge the upstream PR
5. Wait until new images are built
6. Run the Prow bumper once
7. Kick off the `managed-clonerefs` image build
8. Observe that everything works

Did I miss anything?

petr-muller:
Yeah - I think I only want to disable autobumping so that this does not land when nobody is looking, potentially waking up to a broken Prow or something. I agree that e2e should give a good enough signal.

petr-muller (approved):
/test validate-test-infra

PR #2682, opened 2022-02-17
better-retester: persist cache

hongkailiu:
/cc @petr-muller

### DIFF
--- a/cmd/better-retester/retester.go
+++ b/cmd/better-retester/retester.go
@@ -2,29 +2,99 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"strings"
 	"time"
 
 	githubql "github.com/shurcooL/githubv4"
 	"github.com/sirupsen/logrus"
 
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/github"
 	"k8s.io/test-infra/prow/tide"
+	"sigs.k8s.io/yaml"
 )
 
-type pullRequest struct {
-	prSha             string
-	baseSha           string
-	retestsForPrSha   int
-	retestsForBaseSha int
+type PullRequest struct {
+	PRSha              string      `json:"pr_sha,omitempty"`
+	BaseSha            string      `json:"base_sha,omitempty"`
+	RetestsForPrSha    int         `json:"retests_for_pr_sha,omitempty"`
+	RetestsForBaseSha  int         `json:"retests_for_base_sha,omitempty"`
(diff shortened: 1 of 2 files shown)


cmd/better-retester/retester.go
```diff
@@ -81,10 +135,9 @@ func (c *retestController) sync() error {
 		errs = append(errs, c.retestOrBackoff(pr))
 	}
 
-	// TODO: Load the backoff records from disk after start (using an option) and survive when it is not present
-	// TODO: Persist the backoff records after each sync (atomically, so that we do not corrupt it on interrupt)
```
hongkailiu, line 85: atomically: not sure if I did it right.

hongkailiu: Google a bit.
It should be atomic now.

cmd/better-retester/retester.go
```diff
@@ -2,29 +2,95 @@ package main
...
+	}
+	// write to a temp file and rename it to the cache file to ensure "atomic write":
+	// either it is complete or nothing
+	tmpFile, err := ioutil.TempFile("", "backoffCache")
+	if err != nil {
+		return fmt.Errorf("failed to create a temp file: %w", err)
+	}
+	tmp := tmpFile.Name()
+	defer func() {
+		if err := os.Remove(tmp); err != nil {
```
petr-muller, line 82: Won't this error out by default when we succeed on `os.Rename` afterwards? After that the `tmp` path won't exist anymore.

cmd/better-retester/retester.go
```diff
@@ -54,11 +121,14 @@ func prUrl(pr tide.PullRequest) string {
 }
 
 func (c *retestController) sync() error {
+	if err := c.backoff.loadFromDisk(); err != nil {
```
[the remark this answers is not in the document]
hongkailiu: Updated.

cmd/better-retester/retester.go
hongkailiu: Updated.

cmd/better-retester/retester.go
```diff
@@ -2,29 +2,95 @@ package main
...
+	bytes, err := ioutil.ReadFile(b.file)
+	if err != nil {
+		return fmt.Errorf("failed to read file %s: %w", b.file, err)
+	}
+	cache := map[string]*PullRequest{}
+	if err := yaml.Unmarshal(bytes, &cache); err != nil {
+		return fmt.Errorf("failed to unmarshal: %w", err)
+	}
+	for key, pr := range cache {
+		if age := metav1.Now().Sub(pr.LastConsideredTime.Time); age > b.cacheRecordAge {
```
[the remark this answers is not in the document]
hongkailiu: Saw the comment on it:

https://github.com/kubernetes/apimachinery/blob/276a8a7530a30f42c82b1f0fa2261a905130bf97/pkg/apis/meta/v1/time.go#L24-L33

Keep the same way to marshal/unmarchal as k8s.
an bad idea?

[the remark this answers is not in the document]
hongkailiu: Sorry. Misunderstood your question.
Updated.

petr-muller (approved):
sweet

PR #2688, opened 2022-02-21
repo init ui: run 'make jobs' prior to commit

smg247:
We need to generate the jobs prior to the commit/pr creation. I also thought of utilizing the the tools that `make jobs` runs inside the repo-init binary itself, but it is doing a lot and most of it would need to be refactored to be accessible here. Ultimately, I decided that the `make jobs` target is better to use as it will remain up to date and is the way that we currently onboard a new repo.

We also need to validate that the tests execute in an image that we have specified so that context needs to be built up for the validation.

Since `make jobs` takes some time I added an info message to make it more user friendly.

### DIFF
--- a/cmd/repo-init/api.go
+++ b/cmd/repo-init/api.go
@@ -422,7 +422,6 @@ func unmarshalValidationRequest(data []byte) (validationType, interface{}, error
 func (s server) validateConfig(w http.ResponseWriter, r *http.Request) {
 	logger := s.logger.WithField("handler", "configValidationHandler")
 	body, err := ioutil.ReadAll(r.Body)
-
 	if err != nil {
 		logger.WithError(err).Error("Error while reading request body")
 		w.WriteHeader(http.StatusInternalServerError)
@@ -436,6 +435,7 @@ func (s server) validateConfig(w http.ResponseWriter, r *http.Request) {
 		_, _ = w.Write([]byte("Invalid validation request"))
 		return
 	}
+	logger.WithField("validationType", validationType).Debug("validating the config")
 
 	var validationErrors []error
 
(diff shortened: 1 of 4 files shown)


smg247:
/cc @bbguimaraes
I see that you worked on the `validation/graph.go` that I am utilizing here.

cmd/repo-init/api.go
```diff
@@ -604,6 +629,14 @@ func (s server) generateConfig(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	s.logger.Debug("running 'make jobs' prior to commit")
+	makeJobs := exec.Command("make", "jobs")
```
petr-muller, line 633: Will this work though when we run this inside a Pod? `make jobs` runs `podman run`...

smg247: No, I suppose it will not work...I wasn't actually aware of that.
I think this means the only way to achieve this would be a refactor of: `ci-operator-prowgen`, `sanitize-prow-jobs`, and `ci-operator-checkconfig` to allow us to directly use them within the `repo-init` binary. Do you see any way around that?

smg247: Or do we simply need to include `podman` in the dockerfile for `repo-init`? not sure if that would solve the issue...

smg247: Or does that just mean we need to include `podman` in the `repo-init` dockerfile?

smg247: I went ahead and added `podman` to the dockerfile. If we need to go the other route I am going to do an entirely new PR and close this one.

petr-muller: I never knew kube/ocp well enough to know whether you can successfuly run `podman run` in containers ;) If it does not work, we could do what `autoconfigbrancher` does:

1. Inject the necessary binaries into the `repo-init-ui` image ([link](https://github.com/openshift/release/blob/d1118ae65699cab930f93c741121bdd1c2305874/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml#L182-L210))
2. Execute the binaries themselves ([link](https://github.com/openshift/ci-tools/blob/3a9d6085e8ea08c322794d9e27084e813c52e647/cmd/autoconfigbrancher/main.go#L197-L203))

petr-muller (approved):

PR #2690, opened 2022-02-21
autoowners: stop putting fetch time into header

petr-muller:
Technically a nice idea, but it results in a daily super-churn: https://github.com/openshift/release/pull/26399/files

I considered to make the tool smarter (compare the existing OWNERS files and only change the header when the content fails) but because we have two types of OWNERS files, we need to handle cases like files missing and varius errors etc, I decided it is not worth the hassle. The information about when the file was last changed will be stored in `git` so putting it to header is just convenience.

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -263,10 +263,10 @@ func writeOwners(orgRepo orgRepo, httpResult httpResult, cleaner ownersCleaner, 
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


petr-muller:
Wow. Removed an item from `Sprintf`, failed two CI jobs. Very senior. Much experienced.

/woof

hongkailiu (approved):

hongkailiu:
> so putting it to header is just convenience.

Convenience comes with a price. 😄

PR #2695, opened 2022-02-24
determinize-prow-config: use strict mode to fail when there is an unknown key

smg247:
In order to not have a situation where running `make prow-config` causes us to silently lose configuration due to improper indentation or misnaming a key we should load the config in strict mode.

There is an existing function in test-infra to do this, but `agent.Start` doesn't utilize it. I considered adding a flag to that method to trigger that behavior, but we don't need to be using `agent.Start` here anyways. We can just load the config once, in strict mode, and then let the rest of the functionality follow as normal.


petr-muller:
Sweet!

PR #1007, opened 2020-07-17
e2e: add a test that runs a template

stevekuznetsov:
test: pare down template

Remove superfluous parts of the template and add in executions that will
let us inspect the output of the template by comparing output artifacts.

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

test: update expected integration output

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

e2e: add a test that runs a template

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

/assign @bbguimaraes @petr-muller

PR #2696, opened 2022-02-28
cmd/ci-operator: Trust suite names in writeJUnit

wking:
Way back in 12c1a0d704, it seemed like there might be a single suite, so clobbering the name with the writeJUnit argument was reasonable.  But 26ebc68367 (#1007) added sorting code that only makes sense if some JUnit may contain multiple suites.  Only clobbering the name on whichever suite happened to be first (before sorting!) doesn't seem like an internally-consistent approach.

With this commit, I'm pivoting to consistently trust the suite data to contain names, and I've updated any suites I've found in the code to declare those names as well.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -1548,6 +1548,7 @@ func (o *options) writeFailingJUnit(errs []error) {
 	suites := &junit.TestSuites{
 		Suites: []*junit.TestSuite{
 			{
+				Name:      "job",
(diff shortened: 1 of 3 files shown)


bbguimaraes:
I don't think we ever report more than one suite.

The current naming seems very arbitrary (why "operator" on success but "job" on (a certain type of) failure?), so I don't see a problem with making this change.

/lgtm

PR #2697, opened 2022-03-01
pkg/steps/multi_stage: Add subTests for phase completion

wking:
Make it easier to see "all the pre steps completed", etc., by declaring steps/test-cases for each phase (pre, test, and post).

### DIFF
--- a/pkg/steps/multi_stage.go
+++ b/pkg/steps/multi_stage.go
@@ -152,12 +152,12 @@ func (s *multiStageTestStep) run(ctx context.Context) error {
 		return err
 	}
 	var errs []error
-	if err := s.runSteps(ctx, s.pre, env, true, false, secretVolumes, secretVolumeMounts); err != nil {
+	if err := s.runSteps(ctx, "pre", s.pre, env, true, false, secretVolumes, secretVolumeMounts); err != nil {
 		errs = append(errs, fmt.Errorf("%q pre steps failed: %w", s.name, err))
-	} else if err := s.runSteps(ctx, s.test, env, true, len(errs) != 0, secretVolumes, secretVolumeMounts); err != nil {
+	} else if err := s.runSteps(ctx, "test", s.test, env, true, len(errs) != 0, secretVolumes, secretVolumeMounts); err != nil {
 		errs = append(errs, fmt.Errorf("%q test steps failed: %w", s.name, err))
 	}
-	if err := s.runSteps(context.Background(), s.post, env, false, len(errs) != 0, secretVolumes, secretVolumeMounts); err != nil {
+	if err := s.runSteps(context.Background(), "post", s.post, env, false, len(errs) != 0, secretVolumes, secretVolumeMounts); err != nil {
 		errs = append(errs, fmt.Errorf("%q post steps failed: %w", s.name, err))
 	}
 	return utilerrors.NewAggregate(errs)
@@ -388,6 +388,7 @@ func commandConfigMapForTest(testName string) string {
 
 func (s *multiStageTestStep) runSteps(
 	ctx context.Context,
+	phase string,
 	steps []api.LiteralTestStep,
 	env []coreapi.EnvVar,
 	shortCircuit bool,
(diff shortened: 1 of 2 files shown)


pkg/steps/multi_stage.go
```diff
@@ -431,6 +434,20 @@ func (s *multiStageTestStep) runSteps(
...
+	phaseStep := api.CIOperatorStepDetailInfo{
+		StepName:    fmt.Sprintf("%s phase", phase),
+		Description: fmt.Sprintf("The collected steps of multi-stage phase %s.", phase),
+		StartedAt:   &start,
+		FinishedAt:  &finished,
+		Duration:    &duration,
+		Failed:      utilpointer.BoolPtr(len(errs) > 0),
+	}
+	s.subSteps = append(s.subSteps, phaseStep)
+	s.subTests = append(s.subTests, notifier.SubTests(phaseStep.StepName)...)
```
bbguimaraes, line 449: You don't need these gymnastics.  `TestContainerNotifier` will clear the list reported by `SubTests` after each call in `runPod`, so this list will always be empty:

https://github.com/openshift/ci-tools/blob/a56c10fe8957eb197c70037d304d35f9089ef700/pkg/steps/artifacts.go#L110-L113

Unless you intended to also report them under each phase, in which case you'll need to create a new type of notifier and/or accumulate the lists manually.

bbguimaraes (commented):
The addition of the `phase` argument to `runSteps` looks good.  The notifier change probably does not do what is intended.

If you want to move the first part to a separate PR, it can go in immediately.  One extra thing I'd added (I've had a branch with an almost identical change sitting here forever) was a log message at the top of `runSteps` saying something to the effect of 'Running multi-stage phase "$phase"', useful if you're not familiar with the test/workflow used.

pkg/steps/multi_stage.go
wking: I was unclear on the distinction between `subSteps` and `subTests`, because they sound fairly similar to me, and the existing `runPod` code seemed to append to both together, and neither `subSteps` nor `SubSteps` seems to have godocs anywhere.  I haven't traced out `subSteps` consumption; maybe I can leave my phase test-cases out of `subSteps`?

For `subTests`, my goal is just to get these into the JUnit.  Without a notifier, should I just create a `*junit.TestCase` directly here?

bbguimaraes: `subSteps` is the list that ultimately populates `ci-operator-step-graph.json` in the artifacts.  It was the beginning of an effort to provide a better visualization in Deck, but didn't progress much.  It's mostly informative (for anyone who knows where to look).

`subTests` is what ends up in the JUnit output.

You can see both in the E2E tests, e.g.:
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ci-tools/2697/pull-ci-openshift-ci-tools-master-e2e/1498455470455656448/artifacts/e2e/e2e/artifacts/TestMultiStage_step_with_best_effort_post-step/ci-operator-step-graph.json
- https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ci-tools/2697/pull-ci-openshift-ci-tools-master-e2e/1498455470455656448/artifacts/e2e/e2e/artifacts/TestMultiStage_step_with_best_effort_post-step/_junit_operator.xml

bbguimaraes: `subSteps` is a simple list we generate directly here, it exists only in multi-stage tests.  `subTests` is more general and is used by many types of (`ci-operator`) steps.  For test steps, it is generated by these notifiers, which coordinate with the execution code to monitor the test pod.

(this needs to be documented somewhere, I'm taking notes)

bbguimaraes: The problem here is the notifier expects `SubTests` to be called once per pod, it clears its internal state on each call.  So you will get the JUnit information for each step in `runPod`, but the later call for the phase you are adding in `runSteps` will always be empty.  Depending on what you want the format of the phase in JUnit to be, it might be enough to just manufacture a `TestCase` instead of using the notifier.

PR #2703, opened 2022-03-03
censor access tokens in repo-init

smg247:
We noticed that the access token was being logged when pushing the commit at the end of the repo-init flow.
Now that log looks like `{"args":"['-C', '/var/folders/8l/vb1frxqj6r54xmrswnvctpm40000gn/T/repo-manager-release2106839879', 'push', '-f', 'https://smg247:****************************************@github.com/smg247/release.git', 'HEAD:new-ci-config-1646332911']","cmd":"git","component":"unset","level":"info","msg":"running command","severity":"info","time":"2022-03-03T13:41:51-05:00"}`

### DIFF
--- a/cmd/repo-init/api.go
+++ b/cmd/repo-init/api.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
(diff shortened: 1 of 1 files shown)


smg247:
> "Logged" means we potentially pushed that to AWS log storage, but we did not show this to any user or exposed it to the world, right?

Yes, thats correct

hongkailiu (approved):

PR #2707, opened 2022-03-07
ci-operator: improve error message for invalid `.ci-operator.yaml`

petr-muller:
Improve (make explicit) the error message given when `.ci-operator.yaml` is present and is a valid YAML, but not the yaml we expect

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/hashicorp/go-retryablehttp"
+	"github.com/openshift/ci-tools/pkg/validation"
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


bbguimaraes:
A slightly pedantic comment, but code like this usually goes in `pkg/validation`.

petr-muller:
@bbguimaraes Seems like doing that introduces a nasty import cycle, because `pkg/validation` *tests* already import `pkg/defaults` to create the test inputs to be validated:

https://github.com/openshift/ci-tools/blob/c5d30c1fc1fba53702be99863fbfeefc308fa4b5/pkg/validation/graph_test.go#L164

How opposed are you to the original version where the validation code was part of `pkg/defaults`. No easy solution comes to mind for breaking the import cycle.

bbguimaraes:
> How opposed are you to the original version where the validation code was part of `pkg/defaults`. No easy solution comes to mind for breaking the import cycle.

Works for me, we can think of a better solution if/when this functionality needs to be used somewhere else.

PR #2709, opened 2022-03-08
local-slack-bot: use jira bearer token, and modify certs to work on mac

smg247:
The jira-client now requires a bearer token rather than user and password, so we need to update the local script to adjust for that.

I also couldn't add the cert in the same manner on my mac so I created a mac specific path.

### DIFF
--- a/hack/local-slack-bot.sh
+++ b/hack/local-slack-bot.sh
@@ -29,7 +29,7 @@ function OC() {
 
 os::log::info "Extracting production data we need to run the server..."
 OC extract secret/slack-credentials-dptp-bot-alpha --keys oauth_token,signing_secret --to "${data}"
-OC extract secret/jira-credentials-dptp-bot --keys password --to "${data}"
+OC extract secret/jira-token-dptp-bot --keys token --to "${data}"
 
 os::log::info "Setting up the regular proxy for outgoing traffic..."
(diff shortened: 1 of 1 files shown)


smg247:
/cc @psalajova
for visibility

hack/local-slack-bot.sh
```diff
@@ -44,8 +44,13 @@ done
...
 os::log::info "Installing the proxy's certificate..."
 http_proxy=localhost:7777 https_proxy=localhost:7777 wget --output-document "${data}/mitm.pem" mitm.it/cert/pem
-sudo cp "${data}/mitm.pem" /etc/pki/ca-trust/source/anchors/
-sudo update-ca-trust
+
+if [ "$(uname -s)" != "Darwin" ]; then
+  sudo cp "${data}/mitm.pem" /etc/pki/ca-trust/source/anchors/
+  sudo update-ca-trust
+else # special logic for MacOS
+  sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain "${data}/mitm.pem"
```
hongkailiu, line 52: nit: `sudo` looks demanding (and I do not see `remove-trusted-cert` in tear down) but it was there before the PR.
Local dev. env. is hard. I know.

hongkailiu (approved):

PR #2710, opened 2022-03-08
Update hive api so resumeTimeout doesn't get cleared

alvaroaleman:
It is currently being removed by the clusterimageset-updater, e.G. here:
https://github.com/openshift/release/commit/13330bf9b7fffe0fa46f5cfef1aacb49730517fc

/cc @petr-muller @hongkailiu


test/integration/clusterimageset-updater/output/pools/4-5-auto_clusterpool.yaml
```diff
@@ -15,6 +15,7 @@ metadata:
   namespace: ci-cluster-pool
 spec:
   baseDomain: hive.aws.ci.openshift.org
+  hibernationConfig: null
```
hongkailiu, line 18: Maybe add something like the following to the input to ensure it does not get removed in the output?

```
  hibernationConfig:
    resumeTimeout: 20m
```

alvaroaleman: Added

hongkailiu (approved):
Thanks for the fix.

PR #2713, opened 2022-03-09
sprint-automation: add created and last updated to the card digests

smg247:
These dates will be helpful in the intake role to recognize issues that you have already viewed, but were recently updated and need another look.
They will be helpful in the "Cards Awaiting Acceptance" section so we have a good idea of how long the card has been waiting for by looking at the `Last updated`

### DIFF
--- a/cmd/sprint-automation/main.go
+++ b/cmd/sprint-automation/main.go
@@ -562,6 +562,8 @@ func sendTriageBuild02Upgrade(slackClient *slack.Client, version, stableDuration
 	return nil
 }
 
+const dateFormat = "Mon, 02 Jan 2006"
+
 func blockForIssue(issue jiraapi.Issue) slack.Block {
 	// we really don't want these things to line wrap, so truncate the summary
 	cutoff := 85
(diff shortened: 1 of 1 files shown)


cmd/sprint-automation/main.go
```diff
@@ -562,20 +562,28 @@ func sendTriageBuild02Upgrade(slackClient *slack.Client, version, stableDuration
...
 		ContextElements: slack.ContextElements{
 			Elements: []slack.MixedElement{
 				&slack.TextBlockObject{
 					Type: slack.MarkdownType,
-					Text: fmt.Sprintf("<https://issues.redhat.com/browse/%s|*%s*>: %s", issue.Key, issue.Key, summary),
+					Text: fmt.Sprintf("<https://issues.redhat.com/browse/%s|*%s*>: %s \n", issue.Key, issue.Key, summary),
+				},
+				&slack.TextBlockObject{
+					Type: slack.PlainTextType,
+					Text: fmt.Sprintf("Created on: %s  Last updated: %s", created, updated),
```
hongkailiu, line 586: Is it easy to get the author who did the update?
Do we want to avoid robot updates?

smg247: We do have access to the `changelog` so we could theoretically ignore updates by certain users in this digest. I do wonder where robot updates fit into our workflow though...would we want to not surface those as an updated card? For intake, would we not want to see that there had been a change just because it was a robot change?

I could also potentially add the date of the most recent comment here (assuming one exists), but the overall goal was just to surface things that we **know** haven't changed since we have last looked.

smg247: /hold cancel
I will look into adding the `updateBy`, but I will do it in a follow up PR. I am intake this week, and I want to see this in action tomorrow :)

PR #2708, opened 2022-03-08
Extract sort logic for SelectJobsForChangedRegistry

nesty156:
Extract sort logic and created tests
First part of : [DPTP-2743](https://issues.redhat.com/browse/DPTP-2743)

PR #2714, opened 2022-03-09
Creating more relevant sorting logic for SelectJobsForChangedRegistry

nesty156:
Creating logic for moreRelevant method based on tests created in [this PR](https://github.com/openshift/ci-tools/pull/2708)
Second part of : [DPTP-2743](https://issues.redhat.com/browse/DPTP-2743)

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -57,6 +57,22 @@ type Loggers struct {
 	Job, Debug logrus.FieldLogger
 }
 
+// Number of openshift versions
+var numVersion = 50
+
+// Global map that contains relevance of known branches
+var relevancy = map[string]int{
+	"master": numVersion + 1,
+	"main":   numVersion + 1,
+}
+
+func init() {
+	for i := 1; i < numVersion; i++ {
+		relevancy[fmt.Sprintf("release-4.%d", i)] = i
+		relevancy[fmt.Sprintf("openshift-4.%d", i)] = i
+	}
+}
+
 // NewProwJobClient creates a ProwJob client with a dry run capability
 func NewProwJobClient(clusterConfig *rest.Config, dry bool) (ctrlruntimeclient.Client, error) {
 	if dry {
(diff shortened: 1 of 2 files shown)


pkg/rehearse/jobs.go
```diff
@@ -714,7 +714,32 @@ func SelectJobsForChangedRegistry(regSteps []registry.Node, allPresubmits presub
...
+		if strings.HasPrefix(branchJ, "release-4.") || strings.HasPrefix(branchJ, "openshift-4.") {
+			numberI := strings.TrimPrefix(branchI, "release-4.")
+			if numberI == branchI {
+				numberI = strings.TrimPrefix(branchI, "openshift-4.")
+			}
+			numberJ := strings.TrimPrefix(branchJ, "release-4.")
+			if numberJ == branchJ {
+				numberJ = strings.TrimPrefix(branchJ, "openshift-4.")
+			}
+			intI, _ := strconv.Atoi(numberI)
```
petr-muller, line 731: No matter what, you should not ignore errors - we need to correctly handle a case like `release-4.x`

pkg/rehearse/jobs.go
```diff
@@ -714,7 +714,32 @@ func SelectJobsForChangedRegistry(regSteps []registry.Node, allPresubmits presub
...
+			if intI > intJ {
+				return true
+			} else {
+				return false
+			}
+		} else {
+			return true
+		}
+	}
+	return branchI > branchJ
```
petr-muller, line 742: I was thinking about this, and I think we'll want to optimize this - because it will be used in a less function of a sort algorithm, it will potentially be called many times, so we can try being a bit smart about this. How about pregenerating  a map from a known branch names to numbers? You would do something like this once:

```go
branches := map[string]int{}
for i := 1; i < 50; i++ {
  branches[fmt.Sprintf("release-4.%d", i)] = i
  branches[fmt.Sprintf("openshift-4.%d", i)] = i
}
```

And then you'd only used that map instead of parsing the strings all the time. You could go even further, add `main` and `master` mapped to `100` or something. And you could even more further, consider the "not in map" case (which will return the default value of `0`) too. Then you could reduce the whole `moreRelevant` to a comparison of integers fetched from a map.

nesty156: > Then you could reduce the whole moreRelevant to a comparison of integers fetched from a map.

I don't know how to compare two values ​​with 0. So far, I'm just submitting your suggestions for optimizing the moreRelevant method.

pkg/rehearse/jobs.go
```diff
@@ -714,7 +714,19 @@ func SelectJobsForChangedRegistry(regSteps []registry.Node, allPresubmits presub
...
 
 func moreRelevant(one, two *config.DataWithInfo) bool {
-	return one.Info.Filename > two.Info.Filename
+	branchI := one.Info.Metadata.Branch
+	branchJ := two.Info.Metadata.Branch
+	branches := map[string]int{}
+	for i := 1; i < 50; i++ {
+		branches[fmt.Sprintf("release-4.%d", i)] = i
+		branches[fmt.Sprintf("openshift-4.%d", i)] = i
+	}
```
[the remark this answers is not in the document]
nesty156: In Implementation Details of [DPTP-2743](https://issues.redhat.com/browse/DPTP-2743)
I have following bullet point:

- the the banch does not match, just prefer the "higher" based on string comparison
Now in my implementation it works by comparing 0 to 0

pkg/rehearse/jobs.go
nesty156: In Implementation Details of [DPTP-2743](https://issues.redhat.com/browse/DPTP-2743)
I have following bullet point:

- the the banch does not match, just prefer the "higher" based on string comparison
Now in my implementation it works by comparing 0 to 0

petr-muller (approved):

petr-muller:
Thank you!

PR #2715, opened 2022-03-10
repo-init: create repo's prow configuration

hongkailiu:
https://issues.redhat.com/browse/DPTP-2759

/cc @openshift/test-platform

### DIFF
--- a/cmd/repo-init/main.go
+++ b/cmd/repo-init/main.go
@@ -17,7 +17,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/docker/docker/pkg/fileutils"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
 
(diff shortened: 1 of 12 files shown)


bbguimaraes:
I know this is WIP, but I tried going quickly through the changes and, after reading the Jira issue and the commits, it's still not obvious to me:

- Why the program stopped working as expected (or even if it did).
- How this change addresses whatever problems the implementation has.

A short description that answers these questions would make going through the code much easier.

hongkailiu:
> * Why the program stopped working as expected (or even if it did).

Once we started to shard the Prow config to org/repo sub-folder structure, the tide config for the new repo stopped working.

> * How this change addresses whatever problems the implementation has.

Let me reply this with comment on the relevant code.

cmd/repo-init/main.go
```diff
@@ -462,25 +462,46 @@ func fetchOrDefaultWithPrompt(msg, def string) string {
...
+		if _, err := os.Stat(src); errors.Is(err, os.ErrNotExist) {
+			continue
+		}
+		if _, err := fileutils.CopyFile(src, path.Join(tmpDir, fmt.Sprintf("%d%s", i, ciopconfig.SupplementalProwConfigFileName))); err != nil {
+			return fmt.Errorf("failed to copy file: %w", err)
+		}
+	}
+
 	agent := prowconfig.Agent{}
-	if err := agent.Start(configPath, "", nil, ""); err != nil {
```
hongkailiu, line 468: The prowconfigs for org/repo have been sharded into its own folders.
But we load only the global one.
So we did not have the config for `openshift/cluster-version-operator` and `openshift/ci-tools`.

This PR copies them to a tmp folder and load them as supplemental configs.

cmd/repo-init/main.go
```diff
@@ -462,25 +462,46 @@ func fetchOrDefaultWithPrompt(msg, def string) string {
...
 
 	prowConfig := agent.Config()
-	editProwConfig(prowConfig, config)
 
-	data, err := yaml.Marshal(prowConfig)
-	if err != nil {
-		return fmt.Errorf("could not marshal Prow configuration: %w", err)
-	}
-
-	return ioutil.WriteFile(configPath, data, 0644)
```
hongkailiu, line 480: We write the updated one into the globe one.
Because we never updated anything, we write the same config back to the global config.
Otherwise, we would notice the bug earlier.

The PR writes the config for the new repo into its own folder instead.

test/integration/repo-init/expected/core-services/prow/02_config/_config.yaml
```diff
@@ -45,49 +45,45 @@ tide:
   max_goroutines: 20
   queries:
   - includedBranches:
-    - openshift-4.1
```
hongkailiu, line 48: The changes on the global one should not happen happened, as expected: The config for the new repo is in its own folder.

test/integration/repo-init/expected/core-services/prow/02_config/org/other/_prowconfig.yaml
```diff
@@ -0,0 +1,21 @@
+tide:
```
hongkailiu, line 1: The sharded one has been generated as expected.

cmd/repo-init/main.go
bbguimaraes: Is this done because the cost of loading all the files is prohibitive?

We have `release.go` in `pkg/config/` for shared code that operates on `openshift/release`, would it make sense to put the loading code there so it doesn't diverge again in the future?  It would be preferable to leaving the possibility for a future bug like this, if the cost is not too high.

bbguimaraes:
> Once we started to shard the Prow config to org/repo sub-folder structure, the tide config for the new repo stopped working.

Thanks, that is exactly the information I was looking for.  Left one comment inline.

cmd/repo-init/main.go
hongkailiu: Modification:
- Loading all of them now.
- Moving the loading code to `release.go`.

PR #2720, opened 2022-03-14
return correct ClusterType for QE's cluster profiles

jianlinliu:
Some following fix for https://github.com/openshift/ci-tools/pull/2716

PR #2721, opened 2022-03-14
provide different secrets for QE dedicated cluster profiles

jianlinliu:
Follow up https://github.com/openshift/ci-tools/pull/2720, after testing, my prow job is still using the `azure4` secrets, the installation did not land into QE dedicated cluster profiles.


droslean (approved):

staebler:
> @patrickdillon @staebler pls help review, also cc @yufchang

I do not know enough about this repo to understand exactly what this change is doing. I cannot offer a meaningful review.

jianlinliu:
> use an existing cluster profile (azure4)
Per my understanding, QE's cluster profile name is `azure-qe`, while using the same cluster type - `azure4`

PR #2722, opened 2022-03-14
refine PD queries to look at specific times of the day to solve monday morning issue with US secondary

smg247:
Fixes: https://issues.redhat.com/browse/DPTP-2403

I also moved the call to `ensureGroupMembership` up, because a failure to send one of the other messages doesn't need to prevent that from happening.

### DIFF
--- a/cmd/sprint-automation/main.go
+++ b/cmd/sprint-automation/main.go
@@ -123,6 +123,10 @@ func main() {
 		logrus.WithError(err).Fatal("Could not post team digest to Slack.")
 	}
 
+	if err := ensureGroupMembership(slackClient, userIdsByRole); err != nil {
+		logrus.WithError(err).Fatal("Could not ensure Slack group membership.")
+	}
+
 	if err := sendIntakeDigest(slackClient, jiraClient, userIdsByRole[roleIntake]); err != nil {
 		logrus.WithError(err).Fatal("Could not post @dptp-intake digest to Slack.")
 	}
@@ -131,10 +135,6 @@ func main() {
 		if err := sendNextWeeksRoleDigest(pagerDutyClient, slackClient); err != nil {
 			logrus.WithError(err).Fatal("Could not post next week's role digest to Slack.")
 		}
-	}
-
-	if err := ensureGroupMembership(slackClient, userIdsByRole); err != nil {
-		logrus.WithError(err).Fatal("Could not ensure Slack group membership.")
 	}
 
 	if err := addSchemes(); err != nil {
@@ -245,6 +245,9 @@ func users(client *pagerduty.Client, slackClient *slack.Client) (map[string]stri
 func usersOnCallAtTime(client *pagerduty.Client, slackClient *slack.Client, when time.Time) (map[string]string, []error) {
 	var errors []error
 	userIdsByRole := map[string]string{}
+	// 7 am UTC is when our PD day begins, and US on-call ends at 10pm UTC. Query 8 am - 9 pm for safe results
+	dayStart := time.Date(when.Year(), when.Month(), when.Day(), 8, 0, 1, 0, time.UTC)
+	dayEnd := dayStart.Add(13 * time.Hour).Add(-2 * time.Second)
 	for _, item := range []struct {
 		role         string
 		query        string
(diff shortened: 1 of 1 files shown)


cmd/sprint-automation/main.go
```diff
@@ -245,6 +245,9 @@ func users(client *pagerduty.Client, slackClient *slack.Client) (map[string]stri
 func usersOnCallAtTime(client *pagerduty.Client, slackClient *slack.Client, when time.Time) (map[string]string, []error) {
 	var errors []error
 	userIdsByRole := map[string]string{}
+	// 7 am UTC is when our PD day begins, and US on-call ends at 10pm UTC. Query 8 am - 9 pm for safe results
```
hongkailiu, line 248: > US on-call ends at 10pm UTC.

Is this the calculation?
18:00 (EDT) = 22:00 (UTC)

Do we need to modify the query again if we have a teammate from PDT?
I guess not since PDT starts work as well at range of the current query.

cmd/sprint-automation/main.go
```diff
@@ -245,6 +245,9 @@ func users(client *pagerduty.Client, slackClient *slack.Client) (map[string]stri
 func usersOnCallAtTime(client *pagerduty.Client, slackClient *slack.Client, when time.Time) (map[string]string, []error) {
```
hongkailiu, line 245: nit:

```suggestion
func usersOnCallAtTime(client *pagerduty.Client, slackClient *slack.Client, year, month, day int) (map[string]string, []error) {
```

cmd/sprint-automation/main.go
```diff
@@ -253,32 +256,32 @@ func usersOnCallAtTime(client *pagerduty.Client, slackClient *slack.Client, when
...
+			since: dayStart,
+			until: dayEnd,
 		},
 		{
 			role:  roleHelpdesk,
 			query: primaryOnCallQuery,
-			since: when.Add(-7 * 24 * time.Hour).Add(-1 * time.Second),
-			until: when.Add(-7 * 24 * time.Hour),
+			since: dayStart.Add(-7 * 24 * time.Hour),
+			until: dayStart.Add(-7 * 24 * time.Hour).Add(1 * time.Second),
```
hongkailiu, line 278: to keep the consistency:

```suggestion
			until: dayEnd.Add(-7 * 24 * time.Hour),
```

cmd/sprint-automation/main.go
```diff
@@ -253,32 +256,32 @@ func usersOnCallAtTime(client *pagerduty.Client, slackClient *slack.Client, when
...
 		},
 		{
 			role:  roleIntake,
 			query: primaryOnCallQuery,
-			since: when.Add(-2 * 7 * 24 * time.Hour).Add(-1 * time.Second),
-			until: when.Add(-2 * 7 * 24 * time.Hour),
+			since: dayStart.Add(-2 * 7 * 24 * time.Hour),
+			until: dayStart.Add(-2 * 7 * 24 * time.Hour).Add(1 * time.Second),
```
hongkailiu, line 284: to keep the consistency:

```suggestion
			until: dayEnd.Add(-2 * 7 * 24 * time.Hour),
```

hongkailiu:
> I also moved the call to `ensureGroupMembership` up, because a failure to send one of the other messages doesn't need to prevent that from happening.

Could you please elaborate? e.g., prevent what from happening?

smg247:
> > I also moved the call to `ensureGroupMembership` up, because a failure to send one of the other messages doesn't need to prevent that from happening.
> 
> 
> 
> Could you please elaborate? e.g., prevent what from happening?

Prevent the `ensureGroupMembership` method from running. That's why we had to manually set the helpdesk and triage aliases this morning.

cmd/sprint-automation/main.go
smg247: Yeah, I just got that value from looking at the L3 support schedule on PD. EU starts at 7 am UTC, and US ends at 10pm UTC (so 6pm eastern). Even when we had team members on the west coast those hours were the same, but if that were to change in the future the query should still work as is. The point of it is just to make sure that we hit at least a second in all of the possible blocks (currently EU and US) to pick everything up.

hongkailiu (approved):

PR #2728, opened 2022-03-17
Add analyze-required-install to analyze installation status across jo…

xueqzhan:
…bs for a payload

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
@@ -12,6 +12,8 @@ import (
 const (
 	// AggregationIDLabel is the name of the label for the aggregation id in prow job
 	AggregationIDLabel = "release.openshift.io/aggregation-id"
+	// PayloadInvocationIDLabel is the name of the label for the payload invocation id in prow job
+	PayloadInvocationIDLabel = "release.openshift.io/aggregation-id"
 )
 
 func NewPayloadAnalysisJobLocatorForPR(
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
@@ -11,6 +11,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/clock"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
+	"github.com/openshift/ci-tools/pkg/junit"
 )
 
 type JobRunGetter interface {
@@ -119,4 +120,26 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
 	fmt.Printf("found %d finished jobRuns: %v and %d unfinished jobRuns: %v\n",
 		len(finishedJobRunNames), strings.Join(finishedJobRunNames, ", "), len(unfinishedJobRunNames), strings.Join(unfinishedJobRunNames, ", "))
 	return finishedJobRuns, unfinishedJobRuns, finishedJobRunNames, unfinishedJobRunNames, nil
+}
+
+// OutputTestCaseFailures prints detailed test failures
+func OutputTestCaseFailures(parents []string, suite *junit.TestSuite) {
+	currSuite := append(parents, suite.Name)
+	for _, testCase := range suite.TestCases {
+		if testCase.FailureOutput == nil {
+			continue
+		}
+		if len(testCase.FailureOutput.Output) == 0 && len(testCase.FailureOutput.Message) == 0 {
+			continue
+		}
+		fmt.Printf("Test Failed! suite=[%s], testCase=%v\nMessage: %v\n%v\n\n",
+			strings.Join(currSuite, "  "),
+			testCase.Name,
+			testCase.FailureOutput.Message,
+			testCase.SystemOut)
+	}
+
+	for _, child := range suite.Children {
+		OutputTestCaseFailures(currSuite, child)
+	}
 }
--- a/pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
@@ -1,0 +1,466 @@
+package jobruntestcaseanalyzer
+
+import (
+	"context"
+	"encoding/xml"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/storage"
+	"gopkg.in/yaml.v2"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
+	"github.com/openshift/ci-tools/pkg/junit"
+)
+
+type testIdentifier struct {
+	testSuites []string
+	testName   string
+}
+
+var (
+	installTestSuites     = []string{"cluster install"}
+	installTest           = "install should succeed: overall"
+	installTestIdentifier = testIdentifier{testSuites: installTestSuites, testName: installTest}
+)
+
+// JobGetter gets related jobs for further analysis
+type JobGetter interface {
+	GetJobs(ctx context.Context) ([]jobrunaggregatorapi.JobRow, error)
+}
+
+type testCaseAnalyzerJobGetter struct {
+	platform       string
+	infrastructure string
+	network        string
+	jobGCSPrefixes *[]jobGCSPrefix
+	ciDataClient   jobrunaggregatorlib.CIDataClient
+}
+
+// GetJobs find all related jobs for the test case analyzer
+// For PR payload, this contains jobs correspond to the list of jobGCSPreix passed
+// For release-controller generated payload, this contains all jobs meeting selection criteria
+// from command args.
+func (s *testCaseAnalyzerJobGetter) GetJobs(ctx context.Context) ([]jobrunaggregatorapi.JobRow, error) {
+	jobs, err := s.ciDataClient.ListAllJobs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to list all jobs: %w", err)
+	}
+
+	// if PR payload, only find the exact jobs
+	if len(*s.jobGCSPrefixes) > 0 {
+		jobNames := sets.String{}
+		for i := range *s.jobGCSPrefixes {
+			jobGCSPrefix := (*s.jobGCSPrefixes)[i]
+			jobNames.Insert(jobGCSPrefix.jobName)
+		}
+		jobs = s.filterJobsByNames(jobNames, jobs)
+	} else {
+		// Non PR payload, select by criteria
+		jobs = s.filterJobsForPayload(jobs)
+	}
+	return jobs, nil
+}
+
+func getJobInfrastructure(name string) string {
+	if strings.Contains(name, "upi") {
+		return "upi"
+	}
+	return "ipi"
+}
+
+func (s *testCaseAnalyzerJobGetter) filterJobsForPayload(allJobs []jobrunaggregatorapi.JobRow) []jobrunaggregatorapi.JobRow {
+	jobs := []jobrunaggregatorapi.JobRow{}
+	for i := range allJobs {
+		job := allJobs[i]
+		if (len(s.platform) != 0 && job.Platform != s.platform) ||
+			(len(s.network) != 0 && job.Network != s.network) ||
+			(len(s.infrastructure) != 0 && s.infrastructure != getJobInfrastructure(job.JobName)) {
+			continue
+		}
+		jobs = append(jobs, job)
+	}
+	return jobs
+}
+
+func (s *testCaseAnalyzerJobGetter) filterJobsByNames(jobNames sets.String, allJobs []jobrunaggregatorapi.JobRow) []jobrunaggregatorapi.JobRow {
+	ret := []jobrunaggregatorapi.JobRow{}
+	for i := range allJobs {
+		curr := allJobs[i]
+		if jobNames.Has(curr.JobName) {
+			ret = append(ret, curr)
+		}
+	}
+	return ret
+}
+
+// TestCaseChecker checks if a test passes certain criteria across all job runs
+type TestCaseChecker interface {
+	// CheckTestCase returns a test suite based on whether a test has passed certain criteria across job runs
+	CheckTestCase(ctx context.Context, jobRunJunits map[jobrunaggregatorapi.JobRunInfo]*junit.TestSuites) *junit.TestSuite
+}
+
+type minimumRequiredPassesTestCaseChecker struct {
+	id testIdentifier
+	// testNameSuffix is a string that will be appended to the test name for the test case to
+	// be created. This might include variant info like platform, network and infrastructure etc.
+	testNameSuffix         string
+	requiredNumberOfPasses int
+}
+
+type testStatus int
+
+const (
+	testSkipped testStatus = iota
+	testPassed
+	testFailed
+)
+
+func getTestStatus(id testIdentifier, testSuite *junit.TestSuite) testStatus {
+	if len(id.testSuites) == 0 || id.testSuites[0] != testSuite.Name {
+		return testSkipped
+	}
+	// We have a top level suite match, search for test case
+	if len(id.testSuites) == 1 {
+		for _, testCase := range testSuite.TestCases {
+			// Exact match will result in either pass or fail
+			if testCase.Name == id.testName {
+				if testCase.FailureOutput == nil {
+					return testPassed
+				} else {
+					return testFailed
+				}
+			}
+		}
+		return testSkipped
+	}
+	// Search next level
+	next := id
+	next.testSuites = id.testSuites[1:]
+	for _, childSuite := range testSuite.Children {
+		if len(next.testSuites) > 0 && next.testSuites[0] == childSuite.Name {
+			status := getTestStatus(next, childSuite)
+			if status == testPassed || status == testFailed {
+				return status
+			}
+		}
+	}
+	return testSkipped
+}
+
+func (r minimumRequiredPassesTestCaseChecker) addTestResultToDetails(currDetails *jobrunaggregatorlib.TestCaseDetails,
+	jobRun jobrunaggregatorapi.JobRunInfo, status testStatus) {
+	switch status {
+	case testPassed:
+		currDetails.Passes = append(
+			currDetails.Passes,
+			jobrunaggregatorlib.TestCasePass{
+				JobRunID:       jobRun.GetJobRunID(),
+				HumanURL:       jobRun.GetHumanURL(),
+				GCSArtifactURL: jobRun.GetGCSArtifactURL(),
+			})
+	case testFailed:
+		currDetails.Failures = append(
+			currDetails.Failures,
+			jobrunaggregatorlib.TestCaseFailure{
+				JobRunID:       jobRun.GetJobRunID(),
+				HumanURL:       jobRun.GetHumanURL(),
+				GCSArtifactURL: jobRun.GetGCSArtifactURL(),
+			})
+	default:
+		currDetails.Skips = append(
+			currDetails.Skips,
+			jobrunaggregatorlib.TestCaseSkip{
+				JobRunID:       jobRun.GetJobRunID(),
+				HumanURL:       jobRun.GetHumanURL(),
+				GCSArtifactURL: jobRun.GetGCSArtifactURL(),
+			})
+	}
+}
+
+// addToTestSuiteFromSuiteNames adds embedded TestSuite from array of suite names to top suite.
+// It returns the bottom level suite for easy access for callers.
+func addToTestSuiteFromSuiteNames(topSuite *junit.TestSuite, suiteNames []string) *junit.TestSuite {
+	previousSuite := topSuite
+	for _, suiteName := range suiteNames {
+		currentSuite := &junit.TestSuite{
+			Name:      suiteName,
+			TestCases: []*junit.TestCase{},
+		}
+		previousSuite.Children = append(previousSuite.Children, currentSuite)
+		previousSuite = currentSuite
+	}
+	return previousSuite
+}
+
+func updateTestCountsInSuite(suite *junit.TestSuite) {
+	var numTests, numFailed uint
+	for _, test := range suite.TestCases {
+		numTests++
+		if test.FailureOutput != nil {
+			numFailed++
+		}
+	}
+	for _, child := range suite.Children {
+		updateTestCountsInSuite(child)
+		numTests += child.NumTests
+		numFailed += child.NumFailed
+	}
+	suite.NumTests = numTests
+	suite.NumFailed = numFailed
+}
+
+// CheckTestCase returns a test case based on whether a test has passed certain criteria across job runs
+func (r minimumRequiredPassesTestCaseChecker) CheckTestCase(ctx context.Context, jobRunJunits map[jobrunaggregatorapi.JobRunInfo]*junit.TestSuites) *junit.TestSuite {
+	topSuite := &junit.TestSuite{
+		Name:      "minimum-required-passes-checker",
+		TestCases: []*junit.TestCase{},
+	}
+	bottomSuite := addToTestSuiteFromSuiteNames(topSuite, r.id.testSuites)
+
+	testName := fmt.Sprintf("test '%s' has required number of successful passes across payload jobs", r.id.testName)
+	if len(r.testNameSuffix) > 0 {
+		testName += fmt.Sprintf(" for %s", r.testNameSuffix)
+	}
+	testCase := &junit.TestCase{
+		Name: testName,
+	}
+	bottomSuite.TestCases = append(bottomSuite.TestCases, testCase)
+
+	start := time.Now()
+	successCount := 0
+	currDetails := &jobrunaggregatorlib.TestCaseDetails{
+		Name:          r.id.testName,
+		TestSuiteName: strings.Join(r.id.testSuites, jobrunaggregatorlib.TestSuitesSeparator),
+	}
+	for jobRun, testSuites := range jobRunJunits {
+		status := testSkipped
+		// Now go through test suites check test result
+		for _, testSuite := range testSuites.Suites {
+			status = getTestStatus(r.id, testSuite)
+			found := false
+			switch status {
+			case testPassed:
+				found = true
+				successCount++
+			case testFailed:
+				found = true
+			}
+			if found {
+				break
+			}
+		}
+		r.addTestResultToDetails(currDetails, jobRun, status)
+	}
+	currDetails.Summary = fmt.Sprintf("Total job runs: %d, passes: %d, failures: %d, skips %d", len(jobRunJunits), len(currDetails.Passes), len(currDetails.Failures), len(currDetails.Skips))
+	detailsYaml, err := yaml.Marshal(currDetails)
+	if err != nil {
+		return nil
+	}
+	testCase.Duration = time.Since(start).Seconds()
+	testCase.SystemOut = string(detailsYaml)
+	if successCount < r.requiredNumberOfPasses {
+		testCase.FailureOutput = &junit.FailureOutput{
+			Message: fmt.Sprintf("required minimum successful count %d, got %d", r.requiredNumberOfPasses, successCount),
+		}
+	}
+	updateTestCountsInSuite(topSuite)
+	return topSuite
+}
+
+// JobRunTestCaseAnalyzerOptions
+// 1. either gets a list of jobs from big query that meet the passed criteria: platform, network, or uses the passed jobs
+// 2. finds job runs for matching jobs for the specified payload tag
+// 3. runs all test case checkers and constructs a synthetic junit
+type JobRunTestCaseAnalyzerOptions struct {
+	payloadTag string
+	workingDir string
+	// jobRunStartEstimate is used by job run locator to calculate the time window to search for job runs.
+	jobRunStartEstimate time.Time
+	timeout             time.Duration
+	ciDataClient        jobrunaggregatorlib.CIDataClient
+	ciGCSClient         jobrunaggregatorlib.CIGCSClient
+	gcsClient           *storage.Client
+	testCaseCheckers    []TestCaseChecker
+	payloadInvocationID string
+	jobGCSPrefixes      *[]jobGCSPrefix
+	jobGetter           JobGetter
+}
+
+func (o *JobRunTestCaseAnalyzerOptions) findJobRunsWithRetry(ctx context.Context,
+	jobName string, jobRunLocator jobrunaggregatorlib.JobRunLocator) ([]jobrunaggregatorapi.JobRunInfo, error) {
+	errorsInARow := 0
+	for {
+		jobRuns, err := jobRunLocator.FindRelatedJobs(ctx)
+		if err != nil {
+			if errorsInARow > 20 {
+				fmt.Printf("give up finding job runs for %s after retries: %v", jobName, err)
+				return nil, err
+			}
+			errorsInARow++
+			fmt.Printf("error finding job runs for %s: %v", jobName, err)
+		} else {
+			return jobRuns, nil
+		}
+
+		fmt.Printf("   waiting and will attempt to find related jobs for %s in a minute\n", jobName)
+		select {
+		case <-ctx.Done():
+			// Simply return. Caller will check ctx and return error
+			return nil, ctx.Err()
+		case <-time.After(1 * time.Minute):
+			continue
+		}
+	}
+}
+
+// GetRelatedJobRuns gets all related job runs for analysis
+func (o *JobRunTestCaseAnalyzerOptions) GetRelatedJobRuns(ctx context.Context) ([]jobrunaggregatorapi.JobRunInfo, error) {
+	var jobRunsToReturn []jobrunaggregatorapi.JobRunInfo
+	jobs, err := o.jobGetter.GetJobs(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get related jobs: %w", err)
+	}
+
+	waitGroup := sync.WaitGroup{}
+	resultCh := make(chan []jobrunaggregatorapi.JobRunInfo, len(jobs))
+	for i := range jobs {
+		job := jobs[i]
+		var jobRunLocator jobrunaggregatorlib.JobRunLocator
+
+		if len(o.payloadTag) > 0 {
+			jobRunLocator = jobrunaggregatorlib.NewPayloadAnalysisJobLocatorForReleaseController(
+				job.JobName,
+				o.payloadTag,
+				o.jobRunStartEstimate,
+				o.ciDataClient,
+				o.ciGCSClient,
+				o.gcsClient,
+				"origin-ci-test",
+			)
+		}
+		if len(o.payloadInvocationID) > 0 {
+			jobRunLocator = jobrunaggregatorlib.NewPayloadAnalysisJobLocatorForPR(
+				job.JobName,
+				o.payloadInvocationID,
+				jobrunaggregatorlib.PayloadInvocationIDLabel,
+				o.jobRunStartEstimate,
+				o.ciDataClient,
+				o.ciGCSClient,
+				o.gcsClient,
+				"origin-ci-test",
+				(*o.jobGCSPrefixes)[i].gcsPrefix,
+			)
+		}
+
+		fmt.Printf("  launching findJobRunsWithRetry for %q\n", job.JobName)
+
+		waitGroup.Add(1)
+
+		go func() {
+			defer waitGroup.Done()
+			jobRuns, err := o.findJobRunsWithRetry(ctx, job.JobName, jobRunLocator)
+			if err == nil {
+				resultCh <- jobRuns
+			}
+		}()
+	}
+	waitGroup.Wait()
+	close(resultCh)
+
+	// drain the result channel first
+	for jobRuns := range resultCh {
+		jobRunsToReturn = append(jobRunsToReturn, jobRuns...)
+	}
+	select {
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	default:
+		break
+	}
+	return jobRunsToReturn, nil
+}
+
+func (o *JobRunTestCaseAnalyzerOptions) runTestCaseCheckers(ctx context.Context,
+	finishedJobRuns []jobrunaggregatorapi.JobRunInfo, unfinishedJobRuns []jobrunaggregatorapi.JobRunInfo) *junit.TestSuite {
+	suiteName := "payload-cross-jobs"
+	topSuite := &junit.TestSuite{
+		Name:      suiteName,
+		TestCases: []*junit.TestCase{},
+	}
+
+	allJobRuns := append(finishedJobRuns, unfinishedJobRuns...)
+	jobRunJunitMap := map[jobrunaggregatorapi.JobRunInfo]*junit.TestSuites{}
+	for i := range allJobRuns {
+		jobRun := allJobRuns[i]
+
+		testSuites, err := jobRun.GetCombinedJUnitTestSuites(ctx)
+		if err != nil {
+			continue
+		}
+		jobRunJunitMap[jobRun] = testSuites
+	}
+	for _, checker := range o.testCaseCheckers {
+		testSuite := checker.CheckTestCase(ctx, jobRunJunitMap)
+		topSuite.Children = append(topSuite.Children, testSuite)
+		topSuite.NumTests += testSuite.NumTests
+		topSuite.NumFailed += testSuite.NumFailed
+	}
+	return topSuite
+}
+
+func (o *JobRunTestCaseAnalyzerOptions) Run(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, o.timeout)
+	defer cancel()
+
+	matchID := o.payloadTag
+	if len(matchID) == 0 {
+		matchID = o.payloadInvocationID
+	}
+
+	outputDir := filepath.Join(o.workingDir, matchID)
+	if err := os.MkdirAll(outputDir, 0755); err != nil {
+		return fmt.Errorf("error creating output directory %q: %w", outputDir, err)
+	}
+
+	// if it hasn't been more than hour since the jobRuns started, the list isn't complete.
+	readyAt := o.jobRunStartEstimate.Add(1 * time.Hour)
+
+	durationToWait := o.timeout - 20*time.Minute
+	timeToStopWaiting := o.jobRunStartEstimate.Add(durationToWait)
+
+	fmt.Printf("Analyzing test status for job runs for %q.  now=%v, ReadyAt=%v, timeToStopWaiting=%v.\n", matchID, time.Now(), readyAt, timeToStopWaiting)
+
+	err := jobrunaggregatorlib.WaitUntilTime(ctx, readyAt)
+	if err != nil {
+		return err
+	}
+	finishedJobRuns, unfinishedJobRuns, _, _, err := jobrunaggregatorlib.WaitAndGetAllFinishedJobRuns(ctx, timeToStopWaiting, o, outputDir)
+	if err != nil {
+		return err
+	}
+
+	testSuite := o.runTestCaseCheckers(ctx, finishedJobRuns, unfinishedJobRuns)
+	jobrunaggregatorlib.OutputTestCaseFailures([]string{"root"}, testSuite)
+
+	// Done with all tests
+	junitXML, err := xml.Marshal(testSuite)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(filepath.Join(outputDir, "junit-test-case-analysis.xml"), junitXML, 0644); err != nil {
+		return err
+	}
+	if testSuite.NumFailed > 0 {
+		return fmt.Errorf("Some test checker failed.  See above for details.")
+	}
+	return nil
+}
--- a/pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
+++ b/pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
@@ -1,0 +1,352 @@
+package jobruntestcaseanalyzer
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
+)
+
+const (
+	installTestGroup                  string = "install"
+	defaultMinimumSuccessfulTestCount int    = 1
+	// maxTimeout is our guess of the maximum duration for a job run
+	maxTimeout time.Duration = 4*time.Hour + 35*time.Minute
+)
+
+var (
+	knownPlatforms = sets.String{
+		"aws":     sets.Empty{},
+		"azure":   sets.Empty{},
+		"gcp":     sets.Empty{},
+		"libvirt": sets.Empty{},
+		"metal":   sets.Empty{},
+		"ovirt":   sets.Empty{},
+		"vsphere": sets.Empty{},
+	}
+	knownNetworks        = sets.String{"ovn": sets.Empty{}, "sdn": sets.Empty{}}
+	knownInfrastructures = sets.String{"upi": sets.Empty{}, "ipi": sets.Empty{}}
+	knownTestGroups      = sets.String{installTestGroup: sets.Empty{}}
+)
+
+type jobGCSPrefix struct {
+	jobName   string
+	gcsPrefix string
+}
+
+type jobGCSPrefixSlice struct {
+	values *[]jobGCSPrefix
+}
+
+func (s *jobGCSPrefixSlice) String() string {
+	if len(*s.values) == 0 {
+		return ""
+	}
+	var jobPairs []string
+	for _, value := range *s.values {
+		jobPairs = append(jobPairs, fmt.Sprintf("%s=%s", value.jobName, value.gcsPrefix))
+	}
+	return strings.Join(jobPairs, ",")
+}
+
+func (s *jobGCSPrefixSlice) Set(value string) error {
+	if len(value) == 0 {
+		*s.values = nil
+		return nil
+	}
+	jobPairs := strings.Split(value, ",")
+	if len(jobPairs) == 0 {
+		return fmt.Errorf("need at least one GCS prefix configured with explicit-gcs-prefixes")
+	}
+	for _, jobPair := range jobPairs {
+		jStrs := strings.Split(jobPair, "=")
+		if len(jStrs) != 2 {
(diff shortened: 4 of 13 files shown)


pkg/jobrunaggregator/jobruninstallanalyzer/cmd.go
```diff
@@ -0,0 +1,152 @@
...
+
+const kubeTimeSerializationLayout = time.RFC3339
+
+func (f *JobRunsInstallAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
+	f.DataCoordinates.BindFlags(fs)
+	f.Authentication.BindFlags(fs)
+
+	fs.StringVar(&f.PayloadTag, "payload-tag", f.PayloadTag, "The payload tag to analyze installation status, like 4.9.0-0.ci-2021-07-19-185802")
+	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.Platform, "platform", f.Platform, "The platform is used to narrow down a subset of the jobs to analyze")
```
deads2k, line 50: provide example values please.

pkg/jobrunaggregator/jobruninstallanalyzer/cmd.go
```diff
@@ -0,0 +1,152 @@
...
+	fs.StringVar(&f.Infrastructure, "infrastructure", f.Infrastructure, "The infrastructure is used to narrow down a subset of the jobs to analyze")
```
deads2k, line 51: provide example values please.

pkg/jobrunaggregator/jobruninstallanalyzer/cmd.go
```diff
@@ -0,0 +1,152 @@
...
+	fs.StringVar(&f.Network, "network", f.Network, "The network is used to narrow down a subset of the jobs to analyze")
```
deads2k, line 52: example values.

pkg/jobrunaggregator/jobruninstallanalyzer/cmd.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"k8s.io/apimachinery/pkg/util/clock"
+
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
+)
+
+type JobRunsInstallAnalyzerFlags struct {
```
deads2k, line 17: to work with /payload commands, you'll need `--aggregation-id` and `--explicit-gcs-prefix`

pkg/jobrunaggregator/jobruninstallanalyzer/analyzer.go
```diff
@@ -0,0 +1,305 @@
...
+	jobRunStartEstimate      time.Time
+	clock                    clock.Clock
+	timeout                  time.Duration
+	platform                 string
+	infrastructure           string
+	network                  string
+	ciDataClient             jobrunaggregatorlib.CIDataClient
+	ciGCSClient              jobrunaggregatorlib.CIGCSClient
+	gcsClient                *storage.Client
+	minimumSuccessfulInstall int
```
deads2k, line 40: the command line arg looks correct, but at this leave, please create a 

```go
  requiredPasses []RequiredPasses
}

type RequiredPasses struct{
  testName string
  requiredNumberOfPasses string
}

```

pkg/jobrunaggregator/jobruninstallanalyzer/analyzer.go
```diff
@@ -0,0 +1,305 @@
...
+}
+
+func (o *JobRunInstallAnalyzerOptions) getRelatedJobRuns(ctx context.Context) ([]jobrunaggregatorapi.JobRunInfo, error) {
+	var jobRunsToReturn []jobrunaggregatorapi.JobRunInfo
+	jobs, err := o.ciDataClient.ListAllJobs(ctx)
```
deads2k, line 45: I don't see how you will be able to invert in this direction when this is run via a /payload command.

You need to be able to inspect a list of JobNames (release controller knows them all and so does the payload command) and select which ones match.  This means adding a `--jobNames=job1,job2,job3,job4` etc

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
```diff
@@ -10,7 +10,10 @@ import (
 )
 
 func GetPayloadTagFromProwJob(prowJob *prowjobv1.ProwJob) string {
```
deads2k, line 12: wasn't expecting this to change.

pkg/jobrunaggregator/jobruninstallanalyzer/analyzer.go
```diff
@@ -0,0 +1,305 @@
...
+		relatedJobRuns, err := o.getRelatedJobRuns(ctx)
+		if err != nil {
+			return finishedJobRuns, unfinishedJobRuns, err
+		}
+
+		if len(relatedJobRuns) == 0 {
+			return finishedJobRuns, unfinishedJobRuns, fmt.Errorf("job runs for payload %q: found no related jobRuns", o.payloadTag)
+		}
+
+		for i := range relatedJobRuns {
```
deads2k, line 221: this logic the same as the  other analyzer?  if so, make a helper please.

pkg/jobrunaggregator/jobruninstallanalyzer/analyzer.go
```diff
@@ -0,0 +1,305 @@
...
+			continue
+		}
+
+		// Any successful job run implies a successful install
+		if prowJob.Status.State == prowv1.SuccessState {
+			successCount++
+			continue
+		}
+
+	loop:
```
deads2k, line 132: no, refactor to eliminate this.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
```diff
@@ -9,12 +9,19 @@ import (
 	prowjobv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 )
 
-func GetAggregationIDFromProwJob(prowJob *prowjobv1.ProwJob) string {
-	return prowJob.Labels["release.openshift.io/aggregation-id"]
+const (
+	// AggregationIDLabel is the name of the label for the aggregation id in prow job
+	AggregationIDLabel = "release.openshift.io/aggregation-id"
+	// PayloadInvocationIDLabel is the name of the label for the payload invocation id in prow job
+	PayloadInvocationIDLabel = "release.openshift.io/aggregation-id"
```
xueqzhan, line 16: For now the payload invocation label is set to "release.openshift.io/aggregation-id" just for testing purpose. When this is set by the /payload implementation, we can change it to the matching value.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
```diff
@@ -9,12 +9,20 @@ import (
 	prowjobv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 )
 
-func GetPayloadTagFromProwJob(prowJob *prowjobv1.ProwJob) string {
-	return prowJob.Labels["release.openshift.io/analysis"]
+// GetPayloadTagFromProwJob gets the payload tag from prow jobs. For aggregated jobs,
+// the label is release.openshift.io/analysis. For non-aggregated jobs, the label
+// is release.openshift.io/tag
+func GetPayloadTagFromProwJob(prowJob *prowjobv1.ProwJob, isAggregated bool) string {
```
deads2k, line 15: don't use bool flags.  Create two methods.

pkg/jobrunaggregator/jobrunaggregatorlib/spyglass_summary.go
```diff
@@ -0,0 +1,88 @@
...
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
+)
+
+func htmlForJobRuns(ctx context.Context, finishedJobsToAggregate, unfinishedJobsToAggregate []jobrunaggregatorapi.JobRunInfo) string {
```
deads2k, line 11: straigth move?

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
```diff
@@ -9,12 +9,20 @@ import (
...
+	payloadTag := prowJob.Labels["release.openshift.io/analysis"]
+	if len(payloadTag) == 0 && !isAggregated {
+		payloadTag = prowJob.Annotations["release.openshift.io/tag"]
```
deads2k, line 18: Can we simply always use this value?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+// 4. constructs a synthentic junit that check minimum required succuessful passes for configured tests
+type JobRunTestCaseAnalyzerOptions struct {
+	payloadTag          string
+	workingDir          string
+	jobRunStartEstimate time.Time
+	clock               clock.Clock
+	timeout             time.Duration
+	platform            string
+	infrastructure      string
+	network             string
```
deads2k, line 44: these are used only for job name selection?  if so, please create a 

```go
type JobNameSelector interface{
    ListJobNames(ctx)
```

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,241 @@
...
+		return fmt.Errorf("cannot specify both --payload-tag and --payload-invocation-id")
+	}
+	if len(f.PayloadTag) == 0 && len(f.PayloadInvocationID) == 0 {
+		return fmt.Errorf("exactly one of --payload-tag or --payload-invocation-id must be specified")
+	}
+	if len(f.PayloadInvocationID) > 0 && len(f.JobGcsPrefixes) == 0 {
+		return fmt.Errorf("if --payload-invocation-id is specified, you must specify --explicit-gcs-prefixes")
+	}
+
+	bigQueryClient, err := f.Authentication.NewBigQueryClient(context.Background(), f.DataCoordinates.ProjectID)
```
deads2k, line 163: this doesn't belong in a validate step.  You can do this in `ToOptions` if you like.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,241 @@
...
+	WorkingDir                  string
+	PayloadTag                  string
+	Timeout                     time.Duration
+	EstimatedJobStartTimeString string
+	Platform                    string
+	Infrastructure              string
+	Network                     string
+	MinimumSuccessfulInstall    int
+	PayloadInvocationID         string
+	JobGcsPrefixes              []jobGcsPrefix
```
deads2k, line 71: GCS

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,241 @@
...
+}
+
+func (s *jobGcsPrefixSlice) Set(value string) error {
+	if len(value) == 0 {
+		*s.values = nil
+		return nil
+	}
+	jobPairs := strings.Split(value, ",")
+	for _, jobPair := range jobPairs {
+		jStrs := strings.Split(jobPair, "~")
```
deads2k, line 48: is that a tilde?  Why not equals?

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
deads2k, line 71: I recommend using `[]string` with validation on the format.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+			continue
+		}
+
+		var jobRunLocator jobrunaggregatorlib.JobRunLocator
+
+		if len(o.payloadTag) > 0 {
+			jobRunLocator = jobrunaggregatorlib.NewPayloadAnalysisJobLocatorForReleaseController(
+				job.JobName,
+				o.payloadTag,
+				false,
```
deads2k, line 114: this is unreadable.  either assign to variable or comment or use a different method name.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+			} else {
+				jobRunsToReturn = append(jobRunsToReturn, jobRuns...)
+				break
+			}
+
+			fmt.Printf("   waiting and will attempt to find related jobs in a minute\n")
+			select {
+			case <-ctx.Done():
+				return nil, ctx.Err()
+			case <-time.After(1 * time.Minute):
```
deads2k, line 155: retrying in the nested loop can lead to pathological retry behavior, right?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
+	"github.com/openshift/ci-tools/pkg/junit"
+)
+
+const (
+	installTest = "cluster install should succeed"
+)
+
+type requiredPasses struct {
+	testName               string
```
deads2k, line 27: you need a `testSuite []string` to uniquely identify

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+	requiredNumberOfPasses int
```
deads2k, line 28: make this a pass/fail interface that takes the number of successes and the number of failures

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+			Name: testName,
+		}
+		testSuite.TestCases = append(testSuite.TestCases, testCase)
+		testSuite.NumTests++
+
+		successCount := o.getTestSuccessCount(ctx, allJobRuns, requiredPass.testName)
+		if successCount < requiredPass.requiredNumberOfPasses {
+			testCase.FailureOutput = &junit.FailureOutput{
+				Message: fmt.Sprintf("required minimum successful count %d, got %d", requiredPass.requiredNumberOfPasses, successCount),
+				Output:  testCase.SystemOut,
```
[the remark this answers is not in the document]
deads2k: > is something setting SystemOut? If not, please add details about every job run included.

and we want to see it for successful runs too

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,241 @@
...
+
+	fs.StringVar(&f.WorkingDir, "working-dir", f.WorkingDir, "The directory to store caches, output, and the like.")
+	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for analyzing job to complete.")
+	fs.Var(&jobGcsPrefixSlice{&f.JobGcsPrefixes}, "explicit-gcs-prefixes", "a list of gcs prefixes for jobs created for payload. Only used by per PR payload promotion jobs. The format is comma-separated elements, each consisting of job name and gcs prefix separate by tilde, like openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade~logs/openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade")
+}
+
+func NewJobRunsTestCaseAnalyzerCommand() *cobra.Command {
+	f := NewJobRunsTestCaseAnalyzerFlags()
+
+	cmd := &cobra.Command{
```
deads2k, line 108: Add realistic examples of how you'd like the release controller to call this.

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
+)
+
+type JobRunGetter interface {
+	// GetRelatedJobRuns gets all related job runs for analysis
+	GetRelatedJobRuns(ctx context.Context) ([]jobrunaggregatorapi.JobRunInfo, error)
+}
+
+// WaitAndGetAllJobRuns waits
+func WaitAndGetAllJobRuns(ctx context.Context,
```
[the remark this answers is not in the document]
xueqzhan: yes, straight move.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+type requiredPasses struct {
+	testName               string
+	requiredNumberOfPasses int
+}
+
+// JobRunTestCaseAnalyzerOptions
+// 1. get a list of jobs from big query that meet the passed criteria: platform, network etc.
+// 2. finds jobruns for matching jobs for the specified payload tag
+// 3. reads all junit for the each jobrun
+// 4. constructs a synthentic junit that check minimum required succuessful passes for configured tests
```
DennisPeriquet, line 35: s/succuessful/successful/

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+	}
+	return jobs, nil
+}
+
+func getJobInfrastructure(name string) string {
+	if strings.Contains(name, "upi") {
+		return "upi"
+	}
+	return "ipi"
```
DennisPeriquet, line 74: This is probably a nit, but I'd explicitly check if "ipi" occurs within `names` too and return "ipi" if true.

I realize there is only "ipi" and "upi" today so it doesn't matter; but if a new one is added, this code will return "ipi" for anything not "upi".

pkg/jobrunaggregator/jobrunaggregatorlib/spyglass_summary.go
xueqzhan: straight move

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
xueqzhan: Good idea. That removes the isAggregated flag as well.

xueqzhan: Turned out this is still needed. Otherwise we will be picking up the informing run of this job as well.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: Job names do not always have "upi" or "ipi". When upi is in the name, it is upi, otherwise, it is considered ipi.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: Not really. They are also used to create specific test names later. So it makes sense to have them as variables.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
xueqzhan: The clients are used to dynamically get all networks and platforms for validation purposes. Also, ToOptions is done after Validate.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,268 @@
...
+
+	f.BindFlags(cmd.Flags())
+
+	return cmd
+}
+
+// Validate checks to see if the user-input is likely to produce functional runtime options
+func (f *JobRunsTestCaseAnalyzerFlags) Validate() error {
+	if len(f.WorkingDir) == 0 {
+		return fmt.Errorf("missing --working-dir: like -install-analyzer-working-dir")
```
dgoodwin, line 166: What does like -install-analyzer-working-dir mean here.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,268 @@
...
+
+const kubeTimeSerializationLayout = time.RFC3339
+
+func (f *JobRunsTestCaseAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
+	f.DataCoordinates.BindFlags(fs)
+	f.Authentication.BindFlags(fs)
+
+	fs.StringVar(&f.PayloadTag, "payload-tag", f.PayloadTag, "The payload tag to analyze test case status, like 4.9.0-0.ci-2021-07-19-185802")
+	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.Platform, "platform", f.Platform, "The platform used to narrow down a subset of the jobs to analyze, like vsphere")
```
dgoodwin, line 99: Suggest replacing 'like vsphere' with '(ex: aws|gcp|azure|vsphere)' Similar for the use of "like" in most of these usage comments.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,412 @@
...
+		return nil, ctx.Err()
+	default:
+		break
+	}
+	return jobRunsToReturn, nil
+}
+
+func (o *JobRunTestCaseAnalyzerOptions) runTestCaseCheckers(ctx context.Context,
+	finishedJobRuns []jobrunaggregatorapi.JobRunInfo, unfinishedJobRuns []jobrunaggregatorapi.JobRunInfo) *junit.TestSuite {
+	suiteName := "test case checker"
```
dgoodwin, line 339: Would "payload-cross-job" be an ok suite name?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,412 @@
...
+	if len(o.platform) != 0 {
+		suiteName += fmt.Sprintf(" for platform %s", o.platform)
+	}
+	if len(o.network) != 0 {
+		suiteName += fmt.Sprintf(" with network %s", o.network)
+	}
+	if len(o.infrastructure) != 0 {
+		suiteName += fmt.Sprintf(" and infrastructure %s", o.infrastructure)
+	}
```
dgoodwin, line 348: This feels like too much granularity leaking into suite names, I'd expect fewer test suites with many tests within, in this case it looks like we're going to be close to 1-1 suite-test relationship. Could we move this info into the test name instead and use the suite for a generic bucket for all of these types of tests? This will give us better insight into how a test is behaving on each platform in Sippy, which is also not well setup for dynamic suite names.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,412 @@
...
+		}
+		r.addTestResultToDetails(currDetails, jobRun, status)
+	}
+	currDetails.Summary = fmt.Sprintf("Total job runs: %d, passes: %d, failures: %d, skips %d", len(jobRuns), len(currDetails.Passes), len(currDetails.Failures), len(currDetails.Skips))
+	detailsYaml, err := yaml.Marshal(currDetails)
+	if err != nil {
+		return nil
+	}
+	testCase.Duration = time.Since(start).Seconds()
+	testCase.SystemOut = string(detailsYaml)
```
dgoodwin, line 152: I actually thought setting SystemOut = automatic failure, but may have been mistaken or that was in another area.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,412 @@
...
+
+	"k8s.io/apimachinery/pkg/util/clock"
+
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
+	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
+	"github.com/openshift/ci-tools/pkg/junit"
+)
+
+type testIdentifier struct {
+	testSuites []string
```
dgoodwin, line 25: What is the scenario in which we might be looking for a test amidst multiple suites? Is this anticipating openshift-tests vs openshift-tests-upgrade?

dgoodwin (commented):
This is going to be really nice to see in play. Mostly just nits looks sound to me.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,412 @@
...
+
+type minimumRequiredPassesTestCaseChecker struct {
+	id                     testIdentifier
+	requiredNumberOfPasses int
+}
+
+type testStatus int
+
+const (
+	testSkipped testStatus = iota
```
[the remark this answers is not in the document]
xueqzhan: For PayloadInvocationIDLabel, we will need to ask DPTP to make corresponding changes for PR payload command to hook this up. We should have the release job config change for release-controller first as a reference.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
xueqzhan: Typo with the extra -

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: This is where I need some suggestions. 

This is the current format:

<testsuite name="test case checker for platform aws with network ovn and infrastructure ipi" tests="1" skipped="0" failures="1" time="0">
    <testcase name="test 'cluster install/install should succeed: overall/platform: aws network: ovn infrastructure ipi' should have required number of successful passes across jobRuns" time="0">
    </testcase>
</testsuite>

I am thinking of changing this to:
<testsuite name="payload-cross-job" tests="1" skipped="0" failures="1" time="0">
    <testcase name="test 'cluster install/install should succeed: overall/platform: aws network: ovn infrastructure ipi' should have required number of successful passes across jobRuns" time="0">
    </testcase>
</testsuite>

Any improvement on the test name?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: That is FailureOutput I think.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: Test suite can be embedded. This design allows one to match internal test suites.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,404 @@
...
+					return testPassed
+				} else {
+					return testFailed
+				}
+			}
+		}
+		return testSkipped
+	}
+	// Search next level
+	id.testSuites = id.testSuites[1:]
```
deads2k, line 74: don't rewrite your parameters, create a new instance.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,404 @@
...
+
+// TestCaseChecker checks if a test passes certain criteria across all job runs
+type TestCaseChecker interface {
+	// CheckTestCase returns a test case based on whether a test has passed certain criteria across job runs
+	CheckTestCase(ctx context.Context, jobRuns []jobrunaggregatorapi.JobRunInfo) *junit.TestCase
+}
+
+type minimumRequiredPassesTestCaseChecker struct {
+	id                     testIdentifier
+	testNameSuffix         string
```
deads2k, line 43: what is this and what is it for?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,404 @@
...
+func (r minimumRequiredPassesTestCaseChecker) CheckTestCase(ctx context.Context, jobRuns []jobrunaggregatorapi.JobRunInfo) *junit.TestCase {
+	testName := fmt.Sprintf("test '%s/%s/%s' should have required number of successful passes across jobRuns", strings.Join(r.id.testSuites, ":"), r.id.testName, r.testNameSuffix)
+	testCase := &junit.TestCase{
+		Name: testName,
+	}
+	start := time.Now()
+	successCount := 0
+	currDetails := &jobrunaggregatorlib.TestCaseDetails{
+		Name:          r.id.testName,
+		TestSuiteName: strings.Join(r.id.testSuites, ":"),
```
deads2k, line 124: is this how we do this elsewhere?  I honestly do not know.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
deads2k: > Not really. They are also used to create specific test names later. So it makes sense to have them as variables.

Then perhaps doc-ing them is appropriate

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,404 @@
...
+				"origin-ci-test",
+				(*o.jobGCSPrefixes)[i].gcsPrefix,
+			)
+		}
+
+		fmt.Printf("  launching findJobRunsWithRetry for %q\n", job.JobName)
+
+		waitGroup.Add(1)
+
+		go o.findJobRunsWithRetry(ctx, &waitGroup, resultCh, job.JobName, jobRunLocator)
```
deads2k, line 320: restructure for

```
go func(){
  defer waitGroup.Done()
  o.findJobRunsWithjRetry
}()

Comment (deads2k) [inline pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go:320]: probably add to the channel here to keep the signature easier to code to.

File: pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go (hunk 35 above)
Comment (xueqzhan) [reply pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go:43]: Need the identifiers (platform, network etc) for the test names inside TestCaseChecker. I can add another function (SetTestNameSuffix()) to the interface if it is better. Or another better way in go?

File: pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go [hunk 38]
```diff
@@ -0,0 +1,404 @@
...
+				JobRunID:       jobRun.GetJobRunID(),
+				HumanURL:       jobRun.GetHumanURL(),
+				GCSArtifactURL: jobRun.GetGCSArtifactURL(),
+			})
+	}
+}
+
+// CheckTestCase returns a test case based on whether a test has passed certain criteria across job runs
+func (r minimumRequiredPassesTestCaseChecker) CheckTestCase(ctx context.Context, jobRuns []jobrunaggregatorapi.JobRunInfo) *junit.TestCase {
+	testName := fmt.Sprintf("test '%s/%s/%s' should have required number of successful passes across jobRuns", strings.Join(r.id.testSuites, ":"), r.id.testName, r.testNameSuffix)
```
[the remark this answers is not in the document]
xueqzhan: So we need to aggregate the install analyzer? How does that work? Running the install analyzer multiple times will not result in improvement. 

In terms of test name, let me know if you have a better way. I am thinking of following Devan's suggestion above. So it will be 

`test 'cluster install.install should succeed' has required number of passes across payload jobs for platform:aws network:sdn infrastructure:ipi`

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: I do not know either. Devan mentioned "." has been used. But that is between suite and test names. How about between levels of test suites? Where are we using the embedded suites?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,404 @@
...
+			jobGCSPrefix := (*o.jobGCSPrefixes)[i]
+			jobNames = append(jobNames, jobGCSPrefix.jobName)
+		}
+		jobs, err = o.filterJobsByNames(jobNames, jobs)
+		if err != nil {
+			return nil, fmt.Errorf("failed to filter jobs by names: %w", err)
+		}
+	} else {
+		// Non PR payload, select by criteria
+		jobs = o.filterJobsForPayload(jobs)
```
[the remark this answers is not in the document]
xueqzhan: We do not filter by release. Different payload runs have different payload invocation ID. For jobs not specifically for that release, it will not get any job runs by using that invocation ID.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,421 @@
...
+
+type testStatus int
+
+const (
+	testSkipped testStatus = iota
+	testPassed
+	testFailed
+)
+
+func getTestStatus(id testIdentifier, testSuite *junit.TestSuite) testStatus {
```
deads2k, line 126: If you pass a testSuite containing multiple testcases, only the first matching hit is checked.  This case needs to produce an error.

deads2k: > If you pass a testSuite containing multiple testcases, only the first matching hit is checked. This case needs to produce an error.

ok, I think I see it.  It's an unfortunate naming issue in the testSuite versus ID.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,421 @@
...
+				JobRunID:       jobRun.GetJobRunID(),
+				HumanURL:       jobRun.GetHumanURL(),
+				GCSArtifactURL: jobRun.GetGCSArtifactURL(),
+			})
+	}
+}
+
+// CheckTestCase returns a test case based on whether a test has passed certain criteria across job runs
+func (r minimumRequiredPassesTestCaseChecker) CheckTestCase(ctx context.Context, jobRuns []jobrunaggregatorapi.JobRunInfo) *junit.TestCase {
+	testName := fmt.Sprintf("test '%s.%s' has required number of successful passes across payload jobs", strings.Join(r.id.testSuites, jobrunaggregatorlib.TestSuitesSeparator), r.id.testName)
```
deads2k, line 188: I'm a little surprised that we don't keep the suite matching.  Doesn't aggregation normally keep the suites matching?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,421 @@
...
+
+	// Done with all tests
+	junitXML, err := xml.Marshal(testSuite)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(filepath.Join(outputDir, "junit-test-case-analysis.xml"), junitXML, 0644); err != nil {
+		return err
+	}
+	return nil
```
deads2k, line 420: if you have failures, don't you have to exit non-zero so that the job run will show up failed?  analyzer.go does this as I recall.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,290 @@
...
+	fs.StringVar(&f.WorkingDir, "working-dir", f.WorkingDir, "The directory to store caches, output, and the like.")
+	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for analyzing job to complete.")
+	fs.Var(&jobGCSPrefixSlice{&f.JobGCSPrefixes}, "explicit-gcs-prefixes", "a list of gcs prefixes for jobs created for payload. Only used by per PR payload promotion jobs. The format is comma-separated elements, each consisting of job name and gcs prefix separated by =, like openshift-machine-config-operator=3028-ci-4.11-e2e-aws-ovn-upgrade~logs/openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade")
+}
+
+func NewJobRunsTestCaseAnalyzerCommand() *cobra.Command {
+	f := NewJobRunsTestCaseAnalyzerFlags()
+
+	/* Example runs for release controller:
+	 ./job-run-aggregator analyze-required-install
```
deads2k, line 115: `analyzer-required-test` with a repeatable argument that allows specifying something like

`--required-test separated|||suite|||name.testCaseName,numPasses,<integer>` with separation logic (there's probably a CSV library?).

deads2k: I could probably also see a special string for "install" and "upgrade" since we are fairly likely to change those test names over time.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,462 @@
...
+
+type testIdentifier struct {
+	testSuites []string
+	testName   string
+}
+
+var (
+	installTestSuites     = []string{"cluster install"}
+	installTest           = "install should succeed: overall"
+	installTestIdentifier = testIdentifier{installTestSuites, installTest}
```
deads2k, line 33: nit: we initialize with named instance members.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
deads2k: Please add godoc explaining this.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,462 @@
...
+				JobRunID:       jobRun.GetJobRunID(),
+				HumanURL:       jobRun.GetHumanURL(),
+				GCSArtifactURL: jobRun.GetGCSArtifactURL(),
+			})
+	}
+}
+
+// addToTestSuiteFromSuiteNames adds embedded TestSuite from array of suite names to top suite.
+// It returns the bottom level suite for easy access for callers.
+func addToTestSuiteFromSuiteNames(topSuite *junit.TestSuite, suiteNames []string) *junit.TestSuite {
```
deads2k, line 190: a bit of a mind bender this one, but I gather you're mutating so you can build up over time?  I don't personally like it, but perhaps @dgoodwin prefers it this way.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,462 @@
...
+		}
+		previousSuite.Children = append(previousSuite.Children, currentSuite)
+		previousSuite = currentSuite
+	}
+	return previousSuite
+}
+
+func updateTestCountsInSuite(suite *junit.TestSuite) {
+	for _, test := range suite.TestCases {
+		suite.NumTests++
```
deads2k, line 205: please calculate the numbers in a method scoped variable and then do a direct assign.  Done like this, if the method is called twice, the counts will be wrong.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,462 @@
...
+		if test.FailureOutput != nil {
+			suite.NumFailed++
+		}
+	}
+	for _, child := range suite.Children {
+		updateTestCountsInSuite(child)
+		suite.NumTests += child.NumTests
```
deads2k, line 212: same for these.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,462 @@
...
+	// Done with all tests
+	junitXML, err := xml.Marshal(testSuite)
+	if err != nil {
+		return err
+	}
+	if err := ioutil.WriteFile(filepath.Join(outputDir, "junit-test-case-analysis.xml"), junitXML, 0644); err != nil {
+		return err
+	}
+	if testSuite.NumFailed > 0 {
+		return fmt.Errorf("Some test checker failed.  Check JUnit files for details")
```
deads2k, line 459: Future you is going to be annoyed if the log output for the job doesn't write out the failure.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,320 @@
...
+	f.DataCoordinates.BindFlags(fs)
+	f.Authentication.BindFlags(fs)
+
+	fs.Var(&f.TestGroups, "test-group", "One or more test groups to analyze, like install")
+	fs.StringVar(&f.PayloadTag, "payload-tag", f.PayloadTag, "The payload tag to analyze test case status, like 4.9.0-0.ci-2021-07-19-185802")
+	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.Platform, "platform", f.Platform, "The platform used to narrow down a subset of the jobs to analyze, ex: aws|gcp|azure|vsphere")
+	fs.StringVar(&f.Infrastructure, "infrastructure", f.Infrastructure, "The infrastructure used to narrow down a subset of the jobs to analyze, ex: upi|ipi")
+	fs.StringVar(&f.Network, "network", f.Network, "The network used to narrow down a subset of the jobs to analyze, ex: sdn|ovn")
+	fs.IntVar(&f.MinimumSuccessfulInstall, "minimum-successful-install", 1, "minimum number of successful installations among jobs meeting criteria")
```
deads2k, line 138: set the default value in the `NewJobRunsTestCaseAnalyzerFlags` function

deads2k, line 138: this is `minimum-successful-count`, right?  versus a possible future `minimum-successful-percent` or `minimum-successful-likelihood`

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,471 @@
...
+		updateTestCountsInSuite(child)
+		numTests += child.NumTests
+		numFailed += child.NumFailed
+	}
+	suite.NumTests = numTests
+	suite.NumFailed = numFailed
+}
+
+// CheckTestCase returns a test case based on whether a test has passed certain criteria across job runs
+func (r minimumRequiredPassesTestCaseChecker) CheckTestCase(ctx context.Context, jobRunJunits *map[jobrunaggregatorapi.JobRunInfo]*junit.TestSuites) *junit.TestSuite {
```
deads2k, line 223: why a pointer to a map?

deads2k:
one remaining thing I'd like fixed.  I leave lgtm with @dgoodwin 

/approve
/assign @dgoodwin

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: Only if I knew better.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,471 @@
...
+}
+
+// JobRunTestCaseAnalyzerOptions
+// 1. either gets a list of jobs from big query that meet the passed criteria: platform, network, or uses the passed jobs
+// 2. finds job runs for matching jobs for the specified payload tag
+// 3. runs all test case checkers and constructs a synthetic junit
+type JobRunTestCaseAnalyzerOptions struct {
+	payloadTag          string
+	workingDir          string
+	jobRunStartEstimate time.Time
```
dgoodwin, line 287: Could use godoc for why we need a job run start estimate.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,471 @@
...
+	clock               clock.Clock
```
dgoodwin, line 288: godoc as well.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -0,0 +1,471 @@
...
+	if err := os.MkdirAll(outputDir, 0755); err != nil {
+		return fmt.Errorf("error creating output directory %q: %w", outputDir, err)
+	}
+
+	// if it hasn't been more than hour since the jobRuns started, the list isn't complete.
+	readyAt := o.jobRunStartEstimate.Add(1 * time.Hour)
+
+	// Our guess of the max time for jobs
+	durationToWait := o.timeout - 20*time.Minute
+	if durationToWait > (4*time.Hour + 15*time.Minute) {
```
dgoodwin, line 440: I would output a warning here if you're overriding the users requested timeout. This actually might be better off getting rejected in the CLI validation, as this seems like it might be doing something the user doesn't expect and won't notice.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,322 @@
...
+}
+
+const kubeTimeSerializationLayout = time.RFC3339
+
+func (f *JobRunsTestCaseAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
+	f.DataCoordinates.BindFlags(fs)
+	f.Authentication.BindFlags(fs)
+
+	fs.Var(&f.TestGroups, "test-group", "One or more test groups to analyze, like install")
+	fs.StringVar(&f.PayloadTag, "payload-tag", f.PayloadTag, "The payload tag to analyze test case status, like 4.9.0-0.ci-2021-07-19-185802")
```
dgoodwin, line 135:
```suggestion
	fs.StringVar(&f.PayloadTag, "payload-tag", f.PayloadTag, "The release controller payload tag to analyze test case status, like 4.9.0-0.ci-2021-07-19-185802")
```

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,322 @@
...
+	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
```
dgoodwin, line 136: I'd love to see why the caller specifies a job start time in the help here, I assume it's to filter query results so we don't pull too much? Tracing this down through the 5-10 layers it's passed, the name changes at each layer and in the end there look to be some strange assumptions about adding another hour to the start and four to the end. Caller may want to be aware of that if they were looking at the help for this command.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,322 @@
...
+	    --google-service-account-credential-file=credential.json
+	    --test-group=install
+	    --payload-invocation-id=09406e30ea661e228c17120f28eff3c6
+	    --job-start-time=2022-03-18T13:10:20Z
+	    --minimum-successful-count=10
+	    --explicit-gcs-prefixes=periodic-ci-openshift-release-master-ci-4.11-e2e-aws-ovn-upgrade=logs/openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade,periodic-ci-openshift-release-master-ci-4.11-upgrade-from-stable-4.10-e2e-aws-ovn-upgrade=/logs/openshift-machine-config-operator-3028-ci-4.11-upgrade-from-stable-4.10-e2e-aws-ovn-upgrade
+	*/
+	cmd := &cobra.Command{
+		Use:          "analyze-test-case",
+		Long:         `Analyze status of a test case of certain group to make sure they meet minimum criteria specified`,
```
dgoodwin, line 174: I'd love to see this expanded to fully explain the feature. 
 
"Group" of what?
Where is it used?
Elaborate on the different uses between payload and PR payload.
Describe an example of how it's used given our current install test checks.
Why does this command exist? What value is it providing? How could someone use it in the future for more?

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -0,0 +1,322 @@
...
+
+	fs.Var(&f.TestGroups, "test-group", "One or more test groups to analyze, like install")
+	fs.StringVar(&f.PayloadTag, "payload-tag", f.PayloadTag, "The payload tag to analyze test case status, like 4.9.0-0.ci-2021-07-19-185802")
+	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.Platform, "platform", f.Platform, "The platform used to narrow down a subset of the jobs to analyze, ex: aws|gcp|azure|vsphere")
+	fs.StringVar(&f.Infrastructure, "infrastructure", f.Infrastructure, "The infrastructure used to narrow down a subset of the jobs to analyze, ex: upi|ipi")
+	fs.StringVar(&f.Network, "network", f.Network, "The network used to narrow down a subset of the jobs to analyze, ex: sdn|ovn")
+	fs.IntVar(&f.MinimumSuccessfulTestCount, "minimum-successful-count", defaultMinimumSuccessfulTestCount, "minimum number of successful test counts among jobs meeting criteria")
+	usage := fmt.Sprintf("mutually exclusive to --payload-tag.  Matches the .label[%s] on the prowjob, which is a UID", jobrunaggregatorlib.PayloadInvocationIDLabel)
```
dgoodwin, line 141: What is it set by?

dgoodwin (commented):
Looking great just mostly just looking for clarification for the next person to touch this code, or add more checks.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
xueqzhan: This will be set by PR payload command.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
xueqzhan: Turned out it is not needed. Inherited code.

dgoodwin:
/lgtm

/hold

Do a hold cancel if this is safe to let merge prior to a hack day and long weekend, I'd be tempted to cancel it on Tuesday though when everyone is back.

PR #2729, opened 2022-03-18
Change retestOrBackoff functionality: Added option enable-on-repo

nesty156:
First part of changing retestOrBackoff. Option --enable-on-repo will create list of repositories. In this repos retestOrBackoff will be commenting instead of logging.
- I created option enable-on-repo, basic test for gaterOptions method.
- Edit option and retestController struct to contain the list of repositories.

### DIFF
--- a/cmd/better-retester/main.go
+++ b/cmd/better-retester/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
@@ -22,6 +23,8 @@ type githubClient interface {
 	QueryWithGitHubAppsSupport(ctx context.Context, q interface{}, vars map[string]interface{}, org string) error
 }
 
+type arrayFlags []string
+
 type options struct {
 	config configflagutil.ConfigOptions
 	github prowflagutil.GitHubOptions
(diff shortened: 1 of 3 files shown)


cmd/better-retester/main.go
```diff
@@ -33,6 +36,17 @@ type options struct {
 
 	cacheFile      string
 	cacheRecordAge time.Duration
+
+	enableOnRepo arrayFlags
```
petr-muller, line 40: You can use Prow's `flagutil.Strings` and then you do not need to reimplement `Set/String` ([example usage](https://github.com/openshift/ci-tools/blob/master/cmd/determinize-ci-operator/main.go#L29))

petr-muller: Also, the member is a container, so its name should indicate it (`enableOnRepos`)

petr-muller:
Ah sure, mmnt, not your fault. The base branch is stale

petr-muller:
https://github.com/openshift/ci-tools/pull/2730 should fix it

cmd/better-retester/main_test.go
```diff
@@ -0,0 +1,28 @@
...
+	}
+	return true
+}
+func TestGatherOptions(t *testing.T) {
+	expected := []string{"org/repo", "other-org/other-repo"}
+	os.Args = []string{"cmd", "--enable-on-repo=org/repo", "--cache-record-age=100h", "--enable-on-repo=other-org/other-repo"}
+
+	actual := gatherOptions()
+
+	if !stringSlicesEqual(actual.enableOnRepos.Strings(), expected) {
```
petr-muller, line 25: You should just use `cmp.Diff` like all other unit  tests do: https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff

PR #2732, opened 2022-03-21
use separate schedules for help desk and intake from triage

smg247:
I was able to add new schedules for `DPTP Help Desk` and `DPTP Intake` in PagerDuty. They follow the existing `DPTP Primary On-Call` rotation exactly, but a week and two weeks behind respectively. Now, it will be simple to add overrides to the individual schedules when someone needs a shift covered. Bonus, the logic is actually a lot simpler now in `sprint-automation`.

Completes: https://issues.redhat.com/browse/DPTP-2433


petr-muller:
/retest

b03 was in trouble

petr-muller (approved):

PR #2727, opened 2022-03-17
Create QE dedicated GCP cluster profile "gcp-qe"

jianli-wei:
After discuss with DPP and Installer dev team, QE should use our dedicated cluster profiles.

release openshift/release#27067

Profile | Profile String | Cloud Type
-- | -- | --
ClusterProfileGCPQE | gcp-qe | gcp

PR #2733, opened 2022-03-22
Create QE dedicated GCP cluster profile "gcp-qe" (using "SecretVolumeSource")

jianli-wei:
Cont. to https://github.com/openshift/ci-tools/pull/2727.

1. With the last PR, "ProjectedVolumeSource" is expected for the new cluster profile, but jobs using the profile would fail due to lack of the config-map "gcp-qe".
2. On considering that, the secrets data structures defined on [Vault](https://vault.ci.openshift.org/ui/vault/auth?with=oidc%2F) of different cloud providers are similar, I'd like to try "SecretVolumeSource" for the GCP profile and see if it can work. Or any suggestion? Thanks in advance!

### DIFF
--- a/cmd/sprint-automation/main.go
+++ b/cmd/sprint-automation/main.go
@@ -175,6 +175,8 @@ const (
 	primaryOnCallQuery     = "DPTP Primary On-Call"
 	secondaryUSOnCallQuery = "DPTP Secondary On-Call (US)"
 	secondaryEUOnCallQuery = "DPTP Secondary On-Call (EU)"
+	helpdeskQuery          = "DPTP Help Desk"
+	intakeQuery            = "DPTP Intake"
 	roleTriagePrimary      = "@dptp-triage Primary"
 	roleTriageSecondaryUS  = "@dptp-triage Secondary (US)"
 	roleTriageSecondaryEU  = "@dptp-triage Secondary (EU)"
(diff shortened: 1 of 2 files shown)


petr-muller:
Note that the existing GCP cluster profiles utilize the ConfigMap to store some content relevant to the account, such as the name of the GCP project and default public hosted zone. This content is then used by steps (example: https://github.com/openshift/release/blob/5403fcfcd8aa2aa1ca7eb36d5428db4fd7e3c3d9/ci-operator/step-registry/ipi/conf/gcp/ipi-conf-gcp-commands.sh#L9-L10), so you need to make sure this content is present in order for your GCP cluster profile to work with existing step registry content.

jianli-wei:
> Note that the existing GCP cluster profiles utilize the ConfigMap to store some content relevant to the account, such as the name of the GCP project and default public hosted zone. This content is then used by steps (example: https://github.com/openshift/release/blob/5403fcfcd8aa2aa1ca7eb36d5428db4fd7e3c3d9/ci-operator/step-registry/ipi/conf/gcp/ipi-conf-gcp-commands.sh#L9-L10), so you need to make sure this content is present in order for your GCP cluster profile to work with existing step registry content.

Yes, thanks! I did check the existing code on how the contents of ${CLUSTER_PROFILE_DIR} is used in terms of GCP, and all used keys were configured on [Vault](https://vault.ci.openshift.org/ui/vault/secrets/kv/show/selfservice/openshift-qe/gcp-qe).

jianlinliu:
> Note that the existing GCP cluster profiles utilize the ConfigMap to store some content relevant to the account, such as the name of the GCP project and default public hosted zone.

I think we can save the name of the GCP project and default public hosted zone in vault, then get them from CLUSTER_PROFILE_DIR, but not have to utilize the ConfigMap, right? Or is there some guide to tell us how to provide the ConfigMap for a new GCP cluster profile?

jianlinliu:
The PR looks good to me, need DPTP team's final review.

petr-muller:
> I think QE can save the name of the GCP project and default public hosted zone in vault, then get them from CLUSTER_PROFILE_DIR, but not have to utilize the ConfigMap, right? Or is there some guide to tell us how to provide the ConfigMap for a new GCP cluster profile? 

Yeah, if you provide the necessary keys in vault, that should work too. There's no "guide".

cmd/sprint-automation/main.go
```diff
@@ -175,6 +175,8 @@ const (
 	primaryOnCallQuery     = "DPTP Primary On-Call"
 	secondaryUSOnCallQuery = "DPTP Secondary On-Call (US)"
 	secondaryEUOnCallQuery = "DPTP Secondary On-Call (EU)"
+	helpdeskQuery          = "DPTP Help Desk"
+	intakeQuery            = "DPTP Intake"
```
petr-muller, line 179: This should not be here, looks like this some kind of `git commit --amend` that went wrong. Please clean up the PR so it does not contain unrelated changes.

petr-muller: /hold

PR #2734, opened 2022-03-22
Change retestOrBackoff functionality: Added method createComment, Edited retestOrBackoff funcionality

nesty156:
Second part of changing retestOrBackoff. Option --enable-on-repo will create list of repositories. In this repos retestOrBackoff will be commenting instead of logging.

- I created method createComment.
- Edit retestOrBackoff funcionality to start commenting instead of logging on repositories from the list

Previous PR: https://github.com/openshift/ci-tools/pull/2729

### DIFF
--- a/cmd/better-retester/retester.go
+++ b/cmd/better-retester/retester.go
@@ -15,7 +15,9 @@ import (
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/config"
+	prowflagutil "k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/git/v2"
 	"k8s.io/test-infra/prow/github"
 	"k8s.io/test-infra/prow/tide"
@@ -108,10 +110,10 @@ type retestController struct {
 	usesGitHubApp bool
 	backoff       *backoffCache
 
-	commentOnRepos []string
+	commentOnRepos sets.String
 }
 
-func newController(ghClient githubClient, cfg config.Getter, gitClient git.ClientFactory, usesApp bool, cacheFile string, cacheRecordAge time.Duration, enableOnRepos []string) *retestController {
+func newController(ghClient githubClient, cfg config.Getter, gitClient git.ClientFactory, usesApp bool, cacheFile string, cacheRecordAge time.Duration, enableOnRepos prowflagutil.Strings) *retestController {
 	logger := logrus.NewEntry(logrus.StandardLogger())
 	ret := &retestController{
 		ghClient:       ghClient,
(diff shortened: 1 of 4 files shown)


petr-muller:
/retest

The build was affected by our b03 brownouts, not your fault

cmd/better-retester/retester.go
```diff
@@ -211,21 +211,46 @@ func (b *backoffCache) check(pr tide.PullRequest, baseSha string) (retestBackoff
...
 	if err != nil {
 		return err
 	}
 
+	repo := fmt.Sprintf("%s/%s", pr.Repository.Owner.Login, pr.Repository.Name)
 	action, message := c.backoff.check(pr, baseSha)
 	switch action {
 	case retestBackoffHold:
-		c.logger.Infof("%s: %s (%s)", prUrl(pr), "/hold", message)
+		if contains(c.commentOnRepos, repo) {
```
petr-muller, line 241: Let's make `commentOnRepos` a set, then it has a `Has` method and you do not need to reimplement. The `flagutil.Strings` even has a method to make the set for you:

https://github.com/kubernetes/test-infra/blob/7092a56c8e09bb0c23d782f7a165ed923fa4915c/prow/flagutil/strings.go#L52-L55

Make sure you only create the set once though - ideally the `retestController` has the set as a member, and you make it from `flagutil.Strings` when you make an instance of `retestControler`

cmd/better-retester/retester.go
```diff
@@ -211,21 +211,46 @@ func (b *backoffCache) check(pr tide.PullRequest, baseSha string) (retestBackoff
...
+		}
 	case retestBackoffPause:
 		c.logger.Infof("%s: %s (%s)", prUrl(pr), "no comment", message)
 	case retestBackoffRetest:
-		c.logger.Infof("%s: %s (%s)", prUrl(pr), "/retest-required", message)
+		if contains(c.commentOnRepos, repo) {
+			c.createComment(pr, "/retest-required", message)
+		} else {
+			c.logger.Infof("%s: %s (%s)", prUrl(pr), "/retest-required", message)
+		}
```
petr-muller, line 253: Let's make sure we only have this logic once - the only bit that is different is the command, otherwise the logic is the same.

vendor/k8s.io/test-infra/prow/github/fakegithub/fakegithub.go
```diff
@@ -1108,6 +1108,10 @@ func (f *FakeClient) ListCurrentUserOrgInvitations() ([]github.UserOrgInvitation
 	return ret, nil
 }
 
+func (f *FakeClient) QueryWithGitHubAppsSupport(ctx context.Context, q interface{}, vars map[string]interface{}, org string) error {
+	return nil
+}
+
```
petr-muller, line 1114: You cannot simply change files under `vendor`: this is upstream code we vendor. See https://go.dev/ref/mod#vendoring for more details. If you need this change, you have two options:

1) Wrap the `FakeClient` in a downstream struct and add the missing method there:

```go
type MyFakeClient FakeClient
func (f *MyFakeClient) QueryWithGitHubAppsSupport(...
```

2) Add the change to https://github.com/kubernetes/test-infra/blob/master/prow/github/fakegithub/fakegithub.go by filing a PR to k/test-infra, and after it merges, file a PR here to bump the vendoring.

The latter is the *rght* way but it's gonna take a bit longer (especially to this feature branch) so I'd personally go with the former.

cmd/better-retester/retester_test.go
```diff
@@ -0,0 +1,86 @@
...
+				logger:         logger,
+				backoff:        &backoffCache{cache: map[string]*PullRequest{}, logger: logger},
+				commentOnRepos: enableOnRepos.StringSet(),
+			},
+			expected: "",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			tc.c.retestOrBackoff(tc.pr)
```
petr-muller, line 76: You should also check for whether the error is returned, and if yes, whether it is expected.

petr-muller (commented):
Looks good! One more nit, plus there's also a failing test, which only wants you to run `make generate` to reorder some code.

petr-muller (approved):

PR #2735, opened 2022-03-22
DPTP-2760 Resolve console host url while completing options

droslean:
/cc @openshift/test-platform 

Description of the problem:
If ci-operator fails to resolve the config from the config resolver, it tries to report the error and exit (see [ci-operator/main.go#L541-L543](https://github.com/openshift/ci-tools/blob/master/cmd/ci-operator/main.go#L541-L543)). This will generate another error while reporting the results to the results server: `response for the report was not 200: cluster field in request is empty` which is expected since by that time the cluster field is empty because it's being resolved later in the code.

This PR changes the resolving method while completing the options instead.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -483,6 +483,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 }
 
 func (o *options) Complete() error {
+	o.resolveConsoleHost()
 	jobSpec, err := api.ResolveSpecFromEnv()
 	if err != nil {
 		if len(o.gitRef) == 0 {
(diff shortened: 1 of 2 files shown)


cmd/ci-operator/main.go
```diff
@@ -483,6 +483,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 }
 
 func (o *options) Complete() error {
+	o.resolveConsoleHost()
```
petr-muller, line 486: I don't think this is sufficient to fully fix DPTP-2760. There are still cases (when `resolveConsoleHost` fails mostly) when `consoleHost` is empty, and then the report will still fail.

This patch is good, but I think we should also augment the actual reporter to handle the case of `consoleHost` being empty by e.g. putting an `unknown` there:

https://github.com/openshift/ci-tools/blob/e1f2f0d34b7c2697040c954abc696302671bb729/pkg/results/report.go#L46-L64

petr-muller (approved):

PR #2738, opened 2022-03-25
Create QE dedicated GCP cluster profile "gcp-qe" (using "SecretVolumeSource")

jianli-wei:
Cont. to https://github.com/openshift/ci-tools/pull/2727.

1. With the last PR, "ProjectedVolumeSource" is expected for the new cluster profile, but jobs using the profile would fail due to lack of the config-map "gcp-qe".
2. On considering that, the secrets data structures defined on [Vault](https://vault.ci.openshift.org/ui/vault/auth?with=oidc%2F) of different cloud providers are similar, I'd like to try "SecretVolumeSource" for the GCP profile and see if it can work. Or any suggestion? Thanks in advance!

### DIFF
--- a/pkg/prowgen/podspec.go
+++ b/pkg/prowgen/podspec.go
@@ -329,6 +329,7 @@ func generateClusterProfileVolume(profile cioperatorapi.ClusterProfile, clusterT
 		cioperatorapi.ClusterProfileAzureQE,
 		cioperatorapi.ClusterProfileAzureMagQE,
 		cioperatorapi.ClusterProfileEquinixOcpMetal,
+		cioperatorapi.ClusterProfileGCPQE,
 		cioperatorapi.ClusterProfileIBMCloud,
 		cioperatorapi.ClusterProfileLibvirtS390x,
 		cioperatorapi.ClusterProfileLibvirtPpc64le,


jianlinliu:
When squash commits for https://github.com/openshift/ci-tools/pull/2733, it get messed up, so close that one, and open this new one.

bbguimaraes:
This is unlikely to be what you want.

Your problem is the original profile (`gcp`, I suppose) is a combination of a `Secret` and a `ConfigMap`.  You added a profile called `gcp-qe`, and the way you structured the code assumes a `ConfigMap` named `cluster-profile-gcp-qe` exists (see line 372 of the file you are editing).  You have now changed it to require just the secret, but presumably a `ConfigMap` is required for your new profile to work properly just as it is for the original.

jianlinliu:
> This is unlikely to be what you want.
> 
> Your problem is the original profile (`gcp`, I suppose) is a combination of a `Secret` and a `ConfigMap`. You added a profile called `gcp-qe`, and the way you structured the code assumes a `ConfigMap` named `cluster-profile-gcp-qe` exists (see line 372 of the file you are editing). You have now changed it to require just the secret, but presumably a `ConfigMap` is required for your new profile to work properly just as it is for the original.

There was some similar discussion in https://github.com/openshift/ci-tools/pull/2733#issuecomment-1077879014

bbguimaraes:
It would be great if anyone gave DPTP a hint on how those are created too.

If you made sure all required files are passed via the secret, that might work.

/approve

jianlinliu:
> It would be great if anyone gave DPTP a hint on how those are created too.
> 
> If you made sure all required files are passed via the secret, that might work.
> 
> /approve

Seem like need to write /approve in a separate comment.

PR #2739, opened 2022-03-25
guarantee that `BASETMPDIR` starts with `TMPDIR/openshift/...`

smg247:
The `entrypoint-wrapper` integration test is failing locally on my mac with the following output:

```
FAILURE after 0.000s: test/integration/entrypoint-wrapper.sh:212: executing 'run_test test_shared_dir' expecting success: the command returned the wrong error code
[INFO] Verifying SHARED_DIR is set correctly
1c1
< /var/folders/8l/vb1frxqj6r54xmrswnvctpm40000gn/T/openshift/test-integration/tmp/secret
---
> /var/folders/8l/vb1frxqj6r54xmrswnvctpm40000gn/T//openshift/test-integration/tmp/secret

There was no error output from the command.
[ERROR] PID 13828: hack/lib/cmd.sh:6: `return "${return_code}"` exited with status 1.
```

This is due to the way the `TMPDIR` env variable is set on MacOS. With this change we guarantee that there will be exactly 1 `/` between the TMPDIR and `openshift` in `BASETMPDIR`.

### DIFF
--- a/cmd/sprint-automation/main.go
+++ b/cmd/sprint-automation/main.go
@@ -298,20 +298,41 @@ func userOnCallDuring(client *pagerduty.Client, query string, since, until time.
 		return nil, fmt.Errorf("could not query PagerDuty for the %s on-call schedule: %w", query, err)
 	}
 	if len(scheduleResponse.Schedules) != 1 {
-		return nil, fmt.Errorf("did not get exactly one schedule when querying PagerDuty for the %s on-call schedule: %v", query, scheduleResponse.Schedules)
+		return nil, fmt.Errorf("did not get exactly one schedule when querying PagerDuty for the '%s' on-call schedule: %v", query, scheduleResponse.Schedules)
 	}
+	schedule := scheduleResponse.Schedules[0]
 
-	users, err := client.ListOnCallUsers(scheduleResponse.Schedules[0].ID, pagerduty.ListOnCallUsersOptions{
+	users, err := client.ListOnCallUsers(schedule.ID, pagerduty.ListOnCallUsersOptions{
 		Since: since.String(),
 		Until: until.String(),
 	})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


bbguimaraes:
I'd rather fix the test &mdash; it's incorrect &mdash; than the test suite.

It looks like there isn't a Mac equivalent to `readlink`/`realpath`, but this should work with just `bash` (I don't have a way to test it):

```bash
diff <(echo "$v") <(cd "${TMPDIR}"/secret && pwd -P)
```

---

```
$ TMPDIR=/tmp/// test/integration/entrypoint-wrapper.sh
…
FAILURE after 0.037s: test/integration/entrypoint-wrapper.sh:212: executing 'run_test test_shared_dir' expecting success: the command returned the wrong error code
[INFO] Verifying SHARED_DIR is set correctly
1c1
< /tmp/openshift/entrypoint-wrapper/tmp/secret
---
> /tmp////openshift/entrypoint-wrapper/tmp/secret
…
```

vs.

```
$ TMPDIR=/tmp/// test/integration/entrypoint-wrapper.sh
SUCCESS after 0.078s: test/integration/entrypoint-wrapper.sh:211: executing 'run_test test_mkdir' expecting success
SUCCESS after 0.048s: test/integration/entrypoint-wrapper.sh:212: executing 'run_test test_shared_dir' expecting success
…
```

smg247:
@bbguimaraes using `pwd -P` resulted in
```
[INFO] Verifying SHARED_DIR is set correctly
1c1
< /var/folders/8l/vb1frxqj6r54xmrswnvctpm40000gn/T/openshift/test-integration/tmp/secret
---
> /private/var/folders/8l/vb1frxqj6r54xmrswnvctpm40000gn/T/openshift/test-integration/tmp/secret
```
but just dropping the `-P` allowed it to pass on my mac.

PR #2742, opened 2022-03-29
Pod Scaler Producer: use less memory at one time

smg247:
`pod-scaler-producer` is using a ton of memory creating histograms. I recently increased the memory limit to `6Gi`, and it still is not enough. A `pprof` showed that this memory is largely being consumed in the binning of the histograms. I believe that simply querying one cluster's prometheus at a time, storing the cached data and then clearing it when done, will significantly reduce memory consumption.

### DIFF
--- a/cmd/pod-scaler/producer.go
+++ b/cmd/pod-scaler/producer.go
@@ -107,7 +107,6 @@ func produce(clients map[string]prometheusapi.API, dataCache cache, ignoreLatest
 				lock: &sync.RWMutex{},
 				data: cache,
 			}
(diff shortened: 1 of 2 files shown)


petr-muller (approved):
LGTM but I'd like Bruno to take a look

smg247:
After some further research I am concerned that this approach won't work correctly. It seems that GCS bucket objects are immutable, so while I had assumed that the `storeCache` method was doing an append, it is actually doing a full re-write each time. Looks like I am going to have to come up with a different approach here.

PR #2747, opened 2022-03-31
Update the workflow correlation for AWSOSDMSP Clusterprofile

samvarankashyap:
$subject

### DIFF
--- a/cmd/repo-init/main.go
+++ b/cmd/repo-init/main.go
@@ -101,9 +101,10 @@ var (
 	// This is supposed to be a simple, non-exhaustive list of profiles that are
 	// commonly used.  Advanced users can edit the configuration themselves.
 	clusterProfiles = map[api.ClusterProfile]string{
-		api.ClusterProfileAWS:   "ipi-aws",
-		api.ClusterProfileAzure: "ipi-azure",
-		api.ClusterProfileGCP:   "ipi-gcp",
+		api.ClusterProfileAWS:       "ipi-aws",
+		api.ClusterProfileAWSOSDMSP: "osd-create",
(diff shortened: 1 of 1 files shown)


cmd/repo-init/main.go
```diff
@@ -101,9 +101,10 @@ var (
 	// This is supposed to be a simple, non-exhaustive list of profiles that are
 	// commonly used.  Advanced users can edit the configuration themselves.
 	clusterProfiles = map[api.ClusterProfile]string{
-		api.ClusterProfileAWS:   "ipi-aws",
-		api.ClusterProfileAzure: "ipi-azure",
-		api.ClusterProfileGCP:   "ipi-gcp",
+		api.ClusterProfileAWS:       "ipi-aws",
+		api.ClusterProfileAWSOSDMSP: "osd-create",
+		api.ClusterProfileAzure:     "ipi-azure",
+		api.ClusterProfileGCP:       "ipi-gcp",
```
petr-muller, line 107: I don't think the `osd-create` is a commonly used workflow, so I think there's no reason why it should be in this list. Can you explain why do you think this change is needed?

samvarankashyap: @petr-muller 
Within CVP MSP workflows we would be using osd-create workflow for frequent provisioning of osd clusters for all operator testing efforts. 
I thought it would be appropriate to place it here. 
If you think its not relevant i can close this PR. 
Thank you.

petr-muller: repo-init is a tool used by owners of repositories newly onboarded to OpenShift CI when they create their first CI setup. Unless you expect many new repositories to use his workflow in their CI jobs, I don't think it should be included.

PR #1509, opened 2020-12-09
[DPTP-1721] make repo-init tool to generate multi-stage jobs instead of templates

droslean:
This PR makes the `repo-init` tool to generate multi-stage jobs instead of template ones. 
It is currently allows to create a job using only `gcp` `aws` and `azure4` cluster profiles.

/cc @openshift/openshift-team-developer-productivity-test-platform 

Signed-off-by: Nikolaos Moraitis <nmoraiti@redhat.com>

PR #2751, opened 2022-04-04
repo-init: use correct list of cluster profiles

bbguimaraes:
This change addresses a number of problems:

- The PR used as an example in our documentation for adding cluster
  profiles incorrectly added a profile to `repo-init`'s list, so others
  have occasionally followed:

  https://docs.ci.openshift.org/docs/how-tos/adding-a-cluster-profile/#registering-a-new-profile

- `repo-init` and its "frontend" unfortunately do not share most of
  their code.

- As a result, they have used independent lists of profiles, which have
  advanced inconsistently.

To that end, I've unified the list of profiles used by both programs.
Since https://github.com/openshift/ci-tools/pull/1509, this list has
been a very reduced subset of all available profiles, so I've restricted
it to those three profiles and added a clarifying comment to discourage
future additions.

### DIFF
--- a/cmd/repo-init/main.go
+++ b/cmd/repo-init/main.go
@@ -15,6 +15,7 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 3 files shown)


bbguimaraes:
> To that end, I've unified the list of profiles used by both programs.
> Judging by the fact that the list originally added in
> bf324cf was already a very reduced
> subset of all the profiles, I've restricted it to those three profiles
> and added a clarifying comment to discourage future additions.

I've corrected this.  The initial implementation used the full profile list and the `pkg/api` data, but was changed to the subset during https://github.com/openshift/ci-tools/pull/1509.

The implementation of this PR has not changed.

/hold cancel

bbguimaraes:
I'm going to blame these on our incredibly robust servers.

/retest-required

cmd/repo-init/main.go
```diff
@@ -95,6 +96,24 @@ const (
...
+	clusterProfiles = map[api.ClusterProfile]string{
+		api.ClusterProfileAWS:   "ipi-aws",
+		api.ClusterProfileAzure: "ipi-azure",
+		api.ClusterProfileGCP:   "ipi-gcp",
+	}
+	clusterProfileList = func() (ret []api.ClusterProfile) {
+		for k := range clusterProfiles {
+			ret = append(ret, k)
+		}
+		sort.Slice(ret, func(i, j int) bool { return ret[i] < ret[j] })
```
bbguimaraes, line 112: Note: can't use `sort.Strings` here because Go is silly.

PR #2752, opened 2022-04-04
pkg/validation: reuse cluster profile list

bbguimaraes:
The `ClusterProfiles` list/function was added as part of the `repo-init`
program in bf324cf9fc816a26b63b4d3654cb465261f0d735.  It unfortunately
duplicated the list already present in `validateClusterProfiles` in
`pkg/validation`, added in 99a762598c1b07d340a96b0672485269c33051c0.

Later, 5405c167af08c0123523c5659795d4ef40320732 replaced the usage of
`ClusterProfiles` with its own small subset of profiles, and the function
hasn't been used since.

An authoritative list of profiles is still a useful reference and may be
reused in the future, so I am removing the inline list from
`pkg/validation` and making it now the sole user of the `pkg/api` function.

Other than inconsistent ordering (corrected in
e259fa4345c647c04e36bb7b3f299665a1fd8e93), the two
lists were identical.


droslean (approved):

pkg/validation/test.go
```diff
@@ -398,67 +398,17 @@ func validateTestStepDependencies(config *api.ReleaseBuildConfiguration) []error
...
+		if _, ok := v.validClusterProfiles[p]; ok {
+			return nil
+		}
+	} else {
+		for _, x := range api.ClusterProfiles() {
+			if x == p {
+				return nil
+			}
+		}
 	}
```
bbguimaraes, line 412: @droslean: sorry for the abrupt push, the only difference to the version you approved was moving the `for` into the `else` here (a (des-des-)optimization, not a functional change).

PR #2753, opened 2022-04-04
pkg/prowgen: invert profile `ConfigMap` list

bbguimaraes:
Profiles without `ConfigMap`s are clearly the common case.  Invert the
filtering rules so that virtually every new profile does not have to
change this area.

Looking at `openshift/release`, the pattern here seems to be:

- GCP and oVirt profiles legitimately use a CM.
- The remaining AWS and Azure profiles in this list maybe do or do not
  use CMs, but are only used in ancient 3.11 jobs anyway.


droslean (approved):

bbguimaraes:
I'll have to look into that `pod-scaler` test at some point…

/retest-required

PR #2754, opened 2022-04-04
During aggregation test, do not consider 'initialize' test as a failure

DennisPeriquet:
For https://issues.redhat.com/browse/TRT-176

This test always fails on a failure in an aggregation test so no point in reporting it too.


deads2k:
Try modifying in this spot: https://github.com/openshift/ci-tools/blob/049f3d99770adabe1652eeea03bbd5e680d3cdda/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go#L531

PR #2757, opened 2022-04-05
prow-job-dispatcher: add options for enabling/disabling clusters

petr-muller:
Add:
- `--enable-cluster`
- `--disable-cluster`
- `--default-cluster`
    
These options should be helpful for failovers etc. The implementation of
`prow-job-dispatcher` is a little messy, so I ended up removing the
disabled clusters from the structure before the dispatching, and then I
return them back.

Part of: [DPTP-2064](https://issues.redhat.com/browse/DPTP-2064)

/cc @hongkailiu @jmguzik @openshift/test-platform

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -22,8 +22,10 @@ import (
 	"k8s.io/test-infra/prow/cmd/generic-autobumper/bumper"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/config/secret"
+	"k8s.io/test-infra/prow/flagutil"
 	"sigs.k8s.io/yaml"
 
+	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/dispatcher"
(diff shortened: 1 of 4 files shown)


smg247:
did you consider the inverse of this? something like `exclude-clusters` might end up being simpler long-term if we end up with many clusters. Likely we would only ever need to exclude one or two rather than having to name a bunch of included clusters.

petr-muller:
@smg247 good point. I decided for this approach when I  started, because I wanted to also support the un-failover use case. But that got lost in iterations ;) You are right, let me try something else.

/hold

petr-muller:
@smg247 PTAL

/retitle prow-job-dispatcher: add options for enabling/disabling clusters
/hold cancel

hongkailiu:
Distracted a lot in the last two days.
I will take a look at this tomorrow.

petr-muller:
/retest-required

Remaining retests: 2 against base HEAD b01e35017845de4ab46f2bdf88f1bb4604e3ee1b and 8 for PR HEAD 66df0f2d910d63d8815671a2150b30dabaf498ca in total

PR #2771, opened 2022-04-14
Add '--aggr-data-start-time' parameter to target specific aggregation data

DennisPeriquet:
This is for TRT-132.

Today, aggregation data from BigQuery is taken by starting at roughly the time when the aggregation job was run, going back 10 days, and then using 7 days worth of data.

This change allows us to choose an arbitrary time instead of the time when the aggregation job was run.  This allows us to target data at a specific point in time.  For example, if there was some problem that resulted in gaps in the data, we can move to a different day and avoid the gap.

The value of the `--aggr-data-start-time` will be used as the day where we backoff by 10 days and use 7 days worth of aggregation data; example:

```
./job-run-aggregator analyze-job-runs \
    --job periodic-ci-openshift-release-master-ci-4.11-e2e-azure-ovn-upgrade \
    --payload-tag 4.11.0-0.nightly-2022-04-13-115124 \
    --explicit-gcs-prefix https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/ \
    --job-start-time 2022-04-13T12:03:59Z \
    --working-dir dp-debug1 \
    --aggr-data-start-time 2022-04-07T12:03:59Z \
    --timeout 4h40m \
    --google-service-account-credential-file openshift-ci-data-read-only.json
```

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
@@ -25,6 +26,7 @@ type JobRunsAnalyzerFlags struct {
 	ExplicitGCSPrefix           string
 	Timeout                     time.Duration
 	EstimatedJobStartTimeString string
+	AggrDataTime                string
 }
 
 func NewJobRunsAnalyzerFlags() *JobRunsAnalyzerFlags {
@@ -50,7 +52,8 @@ func (f *JobRunsAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&f.AggregationID, "aggregation-id", f.AggregationID, "mutually exclusive to --payload-tag.  Matches the .label[release.openshift.io/aggregation-id] on the prowjob, which is a UID")
 	fs.StringVar(&f.ExplicitGCSPrefix, "explicit-gcs-prefix", f.ExplicitGCSPrefix, "only used by per PR payload promotion jobs.  This overrides the well-known mapping and becomes the required prefix for the GCS query")
 	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for aggregation to complete.")
-	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC3399: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.AggrDataTime, "aggr-data-start-time", f.AggrDataTime, fmt.Sprintf("Aggregation data start time (used to target specific aggregation data) in RFC3399: %s", kubeTimeSerializationLayout))
 }
 
 func NewJobRunsAnalyzerCommand() *cobra.Command {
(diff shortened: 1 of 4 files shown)


DennisPeriquet:
With many payloads failing, I'm not sure it's a good idea to introduce this just yet.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
```diff
@@ -113,7 +115,12 @@ func (f *JobRunsAnalyzerFlags) Validate() error {
 	if len(f.AggregationID) > 0 && len(f.ExplicitGCSPrefix) == 0 {
 		return fmt.Errorf("if --aggregation-id is specified, you must specify --explicit-gcs-prefix")
 	}
-
+	if len(f.AggrDataTime) > 0 {
+		_, err := time.Parse(kubeTimeSerializationLayout, f.AggrDataTime)
+		if err != nil {
+			return fmt.Errorf("The --aggr-data-start-time value needs to be a valid time")
```
dgoodwin, line 121: You should include the error message just in case there were any details in there. Suggest return errors.Wrap(err, "your current message").

pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
```diff
@@ -51,6 +52,7 @@ func (f *JobRunsAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&f.ExplicitGCSPrefix, "explicit-gcs-prefix", f.ExplicitGCSPrefix, "only used by per PR payload promotion jobs.  This overrides the well-known mapping and becomes the required prefix for the GCS query")
 	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for aggregation to complete.")
 	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.AggrDataTime, "aggr-data-start-time", f.AggrDataTime, fmt.Sprintf("Aggregation data start time (used to target specific aggregation data) in RFC822Z: %s", kubeTimeSerializationLayout))
```
dgoodwin, line 55: Is there any reason not to use RFC3339     = "2006-01-02T15:04:05Z07:00", this seems much more common and intuitive to me. I was not immediately aware what this format was.

DennisPeriquet: apparently, we're already using RFC3339 but not saying it in the text.  Line 41 shows `const kubeTimeSerializationLayout = time.RFC3339`.  So I fixed the text to match.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
DennisPeriquet: yes, it looks good:
```
FATA[0000] Flags are invalid                             error="The --aggr-data-start-time value needs to be a valid time: parsing time \"2022-04-07T12:03:59errorZ\" as \"2006-01-02T15:04:05Z07:00\": cannot parse \"errorZ\" as \"Z07:00\""
```

DennisPeriquet:
@deads2k indicated a parameter like this might be a maintenance headache (hence my `/hold`).  In DM, deads2k mentions, `...find a way to indicate which numbers you want to use as a look-aside to the data itself`.  Can you please elaborate on what that means?

dgoodwin:
> @deads2k indicated a parameter like this might be a maintenance headache (hence my `/hold`). In DM, deads2k mentions, `...find a way to indicate which numbers you want to use as a look-aside to the data itself`. Can you please elaborate on what that means?

I believe he means instead of explicitly having to be configured with a lookback time, modifying the code itself to *detect* when it doesn't have the required two weeks? of data, and automatically finding a point in time previously where we could resume from. This will be tricky, it's relatively simple if there's NO data prior to now, we could just go back to the first time we see prow job timestamps appearing, but what if there's just one or two days of data when we started aggregation up again, in which case you'd still want to say "this isn't enough", and go find a point in time where there is a lookback. Or, maybe we do have two weeks of data, but CI was broken for a day and nothing got run for this job, we'd still want to use that two weeks despite the small day or two gap.

I'm not sure it's a maintenance headache, this would be used as a fallback when we shutoff aggregation for something like 4.10, then got into trouble and wanted to turn in back on. The maintenance then would be to manually find the startup point, configure the cronjob, and set some kind of reminder for two weeks to remove it. It might *never* get used if we're lucky. We should discuss in scrum.

DennisPeriquet:
we may never use this so I'm closing the Jira and this PR.

PR #2777, opened 2022-04-19
ci-operator multi-stage: move to separate package

bbguimaraes:
This has become a world of its own (from about 200 to 1000 lines and
growing).  Move to its own package so it can be separated into several
files.

/hold

~Includes https://github.com/openshift/ci-tools/pull/2776.~


bbguimaraes:
I know it's colossal but, just like the last one, it's mostly renaming and code shuffling.  There are no functional changes here (those will come later =), the two commits are:

0. Move the single file to its own package, adjust a few `import`s.
1. Move entire functions, unchanged, to their own files, add required `import`s to each of them.

hongkailiu (approved):

PR #2783, opened 2022-04-21
Split some aggregator code into common library

xueqzhan:


pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
```diff
@@ -87,79 +88,10 @@ func (o *JobRunAggregatorAnalyzerOptions) Run(ctx context.Context) error {
...
-		}
-
-		if len(unfinishedJobNames) > 0 {
-			fmt.Printf("%q for %q: found %d unfinished related jobRuns: %v\n", o.jobName, o.payloadTag, len(unfinishedJobNames), strings.Join(unfinishedJobNames, ", "))
-			time.Sleep(2 * time.Minute)
-			continue
-		}
-
-		break
+	finishedJobsToAggregate, _, finishedJobRunNames, unfinishedJobNames, err := jobrunaggregatorlib.WaitAndGetAllJobRuns(ctx,
```
deads2k, line 91: separate the wait and the get please.

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -0,0 +1,109 @@
...
+
+		// ready or not, it's time to check
+		if clock.Now().After(timeToStopWaiting) {
+			fmt.Printf("waited long enough. Ready or not, here I come. (readyOrNot=%v now=%v)\n", timeToStopWaiting, clock.Now())
+			break
+		}
+
+		if len(unfinishedJobRunNames) > 0 {
+			fmt.Printf("found %d unfinished related jobRuns: %v\n", len(unfinishedJobRunNames), strings.Join(unfinishedJobRunNames, ", "))
+			time.Sleep(2 * time.Minute)
```
deads2k, line 99: This is probably a move (this screams "David wrote this fast"), but please use a construct like

```go
select{
case <- time.After():
    continue
case <- ctx.Done():
    return ctx.Error()
}
```

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
```diff
@@ -9,12 +9,19 @@ import (
...
 
-func GetPayloadTagFromProwJob(prowJob *prowjobv1.ProwJob) string {
+// GetPayloadTagFromProwJobForAggregatedJob gets the payload tag from prow jobs for aggregated jobs.
+func GetPayloadTagFromProwJobForAggregatedJob(prowJob *prowjobv1.ProwJob) string {
 	return prowJob.Labels["release.openshift.io/analysis"]
 }
 
+// GetPayloadTagFromProwJobForNonAggregatedJob gets the payload tag from prow jobs for non-aggregated jobs.
+func GetPayloadTagFromProwJobForNonAggregatedJob(prowJob *prowjobv1.ProwJob) string {
+	return prowJob.Annotations["release.openshift.io/tag"]
```
deads2k, line 19: does this tag exist on all release controller created prowjobs?  Do we need to use "release.openshift.io/analysis" for some reason?

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
```diff
@@ -9,12 +9,19 @@ import (
...
+}
+
 func NewPayloadAnalysisJobLocatorForReleaseController(
 	jobName, payloadTag string,
+	isAggregated bool,
```
deads2k, line 24: this results in unreadable function calls. Create two methods or collapse the cases

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
xueqzhan: Yes, we need to use "release.openshift.io/analysis" just to get the aggregated jobs. Otherwise we will be aggregating the informing jobs as well.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
xueqzhan: Done

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
xueqzhan: Done

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
xueqzhan: Done

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
deads2k: > Yes, we need to use "release.openshift.io/analysis" to get just the aggregated jobs. Otherwise we will be aggregating the informing jobs as well.

I thought we limited by job type.

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -0,0 +1,149 @@
...
+	for {
+		fmt.Println() // for prettier logs
+		unfinishedJobRunNames = []string{}
+
+		if clock.Now().Before(readyAt) {
+			fmt.Printf("waiting until %v to collect more jobRuns before assessing finished or not. (now=%v)\n", readyAt, clock.Now())
+			time.Sleep(2 * time.Minute)
+			continue
+		}
+		fmt.Printf("it is %v, finished waiting until %v.\n", clock.Now(), readyAt)
```
deads2k, line 39: this is the initial wait.  So you have this wait as one method, then the rest as the get all jobs.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
```diff
@@ -9,12 +9,17 @@ import (
...
 )
 
-func GetAggregationIDFromProwJob(prowJob *prowjobv1.ProwJob) string {
-	return prowJob.Labels["release.openshift.io/aggregation-id"]
+const (
+	// AggregationIDLabel is the name of the label for the aggregation id in prow job
+	AggregationIDLabel = "release.openshift.io/aggregation-id"
+)
+
+func getMatchIDFromProwJob(prowJob *prowjobv1.ProwJob, label string) string {
```
deads2k, line 17: this method doesn't look useful

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
xueqzhan: In my current change, all wait (ReadyAt and timeToStopWaiting) are both in the wait function. By the time this function returns, the wait is over, either all jobs done or the timeToStopWaiting reached. Did you mean to have something different?

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
xueqzhan: Not for now. But this will be used later for matching either SheduleID or payload invocation ID. So it is useful.

deads2k: > Not for now. But this will be used later for matching either SheduleID or payload invocation ID. So it is useful.

A method that does nothing but lookup a map value isn't one we should have.

deads2k: When the map value was fixed, it made some sense.  But if you're passing the map value, it's not doing anythign for you.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
xueqzhan: I am still going to differentiate between aggregated vs non-aggregated in this PR. If we want to use "release.openshift.io/tag" for all jobs, I will test the path in a separate PR.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
```diff
@@ -36,11 +43,17 @@ func NewPayloadAnalysisJobLocatorForReleaseController(
 }
 
 type releaseControllerProwJobMatcher struct {
-	payloadTag string
+	payloadTag   string
+	isAggregated bool
```
deads2k, line 47: you're plumbing this through, but as we discussed on slack with brad, the distinction doesn't matter, so this is adding complexity without benefit.

xueqzhan: As of now, aggregator is only coded to work with 10 jobs. So we need to differentiate aggregated and non-aggregated jobs. Otherwise, aggregator will panic when processing old data.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
xueqzhan: Items added to lookup table. So we can use a single payload tag label now.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_release_jobs.go
xueqzhan: Done

PR #2785, opened 2022-04-22
Update slackbot readme

psalajova:
Updated the README file so it describes the new messaging functionality and also existing functionality. Not sure if it is detailed enough.

https://issues.redhat.com/browse/DPTP-2746
/cc smg247


smg247:
I think this is a good level of detail. Someone new to the project would be able to get a high-level overview of what the point of the `slack-bot` was, and if they need more detail they should look at the code.

/lgtm

PR #2788, opened 2022-04-23
Upgrade to go 1.18

hongkailiu:
/cc @openshift/test-platform


smg247:
@hongkailiu we actually need this one so that `test-infra` can be bumped. Do you know how to fix the test failures?

hongkailiu:
> @hongkailiu we actually need this one so that `test-infra` can be bumped. Do you know how to fix the test failures?

Tried several things yesterday. No luck.
I can reproduce it locally (with go 1.18) tough.

smg247:
I think we can just change `pkg/html/html.go#51` to `if _, err := fmt.Fprint(w, pageEnd); err != nil {` now...

PR #2793, opened 2022-04-27
Update `test-infra` dependency

smg247:
Bumping `test-infra`. Quite a few breaking changes here, but they were fairly straightforward to port. A noteworthy change to the secret censor now results in 'X' instead of '*' in the output.


smg247:
https://github.com/openshift/release/pull/28126 will take care of the `breaking-changes` test failure

pkg/steps/multi_stage/testdata/zz_fixture_TestGeneratePods.yaml
```diff
@@ -103,6 +103,7 @@
       image: sidecar
       name: sidecar
       resources: {}
+      terminationMessagePolicy: FallbackToLogsOnError
```
hongkailiu, line 106: I bet this is the source of breaking

hongkailiu: I lost the bet.

hongkailiu:
LGTM.

I guess we need to fix the breaking?

pkg/steps/multi_stage/testdata/zz_fixture_TestGeneratePods.yaml
smg247: It's actually the new config I am specifically trying to get by bumping this that is causing the breaking changes.

smg247:
> LGTM.
> 
> I guess we need to fix the breaking?

https://github.com/openshift/release/pull/28126 will do it

hongkailiu:
/lgtm
/hold

Cancel whenever you are ready.

PR #2790, opened 2022-04-25
ci-operator multi-stage: add flags member

bbguimaraes:
`ci-tools` has a chronic problem with cryptic booleans everywhere.
More configuration options will be added here soon, add a flags member
in preparation.

PR #2795, opened 2022-04-27
ci-operator multi-stage: simplify best-effort test

bbguimaraes:
Return the set directly instead of a closure.

The set bound to the closure is only ever populated based on the input
configuration, there is no need to have dynamic behavior after
construction.

/hold

~Includes https://github.com/openshift/ci-tools/pull/2790 to use the new `flags` member.~


bbguimaraes:
https://issues.redhat.com/browse/DPTP-2788

PR #2803, opened 2022-05-06
entrypoint-wrapper: add option to wait for file

bbguimaraes:
This can be used to synchronize the wrapped command with some external
process, e.g. the establishment of a VPN connection.

### DIFF
--- a/cmd/entrypoint-wrapper/main.go
+++ b/cmd/entrypoint-wrapper/main.go
@@ -17,6 +17,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify.v1"
 
 	coreapi "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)


cmd/entrypoint-wrapper/main.go
```diff
@@ -170,6 +191,40 @@ func copyDir(dst, src string) error {
 	return nil
 }
 
+func waitForFile(path string, timeout time.Duration) error {
```
droslean, line 194: Can we just use https://github.com/kubernetes/test-infra/blob/master/prow/config/agent.go#L140 instead and pass the extra logic in the callback fn?

bbguimaraes: No, because it for some unknown reason does not give the event function… the event.  The way the `entrypoint-wrapper` function is set up, there would be no ambiguity since we guarantee the file does not previously exist and only watch a single file, but making those assumptions implicit would be sloppy programming, IMO.

It would also help very little in reducing this already tiny (~30-line) function, and the acrobatics to set both callbacks up only to then cancel everything on the first event would make it less clear in the end.

bbguimaraes:
More `pod-scaler` failures =/

/retest

jmguzik (approved):
Looks nice.

PR #2804, opened 2022-05-06
Validate the name of all releases

hongkailiu:
https://coreos.slack.com/archives/GB7NB0CUC/p1651853180600959

Check all releases, instead of only integration releases.

/cc @openshift/test-platform

### DIFF
--- a/pkg/validation/release.go
+++ b/pkg/validation/release.go
@@ -18,6 +18,9 @@ func validateReleases(fieldRoot string, releases map[string]api.UnresolvedReleas
 		names.Insert(name)
 	}
 	for _, name := range names.List() {
+		if err := partOfImageStreamName(name); err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%s]: the release name is not valid: %w", fieldRoot, name, err))
+		}
 		release := releases[name]
 		if hasTagSpec {
 			for _, incompatibleName := range []string{api.LatestReleaseName, api.InitialReleaseName} {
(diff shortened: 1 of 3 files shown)


hongkailiu:
https://issues.redhat.com/browse/DPTP-2858

pkg/validation/release.go
```diff
@@ -18,6 +18,9 @@ func validateReleases(fieldRoot string, releases map[string]api.UnresolvedReleas
 		names.Insert(name)
 	}
 	for _, name := range names.List() {
+		if err := partOfImageStreamName(name); err != nil {
+			validationErrors = append(validationErrors, fmt.Errorf("%s[%s]: the release name is not a valid part of an image stream name: %w", fieldRoot, name, err))
```
bbguimaraes, line 22: Users generally don't know the implementation details of release image streams, so this is too specific.  Just saying it's invalid because it has a period should be enough.

pkg/validation/release_test.go
```diff
@@ -155,7 +156,7 @@ func TestValidateReleases(t *testing.T) {
 			},
 		},
 		{
-			name: "invalid name of integration release",
+			name: "invalid release name",
```
bbguimaraes, line 159: Add some cases here covering the other types that are now validated, just for completeness.

pkg/validation/release_test.go
```diff
@@ -171,7 +172,7 @@ func TestValidateReleases(t *testing.T) {
 				},
 			},
 			output: []error{
-				errors.New("the name of an integration release must not contain '.' but found root.ocp-4.11"),
+				fmt.Errorf("root[ocp-4.11]: the release name is not a valid part of an image stream name: %w", fmt.Errorf("must not contain '.'")),
```
bbguimaraes, line 175: Not terribly important, but since you'll need to change it: I think you can just inline the inner `fmt.Errorf()` in the message string of the outer one.

bbguimaraes (changes requested):

pkg/validation/release.go
hongkailiu: Done.

pkg/validation/release_test.go
hongkailiu: Done.

pkg/validation/release_test.go
hongkailiu: Tried before my first commit.
I cannot make it work.

hongkailiu:
I will fix the breaking the next Monday.

PR #2805, opened 2022-05-09
Introduce a watcher for a symlink release repo path

droslean:
/cc @bbguimaraes @openshift/test-platform 

- Adds a `--release-repo-path` flag
- Start watcher for the release repo folder symlink destination

If the symlink of the release repo path will change, then the watcher starts watching the new destination and reload the configs once.

https://issues.redhat.com/browse/DPTP-2531

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -7,10 +7,12 @@ import (
 	"io/fs"
 	"net/http"
 	"os"
+	"path/filepath"
 	"strconv"
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify.v1"
 
 	prowConfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/flagutil"
@@ -26,11 +28,17 @@ import (
 	"github.com/openshift/ci-tools/pkg/webreg"
 )
 
+const (
+	releaseRepoConfigsPath  = "ci-operator/config"
+	releaseRepoRegistryPath = "ci-operator/step-registry"
+)
+
 type options struct {
 	configPath             string
 	registryPath           string
 	logLevel               string
 	address                string
+	releaseRepoGitSyncPath string
 	port                   int
 	uiAddress              string
 	uiPort                 int
@@ -49,6 +57,7 @@ func gatherOptions() (options, error) {
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	fs.StringVar(&o.configPath, "config", "", "Path to config dirs")
 	fs.StringVar(&o.registryPath, "registry", "", "Path to registry dirs")
+	fs.StringVar(&o.releaseRepoGitSyncPath, "release-repo-git-sync-path", "", "Path to release repository dir")
 	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
 	fs.StringVar(&o.address, "address", ":8080", "DEPRECATED: Address to run server on")
 	fs.StringVar(&o.uiAddress, "ui-address", ":8082", "DEPRECATED: Address to run the registry UI on")
@@ -65,29 +74,50 @@ func gatherOptions() (options, error) {
 	return o, nil
 }
 
-func validateOptions(o options) error {
+func validateOptions(o *options) error {
 	_, err := logrus.ParseLevel(o.logLevel)
 	if err != nil {
 		return fmt.Errorf("invalid --log-level: %w", err)
 	}
-	if o.configPath == "" {
-		return fmt.Errorf("--config is required")
+
+	if o.releaseRepoGitSyncPath != "" && (o.configPath != "" || o.registryPath != "") {
+		return fmt.Errorf("--release-repo-path is mutually exclusive with --config and --registry")
 	}
-	if _, err := os.Stat(o.configPath); err != nil {
-		if os.IsNotExist(err) {
-			return fmt.Errorf("--config points to a nonexistent directory: %w", err)
+
+	if o.releaseRepoGitSyncPath == "" {
+		if o.configPath == "" {
+			return fmt.Errorf("--config is required")
 		}
-		return fmt.Errorf("Error getting stat info for --config directory: %w", err)
-	}
-	if o.registryPath == "" {
-		return fmt.Errorf("--registry is required")
-	}
-	if _, err := os.Stat(o.registryPath); err != nil {
-		if os.IsNotExist(err) {
-			return fmt.Errorf("--registry points to a nonexistent directory: %w", err)
+
+		if _, err := os.Stat(o.configPath); err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("--config points to a nonexistent directory: %w", err)
+			}
+			return fmt.Errorf("error getting stat info for --config directory: %w", err)
 		}
-		return fmt.Errorf("Error getting stat info for --registry directory: %w", err)
+
+		if o.registryPath == "" {
+			return fmt.Errorf("--registry is required")
+		}
+
+		if _, err := os.Stat(o.registryPath); err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("--registry points to a nonexistent directory: %w", err)
+			}
+			return fmt.Errorf("error getting stat info for --registry directory: %w", err)
+		}
+	} else {
+		if _, err := os.Stat(o.releaseRepoGitSyncPath); err != nil {
+			if os.IsNotExist(err) {
+				return fmt.Errorf("--release-repo-path points to a nonexistent directory: %w", err)
+			}
+			return fmt.Errorf("error getting stat info for --release-repo-path directory: %w", err)
+		}
+
+		o.configPath = filepath.Join(o.releaseRepoGitSyncPath, releaseRepoConfigsPath)
+		o.registryPath = filepath.Join(o.releaseRepoGitSyncPath, releaseRepoRegistryPath)
 	}
+
 	if o.validateOnly && o.flatRegistry {
 		return errors.New("--validate-only and --flat-registry flags cannot be set simultaneously")
 	}
@@ -119,18 +149,44 @@ func main() {
 	if err != nil {
 		logrus.WithError(err).Fatal("failed go gather options")
 	}
-	if err := validateOptions(o); err != nil {
+	if err := validateOptions(&o); err != nil {
 		logrus.Fatalf("invalid options: %v", err)
 	}
 	level, _ := logrus.ParseLevel(o.logLevel)
 	logrus.SetLevel(level)
 
-	configAgent, err := agents.NewConfigAgent(o.configPath, agents.WithConfigMetrics(configresolverMetrics.ErrorRate))
+	configAgentOption := func(*agents.ConfigAgentOptions) {}
+	registryAgentOption := func(*agents.RegistryAgentOptions) {}
+	if o.releaseRepoGitSyncPath != "" {
+		eventCh := make(chan fsnotify.Event)
+		errCh := make(chan error)
+
+		universalSymlinkWatcher := &agents.UniversalSymlinkWatcher{
+			EventCh:   eventCh,
+			ErrCh:     errCh,
+			WatchPath: o.releaseRepoGitSyncPath,
+		}
+
+		configAgentOption = func(opt *agents.ConfigAgentOptions) {
+			opt.UniversalSymlinkWatcher = universalSymlinkWatcher
+		}
+		registryAgentOption = func(opt *agents.RegistryAgentOptions) {
+			opt.UniversalSymlinkWatcher = universalSymlinkWatcher
+		}
+
+		watcher, err := universalSymlinkWatcher.GetWatcher()
+		if err != nil {
+			logrus.Fatalf("Failed to get the universal symlink watcher: %v", err)
+		}
+		interrupts.Run(watcher)
+	}
+
+	configAgent, err := agents.NewConfigAgent(o.configPath, agents.WithConfigMetrics(configresolverMetrics.ErrorRate), configAgentOption)
 	if err != nil {
 		logrus.Fatalf("Failed to get config agent: %v", err)
 	}
 
-	registryAgent, err := agents.NewRegistryAgent(o.registryPath, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry))
+	registryAgent, err := agents.NewRegistryAgent(o.registryPath, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry), registryAgentOption)
 	if err != nil {
 		logrus.Fatalf("Failed to get registry agent: %v", err)
 	}
--- a/pkg/load/agents/utils.go
+++ b/pkg/load/agents/utils.go
@@ -3,62 +3,145 @@ package agents
 import (
 	"context"
 	"fmt"
+	"os"
+	"path/filepath"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
-	log "github.com/sirupsen/logrus"
 	"gopkg.in/fsnotify.v1"
 
 	"k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/interrupts"
 )
 
-func startWatchers(path string, callback func() error, recordError func(string)) error {
-	cms, dirs, err := config.ListCMsAndDirs(path)
-	if err != nil {
-		return err
-	}
-	errFunc := func(err error, msg string) {
-		recordError(msg)
-		log.WithError(err).Error(msg)
-	}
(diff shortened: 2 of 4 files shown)


pkg/load/agents/utils.go
```diff
@@ -62,3 +63,72 @@ func startWatchers(path string, callback func() error, recordError func(string))
...
+	if err != nil {
+		return nil, err
+	}
+
+	resolveSymlink := func() string { return filepath.Join(filepath.Dir(path), oldSymlink) }
+
+	if err = w.Add(resolveSymlink()); err != nil {
+		return nil, fmt.Errorf("failed to add %s to watcher: %v", path, err)
+	}
+	logrus.Infof("Watching %s", resolveSymlink())
```
bbguimaraes, line 100: Should probably log every transition since we're doing this.

cmd/ci-operator-configresolver/main.go
```diff
@@ -119,18 +148,18 @@ func main() {
...
 	logrus.SetLevel(level)
 
-	configAgent, err := agents.NewConfigAgent(o.configPath, agents.WithConfigMetrics(configresolverMetrics.ErrorRate))
+	configAgent, err := agents.NewConfigAgent(o.configPath, agents.WithConfigMetrics(configresolverMetrics.ErrorRate), agents.WithReleaseRepoPath(o.releaseRepoPath))
 	if err != nil {
 		logrus.Fatalf("Failed to get config agent: %v", err)
 	}
 
-	registryAgent, err := agents.NewRegistryAgent(o.registryPath, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry))
+	registryAgent, err := agents.NewRegistryAgent(o.registryPath, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry), agents.WithRegistryReleaseRepoPath(o.releaseRepoPath))
```
bbguimaraes, line 162: I know this is not new, but with separate agents/watchers, can we get into a situation where a request is serviced after `configAgent` has reloaded the files but `registryAgent` hasn't yet (or the reverse)?

pkg/load/agents/utils.go
```diff
@@ -62,3 +63,72 @@ func startWatchers(path string, callback func() error, recordError func(string))
...
+	}
+
+	oldSymlink, err := os.Readlink(path)
+	if err != nil {
+		return nil, err
+	}
+
+	resolveSymlink := func() string { return filepath.Join(filepath.Dir(path), oldSymlink) }
+
+	if err = w.Add(resolveSymlink()); err != nil {
```
bbguimaraes, line 97: I don't think this is what we want.  It will generate an event when the old directory is removed, not when the symbolic link is updated.

pkg/load/agents/utils.go
```diff
@@ -62,3 +63,72 @@ func startWatchers(path string, callback func() error, recordError func(string))
...
+
+// GetSymlinkWatcher returns a function that watches if the destination of a single symbolic link changed and run
+// the event function against the new destination.
+func GetSymlinkWatcher(eventFunc func() error, errFunc func(error, string), path string) (func(ctx context.Context), error) {
+	w, err := fsnotify.NewWatcher()
+	if err != nil {
+		return nil, err
+	}
+
+	oldSymlink, err := os.Readlink(path)
```
bbguimaraes, line 90: There is a race here.  The link can change between this and the `Add` below.

bbguimaraes (commented):
Some comments.  This is very tricky to get right, as we've seen in production with the existing implementation.

I know this is the status quo, and thanks for working on making it better, but the fact that this is a critical bug which is not caught in any of our tests is really bad, we need some unit/integration tests.

It's also kind of sad that the upstream client does exactly what's necessary, but has a hard-coded path for the symlink (the `ConfigMap` mount).  I don't know how much of an interest there would be in supporting `git-sync` there.

cmd/ci-operator-configresolver/main.go
droslean: The way the code is designed, we will run 2 identical watchers, but I don't think we will get into a situation where one is updated and the other not since they are programmed to respond to the same event.

cmd/ci-operator-configresolver/main.go
```diff
@@ -49,6 +56,7 @@ func gatherOptions() (options, error) {
 	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
 	fs.StringVar(&o.configPath, "config", "", "Path to config dirs")
 	fs.StringVar(&o.registryPath, "registry", "", "Path to registry dirs")
+	fs.StringVar(&o.releaseRepoPath, "release-repo-path", "", "Path to release repository dir")
```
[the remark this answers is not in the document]
droslean: Something like `--symlink-release-repo-path` should be descriptive enough.

pkg/load/agents/utils.go
```diff
@@ -62,3 +63,72 @@ func startWatchers(path string, callback func() error, recordError func(string))
...
+	if err != nil {
+		return nil, err
+	}
+
+	resolveSymlink := func() string { return filepath.Join(filepath.Dir(path), oldSymlink) }
```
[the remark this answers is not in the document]
droslean: Can you elaborate more on this one? Do you mean that the `path` variable won't be available inside the closure?

pkg/load/agents/utils.go
```diff
@@ -62,3 +63,72 @@ func startWatchers(path string, callback func() error, recordError func(string))
...
+
+	if err = w.Add(resolveSymlink()); err != nil {
+		return nil, fmt.Errorf("failed to add %s to watcher: %v", path, err)
```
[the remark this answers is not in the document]
droslean: There is no open watcher actually. When the folder is removed, the watcher will stop watching it as well.

cmd/ci-operator-configresolver/main.go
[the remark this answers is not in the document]
bbguimaraes: I think we need to be more explicit.  `release-repo-git-sync-path` or something similar.

cmd/ci-operator-configresolver/main.go
bbguimaraes: What if this happens?

- `configAgent` is updated to include a test which uses a new registry component
- request arrives here
- `registryAgent` is updated to include the new registry component

pkg/load/agents/utils.go
bbguimaraes: This could happen:

- `oldSymlink = "d0"`
- `git-sync` updates the directory and removes `d0`
- `w.Add` below will try to add `d0` and fail

pkg/load/agents/utils.go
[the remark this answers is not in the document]
bbguimaraes: `path` does not change, so neither does `filepath.Dir(path)`.  It can be calculated once outside the closure, instead of in each call:

```go
path := "test/test"
f := func() string { return filepath.Dir(path) }
for { f() }
```
```asm
0000000000462220 <main.main>:
…
  462234:       48 8d 05 c1 0e 01 00    lea    rax,[rip+0x10ec1]        # 4730fc <go.string.*+0xbac>
  46223b:       bb 09 00 00 00          mov    ebx,0x9
  462240:       e8 7b fe ff ff          call   4620c0 <path/filepath.Dir>
  462245:       eb ed                   jmp    462234 <main.main+0x14>
…
```

pkg/load/agents/utils.go
bbguimaraes: That's because the operation is done on the directory, not the link.

https://github.com/kubernetes/git-sync/blob/c71d1cb890de054a8be2ee69d023a8c5cf4e1ad1/cmd/git-sync/main.go#L879-L888

```
$ mkdir test; cd test
$ mkdir a b
$ ln -s a l
$ inotifywait --monitor . &
[1] 2768156
$ Setting up watches.
Watches established.

$ ln -snf b tmp
./ CREATE tmp
$ mv -T tmp l
./ MOVED_FROM tmp
./ MOVED_TO l
```

pkg/load/agents/utils.go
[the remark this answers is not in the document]
droslean: Ok I will make sure that the watcher is closed before throwing the error.

cmd/ci-operator-configresolver/main.go
droslean: When the new symlink destination will be created, both of the agents will respond in the same event and reload their configs once.

pkg/load/agents/utils.go
droslean: `w.Add` will add the new symlink instead.

pkg/load/agents/utils.go
droslean: Ok, that was introduced in the latest version of `git-sync`. It seems that I was using an older version. Let me check and update the code based on the new behaviour.

droslean: I think the problem here is how golang is watching the folder. When we start watch the `release` folder which has `rev-123` symlink, `git-sync` creates a new symlink with the new destination `rev-345` as `tmp-link` and then renames the `tmp-link` to `release`, but we still get an event from the old `release` since the new rename event is for `tmp-link` which we don't watch. So if we could just watch the destination folder instead, we can assume that if is deleted, then a new one is created and we can get the new destination name from the new `release` symlink.

bbguimaraes: Yes, which is why we need to monitor the directory:
```go
w, err := fsnotify.NewWatcher()
// …
err := w.Add(".")
// …
for e := range w.Events {
	fmt.Printf("%s\n", e)
}
```

```
$ ./test &
[1] 2683467
$ ln -s a tmp
"./tmp": CREATE
$ mv tmp l
"./tmp": RENAME
"./l": CREATE
$ ln -s b tmp
"./tmp": CREATE
$ mv tmp l
"./tmp": RENAME
"./l": CREATE
$ ln -s a tmp
"./tmp": CREATE
$ mv tmp l
"./tmp": RENAME
"./l": CREATE
```

This is the intended usage of `ConfigMap` mounts, as described [here](https://deploy-preview-233--zealous-shirley-0f56b3.netlify.app/docs/architecture/ci-operator-internals/configuration-updates/#atomicwriter).  It is also how the [`test-infra` agent works](https://github.com/kubernetes/test-infra/blob/07586f2d13d815a4e3053edac7098c36d95a71cb/prow/config/agent.go#L79.).  The recent `entrypoint-wrapper` addition is another [example](https://github.com/openshift/ci-tools/blob/d9519e391da9003ee75b5700e359a8bb9ed7aefb/cmd/entrypoint-wrapper/main.go#L194).

cmd/ci-operator-configresolver/main.go
bbguimaraes: Right, but they operate without knowledge of each other.  How do we prevent the situation I described?  Suppose we are updating to a version that adds both a registry component and a `ci-operator` configuration file which uses it.

pkg/load/agents/utils.go
bbguimaraes: It will not, `oldSymlink` will not have changed.

cmd/ci-operator-configresolver/main.go
droslean: But both agents will be updated simultaneously.

pkg/load/agents/utils.go
droslean: I am not sure how this will improve the situation. Why do we need to watch the entire directory? We will still ignore all the events and do the same thing as we do now.

pkg/load/agents/utils.go
droslean: At this point, we just initialize the current symlink. By the time we will add the new folder to the watcher, the `oldSymlink` will change.

cmd/ci-operator-configresolver/main.go
bbguimaraes: Our agents are essentially doing (independently):

```go
for {
    waitForChanges()
    newConfig = loadConfig()
}
```

- [`configAgent`](https://github.com/openshift/ci-tools/blob/d9519e391da9003ee75b5700e359a8bb9ed7aefb/pkg/load/agents/configAgent.go#L221-L243)
- [`registryAgent`](https://github.com/openshift/ci-tools/blob/master/pkg/load/agents/registryAgent.go#L129-L155)

Because there is no synchronization between them, the situation in https://github.com/openshift/ci-tools/pull/2805#discussion_r873811638 would be:

- both agents are notified of the change
- `configAgent` finishes `newConfig = loadConfig()`
- a request arrives
- `registryAgent` finishes `newConfig = loadConfig()`

The request will fail, since the configuration returned by `configAgent` references a registry component which the `registryAgent` does not (yet) know.

droslean: They are doing it independently but the same event will trigger them. With some divine power, they shouldn't race :). Let me check what it takes to refactor the whole thing to use the same watcher.

pkg/load/agents/utils.go
droslean: It is also safe because we can't rely on reloading the configs on a `CREATE` event for example. We will still rely on the `git-sync` implementation to figure out when is the right time to reload our configs. I will change the code to watch the entire folder instead, but I will need to see the `git-sync` code to determine when we will trigger the reload. I guess on `RENAME`.

droslean: Not to mention that we will have to implement a new watcher from scratch since the fsnotify won't watch any changes inside the directory.

cmd/ci-operator-configresolver/main.go
bbguimaraes: To be clear: this problem has always existed in `ci-operator-configresolver`.  It is also somewhat improved by the use of a single directory instead of multiple `ConfigMap` mounts.  Before we had _no guarantee_ about when the registry update would happen in relation to the configuration update.

But if we're making changes here in order to solve the configuration update problem (which is the goal of DPTP-2531), we should at least have a plan for what to do about this.

droslean: Before those agents were watching two different things. Now the agents are watching the exact same thing. It is quite unlike that they will race. Either way, I think it should be better to have only 1 watcher instead. I will do the changes.

droslean:
@bbguimaraes with the latest updates on this PR, we will run only one watcher, and the config and registry agents will just receive from that channel in order to update their configs. One thing that didn't change is the way we watch the release repo. We still waiting for a `REMOVE` event and getting the new symlink location to load the config since there is no good way to watch the entire folder.

bbguimaraes:
I'm struggling with how to make this clear.  The problem is not that the updates are triggered or started separately, it's that they are _applied_ separately.  We cannot have one agent do `a.config = newConfig` before the other: that has to be an atomic step, otherwise any request which occurs between the two can fail.

But, to reiterate: I don't think we need to fix this particular problem in this PR.

bbguimaraes:
For the watcher, this works for me:
```go
func main() {
	w, _ := fsnotify.NewWatcher()
	w.Add(os.Args[1])
	for e := range w.Events {
		fmt.Printf("event: %s\n", e)
		if e.Name != os.Args[2] { continue }
		d, _ := os.Readlink(e.Name)
		fmt.Println("new path:", d)
	}
}
```
```
$ git init --quiet test.git
$ git -C test.git commit --quiet --allow-empty --message initial
$ git-sync --repo $PWD/test.git --root $PWD/repo --wait 1 &> /dev/null &
[1] 3818322
$ (cd repo && ../test) &
[2] 3818545
$ git -C test.git commit --quiet --allow-empty --message $RANDOM
event: "./432a42870a34551a1376d43014ccb5ae71c7e587": CREATE
event: "./tmp-link": CREATE
event: "./tmp-link": RENAME
event: "./test.git": CREATE
new path: 432a42870a34551a1376d43014ccb5ae71c7e587
event: "./689a49faff9b27cda2f594ec7ea7e607d279cb4a": REMOVE
$ git -C test.git commit --quiet --allow-empty --message $RANDOM
event: "./451516cd48ce8a291e6783829756397c1b474972": CREATE
event: "./tmp-link": CREATE
event: "./tmp-link": RENAME
event: "./test.git": CREATE
new path: 451516cd48ce8a291e6783829756397c1b474972
event: "./432a42870a34551a1376d43014ccb5ae71c7e587": REMOVE
```

droslean:
@bbguimaraes It is either we trigger the config reloading when a new folder is being created or if the old one is removed (current implementation). If we want to go with the first option, my concern is that it can race because it is not guaranteed that `git-sync` finished its magic when creating a new folder. As you can see in your example the removal of the old folder is the last event, so we assume that the new folder is ready to rock. If you insist I can implement the first option and check for races either way. WDYT?

bbguimaraes:
That is the protocol and how it is intended to be used.  Again, see https://deploy-preview-233--zealous-shirley-0f56b3.netlify.app/docs/architecture/ci-operator-internals/configuration-updates/#atomicwriter.

The directory is populated, _then_ the symlink is atomically updated, exactly to allow this usage.

droslean:
> The directory is populated, then the symlink is atomically updated, exactly to allow this usage.

This is exactly my concern. If we trigger the reload when the new folder is created, we could end up reloading the old configs because the symlink is updated after that event. We could race.

droslean:
@bbguimaraes updated based on https://github.com/openshift/ci-tools/pull/2805#issuecomment-1137302557

PR #2807, opened 2022-05-10
Pr Reminder tool

smg247:
See `README.md` for a description of this new tool.
This is quite a large PR, but I don't see anyway to break this up into a smaller, useful chunk.
The periodic to run this will be added in a forthcoming PR.

For [DPTP-2745](https://issues.redhat.com/browse/DPTP-2745)
/cc @openshift/test-platform

Screen Shot 2022-05-10 at 10 02 30 AM

### DIFF
--- a/cmd/pr-reminder/main.go
+++ b/cmd/pr-reminder/main.go
@@ -109,23 +109,23 @@ func (p prRequest) link() string {
 }
 
 const (
-	recent = " :large_green_circle:"
-	normal = " :large_orange_circle:"
-	old    = " :red_circle:"
+	recent = ":large_green_circle:"
+	normal = ":large_orange_circle:"
+	old    = ":red_circle:"
 )
 
 func (p prRequest) createdUpdatedMessage() string {
-	message := fmt.Sprintf("Created: %s | Updated: %s", p.created.Format(time.RFC1123), p.lastUpdated.Format(time.RFC1123))
+	var recency string
 	// PRs that have been updated in the last day should be called out
 	now := time.Now()
 	if p.created.After(now.Add(-time.Hour * 24 * 2)) {
-		message += recent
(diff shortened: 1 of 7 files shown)


cmd/pr-reminder/main.go
```diff
@@ -0,0 +1,344 @@
...
+
+const (
+	recent = ":large_green_circle:"
+	normal = ":large_orange_circle:"
+	old    = ":red_circle:"
+)
+
+func (p prRequest) recency() string {
+	now := time.Now()
+	if p.Created.After(now.Add(-time.Hour * 24 * 2)) {
```
jmguzik, line 129: maybe const out of this value?

cmd/pr-reminder/main.go
```diff
@@ -0,0 +1,344 @@
...
+		return recent
+	} else if p.Created.After(now.Add(-time.Hour * 24 * 7)) {
```
jmguzik, line 131: as above

cmd/pr-reminder/main.go
```diff
@@ -0,0 +1,344 @@
...
+		}
+	}
+	if len(usersMissingGithubId) > 0 {
+		return nil, fmt.Errorf("no githubId found for user(s): %v", usersMissingGithubId)
+	}
+
+	return users, nil
+}
+
+func messageUser(user user, slackClient slackClient) error {
```
jmguzik, line 278: This is just a suggestion: I would split this function into smaller chunks. It seems to be quite long.

smg247: It is quite long, but its 90% simple boiler plate that the slack client requires to send a message. I think breaking it up would just make it even more time consuming to read.

jmguzik (approved):
Really nice work!

PR #2814, opened 2022-05-12
backporter: fix a logging

hongkailiu:
To debug for
https://coreos.slack.com/archives/CBN38N3MW/p1652351363015799

and found no logs was printed

/cc @openshift/test-platform


smg247:
The `e2e` is probably a flake, but it looks like the `unit` does need to be updated.

PR #2815, opened 2022-05-13
Before branching day: step 1

danilo-gemoli:
This is a tool that tries to automate the first step of phase "Before branching day" .
Check [Centralized Release Branching and Config Management V3](https://docs.google.com/document/d/1Z6ejnDCOCvNv9PWkyNPzVbjuLbDMAAT5GEeDpzb0SMs/edit#heading=h.r9xn02r1cyfn)
/cc @jmguzik @droslean

### DIFF
--- a/pkg/branchcuts/bumper/bumper.go
+++ b/pkg/branchcuts/bumper/bumper.go
@@ -1,0 +1,110 @@
+package bumper
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/sirupsen/logrus"
+
+	"github.com/openshift/ci-tools/pkg/util"
+)
+
+type Bumper[T any] interface {
+	GetFiles() ([]string, error)
+
+	Unmarshall(file string) (T, error)
+
+	BumpFilename(filename string, obj T) (string, error)
+
+	BumpContent(obj T) (T, error)
+
+	Marshall(obj T, bumpedFilename, dir string) error
+}
+
+type BumpingOptions struct {
+	OutDir string
+}
+
+// Bump bumps files using the Bumpers b according to the BumpingOptions.
+func Bump[T any](b Bumper[T], o *BumpingOptions) error {
+	filesCh := make(chan string)
+	produce := func() error {
+		defer close(filesCh)
+		files, err := b.GetFiles()
+		logrus.Debugf("files: %+v", files)
+		if err != nil {
+			return err
+		}
+		for _, f := range files {
+			filesCh <- f
+		}
+		return nil
+	}
+	errsChan := make(chan error)
+	map_ := func() error {
+		for f := range filesCh {
+			if err := BumpObject(b, f, o.OutDir); err != nil {
+				errsChan <- err
+			}
+		}
+		return nil
+	}
+	return util.ProduceMap(0, produce, map_, errsChan)
+}
+
+func BumpObject[T any](b Bumper[T], file, outDir string) error {
+	logrus.Infof("bumping config %s", file)
+
+	srcFileFullPath := file
+
+	obj, err := b.Unmarshall(srcFileFullPath)
+	if err != nil {
+		logrus.WithError(err).Errorf("failed to unmarshall file %s", srcFileFullPath)
+		return fmt.Errorf("unmarshall file %s: %w", file, err)
+	}
+
+	filename := path.Base(file)
+
+	logrus.Infof("bumping filename %s", filename)
+	bumpedFilename, err := b.BumpFilename(filename, obj)
+	if err != nil {
+		logrus.WithError(err).Errorf("error bumping file %s", bumpedFilename)
+		return fmt.Errorf("bump filename: %w", err)
+	}
+	logrus.Infof("bumped filename %s", bumpedFilename)
+
+	outDir = getOutDir(file, outDir)
+	logrus.Debugf("out dir: %s", outDir)
+	dstFileFullPath := path.Join(outDir, bumpedFilename)
+
+	if _, err := os.Stat(dstFileFullPath); err == nil {
+		logrus.WithError(err).Warnf("file %s already exists, skipping", dstFileFullPath)
+		return fmt.Errorf("file %s already exists", dstFileFullPath)
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("file exists: %w", err)
+	}
+
+	logrus.Infof("bumping obj")
+	bumpedObj, err := b.BumpContent(obj)
+	if err != nil {
+		logrus.WithError(err).Error("error bumping obj")
+		return fmt.Errorf("bump object: %w", err)
+	}
+
+	logrus.Infof("marshalling obj %s to %s", bumpedFilename, outDir)
+	if err := b.Marshall(bumpedObj, bumpedFilename, outDir); err != nil {
+		logrus.WithError(err).Error("error marshalling2 obj")
+		return fmt.Errorf("marshall obj: %w", err)
+	}
+
+	return nil
+}
+
+func getOutDir(file string, dir string) string {
+	if dir != "" {
+		return dir
+	}
+	return path.Dir(file)
+}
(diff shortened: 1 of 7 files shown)


droslean (commented):
Nice, but I don't think that the folder structure should be designed like that. Just put each tool into his own folder in `ci-tools/cmd`. Also, none of the tools should execute something from an external `Makefile` since all logic already exists programmatically.

cmd/branchingconfigmanagers/release-jobs-config-manager/internal/bumper/bumper.go
```diff
@@ -0,0 +1,127 @@
+package bumper
```
droslean, line 1: There is no need for the whole package. Please check: https://github.com/openshift/ci-tools/blob/master/pkg/config/load.go#L242

cmd/branchingconfigmanagers/release-jobs-config-manager/internal/bumper/default-bumper.go
```diff
@@ -0,0 +1,91 @@
+package bumper
```
droslean, line 1: You can remove this as well.

The tool should construct a ci-operator config as an object, make the changes in the struct and then bump using https://github.com/openshift/ci-tools/blob/master/pkg/config/load.go#L242

cmd/branchingconfigmanagers/release-jobs-config-manager/cmd/rpm-deps-mirroring-services/main.go
```diff
@@ -0,0 +1,169 @@
...
+	if err = o.adjustPaths(); err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
+	}
+
+	logrus.SetLevel(logrus.Level(o.logLevel))
+	logrus.Debugf("using options %+v", o)
+
+	recStatus, err := reconcile(time.Now(), o)
+	switch recStatus {
```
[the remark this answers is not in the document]
danilo-gemoli: So do we only go on iff everything went ok (exitCode == ReconcileOK)?

cmd/branchingconfigmanagers/release-jobs-config-manager/internal/bumper/default-bumper.go
danilo-gemoli: Let's consider [this](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/release/openshift-release-master__nightly-4.11.yaml) config to bump from version 4.11 to 4.12; if I have understood it correctly, I should have read this file using something like [this](https://github.com/openshift/ci-tools/blob/master/pkg/config/load.go#L271) (or similar) and then making the changes on the object. Since the version "4.11" is placed into a lot of different fields, how can I know all of them beforehand? What if people start putting version into a field we have not taken into the account? At that point I could use the reflection to overcome that problem, but what would then be the benefits of doing so instead of my actual approach? Furthermore, the function you are mentioning seems to 'only' write the file with doing the actual version bumping, am I missing something?

cmd/branchingconfigmanagers/release-jobs-config-manager/cmd/rpm-deps-mirroring-services/main.go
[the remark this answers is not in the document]
droslean: If you want the program to exit on that specific error, you can use `logrus.WIthError(err).Fatal("....")`. Fatal will terminate the code with a non-zero return code.

cmd/branchingconfigmanagers/release-jobs-config-manager/internal/bumper/default-bumper.go
droslean: Yes. You can load the config as a struct, change the required versions and then write the config to a file. The fields that hold the version should be specific and we can make a list of them.

danilo-gemoli: After a brief discussion with @droslean I agreed to follow his approach. The main motivation behind the decision is that config file could be represented in other formats than just plain text files. So the assumption to read them line by line could not work in every scenario.

jmguzik:
@danilo-gemoli what about this one?

pkg/branchcuts/bumper/bumper.go
```diff
@@ -0,0 +1,110 @@
+package bumper
```
jmguzik, line 1: Are these bumper files used in the second PR?

jmguzik, line 1: This one will be skipped (and I think it was skipped) by your bumper logic: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/release/openshift-release-master__nightly-4.13.yaml#L560

jmguzik (changes requested):
Some remarks.

pkg/branchcuts/bumper/bumper.go
danilo-gemoli: Yes, that's why I have created this abstraction. It will become more obvious when you will be reviewing the second and third PRs.

danilo-gemoli: Yes, you are correct. I know those tools could potentially miss some other sections as well, especially when it comes to handle non-structured text. As long as the field that it is going to be modified is well defined and it has a structure it is quite easy to perform the replacement. In all other cases I simply do not have strategies other than a naive text/regex replacement and that is dangerous because you could end up modifying text that it is not supposed to be modified.

Example:
Suppose you want to replace `release-4.13` with `release-4.14` using the `sed` expression `s/4\.13/4\.14/g` inside the following script:
```sh
curl -L https://github.com/openshift-cnv/cnv-ci/tarball/release-4.13 -o /tmp/cnv-ci.tgz
# Do some work here
# ...
# I'm going to do this and that because of a bug since version 4.13
# ...
```
That would be transformed into:
```diff
- curl -L https://github.com/openshift-cnv/cnv-ci/tarball/release-4.13 -o /tmp/cnv-ci.tgz
+ curl -L https://github.com/openshift-cnv/cnv-ci/tarball/release-4.14 -o /tmp/cnv-ci.tgz 
# Do some work here
# ...
- # I'm going to do this and that because of a bug since version 4.13
+ # I'm going to do this and that because of a bug since version 4.14
# ...
```
While the first replacement is probably good, the second one for sure it is not as we have changed the semantic of the original text.
Of course we could decide to set some rules as "skip comments if it is a bash script" but I have realized you won't never be 100% sure you made the correct decision if we go down this way.
I bet this is a problem that even @petr-muller faced again and again.
I am obviously open to any advises.

petr-muller: TBH I never worried too much about substitutions like this (which likely resulted in some wrong results but we were always on that fine line between process efficiency and correctness). I think the only place where we did `sed` substitution was bumping relesae payload jobs, and that was a basically a black box incantation that nobody eventually understood after a few bumps :)

That said, I think a reasonable approach is to not have ambition to bump versions in unstructured texts. You could have a contract that some part of the CI system exposes the "bumped" values in e.g. an envvar, and if people's payloads want to be handled by the tooling, they should use the nevvar. If they don't, they won't be bumped.

danilo-gemoli: and I agree. You should either follow the "contract" we established beforehand or those files (or at least some of their stanzas/sections/fields/whatever) won't be bumped to the next version.
That's it, unless we don't want to start training a neural network... that won't be 100% accurate either! :)

jmguzik: > `+ # I'm going to do this and that because of a bug since version 4.14`

Most of these problems will around curl as I noticed in the past so I would definitely welcome some regexp that is taking care at least of this case.

> which likely resulted in some wrong results but we were always on that fine line between process efficiency and correctness
Generally, I would say yes and agree, but on the other hand, as we are not that experienced still, we would like to avoid possible situations when something is not working because we put the accent on efficiency instead of correctness. I also feel now more eyes are watching the process and there is a bigger pressure to do it right.

@danilo-gemoli with that being said. I am ok to merge it as an initial version that could be corrected in the future.

jmguzik (approved):

PR #2817, opened 2022-05-13
Ensure importing release on promoting to official images

hongkailiu:
https://issues.redhat.com/browse/DPTP-2857

/cc @openshift/test-platform

### DIFF
--- a/pkg/validation/config.go
+++ b/pkg/validation/config.go
@@ -152,7 +152,14 @@ func (v *Validator) validateConfiguration(ctx *configContext, config *api.Releas
 
 	// Validate promotion
 	if config.PromotionConfiguration != nil {
-		validationErrors = append(validationErrors, validatePromotionConfiguration("promotion", *config.PromotionConfiguration)...)
+		validationErrors = append(validationErrors,
+			validatePromotionConfiguration("promotion",
+				*config.PromotionConfiguration,
+				api.PromotesOfficialImages(config, api.WithOKD),
+				len(api.ImageTargets(config)) > 0,
+				api.ExtractPromotionNamespace(config),
+				config.ReleaseTagConfiguration,
+				config.Releases)...)
 	}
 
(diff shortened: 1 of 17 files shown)


pkg/validation/config.go
```diff
@@ -402,6 +407,21 @@ func validatePromotionConfiguration(fieldRoot string, input api.PromotionConfigu
...
 		validationErrors = append(validationErrors, fmt.Errorf("%s: both name and tag defined", fieldRoot))
 	}
+
+	if promotesOfficialImages {
+		var latestRelease bool
+		if releaseTagConfiguration != nil {
+			latestRelease = true
+		}
+		if releases != nil {
+			if _, ok := releases["latest"]; ok {
```
bbguimaraes, line 417: Fun fact: map indexing has an implicit `nil` check in Go.  This can all be just:

```go
if _, ok := releases["latest"]; !ok && releaseTagConfiguration == nil {
```

pkg/validation/config.go
```diff
@@ -402,6 +407,21 @@ func validatePromotionConfiguration(fieldRoot string, input api.PromotionConfigu
...
+				latestRelease = true
+			}
+		}
+		if !latestRelease {
+			validationErrors = append(validationErrors, fmt.Errorf("promotion to the integration image stream or namespace requires defining tag_specification or releases[\"latest\"]"))
```
bbguimaraes, line 422: This is also too implementation-specific for a user message.

Importing a release is just a proxy, the validation error should describe the _purpose_ of the verification and what we expect from the configuration, not the mechanics of what was checked.

bbguimaraes (commented):
Those repositories in the failed tests have `promotion` but not `images`, I suppose for automatic branching?

pkg/validation/config.go
hongkailiu: At least my purpose is to ask our users to add the `tag_specification` or `releases`.
Are you suggesting our purpose is to add an e2e tests instead?

PR #2819, opened 2022-05-16
Allow add tests from openshift-tests & openshift-tests-private to testgrid

liangxia:
QE would like to add several tests from openshift-tests and openshift-tests-private (both are under openshift/) to testgrid


jhou1:
Thank you
/lgtm

PR #2822, opened 2022-05-17
Flake aggregated disruption failures.

dgoodwin:
These tests have been live for about a week and are already failing
payloads for what looks like fairly routine changes in disruption on
Azure. Move to flakes for now so we can continue watching how often
these will trigger and iterate, without blocking payloads.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
@@ -172,13 +172,20 @@ func (o *JobRunAggregatorAnalyzerOptions) Run(ctx context.Context) error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


dgoodwin:
Based on Sippy data it looks like these do occasionally fail on AWS or GCP maybe 1-2 times a week, whereas Azure is 10x that.

deads2k:
Can we just skip evaluating this on azure?  I think that will be more effective than flaking on everything.  Also, you have not changed the process exit code, so i think we'll still fail.

The PR that introduced this may be able to simply search for "azure" in the job name.  ugly, but cheap.

deads2k:
/lgtm
/hold

Compare to https://github.com/openshift/ci-tools/pull/2823

I'm ok with either, but I think you might still want the data itself computed, just not able to fail tests.

PR #2826, opened 2022-05-19
Tide config shard manager for automatic branch cuts: GA event

jmguzik:

### DIFF
--- a/cmd/branchingconfigmanagers/tide-config-manager/main.go
+++ b/cmd/branchingconfigmanagers/tide-config-manager/main.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"path"
 	"strings"
@@ -11,6 +12,7 @@ import (
 
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/afero"
+	"gopkg.in/yaml.v2"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -25,8 +27,9 @@ type options struct {
 	prowConfigDir            string
 	shardedProwConfigBaseDir string
 	lifecycleConfigFile      string
-	overwriteTimeRaw         string
-	overwriteTime            *time.Time
+	excludedReposFile        string
+	overrideTimeRaw          string
+	overrideTime             *time.Time
 }
 
 func gatherOptions() (*options, error) {
@@ -37,7 +40,8 @@ func gatherOptions() (*options, error) {
 	fs.StringVar(&o.prowConfigDir, "prow-config-dir", "", "Path to the Prow configuration directory.")
 	fs.StringVar(&o.shardedProwConfigBaseDir, "sharded-prow-config-base-dir", "", "Basedir for the sharded prow config. If set, org and repo-specific config will get removed from the main prow config and written out in an org/repo tree below the base dir.")
 	fs.StringVar(&o.lifecycleConfigFile, "lifecycle-config", "", "Path to the lifecycle config file")
-	fs.StringVar(&o.overwriteTimeRaw, "overwrite-time", "", "Act as if this was the current time, must be in RFC3339 format")
+	fs.StringVar(&o.excludedReposFile, "excluded-repos-config", "", "Path to the GA's excluded repos config file.")
+	fs.StringVar(&o.overrideTimeRaw, "override-time", "", "Act as if this was the current time, must be in RFC3339 format")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("could not parse input")
 	}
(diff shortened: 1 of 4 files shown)


cmd/branchingconfigmanagers/tide-config-manager/main.go
```diff
@@ -237,6 +237,136 @@ func (cfe codeFreezeEvent) GetDataFromProwConfig(pc *prowconfig.ProwConfig) {
...
+	branches := sets.NewString(q.IncludedBranches...)
+	if reqLabels.Has(cherryPickApproved) {
+		if branches.Has(gae.releasePast) {
+			branches.Insert(gae.releaseCurrent)
+		}
+		if branches.Has(gae.openshiftPast) {
+			branches.Insert(gae.openshiftCurrent)
+		}
+		if branches.Intersection(gae.openshiftReleaseBranches).Len() == 0 && !gae.noXYAllowList.Has(q.Repos[0]) {
+			fmt.Printf("Suspicious cherry-pick-approved query (without %s): %s\n", gae.current, q.Repos)
```
jmguzik, line 322: I will create a follow-up card to get rid of these prints that I have here, but they are complicated cases that need further analysis. Usually, it means the config is broken, or something is missing. This requires closer investigation.

cmd/branchingconfigmanagers/tide-config-manager/main_test.go
```diff
@@ -41,25 +73,15 @@ func TestReconcile(t *testing.T) {
...
-  - includedBranches:
-    - master
-    labels:
-    - docs-approved
-    - px-approved
-    - qe-approved
-    repos:
-    - openshift/dummy
-`},
+				path: prepareProwConfig(repos, []string{docsApproved, pxApproved, qeApproved}, []string{masterBranch})},
```
jmguzik, line 84: @droslean could not use here testhelper as the common function in pkg is not ready for it, and it would need further refactoring. Instead, I created a simple mechanism, not to paste configs every time. I think that it's enough, let me know if ok for you.

cmd/branchingconfigmanagers/tide-config-manager/main.go
```diff
@@ -237,6 +237,136 @@ func (cfe codeFreezeEvent) GetDataFromProwConfig(pc *prowconfig.ProwConfig) {
...
+	openshiftCurrent              string
+	releaseFuture                 string
+	openshiftFuture               string
+	past                          string
+	current                       string
+	future                        string
+}
+
+func newGeneralAvailabilityEvent(past, current, future string) generalAvailabilityEvent {
+	noXYAllowList := sets.NewString(
```
droslean, line 258: That list should be controllable with flags or something outside the code.

cmd/branchingconfigmanagers/tide-config-manager/main_test.go
droslean: Why the testhelper pkg is not ready for it? What it would take to make it ready?

cmd/branchingconfigmanagers/tide-config-manager/main.go
jmguzik: Yes, but the ticket that is going to figure out is in the backlog. It's still a prototype.

cmd/branchingconfigmanagers/tide-config-manager/main_test.go
jmguzik: Bad working myabe. Look at `shardprowconfig.ShardProwConfig`. It's not exposing the data in any way. I could refactor it, but it is code-shared with other tool which does not take the benefit of `testhelper`. So, it would be refactoring of 2 tools just to make it available in tests.

cmd/branchingconfigmanagers/tide-config-manager/main.go
jmguzik: I guess local config could be created for now, what do you think?

droslean: A local config should be ok, and we can have it in release repo as well, so we can control its data.

cmd/branchingconfigmanagers/tide-config-manager/main_test.go
droslean: I think it is better to refactor the reconcile function. The fact that it is taking a target as an argument smells. In the tests, you could define the config and then after running reconcile you can check `testhelper.CompareWithFixture(t, tt.args.config)`. The expected config should exist in the `testdata` folder for better readability. Your code shouldn't walk in any directory, instead you can just load the prow config and loop through the shared configuration and reconcile them. Note that the code should also reconcile the non-sharded config as well.

jmguzik: So to make what you want, I should refactor `shardprowconfig.ShardProwConfig` and determinize-prow-config tool and my new tool. I think it's slightly beyond the objective for this task.

jmguzik: Nope, I inject additional operations to the `shardprowconfig.ShardProwConfig` so it has everything with the tool.

jmguzik: Just to capture the explanation why `testhelper.CompareWithFixture` is not used in this particular place:
https://github.com/openshift/ci-tools/blob/88cd94e25cd685b4da4bd19dbb44aa7a8bd74010/pkg/api/shardprowconfig/shardprowconfig.go#L119
In the above place `shardprowconfig.ShardProwConfig` function cleans tide queries from the config which are needed to make the comparison.

Queries saving to the disk is done in the same function:
https://github.com/openshift/ci-tools/blob/88cd94e25cd685b4da4bd19dbb44aa7a8bd74010/pkg/api/shardprowconfig/shardprowconfig.go#L122
so changing that would mean refactor not only here, but also in the other tool which is using this function.

It was agreed that the above work would be extracted as a side-job to a separate card as to be done in the future as code cleaning/improvement.

cmd/branchingconfigmanagers/tide-config-manager/main.go
jmguzik: Done, extracted to a yaml config file. I am writing docs/howto file for that tool, but it will come in a separate PR (with an explanation of format etc). Generally, the config will look like this:
```
NoXYAllowList:
  - openshift/cluster-logging-operator
.....
ExcludedAllowList:
  - openshift/assisted-installer
....
```

jmguzik:
@droslean ping.
I think the only thing left from your comments is an issue regarding unit tests, but I would create a Jira card for the future for that once we merge this work.

droslean (approved):

PR #2827, opened 2022-05-19
Log pass/fail disruption info for debugging.

dgoodwin:


xueqzhan:
Do you need this for the standard deviation checks too (checkDisruptionMean)?

dgoodwin:
Probably a good idea, I'm going to move this up in the stack so we log it once for each job run, for each backend. The current messages are pretty clear what the thresholds are I just can't easily see what each job reported. Will comment again once I get this updated, thanks!

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -339,8 +339,12 @@ func (a *weeklyAverageFromTenDays) checkPercentileDisruption(jobRunIDToAvailabil
 
 	for jobRunID, disruption := range jobRunIDToAvailabilityResultForBackend {
 		if float64(disruption.SecondsUnavailable) > threshold {
+			fmt.Printf("fail: jobrun=%s backend=%s observed=%ds threshold=%.2fs thresholdPercentile=%d\n", jobRunID, disruption.ServerName,
```
deads2k, line 342: you want to add these to the summary as well?

dgoodwin: Yeah scratch moving it up, It'll be nice to see it right in the test message instead of the build log. Working on it now.

dgoodwin:
Updated to include the jobID=Xs for each success/fail in most all messages from both mean and percentile testing.

PR #2830, opened 2022-05-19
Fix DPTP-2823 Add .snyk file

bear-redhat:
[DPTP-2823](https://issues.redhat.com//browse/DPTP-2823) We are adding this file to exclude `vendor/` and tests.


bear-redhat:
JIRA ticket: DPTP-2823

PR #2837, opened 2022-05-24
pkg/prowgen: handle missing decoration

bbguimaraes:


pkg/prowgen/jobbase.go
```diff
@@ -106,7 +106,11 @@ func NewProwJobBaseBuilderForTest(configSpec *cioperatorapi.ReleaseBuildConfigur
 
 	maxCustomDuration := time.Hour * 8
 	if test.Timeout != nil && test.Timeout.Duration <= maxCustomDuration {
-		p.base.UtilityConfig.DecorationConfig.Timeout = test.Timeout
+		u := &p.base.UtilityConfig
+		if u.DecorationConfig == nil {
```
bbguimaraes, line 110: This happens when none of these apply: https://github.com/openshift/ci-tools/blob/91eae38acfb9f7a80bb77c5c36b04c2b602ebdbf/pkg/prowgen/jobbase.go#L64-L68.

PR #2841, opened 2022-05-27
github-ldap-user-group-creator: whitelist github IDs for robots

hongkailiu:
Found those (see `githubRobotIds`) in the logs.

/hold

Give DPP more time to confirm if https://github.com/RH-Cachito is a robot
https://coreos.slack.com/archives/CBUT43E94/p1653679694670849

1. Ignore those known robot accounts while logging
2. Error instead of warning if seeing more in the future

/cc @smg247


hongkailiu:
/hold cancel

"RH-Cachito is a bot account", confirmed by @jupierce

PR #2854, opened 2022-06-07
ci-operator: no more dirty versions

hongkailiu:
https://issues.redhat.com/browse/DPTP-2777

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -125,9 +125,7 @@ cmd/vault-secret-collection-manager/index.js: cmd/vault-secret-collection-manage
 # Example:
 #   make production-install
 production-install: cmd/vault-secret-collection-manager/index.js cmd/pod-scaler/frontend/dist cmd/repo-init/frontend/dist
-	rm -f cmd/pod-scaler/frontend/dist/dummy # we keep this file in git to keep the thing compiling without static assets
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -125,9 +125,7 @@ cmd/vault-secret-collection-manager/index.js: cmd/vault-secret-collection-manage
 # Example:
 #   make production-install
 production-install: cmd/vault-secret-collection-manager/index.js cmd/pod-scaler/frontend/dist cmd/repo-init/frontend/dist
-	rm -f cmd/pod-scaler/frontend/dist/dummy # we keep this file in git to keep the thing compiling without static assets
-	rm -f cmd/repo-init/frontend/dist/dummy
-	hack/install.sh
+	hack/install.sh no-race remove-dummy
```
smg247, line 128: why do we need to specify `no-race` now?

hongkailiu: It did not change the behaviour.
https://github.com/openshift/ci-tools/blob/90ba25a3545f306425f18e0871c60924f0a30067/hack/install.sh#L13-L19

Any string except `race` will do the job.
We need add it as the first arg now because we need to provide the 2nd arg.

PR #2855, opened 2022-06-08
FIX: prevent log to interpret bash meta-chars

danilo-gemoli:
A better version of [#2851](https://github.com/openshift/ci-tools/pull/2851)

/cc @bbguimaraes

### DIFF
--- a/hack/lib/log/output.sh
+++ b/hack/lib/log/output.sh
@@ -97,8 +97,8 @@ function os::log::internal::prefix_lines() {
 
 	local old_ifs="${IFS}"
 	IFS=$'\n'
-	for line in "${content}"; do
+	while read -r line; do
 		echo "${prefix} ${line}"
-	done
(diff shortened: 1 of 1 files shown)


bbguimaraes:
What problem are these pull requests trying to solve?

If a test needs to output special characters, it should escape them.  That is the API of these functions, which already have countless users — do we know that none of them relies on the expansion?

In addition, the source for this library is in `openshift/origin`: https://github.com/openshift/origin/blob/master/hack/lib/log/output.sh.  I'd really rather not maintain a fork of it.

bbguimaraes:
It's not entirely clear to me whether it's intentional or an artifact of the `bash` implementation.

However, I don't want us to carry a patch in `ci-tools`, and I don't think we can realistically prove that there isn't a client somewhere which wants the expansion in order to take this upstream to `openshift/origin`.

PR #2856, opened 2022-06-08
Fix some snyk code issues

hongkailiu:
https://issues.redhat.com/browse/DPTP-2899

The PR fixed all [high] and some [medium] issues.

The left ones are all

```
[Medium] Clear Text Logging
```

I have no idea how to fix them.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -664,8 +664,8 @@ func (o *options) Complete() error {
 		}
 	}
 	if o.pushSecretPath != "" {
-		if o.pushSecret, err = getDockerConfigSecret(api.RegistryPushCredentialsCICentralSecret, o.pushSecretPath); err != nil {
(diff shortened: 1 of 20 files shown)


bbguimaraes:
Although there are some actual problems being corrected, a lot of this is quite silly (the "vulnerabilities", not the work).  MD5 is perfectly fine for what we are doing, Github has a strict policy for user names ("[u]sername may only contain alphanumeric characters or single hyphens, and cannot begin or end with a hyphen"), etc.

The first commit is a bit messy, can you take the renames out of it?

Can we even make the `prpq_reconciller` change while maintaining compatibility with the existing data?

/hold

hongkailiu:
> a lot of this is quite silly (the "vulnerabilities", not the work)

Thanks for the review.
I spent too much time on those silly things.

Let us hold it until we are sure about using synk at all.
If we still want it, then I will address your comments.

jmguzik:
@hongkailiu I think it is resolved elsewhere and you can close

hongkailiu:
Seem that we do not care about it any more.

PR #2857, opened 2022-06-08
`check-gh-automation`: don't check repos where the config was simply modified

smg247:
We are running this check a lot, and since we now have a periodic to consistently verify that repos that are already a part of the CI are accessible by our automation we don't need the presubmit to check repos for which the configs have merely been modified.

I used https://git-scm.com/docs/git-diff-tree#Documentation/git-diff-tree.txt---diff-filterACDMRTUXB82308203 to only flag repos for which the configs were: added, moved, or renamed.

### DIFF
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -171,7 +171,7 @@ func GetAllConfigsFromSHA(releaseRepoPath, sha string, logger *logrus.Entry) (*R
 }
 
 func GetChangedTemplates(path, baseRev string) ([]string, error) {
-	changes, err := getRevChanges(path, TemplatesPath, baseRev)
+	changes, err := getRevChanges(path, TemplatesPath, baseRev, false)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


pkg/config/release.go
```diff
@@ -227,20 +227,25 @@ func GetChangedRegistrySteps(path, baseRev string, graph registry.NodeByName) ([
...
 // `root`.  Paths are relative to `root`.
-func getRevChanges(root, path, base string) ([]string, error) {
+// If 'ignoreModified' is true it will only check for relevant added, moved, or copied files
+func getRevChanges(root, path, base string, ignoreModified bool) ([]string, error) {
 	// Sample output (with abbreviated hashes) from git-diff-tree(1):
 	// :100644 100644 bcd1234 0123456 M file0
-	cmd := []string{"diff-tree", "-r", "--diff-filter=d", base + ":" + path, "HEAD:" + path}
+	filter := "--diff-filter=d"
+	if ignoreModified {
+		filter = "--diff-filter=ACR"
```
hongkailiu, line 246: Why removed `d`?

smg247: the lowercase `d` means everything **but** deleted. using uppercase letters means we **only** want the options supplied. In this case: added, moved, or copied.

PR #2866, opened 2022-06-13
Before branching day: step 2

danilo-gemoli:
This is a tool that tries to automate the second step of phase "Before branching day" .
Check [Centralized Release Branching and Config Management V3](https://docs.google.com/document/d/1Z6ejnDCOCvNv9PWkyNPzVbjuLbDMAAT5GEeDpzb0SMs/edit#heading=h.r9xn02r1cyfn)

/cc @jmguzik @droslean

### DIFF
--- a/cmd/branchingconfigmanagers/generated-release-gating-jobs/main.go
+++ b/cmd/branchingconfigmanagers/generated-release-gating-jobs/main.go
@@ -1,0 +1,84 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/openshift/ci-tools/pkg/api/ocplifecycle"
+	"github.com/openshift/ci-tools/pkg/branchcuts/bumper"
+	cioperatorcfg "github.com/openshift/ci-tools/pkg/config"
+)
+
+const (
+	releaseJobsPath = "ci-operator/config/openshift/release"
+)
+
+type options struct {
+	curOCPVersion    string
+	releaseRepoDir   string
+	logLevel         int
+	newIntervalValue int
+}
+
+func gatherOptions() (*options, error) {
+	var errs []error
+	o := &options{}
+	flag.StringVar(&o.curOCPVersion, "current-release", "", "Current OCP version")
+	flag.StringVar(&o.releaseRepoDir, "release-repo", "", "Path to 'openshift/release/ folder")
+	flag.IntVar(&o.newIntervalValue, "interval", 168, "New interval to set")
+	flag.IntVar(&o.logLevel, "log-level", int(logrus.DebugLevel), "Log level")
+	flag.Parse()
+
+	if _, err := ocplifecycle.ParseMajorMinor(o.curOCPVersion); o.curOCPVersion != "" && err != nil {
+		errs = append(errs, fmt.Errorf("error parsing current-release %s", o.curOCPVersion))
+	}
+
+	if o.newIntervalValue < 0 {
+		errs = append(errs, errors.New("error parsing interval: value is not a positive integer"))
+	}
+
+	if o.releaseRepoDir != "" {
+		if !path.IsAbs(o.releaseRepoDir) {
+			errs = append(errs, errors.New("error parsing release repo path: path has to be absolute"))
+		}
+	} else {
+		errs = append(errs, errors.New("error parsing release repo path: path is mandatory"))
+	}
+
+	return o, utilerrors.NewAggregate(errs)
+}
+
+func main() {
+	o, err := gatherOptions()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to gather options")
+	}
+
+	logrus.SetLevel(logrus.Level(o.logLevel))
+	logrus.Debugf("using options %+v", o)
+
+	if err := reconcile(time.Now(), o); err != nil {
+		logrus.WithError(err).Fatal("failed to reconcile the status")
+	}
+	logrus.Info("status reconciled")
+}
+
+func reconcile(now time.Time, o *options) error {
+	logrus.Debugf("using options %+v", o)
+	releaseJobsDir := path.Join(o.releaseRepoDir, releaseJobsPath)
+	b, err := bumper.NewGeneratedReleaseGatingJobsBumper(o.curOCPVersion, releaseJobsDir, o.newIntervalValue)
+	if err != nil {
+		return fmt.Errorf("new bumper: %w", err)
+	}
+	if err := bumper.Bump[*cioperatorcfg.DataWithInfo](b, &bumper.BumpingOptions{}); err != nil {
+		return fmt.Errorf("bumper: %w", err)
+	}
+	return nil
+}
--- a/pkg/branchcuts/bumper/gen-release-jobs-bumper.go
+++ b/pkg/branchcuts/bumper/gen-release-jobs-bumper.go
@@ -1,0 +1,241 @@
+package bumper
+
+import (
+	"fmt"
+	"io/fs"
+	"path"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/sirupsen/logrus"
+
+	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
+	"github.com/openshift/ci-tools/pkg/api/ocplifecycle"
+	cioperatorcfg "github.com/openshift/ci-tools/pkg/config"
+)
+
+const (
+	releaseJobsRegexPatternFormat = `openshift-release-master__(okd|ci|nightly|okd-scos)-%s.*\.yaml`
+	ocpReleaseEnvVarName          = "OCP_VERSION"
+)
+
+type GeneratedReleaseGatingJobsBumper struct {
+	mm               *ocplifecycle.MajorMinor
+	getFilesRegexp   *regexp.Regexp
+	jobsDir          string
+	newIntervalValue int
+}
+
+var _ Bumper[*cioperatorcfg.DataWithInfo] = &GeneratedReleaseGatingJobsBumper{}
+
+func NewGeneratedReleaseGatingJobsBumper(ocpVer, jobsDir string, newIntervalValue int) (*GeneratedReleaseGatingJobsBumper, error) {
+	mm, err := ocplifecycle.ParseMajorMinor(ocpVer)
+	if err != nil {
+		return nil, fmt.Errorf("parse release: %w", err)
+	}
+	mmRegexp := fmt.Sprintf("%d\\.%d", mm.Major, mm.Minor)
+	getFilesRegexp := regexp.MustCompile(fmt.Sprintf(releaseJobsRegexPatternFormat, mmRegexp))
+	return &GeneratedReleaseGatingJobsBumper{
+		mm,
+		getFilesRegexp,
+		jobsDir,
+		newIntervalValue,
+	}, nil
+}
+
+func (b *GeneratedReleaseGatingJobsBumper) GetFiles() ([]string, error) {
+	files := make([]string, 0)
+	err := filepath.Walk(b.jobsDir, func(path string, info fs.FileInfo, err error) error {
+		if b.getFilesRegexp.Match([]byte(path)) {
+			files = append(files, path)
+		}
+		return nil
+	})
+	return files, err
+}
+
+func (b *GeneratedReleaseGatingJobsBumper) Unmarshall(file string) (*cioperatorcfg.DataWithInfo, error) {
+	cfgDataByFilename, err := cioperatorcfg.LoadDataByFilename(file)
+	if err != nil {
+		return nil, err
+	}
+
+	filename := path.Base(file)
+	dataWithInfo, ok := cfgDataByFilename[filename]
+	if !ok {
+		logrus.WithError(err).Errorf("failed to get config %s", file)
+		return nil, fmt.Errorf("can't get data %s", filename)
+	}
+	return &dataWithInfo, nil
+}
+
+func (b *GeneratedReleaseGatingJobsBumper) BumpFilename(
+	filename string,
+	dataWithInfo *cioperatorcfg.DataWithInfo) (string, error) {
+	newVariant, err := ReplaceWithNextVersion(dataWithInfo.Info.Metadata.Variant, b.mm.Major)
+	if err != nil {
+		return "", err
+	}
+	dataWithInfo.Info.Metadata.Variant = newVariant
+	return dataWithInfo.Info.Metadata.Basename(), nil
+}
+
+/*
+	Candidate bumping fields:
+	.base_images.*.name
+	.releases.*.{release,candidate}.version
+	.releases.*.prerelease.version_bounds.{lower,upper}
+	.tests[].steps.test[].env[].default
+*/
+func (b *GeneratedReleaseGatingJobsBumper) BumpContent(dataWithInfo *cioperatorcfg.DataWithInfo) (*cioperatorcfg.DataWithInfo, error) {
+	major := b.mm.Major
+	config := &dataWithInfo.Configuration
+	if err := bumpBaseImages(config, major); err != nil {
+		return nil, err
+	}
+
+	if err := bumpReleases(config, major); err != nil {
+		return nil, err
+	}
+
+	if err := bumpTests(config, major); err != nil {
+		return nil, err
+	}
+
+	if err := ReplaceWithNextVersionInPlace(&config.Metadata.Variant, major); err != nil {
+		return nil, err
+	}
+
+	if config.Tests != nil {
+		for i := 0; i < len(config.Tests); i++ {
+			if config.Tests[i].Interval != nil {
+				*config.Tests[i].Interval = strconv.Itoa(b.newIntervalValue) + "h"
+			}
+		}
+	}
+
+	return dataWithInfo, nil
+}
+
+func bumpBaseImages(config *cioperatorapi.ReleaseBuildConfiguration, major int) error {
+	if config.BaseImages == nil {
+		return nil
+	}
+
+	bumpedImages := make(map[string]cioperatorapi.ImageStreamTagReference)
+	for k := range config.BaseImages {
+		image := config.BaseImages[k]
+		if err := ReplaceWithNextVersionInPlace(&image.Name, major); err != nil {
+			return err
+		}
+		bumpedImages[k] = image
+	}
+	config.BaseImages = bumpedImages
+	return nil
+}
+
+func bumpReleases(config *cioperatorapi.ReleaseBuildConfiguration, major int) error {
+	if config.Releases == nil {
+		return nil
+	}
+
+	bumpedReleases := make(map[string]cioperatorapi.UnresolvedRelease)
+	for k := range config.Releases {
+		release := config.Releases[k]
+		if release.Release != nil {
+			if err := ReplaceWithNextVersionInPlace(&release.Release.Version, major); err != nil {
+				return err
+			}
+		}
+		if release.Candidate != nil {
+			if err := ReplaceWithNextVersionInPlace(&release.Candidate.Version, major); err != nil {
+				return err
+			}
+		}
+		if release.Prerelease != nil {
+			if err := ReplaceWithNextVersionInPlace(&release.Prerelease.VersionBounds.Upper, major); err != nil {
+				return err
+			}
+			if err := ReplaceWithNextVersionInPlace(&release.Prerelease.VersionBounds.Lower, major); err != nil {
+				return err
+			}
+		}
+		bumpedReleases[k] = release
+	}
+	config.Releases = bumpedReleases
+	return nil
+}
+
+func bumpTests(config *cioperatorapi.ReleaseBuildConfiguration, major int) error {
+	if config.Tests == nil {
+		return nil
+	}
+
+	for i := 0; i < len(config.Tests); i++ {
+		test := config.Tests[i]
+
+		if test.MultiStageTestConfiguration == nil {
+			continue
+		}
+
+		if err := bumpTestSteps(test.MultiStageTestConfiguration.Pre, major); err != nil {
+			return err
+		}
+		if err := bumpTestSteps(test.MultiStageTestConfiguration.Test, major); err != nil {
+			return err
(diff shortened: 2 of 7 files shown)


cmd/branchingconfigmanagers/generated-release-gating-jobs/main.go
```diff
@@ -0,0 +1,158 @@
...
+	switch recStatus {
+	case mgr.ReconcileOk:
+		logrus.Info("status reconciled")
+	default:
+		logrus.WithError(err).Error("failed to reconcile the status")
+		os.Exit(reconcileErrorExitStatus)
+	}
+}
+
+func reconcile(now time.Time, o *options) (mgr.ReconcileStatus, error) {
```
droslean, line 128:
```suggestion
func reconcile(now time.Time, o *options) error {
```

cmd/branchingconfigmanagers/generated-release-gating-jobs/main.go
```diff
@@ -0,0 +1,158 @@
...
+	if !ocpLifecycle.IsBeforeCodeFreeze(howManyDaysBefore) {
+		return mgr.ReconcileSkipped, nil
+	}
+
+	currentVersion, futureVersion, err := ocpLifecycle.GetOCPVersions()
+	if err != nil {
+		return mgr.ReconcileErr, fmt.Errorf("OCP Lifecycle versions: %w", err)
+	}
+
+	o.tryOverrideOCPVersions(currentVersion.GetVersion(), futureVersion.GetVersion())
```
droslean, line 145: The `tryOverrideOCPVersions()` name sounds like we should check for any errors.

pkg/branchcuts/bumper/errors.go
```diff
@@ -0,0 +1,73 @@
+package bumper
```
droslean, line 1: There is no need for a logical use case for the whole `errors.go` file.

pkg/branchcuts/config-manager/manager.go
```diff
@@ -0,0 +1,9 @@
+package configmanager
+
+type ReconcileStatus uint8
```
droslean, line 3: We don't care about any status or exit code.

hack/lib/branchcuts/search_yaml_fields.py
```diff
@@ -0,0 +1,73 @@
+#!/usr/bin/python
```
droslean, line 1: Why do we need this script?

hack/lib/branchcuts/test-common.sh
```diff
@@ -0,0 +1,152 @@
+readonly src_test_files_relative_dir="data"
```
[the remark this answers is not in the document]
danilo-gemoli: This script contains common code for the integration tests. See `test/integration/br-cuts-*`

pkg/branchcuts/bumper/repo-bumper.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+	return file, nil
+}
+
+func (b *RepoBumper) Marshall(file *ini.File, bumpedFilename, dir string) error {
+	filePath := path.Join(dir, bumpedFilename)
+	return saveIniFile(filePath, file)
+}
+
+func saveIniFile(path string, f *ini.File) error {
```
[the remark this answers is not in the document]
danilo-gemoli: I know, I did it because the library has a bug and it appends two '\n' at the end of the file when saving. I have already fixed it with [#321](https://github.com/go-ini/ini/pull/321) but I am not sure if upgrading would be a good idea.

cmd/branchingconfigmanagers/generated-release-gating-jobs/main.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	return o, utilerrors.NewAggregate(errs)
+}
+
+func main() {
+	o, err := gatherOptions()
+	if err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
```
droslean, line 105:
```suggestion
		logrus.WithError(err).Fatal("failed to gather options")
```

You can remove the `os.Exit`

cmd/branchingconfigmanagers/generated-release-gating-jobs/main.go
```diff
@@ -0,0 +1,155 @@
...
+
+	if err = o.adjustPaths(); err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
+	}
+
+	logrus.SetLevel(logrus.Level(o.logLevel))
+	logrus.Debugf("using options %+v", o)
+
+	err = reconcile(time.Now(), o)
```
droslean, line 116: This whole code block can be written as:
```go
if err := reconcile(time.Now(), o); err != nil{
	logrus.WithError(err).Fatal("failed to reconcile the status")
}
logrus.Info("status reconciled")
```

cmd/branchingconfigmanagers/release-controller-config-manager/Makefile
```diff
@@ -0,0 +1,7 @@
+build:
```
droslean, line 1: You can remove this file.

cmd/branchingconfigmanagers/release-controller-config-manager/.gitignore
```diff
@@ -0,0 +1 @@
+release-controller-config-manager
```
droslean, line 1: You can remove this file

cmd/branchingconfigmanagers/release-controller-config-manager/main.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	return o, utilerrors.NewAggregate(errs)
+}
+
+func main() {
+	o, err := gatherOptions()
+	if err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
```
droslean, line 98:
```suggestion
		logrus.WithError(err).Fatal("failed to gather options")
```
You can remove the `os.Exit`

cmd/branchingconfigmanagers/release-controller-config-manager/main.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	if err = o.adjustPaths(); err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
```
droslean, line 103:
```suggestion
		logrus.WithError(err).Fatal("failed to adjust paths")
```
You can remove the `os.Exit`

cmd/branchingconfigmanagers/release-controller-config-manager/main.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	logrus.SetLevel(logrus.Level(o.logLevel))
+	logrus.Debugf("using options %+v", o)
+
+	err = reconcile(time.Now(), o)
```
droslean, line 109: This code block can be written as:
```go
if err := reconcile(time.Now(), o); err != nil{
	logrus.WithError(err).Fatal("failed to reconcile the status")
}
logrus.Info("status reconciled")
```

cmd/branchingconfigmanagers/rpm-deps-mirroring-services/.gitignore
```diff
@@ -0,0 +1 @@
+rpm-deps-mirroring-services
```
droslean, line 1: You can remove this file

cmd/branchingconfigmanagers/rpm-deps-mirroring-services/Makefile
```diff
@@ -0,0 +1,7 @@
+build:
```
droslean, line 1: You can remove this file.

cmd/branchingconfigmanagers/rpm-deps-mirroring-services/main.go
```diff
@@ -0,0 +1,157 @@
...
+
+	if err = o.adjustPaths(); err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
+	}
+
+	logrus.SetLevel(logrus.Level(o.logLevel))
+	logrus.Debugf("using options %+v", o)
+
+	err = reconcile(time.Now(), o)
```
droslean, line 108:
```suggestion
if err := reconcile(time.Now(), o); err != nil{
	logrus.WithError(err).Fatal("failed to reconcile the status")
}
logrus.Info("status reconciled")
```

cmd/branchingconfigmanagers/rpm-deps-mirroring-services/main.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	return o, utilerrors.NewAggregate(errs)
+}
+
+func main() {
+	o, err := gatherOptions()
+	if err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
```
droslean, line 97:
```suggestion
		logrus.WithError(err).Fatal("failed to gather options")
```

cmd/branchingconfigmanagers/rpm-deps-mirroring-services/main.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if err = o.adjustPaths(); err != nil {
+		logrus.WithError(err).Error("failed to gather options")
+		os.Exit(optsParsingErrorExitStatus)
```
droslean, line 102:
```suggestion
		logrus.WithError(err).Error("failed to adjust paths")
```

pkg/api/ocplifecycle/codefreeze.go
```diff
@@ -0,0 +1,71 @@
...
+
+	logrus.Debugf("parsing future OCP version %s", after.ProductVersion)
+	futureVersion, err := ParseMajorMinor(after.ProductVersion)
+	if err != nil {
+		return nil, nil, err
+	}
+	return currentVersion, futureVersion, nil
+}
+
+func isCloseToCodeFreeze(codeFreezeDate, now time.Time, howManyDaysBefore int) bool {
```
droslean, line 63: This function can call `time.Now()` itself instead of getting it as an argument, but that won't be testable. Either way, the argument shouldn't be named `now` since it is not representing the present time.

pkg/branchcuts/bumper/gen-release-jobs-bumper.go
```diff
@@ -0,0 +1,241 @@
...
+			if err := ReplaceWithNextVersionInPlace(env.Default, major); err != nil {
+				return err
+			}
+		}
+		multistageTest.Environment[i] = env
+	}
+	return nil
+}
+
+func (b *GeneratedReleaseGatingJobsBumper) Marshall(dataWithInfo *cioperatorcfg.DataWithInfo, bumpedFilename, dir string) error {
```
droslean, line 233: We can use a better name for this method since it is not marshalling anything.

pkg/branchcuts/bumper/gen-release-jobs-bumper_test.go
```diff
@@ -0,0 +1,174 @@
...
+				t.Errorf("Unexpected error: %s", err.Error())
+			}
+			if diff := cmp.Diff(&test.wantConfig, &result.Configuration); diff != "" {
+				t.Errorf("Configurations are different: %s", diff)
+			}
+		})
+	}
+}
+
+func strRef(i string) *string {
```
droslean, line 172:
```suggestion
func strPntr(i string) *string {
```

pkg/branchcuts/bumper/release-controller-config-bumper.go
```diff
@@ -0,0 +1,212 @@
...
+			absPath := filepath.Join(b.jobsDir, filename)
+			files = append(files, absPath)
+		}
+	}
+
+	return files, err
+}
+
+func (b *ReleaseControllerConfigBumper) Unmarshall(file string) (*ReleaseConfig, error) {
+	raw, err := gzip.ReadFileMaybeGZIP(file)
```
droslean, line 61: Is there any case that the file will be gzipped?

pkg/branchcuts/bumper/release-controller-config-bumper.go
```diff
@@ -0,0 +1,212 @@
...
+		if b.getFilesRegexp.Match([]byte(filename)) {
+			absPath := filepath.Join(b.jobsDir, filename)
+			files = append(files, absPath)
+		}
+	}
+
+	return files, err
+}
+
+func (b *ReleaseControllerConfigBumper) Unmarshall(file string) (*ReleaseConfig, error) {
```
droslean, line 60: Let's use a better name. This function return a `*ReleaseConfig`

pkg/branchcuts/bumper/release-controller-config-bumper.go
```diff
@@ -0,0 +1,212 @@
...
+		if err != nil {
+			return err
+		}
+		verify[newKey] = v
+	}
+	releaseConfig.Verify = verify
+	return nil
+}
+
+func (b *ReleaseControllerConfigBumper) Marshall(releaseConfig *ReleaseConfig,
```
droslean, line 195: Let's use a better name here as well.

pkg/branchcuts/bumper/release-controller-config-bumper.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to marshall release config")
+	}
+	if err := ioutil.WriteFile(absolutePath, raw, 0666); err != nil {
+		return fmt.Errorf("failed to write file %s", absolutePath)
+	}
+	return nil
+}
+
+func setMapIfNil[T any](x *map[string]T, y map[string]T) {
```
droslean, line 208: We don't need to do anything like this. If a map is nil, you can just check it and `make()` it.

pkg/branchcuts/bumper/bumper.go
```diff
@@ -0,0 +1,106 @@
...
+
+	Marshall(obj T, bumpedFilename, dir string) error
+}
+
+type BumpingOptions struct {
+	OutDir string
+}
+
+// Bump bumps files using the Bumpers b according to the BumpingOptions.
+func Bump[T any](b Bumper[T], o *BumpingOptions) (bool, error) {
```
droslean, line 31: The whole function is not needed.

pkg/branchcuts/bumper/bumper.go
```diff
@@ -0,0 +1,106 @@
...
+	"fmt"
+	"os"
+	"path"
+
+	"github.com/sirupsen/logrus"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+type Bumper[T any] interface {
```
droslean, line 14: I don't think that you need to specify the type here, and I don't believe that you need the whole interface. Each of the tools uses those methods in a different way.  You could have the `Bump` as a function that gets a list of objects and destinations to bump.

droslean:
/hold

I don't feel confident in merging this PR because is HUGE. I would prefer to split this PR into 4 smaller ones.

jmguzik:
@danilo-gemoli what about this one?

pkg/branchcuts/bumper/replace_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+func TestReplaceWithNextVersion(t *testing.T) {
+	tests := []struct {
+		name     string
+		line     string
+		major    int
+		expected string
+	}{
+		{
+			"Bump to the next version properly",
```
droslean, line 17: Let's add the fields in all test cases to be more readable.

pkg/branchcuts/bumper/replace.go
```diff
@@ -0,0 +1,78 @@
...
+// Matches is an array of regex matches that could be obtained by the following example:
+// string: ocp_4.5-4.6-4.5
+// pattern: 4\.\d+
+// matches: [3]string{}
+// matches[0]: []string{ "4.5", "5" }
+// matches[1]: []string{ "4.6", "6" }
+// matches[2]: []string{ "4.5", "5" }
+//
+// Given the previous input, the function returns []int{6, 5}
+func uniqueSortedMinors(matches [][]string) ([]int, error) {
```
droslean, line 56: This looks really smelly and adds a lot of complexity for no reason. Are you sure that this is the only way of doing this?

danilo-gemoli: When it comes to bump filenames we have to handle some cases like this:
`openshift-release-master__stable-4.7-upgrade-from-stable-4.6-from-stable-4.5.yaml`
that should be transformed into:
`openshift-release-master__stable-4.8-upgrade-from-stable-4.7-from-stable-4.6.yaml`.

In order to cope with that properly you should start from the biggest minor way to the smallest and keep bumping each of them (bottom-up simply doesn't work). So you catch all the versions `{"4.6","4.7","4.8",}` then  all the minors out of them `{"6","7","8"}` and finally in a descending order `{"8","7","6"}`. The routine `uniqueSortedMinors` tries to accomplish that. By the way I'm going to use the `sets` the reduce the lines of code if we stick with this approach.

That's not the only way, we could start from the minor of the version we're targeting, that is the future release, and repeat the processed I've described before. So if we're bumping to `4.13` (future release) the code would look like as follow:
```golang
// As an example: 
// line := "openshift-release-master__stable-4.12-upgrade-from-stable-4.11-from-stable-4.10.yaml"
major, minor := 4, 13
for minor := minor-1; minor >= 0; minor-- {
    curVersion := fmt.Sprintf("%d.%d", major, minor)
    nextVersion := fmt.Sprintf("%d.%d", major, minor+1)
    line = strings.ReplaceAll(line, curVersion, nextVersion)
}
// Result: 
// line := "openshift-release-master__stable-4.13-upgrade-from-stable-4.12-from-stable-4.11.yaml"

```
 That is going to work too and it will cycle a little bit more, but to be honest I don't think it doesn't make (and it won't make in the future too unless numbers become huge) any difference in terms of performance here.

pkg/branchcuts/bumper/replace_test.go
danilo-gemoli: Sure.

pkg/branchcuts/bumper/replace.go
droslean: Why we can't just increment any `4.x` value?

jmguzik: @droslean I don't think incrementing any `4.x` is a correct behavior if we are doing a branch cut for eg 4.12. Maybe it covers most of the cases, but not sure, this could lead to bumping something that is hard coded for example. That being said, it is true that normally we operate in ranges `4.x-2 `to `4.x+2`, and bumping anything in that range I would consider safe.

danilo-gemoli: We can bump any `4.x` value but the order the increment is done matters.

Consider again this line:
`openshift-release-master__stable-4.12-upgrade-from-stable-4.11-from-stable-4.10.yaml`
that has to be transformed to:
`openshift-release-master__stable-4.13-upgrade-from-stable-4.12-from-stable-4.11.yaml`

If we bump it in an order different from descending, the following scenario could happen:
1. Bumping from `4.10` to `4.11`
`openshift-release-master__stable-4.12-upgrade-from-stable-4.11-from-stable-4.11.yaml`
2. Bumping from `4.11` to `4.12`
`openshift-release-master__stable-4.12-upgrade-from-stable-4.12-from-stable-4.12.yaml`
3. Bumping from `4.12` to `4.13`
`openshift-release-master__stable-4.13-upgrade-from-stable-4.13-from-stable-4.13.yaml`

That's not what we expect.

droslean: Bumping from `4.10` to `4.11`:
`openshift-release-master__stable-($TO+0.1)-upgrade-from-stable-$TO-from-stable-$TO.yaml`

droslean: Either way. I am still not convinced that this is the best way. Why do we need to sort them? And what is this 2d string array? Ultimately, we just need to match a regex and increment it based on the given `FROM` and `TO` versions. This applies to everything.

danilo-gemoli: > Bumping from 4.10 to 4.11:
> `openshift-release-master__stable-($TO+0.1)-upgrade-from-stable-$TO-from-stable-$TO.yaml`

If what you intended was `TO="4.11"` then it will result in: 
`openshift-release-master__stable-4.12-upgrade-from-stable-4.11-from-stable-4.11.yaml`
and I'm not sure it's what we want.

If the starting string is:
`openshift-release-master__stable-4.12-upgrade-from-stable-4.11-from-stable-4.10.yaml`

the pattern it follows is:
`N=12`
`openshift-release-master__stable-4.$(( N ))-upgrade-from-stable-4.$(( N-1 ))-from-stable-4.$(( N-2 )).yaml`

therefore it should be transformed to (every minor version incremented by 1):
`openshift-release-master__stable-4.$(( N+1 ))-upgrade-from-stable-4.$(( N ))-from-stable-4.$(( N-1 )).yaml`

that is:
`openshift-release-master__stable-4.13-upgrade-from-stable-4.12-from-stable-4.11.yaml`

> Why do we need to sort them?

According to my approach we need to sort them because of [this](https://github.com/openshift/ci-tools/pull/2866#discussion_r1033431826).

> And what is this 2d string array?

It is the result of function [Regexp.FindAllStringSubmatch](https://pkg.go.dev/regexp#Regexp.FindAllStringSubmatch) from the standard library.

> I am still not convinced that this is the best way.

Feel free to suggest any code so we can discuss on it together.

jmguzik: The reason I do not like this function is not necessarily the method or the complexity but the input, I feel it could be better than `[][]string` which is described not in the code but in the comment. I am ok, to extract this discussion and upgrade of this mechanism to another PR or minor Jira issue, just to prevent this whole tool from rotting.

jmguzik (approved):

jmguzik:
@danilo-gemoli please prepare alternative step descriptions for your tools in the branching document. Please do not remove existing ones. Document where your tools still need manual intervention (eg these bumps in comments or commands). You can just simply do it or create a Jira task if needed.

danilo-gemoli:
/override ci/prow/e2e
we should reduce e2e tests flakiness ASAP

jmguzik (approved):

PR #2872, opened 2022-06-16
upkeep: updated generated job names

eggfoobar:
Single-node jobs have been updated to reference OVN in the names, re-ran generate job names to update with the changes.

Signed-off-by: ehila <ehila@redhat.com>


eggfoobar:
/hold
/cc @DennisPeriquet @neisw

Waiting on guidance for updating the BigQuery tables, this is just the update to the generated name update text file.

Ran:
```sh
./job-run-aggregator generate-job-names > ./pkg/jobrunaggregator/jobtableprimer/generated_job_names.txt
```

eggfoobar:
Agreed next steps are to automate the process of priming the table names after the names have been generated, created PR to add cronjob for updating table names periodically https://github.com/openshift/continuous-release-jobs/pull/1167 

@DennisPeriquet @neisw the idea here being that every 4 hours the cron job would run to update big query based off of the list in this repo, so we would open a PR to update the list and within the next 4 hours that list will be updated in BigQuery. Do we have any thoughts on the 4 hour frequency, do we want to increase/decrease it?

eggfoobar:
No longer needed, process is now automated. See https://github.com/openshift/ci-tools/pull/2936

PR #2873, opened 2022-06-16
Pod Scaler: cap memory requests at 20Gi

smg247:
We need to cap the amount of memory that the `pod-scaler` requests to `20Gi`. I made this amount configurable and also made the existing `cpu-cap` configurable.

For: [DPTP-2928](https://issues.redhat.com/browse/DPTP-2928)

/cc @openshift/test-platform

### DIFF
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -32,7 +32,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/steps"
 )
 
-func admit(port, healthPort int, certDir string, client buildclientv1.BuildV1Interface, loaders map[string][]*cacheReloader, mutateResourceLimits bool) {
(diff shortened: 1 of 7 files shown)


cmd/pod-scaler/admission.go
```diff
@@ -243,16 +243,23 @@ func preventUnschedulable(resources *corev1.ResourceRequirements) {
 	if resources.Requests == nil {
 		return
 	}
-	if _, ok := resources.Requests[corev1.ResourceCPU]; !ok {
-		return
+
+	if _, ok := resources.Requests[corev1.ResourceCPU]; ok {
+		// 10 CPU is a ballpark number. Current build farm nodes have 16 CPU so pods get unschedulable
+		// around 14 CPU
+		// TODO(DPTP-2525): Make configurable and perhaps cluster-specific?
```
droslean, line 250: IMHO this could be done in this PR. At least make it configurable and as a follow-up, we can determine the desired limit per cluster.

smg247: yeah, I suppose it would be nice if both the CPU and memory caps were configurable

PR #2875, opened 2022-06-17
ci-operator: add build pending timeout period

bbguimaraes:
This follows the similar implementation (and uses the same time period)
in pkg/steps/template.go.

---

Example (with timeout changed to `1s`):

```yaml
resources:
  src:
    requests:
      memory: 1T
```

```
INFO[2022-06-17T18:22:59Z] Building src
INFO[2022-06-17T18:23:05Z] build didn't start running within 1s (phase: Pending):
Found 1 events for Pod src-build:
* 0x : 0/23 nodes are available: 1 node(s) had taint {node-role.kubernetes.io/ci-builds-worker: ci-builds-worker}, that the pod didn't tolerate, 1 node(s) had taint {node-role.kubernetes.io/ci-prowjobs-worker: ci-prowjobs-worker}, that the pod didn't tolerate, 1 node(s) were unschedulable, 2 Insufficient memory, 2 node(s) had taint {ci.openshift.io/ci-search: true}, that the pod didn't tolerate, 3 node(s) had taint {node-role.kubernetes.io/infra: }, that the pod didn't tolerate, 3 node(s) had taint {node-role.kubernetes.io/master: }, that the pod didn't tolerate, 4 node(s) had taint {node-role.kubernetes.io/ci-longtests-worker: ci-longtests-worker}, that the pod didn't tolerate, 6 node(s) had taint {node-role.kubernetes.io/ci-tests-worker: ci-tests-worker}, that the pod didn't tolerate.
```

---

There is an unfortunate layering violation here in that we are forced to
get to the build pod through the `pod-name` annotation and examine it.
I could not find a way to do this through the `Build` object (as is
possible for logs, for example).  A pending build has very little
information:

```
status:
  conditions:
  - lastTransitionTime: "2022-06-17T11:06:41Z"
    lastUpdateTime: "2022-06-17T11:06:41Z"
    status: "False"
    type: New
  - lastTransitionTime: "2022-06-17T11:06:41Z"
    lastUpdateTime: "2022-06-17T11:06:41Z"
    status: "True"
    type: Pending
  output: {}
  outputDockerImageReference: image-registry.openshift-image-registry.svc:5000/bbguimaraes0/pipeline:src
  phase: Pending
```

It is only by examining the build pod (reusing the existing code which
handles test pods, at least) that the cause can be determined.

---

/hold

It's Friday.


pkg/defaults/defaults.go
```diff
@@ -244,19 +244,19 @@ func fromConfig(
 			step = steps.InputImageTagStep(&conf, client, jobSpec)
 			inputImages[conf.InputImage] = struct{}{}
 		} else if rawStep.PipelineImageCacheStepConfiguration != nil {
-			step = steps.PipelineImageCacheStep(*rawStep.PipelineImageCacheStepConfiguration, config.Resources, buildClient, jobSpec, pullSecret)
+			step = steps.PipelineImageCacheStep(*rawStep.PipelineImageCacheStepConfiguration, config.Resources, buildClient, podClient, jobSpec, pullSecret)
```
droslean, line 247: thoughts for food... We should really need to implement a stepFactory or something similar. I hate the fact that we keep passing clients everywhere.

PR #2880, opened 2022-06-29
Require workflow use in #forum-testplatform

smg247:
Now that workflows exist we should gently remind users that post a direct message to use the workflows instead. We just need to check that the bot_id in the message is non-empty to verify that a workflow is being used. Note that actually deleting their message requires a channel admin's slack token which we don't currently posses. If we have to use a heavier handed approach here we will need to work that out.

I also added a few flags to slack-bot to make local testing a little nicer.

For: [DPTP-2951](https://issues.redhat.com/browse/DPTP-2951)

Screen Shot 2022-06-29 at 1 00 04 PM


smg247:
https://github.com/openshift/ci-tools/pull/2881 will fix the e2e flakes

smg247:
It appears my attempt to fix the `e2e` flakes didn't pan out. I am going to continue looking into what is going on there, but in the meantime those shouldn't block this PR.

/override ci/prow/e2e

pkg/slack/events/helpdesk/helpdesk-message.go
```diff
@@ -50,27 +45,37 @@ func Handler(client messagePoster, keywordsConfig KeywordsConfig) events.Partial
...
+			if event.Channel != forumChannelId {
+				log.Debugf("not in correct channel. wanted: %s, message was in: %s", forumChannelId, event.Channel)
 				return false, nil
 			}
-			if !strings.Contains(event.Text, helpdeskMention) {
+
+			var response []slack.Block
+			notifyToUseWorkflow := requireWorkflowsInForum && event.ThreadTimeStamp == "" && event.BotID == ""
+			if notifyToUseWorkflow {
+				log.Debugf("Top level message not from a workflow, notifying user")
```
hongkailiu, line 56:
```suggestion
				log.Debug("Top level message not from a workflow, notifying user ...")
```

pkg/slack/events/helpdesk/helpdesk-message.go
```diff
@@ -50,27 +45,37 @@ func Handler(client messagePoster, keywordsConfig KeywordsConfig) events.Partial
...
-			if event.Channel != channelId {
-				log.Debugf("not in correct channel. wanted: %s, message was in: %s", channelId, event.Channel)
+			if event.Channel != forumChannelId {
+				log.Debugf("not in correct channel. wanted: %s, message was in: %s", forumChannelId, event.Channel)
 				return false, nil
 			}
-			if !strings.Contains(event.Text, helpdeskMention) {
+
+			var response []slack.Block
+			notifyToUseWorkflow := requireWorkflowsInForum && event.ThreadTimeStamp == "" && event.BotID == ""
```
hongkailiu, line 54: Is `event.ThreadTimeStamp == "" && event.BotID == ""` the indicator that workflow is not used?

smg247: the `BotID` shows us that either a workflow was used or the post was from some bot (which currently isn't a thing that occurs in #forum-testplatform), and the lack of  `ThreadTimeStamp` shows us that it is a top-level message and not a reply.

PR #2886, opened 2022-07-05
verify user is non-nil before posting reply

smg247:
I am not completely sure how it is happening, but it appears that the bot will post a top-level message that looks like
Screen Shot 2022-07-05 at 4 43 16 PM
when a workflow is used.

From looking through the logs it was referencing a message with no `bot-id` **and** no `user`. We only need to do this auto-reply when the user is non-nil.

/cc @openshift/test-platform


smg247:
The `e2e` failure is a flake I am working on with https://issues.redhat.com/browse/DPTP-2976

/override ci/prow/e2e

jmguzik (approved):

PR #2904, opened 2022-07-19
Fix issues found by snyk-deps

bear-redhat:


.snyk
```diff
@@ -2,4 +2,48 @@ exclude:
   global:
     - "vendor/**"
     - "**/*_test.go"
-
+ignore:
+  'snyk:lic:golang:github.com:hashicorp:vault:api:MPL-2.0':
```
droslean, line 6: Can we do something like `snyk:lic:golang:github.com:*:MPL-2.0` here?

bear-redhat: Tried, not working

PR #2911, opened 2022-07-23
Observer pods

jmguzik:
/cc @jmguzik 
/hold


jmguzik:
As discussed today, for now, temporarily e2e test will be removed (I removed file from the PR, and left the rest of the files) and it will be the task for the next sprint to implement it properly.

jmguzik:
/cc @danilo-gemoli @bbguimaraes 
were involved in the discussion.

jmguzik:
As discussed with @bear-redhat we should not care so much about deps.
Not sure is optional or not so:
/override ci/prow/snyk-deps

danilo-gemoli:
I hope this will not break anything
/lgtm

jmguzik:
> I hope this won't break anything /lgtm

My hope is the same xD.

test/e2e/observer/e2e_test.go
```diff
@@ -0,0 +1,80 @@
+//go:build e2e
```
jmguzik: This file was removed and should be re-added once we figure out e2e testing.

PR #2917, opened 2022-07-28
[WIP] ci-secret-generator: read only `stdout`

bbguimaraes:


cmd/ci-secret-generator/main.go
```diff
@@ -120,10 +120,19 @@ func (o *options) validateConfig() error {
 }
 
 func executeCommand(command string) ([]byte, error) {
-	out, err := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command).CombinedOutput()
+	cmd := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command)
+	err := cmd.Wait()
```
hongkailiu, line 124: Need to call `Start()` first like [the example](https://pkg.go.dev/os/exec#example-Cmd.Start)?

I think `cmd.Output()` would fix the bug but in a long run `cmd.Run()` with checking on error output is a better solution.

Steal from https://stackoverflow.com/questions/1877045/how-do-you-get-the-output-of-a-system-command-in-go:
```
cmd := exec.Command("date")
var outb, errb bytes.Buffer
cmd.Stdout = &outb
cmd.Stderr = &errb
err := cmd.Run()
if err != nil {
    log.Fatal(err)
}
fmt.Println("out:", outb.String(), "err:", errb.String())
```

bbguimaraes:
Superseded by https://github.com/openshift/ci-tools/pull/2922.

/close

PR #2920, opened 2022-07-29
ci-secret-generator: add unit tests for executeCommand

hongkailiu:
https://issues.redhat.com/browse/DPTP-3034

/cc @danilo-gemoli @bbguimaraes

### DIFF
--- a/cmd/ci-secret-generator/main_test.go
+++ b/cmd/ci-secret-generator/main_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
@@ -224,6 +225,50 @@ func TestValidateContexts(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			if err := validateContexts(itemContextsFromConfig(tc.cfg), tc.bootstrapCfg); err != nil {
 				t.Errorf("validation failed unexpectedly: %v", err)
+			}
+		})
+	}
+}
+func TestExecuteCommand(t *testing.T) {
+	testCases := []struct {
+		name          string
+		cmd           string
+		expected      []byte
+		expectedError error
+	}{
+		{
+			name:     "basic case",
+			cmd:      "echo basic case",
+			expected: []byte("basic case\n"),
+		},
+		{
+			name:          "error on no output",
+			cmd:           "true",
+			expectedError: fmt.Errorf("command \"true\" returned no output"),
+		},
+		{
+			name:          "error on cmd failure",
+			cmd:           "false",
+			expectedError: fmt.Errorf(" : exit status 1"),
+		},
+		{
+			name:          "error if stderr is not empty",
+			cmd:           ">&2 echo some error",
+			expectedError: fmt.Errorf("command \">&2 echo some error\" has error output"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.name == "error if stderr is not empty" {
+				t.Skip("Skipping a test before https://issues.redhat.com/browse/DPTP-3033 is complete:", tc.name)
+			}
+			actual, actualError := executeCommand(tc.cmd)
+			if diff := cmp.Diff(tc.expected, actual); diff != "" {
+				t.Errorf("%s: mismatch (-expected +actual), diff: %s", tc.name, diff)
+			}
+			if diff := cmp.Diff(tc.expectedError, actualError, testhelper.EquateErrorMessage); diff != "" {
+				t.Errorf("%s: mismatch (-expected +actual), diff: %s", tc.name, diff)
 			}
 		})
 	}


cmd/ci-secret-generator/main_test.go
```diff
@@ -228,3 +229,47 @@ func TestValidateContexts(t *testing.T) {
...
+		{
+			name:          "error if stderr is not empty",
+			cmd:           ">&2 echo some error",
+			expectedError: fmt.Errorf("command \">&2 echo some error\" has error output"),
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.name == "error if stderr is not empty" {
```
hongkailiu, line 263: @danilo-gemoli 

You will remove the skip when DPTP-3033 is done.

cmd/ci-secret-generator/main_test.go
```diff
@@ -228,3 +229,47 @@ func TestValidateContexts(t *testing.T) {
...
+		},
+		{
+			name:          "error on no output",
+			cmd:           "true",
+			expectedError: fmt.Errorf("command \"true\" returned no output"),
+		},
+		{
+			name:          "error on cmd failure",
+			cmd:           "false",
+			expectedError: fmt.Errorf(" : exit status 1"),
```
bbguimaraes, line 252: This message is strange, but I suppose this just tests the current behavior.

cmd/ci-secret-generator/main_test.go
```diff
@@ -228,3 +229,47 @@ func TestValidateContexts(t *testing.T) {
...
+			if tc.name == "error if stderr is not empty" {
+				t.Skip("Skipping a test before https://issues.redhat.com/browse/DPTP-3033 is complete:", tc.name)
+			}
+			actual, actualError := executeCommand(tc.cmd)
+			if diff := cmp.Diff(tc.expected, actual); diff != "" {
+				t.Errorf("%s: actual does not match expected, diff: %s", tc.name, diff)
+			}
+			if diff := cmp.Diff(tc.expectedError, actualError, testhelper.EquateErrorMessage); diff != "" {
+				t.Errorf("%s: actual does not match expected, diff: %s", tc.name, diff)
+			}
```
bbguimaraes, line 272: No need to push another version just for this, but `T::Run` already adds the test case name (the first argument) to the error messages.

bbguimaraes: OK, this may be worth an edit after all: the diff is `expected → actual`, but the message says "actual does not match expected" — the latter is what I normally expect ([e.g.](https://doc.qt.io/qt-5/qtest.html#QCOMPARE)).

cmd/ci-secret-generator/main_test.go
hongkailiu: Yes, it is.

cmd/ci-secret-generator/main_test.go
hongkailiu: Shit. That is why I always got confused when I have to debug the diff.

https://github.com/google/go-cmp/blob/master/cmp/example_test.go

Let me fix it.

cmd/ci-secret-generator/main_test.go
```diff
@@ -228,3 +229,47 @@ func TestValidateContexts(t *testing.T) {
...
+		{
+			name:          "error on cmd failure",
+			cmd:           "false",
+			expectedError: fmt.Errorf(" : exit status 1"),
+		},
+		{
+			name:          "error if stderr is not empty",
+			cmd:           ">&2 echo some error",
+			expectedError: fmt.Errorf("command \">&2 echo some error\" has error output"),
+		},
```
danilo-gemoli, line 258: Is this to be considered as an error even if the command return 0 as exit status?

cmd/ci-secret-generator/main_test.go
bbguimaraes: Yes, it's moderately annoying that the division between want/got and got/want is close to 50% in these things.

I'm fine with either (I bet we have both in our code), but the message has to reflect the order, otherwise it's impossible to decode the error message, for exactly this reason.

cmd/ci-secret-generator/main_test.go
hongkailiu: The reason is that it is likely that we need to handle the error message even if it returns 0.
The deprecation msg is an example.
We do not want to wait until it errors out to handle it. It would be too late and we would have pressure to look for solutions.
Instead, we want to handle it once we see the deprecation message, at the time, the command might still return 0.

bbguimaraes: We didn't coordinate this, but I had the same idea.

This would allow us to catch future deprecation messages (or similar problems), which we would otherwise never see.  Unless we find some hostile program whose `stderr` we cannot control, I think this would be helpful.

cmd/ci-secret-generator/main_test.go
hongkailiu: Updated with a clear - and + meaning as the example.

PR #2922, opened 2022-07-29
executeCommand doesn't mix std streams anymore

danilo-gemoli:
/cc @hongkailiu @bbguimaraes

### DIFF
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -141,14 +141,9 @@ func executeCommand(command string) ([]byte, error) {
 	if err := cmd.Run(); err != nil {
 		stderr := errBuf.Bytes()
 		stdout := outBuf.Bytes()
-		partialStreams := true
 		// The command completed with non zero exit code, standard streams *should* be available.
-		if _, ok := err.(*exec.ExitError); ok {
-			partialStreams = false
-		}
-		// At this point we can't easily tell why the command failed,
-		// there is no guarantee neither stdout nor stderr have been properly filled.
-		return nil, fmtExecCmdErr(execCmdRunErrAction, command, err, stdout, stderr, partialStreams)
+		_, partialStreams := err.(*exec.ExitError)
+		return nil, fmtExecCmdErr(execCmdRunErrAction, command, err, stdout, stderr, !partialStreams)
 	}
 
 	stderr := errBuf.Bytes()
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -23,8 +23,8 @@ import (
 
 const (
 	execCmdRunErrAction            = "run"
-	execCmdValidateStdoutErrAction = "validate stdout"
-	execCmdValidateStderrErrAction = "validate stderr"
+	execCmdValidateStdoutErrAction = "validate stdout of"
+	execCmdValidateStderrErrAction = "validate stderr of"
 	execCmdErrFmt                  = "failed to %s command %q: %w\n%s:\n%s\n%s:\n%s"
 )
 
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -120,17 +120,36 @@ func (o *options) validateConfig() error {
 }
 
 func executeCommand(command string) ([]byte, error) {
-	out, err := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command).CombinedOutput()
-	if err != nil {
-		return nil, fmt.Errorf("%s : %w", string(out), err)
+	cmd := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command)
+	var outBuf, errBuf bytes.Buffer
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	if err := cmd.Run(); err != nil {
+		// The command completed with non zero exit code, standard streams *should* be available.
+		if _, ok := err.(*exec.ExitError); ok {
+			errStr := errBuf.String()
+			if len(errStr) == 0 {
+				return nil, fmt.Errorf(" : %w", err)
+			}
+			return nil, fmt.Errorf("%s: %w", errStr, err)
+		}
+		// At this point we can't easily tell why the command failed,
+		// there is no guarantee neither stdout nor stderr are valid.
+		return nil, fmt.Errorf("failed to run the command: %w", err)
 	}
-	if len(out) == 0 || len(bytes.TrimSpace(out)) == 0 {
+
+	if len(errBuf.Bytes()) != 0 {
+		return nil, fmt.Errorf("command %q has error output", command)
+	}
+
+	stdout := outBuf.Bytes()
+	if len(stdout) == 0 || len(bytes.TrimSpace(stdout)) == 0 {
 		return nil, fmt.Errorf("command %q returned no output", command)
 	}
-	if string(bytes.TrimSpace(out)) == "null" {
+	if string(bytes.TrimSpace(stdout)) == "null" {
 		return nil, fmt.Errorf("command %s returned 'null' as output", command)
 	}
-	return out, nil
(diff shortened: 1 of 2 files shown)


cmd/ci-secret-generator/main.go
```diff
@@ -120,17 +120,36 @@ func (o *options) validateConfig() error {
...
-	out, err := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command).CombinedOutput()
-	if err != nil {
-		return nil, fmt.Errorf("%s : %w", string(out), err)
+	cmd := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command)
+	var outBuf, errBuf bytes.Buffer
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	if err := cmd.Run(); err != nil {
+		// The command completed with non zero exit code, standard streams *should* be available.
+		if _, ok := err.(*exec.ExitError); ok {
```
bbguimaraes, line 129: Is this test following some standard recommendation?  From my reading of the code, it is possible to have another type of error and at least partial content in `Stdout`/`Stderr` (e.g. a `read` error in the middle of the execution).

cmd/ci-secret-generator/main.go
```diff
@@ -120,17 +120,36 @@ func (o *options) validateConfig() error {
...
+			errStr := errBuf.String()
+			if len(errStr) == 0 {
+				return nil, fmt.Errorf(" : %w", err)
+			}
+			return nil, fmt.Errorf("%s: %w", errStr, err)
+		}
+		// At this point we can't easily tell why the command failed,
+		// there is no guarantee neither stdout nor stderr are valid.
+		return nil, fmt.Errorf("failed to run the command: %w", err)
```
bbguimaraes, line 138: I know some of these already exist, but let's turn them into decent messages, since we're changing this code.

They should always include the command which was executed (or not, if the caller does, but we have both here).

bbguimaraes (commented):
/hold

Needs https://github.com/openshift/ci-tools/pull/2920 and tests.

cmd/ci-secret-generator/main.go
danilo-gemoli: According to the `cmd.Run` documentation, the error of type `(*exec.ExitError)` is returned when the process being invoked did not complete successfully and, as far I have understood, it translates into what follows:

- the process completed
- its exit code differs from 0
- `cmd.Run` was able to both check the exit code and copy std streams into buffers

The best idiomatic, maybe yet outdated, approach I managed to find is [example_test.go](https://go-review.googlesource.com/c/go/+/213337/1/src/os/exec/example_test.go)

Do we care catching any execution path where `stderr` and/or `stdout` are not available?
IMHO, we should be fine handling the case we are sure they are available and just going past the others.

cmd/ci-secret-generator/main.go
danilo-gemoli: Yeah, that code sucks and it exists only to pass tests.

cmd/ci-secret-generator/main.go
bbguimaraes: My comment was specifically about:

```
// At this point we can't easily tell why the command failed,
// there is no guarantee neither stdout nor stderr are valid.
```

`outBuf` and `errBuf` are just byte arrays, they are always valid.  They may contain incomplete data in case the error was caused by the copies (and we can indicate that in the error message), but there is no reason to not display them.

cmd/ci-secret-generator/main.go
danilo-gemoli: Look fine, I'll do that.

cmd/ci-secret-generator/main.go
danilo-gemoli: It sounds fine to me, I would propose something like:
```diff
failed to $action command "$command": $err
- output:
+ output (may be incomplete):
$stdout
- error output:
+ error output (may be incomplete):
$stderr
```

cmd/ci-secret-generator/main.go
```diff
@@ -120,17 +120,34 @@ func (o *options) validateConfig() error {
...
+	if len(errBuf.Bytes()) != 0 {
+		return nil, fmt.Errorf("command %q has error output", command)
+	}
+
+	stdout := outBuf.Bytes()
+	if len(stdout) == 0 || len(bytes.TrimSpace(stdout)) == 0 {
 		return nil, fmt.Errorf("command %q returned no output", command)
 	}
-	if string(bytes.TrimSpace(out)) == "null" {
+	if string(bytes.TrimSpace(stdout)) == "null" {
```
hongkailiu, line 147: Not related to the PR.
I wonder why "null" leads to an error. Unlikely to cause any damage though.

hongkailiu:
The code looks good to me.

We need to remove the skipped test and fix the broken ones with the updated error message.

cmd/ci-secret-generator/main.go
danilo-gemoli: Really have no idea why, I need to dig deeper eventually.

bbguimaraes: My guess is it's for `jq`, we have some instances in the configuration file.

(why the check is in the program, not the script, I do not understand)

bbguimaraes:
```
$ jq '.test' <<< '{}'
null
```

cmd/ci-secret-generator/main_test.go
```diff
@@ -242,27 +290,33 @@ func TestExecuteCommand(t *testing.T) {
 			expected: []byte("basic case\n"),
 		},
 		{
-			name:          "error on no output",
-			cmd:           "true",
-			expectedError: fmt.Errorf("command \"true\" returned no output"),
+			name: "error on no output",
+			cmd:  "true",
+			expectedError: fmtExecCmdErr(execCmdValidateStdoutErrAction, "true",
```
hongkailiu, line 295: Could we display the error directly, instead of using `fmtExecCmdErr`?

cmd/ci-secret-generator/main.go
```diff
@@ -120,17 +133,54 @@ func (o *options) validateConfig() error {
...
+	cmd.Stderr = &errBuf
+
+	if err := cmd.Run(); err != nil {
+		stderr := errBuf.Bytes()
+		stdout := outBuf.Bytes()
+		partialStreams := true
+		// The command completed with non zero exit code, standard streams *should* be available.
+		if _, ok := err.(*exec.ExitError); ok {
+			partialStreams = false
+		}
```
bbguimaraes, line 148:
```suggestion
		// The command completed with non zero exit code, standard streams *should* be available.
		_, partialStreams := err.(*exec.ExitError)
```

cmd/ci-secret-generator/main.go
```diff
@@ -120,17 +133,54 @@ func (o *options) validateConfig() error {
...
+		// At this point we can't easily tell why the command failed,
+		// there is no guarantee neither stdout nor stderr have been properly filled.
```
bbguimaraes, line 150: Comment no longer applies.

cmd/ci-secret-generator/main_test.go
danilo-gemoli: Sure

cmd/ci-secret-generator/main.go
hongkailiu: I feel `partialStreams` is negated comparing to the original logic.

danilo-gemoli: I'm not sure about that: if the error is `*exec.ExitError` then streams should be ok, and that means `partialStreams = false`

bbguimaraes: Yes, sorry, edited too quickly.  But you get the point, there's no need for the conditional.

cmd/ci-secret-generator/main.go
danilo-gemoli: We should consider opening a PR to refactor that.

PR #2927, opened 2022-08-03
add upgrade filter for install test groups

neisw:


pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer._test.go
```diff
@@ -0,0 +1,73 @@
+package jobruntestcaseanalyzer
```
stbenjam, line 1: This file should be named analyzer_test.go, there's an extra dot right now

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer._test.go
```diff
@@ -0,0 +1,73 @@
...
+func TestGetJobs(t *testing.T) {
+
+	tests := map[string]struct {
+		allJobs          []jobrunaggregatorapi.JobRow
+		expectedJobNames sets.String
+		filters          string
+	}{
+		"test upgrade filter":   {expectedJobNames: sets.String{"name-without-it": sets.Empty{}, "name-with-something-else": sets.Empty{}}, filters: "upgrade"},
+		"test no filter":        {expectedJobNames: sets.String{"name-without-it": sets.Empty{}, "name-with-something-else": sets.Empty{}, "name-with-upgrade": sets.Empty{}}},
+		"test multiple filters": {expectedJobNames: sets.String{"name-without-it": sets.Empty{}}, filters: "upgrade,something"},
```
stbenjam, line 22: Nit, but it's clearer to me and also IMHO generally a better test if the cases look more like the real world data.  Can we use real job names here?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -56,7 +57,7 @@ func (s *testCaseAnalyzerJobGetter) GetJobs(ctx context.Context) ([]jobrunaggreg
 	}
 
 	// if PR payload, only find the exact jobs
-	if len(*s.jobGCSPrefixes) > 0 {
+	if s.jobGCSPrefixes != nil && len(*s.jobGCSPrefixes) > 0 {
```
stbenjam, line 60: Thanks for fixing this!

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -145,6 +146,8 @@ func (f *JobRunsTestCaseAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&f.WorkingDir, "working-dir", f.WorkingDir, "The directory to store caches, output, and the like.")
 	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for analyzing job to complete.")
 	fs.Var(&jobGCSPrefixSlice{&f.JobGCSPrefixes}, "explicit-gcs-prefixes", "a list of gcs prefixes for jobs created for payload. Only used by per PR payload promotion jobs. The format is comma-separated elements, each consisting of job name and gcs prefix separated by =, like openshift-machine-config-operator=3028-ci-4.11-e2e-aws-ovn-upgrade~logs/openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade")
+
+	fs.StringVar(&f.JobNameFilters, "job-name-filter", f.JobNameFilters, "Applied only when --explicit-gcs-prefixes is not specified.  A comma separated list of substrings used to filter JobNames from the analysis")
```
stbenjam, line 150: I'd suggest using `StringArrayVar` instead to just get an array, and you can avoid splitting you do later on.

stbenjam (commented):
Just a few nits

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer._test.go
neisw: Good thought and I agree, thanks.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer._test.go
neisw: Nice catch

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
neisw: Thanks!

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -86,6 +87,21 @@ func (s *testCaseAnalyzerJobGetter) filterJobsForPayload(allJobs []jobrunaggrega
...
+				if strings.Contains(job.JobName, key) {
+					isFiltered = true
+					break
+				}
+			}
+
+			if isFiltered {
+				continue
+			}
+		}
```
stbenjam, line 103:
```suggestion
		}
```

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -86,6 +87,21 @@ func (s *testCaseAnalyzerJobGetter) filterJobsForPayload(allJobs []jobrunaggrega
...
+			isFiltered := false
+			for key := range s.filters {
+				if strings.Contains(job.JobName, key) {
+					isFiltered = true
+					break
+				}
+			}
+
+			if isFiltered {
+				continue
```
stbenjam, line 101: This pattern works fine, but Golang supports named loops to solve this problem as well:

```golang
Outer:
	for i := range allJobs {
		// ...
		if s.filters != nil {
			for key := range s.filters {
				if strings.Contains(job.JobName, key) {
					continue Outer
				}
			}
		}
	}
```

stbenjam:
/lgtm

You'll have to get @deads2k  to leave the final /approve.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
deads2k: That will likely vary by reviewer.  I prefer to see a construct as written here or a helper method with the nested loop to read easier.  Named loops make some people twitch.  I don't recall explicitly writing down something against though.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -41,6 +41,7 @@ type testCaseAnalyzerJobGetter struct {
 	platform       string
 	infrastructure string
 	network        string
+	filters        sets.String
```
deads2k, line 44: All the other criteria are positive criteria, this is the first excluding criteria.  That seems worth a comment or a name like `exclusionaryFilters`

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -145,6 +146,8 @@ func (f *JobRunsTestCaseAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&f.WorkingDir, "working-dir", f.WorkingDir, "The directory to store caches, output, and the like.")
 	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for analyzing job to complete.")
 	fs.Var(&jobGCSPrefixSlice{&f.JobGCSPrefixes}, "explicit-gcs-prefixes", "a list of gcs prefixes for jobs created for payload. Only used by per PR payload promotion jobs. The format is comma-separated elements, each consisting of job name and gcs prefix separated by =, like openshift-machine-config-operator=3028-ci-4.11-e2e-aws-ovn-upgrade~logs/openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade")
+
+	fs.StringArrayVar(&f.JobNameFilters, "job-name-filter", f.JobNameFilters, "Applied only when --explicit-gcs-prefixes is not specified.  The flag can be specified multiple times to create a list of substrings used to filter JobNames from the analysis")
```
deads2k, line 150: These are negative filters, right?  If a job matches one of these filters, it is not inspected.  That is the opposite of the `--platform` for instance.  I think choosing a name that matches that exclusionary nature is important and the help should be specific, perhaps even including a brief example.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
stbenjam: I find them concise and make it much clearer which loop is being continued. 🤷🏻  Anyway, I don't hold up reviews on it, but I think most people don't use it since they don't know it's there.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
neisw: went with `exclude-job-names` && `ExcludeJobNames` if that is agreeable.  Also added example in the comments

--exclude-job-names=upgrade
--exclude-job-names=ipv6

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
neisw: does `excludeJobNames` work?

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer_test.go
```diff
@@ -0,0 +1,84 @@
...
+			if err := fs.Parse(args); err != nil {
+				t.Fatalf("%s flag set parse returned error %#v", name, err)
+			}
+
+			if f.ExcludeJobNames != nil && len(f.ExcludeJobNames) > 0 {
+				jobGetter.excludeJobNames = sets.String{}
+				jobGetter.excludeJobNames.Insert(f.ExcludeJobNames...)
+			}
+
+			returnedJobs, err := jobGetter.GetJobs(context.TODO())
```
dgoodwin, line 56: I would suggest testing at the filterJobsForPayload level rather than GetJobs, which would eliminate the need for that large'ish ciDataClientTester. We're not interested in testing 

Alternatively, gomock would make an easy substitution, however it does not appear to be a dep here already so it would need to be added.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -86,11 +87,28 @@ func (s *testCaseAnalyzerJobGetter) filterJobsForPayload(allJobs []jobrunaggrega
...
+		}
+
 		jobs = append(jobs, job)
 	}
 	return jobs
 }
 
+func (s *testCaseAnalyzerJobGetter) isJobNameFiltered(jobName string) bool {
+
+	if s.excludeJobNames != nil {
```
deads2k, line 102: minor: kube style would be to invert this and have

```go
if s.excludeJobNames == nil{
  return false
}

for key := range..
```

Also note that you don't need the if block at all for golang, you can iterate over a nil map: https://go.dev/play/p/FStUZvQBOT2

deads2k:
with name updates, I'm ok with this.

/approve

/assign @dgoodwin 

leaving lgtm with @dgoodwin or delegate.

dgoodwin:
Could we slip in the command to autogenerate the mock, maybe in hack/update-generated.sh

neisw:
/assign @alvaroaleman 

Could you give a review of the go.mod change and addition of golang/mock

alvaroaleman:
I am not an approver here
/unassign
/uncc

hongkailiu:
I do not have much experience of https://github.com/golang/mock
The change looks reasonable to me on the go.mod and I assume that it is used only in `jobrunaggregator`.

I will approve after @dgoodwin or @deads2k adds back LGTM.

dgoodwin:
/lgtm

Thanks @hongkailiu

PR #2930, opened 2022-08-04
ci-operator: allow multi-stage steps to opt out of SA credentials

stevekuznetsov:

https://issues.redhat.com/browse/DPTP-3038

### DIFF
--- a/pkg/steps/multi_stage/gen.go
+++ b/pkg/steps/multi_stage/gen.go
@@ -124,7 +124,14 @@ func (s *multiStageTestStep) generatePods(
 		delete(pod.Labels, base_steps.ProwJobIdLabel)
 		pod.Annotations[base_steps.AnnotationSaveContainerLogs] = "true"
 		pod.Labels[MultiStageTestLabel] = s.name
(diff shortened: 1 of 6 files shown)


pkg/steps/multi_stage/gen.go
```diff
@@ -124,7 +124,13 @@ func (s *multiStageTestStep) generatePods(
 		delete(pod.Labels, base_steps.ProwJobIdLabel)
 		pod.Annotations[base_steps.AnnotationSaveContainerLogs] = "true"
 		pod.Labels[MultiStageTestLabel] = s.name
-		pod.Spec.ServiceAccountName = s.name
+		if step.NoKubeconfig == nil || !*step.NoKubeconfig {
```
smg247, line 127: can we just create a variable like `noKubeconfig := step.NoKubeconfig != nil && *step.NoKubeconfig` so that we don't have to keep doing this same logic everywhere?

stevekuznetsov: sounds good - done

PR #2934, opened 2022-08-05
Observer Pods: read-only kubeconfig

danilo-gemoli:
It seems that observer pods don't need to get write access to the kubeconfig.
This PR add "skip manageKubeconfig" functionality to `entrypoint-wrapper` such that observers get the read-only kubeconfig version as soon as it is ready.

There is nothing wrong with the `rw` variant of kubeconfig but this one:
1. the `entrypoint-wrapper` makes an empty temporary `rw` kubeconfig in `/tmp/kubeconfig-xxxx`, it sets `$KUBECONFIG` env variable to its child process (observer pod in such a case) and it finally waits for the real one to be uploaded by someone
2. when the observer pod starts it checks and finds the kubeconfig but it is (most likely) empty at the beginning
3. in the meanwhile someone has just uploaded a proper kubeconfig
4. `entrypoint-wrapper` detects it and updates the temporary `rw` copy
5. at this time only the observer pod gets the real content

We want to avoid step (2.) such that the observer pod gets the kubeconfig iff it's ready and non-empty.

/cc @jmguzik @bbguimaraes

### DIFF
--- a/cmd/entrypoint-wrapper/main.go
+++ b/cmd/entrypoint-wrapper/main.go
@@ -72,6 +72,7 @@ type options struct {
 	waitTimeout    time.Duration
 	cmd            []string
 	client         coreclientset.SecretInterface
+	roKubeconfig   bool
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
@@ -79,6 +80,7 @@ func bindOptions(flag *flag.FlagSet) *options {
 	flag.BoolVar(&opt.dry, "dry-run", false, "Print the secret instead of creating it")
 	flag.StringVar(&opt.waitPath, "wait-for-file", "", "Wait for a file to appear at this path before starting the program")
 	flag.StringVar(&opt.waitTimeoutStr, "wait-timeout", "", "Used with --wait-for-file, maximum wait time before starting the program")
+	flag.BoolVar(&opt.roKubeconfig, "ro-kubeconfig", false, "If set do not copy kubeconfig to a writable location")
 	return opt
 }
 
@@ -129,7 +131,7 @@ func (o *options) run() error {
 	var errs []error
 	ctx, cancel := context.WithCancel(context.Background())
 	go uploadKubeconfig(ctx, o.client, o.name, o.dstPath, o.dry)
-	if err := execCmd(o.cmd); err != nil {
+	if err := execCmd(o.cmd, o.roKubeconfig); err != nil {
 		errs = append(errs, fmt.Errorf("failed to execute wrapped command: %w", err))
 	}
 	// we will upload the secret from the post-execution state, so we know
@@ -225,7 +227,7 @@ func waitForFile(path string, timeout time.Duration) error {
 	}
 }
 
-func execCmd(argv []string) error {
+func execCmd(argv []string, skipManageKubeconfig bool) error {
 	proc := exec.Command(argv[0], argv[1:]...)
 	proc.Stdout = os.Stdout
 	proc.Stderr = os.Stderr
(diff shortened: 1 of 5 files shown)


pkg/steps/multi_stage/gen.go
```diff
@@ -247,6 +248,9 @@ func addSecretWrapper(pod *coreapi.Pod, vpnConf *vpnConf) {
 	container := &pod.Spec.Containers[0]
 	args := container.Args
 	container.Args = make([]string, 0)
+	if isObserver {
+		container.Args = append(container.Args, "--ro-kubeconfig")
```
droslean, line 252: This is wrong here. It can be removed.

pkg/steps/multi_stage/gen.go
```diff
@@ -41,7 +41,7 @@ func (s *multiStageTestStep) generateObservers(
 			Resources: observer.Resources,
 		})
 	}
-	pods, _, err := s.generatePods(adapted, nil, secretVolumes, secretVolumeMounts)
+	pods, _, err := s.generatePods(adapted, nil, secretVolumes, secretVolumeMounts, true)
```
droslean, line 44: Since this is hardcoded as `true`, there is no need to pass it through arguments.

cmd/entrypoint-wrapper/main.go
```diff
@@ -72,13 +72,15 @@ type options struct {
...
 	client         coreclientset.SecretInterface
+	roKubeconfig   bool
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
 	opt := &options{}
 	flag.BoolVar(&opt.dry, "dry-run", false, "Print the secret instead of creating it")
 	flag.StringVar(&opt.waitPath, "wait-for-file", "", "Wait for a file to appear at this path before starting the program")
 	flag.StringVar(&opt.waitTimeoutStr, "wait-timeout", "", "Used with --wait-for-file, maximum wait time before starting the program")
+	flag.BoolVar(&opt.roKubeconfig, "ro-kubeconfig", false, "If set do not copy kubeconfig to a writable location")
```
droslean, line 83: Can we have a better name for the flag? Like `--read-only-kubeconfig`

cmd/entrypoint-wrapper/main.go
```diff
@@ -129,7 +131,7 @@ func (o *options) run() error {
 	var errs []error
 	ctx, cancel := context.WithCancel(context.Background())
 	go uploadKubeconfig(ctx, o.client, o.name, o.dstPath, o.dry)
-	if err := execCmd(o.cmd); err != nil {
+	if err := execCmd(o.cmd, o.roKubeconfig); err != nil {
```
droslean, line 134: nit: we can make this an options method function instead of passing the arg.

cmd/entrypoint-wrapper/main.go
jmguzik: +1

pkg/steps/multi_stage/gen.go
droslean: Maybe it is better to create something like `struct PodOptions{ IsObserver bool}` as an optional argument, and you won't have to change the rest of the non-affected code.

danilo-gemoli: It sounds like a nice idea

cmd/entrypoint-wrapper/main.go
danilo-gemoli: Sure

pkg/steps/multi_stage/gen.go
danilo-gemoli: `entrypoint-wrapper` binary exists in both `.spec.initContainers[1]` and `spec.containers[0]` stanzas, but as an initContainer it is just used as a copy command. It does the real job when the test run, acting as a proxy and doing some hacky stuff like:
- manage kubeconfig
- map `$SHARED_DIR` to its secret
- forward `SIGTERM` and `SIGINT`
- ...

danilo-gemoli:
Closed, it has been superseded by #2959

PR #2939, opened 2022-08-10
waitForBuild: use watch instead of get

hongkailiu:
https://issues.redhat.com/browse/DPTP-3045


bbguimaraes:
Worked fine in my experiments, but it looks like we have more problems in `pod-scaler` E2E tests…

/test e2e

stevekuznetsov (approved):

PR #2945, opened 2022-08-11
ci-operator: handle promotion with no refs

stevekuznetsov:

### DIFF
--- a/pkg/steps/release/promote.go
+++ b/pkg/steps/release/promote.go
@@ -56,7 +56,13 @@ func (s *promotionStep) Run(ctx context.Context) error {
 }
 
 func (s *promotionStep) run(ctx context.Context) error {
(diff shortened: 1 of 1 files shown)


stevekuznetsov:
https://issues.redhat.com/browse/TRT-455

stbenjam:
I would sleep better at night if these had a nil check too instead of assuming callers are doing the right thing:

https://github.com/openshift/ci-tools/blob/master/pkg/steps/release/promote.go#L297-L309

stevekuznetsov:
The mutable options container is passed to each option closure inside of the call-site. It's not of the concern of the caller, and it's entirely self-contained. `nil` is an entirely invalid value there, and if the `options` are `nil`, the program should crash, as it's a programmer error.

PR #2947, opened 2022-08-16
entrypoint-wrapper: add renaming kubeconfig

jmguzik:

### DIFF
--- a/cmd/entrypoint-wrapper/main.go
+++ b/cmd/entrypoint-wrapper/main.go
@@ -13,6 +13,7 @@ import (
 	"os/signal"
 	"path"
 	"path/filepath"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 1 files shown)


jmguzik:
/hold
not ready

cmd/entrypoint-wrapper/main.go
```diff
@@ -326,6 +327,11 @@ func manageKubeconfig(proc *exec.Cmd) error {
 				if err := source.Close(); err != nil {
 					return true, err
 				}
+				newPath := strings.Trim(writableCopy.Name(), tempPrefix)
+				if err := os.Rename(writableCopy.Name(), newPath); err != nil {
+					return true, err
+				}
+				proc.Env = append(proc.Env, fmt.Sprintf("KUBECONFIG=%s", newPath))
```
jmguzik, line 334: It seems that would require mutex set on proc and makes running rename inside a go func() a little bit complicated.

jmguzik: I think we would like to avoid protecting fields right?

cmd/entrypoint-wrapper/main.go
```diff
@@ -298,12 +299,12 @@ func manageHome(proc *exec.Cmd) {
...
 // but changes are not propagated to subsequent steps to limit the amount of possible mistakes
 func manageKubeconfig(proc *exec.Cmd) error {
+	tempPrefix := "temp-"
 	if original, set := os.LookupEnv("KUBECONFIG"); set {
-		writableCopy, err := ioutil.TempFile("", "kubeconfig-*")
+		writableCopy, err := ioutil.TempFile("", tempPrefix+"kubeconfig-*")
 		if err != nil {
 			return fmt.Errorf("could not create unique, writeable $KUBECONFIG copy: %w", err)
 		}
-		proc.Env = append(proc.Env, fmt.Sprintf("KUBECONFIG=%s", writableCopy.Name()))
```
danilo-gemoli, line 306: I'm a little bit perplexed on unsetting `$KUBECONFIG` at the beginning, but I may lack of context. What if tests rely on having an even empty `$KUBECONFIG` as soon as they start? Let's suppose a test does what follows:
- it starts and check if `$KUBECONFIG` exists (otherwise failing), without caring if it is actually empty and/or writable and/or valid, etc. 
- it does its stuff
- then it uses the `$KUBECONFIG` and at this moment it is (hopefully) non-empty and valid  

Such a test would fail if we introduce those changes. Again, I'm not even sure that situation would be possible but it's worth discussing.

danilo-gemoli: @jmguzik just ignore my comment for now as it may not have any sense at all

jmguzik: it is being discussed on our slack channel

jmguzik:
@danilo-gemoli @bbguimaraes PTAL.
I think it's ready. Integration tests are passing, kubeconfig copy is writable.

PR #2951, opened 2022-08-17
Add metrics to pod-scaler-admission

psalajova:
This adds metrics for pod-scaler admission to record an error to prometheus, in case pod-scaler determines that a job needs more than 10 times the memory configured in the spec.

for the first 2 subtasks of https://issues.redhat.com/browse/DPTP-2929

/cc smg247

### DIFF
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -5,11 +5,11 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	"github.com/prometheus/client_golang/prometheus"
 	"net/http"
 	"sort"
 	"strconv"
 
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/sirupsen/logrus"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


cmd/pod-scaler/admission.go
```diff
@@ -217,6 +220,9 @@ func useOursIfLarger(allOfOurs, allOfTheirs *corev1.ResourceRequirements, logger
 			if our.Cmp(their) == 1 {
 				logger.Debugf("determined %s %s of %s to be larger than %s configured", field, pair.resource, our.String(), their.String())
 				(*pair.theirs)[field] = our
+				if our.Value() > (their.Value() * 10) {
+					metrics.RecordError("actual memory 10x more than configured amount", promMetrics.ErrorRate)
```
smg247, line 224: we need to be sure to provide information about which workload this is for. You will have to pass that info into this function as it looks like it is not in the scope of the function currently.

cmd/pod-scaler/admission.go
```diff
@@ -217,11 +224,32 @@ func useOursIfLarger(allOfOurs, allOfTheirs *corev1.ResourceRequirements, logger
 			if our.Cmp(their) == 1 {
 				logger.Debugf("determined %s %s of %s to be larger than %s configured", field, pair.resource, our.String(), their.String())
 				(*pair.theirs)[field] = our
+				recordPodAdmitted(containerName, admittedPodsMetric)
```
smg247, line 227: can this move out of this loop so that it only happens once per pod rather than per metric? that also means it will happen regardless of our determined amount being larger (which is a good thing).

PR #2959, opened 2022-08-22
Observer pods: manage kubeconfig and shared_dir features on entrypoint-wrapper

danilo-gemoli:
This work is preparatory for a future e2e test (an additional PR will follow).

`entrypoint-wrapper` has been modified to opt-out the following features:
- make a `rw` copy of `$KUBECONFIG`
- synchronize `$SHARED_DIR` back to the shared secret
- upload a kubeconfig

`ci-operator` has been modified to configure observer-pods such that:
- they are not allowed to update the shared secret
- no kubeconfig will ever be uploaded
- a kubeconfig will exist in `$KUBECONFIG` as soon as possible

/cc @jmguzik @bbguimaraes 
/hold

### DIFF
--- a/cmd/entrypoint-wrapper/main.go
+++ b/cmd/entrypoint-wrapper/main.go
@@ -41,6 +41,12 @@ var (
 	encoder runtime.Encoder
 )
 
+const (
+	manageKubeconfigMode = "manage-kubeconfig"
+	skipKubeconfigMode   = "skip-kubeconfig"
+	observerMode         = "observer"
+)
+
 func init() {
 	utilruntime.Must(coreapi.AddToScheme(coreScheme))
 	encoder = codecFactory.LegacyCodec(coreapi.SchemeGroupVersion)
@@ -53,7 +59,7 @@ func main() {
 		logrus.WithError(err).Fatal("failed to parse flag set")
 	}
 	opt.cmd = flagSet.Args()
-	if err := opt.complete(); err != nil {
+	if err := opt.complete(flagSet); err != nil {
 		fmt.Fprintln(os.Stderr, "error:", err)
 		os.Exit(1)
 	}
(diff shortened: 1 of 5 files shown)


cmd/entrypoint-wrapper/main.go
```diff
@@ -41,6 +41,13 @@ var (
 	encoder runtime.Encoder
 )
 
+const (
+	skipKubeconfigFlag   = "skip-kubeconfig"
+	rwKubeconfigFlag     = "readwrite-kubeconfig"
```
jmguzik, line 46: do we need this as it is default right now?

danilo-gemoli: Yes, `rw` kubeconfig is the default but +1 for the catch, it is not needed for observers. I made a commit to set it to false, thanks!

bbguimaraes:
This results in too many cases.  There are only three:

- regular test (do everything)
- test without `kubeconfig` (do nothing)
- observer (copy `kubeconfig` only)

but three binary flags give us eight different states.

In addition, this program's only purpose is to be used by `ci-operator`, so I don't think there's any reason to maintain the `--skip-kubeconfig` flag (yet another to manage, establish precedence, etc.).

cmd/entrypoint-wrapper/main.go
```diff
@@ -64,28 +70,31 @@ func main() {
...
+	client           coreclientset.SecretInterface
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
 	opt := &options{}
 	flag.BoolVar(&opt.dry, "dry-run", false, "Print the secret instead of creating it")
 	flag.StringVar(&opt.waitPath, "wait-for-file", "", "Wait for a file to appear at this path before starting the program")
 	flag.StringVar(&opt.waitTimeoutStr, "wait-timeout", "", "Used with --wait-for-file, maximum wait time before starting the program")
-	flag.BoolVar(&opt.skipKubeconfig, "skip-kubeconfig", false, "Skip managing the $KUBECONFIG copy")
+	flag.StringVar(&opt.mode, "mode", manageKubeconfigMode, fmt.Sprintf("Set how kubeconfig should be managed. Allowed values are: %s, %s or %s", manageKubeconfigMode, skipKubeconfigMode, observerMode))
```
jmguzik, line 93: will that break Steve's feature? Probably he would need to update his configs.

jmguzik: What is the default, if you don't provide anything `manageKubeconfigMode` is set? It should.

danilo-gemoli: It should not break Steve's work as I have translated his `--skip-kubeconfig` flag into `--mode=skip-kubeconfig` with no changes in behaviour: to dig deeper take a look here #2930.
If `--mode` was not specified then the default value is `manageKubeconfigMode` that is `"manage-kubeconfig"`. The third parameter of `flag.StringVar` method is the default value.

jmguzik (approved):

jmguzik:
/hold cancel
to trigger the retester for the night

jmguzik:
\o/

PR #2961, opened 2022-08-22
`check-gh-automation`: check that app is installed in repo

smg247:
We can utilize the newly added (https://github.com/kubernetes/test-infra/pull/27181) `IsAppInstalled` method to determine if our app has been installed in the repo.

For: https://issues.redhat.com/browse/DPTP-3041

### DIFF
--- a/cmd/check-gh-automation/main_test.go
+++ b/cmd/check-gh-automation/main_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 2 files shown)


smg247:
/hold
We are going to need to pull in another change from upstream before this works in all cases.

cmd/check-gh-automation/main_test.go
```diff
@@ -116,14 +140,20 @@ func TestCheckRepos(t *testing.T) {
...
+		},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			failing, err := checkRepos(tc.repos, tc.bots, tc.ignore, client, logrus.NewEntry(logrus.New()))
 			if diff := cmp.Diff(tc.expectedErr, err, testhelper.EquateErrorMessage); diff != "" {
 				t.Fatalf("error doesn't match expected, diff: %s", diff)
 			}
-			if diff := cmp.Diff(tc.expected, failing); diff != "" {
+			if diff := cmp.Diff(tc.expected, failing, cmpopts.EquateEmpty()); diff != "" {
```
hongkailiu, line 156: NIT: If I understand it correctly, I would define the `expected` explicitly instead of ignoring the difference of `[]` and `nil`.

hongkailiu (approved):

cmd/check-gh-automation/main_test.go
smg247: The problem with that is if one of the calls to the gh api fails it returns nil, and otherwise will return an empty list. In practice we don't care at all about this difference, the only place it matters is this test. I didn't want to muck up the actual code just to satisfy this test, so I used this approach.

hongkailiu: I was suggesting to define the following in the tests:

> expected:    []string{},

PR #2967, opened 2022-08-23
Observer pods: e2e test

danilo-gemoli:
An e2e test to make sure the observer pod run properly and it also receives a kubeconfig.
This PR makes sense only when #2959 will be merged.

/cc @jmguzik @bbguimaraes 

This needs a rebase
/hold


test/e2e/observer/e2e_test.go
```diff
@@ -0,0 +1,87 @@
...
+		t.Fatalf("failed to create dummy secret dir: %v", err)
+	}
+	if err := ioutil.WriteFile(filepath.Join(clusterProfileDir, "data"), []byte("nothing"), 0644); err != nil {
+		t.Fatalf("failed to create dummy secret data: %v", err)
+	}
+	var testCases = []struct {
+		name    string
+		args    []string
+		env     []string
+		success bool
```
bbguimaraes, line 35: Will there be a failure case?

test/e2e/observer/e2e_test.go
```diff
@@ -0,0 +1,87 @@
...
+				"--target=with-observer",
+				"--secret-dir=" + clusterProfileDir,
+			},
+			env:     []string{defaultJobSpec},
+			success: true,
+			output: []string{
+				`Running step with-observer-observer`,
+				`Running step with-observer-create-kubeconfig`,
+				`Running step with-observer-check-shared-dir`,
+				`Step with-observer-observer succeeded after`,
```
bbguimaraes, line 51: This is a nice side channel.  Can we add cases where the observers fail?

test/e2e/observer/registry/observer/observer-commands.sh
```diff
@@ -7,22 +7,21 @@ set -o xtrace
...
-while true; do
-    if [[ -f "${KUBECONFIG}" ]]; then
-      echo "\$KUBECONFIG exists"
-      break
-    fi
-    echo "\$KUBECONFIG does not exist, waiting..."
-    sleep 1
+echo "do-not-upload-me" >"${SHARED_DIR}/intruder"
+
+for (( i=1; i<=300; i++ )); do
```
jmguzik, line 17: Is it really needed to wait 300s?

jmguzik (approved):
I can't see anything wrong, the tests seem to cover basic cases. Bash also looks OK IMO.
Good work Danilo, now let's prepare a demo for TRT.
LGTM.

test/e2e/observer/registry/observer/observer-commands.sh
danilo-gemoli: I don't think so, it's just an hard limit. After a lot of run I noticed kubelet can take up to ~2mins before propagating the changes from the secret to the pods.

jmguzik:
\o/

PR #2970, opened 2022-08-23
`check-gh-automation`: allow repo to be passed in and create local dev script

smg247:
Now that this tool is getting more complicated, it is nice to have an easy way to run it locally. By adding a new option to supply the repo(s) to be checked we can achieve that.

As a follow up, I want to create a make target in `openshift/release` that allows us to simply check a repo on command. This is useful when we think some changes to the automation access may be causing problems.

### DIFF
--- a/cmd/check-gh-automation/README.md
+++ b/cmd/check-gh-automation/README.md
@@ -1,5 +1,6 @@
 # Check GH Automation
 A tool to check that our bots (`openshift-merge-robot` and `openshift-ci-robot`) have access to repositories that have CI configured.
+It also checks that the app used to run it is installed in the repositories.
 This can be run in multiple modes:
 
 ## Pass Prow Config Options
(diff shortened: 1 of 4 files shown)


cmd/check-gh-automation/README.md
```diff
@@ -1,5 +1,6 @@
 # Check GH Automation
 A tool to check that our bots (`openshift-merge-robot` and `openshift-ci-robot`) have access to repositories that have CI configured.
+It also checks that the app used to run it is installed in the repositories.
```
hongkailiu, line 3: English is so hard to me: Should we use `if` instead of `that` here?

hack/local-check-gh-automation.sh
```diff
@@ -0,0 +1,26 @@
...
+  os::util::describe_return_code "${return_code}"
+  exit "${return_code}"
+}
+trap "cleanup" EXIT
+
+data="${BASETMPDIR}/data"
+mkdir -p "${data}"
+
+function OC() {
+	oc --context app.ci --namespace ci --as system:admin "$@"
```
hongkailiu, line 17: `--as system:admin`: I do not see the oc-cmd that we need it.

cmd/check-gh-automation/README.md
smg247: I think they are both okay. They both sound natural to me, and I found: https://forum.wordreference.com/threads/check-if-check-that.506018/

hack/local-check-gh-automation.sh
smg247: I assumed that we needed that we needed admin perms to grab those secrets, but it looks like that is unnecessary. I'll correct that.

hongkailiu (approved):

PR #2972, opened 2022-08-24
Change base image to internal unpriv nginx

jmguzik:


images/gh-pull-through-cache/Dockerfile
```diff
@@ -1,5 +1,4 @@
-FROM quay.io/centos/centos:stream9
```
jmguzik, line 1: according to this: https://medium.com/kocsistem/how-to-run-nginx-for-root-non-root-5ceb13db6d41
I will have some problems doing this image on my own. Instead I would like to use internal which exists and just switch config. I could use config map but deployment must be per cluster so it would be easier to just have a ready image..

PR #2976, opened 2022-08-25
e2e: observer pods

jmguzik:


danilo-gemoli:
This is going to be replaced by #2967

jmguzik:
ack

PR #2978, opened 2022-08-26
`prowgen`: expose option to disable rehearsals for specific tests

smg247:
We need an option for test maintainers to not have rehearsals for specific tests when using `prowgen`.

Coming next: An option in the repo's `.ci-operator.prowgen` config to disable rehearsals for **all** tests.

For: https://issues.redhat.com/browse/DPTP-3091

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -27,13 +27,13 @@ func IsPromotionJob(jobLabels map[string]string) bool {
 // ReleaseBuildConfiguration describes how release
 // artifacts are built from a repository of source
 // code. The configuration is made up of two parts:
-//  - minimal fields that allow the user to buy into
-//    our normal conventions without worrying about
-//    how the pipeline flows. Use these preferentially
-//    for new projects with simple/conventional build
-//    configurations.
-//  - raw steps that can be used to create custom and
-//    fine-grained build flows
+//   - minimal fields that allow the user to buy into
+//     our normal conventions without worrying about
+//     how the pipeline flows. Use these preferentially
+//     for new projects with simple/conventional build
+//     configurations.
+//   - raw steps that can be used to create custom and
+//     fine-grained build flows
 type ReleaseBuildConfiguration struct {
 	Metadata Metadata `json:"zz_generated_metadata"`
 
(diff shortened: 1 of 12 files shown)


pkg/api/types.go
```diff
@@ -689,6 +689,10 @@ type TestStepConfiguration struct {
 	// Timeout overrides maximum prowjob duration
 	Timeout *prowv1.Duration `json:"timeout,omitempty"`
 
+	// DisableRehearsal will exclude the "can-be-rehearsed" label from the pj spec.
+	// Which will mean the test will not be picked up for rehearsals. Not valid on postsubmits.
+	DisableRehearsal bool `json:"disable_rehearsal,omitempty"`
```
droslean, line 694: I think it would be better not to interact with `ci-operator` at all. We can use the prowgen configuration file, to allow users to disable the rehearsal in specific jobs.

droslean: Check: https://github.com/openshift/ci-tools/blob/master/pkg/config/load.go#L21-L34

smg247: That's what I am doing in https://github.com/openshift/ci-tools/pull/2979, but I liked the idea of making them configurable at the job level. With the prowgen config it's all or nothing

droslean: We can configure them at the job level in the prowgen configuration too. Something like
```yaml
exclude_rehearsals:
- test-1
- test-2
- test-3
```

smg247: Right, that's what I would like to avoid doing though. That would leave us with test names in two separate files. We would have to keep the files in sync and create an entirely new type of validation.

Can you explain the downsides to adding a new Boolean field where I have done it here?

droslean: My concern is that this is a temporary workaround until we change the rehearsal tool to a prow external plugin. Also, the rehearsals have nothing to do with ci-operator, so I think it would be better not to include this logic in the ci-operator config. 
Prowgen is responsible for generating the job and anything related to it should be included in prowgen config only. 

Also, following this logic, we should remove completely the prowgen config and include all of those options in ci-operator config, but this will mess things up. 

I believe not messing up with ci-operator config is the best option here, since this is a temporary workaround.

smg247:
Closing this one and will be moving this functionality to the prowgen config and doing it all in https://github.com/openshift/ci-tools/pull/2979

PR #2979, opened 2022-08-26
`prowgen`: disable rehearsals via the .config.prowgen file

smg247:
This will allow job maintainers to disable rehearsals on **all** jobs for an org or repo by using the `.config.prowgen` file. It also allows for individual tests rehearsals to be disabled via a list of `disabled_rehearsals`.

I considered adding validation that each of the tests in the disabled rehearsals list actually matched a test name, but doing so would be fairly complex and difficult to put in place. Since the job maintainers will clearly be able to see that the label has still been applied if they were to make a typo in the entry in the list, I think it is okay to forgo this validation.

Coming up next: Information about using these new configs will be added to `ci-docs`

For: https://issues.redhat.com/browse/DPTP-3091

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -136,7 +136,7 @@ func (o *options) generateJobsToDir(subDir string, prowConfig map[string]*config
 func generateJobs(resolver registry.Resolver, cache map[string]*config.Prowgen, output map[string]*prowconfig.JobConfig) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 	return func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 		orgRepo := fmt.Sprintf("%s/%s", info.Org, info.Repo)
-		pInfo := &prowgen.ProwgenInfo{Metadata: info.Metadata, Config: config.Prowgen{Private: false, Expose: false}}
+		pInfo := &prowgen.ProwgenInfo{Metadata: info.Metadata, Config: config.Prowgen{Private: false, Expose: false, DisableAllRehearsals: false}}
 		var ok bool
 		var err error
 		var orgConfig, repoConfig *config.Prowgen
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -136,7 +136,7 @@ func (o *options) generateJobsToDir(subDir string, prowConfig map[string]*config
 func generateJobs(resolver registry.Resolver, cache map[string]*config.Prowgen, output map[string]*prowconfig.JobConfig) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 	return func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 		orgRepo := fmt.Sprintf("%s/%s", info.Org, info.Repo)
-		pInfo := &prowgen.ProwgenInfo{Metadata: info.Metadata, Config: config.Prowgen{Private: false, Expose: false, DisableAllRehearsals: false}}
+		pInfo := &prowgen.ProwgenInfo{Metadata: info.Metadata, Config: config.Prowgen{Private: false, Expose: false}}
 		var ok bool
 		var err error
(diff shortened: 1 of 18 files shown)


pkg/prowgen/prowgen.go
```diff
@@ -137,9 +139,9 @@ func testContainsLease(test *cioperatorapi.TestStepConfiguration) bool {
 	return len(api.LeasesForTest(test.MultiStageTestConfigurationLiteral)) > 0
 }
 
-func generatePresubmitForTest(jobBaseBuilder *prowJobBaseBuilder, name string, info *ProwgenInfo, runIfChanged, skipIfOnlyChanged string, optional bool) *prowconfig.Presubmit {
+func generatePresubmitForTest(jobBaseBuilder *prowJobBaseBuilder, name string, info *ProwgenInfo, runIfChanged, skipIfOnlyChanged string, optional, disableRehearsal bool) *prowconfig.Presubmit {
```
droslean, line 142: Can we instead add an optional argument that holds any options including the rehearse boolean? This way you don't have to update all the references of this function, and also allows the code to be more extensible in the future if we want to add more options.

cmd/ci-operator-prowgen/main.go
```diff
@@ -136,7 +136,7 @@ func (o *options) generateJobsToDir(subDir string, prowConfig map[string]*config
 func generateJobs(resolver registry.Resolver, cache map[string]*config.Prowgen, output map[string]*prowconfig.JobConfig) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 	return func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
 		orgRepo := fmt.Sprintf("%s/%s", info.Org, info.Repo)
-		pInfo := &prowgen.ProwgenInfo{Metadata: info.Metadata, Config: config.Prowgen{Private: false, Expose: false}}
+		pInfo := &prowgen.ProwgenInfo{Metadata: info.Metadata, Config: config.Prowgen{Private: false, Expose: false, DisableAllRehearsals: false}}
```
droslean, line 139: nit: Those values will already be false. Having them makes the code more readable though. Your call.

pkg/prowgen/prowgen.go
```diff
@@ -183,10 +207,28 @@ func hashDailyCron(job string) string {
...
+
+type GeneratePeriodicOption func(options *GeneratePeriodicOptions)
+
+func FromConfigSpec(configSpec *cioperatorapi.ReleaseBuildConfiguration) GeneratePeriodicOption {
+	return func(options *GeneratePeriodicOptions) {
+		options.pathAlias = configSpec.CanonicalGoRepository
+	}
+}
+
+func GeneratePeriodicForTest(jobBaseBuilder *prowJobBaseBuilder, info *ProwgenInfo, cron string, interval string, options ...GeneratePeriodicOption) *prowconfig.Periodic {
```
droslean, line 224: `cron` and `interval` should be in the options as well.

pkg/config/load.go
```diff
@@ -31,6 +31,12 @@ type Prowgen struct {
 	// are private.
 	// This field has no effect if private is not set.
 	Expose bool `json:"expose,omitempty"`
+	// DisableAllRehearsals indicates that all jobs will not have their "can-be-rehearsed" label set
+	// and therefore will not be picked up for rehearsals.
+	DisableAllRehearsals bool `json:"disable_all_rehearsals,omitempty"`
```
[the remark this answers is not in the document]
smg247: I don't mind the idea, but I think something like the following make more sense:
```
rehearsals:
  disable_all: false
  disabled_rehearsals:
  - test-1
  - test-2
  - test-3
```
If `disable_all` is true we would ignore the individual ones, otherwise only those would be disabled. I don't want people to have to call out the ones they want enabled.

pkg/prowgen/prowgen.go
smg247: My thinking there was that one of them must be provided, so they aren't really optional. I can move them though if you think thats best.

pkg/config/load.go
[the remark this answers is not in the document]
droslean: Nice! That makes more sense.

droslean (approved):

PR #2985, opened 2022-08-31
promoted-image-governor: don't delete images tagged by sha

stevekuznetsov:

### DIFF
--- a/cmd/promoted-image-governor/main.go
+++ b/cmd/promoted-image-governor/main.go
@@ -423,11 +423,19 @@ func main() {
 		logrus.WithError(err).Fatal("failed to determine absolute CI Operator configuration path")
 	}
 	var promotedTags []api.ImageStreamTagReference
+	var ignoredCommitTags []*regexp.Regexp
 	if err := config.OperateOnCIOperatorConfigDir(abs, func(cfg *api.ReleaseBuildConfiguration, metadata *config.Info) error {
 		for _, isTagRef := range release.PromotedTags(cfg) {
 			promotedTags = append(promotedTags, isTagRef)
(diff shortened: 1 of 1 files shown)


cmd/promoted-image-governor/main.go
```diff
@@ -423,12 +423,20 @@ func main() {
...
 	var promotedTags []api.ImageStreamTagReference
+	var ignoredCommitTags []*regexp.Regexp
 	if err := config.OperateOnCIOperatorConfigDir(abs, func(cfg *api.ReleaseBuildConfiguration, metadata *config.Info) error {
 		for _, isTagRef := range release.PromotedTags(cfg) {
 			promotedTags = append(promotedTags, isTagRef)
 			if _, ok := opts.explains[isTagRef]; ok {
 				opts.explains[isTagRef] = cfg.Metadata.AsString()
 			}
+			if cfg.PromotionConfiguration.TagByCommit {
+				ignoreRegex, err := regexp.Compile(fmt.Sprintf("%s/%s:[0-9a-f]{5,40}", isTagRef.Namespace, isTagRef.Name))
```
hongkailiu, line 434: I guess it would ignore any tags satisfying the regex but that is not a real commit.
Basically, the governor just stopped governing on the stream with `TagByCommit`. 😄 
Staled images are not a concept there.

hongkailiu (approved):

cmd/promoted-image-governor/main.go
stevekuznetsov: I thought about this ... we could make the sha matcher longer I guess? But yes if this becomes an issue we can reconsider

PR #2992, opened 2022-09-08
Result-aggregator: handle pod-scaler admission requests

psalajova:
Adds functionality to result-aggregator to handle requests from pod-scaler admission service and record them to Prometheus.

/cc smg247

### DIFF
--- a/cmd/result-aggregator/main.go
+++ b/cmd/result-aggregator/main.go
@@ -37,7 +37,7 @@ var (
 			Name: "pod_scaler_admission_error_rate",
 			Help: "number of errors, sorted by label/type",
 		},
-		[]string{"test_name", "configured_memory", "actual_memory"},
+		[]string{"workload_name", "configured_memory", "determined_memory"},
 	)
 )
 
@@ -53,9 +53,9 @@ type options struct {
 }
 
 type podScalerRequest struct {
-	testName         string
+	workloadName     string
 	configuredMemory string
-	actualMemory     string
+	determinedMemory string
 }
 
 func gatherOptions() (options, error) {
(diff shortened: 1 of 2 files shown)


cmd/result-aggregator/main.go
```diff
@@ -45,6 +52,12 @@ type options struct {
 	passwdFile  string
 }
 
+type podScalerRequest struct {
+	testName         string
+	configuredMemory string
+	actualMemory     string
```
smg247, line 58: can we change the name of this to `determinedMemory`?

cmd/result-aggregator/main.go
```diff
@@ -45,6 +52,12 @@ type options struct {
 	passwdFile  string
 }
 
+type podScalerRequest struct {
+	testName         string
```
smg247, line 56: can this be called `workloadName`? It is a bit more generic.

cmd/result-aggregator/main.go
```diff
@@ -88,6 +101,19 @@ func validateRequest(request *results.Request) error {
 	return nil
 }
 
+func validatePodScalerRequest(request *podScalerRequest) error {
```
smg247, line 104: We can add a unit test for this function.

smg247 (changes requested):

cmd/result-aggregator/main_test.go
```diff
@@ -148,3 +149,72 @@ func TestLoginHandler(t *testing.T) {
...
+		t.Run(testCase.name, func(t *testing.T) {
+			actual := validatePodScalerRequest(testCase.request)
+			if actual == nil {
+				if testCase.expected != nil {
+					t.Fatalf("got error when test case was ok, error: %v", actual.Error())
+				}
+			} else {
+				if actual.Error() != testCase.expected.Error() {
+					t.Fatalf("error: %v doesn't match expected error: %v", actual.Error(), testCase.expected.Error())
+				}
```
smg247, line 216: You can simplify this by using `cmp.Diff`. See an example here: https://github.com/openshift/ci-tools/blob/6aa22d5527e30fa0c99e19b11064ffb4e94b2e1b/cmd/ci-operator/main_test.go#L1150-L1152

PR #2999, opened 2022-09-12
DPTP-3081 prReminder add held label

psalajova:
When Pr Reminder slack tool lists PRs, this adds additional information if the PR is labeled as held (do-not-merge/hold).

for https://issues.redhat.com/browse/DPTP-3081
/cc smg247

### DIFF
--- a/cmd/pr-reminder/main.go
+++ b/cmd/pr-reminder/main.go
@@ -456,8 +456,5 @@ func messageUser(user user, slackClient slackClient) error {
 
 // getLabelMessage returns a string listing te PR's labels
 func getLabelMessage(labels []string) string {
-	if len(labels) == 0 {
-		return ""
-	}
 	return fmt.Sprintf(":label: labeled: *%v*", strings.Join(labels[:], ", "))
 }
--- a/cmd/pr-reminder/main.go
+++ b/cmd/pr-reminder/main.go
@@ -70,8 +70,12 @@ type config struct {
 	Teams []team `json:"teams"`
 }
 
-// TODO: how to store labels we don't want to filter out?
-const holdLabel = "do-not-merge/hold"
+// getInterestedLabels returns a set of those labels we are interested in when using the PR reminder
+func getInterestedLabels() sets.String {
+	var labels = sets.String{}
+	labels.Insert("do-not-merge/hold")
+	return labels
+}
 
 var orgRepoFormat = regexp.MustCompile(`\w+/\w+`)
 
@@ -340,7 +344,7 @@ func findPrsForUsers(users map[string]user, ghClient prClient) map[string]user {
 						Author:      pr.User.Login,
 						Created:     pr.CreatedAt,
 						LastUpdated: pr.UpdatedAt,
-						Labels:      filterLabels(pr.Labels),
+						Labels:      filterLabels(pr.Labels, getInterestedLabels()),
 					})
 					users[i] = u
 				}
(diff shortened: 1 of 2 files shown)


cmd/pr-reminder/main.go
```diff
@@ -70,6 +70,9 @@ type config struct {
 	Teams []team `json:"teams"`
 }
 
+// TODO: how to store labels we don't want to filter out?
+const holdLabel = "do-not-merge/hold"
```
smg247, line 74: I think a const slice would be good (obviously it would only contain a single element for now)

cmd/pr-reminder/main.go
```diff
@@ -346,6 +351,16 @@ func findPrsForUsers(users map[string]user, ghClient prClient) map[string]user {
 	return users
 }
 
+func filterLabels(labels []github.Label) []string {
```
smg247, line 354: this function can take an `interestedLabels sets.String` parameter so that it can easily be checked with `interestedLabels.Has(label.Name)`

cmd/pr-reminder/main.go
```diff
@@ -424,3 +443,12 @@ func messageUser(user user, slackClient slackClient) error {
...
 	return kerrors.NewAggregate(errors)
 }
+
+// appendLabels checks the PR's labels and returns a message if PR is held,
+// or empty string if the PR is not held.
+func appendLabels(labels []string) string {
+	if len(labels) == 0 {
+		return " " //if PR has no labels, this adds nothing to the slack message
+	}
+	return fmt.Sprintf(":warning: This PR is labeled: *%v*", strings.Join(labels[:], ", "))
```
smg247, line 453: I like the idea of the `warning` emoji, but ultimately we might want to include other labels here and it may be a bit too heavy. It appears there is a `label` emoji, so lets try that.
```suggestion
	return fmt.Sprintf(":label: labeled: *%v*", strings.Join(labels[:], ", "))
```

cmd/pr-reminder/main.go
```diff
@@ -408,6 +423,10 @@ func messageUser(user user, slackClient slackClient) error {
 						Type: slack.MarkdownType,
 						Text: pr.createdUpdatedMessage(),
 					},
+					&slack.TextBlockObject{
+						Type: slack.MarkdownType,
+						Text: appendLabels(pr.Labels),
+					},
```
smg247, line 429: We shouldn't be adding this block at all if there are no labels to include.

cmd/pr-reminder/main.go
```diff
@@ -424,3 +453,11 @@ func messageUser(user user, slackClient slackClient) error {
 
 	return kerrors.NewAggregate(errors)
 }
+
+// getLabelMessage returns a string listing te PR's labels
+func getLabelMessage(labels []string) string {
+	if len(labels) == 0 {
+		return ""
+	}
```
smg247, line 461: you don't need this anymore, this function is only called when labels exist.

cmd/pr-reminder/main.go
```diff
@@ -346,6 +355,19 @@ func findPrsForUsers(users map[string]user, ghClient prClient) map[string]user {
 	return users
 }
 
+// filterLabels filters out those labels from the PR we are not interested in
+// and returns only those that are included in the interestedLabels set
+func filterLabels(labels []github.Label, interestedLabels sets.String) []string {
```
smg247, line 360: I think we should have a unit test for this function.

smg247 (changes requested):

cmd/pr-reminder/main.go
psalajova: yes good catch

PR #3002, opened 2022-09-13
Give warning when no jobs found for testcase analyzer

DennisPeriquet:
Re: [TRT-407](https://issues.redhat.com//browse/TRT-407)

This PR makes it so that if there is no related job data, we don't fatally abort; this makes it so that we can still analyze subsequent tests.

The "some test checker failed,  see above for details" condition occurs when the minimum number of jobs is not met like this:

```
Message: required minimum successful count 2, got 1
name: 'install should succeed: overall'
testsuitename: cluster install
summary: 'Total job runs: 1, passes: 1, failures: 0, skips 0'
passes:
- jobrunid: "1569348024021291008"
  humanurl: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-vsphere-sdn/1569348024021291008
  gcsartifacturl: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-vsphere-sdn/1569348024021291008/artifacts
failures: []
skips: []

WARN[0007] Unable to perform test analysis               error="some test checker failed,  see above for details"
```

This PR also makes it so that condition is no longer a fatal error.

I also removed jobs that lack job data in [release/PR32225](https://github.com/openshift/release/pull/32225) to avoid errors in the log.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
@@ -2,6 +2,7 @@ package jobrunaggregatorlib
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"path/filepath"
@@ -13,6 +14,9 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 	"github.com/openshift/ci-tools/pkg/junit"
 )
+
+var ErrorNoRelatedJobs = errors.New("found no related jobRuns")
+var ErrorTestCheckerFailed = errors.New("some test checker failed,  see above for details")
 
 type JobRunGetter interface {
 	// GetRelatedJobRuns gets all related job runs for analysis
@@ -63,7 +67,7 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
 		}
 
 		if len(relatedJobRuns) == 0 {
-			return finishedJobRuns, unfinishedJobRuns, finishedJobRunNames, unfinishedJobRunNames, fmt.Errorf("found no related jobRuns")
+			return finishedJobRuns, unfinishedJobRuns, finishedJobRunNames, unfinishedJobRunNames, ErrorNoRelatedJobs
 		}
 
 		for i := range relatedJobRuns {
--- a/pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
@@ -481,7 +481,7 @@ func (o *JobRunTestCaseAnalyzerOptions) Run(ctx context.Context) error {
 		return err
 	}
 	if testSuite.NumFailed > 0 {
-		return fmt.Errorf("some test checker failed,  see above for details")
+		return jobrunaggregatorlib.ErrorTestCheckerFailed
 	}
 	return nil
 }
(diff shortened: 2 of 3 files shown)


DennisPeriquet:
CI probably failing due to https://github.com/openshift/ci-tools/pull/3005

But I'm running the test locally and found an error -- investigating ...

```
waited long enough. Ready or not, here I come. (readyOrNot=2022-09-12 19:41:39 +0000 UTC now=2022-09-13 14:21:23.101559843 -0400 EDT m=+5.329542897)
found 1 finished jobRuns: periodic-ci-openshift-release-master-nightly-4.12-e2e-vsphere-sdn1569348024021291008 and 0 unfinished jobRuns: 
Test Failed! suite=[root  payload-cross-jobs  minimum-required-passes-checker  cluster install], testCase=test 'install should succeed: overall' has required number of successful passes across payload jobs for plaftorm:vsphere network:sdn infrastructure:ipi
Message: required minimum successful count 2, got 1
name: 'install should succeed: overall'
testsuitename: cluster install
summary: 'Total job runs: 1, passes: 1, failures: 0, skips 0'
passes:
- jobrunid: "1569348024021291008"
  humanurl: https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-vsphere-sdn/1569348024021291008
  gcsartifacturl: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/logs/periodic-ci-openshift-release-master-nightly-4.12-e2e-vsphere-sdn/1569348024021291008/artifacts
failures: []
skips: []


FATA[0007] Command failed                                error="some test checker failed,  see above for details"
```

@xueqzhan, it seems even with my changes, the condition above will also fatally exit.  I'm trying to understand why it needs to be fatal...

xueqzhan:
> @xueqzhan, it seems even with my changes, the condition above will also fatally exit. I'm trying to understand why it needs to be fatal...

I believe aggregator exit with non-zero code when the overall job fails. We should not stop the script because of that.

DennisPeriquet:
@xueqzhan that makes sense; I added a change to allow that condition to not be fatal.

xueqzhan:
I wonder if adding a junit makes more sense here. When aggregator cannot find more than three jobs, does it fail with a junit? If so, I think a similar junit will be helpful.

DennisPeriquet:
I have never seen an aggregator job that found only 3 jobs.

xueqzhan:
Do we still need this change if your script is not ending on exit errors?

DennisPeriquet:
> Do we still need this change if your script is not ending on exit errors?

I believe yes.  Without the changes in this PR, there will still be cases when the "some test checker failed,  see above for details" condition will be fatal.

xueqzhan:
> > Do we still need this change if your script is not ending on exit errors?
> 
> I believe yes. Without the changes in this PR, there will still be cases when the "some test checker failed, see above for details" condition will be fatal.

Maybe I am missing something. What is the problem of this being fatal? There was an error aggregating jobs. Any error, including no jobs, can be treated as fatal, as long as your script is dealing with it properly. No?

DennisPeriquet:
I'm ok with `job-run-aggregator analyze-job-runs` exiting with an error but if it exits with a fatal error, it exits the calling script esp. since we use the `set -o errexit`.  I like having `set -o errexit` because it catches other errors that are hard to find and should fatally abort.

xueqzhan:
> I'm ok with `job-run-aggregator analyze-job-runs` exiting with an error but if it exits with a fatal error, it exits the calling script esp. since we use the `set -o errexit`. I like having `set -o errexit` because it catches other errors that are hard to find and should fatally abort.

I overlooked. I thought your script will go on to the next variant even if there is a fatal error analyzing previous variant. I think that should be our requirement. If you want to exit to catch other errors, you can disable errexit right before calling aggregator and reenable after.

DennisPeriquet:
no longer relevant

PR #3007, opened 2022-09-14
DPTP-3106 Create PodScalerReporter

psalajova:
This adds a PodScalerReporter as a way for sending http requests from pod-scaler-admission to result-aggregator.

/cc smg247

### DIFF
--- a/pkg/results/report.go
+++ b/pkg/results/report.go
@@ -149,10 +149,11 @@ func (r *reporter) report(request Request) {
 	sendRequest(req, r.client, r.username, r.password)
 }
 
+// PodScalerRequest holds the data from pod-scaler used to report a result to an aggregation server
 type PodScalerRequest struct {
-	workloadName     string
-	configuredMemory string
-	determinedMemory string
+	WorkloadName     string
+	ConfiguredMemory string
+	DeterminedMemory string
 }
 
 type PodScalerReporter struct {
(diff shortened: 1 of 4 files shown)


pkg/results/report.go
```diff
@@ -146,9 +146,63 @@ func (r *reporter) report(request Request) {
 		logrus.Tracef("could not create report request: %v", err)
 		return
 	}
+	sendRequest(req, r.client, r.username, r.password)
+}
+
+type PodScalerRequest struct {
```
smg247, line 152: you need to make sure to remove the `podScalerRequest` from `result-aggregator` and use this one there too.

pkg/results/report.go
```diff
@@ -146,9 +146,63 @@ func (r *reporter) report(request Request) {
...
+		client:   &http.Client{},
+		username: username,
+		password: password,
+		address:  o.address,
+	}, nil
+}
+
+// ReportMemoryConfigurationWarning is used to send the information about memory configuration
+// from pod-scaler-admission to result-aggregator.
+func (r *PodScalerReporter) ReportMemoryConfigurationWarning(workflowName, configuredMemory, determinedMemory string) {
```
smg247, line 180: This method can be unit tested. See: https://github.com/openshift/ci-tools/blob/2ffb68b53d9940bd1afa5ac84f38cb6c001ea50c/pkg/results/report_test.go#L25-L105

smg247 (changes requested):

pkg/results/report_test.go
```diff
@@ -182,3 +182,69 @@ func TestGetUsernameAndPassword(t *testing.T) {
...
+					return
+				}
+
+				data, err := ioutil.ReadAll(request.Body)
+				if err != nil {
+					t.Errorf("failed to read request body: %v", err)
+				}
+
+				if actual := string(data); actual != tc.expected {
+					t.Errorf("expected %v, got %v", tc.expected, actual)
```
smg247, line 234: can we use `cmp.Diff` here? it makes it easier to spot the problems when the test fails.

PR #3008, opened 2022-09-14
[cluster-init] remove the implementation of the registry-pull-credentials-all secret

droslean:
/cc @openshift/test-platform @hongkailiu 

There is literary nothing that is using this secret. I tried to check all pods in all clusters to see if there is any usage of that secret, and also there is no reference in the release repo as well. 

The initial implementation of the `registry-pull-credentials-all` secret was an effort of https://issues.redhat.com/browse/DPTP-1845

Why?
We reached a point where we are in the need to add the quay.io secret in `registry-pull-credentials` which resulted in a duplication of keys in the `-all` secret. Instead of refactoring the code to maintain both secrets and a validated way, this PR instead drops the whole `-all` idea.

follow-up https://github.com/openshift/release/pull/32289


droslean:
Note: `breaking-changes` shouldn't fail.

droslean:
So... That was a part of https://issues.redhat.com/browse/DPTP-1845 which makes much sense. In an effort of maintaining one secret with all credentials.

hongkailiu:
I was expecting the breaking job to fail. But it is not, strange.

hongkailiu:
/lgtm

Left the hold on purpose :)

stevekuznetsov:
Yikes, I barely remember this. As I remember, we had 10x secrets all over the place with the same data, and had issues with inconsistencies between A & B when they should have been identical. I don't care about any particulars for the implementation here - but I imagine the TP team would still benefit from having just one set of pull credentials in once place, instead of 10 sets in 10 places.

PR #3013, opened 2022-09-15
Change minimum number of jobs to 6 instead of 3 for more accurate aggregation

DennisPeriquet:
Re: [TRT-370](https://issues.redhat.com//browse/TRT-370)

The current minimum is only 3 and when there are infra problems, this seems too small because too many job runs don't get to fully run as in [this example](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/aggregated-azure-sdn-upgrade-4.12-minor-release-openshift-release-analysis-aggregator/1570003912218382336).

3 seems too small of a sample for an accurate signal from aggregation; 6 seems more appropriate (allowing up to 4 jobs to fail due to infra issues).


dgoodwin:
YOOOOLOOOOOOOO

/lgtm

We don't know how hard this will hit payloads, but we think it's best to try as the bar seems too low right now. We will iterate if need be.

PR #3014, opened 2022-09-16
Fix missing job names for disruption data uploading.

dgoodwin:
This looks like we've missed accumulating disruption data for a vast
number of 4.11 and 4.12 periodics. Uncovered when we saw that 4.12
single node aws upgrade disruption tests were failing claiming they
could not find any exact or fuzzy match.

[TRT-558](https://issues.redhat.com//browse/TRT-558)


dgoodwin:
This isn't quite as bad as it looks, we were still picking up at least *some* date for most combos, although nowhere near as much as we thought. Full misses can be found by searching: https://search.ci.openshift.org/?search=maxAllowed%3D2.718s&maxAge=48h&context=1&type=bug%2Bissue%2Bjunit&name=4.12&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

They look pretty isolated to mostly single node, ovirt, openstack, and vmware. We were ignoring a lot of data we probably expected to be uploading, but full cache misses were more rare.

None the less this is an important fix, and will likely recover a number of permafailing jobs on disruption tests.

PR #3017, opened 2022-09-19
Add title for test analyzer variant to enhance readability

DennisPeriquet:
Regarding [TRT-562](https://issues.redhat.com//browse/TRT-562)

Apparently, the name of the `job-run-summary.html` file is used as the title for the accordion menu in aggregated jobs.

For test analyzer aggregated jobs, there are several variants and so the title is always the same; this makes it hard to find results for a specific variant.

This PR names the `job-run-summary.html` with more text that is indicative of the variant to make it easier to look at the job results.  There will be no difference for the origin aggregated jobs.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
@@ -92,7 +92,7 @@ func (o *JobRunAggregatorAnalyzerOptions) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	finishedJobsToAggregate, _, finishedJobRunNames, unfinishedJobNames, err := jobrunaggregatorlib.WaitAndGetAllFinishedJobRuns(ctx, timeToStopWaiting, o, o.workingDir)
+	finishedJobsToAggregate, _, finishedJobRunNames, unfinishedJobNames, err := jobrunaggregatorlib.WaitAndGetAllFinishedJobRuns(ctx, timeToStopWaiting, o, o.workingDir, "aggregated")
 	if err != nil {
 		return err
 	}
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/spyglass_summary.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/spyglass_summary.go
@@ -8,10 +8,13 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 )
 
-func htmlForJobRuns(ctx context.Context, finishedJobsToAggregate, unfinishedJobsToAggregate []jobrunaggregatorapi.JobRunInfo) string {
-	html := `<!DOCTYPE html>
+func htmlForJobRuns(ctx context.Context, finishedJobsToAggregate, unfinishedJobsToAggregate []jobrunaggregatorapi.JobRunInfo, jobSummaryInfo string) string {
+	html := fmt.Sprintf(`<!DOCTYPE html>
 <html>
 <head>
+<title>
+job-run-summary for %s
+</title>
 <style>
 a {
 	color: #ff8caa;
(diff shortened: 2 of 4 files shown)


pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -93,8 +94,15 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
...
-		if err := ioutil.WriteFile(filepath.Join(outputDir, "job-run-summary.html"), []byte(summaryHTML), 0644); err != nil {
+		summaryHTML := htmlForJobRuns(ctx, finishedJobRuns, unfinishedJobRuns, variantInfo)
+
+		var jobRunSummaryPrefix string
+		if len(variantInfo) == 0 {
+			jobRunSummaryPrefix = ""
+		} else {
+			jobRunSummaryPrefix = variantInfo + "-"
+		}
+		if err := ioutil.WriteFile(filepath.Join(outputDir, fmt.Sprintf("%sjob-run-summary.html", jobRunSummaryPrefix)), []byte(summaryHTML), 0644); err != nil {
```
stbenjam, line 105: Any reason to use a file name over injecting a <title> tag into the HTML? I personally find it more readable than a file name

pkg/jobrunaggregator/jobrunaggregatorlib/spyglass_summary.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
 )
 
-func htmlForJobRuns(ctx context.Context, finishedJobsToAggregate, unfinishedJobsToAggregate []jobrunaggregatorapi.JobRunInfo) string {
+func htmlForJobRuns(ctx context.Context, finishedJobsToAggregate, unfinishedJobsToAggregate []jobrunaggregatorapi.JobRunInfo, jobSummaryInfo string) string {
 	html := `<!DOCTYPE html>
 <html>
 <head>
```
stbenjam, line 14: The <title> tag I mentioned would be injected here, and spyglass would use that instead

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
DennisPeriquet: I tried to put a `<title>/</title>` inside the `<head></head>` and it only changed the title of the tab and not the accordion menu.

DennisPeriquet: like this:

```
<html>
    <head>
        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
        <title>
            job-run-summary for vsphere-sdn-ipi
        </title>
    </head>
    <body>
        <div id="wrapper">
...
```

stbenjam: I don't think editing it with web developer tools will work since it gets parsed by prow https://github.com/kubernetes/test-infra/pull/25409

DennisPeriquet: But I agree, if we can get the title to work, that'd be a lot simpler.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
```diff
@@ -92,7 +92,7 @@ func (o *JobRunAggregatorAnalyzerOptions) Run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
-	finishedJobsToAggregate, _, finishedJobRunNames, unfinishedJobNames, err := jobrunaggregatorlib.WaitAndGetAllFinishedJobRuns(ctx, timeToStopWaiting, o, o.workingDir)
+	finishedJobsToAggregate, _, finishedJobRunNames, unfinishedJobNames, err := jobrunaggregatorlib.WaitAndGetAllFinishedJobRuns(ctx, timeToStopWaiting, o, o.workingDir, "")
```
stbenjam, line 95:
```suggestion
	finishedJobsToAggregate, _, finishedJobRunNames, unfinishedJobNames, err := jobrunaggregatorlib.WaitAndGetAllFinishedJobRuns(ctx, timeToStopWaiting, o, o.workingDir, "aggregation")
```

so it renders as `job-run-summary for aggregation`

PR #3019, opened 2022-09-20
Pods acquiring imagePullSecrets: race condition mitigation

danilo-gemoli:
This PR is an attempt to mitigate this problem: [DPTP-3134](https://issues.redhat.com/browse/DPTP-3134)

/cc @hongkailiu

FYI: @jupierce

### DIFF
--- a/pkg/util/rbacs.go
+++ b/pkg/util/rbacs.go
@@ -5,12 +5,15 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/sirupsen/logrus"
+
 	coreapi "k8s.io/api/core/v1"
 	rbacapi "k8s.io/api/rbac/v1"
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
+	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/results"
 )
 
@@ -42,6 +45,15 @@ func CreateRBACs(ctx context.Context, sa *coreapi.ServiceAccount, role *rbacapi.
 		return nil
 	}
 
+	hasDockerCfgImagePullSecretSet := func(imagePullSecrets []coreapi.LocalObjectReference) bool {
+		for _, secret := range imagePullSecrets {
+			if api.RegistryPullCredentialsSecret != secret.Name {
+				return true
+			}
+		}
+		return false
+	}
+
 	if err := wait.PollImmediate(retryDuration, timeout, func() (bool, error) {
 		actualSA := &coreapi.ServiceAccount{}
 		if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{
@@ -51,15 +63,14 @@ func CreateRBACs(ctx context.Context, sa *coreapi.ServiceAccount, role *rbacapi.
 			return false, fmt.Errorf("couldn't get service account %s: %w", sa.Name, err)
 		}
 
-		if len(actualSA.ImagePullSecrets) == 0 {
+		if !hasDockerCfgImagePullSecretSet(actualSA.ImagePullSecrets) {
 			return false, nil
 		}
 
 		return true, nil
-	},
-	); err != nil {
-		return results.ForReason("create_dockercfg_secrets").WithError(err).Errorf("timeout while waiting for dockercfg secret creation for service account %q: %v", sa.Name, err)
+	}); err != nil {
+		_ = results.ForReason("create_dockercfg_secrets").WithError(err).Errorf("timeout while waiting for dockercfg secret creation for service account %q: %v", sa.Name, err)
+		logrus.WithError(err).Debugf("timeout while waiting for dockercfg secret creation for service account %q", sa.Name)
 	}
-
 	return nil
 }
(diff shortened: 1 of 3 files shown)


pkg/util/rbacs.go
```diff
@@ -5,61 +5,76 @@ import (
...
 	"github.com/openshift/ci-tools/pkg/results"
 )
 
 // CreateRBACs creates the given service account, role, and role bindings. In addition, it waits until the service account will be updated with an image pull secret.
 // Because the DockerCfgController controller needs some time to create the corresponding secrets we need to wait until this happens, otherwise, the pod that
 // will use this service account will fail to start since there are no credentials to pull any images.
-func CreateRBACs(ctx context.Context, sa *coreapi.ServiceAccount, role *rbacapi.Role, roleBindings []rbacapi.RoleBinding, client ctrlruntimeclient.Client, retryDuration, timeout time.Duration) error {
+func CreateRBACs(ctx context.Context, sa *coreapi.ServiceAccount, role *rbacapi.Role,
+	roleBindings []rbacapi.RoleBinding, client ctrlruntimeclient.Client,
+	retryDuration, timeout time.Duration) (*coreapi.ServiceAccount, error) {
```
hongkailiu, line 25: I cannot see why we need to return the serviceAccount.

pkg/util/rbacs.go
```diff
@@ -5,61 +5,76 @@ import (
...
+		if !hasDockerCfgImagePullSecretSet(actualSa.ImagePullSecrets) {
 			return false, nil
 		}
 
 		return true, nil
-	},
-	); err != nil {
-		return results.ForReason("create_dockercfg_secrets").WithError(err).Errorf("timeout while waiting for dockercfg secret creation for service account %q: %v", sa.Name, err)
+	}); err != nil {
+		logrus.WithError(err).Warnf("timeout while waiting for dockercfg secret creation for service account %q", sa.Name)
```
hongkailiu, line 75:
```suggestion
		logrus.WithError(err).Debugf("timeout while waiting for dockercfg secret creation for service account %q", sa.Name)
```

pkg/util/rbacs.go
```diff
@@ -5,61 +5,76 @@ import (
...
 
-		if len(actualSA.ImagePullSecrets) == 0 {
+		if !hasDockerCfgImagePullSecretSet(actualSa.ImagePullSecrets) {
 			return false, nil
 		}
 
 		return true, nil
-	},
-	); err != nil {
-		return results.ForReason("create_dockercfg_secrets").WithError(err).Errorf("timeout while waiting for dockercfg secret creation for service account %q: %v", sa.Name, err)
```
hongkailiu, line 61: I would not remove this line, just do not return here.

pkg/util/rbacs.go
danilo-gemoli: We are going to create a ServiceAccount having `registry-pull-credentials` secret already attached to it and then we are going to periodically check whether the `DockerConfigurationController` got the chance to set the `*-dockercfg-*` pull secret. I need to return the ServiceAccount because:
- if it has two `imagePullSecrets` it means everything worked properly, so there is no need to do anything
- if it has only one `imagePullSecrets`, that is `registry-pull-credentials`, then it means we need to apply this mitigation and set the `registry-pull-credentials` secret on the pod directly

I left a comment inside the code: https://github.com/danilo-gemoli/ci-tools/blob/bug/dptp-3134-pod-missing-image-pull-secrets/pkg/steps/multi_stage/multi_stage.go#L419

Does it make sense? Am I missing something?

pkg/util/rbacs.go
danilo-gemoli: What is it useful for if we are not going to return it?

hongkailiu: `results.()` sends data to result-aggregator and then prometheus.

pkg/util/rbacs.go
hongkailiu: > if it has only one imagePullSecrets, that is registry-pull-credentials, then it means we need to apply this mitigation and set the registry-pull-credentials secret on the pod directly

danilo-gemoli: If I got it correctly, `imagePullSecrets`s are copied from the SA to the Pod by an Admission Controller, see [here](https://kubernetes.io/docs/reference/access-authn-authz/service-accounts-admin/#serviceaccount-admission-controller). @jupierce 's idea stated here [DPTP-3134](https://issues.redhat.com/browse/DPTP-3134) is about setting an `imagePullSecret` (that we know it exists and works) on the Pod directly, without relying on any mechanism. I am not sure that would implement what you intended though.

hongkailiu: So different ideas:
1. wire the pull secret to SA (regardless of timeout or not ... and the pull secrets generated for the SA will be there too) and trust kubelet to wire SA's pull secrets to pod.
2. wire the pull secret to pod directly when SA waits for the pull secrets and timeout. That is what you are doing now.
3. in the card by Justin: "ensure admission will add imagePullSecrets to a pod". This is different from 2: add-vs-set. BTW, we do not have any admission in the production yet.

3 is the most complete/safest solution since it _adds_ secrets.

No matter which one is chosen, we need to ensure the secret is in the namespace `ci-op-xxxx`. We need to check the code if that is the case.

hongkailiu:
/lgtm

/hold

Let us roll it out next Monday.

hongkailiu:
/lgtm cancel

lint job failed. :)

hongkailiu (approved):

PR #3022, opened 2022-09-21
New jobrunanalyzer historicaldata cmd

eggfoobar:
/assign @xueqzhan 

Added support for analyzing historical data from big query and generating new query results for origin. 

The results will be updated from big query and a report file will be generated in a file to be used in PRs. A `require_review` file will also be created to denote that due to the types of changes detected, a stricter scrutiny needs to be applied in the PR.


pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/types.go
```diff
@@ -0,0 +1,29 @@
...
+	TimeDiff                              time.Duration `json:"-"`
+	DurationP95                           time.Duration `json:"-"`
+	DurationP99                           time.Duration `json:"-"`
+	jobrunaggregatorapi.HistoricalDataRow `json:",inline"`
+}
+
+type compareResults struct {
+	increaseCount   int
+	decreaseCount   int
+	missingDataList string
```
xueqzhan, line 25: What do we need this for?

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -0,0 +1,211 @@
...
+			d.DurationP95 = newP95
+
+			timeDiff := newP99 - oldP99
+			hasPrevData := oldP99 != 0
+			if newP99 > oldP99 && timeDiff > o.leeway && hasPrevData {
+				increaseCount += 1
+				d.TimeDiff = timeDiff
+			}
+			if newP99 < oldP99 {
+				decreaseCount += 1
```
xueqzhan, line 123: Do we care about sudden large decrease in timeDiff?

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
```diff
@@ -0,0 +1,95 @@
...
+	values.Network = c.Network
+	values.Topology = c.Topology
+	values.P95 = c.P95
+	values.P99 = c.P99
+
+	return json.Marshal(values)
+}
+
+func (a *HistoricalDataRow) GetKey() string {
+	return fmt.Sprintf("%s_%s_%s_%s_%s_%s_%s",
```
xueqzhan, line 86: When a certain element (e.g. FromRelease) is empty, do we get double underscore? That might explain why this ordering is different from when I manually query with "Order By" with the same sequence of columns.

xueqzhan (commented):
Looking good. I am looking forward to see this in action with the automatic job! Thank you for doing this @eggfoobar!

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/types.go
eggfoobar: Good catch, that's a left over from changing it to be an array of the jobs, removing it.

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
eggfoobar: I'm not sure, @dgoodwin do you have an opinion on weather we want to trigger a pr review if the data set goes in the opposite direction, so not only do we trigger if we increase by the leeway but we also trigger if we decrease by the leeway?

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
eggfoobar: Yup, that's what happens when they are empty

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
dgoodwin: Yes let's trigger the review, just to pat ourselves on the back if nothing else.

hack/run-job-aggragator-analyzor.sh
```diff
@@ -0,0 +1,104 @@
...
+        --current $TEMP_DIR/current-disruptions.json \
+        --data-type disruptions \
+        --leeway 1m \
+        --google-service-account-credential-file $GC_CREDS
+else
+    ./job-run-aggregator analyze-historical-data  \
+        --current $TEMP_DIR/current-alerts.json \
+        --data-type alerts \
+        --new $ALERTS_FILE \
+        --leeway 1m
```
dgoodwin, line 90: Could we move this to a percentage based calculation? A minute is an eternity in disruption world, and if something was 1s and jumped to 50s, we really need someone looking at that. If something grew more than 25-50% we probably should have someone review.

hack/run-job-aggragator-analyzor.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+## Run this script to locally analyze data for disruption
```
dgoodwin, line 5: Could you elaborate on this part a little bit, I'm slightly confused on the utility of the fork in logic based on gcs credentials or not.

hack/run-job-aggragator-analyzor.sh
dgoodwin: Ultimately I'd like something alerting us elsewhere and we could remove the PR review here, right now no such other thing exists.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
```diff
@@ -0,0 +1,95 @@
...
+	err := json.Unmarshal(data, &values)
+	if err != nil {
+		return err
+	}
+	c.Name = values.BackendName
+	c.Type = "disruption"
+	if values.AlertName != "" {
+		c.Name = values.AlertName
+		c.Type = "alert"
+	}
```
dgoodwin, line 43: I think it might be good to go with two separate types here, one for disrupt and one for alerts. These marshal functions would no longer be needed and the APIs are a little cleaner for anyone reading or using them, no confusion about what fields will or won't be set. More flexibility to adjust them individually in future as well.

If there is any major downside to that though let me know.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -86,6 +91,133 @@ func NewCIDataClient(dataCoordinates BigQueryDataCoordinates, client *bigquery.C
 	}
 }
 
+func (c *ciDataClient) ListDisruptionHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalDataRow, error) {
+	return c.listDisruptionHistoricalData(ctx)
```
dgoodwin, line 95: Nit: Indirection here probably not needed, might be worth putting the code in here instead.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -86,6 +91,133 @@ func NewCIDataClient(dataCoordinates BigQueryDataCoordinates, client *bigquery.C
...
+	for {
+		data := &jobrunaggregatorapi.HistoricalDataRow{}
+		err = disruptionRow.Next(data)
+		if err == iterator.Done {
+			break
+		}
+		if err != nil {
+			return nil, err
+		}
+		// When querying the P99 and P95 data via code there is no single precision trailing zeros as apposed to manually
```
dgoodwin, line 151: opposed

hack/run-job-aggragator-analyzor.sh
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/env bash
```
dgoodwin, line 1: Nit: script rename to run-job-aggregator-analyzer.sh

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -0,0 +1,209 @@
...
+
+	if len(newHistoricalData) == 0 {
+		return fmt.Errorf("new historical data is empty, can not compare")
+	}
+
+	// We check to make sure the current data set doesn't contain the previous release version
+	// If that's the case, then we are in a new release event (i.e. a new branch has been cut)
+	// We then write to a `require_review` file to record why a review would be required.
+	currentVersionUpdate := currentDataContainsPreviousRelease(previousRelease, currentHistoricalData)
+	if currentVersionUpdate {
```
dgoodwin, line 67: This var name looks confusing, maybe newReleaseUpdate?

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -0,0 +1,209 @@
...
+	err = o.renderResultFiles(result)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("successfully compared (%s) with specified leeway of %s\n", o.dataType, o.leeway)
+	return nil
+}
+
+func (o *JobRunHistoricalDataAnalyzerOptions) compareAndUpdate(newData, currentData map[string]jobrunaggregatorapi.HistoricalDataRow, release string, newReleaseEvent bool) compareResults {
```
dgoodwin, line 88: Could use some godoc, this is where the good stuff is.

dgoodwin (commented):
Great stuff Egli this will be a massive help to the disruption testing stack.

I asked for some moderate refactors here, you can hand off if it's too much or carry on as time allows. It's awesome we can quickly get an update PR up in meantime.

Thanks again!

dgoodwin:
TRT-568

hack/run-job-aggragator-analyzor.sh
eggfoobar: Yeah the intention here is to remove that possible fork, I think the interval of the automation will be like 1-2 week. So if something happened and we wish to update things between that interval, we don't need to manually pull down the data and commit, we can just run this and maintain the consistency in the PRs and datasets in terms of ordering.

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
eggfoobar: Sure thing, I was just keeping it inline with some of the other functions I saw

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
eggfoobar: That sounds good, I like that better, was struggling with naming this one

hack/run-job-aggragator-analyzor.sh
eggfoobar: That's a great idea, will update

hack/run-job-aggragator-analyzor.sh
eggfoobar: I'll update the comments for this

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/util.go
```diff
@@ -0,0 +1,155 @@
...
+
+	data, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", "", err
+	}
+	if err := json.Unmarshal(data, &sippyRelease); err != nil {
+		return "", "", err
+	}
+	sorted := []int{}
+	for _, d := range sippyRelease.Releases {
```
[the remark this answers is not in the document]
eggfoobar: That's fair, I think this should be fine for now since we wont get to 5 for some time. I think that will only happen if there is a major change in kubernetes

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/util.go
```diff
@@ -0,0 +1,155 @@
...
+	return
+}
+
+func formatTableOutput(data []parsedJobData, filter bool) string {
+	sort.SliceStable(data, func(i, j int) bool {
+		return data[i].TimeDiff > data[j].TimeDiff
+	})
+	var buffer bytes.Buffer
+	buffer.WriteString("| Name | Release | From | Arch | Network | Platform | Topology | Time Increase |\n")
+	buffer.WriteString("| ---- | ------- | ---- | ---- | ------- | -------- |--------- | ------------- |\n")
```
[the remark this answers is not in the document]
dgoodwin: I love it, thanks!

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -86,6 +91,126 @@ func NewCIDataClient(dataCoordinates BigQueryDataCoordinates, client *bigquery.C
...
+    Network,
+    Topology,
+    IFNULL(SAFE_CAST(ANY_VALUE(P95) AS STRING), "0.0") AS P95,
+    IFNULL(SAFE_CAST(ANY_VALUE(P99) AS STRING), "0.0") AS P99,
+    FROM (
+        SELECT
+            Jobs.Release,
+            Jobs.FromRelease,
+            Jobs.Platform,
+        Jobs.Architecture,
```
DennisPeriquet, line 111: tiny nit: I think these should line up.

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -0,0 +1,223 @@
...
+	}
+
+	fmt.Printf("successfully compared (%s) with specified leeway of %.2f%%\n", o.dataType, o.leeway)
+	return nil
+}
+
+// compareAndUpdate This will compare the recently pulled information and compare it to the currently existing data.
+// We run our comparison by first checking if we are in a newReleaseEvent, meaning master is now pointing on a new release, then we clean out the current data and generate the results new.
+//
+// If we re in a normal cycle, we then run through our regular comparisons, the primary one being P99.
```
DennisPeriquet, line 90: nit: s/we re/we're/

hack/run-job-aggragator-analyzor.sh
eggfoobar: Added more info for reasoning there

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
eggfoobar: Removed that entirely and separated them out into two different structs

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
eggfoobar: Good catch, updated

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
eggfoobar: Thank you, updated

dgoodwin:
Just needs some lint fixes and then I think we're good to go.

PR #3025, opened 2022-09-21
e2e: mitigate observer pods tests flakiness

danilo-gemoli:
The reason of being of this PR is explained here: [DPTP-3140](https://issues.redhat.com/browse/DPTP-3140)

TL;DR: 
observers run concurrently alongside the regular tests but from time to time it happens tests complete before observers have even the chance to be scheduled. In order to keep things simple I have decided to skip e2e tests when that behavior emerges.

/cc @jmguzik @bbguimaraes 

FYI: @smg247

### DIFF
--- a/pkg/steps/multi_stage/run.go
+++ b/pkg/steps/multi_stage/run.go
@@ -23,6 +23,10 @@ import (
 	"github.com/openshift/ci-tools/pkg/util"
 )
 
+const (
+	TerminateObserversLog string = "Signalling observers to terminate..."
+)
+
 func (s *multiStageTestStep) runSteps(
 	ctx context.Context,
 	phase string,
@@ -123,7 +127,7 @@ func (s *multiStageTestStep) runObservers(ctx, textCtx context.Context, pods []c
 	for _, pod := range pods {
 		go func(p coreapi.Pod) {
 			<-ctx.Done()
-			logrus.Info("Signalling observers to terminate...")
+			logrus.Info(TerminateObserversLog)
 			if err := s.client.Delete(context.Background(), &p); err != nil {
 				logrus.WithError(err).Warn("failed to trigger observer to stop")
 			}
--- a/test/e2e/observer/e2e_test.go
+++ b/test/e2e/observer/e2e_test.go
@@ -9,8 +9,10 @@ import (
 	"path"
 	"path/filepath"
 	"regexp"
+	"strings"
 	"testing"
 
+	"github.com/openshift/ci-tools/pkg/steps/multi_stage"
 	"github.com/openshift/ci-tools/test/e2e/framework"
 )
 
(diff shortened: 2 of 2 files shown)


pkg/steps/multi_stage/run.go
```diff
@@ -123,7 +127,7 @@ func (s *multiStageTestStep) runObservers(ctx, textCtx context.Context, pods []c
 	for _, pod := range pods {
 		go func(p coreapi.Pod) {
 			<-ctx.Done()
-			logrus.Info("Signalling observers to terminate...")
+			logrus.Info(TerminateObserversLog)
```
droslean, line 130: Why did you change this to a constant?

test/e2e/observer/e2e_test.go
```diff
@@ -9,26 +9,33 @@ import (
...
 )
 
+var (
+	observerHasFailedRegexp = regexp.MustCompile("Step (.+-observer) failed after")
```
droslean, line 24: This is smelly. Can you elaborate more on why we need to parse the error log to understand which step failed?

jmguzik:
@danilo-gemoli what is with this PR?

danilo-gemoli:
I've put this on hold to work on higher priority tasks. Nobody is complaining anymore so I have to check if the problem still exists.

test/e2e/observer/e2e_test.go
danilo-gemoli: I have noticed (and you can find some logs backing up this assumption [here](https://issues.redhat.com/browse/DPTP-3140)) that sometime the test gets flaky because the observer didn't run at at all. This is more or less what could happen:
- the [test step](https://github.com/openshift/ci-tools/blob/28deee345611ee06adf123ed9736263005add134/test/e2e/observer/failing-observer.yaml#L15) starts
- the [observer](https://github.com/openshift/ci-tools/blob/28deee345611ee06adf123ed9736263005add134/test/e2e/observer/failing-observer.yaml#L25) is supposed to run concurrently but it never gets the chance to. That is a mere timing issue not related with the code
-  the `e2e` test expects the observer to have logged its failure but it finds none
- the whole test fails but it's just a flake

What I'm trying to do here (without doing any invasive change to the `ci-operator` code) is to check for the line the observer is supposed to have logged; if there is none then it means the previous scenario happened and we can therefore ignore it

pkg/steps/multi_stage/run.go
danilo-gemoli: Because it is used in the `e2e` test code as well

test/e2e/observer/e2e_test.go
droslean: I don't have a clear picture of how the observers work, but if we run the observer in parallel, we should have a watcher that monitors the observer's behavior. The fact that the test awaits for error and finds none, is a bug and not a flake. We should fix that instead.

danilo-gemoli: If an observer fails, the `ci-operator` should not consider the whole test to be failed, this is by design and this `e2e` test try to ensure this behavior holds by looking at the logs there were produced. The observer is ephemeral by nature, the `ci-operator` run it at best effort and, from what I have understood, in some cases the whole tests chain executes but the observer(s) does not.

PR #3027, opened 2022-09-21
DPTP-3106 add reporter to admission

psalajova:
Adds the ability to work with report credentials ([this PR](https://github.com/openshift/release/pull/32403)), and pod-scaler-admission can now send requests to result-aggregator using PodScalerReporter.

/cc smg247

### DIFF
--- a/cmd/pod-scaler/admission_test.go
+++ b/cmd/pod-scaler/admission_test.go
@@ -31,11 +31,11 @@ import (
 
 type mockReporter struct {
 	client *http.Client
-	state  string
+	called bool
 }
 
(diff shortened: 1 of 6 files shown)


smg247 (changes requested):

cmd/pod-scaler/admission_test.go
```diff
@@ -28,6 +29,17 @@ import (
 	"github.com/openshift/ci-tools/pkg/testhelper"
 )
 
+type mockReporter struct {
+	client *http.Client
+	state  string
```
smg247, line 34: rather than a string, I would just use a `bool` here. you can just set it to `true` when the method is called.

smg247 (changes requested):

PR #3028, opened 2022-09-21
Increase progress deadline for rpm server

jupierce:
During extreme scale up events on build farms, particularly when clouds are low on capacity, it may take a long time for pods to be scheduled. It's normally counter productive for jobs to fail because of an rpm server timeout, so increase this timeout significantly.

### DIFF
--- a/pkg/steps/rpm_server.go
+++ b/pkg/steps/rpm_server.go
@@ -83,14 +83,14 @@ func (s *rpmServerStep) run(ctx context.Context) error {
 		SuccessThreshold:    1,
 		TimeoutSeconds:      1,
 	}
-	one := int64(1)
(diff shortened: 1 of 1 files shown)


pkg/steps/rpm_server.go
```diff
@@ -83,14 +83,14 @@ func (s *rpmServerStep) run(ctx context.Context) error {
 		SuccessThreshold:    1,
 		TimeoutSeconds:      1,
 	}
-	one := int64(1)
-	two := int32(2)
-	progressDeadline := int32(1200) // It can take 10 minutes for a machine to come up, so double default deadline
```
bbguimaraes, line 88: Ah, those good old days…

PR #3035, opened 2022-09-23
`pj-rehearse-plugin`: rehearsal logic port and refactor

smg247:
The first step in creating an external prow plugin for `pj-rehearse` is to port the logic over and refactor it in a way that is more usable by an event based plugin. I opted to keep the original `pj-rehearse` logic untouched (other than a single function that needed a flag added) so that: there is no risk to it breaking, we can do a soft rollout, and we can continue to use it quickly if we find a problem with the plugin.

The original `pj-rehearse` functionality has been preserved in the port to `pj-rehearse-plugin/rehearse.go`, but the code has been refactored to make it easier to use for the plugin. The changes of note include:

- Accepting a `github.PullRequest` argument to work off of rather than a `JobSpec`, as we will not have access to a `JobSpec` in the plugin
- Splitting the logic between determining the affected jobs and actually triggering them
- Reordering much of the logic to utilize variables much closer to where they are defined
- Using the candidate SHA rather than the `buildId` to name the temporary configmaps
- While the `loggers` struct is still used, both the job and debug logger are the same, this concept isn't relevant anymore and the struct will be removed when we get rid of the original `pj-rehearse`

Currently, this only functions in `dryRun` mode for integration testing.

/cc @openshift/test-platform 
For: https://issues.redhat.com/browse/DPTP-3092


droslean:
I think this approach is more complicated. We can just introduce the event server into the tool and enable it with a specific flag. This way we are keeping backward compatibility and it would be much easier to review and follow-up with the changes.

smg247:
> I think this approach is more complicated. We can just introduce the event server into the tool and enable it with a specific flag. This way we are keeping backward compatibility and it would be much easier to review and follow-up with the changes.

It requires significant refactoring to be utilized by an event server, so I opted to not do that. I could apply this refactor to `pj-rehearse` itself, but I would have to implement, potentially complex, logic to switch between using the `JobSpec` and the `PullRequest`, as well as the `buildId` vs the candidate SHA, and the distinct loggers. Due to all of these changes I thought it would be _safer_ to port the logic over. Since the integration tests are ported over as well we can be sure that we have maintained the basic functionality. I realize that doing it this way results in a larger diff though...

droslean:
I thought it would be easier to opt-in the current logic into the event server. We can manipulate the given `JOB_SPEC` based on the pull request since we are using only a few parts of the job_spec and the `build_id` can be a unique hash. In a follow-up, you can clean this out too. To create a new tool and then migrate the logic in there looks messier. If it's easier for you to do it this way, then first you should opt out the whole logic from the tool, and then migrate it to the new tool.  I just trying to avoid copy-pasting the code around.

smg247:
> If it's easier for you to do it this way, then first you should opt out the whole logic from the tool, and then migrate it to the new tool. I just trying to avoid copy-pasting the code around.

That is a good point. I could use the `pj-rehearse-plugin/rehearse.go` logic in `pj-rehearse` I would just need to create a `github.PullRequest` from the `JobSpec` and also pass a new flag to switch up the logging and waiting logic. 

It avoids duplicated code, but does create quite a bit of extra work. I will think about which way is better if it is an absolute blocker for you to have this code temporarily duplicated due to the port. Keep in mind that the original tool will go away as soon as we get this running and verify that it is functioning correctly, so the duplication will be short lived anyways.

droslean:
We could also introduce a new flag `--run-as-plugin` or something like that, and support both of the implementations. Since the logic is one, then the tool can operate on that logic based on those two preferences, as a job or as a plugin. 

If in your very first PR, you can refactor the logic and opt-out from the tool, then the same tool can run this logic in both ways.

After this, and with a few follow-up PRs to refactor a few areas, then we can just remove the old way of running the tool as a job. 

Creating a new tool to run the same logic differently doesn't make sense to me.

PR #3039, opened 2022-09-26
pr-reminder: modify to not include un-actionable PRs

psalajova:
Modifies the pr-reminder tool so that the slack message that gets sent every morning doesn't include those PRs that have un-actionable labels (`needs-rebase` or `do-not-merge/work-in-progress`).

for https://issues.redhat.com/browse/DPTP-3037

/cc @openshift/test-platform

### DIFF
--- a/cmd/pr-reminder/main.go
+++ b/cmd/pr-reminder/main.go
@@ -19,6 +19,7 @@ import (
 	"k8s.io/test-infra/prow/config/secret"
 	"k8s.io/test-infra/prow/flagutil"
 	"k8s.io/test-infra/prow/github"
+	"k8s.io/test-infra/prow/labels"
 	"k8s.io/test-infra/prow/logrusutil"
 )
 
@@ -72,16 +73,16 @@ type config struct {
 
 // getInterestedLabels returns a set of those labels we are interested in when using the PR reminder
 func getInterestedLabels() sets.String {
-	var labels = sets.String{}
-	labels.Insert("do-not-merge/hold")
-	return labels
+	var prLabels = sets.String{}
+	prLabels.Insert("do-not-merge/hold")
+	return prLabels
 }
 
 // getUnactionablePrLabels returns a set of those labels that mark a PR which can't be reviewed in its current state
 func getUnactionablePrLabels() sets.String {
-	var labels = sets.String{}
-	labels.Insert("do-not-merge/work-in-progress", "needs-rebase")
-	return labels
+	var prLabels = sets.String{}
+	prLabels.Insert(labels.WorkInProgress, labels.NeedsRebase)
+	return prLabels
 }
 
 var orgRepoFormat = regexp.MustCompile(`\w+/\w+`)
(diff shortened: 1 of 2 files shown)


cmd/pr-reminder/main_test.go
```diff
@@ -598,3 +598,50 @@ func Test_filterLabels(t *testing.T) {
 		})
 	}
 }
+
+func Test_hasUnactionableLabels(t *testing.T) {
+	holdLabel := github.Label{Name: "do-not-merge/hold"}
```
droslean, line 603: Check https://pkg.go.dev/k8s.io/test-infra/prow/labels#pkg-constants

cmd/pr-reminder/main.go
```diff
@@ -77,6 +77,13 @@ func getInterestedLabels() sets.String {
 	return labels
 }
 
+// getUnactionablePrLabels returns a set of those labels that mark a PR which can't be reviewed in its current state
+func getUnactionablePrLabels() sets.String {
```
droslean, line 81: I don't think that we need a new function for that. You can specify those as vars or constants.

cmd/pr-reminder/main.go
```diff
@@ -458,3 +465,15 @@ func messageUser(user user, slackClient slackClient) error {
 func getLabelMessage(labels []string) string {
 	return fmt.Sprintf(":label: labeled: *%v*", strings.Join(labels[:], ", "))
 }
+
+// hasUnactionableLabels returns whether a PR has any labels which mark a PR
+// that can't be reviewed in its current state
+func hasUnactionableLabels(labels []github.Label) bool {
+	unactionableLabels := getUnactionablePrLabels()
+	for _, label := range labels {
+		if unactionableLabels.Has(label.Name) {
```
droslean, line 474: Check https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets#String.Intersection

cmd/pr-reminder/main.go
psalajova: We could, I did it like this so the style is consistent with the `getInterestedLabels()` func above, and because it's convenient to have all the labels we need in one set I can work with later on. But I can change it.

cmd/pr-reminder/main.go
psalajova: the arg `labels` has type `[]github.Label` and unactionableLabels is a set of **strings**, and so I don't see a way how to use the Intersection function without first converting `labels` into a set of strings, which I think would complicate the function even more. but maybe I'm not seeing someting?

droslean: Potentially you can create a set from the `[]github.labels` too and compare them.

droslean: You can also check https://pkg.go.dev/k8s.io/test-infra@v0.0.0-20220929080830-4693eb3a938b/prow/github#HasLabels

smg247: Let's not do another API call for this, that would make this a considerably more expensive tool to run.

smg247: Ah, you mean this one: https://github.com/kubernetes/test-infra/blob/29364b9d3b47613dbd8db503fe2997fcedb1a30c/prow/github/helpers.go#L45 that is fine to use

PR #3043, opened 2022-09-26
Drop parsing of disruption data from junit test results.

dgoodwin:
This was a solution for uploading disruption data prior to 4.10, we may
no longer be collecting data for those releases but this is fine as
we do not believe anything is using that data.

I am removing to simplify the code paths in the disruption code base.

[TRT-573](https://issues.redhat.com//browse/TRT-573)


pkg/jobrunaggregator/jobrunbigqueryloader/disruption_junit.go
```diff
@@ -1,130 +0,0 @@
...
-			continue
-		}
-
-		// if the test passed and we DO NOT have an entry already, add one
-		if _, ok := availabilityResultsByName[backendName]; !ok {
-			availabilityResultsByName[backendName] = jobrunaggregatorlib.AvailabilityResult{
-				ServerName:         backendName,
-				SecondsUnavailable: 0,
-			}
-		}
```
dgoodwin, line 92: @deads2k does this look like the spot you were remembering where we might be ingesting 0's by default?

I cannot find an equivalent in the code path for ingesting from backend-disruption json files: https://github.com/openshift/ci-tools/blob/master/pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go#L51

dgoodwin: David confirmed on slack that this, plus ensuring there isn't a console entry in backend-disruption.json files noting 0, should cover it. I confirmed this previously in job runs like https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/27439/pull-ci-openshift-origin-master-e2e-gcp-ovn-upgrade/1573367107456864256/artifacts/e2e-gcp-ovn-upgrade/openshift-e2e-test/artifacts/junit/backend-disruption_20220923-183829.json

PR #3044, opened 2022-09-26
Omit disruptive jobs from the biquery disruption db.

dgoodwin:
Keeping them out of the job names list means they will not be imported
into bigquery, however the jobs themselves continue to run disruption
monitoring, output data files, and show disruption in the spyglass
intervals chart.

[TRT-573](https://issues.redhat.com//browse/TRT-573)


DennisPeriquet:
This LGTM but I have no /lgtm powers.

PR #3046, opened 2022-09-26
Bump aggregator to 6h at least temporarily.

dgoodwin:
We're extending the job limit to 8h in https://github.com/openshift/ci-tools/pull/3045

The aggregator's internal wait time was still 4h, thus we were still
timing out.


soltysh (approved):
/lgtm

droslean (approved):

PR #3052, opened 2022-09-28
feat: add support for commit message being different than pr body

eggfoobar:
Signed-off-by: eggfoobar <eggfoobar@users.noreply.github.com>

### DIFF
--- a/cmd/prcreator/main.go
+++ b/cmd/prcreator/main.go
@@ -13,12 +13,13 @@ import (
 
 type options struct {
 	prcreation.PRCreationOptions
-	prTitle      string
-	prMessage    string
-	prAssignee   string
-	organization string
-	repo         string
-	branch       string
+	prTitle         string
+	prMessage       string
+	prCommitMessage string
+	prAssignee      string
+	organization    string
+	repo            string
+	branch          string
 }
 
 func gatherOptions() (*options, error) {
@@ -26,6 +27,7 @@ func gatherOptions() (*options, error) {
 	opts.PRCreationOptions.AddFlags(flag.CommandLine)
 	flag.StringVar(&opts.prTitle, "pr-title", "", "The title of the PR to create")
 	flag.StringVar(&opts.prMessage, "pr-message", "", "The message of the PR to create")
+	flag.StringVar(&opts.prCommitMessage, "pr-commit-message", "", "The git commit message of the PR to create")
 	flag.StringVar(&opts.prAssignee, "pr-assignee", "", "The assignee of the PR to create")
 	flag.StringVar(&opts.organization, "organization", "openshift", "The GitHub organization in which the PR should be created")
 	flag.StringVar(&opts.repo, "repo", "release", "The name of the repo in which the PR should be created")
@@ -66,6 +68,7 @@ func main() {
 		opts.prTitle,
 		prcreation.PrBody(opts.prMessage),
 		prcreation.PrAssignee(opts.prAssignee),
+		prcreation.PrCommitMessage(opts.prCommitMessage),
 	); err != nil {
 		logrus.WithError(err).Fatal("failed to upsert PR")
 	}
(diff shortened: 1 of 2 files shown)


cmd/prcreator/main.go
```diff
@@ -13,19 +13,21 @@ import (
...
+	repo            string
+	branch          string
 }
 
 func gatherOptions() (*options, error) {
 	opts := options{}
 	opts.PRCreationOptions.AddFlags(flag.CommandLine)
 	flag.StringVar(&opts.prTitle, "pr-title", "", "The title of the PR to create")
 	flag.StringVar(&opts.prMessage, "pr-message", "", "The message of the PR to create")
+	flag.StringVar(&opts.prCommitMessage, "pr-commit-message", "", "The git commit message of the PR to create")
```
hongkailiu, line 30: Could we use `pr-message` instead?

eggfoobar: The previous flag uses that for the PR itself, would changing it to `git-message` be clearer?

eggfoobar: @hongkailiu Any thoughts on that idea?

hongkailiu: Sorry for the late reply.

Before this PR:
`prMessage` is used to init the PRBody.

https://github.com/openshift/ci-tools/blob/e0c715b105fc5b1d3a57f537bc40b1f2cab046a9/cmd/prcreator/main.go#L67

and then the PRBody is used as the commit msg

https://github.com/openshift/ci-tools/blob/e0c715b105fc5b1d3a57f537bc40b1f2cab046a9/pkg/github/prcreation/prcreation.go#L155

and the PR body

https://github.com/openshift/ci-tools/blob/e0c715b105fc5b1d3a57f537bc40b1f2cab046a9/pkg/github/prcreation/prcreation.go#L180

The above logic is usually what happens when we create a PR manually:
- The first line of commit msg becomes the PR title
- The rest of the commit msg becomes the PR body/description

Of course, we can manually modify the PR description while creating a PR.

Your goal is to have different values for the commit msg and the PR body, correct?

Yes, `git-message` sounds better to me since it is not bound to PR. 
Moreover, could you please explain (in the code of `flag.StringVar`) what would happen if it is not set (when the default value "" takes effect)? Intuitively, a user would think it results in empty commit msg but actually it is not the case.

eggfoobar: > Your goal is to have different values for the commit msg and the PR body, correct?

Yes, in some of the PR's we're looking to do, the PR Body will include large markdown tables and we don't want that much information to be part of the commit history itself.

> Yes, `git-message` sounds better to me since it is not bound to PR.

Sounds good, I'll update that now.

> Moreover, could you please explain (in the code of `flag.StringVar`) what would happen if it is not set (when the default value "" takes effect)? Intuitively, a user would think it results in empty commit msg but actually it is not the case.

Sure, my thought is if no value is provided, a default would be that we fall back to the `prBody`. I will add a message that, that will be the default so there's no confusion, or if it's desired we can support explicit empty commit message, something like `--git-message="-"`?

cmd/prcreator/main.go
```diff
@@ -13,19 +13,21 @@ import (
...
+	repo             string
+	branch           string
 }
 
 func gatherOptions() (*options, error) {
 	opts := options{}
 	opts.PRCreationOptions.AddFlags(flag.CommandLine)
 	flag.StringVar(&opts.prTitle, "pr-title", "", "The title of the PR to create")
 	flag.StringVar(&opts.prMessage, "pr-message", "", "The message of the PR to create")
+	flag.StringVar(&opts.gitCommitMessage, "git-message", "", "The git commit message of the PR to create, default is pr-message")
```
hongkailiu, line 30:
```suggestion
	flag.StringVar(&opts.gitCommitMessage, "git-message", "", "The git commit message of the PR to create. If not set, then its value will be composed of other flags")
```

eggfoobar: Sounds good, thanks! Updated

hongkailiu:
/lgtm

Thanks.

PR #3053, opened 2022-09-29
Prowgen: multiarch postsubmit images

danilo-gemoli:
This is related to [DPTP-3029](https://issues.redhat.com/browse/DPTP-3029)

/cc @droslean

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1980,3 +1980,34 @@ type MetadataWithTest struct {
 func (m *MetadataWithTest) JobName(prefix string) string {
 	return m.Metadata.JobName(prefix, m.Test)
 }
+
+type Architecture string
+
+const (
+	AMD64Arch Architecture = "amd64"
+	ARM64Arch Architecture = "arm64"
+)
+
+var archToCluster = map[Architecture]Cluster{
+	ARM64Arch: ClusterARM01,
+}
+
+func (a Architecture) IsValid() bool {
+	return a.GetMappedCluster() != ""
+}
+
+func (a Architecture) GetMappedCluster() Cluster {
+	c, found := archToCluster[a]
+	if !found {
+		return ""
+	}
+	return c
+}
+
+func GetAvailableArchitectures() []string {
+	architectures := make([]string, 0, len(archToCluster))
+	for arch := range archToCluster {
+		architectures = append(architectures, string(arch))
+	}
+	return architectures
+}
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -12,6 +12,9 @@ import (
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
 
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+
+	"github.com/openshift/ci-tools/pkg/api"
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/util"
 	"github.com/openshift/ci-tools/pkg/util/gzip"
@@ -33,6 +36,24 @@ type Prowgen struct {
 	Expose bool `json:"expose,omitempty"`
 	// Rehearsals declares any disabled rehearsals for jobs
 	Rehearsals Rehearsals `json:"rehearsals,omitempty"`
+	// Set which architecture should the images be promoted from
+	AdditionalArchitectures []api.Architecture `json:"additional_architectures"`
+}
+
+func (p *Prowgen) Validate() error {
+	errs := make([]error, 0)
+	invalidArchs := make([]string, 0, len(p.AdditionalArchitectures))
+	for _, arch := range p.AdditionalArchitectures {
+		if !arch.IsValid() {
+			invalidArchs = append(invalidArchs, string(arch))
+		}
+	}
+	if len(invalidArchs) > 0 {
+		e := fmt.Errorf("architectures %s are not valid, available ones are: %s",
+			strings.Join(invalidArchs, ", "), strings.Join(api.GetAvailableArchitectures(), ", "))
+		errs = append(errs, e)
+	}
+	return utilerrors.NewAggregate(errs)
 }
 
 type Rehearsals struct {
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -38,7 +38,7 @@ type ProwgenInfo struct {
 // new jobs are generated with GenerateJobs, the call site should also use
 // Prune() function to remove all stale jobs and label the jobs as simply
 // "generated".
-func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *ProwgenInfo) *prowconfig.JobConfig {
+func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *ProwgenInfo) (*prowconfig.JobConfig, error) {
 	orgrepo := fmt.Sprintf("%s/%s", info.Org, info.Repo)
 	presubmits := map[string][]prowconfig.Presubmit{}
 	postsubmits := map[string][]prowconfig.Postsubmit{}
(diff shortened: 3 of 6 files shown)


pkg/config/load.go
```diff
@@ -33,6 +33,21 @@ type Prowgen struct {
 	Expose bool `json:"expose,omitempty"`
 	// Rehearsals declares any disabled rehearsals for jobs
 	Rehearsals Rehearsals `json:"rehearsals,omitempty"`
+	// Set which cluster and architecture should the images be promoted from
+	Images Images `json:"postsubmit_images"`
+}
+
+type Images struct {
+	// amd64 is the default arch, falsify this field to skip it
+	SkipDefaultArch bool   `json:"skip_default_arch,omitempty"`
```
droslean, line 42: I don't think we need this. We will always build for amd64. We only need to allow users to specify the extra architectures

pkg/config/load.go
```diff
@@ -33,6 +33,21 @@ type Prowgen struct {
...
+	Architectures   []Arch `json:"architectures"`
+}
+type Arch struct {
+	// The name of the architecture, it will be used as suffix for the job
+	// Ex.: openshift-ci-tools-master-images-${arch.Name} -> openshift-ci-tools-master-images-arm64
+	Name string `json:"name,omitempty"`
```
droslean, line 48: This is not needed actually. You can get the name from the name of the architecture that we will hardcode.

pkg/config/load.go
```diff
@@ -33,6 +33,21 @@ type Prowgen struct {
...
+	// The cluster that will run the promotion
+	Cluster string `json:"cluster,omitempty"`
```
droslean, line 50: This is not needed. We should hardcode the clusters/architectures and validate the config as well.

danilo-gemoli: By hardcoding clusters/architectures, do you mean something like this one below?
```golang
func determineCluster(arch string) (string, error) {
  switch arch {
  case "arm64":
    return "arm01", nil
  case "powervs":
    return "pw01", nil
  // More cases here
  // ...
  default:
    return "", fmt.Errorf("arch %s is not supported", arch)
  }
}
```

pkg/config/load.go
```diff
@@ -33,6 +33,21 @@ type Prowgen struct {
 	Expose bool `json:"expose,omitempty"`
 	// Rehearsals declares any disabled rehearsals for jobs
 	Rehearsals Rehearsals `json:"rehearsals,omitempty"`
+	// Set which cluster and architecture should the images be promoted from
+	Images Images `json:"postsubmit_images"`
```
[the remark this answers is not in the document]
danilo-gemoli: Is it ok to take the names from that list and putting it as a suffix for the name of the job? Like:
```golang
fmt.Sprintf("branch-ci-openshift-ci-tools-master-images-%s", "arm64")
```
Or do we need to hardcode it as well?
```golang
func determineArchitectureName(arch string) (string, error) {
  switch arch {
  case "arm64":
    return "arm64", nil
  case "powervs":
    return "powervs", nil
  // More cases here
  // ...
  default:
    return "", fmt.Errorf("arch %s is not supported", arch)
  }
}
```

pkg/config/load.go
danilo-gemoli: See the answer above.

pkg/config/load.go
droslean: Yea, but let's have a constant for each arch and cluster. I think we already do somewhere. Techically, we should have a way to connect the arch with clusters. Now we have `arm64` --> `arm01` cluster, but we are planning to add more.

pkg/config/load.go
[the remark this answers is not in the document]
droslean: Yea I agree with putting the arch in the name suffix

pkg/api/types.go
```diff
@@ -1980,3 +1980,21 @@ type MetadataWithTest struct {
...
 }
+
+type Architecture string
+
+const (
+	AMD64Arch Architecture = "amd64"
+	ARM64Arch Architecture = "arm64"
+)
+
+var archToCluster = map[Architecture]Cluster{
```
droslean, line 1991:
```suggestion
var archToCluster = map[Architecture][]Cluster{
```

But this is not important now. Once we will add the other clusters, we can make the change. Feel free not to address this or write a comment

pkg/api/types.go
```diff
@@ -1980,3 +1980,21 @@ type MetadataWithTest struct {
...
+	ARM64Arch: ClusterARM01,
+}
+
+func (a Architecture) GetDefaultCluster() (Cluster, error) {
+	if c, ok := archToCluster[a]; ok {
+		return c, nil
+	}
+	return "", fmt.Errorf("no predefined cluster for arch: %s", string(a))
```
droslean, line 1999: This has to be validated once the config is loaded, and not after.

pkg/prowgen/prowgen.go
```diff
@@ -98,28 +98,8 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	}
 
 	imageTargets := api.ImageTargets(configSpec)
-
-	if len(imageTargets) > 0 {
```
droslean, line 102: let's don't move them, since this change is not related to the PR.

pkg/prowgen/prowgen.go
```diff
@@ -146,7 +126,68 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
...
+	var presubmitTargets = imageTargets.List()
+	if api.PromotesOfficialImages(configSpec, api.WithOKD) {
+		presubmitTargets = append(presubmitTargets, "[release:latest]")
+	}
+
+	jobBaseGen := newJobBaseBuilder().TestName("images")
+	jobBaseGen.PodSpec.Add(Targets(presubmitTargets...))
+	presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(jobBaseGen, "images", info))
+
+	generatePostsubmitsJobsForArch := func(arch api.Architecture) error {
```
droslean, line 153: Do we need to do this separately? We already generate the postsubmit. We could pass the additional arch config and generate a list of postsubmits instead.

pkg/api/types.go
```diff
@@ -1980,3 +1980,28 @@ type MetadataWithTest struct {
...
+		return false, err
+	}
+	return true, nil
+}
+
+func (a Architecture) GetDefaultCluster() (Cluster, error) {
+	if c, ok := archToCluster[a]; ok {
+		return c, nil
+	}
+	return "", fmt.Errorf("no predefined cluster for arch: %s", string(a))
```
droslean, line 2006: I don't think we should return an error at all. If the cluster is not valid, let's just return false, and perhaps a message with that will include all supported clusters.

pkg/prowgen/prowgen.go
```diff
@@ -110,15 +110,16 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 		presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(jobBaseGen, "images", info))
 
 		if configSpec.PromotionConfiguration != nil {
-			jobBaseGen = newJobBaseBuilder().TestName("images")
-			jobBaseGen.PodSpec.Add(Promotion(), Targets(imageTargets.List()...))
-			postsubmit := generatePostsubmitForTest(jobBaseGen, info)
-			postsubmit.MaxConcurrency = 1
-			if postsubmit.Labels == nil {
-				postsubmit.Labels = map[string]string{}
+			for _, arch := range append([]api.Architecture{api.AMD64Arch}, info.Config.AdditionalArchitectures...) {
```
droslean, line 113: This has to be done inside `generatePostsubmitForPromotion`

pkg/config/load.go
```diff
@@ -33,6 +36,24 @@ type Prowgen struct {
 	Expose bool `json:"expose,omitempty"`
 	// Rehearsals declares any disabled rehearsals for jobs
 	Rehearsals Rehearsals `json:"rehearsals,omitempty"`
+	// Set which architecture should the images be promoted from
+	AdditionalArchitectures []api.Architecture `json:"additional_architectures"`
+}
+
+func (p *Prowgen) Validate() error {
+	errs := make([]error, 0)
+	invalidArchs := make([]string, 0, len(p.AdditionalArchitectures))
```
droslean, line 45: Shall we care to display which archs are invalid? If, so, you can just do it with sets and intersections.

pkg/prowgen/prowgen.go
```diff
@@ -218,6 +217,41 @@ func generatePostsubmitForTest(jobBaseBuilder *prowJobBaseBuilder, info *Prowgen
...
+	}
+	architectures := append([]api.Architecture{api.AMD64Arch}, info.Config.AdditionalArchitectures...)
+	postsubmits := make([]prowconfig.Postsubmit, 0, len(architectures))
+	for _, arch := range architectures {
+		jobBaseBuilder := jobBaseBuilderFactory()
+		var jobBaseGen *prowJobBaseBuilder
+		if arch != api.AMD64Arch {
+			testName := fmt.Sprintf("images-%s", string(arch))
+			cluster, found := arch.GetMappedCluster()
+			if !found {
```
droslean, line 233: You will never hit this since you are validating the config.

danilo-gemoli: Yes, but this is just an execution path, we should not ignore the error because we know we have validated the config beforehand.

pkg/api/types.go
```diff
@@ -1980,3 +1980,32 @@ type MetadataWithTest struct {
...
+	ARM64Arch: ClusterARM01,
+}
+
+func (a Architecture) IsValid() bool {
+	_, found := a.GetMappedCluster()
+	return found
+}
+
+func (a Architecture) GetMappedCluster() (Cluster, bool) {
+	c, found := archToCluster[a]
```
droslean, line 2001: found doesn't make sense here, since you are returning the object.

pkg/config/load.go
```diff
@@ -33,6 +37,20 @@ type Prowgen struct {
 	Expose bool `json:"expose,omitempty"`
 	// Rehearsals declares any disabled rehearsals for jobs
 	Rehearsals Rehearsals `json:"rehearsals,omitempty"`
+	// Set which architecture should the images be promoted from
+	AdditionalArchitectures []api.Architecture `json:"additional_architectures"`
+}
+
+func (p *Prowgen) Validate() error {
+	errs := make([]error, 0)
+	additionalArchs := sets.NewString(util.CastSliceStr[api.Architecture, string](p.AdditionalArchitectures)...)
```
droslean, line 46: You can just iterate on the architectures and validate each one of them

droslean (approved):

PR #3054, opened 2022-09-29
ci-operator multi-stage: log observer errors

bbguimaraes:
This part of the code is already difficult to follow as it is.

`ci-operator` is generally not happy about pods it is following being deleted,
which does not happen frequently due to the PDB used for builds and tests.  Put
the error in the debug log so it can be used to understand how the monitoring
code responds to the deletion.

---

The notifier (which generates JUnit information) is only triggered by terminated
pods:

https://github.com/openshift/ci-tools/blob/68249882ebce83a5f9b093b313adad88d67f67dd/pkg/util/pods.go#L378

which explains results seen in some of the E2E test failures (the test itself is
a very contrived — but possible — scenario).  This should help identify the sort
of information coming from the API server in those situations.


bbguimaraes:
https://issues.redhat.com/browse/DPTP-3140

/cc danilo-gemoli

droslean (approved):

PR #3061, opened 2022-10-04
Treat observers as references in graph

jmguzik:


droslean (approved):

jmguzik:
> Why treat them as references? Those are two different things.
As discussed, this is a hack to enable rehearsals with observers. Hopefully, it will work. The scope is very limited, with potential harm as well. It is not a definite solution, we will get back to this.
/hold cancel

bbguimaraes:

PR #3062, opened 2022-10-04
Use token file in secret test-credentials/ci-operator

hongkailiu:
The breaking will be fixed by https://github.com/openshift/release/pull/32833

As far as I can tell, the secret is used only in the tests of `ci-tools`:

https://github.com/openshift/release/blob/70499fc48ae9e4bbdc621dee9820ddd4c02d433f/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml#L765

The verification will be in two steps:
1. add the tokenFile to the secret but the config file is still with `token`.
2. use the tokenFile in the config file.

We have to do this PR first because otherwise the build-clusters job would fail in 
https://github.com/openshift/release/pull/32642

### DIFF
--- a/cmd/cluster-init/update_bootstrap_secrets.go
+++ b/cmd/cluster-init/update_bootstrap_secrets.go
@@ -85,13 +85,21 @@ func updateSecret(secretGenerator func(options) secretbootstrap.SecretConfig) fu
 }
 
 func generateCiOperatorSecret(o options) secretbootstrap.SecretConfig {
-	return secretbootstrap.SecretConfig{
-		From: map[string]secretbootstrap.ItemContext{
-			kubeconfig: {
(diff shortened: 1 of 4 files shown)


cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -18,6 +18,7 @@ const (
 	testCredentials     = "test-credentials"
 	kubeconfig          = "kubeconfig"
 	config              = "config"
+	token               = "token.txt"
```
smg247, line 21: can this be moved to `main.go` since that is the only usage?

smg247 (changes requested):

cmd/cluster-init/update_bootstrap_secrets.go
hongkailiu: Done.

PR #3063, opened 2022-10-04
`pj-rehearse`: refactor in place to allow for use in prow plugin

smg247:
The replacement for https://github.com/openshift/ci-tools/pull/3035 to refactor `pj-rehearse` in place rather than as a separate tool.

The original `pj-rehearse` functionality has been preserved in the move to `pj-rehearse/rehearse.go`, but the code has been refactored to make it easier to use for the plugin. The changes of note include:

- Accepting a `rehearsalCandidate` argument to work off of rather than a `JobSpec`, as we will not have access to a `JobSpec` in the plugin. We can form a `rehearsalCandidate` from a `JobSpec` or a `github.PullRequest`.
- Splitting the logic between determining the affected jobs and actually triggering them
- Reordering much of the logic to utilize variables much closer to where they are defined
- Using the candidate SHA rather than the `buildId` to name the temporary configmaps

/cc @openshift/test-platform


cmd/pj-rehearse/rehearse.go
```diff
@@ -0,0 +1,588 @@
+package main
```
droslean, line 1: Moving the code to a different file makes the PR really hard to review and also to me it doesn't make any sense to do it since it is still in the same package. Perhaps this should live in `pkg/rehearse`

test/integration/pj-rehearse.sh
```diff
@@ -43,6 +43,9 @@ export JOB_SPEC='{"type":"presubmit","job":"pull-ci-openshift-release-master-reh
 actual="${workdir}/rehearsals.yaml"
 os::cmd::expect_success "pj-rehearse --dry-run=true --candidate-path ${repo} --rehearsal-limit 20 > ${actual}"
 os::integration::sanitize_prowjob_yaml ${actual}
+# Substitute the SHA in the job names to a known SHA for comparison
+sed -i -E -e "s/-${candidate_sha}-/-4de8ab7c20656998264a2593116288f5eb070b32-/g" ${actual}
```
droslean, line 47: This will break the manifest since the limit is to 63 chars

smg247: I assume you mean using the SHA will break it and not the substitution for the integration tests. In that case, do you have any idea of what else we can use absent the `BuildID`?

droslean: Yea I mean that this will generate https://github.com/openshift/ci-tools/pull/3063/files#diff-e0b6c8feb0bd8af9facc04c0a2e3384bf1258e99ea6c5404cdf25abb998af9bfR132 which is not allowed in Kubernetes. We could use something else for the tests though. Maybe a 5digit hash code with the job itself as an input?

smg247: Is it important that it is unique (and deterministic) to a specific HEAD of the PR? I thought that was why the `BuildID` was being used in the first place, but I could have made a poor assumption. If that _is_ the case, maybe we can do a hash code based on the SHA...

droslean: For the record: The limit is 253 chars for the object names, and 64 chars is for the label values.

smg247:
/hold I want to watch the rollout when I have time to verify it is working correctly
Since this is retesting it will be about 2.5 hours if I let it merge now, and that puts it right at 5 pm eastern, not a good time to have a potential issue ;)

PR #3064, opened 2022-10-05
Update formatting to prep for golang version 1.19

smg247:
I know that we aren't using `go` version `1.19` yet, but I started using it locally and it seems `gofmt` likes to format comments differently now. This change should be backwards compatible and will just be one less thing we have to do when we bump versions.

/cc @openshift/test-platform

### DIFF
--- a/pkg/util/sync.go
+++ b/pkg/util/sync.go
@@ -63,11 +63,11 @@ func ProduceMap(
 // The pipeline is constituted of a single producer, a single reducer, and `n`
 // mapper workers, i.e.:
 //
-//       producer
-//      / /   \ \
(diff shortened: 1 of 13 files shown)


pkg/util/sync.go
```diff
@@ -63,11 +63,11 @@ func ProduceMap(
...
-//       producer
-//      / /   \ \
-//     m m  …  m m ---> done
-//      \ \   / /        |
-//       reducer <-------'
+//	  producer
+//	 / /   \ \
+//	m m  …  m m ---> done
+//	 \ \   / /        |
+//	  reducer <-------'
```
bbguimaraes, line 70: `>:(`

PR #3068, opened 2022-10-06
Allow setting always_run in the ci-operator config

zaneb:
If there is an existing job definition then the prow job generation preserves the value of the always_run setting. However, when generating a new job it is always set to true unless forced to false by a run_if_changed or skip_if_always_changed directive in the config.

The result of this is that when copying or renaming job configs, jobs that were intended to only ever be run manually always get enabled on every PR and nobody ever notices before merging the configuration change.

Allow setting "always_run: false" in the config so that we at least have the option to set it in the config and thus have it preserved when jobs are renamed or duplicated.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -687,6 +687,9 @@ type TestStepConfiguration struct {
 	// ClusterClaim claims an OpenShift cluster and exposes environment variable ${KUBECONFIG} to the test container
 	ClusterClaim *ClusterClaim `json:"cluster_claim,omitempty"`
 
+	// AlwaysRun can be set to false to disable running the job on every PR
+	AlwaysRun *bool `json:"always_run,omitempty"`
+
 	// RunIfChanged is a regex that will result in the test only running if something that matches it was changed.
 	RunIfChanged string `json:"run_if_changed,omitempty"`
 
--- a/pkg/api/zz_generated.deepcopy.go
+++ b/pkg/api/zz_generated.deepcopy.go
@@ -1847,6 +1847,11 @@ func (in *TestStepConfiguration) DeepCopyInto(out *TestStepConfiguration) {
 		*out = new(ClusterClaim)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.AlwaysRun != nil {
+		in, out := &in.AlwaysRun, &out.AlwaysRun
+		*out = new(bool)
+		**out = **in
+	}
 	if in.Timeout != nil {
 		in, out := &in.Timeout, &out.Timeout
 		*out = new(v1.Duration)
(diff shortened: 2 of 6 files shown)


droslean:
@zaneb 

> The result of this is that when copying or renaming job configs, jobs that were intended to only ever be run manually always get enabled on every PR and nobody ever notices before merging the configuration change.

This is not true. When you rename/remove a job, there is another component that reconciles the jobs in every PR. Can you elaborate more on what problem this PR is trying resolve?

zaneb:
> Can you elaborate more on what problem this PR is trying resolve?

Sure. You have a config with neither `run_if_changed` nor `skip_if_only_changed`. You generate the job and manually set `always_run: false`.

Now you manually copy the config to another branch (either a release branch, or another variant), either by copying the whole file or by copying-and-pasting that job config into another file. You run `make jobs` to generate the job, which it does with `always_run: true`. You fail to notice this among the 50 lines of new YAML. Now the most unimportant jobs are running all the time.

Some examples of having to fix the config after this happened:

* https://github.com/openshift/release/pull/32652 (due to copy)
* https://github.com/openshift/release/pull/32911 (due to rename)

If it were possible to set `always_run: false` _in the config_ (as it is possible to set `run_if_changed` or `skip_if_onlys_changed`), then `make jobs` would have generated the expected on-demand job.

droslean:
@zaneb You can just set `optional: true` and then `make jobs`

zaneb:
> You can just set `optional: true` and then `make jobs`

That would set `always_run: false`? It doesn't look that way from the code:

```
AlwaysRun: opts.runIfChanged == "" && opts.skipIfOnlyChanged == "",
```

droslean:
> > You can just set `optional: true` and then `make jobs`
> 
> That would set `always_run: false`? It doesn't look that way from the code:
> 
> ```
> AlwaysRun: opts.runIfChanged == "" && opts.skipIfOnlyChanged == "",
> ```

Yea it will actually. Of course this is controlled also by the other options, but if the job is optional, we will generate `always_run: false`. 

Here are some examples:
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/baremetal-operator/openshift-baremetal-operator-master.yaml#L79-L81    ----->    https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/baremetal-operator/openshift-baremetal-operator-master-presubmits.yaml#L79

https://github.com/openshift/release/blob/master/ci-operator/config/openshift/openshift-controller-manager/openshift-openshift-controller-manager-master.yaml#L68-L73 --->  https://github.com/openshift/release/blob/master/ci-operator/jobs/openshift/openshift-controller-manager/openshift-openshift-controller-manager-master-presubmits.yaml#L150

zaneb:
> Yea it will actually. Of course this is controlled also by the other options, but if the job is optional, we will generate `always_run: false`.

OK, I just tried it and... it didn't :man_shrugging:

droslean:
> > Yea it will actually. Of course this is controlled also by the other options, but if the job is optional, we will generate `always_run: false`.
> 
> OK, I just tried it and... it didn't man_shrugging

Do you mean that with `optional: true` we generate the job with `always_run: true`

zaneb:
> Do you mean that with `optional: true` we generate the job with `always_run: true`

Correct, that's what came out of `make jobs`.
Which seems right to me - it's probably fair to say that on-demand jobs should be optional by default, but I don't think the reverse is true.

droslean:
@zaneb So it makes sense to allow people to control the always_run field. Please can you make sure to update also the docs?

/lgtm
/approve

zaneb:
Docs change: https://github.com/openshift/ci-docs/pull/332

PR #3075, opened 2022-10-10
`pj-rehearse`: create external plugin server

smg247:
This adds a server to `pj-rehearse` to act as an external prow plugin. Initially, we will have `preCheck` and `commentOnPrCreation` turned off to do a soft-rollout of this change. Eventually, these will both be turned on.

When a PR is created in configured repos, the plugin determines the affected jobs and comments like:
Screen Shot 2022-10-10 at 3 29 15 PM

When a user adds a comment like `/pj-rehearse` or any of the similar commands, the jobs are triggered.

When a user comments `/pj-rehearse ack` or `/pj-rehearse skip` the `rehearsals-ack` label is added.

For: https://issues.redhat.com/browse/DPTP-3092

/cc @openshift/test-platform @droslean

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -1,16 +1,25 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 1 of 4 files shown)


smg247:
The `e2e` failure is a flake, but I'm not going to bother getting it to pass until I get an `lgtm` here.

droslean (approved):

PR #3080, opened 2022-10-12
DPTP-2428 Don't send acknowledged cards to Intake

psalajova:
If the last comment left on a new card is by TP Intake, this card won't be included in the intake digest on Slack.

for https://issues.redhat.com/browse/DPTP-2428
/cc smg247


smg247:
/override ci/prow/e2e

This test is flaky and is consistently failing right now.

PR #3082, opened 2022-10-12
Increase timeout while waiting the secret creation of the service account

droslean:
/cc @danilo-gemoli @hongkailiu /cc @jupierce @openshift/test-platform  

In theory, there is no way that we create the pod before the actual secret is created. If this is the case, then Kubernetes or Openshift has a core bug in etcd. 

The only case where we stop polling is if we time out. This can happen if the dockercfgcontroller will rollout or the cluster is stressed-up or we are upgrading the cluster.

The fix just increases the timeout from 1 minute to 10 minutes.

It also reverts https://github.com/openshift/ci-tools/pull/3019 and adds one more commit with the timeout change

### DIFF
--- a/pkg/steps/multi_stage/init.go
+++ b/pkg/steps/multi_stage/init.go
@@ -108,14 +108,7 @@ func (s *multiStageTestStep) setupRBAC(ctx context.Context) error {
 	labels := map[string]string{MultiStageTestLabel: s.name}
 	ns := s.jobSpec.Namespace()
 	m := meta.ObjectMeta{Namespace: ns, Name: s.name, Labels: labels}
-	sa := &coreapi.ServiceAccount{
-		ObjectMeta: m,
-		ImagePullSecrets: []coreapi.LocalObjectReference{
-			{
-				Name: api.RegistryPullCredentialsSecret,
-			},
-		},
-	}
+	sa := &coreapi.ServiceAccount{ObjectMeta: m}
 	role := &rbacapi.Role{
 		ObjectMeta: m,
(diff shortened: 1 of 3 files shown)


hongkailiu:
> In theory, there is no way that we create the pod before the actual secret is created. If this is the case, then Kubernetes or Openshift has a core bug in etcd.

> The only case where we stop polling is if we time out. This can happen if the dockercfgcontroller will rollout or the cluster is stressed-up or we are upgrading the cluster.

We do not know what caused the issue: Might be a bug to etcd or something else.
https://issues.redhat.com/browse/DPTP-3134
What we did is to add another pull secret (this is low cost) to mitigate the problem from our side.

droslean:
It is quite unlikely that there is a bug in etcd in this case since we have set a 1min timeout. That means that for some reason (maybe upgrade or node lost) we pass the 1 minute and we never created the secret.

danilo-gemoli:
@droslean unfortunately we have no proof the timeout expired and increasing it doesn't solve the problem either, it just makes it less likely to happen, assuming **that** being the real problem; but again, no evidences were found.
What we did in #3019 was a mere mitigation, a parachute for Pods that have been created without a `SA` to be referenced.
@hongkailiu We should close [DPTP-3134](https://issues.redhat.com/browse/DPTP-3134) and eventually consider this PR in the future, when we will (hopefully) understand the root cause.

droslean:
@danilo-gemoli I believe that there is proof. The fact that the secret was missing means only 2 things. First, the code timeout didn't create the secret (the timeout was so low). Second Openshift etcd bug which is unlikely.

PR #3086, opened 2022-10-14
pr-reminder: show when the latest pr update is less than a day old

psalajova:
Adds a :new: icon to the PR if the PR was updated in the last 24 hours.

/cc smg247

### DIFF
--- a/cmd/pr-reminder/main.go
+++ b/cmd/pr-reminder/main.go
@@ -221,6 +221,13 @@ func (p prRequest) link() string {
 }
 
 func (p prRequest) createdUpdatedMessage() string {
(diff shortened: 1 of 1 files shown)


cmd/pr-reminder/main.go
```diff
@@ -221,18 +221,26 @@ func (p prRequest) link() string {
 }
 
 func (p prRequest) createdUpdatedMessage() string {
+	if time.Since(p.LastUpdated).Hours() <= 24 {
```
smg247, line 224: Let's actually move the emoji to the front of the line so it is more obvious. This will also help to reduce the code duplication here.

PR #3078, opened 2022-10-12
Satisfy new linter warnings

bbguimaraes:
I finally managed to get `golangci-lint` working on my machine again
(apparently, it's very sensitive about the Go version on the system, building
from source made it work again).

Fix a few new issues identified by this new version (`v1.48.0`).

PR #3087, opened 2022-10-14
sprint-automation: address another linter warning

bbguimaraes:
This one only started being identified here after
https://github.com/openshift/ci-tools/pull/3078, for whatever reason.


smg247:
That actually is new code from: https://github.com/openshift/ci-tools/pull/3080.
/lgtm

PR #3088, opened 2022-10-14
pkg/config: add test for registry change detection

bbguimaraes:
This pull request adds tests for the existing implementation of change detection
for registry components.  Perfectly valid infrastructure was already in place
for profile/template/etc. tests (from before that implementation existed), so
this new test is fairly trivial.

37c63e53f tests the old code which detects reference/chain/workflow changes.
Some minor implementation improvements follow, then 3953052d9 extends the test
to also cover the temporary implementation for observer pod introduced in
https://github.com/openshift/ci-tools/pull/3061.

I have a proper graph node type with the correct edges for observers
implemented, but have been wrestling with our `pj-rehearse` integration tests,
so I'm separating this work into its own PR.


droslean (approved):

bbguimaraes:

PR #3095, opened 2022-10-18
add ibmcloud qe cluster profile

MayXuQQ:
add ibmcloud qe cluster profile

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1169,6 +1169,7 @@ const (
 	ClusterProfileGCPLoggingCRIO        ClusterProfile = "gcp-logging-crio"
 	ClusterProfileGCP2                  ClusterProfile = "gcp-openshift-gce-devel-ci-2"
 	ClusterProfileIBMCloud              ClusterProfile = "ibmcloud"
+	ClusterProfileIBMCloudQE            ClusterProfile = "ibmcloud-qe"
 	ClusterProfilePOWERVS               ClusterProfile = "powervs"
 	ClusterProfileLibvirtPpc64le        ClusterProfile = "libvirt-ppc64le"
 	ClusterProfileLibvirtS390x          ClusterProfile = "libvirt-s390x"
@@ -1247,6 +1248,7 @@ func ClusterProfiles() []ClusterProfile {
 		ClusterProfileGCPQE,
 		ClusterProfileHyperShift,
 		ClusterProfileIBMCloud,
+		ClusterProfileIBMCloudQE,
 		ClusterProfilePOWERVS,
 		ClusterProfileKubevirt,
 		ClusterProfileLibvirtPpc64le,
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -1343,6 +1345,8 @@ func (p ClusterProfile) ClusterType() string {
 		return string(CloudGCP)
 	case ClusterProfileIBMCloud:
 		return "ibmcloud"
+	case ClusterProfileIBMCloudQE:
+		return "ibmcloud-qe"
```
jianlinliu, line 1349: Here it is returning `ClusterProfile`, so the value should be "ibmcloud", and this case should be merged to 1346 line.

pkg/api/types.go
```diff
@@ -1597,7 +1603,7 @@ func (p ClusterProfile) Secret() string {
 // LeaseTypeFromClusterType maps cluster types to lease types
 func LeaseTypeFromClusterType(t string) (string, error) {
 	switch t {
-	case "aws", "aws-arm64", "aws-c2s", "aws-china", "aws-usgov", "aws-sc2s", "aws-osd-msp", "aws-outpost", "alibaba", "azure-2", "azure4", "azure-arc", "azure-arm64", "azurestack", "azuremag", "equinix-ocp-metal", "gcp", "libvirt-ppc64le", "libvirt-s390x", "nutanix", "nutanix-qe", "openstack", "openstack-osuosl", "openstack-vexxhost", "openstack-ppc64le", "vsphere", "ovirt", "packet", "packet-edge", "powervs", "kubevirt", "aws-cpaas", "osd-ephemeral":
+	case "aws", "aws-arm64", "aws-c2s", "aws-china", "aws-usgov", "aws-sc2s", "aws-osd-msp", "aws-outpost", "alibaba", "azure-2", "azure4", "azure-arc", "azure-arm64", "azurestack", "azuremag", "equinix-ocp-metal", "gcp", "libvirt-ppc64le", "libvirt-s390x", "nutanix", "nutanix-qe", "openstack", "openstack-osuosl", "openstack-vexxhost", "openstack-ppc64le", "vsphere", "ovirt", "packet", "packet-edge", "powervs", "kubevirt", "aws-cpaas", "osd-ephemeral", "ibmcloud":
```
jianlinliu, line 1606: Per the existing `ibmcloud` cluster profile, it did not add `ibmcloud`, I guess for the new "ibmcloud-qe" cluster profile, does not need it either.

pkg/api/types.go
```diff
@@ -1343,6 +1345,8 @@ func (p ClusterProfile) ClusterType() string {
 		return string(CloudGCP)
 	case ClusterProfileIBMCloud:
 		return "ibmcloud"
+	case ClusterProfileIBMCloudQE:
```
jianlinliu, line 1348: We can merge this `case` onto line 1346

PR #3097, opened 2022-10-18
Extend grace period for observers

jmguzik:
Extend grace period for observers to 1h until param gets exposed in the user configuration


danilo-gemoli:
This PR is following the same path as #3092 
We are going to better refine those changes by introducing two new fields into the `ci-operator` configuration, at least one PR will follow.

/lgtm

bbguimaraes:
=)

PR #3101, opened 2022-10-19
`pj-rehearse`: fix incorrect ref issues with plugin

smg247:
I found a couple of issues related to the refs when testing out the new plugin:

1. Many modified jobs are found that weren't actually present in the PR. This is due to simply using the base SHA from the `PullRequest` rather than determining the SHA for the current head. See [example comment on test PR](https://github.com/openshift/release/pull/33262#issuecomment-1284368180).
2. Triggered rehearsal jobs don't contain the proper org in the `refs`. They currently have the `org` set to that of the fork. I didn't notice this with initial testing as I was testing it in my fork. We need to use the Base org instead.

### DIFF
--- a/cmd/pj-rehearse/server.go
+++ b/cmd/pj-rehearse/server.go
@@ -33,6 +33,7 @@ type githubClient interface {
 	CreateComment(owner, repo string, number int, comment string) error
 	AddLabel(org, repo string, number int, label string) error
 	GetPullRequest(org, repo string, number int) (*github.PullRequest, error)
+	GetRef(org, repo, ref string) (string, error)
 }
 
 type server struct {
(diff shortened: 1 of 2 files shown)


cmd/pj-rehearse/server.go
```diff
@@ -274,6 +281,16 @@ func (s *server) handleIssueComment(l *logrus.Entry, event github.IssueCommentEv
 	}
 }
 
+func (s *server) getCandidate(pullRequest github.PullRequest) (rehearse.RehearsalCandidate, error) {
+	// We can't just use the base SHA from the pullRequest as there may be changes to the branches head in the meantime
+	repo := pullRequest.Base.Repo
+	baseSHA, err := s.ghc.GetRef(repo.Owner.Login, repo.Name, fmt.Sprintf("heads/%s", pullRequest.Base.Ref))
```
smg247, line 287: This method is named oddly. It actually returns a SHA:
https://github.com/kubernetes/test-infra/blob/5022c525793a6f4a741ae47f45e5049b4ec4ad5c/prow/github/client.go#L3359-L3360

droslean (approved):

PR #3103, opened 2022-10-20
retester: add ready endpoints

hongkailiu:
Expect to help to reduce the alerts like https://coreos.slack.com/archives/CHY2E1BL4/p1666295303941069


hongkailiu:
/hold

Maybe it is caused by something else.

PR #3112, opened 2022-10-25
integration: `pj-rehearse` cleanup

bbguimaraes:
This is an independent PR with general improvements to the `pj-rehearse`
integration test suite.  I initially planed to add integration tests following
the unit tests added in https://github.com/openshift/ci-tools/pull/3088, but
found the existing suite in such a poor condition that this preparatory work was
necessary.  It is separated here for independent review.

All commits pass unit and integration tests and try to minimize changes as much
as possible.


bbguimaraes:

PR #3116, opened 2022-10-26
`pj-rehearse`: plugin: don't comment empty job table and streamline acknowledgement

smg247:
We don't want to ever comment an empty job table, if no rehearsable jobs are found we should just say that and immediately ack the rehearsals. I also refactored the code to de-duplicate the acknowledgement.

We also need to fail when the rebase fails. This utilizes the returned `bool` of `MergeWithStrategy` which will be false if the rebase failed.


droslean (approved):

smg247:
Observer e2e is still flaky

PR #3123, opened 2022-11-01
pod-scaler admission: add workload type

psalajova:
This adds workload type for the Prometheus metric (`pod_scaler_admission_high_determined_memory`) to make determining the origin of workloads easier.

for https://issues.redhat.com/browse/DPTP-2929
/cc smg247

### DIFF
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -299,8 +299,5 @@ func determineWorkloadType(annotations, labels map[string]string) string {
 	if _, isProwjob := labels["prow.k8s.io/job"]; isProwjob {
 		return "prowjob"
 	}
-	if _, isRehearsal := labels[rehearse.Label]; isRehearsal {
-		return "rehearsal"
-	}
 	return "undefined"
 }
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -197,7 +197,7 @@ func mutatePodLabels(pod *corev1.Pod, build *buildv1.Build) {
 }
 
 // useOursIfLarger updates fields in theirs when ours are larger
-func useOursIfLarger(allOfOurs, allOfTheirs *corev1.ResourceRequirements, podName string, reporter results.PodScalerReporter, logger *logrus.Entry) {
+func useOursIfLarger(allOfOurs, allOfTheirs *corev1.ResourceRequirements, podName, workloadType string, reporter results.PodScalerReporter, logger *logrus.Entry) {
 	for _, item := range []*corev1.ResourceRequirements{allOfOurs, allOfTheirs} {
 		if item.Requests == nil {
 			item.Requests = corev1.ResourceList{}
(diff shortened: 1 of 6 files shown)


cmd/pod-scaler/admission.go
```diff
@@ -273,36 +273,34 @@ func preventUnschedulable(resources *corev1.ResourceRequirements, cpuCap int64,
...
+		return "build"
+	}
+	if _, isProwjob := labels["prow.k8s.io/job"]; isProwjob {
+		return "prowjob"
+	}
+	if _, isRehearsal := labels[rehearse.Label]; isRehearsal {
+		return "rehearsal"
 	}
-	return fmt.Sprintf("%s-%s", podName, containerName)
+	return "undefined"
```
psalajova, line 305: not sure about "undefined", but I didn't want to return an empty string

cmd/pod-scaler/admission.go
```diff
@@ -273,36 +273,34 @@ func preventUnschedulable(resources *corev1.ResourceRequirements, cpuCap int64,
...
 	}
+	mutateResources(pod.Spec.InitContainers)
+	mutateResources(pod.Spec.Containers)
 }
 
-// determineName returns the string that will be used in Prometheus metrics to identify the workload
-func determineName(podLabels map[string]string, podName, containerName string) string {
-	if value, exists := podLabels["prow.k8s.io/job"]; exists {
-		return value
+// determineWorkloadType returns the workflow type
```
smg247, line 294:
```suggestion
// determineWorkloadType returns the workload type to be used in metrics
```

cmd/pod-scaler/admission.go
```diff
@@ -273,36 +273,34 @@ func preventUnschedulable(resources *corev1.ResourceRequirements, cpuCap int64,
...
+func determineWorkloadType(annotations, labels map[string]string) string {
+	if _, isBuildPod := annotations[buildv1.BuildLabel]; isBuildPod {
+		return "build"
+	}
+	if _, isProwjob := labels["prow.k8s.io/job"]; isProwjob {
+		return "prowjob"
+	}
+	if _, isRehearsal := labels[rehearse.Label]; isRehearsal {
```
smg247, line 302: I am curious why you decided to call out rehearsals specifically, they are just another prowjob. Will this logic ever even be reached?

If we do want to call them out this will need to be before the check for prowjobs.

cmd/pod-scaler/admission.go
smg247: I think that is fine for now.

smg247 (changes requested):

cmd/pod-scaler/admission.go
psalajova: Oh I didn't know rehearsals were a type of prowjob. In that case, we probably don't need this, I just found how to check for the rehearsals in the code above and thought it could be helpful.

smg247: Yeah, lets not call out rehearsals here.

PR #3124, opened 2022-11-01
Also skip okd jobs for gathering disruption/alert data.

dgoodwin:
[TRT-646](https://issues.redhat.com//browse/TRT-646)

### DIFF
--- a/pkg/jobrunaggregator/jobtableprimer/generate_jobnames.go
+++ b/pkg/jobrunaggregator/jobtableprimer/generate_jobnames.go
@@ -223,6 +223,11 @@ func (o *GenerateJobNamesOptions) Run(ctx context.Context) error {
 				continue
 			}
 
+			// OKD jobs are not something we monitor and keep slipping into our disruption data skewing results quite badly.
+			if strings.Contains(curr.Name, "okd") {
+				continue
+			}
+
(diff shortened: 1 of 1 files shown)


pkg/jobrunaggregator/jobtableprimer/generate_jobnames.go
```diff
@@ -223,6 +223,11 @@ func (o *GenerateJobNamesOptions) Run(ctx context.Context) error {
 				continue
 			}
 
+			// OKD jobs are not something we monitor and keep slipping into our disruption data skewing results quite badly.
+			if strings.Contains(curr.Name, "okd") {
```
stbenjam, line 227: Super nitpick: I kind of like a dash. Is there a chance okd could appear in the middle of a word? Probably not but maybe better to be safe?

dgoodwin: I couldn't think of a word but you did on slack, I'll add a leading dash just incase it could appear at end, and send in a followup.

dgoodwin: https://github.com/openshift/ci-tools/pull/3126/files

PR #3126, opened 2022-11-02
skip okd jobs 2

dgoodwin:
- Also skip okd jobs for gathering disruption/alert data.
- Safer detection of okd jobs to skip.


dgoodwin:
/assign @xueqzhan 

Sorry ken small followup from the last one just to be a little safer.

PR #3129, opened 2022-11-03
cluster-display: add #standby for pools

hongkailiu:
One of the things to follow up from the incident last Friday.


wking:
[Lovely][1]; thanks :)

image

[1]: https://docs.ci.openshift.org/docs/how-tos/cluster-claim/#existing-cluster-pools

PR #3132, opened 2022-11-04
test-images-distributor: distribute multiarch IS

danilo-gemoli:
This PR is related to [DPTP-3160](https://issues.redhat.com/browse/DPTP-3160)
`test-images-distributor` now distributes `ISTag`s coming from multiarch namespaces to the proper clusters, without modifying the existing behaviour.

Examples: 
ISTag `ci-arm64/ci-operator:latest` would be replicated on `arm01` but it won't on `build0x`
ISTag `ci/ci-operator:latest` would be replicated **only** on `amd64` clusters: `build0x`

/cc @droslean  @hongkailiu

### DIFF
--- a/pkg/controller/test-images-distributor/test_images_distributor.go
+++ b/pkg/controller/test-images-distributor/test_images_distributor.go
@@ -35,6 +35,38 @@ import (
 
 const ControllerName = "test_images_distributor"
 
+// Everything but default arch (amd64)
+var multiarchSuffixToCluster = map[string]sets.String{
+	"-arm64": sets.NewString(string(api.ClusterARM01)),
+}
+
+func isMultiarchNamespace(namespace string) bool {
+	for suffix := range multiarchSuffixToCluster {
+		if strings.HasSuffix(namespace, suffix) {
+			return true
+		}
+	}
+	return false
+}
+
+func isNamespaceAllowedOnCluster(namespace, cluster string) bool {
+	for suffix, clusters := range multiarchSuffixToCluster {
+		if strings.HasSuffix(namespace, suffix) {
+			return clusters.Has(cluster)
+		}
+	}
+	return false
+}
+
+func isAmd64Cluster(cluster string) bool {
+	for _, clusters := range multiarchSuffixToCluster {
+		if clusters.Has(cluster) {
+			return false
+		}
+	}
+	return true
+}
+
 func AddToManager(mgr manager.Manager,
 	registryClusterName string,
 	registryManager manager.Manager,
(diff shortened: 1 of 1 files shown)


pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -291,6 +330,22 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
...
+	// One of the following is allowed:
+	// - multiarch namespaces to distribute on the proper non-amd64 clusters (ex.: ci-arm64 on arm01)
+	// or
+	// - non-multiarch namespaces to distribute on any amd64 clusters (ex.: ci on build03)
+	if isMultiarchNamespace(decoded.Namespace) {
+		if !isNamespaceAllowedOnCluster(decoded.Namespace, cluster) {
+			log.Debug("multiarch imageStreamTag not allowed on cluster")
+			return nil
+		}
+	} else {
```
droslean, line 342: you mean `else if` here?

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -35,6 +35,45 @@ import (
 
 const ControllerName = "test_images_distributor"
 
+// Everything but default arch (amd64)
+var multiarchSuffixToCluster = map[string][]string{
```
droslean, line 39: That would be easier if it was `map[string]sets.String`

pkg/controller/test-images-distributor/test_images_distributor.go
danilo-gemoli: Yes, initially I wrote only one `if` condition but it was unreadable. This is the easiest form to understand, if we agree on that I can refine it a little bit.

pkg/controller/test-images-distributor/test_images_distributor.go
```diff
@@ -291,6 +323,20 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *
 	*log = *log.WithField("cluster", cluster).WithField("namespace", decoded.Namespace).WithField("name", decoded.Name)
 	log.Info("Starting reconciliation")
 
+	// One of the following is allowed:
+	// - multiarch namespaces to distribute on the proper non-amd64 clusters (ex.: ci-arm64 on arm01)
+	// or
+	// - non-multiarch namespaces to distribute on any amd64 clusters (ex.: ci on build03)
+	if isMultiarchNamespace(decoded.Namespace) {
```
droslean, line 330: Why do we need to check both cases? If the namespace is multi-arch so be it, otherwise it is amd64 by default.

droslean (approved):

pkg/controller/test-images-distributor/test_images_distributor.go
droslean: Oh, that extra check will deny the distribution of the amd64 images to arm01... Nice.

PR #3136, opened 2022-11-05
Enable taking cluster-claim from workflows

osherdp:
Up until now, cluster-claim stanza was only applicable on the job-config level. It is sometimes desired to configure the cluster-claim parameters in one place and reuse it across different job definitions.

This PR adds a propagation of cluster-claim from a workflow if it's defined, with respect to it being overridden by job configuration if it exists.

Implements [DPTP-2845](https://issues.redhat.com/browse/DPTP-2845)


osherdp:
Seems like an irrelevant flake 
/retest

hongkailiu (commented):
Thanks for the PR.
The code looks good to me.

- Could we add an e2e test for the new cases?
https://github.com/openshift/ci-tools/blob/master/test/e2e/multi-stage/e2e_test.go

- cluster_profile and cluster_claim are mutually exclusive
https://docs.ci.openshift.org/docs/architecture/ci-operator/#testing-with-a-cluster-from-a-cluster-pool
This PR provides another way of introducing cluster_claim.
Is there any validation which prevent a workflow from using both?

(I just realized that) An even more implicit case is using one in the workflow and the other in the ci-op's config.
This one might be hard to implement and might have violation before this PR, i.e., define cluster_profile in a workflow, and use the workflow without overriding the cluster_profile and use cluster_claim in the test.

I will ask around if we have validation on workflows.

pkg/registry/resolver_test.go
```diff
@@ -711,6 +713,202 @@ func TestResolve(t *testing.T) {
...
+				Post: []api.TestStep{{
+					LiteralTestStep: &api.LiteralTestStep{
+						As:       "ipi-teardown",
+						From:     "installer",
+						Commands: "openshift-cluster destroy",
+						Resources: api.ResourceRequirements{
+							Requests: api.ResourceList{"cpu": "1000m"},
+							Limits:   api.ResourceList{"memory": "2Gi"},
+						}},
+				}},
```
bbguimaraes, line 778: These are just noise in the definitions of the input/output, they can be removed in all test cases.

osherdp:
> > [#3136 (review)](https://github.com/openshift/ci-tools/pull/3136#pullrequestreview-1171033978)
> 
> Configuration files are resolved before they are validated, so the existing validation should take care of this new case.
> 
> We should also add a `prowgen` test (see `test/integration/ci-operator-prowgen`) to guarantee the jobs are given the required credentials.

Added integration tests
I also had to fix the code, so good catch!

osherdp:
/hold
until finishing reviewing process

osherdp:
@bbguimaraes @hongkailiu can you help me understand why the e2e test fails?
I assume it means my added logic doesn't really work, but I'm not sure what it's missing looking at the gathered logs

bbguimaraes:
Well, the specific failure there is because you are looking for a string which is not present in the output (seems to be just a problem of the literal string "4.6" vs. "4.7").

But more importantly, nothing in the E2E tests seems to be testing the claiming process, and from the output it seems not be happening at all — cf.: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_loki/75/pull-ci-openshift-loki-main-operator-install/1592991183250919424#1:build-log.txt%3A30.

osherdp:
> Well, the specific failure there is because you are looking for a string which is not present in the output (seems to be just a problem of the literal string "4.6" vs. "4.7").
> 
> But more importantly, nothing in the E2E tests seems to be testing the claiming process, and from the output it seems not be happening at all — cf.: https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_loki/75/pull-ci-openshift-loki-main-operator-install/1592991183250919424#1:build-log.txt%3A30.

correct. I think the 4.6 vs 4.7 is another thing that should happen (image override) in the case of cluster-claim, but the root-cause here is that no cluster-claim gets propagated at all
Meaning:
* I'll add some more lines to verify claiming actually happens (probably add that to the other tests as well)
* Before point 1, I might need your help in understanding how to debug it to make claiming happen. Maybe the resolver isn't being triggered in this e2e test?

pkg/api/types.go
```diff
@@ -1052,6 +1054,8 @@ type DependencyOverrides map[string]string
 type MultiStageTestConfigurationLiteral struct {
 	// ClusterProfile defines the profile/cloud provider for end-to-end test steps.
 	ClusterProfile ClusterProfile `json:"cluster_profile"`
+	// ClusterClaim claims an OpenShift cluster and exposes environment variable ${KUBECONFIG} to the test container
+	ClusterClaim *ClusterClaim `json:"cluster_claim,omitempty"`
```
bbguimaraes, line 1058: This is the problem: this field does not belong here.  `cluster_claim` is a field at the test level (https://docs.ci.openshift.org/docs/architecture/ci-operator#testing-with-a-cluster-from-a-cluster-pool), not at the multi-stage test definition level (i.e. the `steps` field).

`MultiStageTestConfiguration` should have a cluster claim field to enable what you are proposing (we can live with the asymmetry, IMO), but the `*Literal` counterpart should not.  No multi-stage test definition should have a `cluster_claim` (the confusion here in the tests is proof).  We'll likely need to adjust the resolution interface, since it is currently strictly `MultiStageTestConfiguration` → `*Literal`.

osherdp:
Closing as I can't find the time to complete this feature
I'll be glad if someone (maybe the future me) have the time to continue with this effort (probably not too much is missing, but I've got to a bit of a rabbit hole in regard to latest comment)

PR #3137, opened 2022-11-08
[prowgen] merge configuration defaults when both org and repo configs are specified

droslean:
/cc @openshift/test-platform 


pkg/config/load.go
```diff
@@ -56,6 +56,24 @@ func (p *Prowgen) Validate() error {
 	return utilerrors.NewAggregate(errs)
 }
 
+func (p *Prowgen) MergeDefaults(defaults *Prowgen) error {
```
smg247, line 59: this will never actually error. you can remove it from the sig.

PR #3138, opened 2022-11-08
`check-gh-automation`: stop mis-categorizing the prowgen config as a repo

smg247:
We shouldn't add prowgen config files that are named `.config.prowgen` as repos to check in the tool.

For: https://issues.redhat.com/browse/DPTP-3204

### DIFF
--- a/cmd/check-gh-automation/main.go
+++ b/cmd/check-gh-automation/main.go
@@ -196,6 +196,10 @@ func gatherModifiedRepos(releaseRepoPath string, logger *logrus.Entry) []string 
 	for _, c := range configs {
 		path := strings.TrimPrefix(c, config.CiopConfigInRepoPath+"/")
 		split := strings.Split(path, "/")
+		if split[1] == ".config.prowgen" {
+			continue
+		}
+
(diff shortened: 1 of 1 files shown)


cmd/check-gh-automation/main.go
```diff
@@ -196,7 +196,11 @@ func gatherModifiedRepos(releaseRepoPath string, logger *logrus.Entry) []string
 	for _, c := range configs {
 		path := strings.TrimPrefix(c, config.CiopConfigInRepoPath+"/")
 		split := strings.Split(path, "/")
-		orgRepos.Insert(fmt.Sprintf("%s/%s", split[0], split[1]))
+		// Check if this is a config file like ".config.prowgen" rather than a repo
+		isConfigFile := strings.HasPrefix(split[1], ".")
```
droslean, line 200: We could just `continue` if the name is `.config.prowgen`

smg247: Yeah, thats fine too. I wanted to avoid issues if we created other types of config files in the future, but I suppose that is not very likely so I can just tailor it to this case.

droslean (approved):

PR #3139, opened 2022-11-08
Bump to latest available test-infra

osherdp:
I wanted to test another feature (config-sharding for slack-reporter-configs) and saw a couple of unrelated changes that are required. This includes:
* Updating kubernetes/test-infra to the latest version. I also had to pin version of opencontainers/runc to ``v1.0.0-rc9`` because the function ``rsystem.RunningInUserNS`` has been removed since then (and ``github.com/docker/docker`` is still using it).
* GitHub issues are now closed together with a closing reason that can be either "completed" or "not_planned". Tests in this repo have been updated to having that field with "completed" value as we're only closing in the regular way.


osherdp:
Depends on https://github.com/openshift/ci-chat-bot/pull/324 and https://github.com/openshift/release-controller/pull/471

PR #3146, opened 2022-11-09
Transition to use new release disruption data with enough count

xueqzhan:
During the release transition, we will continue using old release data set when new release data set is not enough. 

[TRT-659](https://issues.redhat.com/browse/TRT-659)


pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -11,6 +11,8 @@ import (
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorlib"
 )
 
+const releaseTransitionDataCountThreshold = 60
```
dgoodwin, line 14: Am I correct that this is intended to require at least 60% as many results as we had before? Suggest making this 0.6, maybe renaming to releaseTransitionDataMinThresholdRatio, and godoc to explain that we require at least this many results before we'll flip the data file over to the latest release, otherwise we keep using the old.

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -59,20 +61,34 @@ func (o *JobRunHistoricalDataAnalyzerOptions) Run(ctx context.Context) error {
...
 	if newReleaseUpdate {
-		if err := requireReviewFile("The current data contains previous release version\n"); err != nil {
-			return err
+		newReleaseDataCount := 0
+		for _, data := range newDataMap {
+			if data.GetJobData().Release == currentRelease {
+				newReleaseDataCount++
+			}
+		}
+		if newReleaseDataCount < len(currentDataMap)*releaseTransitionDataCountThreshold/100 {
```
dgoodwin, line 81: The change looks good Ken, our minimum 100 entries looks to happen in the query, so it's already been applied by the time we get to this code, and your logic here should hold, if we didn't have at least 60% of entries with 100+ runs, we stick to old release.

dgoodwin: I would add a very clear output message here for printing, indicate the old and new counts with their release version, and that we require at least X percent, and are sticking with the previous version.

Is there any way we could also get that message into the PR message so the reviewer gets a clear indication as well? They should probably notice anyhow but still, it would be nice if it's not too difficult.

PR #3152, opened 2022-11-15
Handle the global pull secret on OSD clusters

hongkailiu:
https://issues.redhat.com/browse/DPTP-2795

Because of https://issues.redhat.com/browse/OHSS-16577, we want to handle it now and revert https://github.com/openshift/release/blob/0605597e1acbdb663239b03d9d7609cdbce9152c/clusters/app.ci/assets/dptp-controller-manager.yaml#L167

/cc @openshift/test-platform

### DIFF
--- a/pkg/api/secretbootstrap/secretboostrap.go
+++ b/pkg/api/secretbootstrap/secretboostrap.go
@@ -239,3 +239,10 @@ func (c *Config) resolve() error {
 
 	return utilerrors.NewAggregate(errs)
 }
+
+const OSDGlobalPullSecretGroupName = "osd_global_pull_secret"
+
+// OSDGlobalPullSecretGroup returns the list of the OSD cluster names where we need to partially manage the global pull secret
+func (c *Config) OSDGlobalPullSecretGroup() []string {
+	return c.ClusterGroups[OSDGlobalPullSecretGroupName]
+}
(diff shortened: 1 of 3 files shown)


pkg/api/secretbootstrap/secretboostrap.go
```diff
@@ -239,3 +239,10 @@ func (c *Config) resolve() error {
 
 	return utilerrors.NewAggregate(errs)
 }
+
+const OSDGlobalPullSecretGroupName = "osd_global_pull_secret"
+
+// OSDGlobalPullSecretGroup returns the list of the OSD cluster names where we need to partially manage the global pull secret
+func (c *Config) OSDGlobalPullSecretGroup() []string {
+	return c.ClusterGroups[OSDGlobalPullSecretGroupName]
+}
```
bbguimaraes, line 248: The code in the PR LGTM, but I see the pull secrets in all our clusters have the same entries (key names).  Any reason not to do this for all of them?

hongkailiu: Good one.
This is more of doing the things we have to. :) 
1. b01, b02: OCP clusters. Our automation manages the global pull secret as a whole.
2. b03-b05: Will be managed after this PR.
3. arm01, vsphere: not managed by us.
4. hive, app.ci: also OSD. But no e2e jobs run there.

PR #3153, opened 2022-11-15
[prowgen] merge additional_architectures when repo defines them

osherdp:
A continuation of https://github.com/openshift/ci-tools/pull/3137.

This allows resetting the ``additional_architectures`` list in a
repository configuration when it's also defined in the organization
configuration.

Required due to arm64 builds failing for assisted-installer
repositories.


osherdp:
/hold
I don't think it will generate breaking-changes, but just in case... holding

PR #3156, opened 2022-11-18
dptp-controller-manager: use universal watcher

bbguimaraes:
86f1d4db1c added the command-line argument but maintained the usage of the old
clients, which are not appropriate for a `git-sync` mount — they work but are
obscenely inefficient (see production logs).  This change copies the pattern
from `ci-operator-configresolver` (94fd5be7a8) to create agents of the
appropriate type.  Note that the required path manipulation based on the
argument is already present.

/hold


bbguimaraes:
https://issues.redhat.com/browse/DPTP-2988

hongkailiu (approved):

PR #3157, opened 2022-11-18
pkg/steps: handle missing secret in index step

bbguimaraes:
Currently results in a nil pointer dereference:

```
$ ci-operator --namespace bbguimaraes --config optional-operators.yaml --target verify-db-named
…
INFO[2022-11-18T16:37:09Z] Building named-bundle
INFO[2022-11-18T16:37:42Z] Build named-bundle succeeded after 33s
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x1a85a71]

goroutine 970 [running]:
github.com/openshift/ci-tools/pkg/steps.(*indexGeneratorStep).run(0xc000e2c400, {0x241b258, 0xc000940e40})
        /home/bbguimaraes/src/ci-tools/pkg/steps/index_generator.go:73 +0x451
github.com/openshift/ci-tools/pkg/steps.(*indexGeneratorStep).Run(0x0?, {0x241b258, 0xc000940e40})
        /home/bbguimaraes/src/ci-tools/pkg/steps/index_generator.go:37 +0x5e
github.com/openshift/ci-tools/pkg/steps.runStep({0x241b258, 0xc000940e40}, 0xc0009c13b0, 0x0?)
        /home/bbguimaraes/src/ci-tools/pkg/steps/run.go:121 +0xb4
created by github.com/openshift/ci-tools/pkg/steps.Run
        /home/bbguimaraes/src/ci-tools/pkg/steps/run.go:73 +0xd6a
```

The comment (now removed) claims the secret is necessary, but the step works
perfectly fine without it:

```
$ ci-operator --namespace bbguimaraes --config optional-operators.yaml --target verify-db-named
[…]
INFO[2022-11-18T18:26:48Z] Building src-bundle
INFO[2022-11-18T18:26:48Z] Build src-bundle succeeded after 39s
INFO[2022-11-18T18:26:48Z] Building named-bundle
INFO[2022-11-18T18:26:49Z] Build named-bundle succeeded after 33s
INFO[2022-11-18T18:26:49Z] Building ci-index-named-bundle-gen
INFO[2022-11-18T18:27:42Z] Build ci-index-named-bundle-gen succeeded after 53s
[…]
```

In any case, this should not affect CI jobs since the pull secret is already
unconditionally added to all generated jobs.

/hold

### DIFF
--- a/pkg/steps/index_generator.go
+++ b/pkg/steps/index_generator.go
@@ -89,6 +89,12 @@ func (s *indexGeneratorStep) run(ctx context.Context) error {
 	if err != nil {
 		return err
 	}
+	var secrets []buildapi.SecretBuildSource
+	if s.pullSecret != nil {
+		secrets = append(secrets, buildapi.SecretBuildSource{
+			Secret: coreapi.LocalObjectReference{Name: s.pullSecret.Name},
+		})
+	}
 	build := buildFromSource(
 		s.jobSpec, fromTag, s.config.To,
 		buildapi.BuildSource{
@@ -106,9 +112,7 @@ func (s *indexGeneratorStep) run(ctx context.Context) error {
 					}},
 				},
 			},
-			Secrets: []buildapi.SecretBuildSource{{
-				Secret: coreapi.LocalObjectReference{Name: s.pullSecret.Name},
-			}},
+			Secrets: secrets,
 		},
 		fromDigest,
 		"",
@@ -126,9 +130,10 @@ func (s *indexGeneratorStep) run(ctx context.Context) error {
 func (s *indexGeneratorStep) indexGenDockerfile() (string, error) {
 	var dockerCommands []string
 	dockerCommands = append(dockerCommands, "FROM quay.io/operator-framework/upstream-opm-builder AS builder")
-	// pull secret is needed for opm command
-	dockerCommands = append(dockerCommands, "COPY .dockerconfigjson .")
-	dockerCommands = append(dockerCommands, "RUN mkdir $HOME/.docker && mv .dockerconfigjson $HOME/.docker/config.json")
+	if s.pullSecret != nil {
+		dockerCommands = append(dockerCommands, "COPY .dockerconfigjson .")
+		dockerCommands = append(dockerCommands, "RUN mkdir $HOME/.docker && mv .dockerconfigjson $HOME/.docker/config.json")
+	}
 	var bundles []string
 	for _, bundleName := range s.config.OperatorIndex {
 		fullSpec, err := utils.ImageDigestFor(s.client, s.jobSpec.Namespace, api.PipelineImageStream, bundleName)()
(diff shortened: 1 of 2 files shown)


pkg/steps/index_generator_test.go
```diff
@@ -65,16 +67,34 @@ WORKDIR /index-data
 COPY --from=builder index.Dockerfile index.Dockerfile
 COPY --from=builder /database/ database`,
 	}, {
-		name: "multiple bundles",
+		name: "single bundle, no pull secret",
```
bbguimaraes, line 70: The `diff` here is not great.  All existing tests are unchanged except for the extra `pullSecret` field in each `step`.

This new test is the same as the previous, but omits the secret.

hongkailiu:
> The comment (now removed) claims the secret is necessary, but the step works
perfectly fine without it:

Do we want to understand it?
Maybe the testing example does not cover the situation where it is needed.

pkg/steps/index_generator.go
```diff
@@ -89,9 +93,10 @@ func (s *indexGeneratorStep) run(ctx context.Context) error {
 func (s *indexGeneratorStep) indexGenDockerfile() (string, error) {
 	var dockerCommands []string
 	dockerCommands = append(dockerCommands, "FROM quay.io/operator-framework/upstream-opm-builder AS builder")
-	// pull secret is needed for opm command
-	dockerCommands = append(dockerCommands, "COPY .dockerconfigjson .")
-	dockerCommands = append(dockerCommands, "RUN mkdir $HOME/.docker && mv .dockerconfigjson $HOME/.docker/config.json")
+	if s.pullSecret != nil {
+		dockerCommands = append(dockerCommands, "COPY .dockerconfigjson .")
+		dockerCommands = append(dockerCommands, "RUN mkdir $HOME/.docker && mv .dockerconfigjson $HOME/.docker/config.json")
+	}
```
hongkailiu, line 99:
```suggestion
	} else {
	    logrus.Warn("The pull secret is nil in indexGeneratorStep which is probably a result of a non-generated job and thus not mounted onto the build pod")
	}
```

As you pointed out, this should not happen in the generated jobs.
But I would like to log it if it  does happen, to reduce the debugging time.

pkg/steps/index_generator_test.go
hongkailiu: Could we add a new case instead of modifying the existing one?

bbguimaraes: Not entirely since the new implementation alters the content generated for the existing tests, but I've added a duplicate of each previous test which has a pull secret.

pkg/steps/index_generator.go
bbguimaraes: Answering this and the top-level question ("do we want to understand it") here: the comment is incorrect — the secret is not necessary for the `opm` command.

It may be necessary for the specific setup we have (or some of them?), but it's entirely possible to use this functionality without needing a pull secret.

hongkailiu: I think the logging is useful but that is NIT.

Still not clear when the pull secret is needed.
I understand it is not in the scope of this PR.

PR #3161, opened 2022-11-22
ci-operator: allow grace period to be configured in observers

nesty156:
Added `Timeout` and `GracePeriod` to Observer struct. And updated remaining functions to work with new changes.
[DPTP-3184](https://issues.redhat.com/browse/DPTP-3184)

/cc @bbguimaraes


bbguimaraes (commented):
A few minor comments, but that's a great first version.

You'll need to update some automatically generated code because of the new fields which are being added.  That's just a matter of executing the `make` command in the failed job log.

I had also mentioned the validation code, but after reviewing what we have, it's not relevant to observers (we could have some extra ones, even for regular steps, but we don't need to worry about that for this PR).

bbguimaraes (commented):
A few minor comments.  The commit separation looks good, but I see `pkg/api/zz_generated.deepcopy.go` and `pkg/webreg/zz_generated.ci_operator_reference.go` are in the last one.  Move them to the middle commit (the one which adds the fields), otherwise we'll have a commit in the history which does not pass tests, making it harder to go back and test old versions, if necessary.

Also, since this is almost ready: before we can merge this, there are a few pre-requisites as we already have tests using observers.

- Create a pull request for `openshift/release` adding the old default values (`1h`/`24h`) to the existing observer definitions, so they don't get implicitly new behavior after we change the default.  They should be easy to find by just looking for files with `observer` in the name.
- Update the documentation page to mention the new configuration options: https://docs.ci.openshift.org/docs/architecture/ci-operator-internals/observer-pods/.  Ideally, it would be a short description which mentions the values work the same as for regular steps, followed to a link to the documentation page detailing that: https://docs.ci.openshift.org/docs/architecture/timeouts/#step-registry-test-process-timeouts.

PR #3164, opened 2022-11-23
cluster-init: automate members of osd_global_pull_secret group

hongkailiu:
Follow up https://github.com/openshift/ci-tools/pull/3152

### DIFF
--- a/cmd/cluster-init/update_bootstrap_secrets.go
+++ b/cmd/cluster-init/update_bootstrap_secrets.go
@@ -47,6 +47,11 @@ func updateCiSecretBootstrapConfig(o options, c *secretbootstrap.Config) error {
 	for _, groupName := range []string{buildUFarm, "non_app_ci"} {
 		c.ClusterGroups[groupName] = sets.NewString(c.ClusterGroups[groupName]...).Insert(o.clusterName).List()
 	}
+	// build02 and build02 are not OSD clusters and thus they should never be in the group
+	if o.clusterName != string(api.ClusterBuild01) && o.clusterName != string(api.ClusterBuild02) {
+		groupName := secretbootstrap.OSDGlobalPullSecretGroupName
+		c.ClusterGroups[groupName] = sets.NewString(append(c.ClusterGroups[groupName], o.clusterName)...).List()
+	}
 	c.UserSecretsTargetClusters = sets.NewString(c.UserSecretsTargetClusters...).Insert(o.clusterName).List()
 
 	for _, step := range []func(c *secretbootstrap.Config, o options) error{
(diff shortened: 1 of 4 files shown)


cmd/cluster-init/update_bootstrap_secrets.go
```diff
@@ -44,7 +44,11 @@ func updateCiSecretBootstrap(o options) error {
...
 
 func updateCiSecretBootstrapConfig(o options, c *secretbootstrap.Config) error {
-	for _, groupName := range []string{buildUFarm, "non_app_ci"} {
+	for _, groupName := range []string{buildUFarm, "non_app_ci", secretbootstrap.OSDGlobalPullSecretGroupName} {
+		if groupName == secretbootstrap.OSDGlobalPullSecretGroupName &&
+			(o.clusterName == string(api.ClusterBuild01) || o.clusterName == string(api.ClusterBuild02)) {
+			continue
+		}
 		c.ClusterGroups[groupName] = sets.NewString(c.ClusterGroups[groupName]...).Insert(o.clusterName).List()
 	}
```
bbguimaraes, line 53:
```go
for _, groupName := range []string{buildUFarm, "non_app_ci"} {
	c.ClusterGroups[groupName] = sets.NewString(c.ClusterGroups[groupName]...).Insert(o.clusterName).List()
}
if o.clusterName != string(api.ClusterBuild01) && o.clusterName != string(api.ClusterBuild02) {
	groupName := secretbootstrap.OSDGlobalPullSecretGroupName
	c.ClusterGroups[groupName] = sets.NewString(c.ClusterGroups[groupName]...).Insert(o.clusterName).List()
}
```

bbguimaraes: Also, these insertions would be clearer with something along the lines of:

```go
func insert[T any](s []T, i int, x T) []T {
	var z T
	s = append(s, z)
	copy(s[i+1:], s[i:])
	s[i] = x
	return s
}

func insert_sorted_string(s []string, x string) []string {
	if i := sort.SearchStrings(s, x); i == len(s) || s[i] != x {
		s = insert(s, i, x)
	}
	return s
}
```

(and less wasteful, since we currently build a hash table, insert, copy everything to another slice, then sort it, just to add a new element to a list which is known to be sorted)

bbguimaraes: It would be good to have some tests for this merging code, even if we don't implement this ^

bbguimaraes: And to have `build01`/`build02` as a constant somewhere with documentation about why they're excluded here (I suppose it's something more generic, non-OSD DPTP clusters?).

hongkailiu: > these insertions would be clearer 

Hopefully it is better now.

> It would be good to have some tests for this merging code, even if we don't implement this ^

The coverage is in the integration tests, like the most of the function of the tool are.

PR #3166, opened 2022-11-23
Fix dispatcher issue when failover happens and colocation is enabled

jmguzik:
When there is a failover and colocation is enabled (`DetermineE2EByJob`) some jobs are not moved from the broken cluster. This feature detects enabled colocation and replaces it for the time of the incident with optional colocation setting (`IgnoreE2EByJobAssignment`).

`IgnoreE2EByJobAssignment` means the job could be colocated or not; both states are accepted. It was developed to enable gradual colocation (% of jobs in batch PR). The incident is a similar situation, as some of the jobs need to be colocated and others not (these from the failing cluster).

Not using this flag and disabling `DetermineE2EByJob` means a complete rearrangement of jobs, the resulting PR will fallback to not colocated jobs for the time of the incident which may lead to scaling problems and will need to be again reverted after the incident, so it is not the desired direction.

/cc @hongkailiu 

### DIFF
--- a/pkg/dispatcher/config.go
+++ b/pkg/dispatcher/config.go
@@ -208,6 +208,21 @@ func isSSHBastionJob(base prowconfig.JobBase) bool {
 	return false
 }
 
+// IsAnyBuildClusterDisabled returns if any of the build clusters is disabled
+func (config *Config) IsAnyBuildClusterDisabled() bool {
+	if len(config.BuildFarm) == 0 {
+		return true
+	}
+	for provider := range config.BuildFarm {
+		for cluster := range config.BuildFarm[provider] {
+			if config.BuildFarm[provider][cluster].Disabled == true {
+				return true
+			}
+		}
+	}
+	return false
+}
+
 // IsInBuildFarm returns the cloudProvider if the cluster is in the build farm; empty string otherwise.
 func (config *Config) IsInBuildFarm(clusterName api.Cluster) api.Cloud {
 	for cloudProvider, v := range config.BuildFarm {
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -443,6 +443,15 @@ func main() {
 		}
 	}
 
+	if config.DetermineE2EByJob == true && config.IsAnyBuildClusterDisabled() {
+		config.DetermineE2EByJob = false
+		config.IgnoreE2EByJobAssignment = true
+	}
+	if config.IgnoreE2EByJobAssignment == true && !config.IsAnyBuildClusterDisabled() {
+		config.DetermineE2EByJob = true
+		config.IgnoreE2EByJobAssignment = false
+	}
+
 	logrus.Info("Dispatching ...")
 	if err := dispatchJobs(context.TODO(), o.prowJobConfigDir, o.maxConcurrency, config, jobVolumes); err != nil {
 		logrus.WithError(err).Fatal("Failed to dispatch")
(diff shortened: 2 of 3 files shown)


pkg/dispatcher/config.go
```diff
@@ -208,6 +208,21 @@ func isSSHBastionJob(base prowconfig.JobBase) bool {
 	return false
 }
 
+// IsAnyBuildClusterDisabled returns if any of the build clusters is disabled
+func (config *Config) IsAnyBuildClusterDisabled() bool {
+	if len(config.BuildFarm) == 0 {
+		return true
+	}
```
hongkailiu, line 215:
```suggestion
```

I believe this should be `false`: no build farm means no cluster is disabled.

hongkailiu:
| DetermineE2EByJob | IgnoreE2EByJobAssignment | behaviour                                                                      |
|-------------------|--------------------------|--------------------------------------------------------------------------------|
| true              | *                        | job runs on the same cloud as the e2e test                                     |
| false             | true                     | rearrange if the current cloud is not the same as e2e; stay as it is otherwise |
| false             | false                    | rearrange                                                                      |

Is my understanding above correct?

> It was developed to enable gradual colocation (% of jobs in batch PR).

Is it the case when `DetermineE2EByJob=false` and `IgnoreE2EByJobAssignment=true`?
After several runs of the dispatcher, more and more jobs will run on the same cloud as the e2e. Correct?


I have some some feelings about the PR but they make no sense if the above is not correct.

jmguzik:
> DetermineE2EByJob	IgnoreE2EByJobAssignment	behaviour
> true	*	job runs on the same cloud as the e2e test
> false	true	rearrange if the current cloud is not the same as e2e; stay as it is otherwise
> false	false	rearrange
> Is my understanding above correct?
> 
> > It was developed to enable gradual colocation (% of jobs in batch PR).
> 
> Is it the case when `DetermineE2EByJob=false` and `IgnoreE2EByJobAssignment=true`? After several runs of the dispatcher, more and more jobs will run on the same cloud as the e2e. Correct?
> 
> I have some some feelings about the PR but they make no sense if the above is not correct.

Nice table :) If I understand correctly then mostly yes. It's like you described. I introduced `IgnoreE2EByJobAssignment=true` to change jobs gradually, I did this in 4 batches. Now with `DetermineE2EByJob=true` colocation is guarded, that is why if you failover you can't just assign an arbitrary cluster because it can be not according to the colocation rules, so you need `IgnoreE2EByJobAssignment=true` to allow mixed again state.

I have personally not tested  `IgnoreE2EByJobAssignment=true` AND `DetermineE2EByJob=true` (may have some side effect, Idk) so I would assume if one is true the latter should be false.

jmguzik:
> rearrange if the current cloud is not the same as e2e; stay as it is otherwise

**Do not** rearrange if the current cloud is not the same as e2e but also allow if it is ok with colocation rules == accept both states.

jmguzik:
| DetermineE2EByJob | IgnoreE2EByJobAssignment | behaviour 
| :---:   | :---: | :---: |
|true   | false | enforce colocation 
| false | true   | accept both outcomes for a job, colocated but also not colocated (moved during failover eg)   
| false | false   | rearrange all jobs (colocation disabled)

hongkailiu:
/lgtm

/hold

Free to cancel when you are ready.

We are going to merge this PR because it improves the procedure for failover.
We will implement it with another PR for another card as discussed via gmeeting:
- remove `IgnoreE2EByJobAssignment` because it is no longer needed.
- handle broken cluster by removing it from the configure somehow.

PR #2818, opened 2022-05-13
Failed the build upon a file-based catalog index

hongkailiu:
https://issues.redhat.com/browse/DPTP-2692

/cc @openshift/test-platform @gallettilance

PR #3169, opened 2022-11-28
Validate indexGeneratorStep

hongkailiu:
This PR not going to fix https://issues.redhat.com/browse/DPTP-2969. Instead it does what it should have been done in https://github.com/openshift/ci-tools/pull/2818 as the short term solution to https://issues.redhat.com/browse/DPTP-2692

- Reverted https://github.com/openshift/ci-tools/pull/2818/files because the build errored out earlier. 
- Based on [the finding](https://issues.redhat.com/browse/DPTP-2969?focusedCommentId=21311180&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21311180), the job fails at the validation stage of steps if ci-operator detects that index in the config is not database format.

We still need to figure out how to adopt the middle term solution with the help of the operator team.

/cc https://github.com/orgs/openshift/teams/test-platform @gallettilance

### DIFF
--- a/pkg/steps/index_generator.go
+++ b/pkg/steps/index_generator.go
@@ -114,7 +114,6 @@ func (s *indexGeneratorStep) indexGenDockerfile() (string, error) {
 	}
 	opmCommand = fmt.Sprintf("%s]", opmCommand)
 	dockerCommands = append(dockerCommands, opmCommand)
-	dockerCommands = append(dockerCommands, fmt.Sprintf("RUN (! grep -q 'operators.operatorframework.io.index.configs.v1=' %s) || (>&2 echo 'error: This is a file-based catalog index and opm index commands are not possible against this type of index. Please refer to the FBC docs here: https://olm.operatorframework.io/docs/reference/file-based-catalogs/'; exit 1)", IndexDockerfileName))
 	dockerCommands = append(dockerCommands, fmt.Sprintf("FROM %s:%s", api.PipelineImageStream, api.PipelineImageStreamTagReferenceSource))
 	dockerCommands = append(dockerCommands, fmt.Sprintf("WORKDIR %s", IndexDataDirectory))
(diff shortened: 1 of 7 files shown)


pkg/steps/index_generator.go
```diff
@@ -31,7 +35,38 @@ func (s *indexGeneratorStep) Inputs() (api.InputDefinition, error) {
...
-func (*indexGeneratorStep) Validate() error { return nil }
+func (s *indexGeneratorStep) Validate() error {
+	if s.config.BaseIndex != "" {
+		source := fmt.Sprintf("%s:%s", api.PipelineImageStream, s.config.BaseIndex)
+		ok, err := databaseIndex(s.client, source, s.jobSpec.Namespace())
+		if err != nil {
+			return fmt.Errorf("failed to determine if the image %s/%s is sqlite based index: %w", s.jobSpec.Namespace(), source, err)
+		}
+		if !ok {
+			return errors.New("opm index commands which is used by the ci-operator interacts only with database index but the base index is not. Please refer to the FBC docs here: https://olm.operatorframework.io/docs/reference/file-based-catalogs/")
```
smg247, line 46:
```suggestion
			return errors.New("opm index commands, which are used by the ci-operator, interact only with a database index, but the base index is not one. Please refer to the FBC docs here: https://olm.operatorframework.io/docs/reference/file-based-catalogs/")
```

hongkailiu:
/hold

Need to fix the broken e2e.

hongkailiu:
This wont work: if the `base_index` is built on the fly, i.e., specified via `images`, the image does not exist on the validation time.

https://docs.ci.openshift.org/docs/how-tos/testing-operator-sdk-operators/#building-operator-bundles

```
base_index: the base index to add the bundle to. If set, image must be specified in base_images or images. If unspecified, the bundle will be added to an empty index. Requires as to be set.
```

/close

PR #3170, opened 2022-11-28
`pod-scaler` producer: parse retention duration out of full result

smg247:
Starting around 11/02/22 the pod-scaler producer has been throwing errors like the following in all of the OSD managed clusters:
```
{"cluster":"build03","component":"pod-scaler","error":"could not determine Prometheus retention duration: not a valid duration string: \"11d or 90GiB\""}
```
This appears to be due to some recent change in the retention policy that OSD made. We have no better way to grab the duration piece of the policy, but it seems like if it is configured as a duration **or** a size then the duration comes first. This will work for the OSD clusters and the clusters we manage.

I will be monitoring the deployment to verify that this error goes away post-merge.

### DIFF
--- a/cmd/pod-scaler/producer.go
+++ b/cmd/pod-scaler/producer.go
@@ -190,7 +190,10 @@ func (q *querier) execute(ctx context.Context, c *clusterMetadata, until time.Ti
 	if err != nil {
 		return fmt.Errorf("could not query Prometheus runtime info: %w", err)
 	}
-	retention, err := model.ParseDuration(runtime.StorageRetention)
+	storageRetention := runtime.StorageRetention
+	// storageRetention may look like "11d or 90Gib" or "30d" depending on the configuration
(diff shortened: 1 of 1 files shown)


hongkailiu (approved):
Evidence:

```console
oc --context build03 --as system:admin exec -n openshift-monitoring prometheus-k8s-0 -- curl -s http://localhost:9090/api/v1/status/runtimeinfo | jq
{
  "status": "success",
  "data": {
    "startTime": "2022-11-16T08:19:32.703871498Z",
    "CWD": "/prometheus",
    "reloadConfigSuccess": true,
    "lastConfigTime": "2022-11-23T03:59:08Z",
    "corruptionCount": 0,
    "goroutineCount": 2790,
    "GOMAXPROCS": 8,
    "GOGC": "",
    "GODEBUG": "",
    "storageRetention": "11d or 90GiB"
  }
}
```

PR #3171, opened 2022-11-28
Failed the build upon a file-based catalog index (2)

hongkailiu:
Replacing https://github.com/openshift/ci-tools/pull/3169

This PR not going to fix https://issues.redhat.com/browse/DPTP-2969. Instead it does what it should have been done in https://github.com/openshift/ci-tools/pull/2818 as the short term solution to https://issues.redhat.com/browse/DPTP-2692

- Reverted https://github.com/openshift/ci-tools/pull/2818/files because the build errored out earlier. 
- Based on [the finding](https://issues.redhat.com/browse/DPTP-2969?focusedCommentId=21311180&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-21311180), the job fails at the runtime of steps to compose the build if ci-operator detects that index in the config is not database format.

We still need to figure out how to adopt the middle term solution with the help of the operator team.

/cc https://github.com/orgs/openshift/teams/test-platform @grokspawn

### DIFF
--- a/pkg/steps/index_generator_test.go
+++ b/pkg/steps/index_generator_test.go
@@ -59,7 +59,6 @@ func TestIndexGenDockerfile(t *testing.T) {
 COPY .dockerconfigjson .
 RUN mkdir $HOME/.docker && mv .dockerconfigjson $HOME/.docker/config.json
 RUN ["opm", "index", "add", "--mode", "semver", "--bundles", "some-reg/target-namespace/pipeline@ci-bundle0", "--out-dockerfile", "index.Dockerfile", "--generate"]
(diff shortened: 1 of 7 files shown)


hongkailiu:
I have seen what i was looking for in the e2e-oo test.

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3171/pull-ci-openshift-ci-tools-master-e2e-oo/1597400177134538752/artifacts/e2e-oo/e2e-oo/artifacts/TestOptionalOperators_optional_operators_named_bundle/ci-operator.log

```console
{"level":"debug","msg":"The base index image is sqlite based","time":"2022-11-29T01:43:10Z"}
```

pkg/steps/index_generator_test.go
```diff
@@ -117,3 +118,61 @@ COPY --from=builder /database/ database`,
 		})
 	}
 }
+
+func TestValidateIndexGeneratorStep(t *testing.T) {
```
smg247, line 122: I guess this should be called
```suggestion
func TestDatabaseIndex(t *testing.T) {
```

PR #3172, opened 2022-11-28
Add `rehearsals-ack` label on uninterested `openshift-bot` PRs

smg247:
The following jobs will need to have rehearsals skipped (and thus have the label applied) in order to add the `rehearsals-ack`  label to the merge criteria. This is needed so that the `periodic-prow-auto-bump` job can auto merge only if all rehearsals pass.

1. autoconfigbrancher
2. autoowners
3. prow-job-dispatcher

I considered adding this as a parameter to each of these tools, but it seems like something that is unlikely to change as long as we are using the plugin as we will never care about rehearsals for these.

For: https://issues.redhat.com/browse/DPTP-3218

### DIFF
--- a/cmd/autoconfigbrancher/main.go
+++ b/cmd/autoconfigbrancher/main.go
@@ -18,6 +18,7 @@ import (
 	"k8s.io/test-infra/prow/labels"
 
 	"github.com/openshift/ci-tools/pkg/promotion"
+	"github.com/openshift/ci-tools/pkg/rehearse"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


droslean:
@smg247 All auto* tools we have apply the labels programmatically. Can we just add this label there instead?

smg247:
> @smg247 All auto* tools we have apply the labels programmatically. Can we just add this label there instead?

Yeah, I think that would work just fine. Good idea!

droslean (approved):

PR #3173, opened 2022-11-29
Add include-job-names argument to filter jobs by names

xueqzhan:
[TRT-614](https://issues.redhat.com/browse/TRT-614)


dgoodwin:
strange, status never came back

/retest

PR #3176, opened 2022-11-30
ci-secret-generator: validation command

danilo-gemoli:
This PR tries to realize [DPTP-3032](https://issues.redhat.com/browse/DPTP-3032).
A new validation field has been introduced on `ci-secret-generator` configuration. 
It checks the output of the command that was issued before and then, if it succeeds, updates the secret.
/cc @hongkailiu

### DIFF
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -22,10 +22,12 @@ import (
 )
 
 const (
-	execCmdRunErrAction            = "run"
-	execCmdValidateStdoutErrAction = "validate stdout of"
-	execCmdValidateStderrErrAction = "validate stderr of"
-	execCmdErrFmt                  = "failed to %s command %q: %w\n%s:\n%s\n%s:\n%s"
+	execCmdRunErrAction                      = "run"
+	execCmdValidateStdoutErrAction           = "validate stdout of"
+	execCmdValidateStderrErrAction           = "validate stderr of"
+	execCmdErrFmt                            = "failed to %s command %q: %w\n%s:\n%s\n%s:\n%s"
+	execValidationCmdRunErrAction            = "run validation"
+	execValidationCmdValidateStderrErrAction = "validate stderr of validation"
 )
 
 var (
@@ -132,6 +134,33 @@ func (o *options) validateConfig() error {
 	return nil
 }
 
+func validateCommandOutput(cmdOutput []byte, validationCmd string) error {
+	if validationCmd == "" {
+		return nil
+	}
+	cmd := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", validationCmd)
+	var outBuf, errBuf bytes.Buffer
+	cmd.Stdin = bytes.NewReader(cmdOutput)
+	cmd.Stdout = &outBuf
+	cmd.Stderr = &errBuf
+	if err := cmd.Run(); err != nil {
+		stderr := errBuf.Bytes()
+		stdout := outBuf.Bytes()
+		_, partialStreams := err.(*exec.ExitError)
+		return fmtExecCmdErr(execValidationCmdRunErrAction, validationCmd, err, stdout, stderr, !partialStreams)
+	}
+
+	stderr := errBuf.Bytes()
+	stdout := outBuf.Bytes()
+
+	if len(stderr) != 0 {
+		return fmtExecCmdErr(execValidationCmdValidateStderrErrAction, validationCmd,
+			errExecCmdNotEmptyStderr, stdout, stderr, false)
+	}
+
+	return nil
+}
+
 func executeCommand(command string) ([]byte, error) {
 	cmd := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", command)
 	var outBuf, errBuf bytes.Buffer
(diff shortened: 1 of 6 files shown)


cmd/ci-secret-generator/main.go
```diff
@@ -132,6 +134,33 @@ func (o *options) validateConfig() error {
...
+		stderr := errBuf.Bytes()
+		stdout := outBuf.Bytes()
+		_, partialStreams := err.(*exec.ExitError)
+		return fmtExecCmdErr(execValidationCmdRunErrAction, validationCmd, err, stdout, stderr, !partialStreams)
+	}
+
+	stderr := errBuf.Bytes()
+	stdout := outBuf.Bytes()
+
+	if len(stderr) != 0 {
```
danilo-gemoli, line 156: @hongkailiu I'm not sure about this. Should we check the `stderr` other than the exit code?

bbguimaraes: Generation and validation are fundamentally different, so I don't see why we should treat them the same.

This should just look at the exit code and print the combined (`stdout`+`stderr`) output on failure — but see next comments.

cmd/ci-secret-generator/main.go
```diff
@@ -195,6 +224,12 @@ func updateSecrets(config secretgenerator.Config, client secrets.Client) error {
 				errs = append(errs, errors.New(msg))
 				continue
 			}
+			if err := validateCommandOutput(out, field.Validation); err != nil {
```
bbguimaraes, line 227: Do we have examples of validation commands we intend to use and the type of output they generate?  Could we leak secret content via the validation command.  I don't see us doing any sort of censoring here.

bbguimaraes: Would be great if we could test this.

cmd/ci-secret-generator/main.go
```diff
@@ -195,6 +224,12 @@ func updateSecrets(config secretgenerator.Config, client secrets.Client) error {
...
+				msg := fmt.Sprintf("failed to validate command output:\n%s", string(out))
```
bbguimaraes, line 228: We definitely _do not_ want to print `out` here.

bbguimaraes (commented):
This needs documentation.

/hold

cmd/ci-secret-generator/main.go
danilo-gemoli: If secret leaking is the concert you're definitely right. Shouldn't the same  be applied [here](https://github.com/openshift/ci-tools/blob/master/cmd/ci-secret-generator/main.go#L194) by the way? `err` could (even though I'm not 100% sure) leak something and no censoring has been applied as well.

cmd/ci-secret-generator/main.go
danilo-gemoli: It would be great if we could simplify it into something like:
```golang
func validateCommandOutput(cmdOutput []byte, validationCmd string) error {
	if validationCmd == "" {
		return nil
	}
	cmd := exec.Command("bash", "-o", "errexit", "-o", "nounset", "-o", "pipefail", "-c", validationCmd)
	cmd.Stdin = bytes.NewReader(cmdOutput)
	if out, err := cmd.CombinedOutput(); err != nil {
		return fmt.Errorf("failed to run validation command %q: %w\ncombined output:\n%s",
			validationCmd, err, out)
	}
	return nil
}
```

cmd/ci-secret-generator/main.go
hongkailiu: We "committed a suicide" once in the past, generating the `kubeconfig` because there were deprecated error msg in the `stderr`. So we screwed up the secret which led to outage.

The card is to validate the output before updating the secrets, suggested by Bruno.

We use [this](https://github.com/openshift/release/blob/c947321a3b8e2a399ad6cc6fdc950dfea83867ce/core-services/ci-secret-generator/_config.yaml#L2) cmd to generate the token,
ideally, i would like to validate if the generated token is valid, e.g., by `oc login` with the token and `oc whoami`

Another mistake in the past: We need to verify the kubeconfig file has only one context, named as the cluster. I would like to verify that too.

We used to verify this by https://github.com/openshift/ci-tools/blob/master/images/ci-secret-generator/oc_sa_create_kubeconfig.sh

I can definitely feel some difficulty with the current implementation:
We need api-server URL to do oc-login and we have list of clusters.
How do we map the relations?
It is even more challenging to do it as "shell in yaml".

cmd/ci-secret-generator/main.go
hongkailiu: checking on `stdeer` is the simplest validation before implementing the real validation: validation fails if `stderr` is not empty.

I agree that we do not need to check `stderr` if we validate combination of `stdout` and `stderr`.

bbguimaraes: We treat the error output of the generation command specially because of the problem we had where deprecation warnings were missed, leading us to find problems only after the deprecation had happened.

Is that likely to happen in validation commands?  Is that worth the additional complexity of this code and of every validation command which will have to carefully manage its `stderr`?  (I don't know the answer, since I don't know what types of programs will be used for validation and their output)

cmd/ci-secret-generator/main.go
hongkailiu: Your concern of exposing secrets is very valid to me. (It is listed even in the notes from one of the incidents.)
I did not fix the code because I was thinking about the validation and might fix all those together. Moreover, after we fixed the config, the error does not show any more. (I know it is an excuse and does not prevent exposing from happening again. But it is the current reality).

So we have 2 issues now to me:
1. How we can censor the token? The risky is already in our code without the validation. 
    It is hard to achieve because we do not know the token in advance. We need to basically remove all the output which is not friendly at all to the debugging. (Another excuse that I did not fix it)
2. Is validation technically feasible? A typical case is the token: how should we validate it? 
    It is more functional to the card. The validation wont work easily unless we need to extend how the tool works, e.g., we need to provide api server in the validation command. I do not have the solution.

PR #3179, opened 2022-12-02
`pj-rehearse`: remove defunct code for old runner

smg247:
There are a few things that we no longer need in `pj-rehearse` now that the old runner jobs are no longer around:

- the dual job and debug loggers - now it can just be a single logger
- rehearsalLimit
- the job mode is now simplified to a basic mode for integration testing, this also means we can simplify the error messages
- the temporary CM code references `SHA` now instead of `buildId`
- we don't need to be able to create a `RehearsalCandidate` from a `JobSpec` anymore

For: https://issues.redhat.com/browse/DPTP-3208

/cc @openshift/test-platform @droslean 
/ hold so I can watch the rollout

### DIFF
--- a/cmd/pj-rehearse/main.go
+++ b/cmd/pj-rehearse/main.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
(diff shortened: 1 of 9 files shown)


cmd/pj-rehearse/main.go
```diff
@@ -101,19 +90,35 @@ func (o *options) validate() error {
...
-			errs = append(errs, errors.New("--candidate-path was not provided"))
-		}
+	if o.dryRun {
+		errs = append(errs, o.dryRunOptions.validate())
 	}
 
 	return utilerrors.NewAggregate(errs)
 }
 
+type dryRunOptions struct {
```
droslean, line 103: can we move those on `pkg/rehearse/options.go?`

smg247: Do you mean that I should create a new file to move these options to? What do we get out of that move?

smg247: I could just move these into the `options` struct directly, but I thought having a separate struct made it more clear that they were only for the test.

PR #3181, opened 2022-12-05
Use P95 for more meaningful comparison in disruption allowances update PRs

DennisPeriquet:
TRT-705

Instead of P99, we will show previous P95, current P95, time increase, and percent increase.


DennisPeriquet:
we want to keep more fields.

PR #3183, opened 2022-12-05
Show Use P95/P99 and % increase in disruption allowances updates

DennisPeriquet:
TRT-705

The changes in P95 given a better signal to tell if something regressed.  We still show P99 for reference.


pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -142,31 +142,42 @@ func (o *JobRunHistoricalDataAnalyzerOptions) compareAndUpdate(newData, currentD
 		}
 
 		newP99 := getDurationFromString(new.GetP99())
-		newP95 := getDurationFromString(new.GetP99())
+		newP95 := getDurationFromString(new.GetP95())
```
DennisPeriquet, line 145: This was a bug from before -- it's fixed here.

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -142,31 +142,42 @@ func (o *JobRunHistoricalDataAnalyzerOptions) compareAndUpdate(newData, currentD
...
-				d.PercentTimeDiff = percentDiff
+			if newP99 > oldP99 && percentDiffP99 > o.leeway {
+				d.TimeDiffP99 = timeDiffP99
+				d.PercentTimeDiffP99 = percentDiffP99
 				d.PrevP99 = oldP99
 			}
-			if newP99 < oldP99 {
-				decreaseCount += 1
+			if newP95 < oldP95 {
+				decreaseCountP95 += 1
```
dgoodwin, line 180: P99 is what we're actually enforcing, addition of the P95 is just to help us see if there's something that needs manual investigation. 

The decrease count probably should stay with P99, as that seems like it might be what we'd expect when viewing the file.

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -142,31 +142,42 @@ func (o *JobRunHistoricalDataAnalyzerOptions) compareAndUpdate(newData, currentD
...
+				increaseCountP95 += 1
+				d.TimeDiffP95 = timeDiffP95
+				d.PercentTimeDiffP95 = percentDiffP95
+				d.PrevP95 = oldP95
 			}
-			if newP99 > oldP99 && percentDiff > o.leeway {
-				increaseCount += 1
-				d.TimeDiff = timeDiff
-				d.PercentTimeDiff = percentDiff
+			if newP99 > oldP99 && percentDiffP99 > o.leeway {
```
dgoodwin, line 174: Lets keep tracking the increase count on P99 as well, not P95.

dgoodwin (commented):
Looks great otherwise, just a small change to keep using the P99 for the increase/decrease counts as that is what we enforce on in the file.

PR #3187, opened 2022-12-07
`pj-rehearse`: fix filter presubmit logic to not skip jobs

smg247:
The filtering function has a bug where it will entirely skip looking at certain jobs. This can manifest in jobs not ending up in the final result that should have. We need to use some custom looping logic to account for removing elements from the slice in-place. The unit test was not complex enough to catch this error, so I made it more robust.

/cc @openshift/test-platform


hongkailiu (approved):

smg247:
There is an incident with the imagestreams, but that doesn't need to block this PR

PR #3189, opened 2022-12-07
pkg/webreg: render empty chains correctly

bbguimaraes:
These changes fix the rendering of pages such as:

- https://steps.ci.openshift.org/chain/agent-test
    - `Internal Server Error: template: baseTemplate:12:3: executing "baseTemplate" at : error calling chainGraph: runtime error: index out of range [-1]`
- https://steps.ci.openshift.org/workflow/agent-e2e-compact-ipv4
    - `Internal Server Error: template: baseTemplate:22:3: executing "baseTemplate" at : error calling workflowGraph: runtime error: index out of range [-1]`

The underlying problem is that `agent-test` is currently empty:

```yaml
chain:
  as: agent-test
  steps:
#  - ref: baremetalds-e2e-test
  documentation: |-
    This chain encapsulates the e2e tests for the agent installer
```

https://github.com/openshift/release/blob/2e20288732dd6ab00f164f5efcd7cc03968a2511/ci-operator/step-registry/agent/test/agent-test-chain.yaml

Regardless of whether that is a valid and/or useful configuration, the interface should not panic when trying to render it.

The problem is fixed by this seris of commits… with a complete rewrite.  Apologies in advance for inflicting such a review; while I did at some point fix the problem in the original code, it became obvious that there was a better way of doing things.  In order to give some confidence that the rewrite is valid, commits are structured in the following sequence:

- A series of commits improving existing tests and adding several more cases, with no practical implementation changes.
- One commit rewriting the graph creation functions.
    - Rendering functions need only minor adjustments.
    - Existing tests continue to pass (the only required changes are regarding the order in which edges are declared, which is irrelevant for `graphviz`).
    - Tests for the empty cases can now be added, and pass immediately.
- One commit simplifying the rendering functions using `html/template`.
    - While I'm rewriting things…

The end result is less code, more uniformity with no flags or odd cases, (more) documentation, and more tests:

```
$ git diff --shortstat master -- pkg/webreg/graphviz.go
 1 file changed, 218 insertions(+), 196 deletions(-)
$ git diff master -- pkg/webreg/graphviz.go | grep -Pc '^\+\s*//'
51
$ git diff --shortstat master -- pkg/webreg/{graphviz_test.go,testdata}
 12 files changed, 531 insertions(+), 100 deletions(-)
```


bbguimaraes:
https://issues.redhat.com/browse/DPTP-3246

pkg/webreg/graphviz.go
```diff
@@ -4,181 +4,249 @@ import (
...
 )
 
-func addSubgraph(mainGraph graph, name string, root []api.TestStep, chains registry.ChainByName, isChild, linkable, isRoot bool) (graph, int) {
-	sg := subgraph{
-		label:       name,
-		firstNode:   -1,
-		hasSGParent: isChild,
-		linkable:    linkable,
+const (
+	graphvizTemplateStr = `digraph Webreg {
```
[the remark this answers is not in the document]
bbguimaraes: Done.

pkg/webreg/testdata/zz_fixture_TestWorkflowDotFile_step_to_chain.yaml
```diff
@@ -0,0 +1,46 @@
+digraph Webreg {
```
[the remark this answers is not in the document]
bbguimaraes: You're right.  Fixed with:

```diff
--- pkg/webreg/graphviz.go
+++ pkg/webreg/graphviz.go
@@ -184,7 +184,7 @@ func (b *graphBuilder) addEdge(i int) {
 		if b.edge.srcType != noEdge {
 			b.graph.edges = append(b.graph.edges, b.edge)
 		}
-		b.edge = edge{srcType: nodeType, srcNode: b.edge.dstNode}
+		b.edge = edge{srcType: nodeType, srcNode: b.edge.dstNode, dstType: nodeType}
 }
 
 func ChainGraph(name string, chains registry.ChainByName) ([]byte, error) {
```

Note that the files were still generated correctly (which is why I didn't notice this problem), since we only differentiate between `subgraphType` and "anything else".

pkg/webreg/graphviz.go
```diff
@@ -2,183 +2,224 @@ package webreg
...
-	}
-	mainGraph.edges = append(mainGraph.edges, preTestEdge)
-	testPostEdge := edge{
-		src:     testIndex,
-		dst:     postIndex,
-		srcType: subgraphType,
-		dstType: subgraphType}
-	mainGraph.edges = append(mainGraph.edges, testPostEdge)
-	return writeDotFile(mainGraph)
+	b.edge = edge{srcType: nodeType, srcNode: b.edge.dstNode, dstType: nodeType}
```
danilo-gemoli, line 187: Assigning `srcNode: b.edge.dstNode` is a little bit misleading, better to change it into `srcNode: i` to underline that `i`, the node has just been added, is going to become the source for the next edge to come.

danilo-gemoli, line 187: `dstType: nodeType` works but looks like a speculation. How do you know beforehand the next edge head will be of type `nodeType`? (Yeah I know, if it won't then `addSubgraph` will set `b.edge.dstType = subgraphType` so we are fine in any case).

pkg/webreg/graphviz.go
```diff
@@ -2,183 +2,224 @@ package webreg
...
-	// this is used to identify a node that can be linked from is this sg is a src for an edge
-	if lastType == nodeType {
-		sg.lastNode = lastIndex
-	} else {
-		sg.lastNode = mainGraph.subgraphs[lastIndex].lastNode
+	i := len(b.graph.subgraphs)
+	b.graph.subgraphs = append(b.graph.subgraphs, sg)
+	b.edge.srcType = subgraphType
+	b.edge.srcGraph = i
+	if hasSrc && edge != len(b.graph.edges) {
```
danilo-gemoli, line 165: Please add a comment here explaining what's the reason of being of this `if`

pkg/webreg/graphviz.go
```diff
@@ -2,183 +2,224 @@ package webreg
...
+	// - In both of the previous cases, `dstGraph`/`srcGraph` is set
+	//   accordingly.  It contains the index of the chain's sub-graph.  Note
+	//   that this is done by `addSubgraph` at every level and always at the
+	//   end of the function, so that the edge is clipped against the outermost
+	//   chain.
+	// - Whenever a node (i.e. a step) is added to the graph, an edge is added
+	//   based on the current state of the variable.  Its destination is the
+	//   newly added node.  This variable is then changed in preparation for
+	//   the subsequent edge (its source is the one just added).
+	edge edge
```
danilo-gemoli, line 84: The description is outstanding. This is just my personal taste and nothing else: I've noticed `addSubgraph` is recursive and it manipulates `edge` across its calls, it's fine even though I find it easier to follow recursive algorithms that have the state as a parameter rather than a member of a class.
No need to change anything by the way.

danilo-gemoli (changes requested):

pkg/webreg/graphviz.go
bbguimaraes: It's supposed to reflect this comment (or vice versa), which is how I read it: https://github.com/openshift/ci-tools/pull/3189/files#diff-de4c46ee468a303e27ac25c79450556c326e7865b3dfaa6377220b41ee0852dfR80-R83.

```
Whenever a node (i.e. a step) is added to the graph, an edge is added
based on the current state of the variable.  Its destination is the
newly added node.  This variable is then changed in preparation for
the subsequent edge (its source is the one just added).
```

bbguimaraes: This and the other statement together implement the rule: the next edge will have a node as its destination unless we cross a sub-graph boundary.

pkg/webreg/graphviz.go
bbguimaraes: You caught me, this is the most subtle section of the code.  Comments added.

pkg/webreg/graphviz.go
bbguimaraes: > E.g. the edge between `ipi-install-install` and the `test` sub-graph, from the same `step-to-chain` as before:

To illustrate:

```
addSubgraph - pre
    addNode - ipi-deprovision-deprovision
    addSubgraph - ipi-install
        addNode - ipi-install-rbac
        addNode - ipi-install-install
            [edge is set]
        [edge is modified (to make `ipi-install` its source)]
    [edge is modified (to make `pre` its source)]
addSubgraph - test
    addNode - empty
        [the actual edge in the graph is created]
    [edge is modified (to make `test` the destination, see https://github.com/openshift/ci-tools/pull/3189#discussion_r1210419087)]
addSubgraph - post
    addNode - empty
```

danilo-gemoli:
/retest
/lgtm
Thanks! We have now in this PR a deep explanation of the algorithm.

PR #3193, opened 2022-12-09
Skip report entry for no difference in P99 vs. P95

DennisPeriquet:
TRT-702

[This observation](https://github.com/openshift/origin/pull/27601#pullrequestreview-1211575865), noted a place where there was a jump by over 30% in P99 data but the entry was not included in the table.  This change fixes that.

Analysis: the code before this change, does not add the entry into the table when the **P95** value didn't change.  What we actually want is to not add the entry into the table when the **P99** value didn't change since we are enforcing P99 and only using P95 to tell if something is really wrong.


dgoodwin:
Something looks hard broken, I suspect in master. Either something changed or some bad code slipped in, but I can't really make sense of it. I suggest a workflow ticket with dptp maybe? Not sure who exactly to talk to.

PR #3199, opened 2022-12-14
Add possibility of skipping periodic tests to config-brancher & fix e2e

jmguzik:
That will cut around 2% of current spendings once enabled.

### DIFF
--- a/cmd/config-brancher/main.go
+++ b/cmd/config-brancher/main.go
@@ -18,7 +18,8 @@ import (
 type options struct {
 	promotion.FutureOptions
 
-	BumpRelease string
+	BumpRelease   string
+	skipPeriodics bool
 }
 
 func (o *options) Validate() error {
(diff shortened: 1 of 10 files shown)


pkg/api/types.go
```diff
@@ -703,6 +703,9 @@ type TestStepConfiguration struct {
 	// Optional indicates that the job's status context, that is generated from the corresponding test, should not be required for merge.
 	Optional bool `json:"optional,omitempty"`
 
+	// Portable allows to port periodic tests to current and future release despite the demand to skip periodics
+	Portable bool `json:"portable,omitempty"`
```
bbguimaraes, line 707: Would it make sense to not copy other job types to those release branches?  If not, we need [validation](https://github.com/openshift/ci-tools/blob/master/pkg/validation/test.go) so it is not added to non-periodic tests.

cmd/config-brancher/main.go
```diff
@@ -139,13 +141,25 @@ func generateBranchedConfigs(currentRelease, bumpRelease string, futureReleases
...
 
+// removePeriodics removes periodic tests from the configuration
+func removePeriodics(tests *[]api.TestStepConfiguration) {
+	for i := len(*tests) - 1; i >= 0; i-- {
+		if !(*tests)[i].Portable && ((*tests)[i].Cron != nil || (*tests)[i].Interval != nil) {
+			*tests = append((*tests)[:i], (*tests)[i+1:]...)
+		}
+	}
+}
+
```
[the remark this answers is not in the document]
jmguzik: We can probably start the contest who makes this more unreadable =D

pkg/api/types.go
jmguzik: First question: I would like to announce this change and give people the right to keep portability. On non-periodic tests this flag, even enabled makes no effect but it could be added to validation for correctness.
Second question: I will describe it as a part of a branching document.

PR #3202, opened 2022-12-14
Fix ci-tools failing test on missing 4.5 release

stbenjam:
4.5 is EOL, use 4.11 instead for the ci-stream test case.


stbenjam:
/close

It's more than a couple of references.  https://issues.redhat.com/browse/DPTP-3261 tracks it

PR #3208, opened 2022-12-19
Skip periodics for the future and current releases

jmguzik:


jmguzik:
/override ci/prow/e2e 
not needed for flag addition

PR #3213, opened 2022-12-21
Change the way clusters are disabled during failover

jmguzik:
- revert https://github.com/openshift/ci-tools/commit/ef2a28d87bc879fdc3f1bdd27a890c6da4be3544
-  remove `IgnoreE2EByJobAssignment`
- change the way clusters are added and deleted 
- add get HTTP to know provider

/cc @hongkailiu

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -2,9 +2,12 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io/fs"
+	"io/ioutil"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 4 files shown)


jmguzik:
@hongkailiu this is still WIP but I would welcome your entry comments.

cmd/prow-job-dispatcher/main_test.go
```diff
@@ -363,3 +363,94 @@ func equalError(t *testing.T, expected, actual error) {
...
+			config:  &dispatcher.Config{},
+			enabled: sets.NewString("build04"),
+		},
+		{
+			name:    "Add multiple clusters, different providers",
+			config:  &dispatcher.Config{},
+			enabled: sets.NewString("build03", "build04", "build05"),
+		},
+	}
+	for _, tt := range tests {
```
hongkailiu, line 390: What does `tt` stand for?

cmd/prow-job-dispatcher/main_test.go
```diff
@@ -363,3 +363,94 @@ func equalError(t *testing.T, expected, actual error) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			addEnabledClusters(tt.config, tt.enabled, getProvider)
+			for cluster := range tt.enabled {
```
hongkailiu, line 393: I would define an `expect` in `tt` and then compare `tt.expect` and `tt.config` after the function.

cmd/prow-job-dispatcher/main.go
```diff
@@ -375,13 +367,81 @@ func dispatchJobs(ctx context.Context, prowJobConfigDir string, maxConcurrency i
...
+		return "", err
+	}
+	var page pageData
+	if err := json.Unmarshal(body, &page); err != nil {
+		return "", err
+	}
+	for _, data := range page.Data {
+		if c, exists := data["cluster"]; exists && c == cluster {
+			if provider, exists := data["cloud"]; exists {
+				return api.Cloud(strings.ToLower(provider)), nil
```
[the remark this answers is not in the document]
jmguzik: Done

cmd/prow-job-dispatcher/main_test.go
jmguzik: table test but changed to tc, done

cmd/prow-job-dispatcher/main_test.go
jmguzik: I left it as I implemented it; I just need check some things in the conf after the function and I feel this way I have a more flexible approach and do not need to recreate some parts of the structure. Tests are not very complicated so I think this trade-off here is OK. Let me know if you really want it to happen.

PR #3220, opened 2023-01-04
cluster-display: update error handling for cluster list

psalajova:
Updated error handling so that if there's an error regarding one cluster, the error is logged, and the rest of the clusters are _not_ skipped.

for https://issues.redhat.com/browse/DPTP-3086

/cc smg247

### DIFF
--- a/cmd/cluster-display/main.go
+++ b/cmd/cluster-display/main.go
@@ -118,27 +118,13 @@ func getClusterPoolPage(ctx context.Context, hiveClient ctrlruntimeclient.Client
 	return &page, nil
 }
 
-func getClusterPage(ctx context.Context, clients map[string]ctrlruntimeclient.Client, skipHive bool) (*Page, error) {
-	var data []map[string]string
-	for cluster, client := range clients {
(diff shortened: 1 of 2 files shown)


cmd/cluster-display/main.go
```diff
@@ -118,47 +118,66 @@ func getClusterPoolPage(ctx context.Context, hiveClient ctrlruntimeclient.Client
 	return &page, nil
 }
 
-func getClusterPage(ctx context.Context, clients map[string]ctrlruntimeclient.Client, skipHive bool) (*Page, error) {
+type ClusterInfoGetter interface {
+	GetClusterDetails(ctx context.Context, cluster string, client ctrlruntimeclient.Client) (map[string]string, error)
+}
+
+type clusterInfoGetter struct{}
```
smg247, line 125: I think since you are going to have this struct, you should add a `cluster` field to it. That way you don't also have to pass the cluster to the method. You would then just create one of these for each cluster in `getClusterPage`

smg247 (changes requested):

PR #3221, opened 2023-01-04
Standardize check for periodic

smg247:
When the `minimum_interval` field was added in https://github.com/openshift/ci-tools/pull/2980, only some of the checks were updated to include it. An important missing place is in `pj-rehearse`, where we will, consequently, not pick up any jobs that are configured with the field as being affected. In order to solve this problem permanently I have standardized the check for a test being a periodic, and replaced all usages with the new method.

/cc @openshift/test-platform 

For: https://issues.redhat.com/browse/DPTP-3264


bbguimaraes:
Nice.

/lgtm

PR #3225, opened 2023-01-04
ci-secret-bootstrap: log some stats of secrets

hongkailiu:
The [job history](https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/job-history/gs/origin-ci-private/logs/periodic-ci-secret-bootstrap) shows that it takes 35m to execute the job which seems quite long.
IIRC, it took about 2m when we just switch from BitWarden to Vault.
Most of the time is spent on updating the secrets on the cluster.

We want to show some stats info to understand the work.

### DIFF
--- a/cmd/ci-secret-bootstrap/secretstats.go
+++ b/cmd/ci-secret-bootstrap/secretstats.go
@@ -1,0 +1,46 @@
+package main
+
+import (
+	"github.com/montanaflynn/stats"
+	"github.com/sirupsen/logrus"
+
+	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+type secretStats struct {
+	// count is the total number of the secrets
+	count int
+	// median is the median of the size of the data of the secrets
+	// the size is counted by the sum of length of each key and its value in the secret's data
+	median float64
+}
+
+func generateSecretStats(secretsByClusterAndName map[string]map[types.NamespacedName]coreapi.Secret) secretStats {
+	if len(secretsByClusterAndName) == 0 {
+		return secretStats{}
+	}
+	count := 0
+	var size []float64
+	for _, secretMap := range secretsByClusterAndName {
+		count += len(secretMap)
+		for _, secret := range secretMap {
+			size = append(size, float64(getSize(secret)))
+		}
+	}
+	median, err := stats.Median(size)
+	if err != nil {
+		logrus.WithError(err).Error("Failed to calculate the median")
+		return secretStats{}
+	}
+	result := secretStats{count: count, median: median}
+	return result
+}
+
+func getSize(secret coreapi.Secret) int {
+	size := 0
+	for k, v := range secret.Data {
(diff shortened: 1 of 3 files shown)


cmd/ci-secret-bootstrap/secretstats.go
```diff
@@ -0,0 +1,43 @@
...
+
+func generateSecretStats(secretsByClusterAndName map[string]map[types.NamespacedName]coreapi.Secret) (secretStats, error) {
+	if len(secretsByClusterAndName) == 0 {
+		return secretStats{}, nil
+	}
+	count := 0
+	var size []float64
+	for _, secretMap := range secretsByClusterAndName {
+		for _, secret := range secretMap {
+			count = count + 1
```
bbguimaraes, line 25: `count += len(secretMap)` (outside the loop).

bbguimaraes: The names (here and in the resulting struct) could also be more explicit about what it is that is being counted and measured.

cmd/ci-secret-bootstrap/secretstats_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/openshift/ci-tools/pkg/testhelper"
+)
+
+func TestGenerateSecretStats(t *testing.T) {
+	testCases := []struct {
+		name                    string
+		secretsByClusterAndName map[string]map[types.NamespacedName]coreapi.Secret
+		expected                secretStats
+		expectedErr             error
```
bbguimaraes, line 19: There are no tests which set this, but do we even want to return an error?

The only way a median calculation can "fail" is if there are no values (https://github.com/montanaflynn/stats/blob/master/median.go).  We return `0` in that case (which is fine for logs), so there is never a failure case.

hongkailiu: Updated.

cmd/ci-secret-bootstrap/secretstats.go
hongkailiu: Added GoLang docs for the fields.

hongkailiu:
Taken from https://deck-internal-ci.apps.ci.l2s4.p1.openshiftapps.com/view/gs/origin-ci-private/logs/periodic-ci-secret-bootstrap/1611087940132278272

```
{"component":"ci-secret-bootstrap","count":787,"file":"/go/src/github.com/openshift/ci-tools/cmd/ci-secret-bootstrap/main.go:401","func":"main.constructSecrets","level":"info","median":4266,"msg":"Secret stats before fetching user secrets","severity":"info","time":"2023-01-05T19:51:25Z"}

...

{"component":"ci-secret-bootstrap","count":3487,"file":"/go/src/github.com/openshift/ci-tools/cmd/ci-secret-bootstrap/main.go:407","func":"main.constructSecrets","level":"info","median":628,"msg":"Secret stats after fetching user secrets","severity":"info","time":"2023-01-05T19:51:39Z"}

```

PR #3227, opened 2023-01-05
Before branching day: integration tests

danilo-gemoli:
These integration tests cover the tools developed in:
- [x] #2815
- [x] #2866
- [x] #3212 

Run them locally with:
```sh
$ echo -n {rpm-deps-mirroring-services,release-controller-config-manager,generated-release-gating-jobs} \
  | xargs -d' ' -I{} /bin/bash -c 'cd "${CI_TOOLS_REPO}/cmd/branchingconfigmanagers/{}" && go install'
$ make integration SUITE=br-cuts-*
```

/cc @jmguzik @droslean

### DIFF
--- a/test/integration/branchcuts/generated-release-gating-jobs/test1/data/release/ci-operator/config/openshift/release/openshift-release-master__ci-4.10.yaml
+++ b/test/integration/branchcuts/generated-release-gating-jobs/test1/data/release/ci-operator/config/openshift/release/openshift-release-master__ci-4.10.yaml
@@ -1,0 +1,37 @@
+releases:
+  initial:
+    candidate:
+      product: ocp
+      relative: 1
+      stream: ci
+      version: "4.10"
+  latest:
+    candidate:
+      product: ocp
+      stream: ci
+      version: "4.10"
+resources:
+  '*':
+    requests:
+      cpu: 100m
+      memory: 200Mi
+tests:
+- as: e2e-aws
+  interval: 24h
+  steps:
+    cluster_profile: aws-2
+    env:
+      BASE_DOMAIN: aws-2.ci.openshift.org
+    workflow: openshift-e2e-aws
+- as: e2e-aws-techpreview
+  interval: 48h
+  steps:
+    cluster_profile: aws-2
+    env:
+      BASE_DOMAIN: aws-2.ci.openshift.org
+    workflow: openshift-e2e-aws-techpreview
+zz_generated_metadata:
+  branch: master
+  org: openshift
+  repo: release
+  variant: ci-4.10
--- a/hack/lib/branchcuts/test-common.sh
+++ b/hack/lib/branchcuts/test-common.sh
@@ -1,0 +1,56 @@
+readonly src_test_files_relative_dir="data"
+readonly want_test_files_relative_dir="want"
+readonly test_config_relative_path="test_config"
+
+function branchcuts::read_test_config() {
+    local -n _cfg=$1
+    local _test_config_path="${2}"
+    for key_value in $(cat "${_test_config_path}")
+    do
+        if [[ "${key_value}" =~ (.+?)=(.*) ]]
+        then
+            _cfg["${BASH_REMATCH[1]}"]="${BASH_REMATCH[2]}"
+        fi
+    done
+}
+readonly -f branchcuts::read_test_config
+
+function branchcuts::run_tests_template() {
+    local run_cfg_manager_fn="${1}"
+    local _diffs=""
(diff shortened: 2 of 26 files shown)


jmguzik:
LGTM but I am not an expert in the field of integration tests in our repo. I am also not sure if we really need them for such tiny tools.
Anyway, @droslean I give time until tomorrow to complain about this commit, otherwise, I'll LGTM :P

droslean:
@danilo-gemoli Can you elaborate more on why we need all of those `.sh` scripts? The integration tests should just run the binary and we need at least one small script to compare the output.

danilo-gemoli:
Sure. I've added three integration tests, along with their `.sh` scripts:
- `br-cuts-generated-release-jobs.sh`
- `br-cuts-release-controller-configs.sh`
- `br-cuts-rpm-mirroring-services.sh`

Each of which setups the whole test, declares a `cleanup` function, some variables and finally 
registers another function (ex.: `function branchcuts::rpm_mirroring::run_config_manager()`) to run the proper tool.

Each test has its tests cases located in `test/integration/branchcuts/$TOOL_UNDER_TEST/$NTH_TEST_CASE`, where:
- `data` contains the files that the tool is expected to bump
- `want` contains the expected results
- `test_config` is a simple `$KEY=$VALUE` text file that holds the configuration the tool needs to be run with (ex.: `current_release=4.12`, `future_release=4.13`)

The script `hack/lib/branchcuts/test-common.sh` holds the code common to all the tests, `branchcuts::run_tests_template()` being the core function.
That template function loops through every test case, runs a tool, compares the resulting output with
the expected one and finally reports either success or failure.

What follows is an example of execution.
Integration test `br-cuts-generated-release-jobs` is going to be executed:
- `br-cuts-generated-release-jobs.sh` runs and then declares:
  - `$test_cases_dir`: a variable defining where to find the test cases
  - `branchcuts::generated_release_jobs::run_config_manager` a function that knows how to run the tool `generated-release-gating-jobs`
- `branchcuts::run_tests_template` (in `test-common.sh`) is called and therefore, for every test case, the following happens:
  - `test_config` is read and parsed
  - some temporary folders are created; those being like the ones `generated-release-gating-jobs` is expected to read the files from. In this case: `ci-operator/config/openshift/release`
  - the function `branchcuts::generated_release_jobs::run_config_manager` declared beforehand is finally called, so the tool executes
  - the resulting output is going to be compared to the expected result. Let's suppose test case `test1` being run, the expected output is found in `test/integration/branchcuts/generated-release-gating-jobs/test1/want/openshift-release-master__ci-4.11.yaml`
  - failure or success is finally reported

Actual and expected data are already there, nothing is generated on the fly.

droslean:
@danilo-gemoli The only temporary files we should generate are the actual data. The test should have input and expected data, and a script to compare the expected with the actual data.

danilo-gemoli:
Folders structure (like `ci-operator/config/openshift/release`) only is automatically generated at every run.
Is that what you're referring to? Should it exists in advance?

danilo-gemoli:
@droslean I have changed it a little bit in accordance with your advice. Not even the folders structured is generated on the fly anymore. Thanks!

test/integration/branchcuts/generated-release-gating-jobs/test1/data/release/ci-operator/config/openshift/release/openshift-release-master__ci-4.10.yaml
```diff
@@ -0,0 +1,37 @@
+releases:
```
droslean, line 1: There is no need to separate this file into a folder named `test1`.  We need to have all the input files in one place like we have them in the release repo. And to extend different cases you need to add more configs that include those cases. The tool should run only once and generate the desired output.

danilo-gemoli: These test cases are much like the ones you can find into the `cluster-init` integration tests here: [test/integration/cluster-init](https://github.com/openshift/ci-tools/tree/master/test/integration/cluster-init). I can adhere to the same naming conventions `input` and `expected` instead of `data` and `want` and renaming `test1`, `test2`, ... `testN` to something more meaningful but, other than that, both have test cases split into several folders, each of which recreates part of the `openshift/release` repository folders structure.
If you take a look at [test/integration/cluster-init.sh](https://github.com/openshift/ci-tools/blob/master/test/integration/cluster-init.sh) you'll notice that `cluster-init` tool runs more than once.

PR #3228, opened 2023-01-05
dptp-pool-cm: add a new controller hypershift_namespace_reconciler

hongkailiu:


pkg/controller/hypershift_namespace_reconciler/hypershift_namespace_reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *logrus.Entry) error {
+
+	*log = *log.WithField("name", req.Name)
+	log.Info("Starting reconciliation")
+
+	ns := &corev1.Namespace{}
+	if err := r.client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: req.Name}, ns); err != nil {
+		if apierrors.IsNotFound(err) {
```
droslean, line 86: You can use https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate#Funcs to control for which event the controller will be triggered.

pkg/controller/hypershift_namespace_reconciler/hypershift_namespace_reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+		log.WithError(err).Error("Reconciliation failed")
+	} else {
+		log.Info("Finished reconciliation")
+	}
+	return reconcile.Result{}, controllerutil.SwallowIfTerminal(err)
+}
+
+func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *logrus.Entry) error {
+
+	*log = *log.WithField("name", req.Name)
```
droslean, line 81: Why the pointers?

pkg/controller/hypershift_namespace_reconciler/hypershift_namespace_reconciler.go
```diff
@@ -0,0 +1,103 @@
...
+		if apierrors.IsNotFound(err) {
+			log.Info("The namespace is deleted")
+			return nil
+		}
+		return fmt.Errorf("failed to get the namespace %s: %w", req.Name, err)
+	}
+	if ns.Labels == nil {
+		return nil
+	}
+	value, ok := ns.Labels["hypershift.openshift.io/hosted-control-plane"]
```
droslean, line 95: If we need to operate only on the namespaces that hold this label, you need to make this check in the handler instead.

hongkailiu: Done!

pkg/controller/hypershift_namespace_reconciler/hypershift_namespace_reconciler.go
hongkailiu: Added the predicates.
However, we still need the error check here because there is a possibility that the ns is delete after the request is generated.

droslean (approved):

PR #3241, opened 2023-01-11
cluster-display: return zero value when err!=nil

hongkailiu:
/cc @psalajova @smg247


hongkailiu:
https://issues.redhat.com/browse/DPTP-3086

PR #3247, opened 2023-01-13
Limit concurrency of bigquery job uploads, fix memory spikes

dgoodwin:
[TRT-772](https://issues.redhat.com//browse/TRT-772)

Upload jobs for alerts, disruption, and test runs are all seemingly locking up in dpcr cluster. Logs stop, pod stays running forever.

In debugging the lockups in the uploader jobs, I found that removing the top level of concurrency, which appears to be queuing up every job unbounded and attempting to process them all as fast as the cpu can, the problem went away.  Upload times then got a little too long particularly for the test runs without any concurrency though.

This change adds a new implementation for concurrency at the job level bounding to 10 goroutines pulling from a channel until there's nothing left.

Note that there is a second layer of concurrency at the job run level, where we limit to 20 concurrent uploads. I think this means we could be doing up to 20*10 = 200 concurrent uploads to google storage. 

As a side bonus, this implementation no longer spikes up to 6GB as we try to process everything at once. Memory usage now seems to bounce around between 300-600MB.

I have added logging that gives a rough idea how far along the job queue we are. We don't know how many runs each job will have to upload, but it's something to give us an indication of how far along we are.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
@@ -8,6 +8,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/sirupsen/logrus"
 	"github.com/spf13/pflag"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
@@ -81,7 +82,7 @@ func NewDryRunInserter(out io.Writer, table string) BigQueryInserter {
 func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 	srcVal := reflect.ValueOf(src)
 	if srcVal.Kind() != reflect.Slice {
-		fmt.Fprintf(d.out, "INSERT into %v: %v\n", d.table, src)
+		logrus.Debugf("INSERT into %s: %v", d.table, src)
 		return
 	}
 
@@ -89,6 +90,7 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 		return
 	}
 
+	// Accumulate bulk insert debugging into a buffer so it's not mixed with concurrent logging from other goroutines:
 	buf := &bytes.Buffer{}
 	fmt.Fprintf(buf, "BULK INSERT into %v\n", d.table)
 	for i := 0; i < srcVal.Len(); i++ {
@@ -107,12 +109,11 @@ func (d dryRunInserter) Put(ctx context.Context, src interface{}) (err error) {
 			fmt.Fprintf(buf, "\tINSERT into %v: JobName=%v\n", d.table, s.JobName)
 
 		default:
-
 			// If we don't know the type, output something generic.
 			fmt.Fprintf(buf, "\tINSERT into %v: %#v\n", d.table, s)
 		}
 	}
-	fmt.Fprint(d.out, buf.String())
+	logrus.Debug(buf.String())
 
 	return nil
 }
--- a/pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
+++ b/pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sync/semaphore"
@@ -35,9 +36,19 @@ type allJobsLoaderOptions struct {
 	shouldCollectedDataForJobFn shouldCollectDataForJobFunc
 	getLastJobRunWithDataFn     getLastJobRunWithDataFunc
 	jobRunUploader              uploader
+	logLevel                    string
 }
 
 func (o *allJobsLoaderOptions) Run(ctx context.Context) error {
+	start := time.Now()
+
+	// Set log level
+	level, err := logrus.ParseLevel(o.logLevel)
+	if err != nil {
+		logrus.WithError(err).Fatal("Cannot parse log-level")
+	}
+	logrus.SetLevel(level)
+
 	logrus.Infof("Locating jobs")
 
 	jobs, err := o.ciDataClient.ListAllJobs(ctx)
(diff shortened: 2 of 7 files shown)


pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -197,12 +195,13 @@ func (o *jobLoaderOptions) Run(ctx context.Context) error {
 		}
 
 		currentUploaders.Add(1)
+		jrID := jobRunID
 		go func() {
```
DennisPeriquet, line 199: that will work, but what do you think about making it a parameter on the go func instead of a closure?

dgoodwin: changed

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
```diff
@@ -81,38 +81,35 @@ func NewDryRunInserter(out io.Writer, table string) BigQueryInserter {
...
+			logrus.Debugf("   INSERT into %s: JobName=%s\n", d.table, s.JobName)
 
 		default:
-
 			// If we don't know the type, output something generic.
-			fmt.Fprintf(buf, "\tINSERT into %v: %#v\n", d.table, s)
+			logrus.Debugf("   INSERT into %s: %#v\n", d.table, s)
 		}
 	}
-	fmt.Fprint(d.out, buf.String())
```
deads2k, line 115: Printing out here prevents interleaving with other threads. Are you sure you want to allow interleaving?

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -48,46 +59,64 @@ func (o *allJobsLoaderOptions) Run(ctx context.Context) error {
...
+// to the errChan for aggregation in the main thread.
+func (o *allJobsLoaderOptions) processJob(ctx context.Context, workerThread, jobCount int, jobChan <-chan jobrunaggregatorapi.JobRow, errChan chan<- error) {
+	for job := range jobChan {
 		jobLogger := logrus.WithFields(logrus.Fields{
-			"job": job.JobName,
+			"job":    job.JobName,
+			"worker": workerThread,
 		})
+		// log how many job remain to be processed
+		jobLogger.WithField("remaining", fmt.Sprintf("%d/%d", len(jobChan), jobCount)).Info("pulled job from queue")
```
deads2k, line 105: this message is only correct if the for loop adding the jobs to the channel has finished before this is called.

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -48,46 +59,64 @@ func (o *allJobsLoaderOptions) Run(ctx context.Context) error {
...
+		wg.Add(1)
+		go o.processJob(ctx, i, jobCount, jobChan, errChan)
+	}
+
 	for i := range jobs {
 		job := jobs[i]
+		jobChan <- job
+	}
+
+	wg.Wait()
```
deads2k, line 76: What is calling `Done` on the `wg` to be sure that this wait completes?

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -248,7 +280,6 @@ func (o *jobRunLoaderOptions) Run(ctx context.Context) error {
 	}
 
 	if err := o.uploadJobRun(ctx, jobRun); err != nil {
-		o.logger.WithError(err).Error("failed to upload jobrun to bigquery")
```
deads2k, line 251: why suppress the error?

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -48,46 +59,64 @@ func (o *allJobsLoaderOptions) Run(ctx context.Context) error {
...
+		go o.processJob(ctx, i, jobCount, jobChan, errChan)
+	}
+
 	for i := range jobs {
 		job := jobs[i]
+		jobChan <- job
+	}
+
+	wg.Wait()
+	close(jobChan)
```
deads2k, line 77: you need to close the channel before you wait for the jobChan consumer to be finished, right?

pkg/jobrunaggregator/jobrunaggregatorlib/big_query.go
dgoodwin: I'll go back to the buffer but log the output with logrus, main goal was just to have control over whether or not I see these verbosity.

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
dgoodwin: This is a critical catch thanks. I was working off a misleading example online. I added in the wg as a param  to processJob and then got in trouble again with it deadlocking because it was passing it by value. :) It's working now.

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
dgoodwin: Fixed.

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
dgoodwin: I've modified ordering here to populate the channel before we start the routines to process it.

pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
dgoodwin: This was from a recent change I did and got carried away, the error was logged at three levels. I don't have a personal rule on consistency for logging them low level or higher up so nothing gets missed, but in this case opted to kill off this higher level logging as it was covered lower.

dgoodwin:
/hold cancel

I think this is ready now. The wg.Done() was indeed missing, and then I did it wrong on my first attempt as well. I think it's healthy now and I'm successfully updating alerts and disruption locally so far.

PR #3252, opened 2023-01-18
pkg/util: improve pod pending check

bbguimaraes:
The goal of this function is to prevent a pod destined to be permanently
pending from delaying a test until the end of its configured timeout period
(usually several hours).

Any container can potentially be the cause of the problem, the current
distinction between infrastructure/test ones is not very useful.  The new
implementation checks all of them, at every point in the execution of the
pod.

/hold

Works fine in my tests and local E2E executions, but let's gather some extra
samples.


bbguimaraes:
Observer test.

/test e2e

pkg/util/pods.go
```diff
@@ -332,20 +323,67 @@ func podJobIsFailed(pod *corev1.Pod) bool {
...
-		if s.Name != "cp-secret-wrapper" && s.State.Running != nil {
-			return true
+		if s.State.Running != nil {
+			return nil
+		} else if w := s.State.Waiting; w != nil {
+			return check(prev, s.Name)
+		} else if t := s.State.Terminated; t != nil {
+			prev = t.FinishedAt.Time
+		} else {
+			panic(fmt.Sprintf("invalid container status: %#v", s))
```
danilo-gemoli, line 376: Do we ever `recover()` somewhere or does this cause the entire application to fail?

pkg/util/pods.go
```diff
@@ -193,16 +189,11 @@ func waitForPodCompletionOrTimeout(ctx context.Context, podClient kubernetes.Pod
...
-				if podHasStarted(pod) {
-					podSeenRunning = true
-				} else if time.Since(pod.CreationTimestamp.Time) > podStartTimeout {
-					message := fmt.Sprintf("pod didn't start running within %s: %s\n%s", podStartTimeout, getReasonsForUnreadyContainers(pod), getEventsForPod(ctx, pod, podClient))
-					logrus.Infof(message)
-					notifier.Complete(name)
-					return pod, errors.New(message)
-				}
+			if err := checkPending(*pod, pendingTimeout, time.Now()); err != nil {
+				err = fmt.Errorf("pod pending timeout: %w: %s\n%s", err, getReasonsForUnreadyContainers(pod), getEventsForPod(ctx, pod, podClient))
```
danilo-gemoli, line 193: So far, the pod (one of its containers) is stuck for whatever reason. What is going to happen next? Will it be eventually recollected by `k8s` somewhere in the future? If it won't, does it make sense to "manually" terminate it?

pkg/util/pods.go
bbguimaraes: The Kubernetes API guarantees that exactly one of the three states will be present.  This is an `assert`ion to detect unsatisfied preconditions, either from a real cluster or in tests.

pkg/util/pods.go
bbguimaraes: What happens exactly depends on the nature of the original problem, but in all cases the pod will eventually be removed when the TTL of the temporary namespace expires.

Note that this PR changes nothing about how this is handled.

danilo-gemoli: > removed when the TTL of the temporary namespace expires

It looks fine.

pkg/util/pods.go
danilo-gemoli: `panic()`-ing [disrupts the regular execution flow](https://go.dev/blog/defer-panic-and-recover) in favor of a `try {} catch {}`-ish style. It has definitely its use cases but I would rather prefer to return and handle a regular `error`, unless you are planning to start converting the codebase into something that follows [this](https://en.wikipedia.org/wiki/Design_by_contract).

bbguimaraes: There are two possibilities for this code to be reached:

- Synthetic data which are incorrect.
- Data from a server which has violated its promise, such that there is no telling what the actual state is.

I don't see how these can be recovered from and what the calling code would do to continue operating.

danilo-gemoli: > possibilities for this code to be reached:
- kubernetes introduces a new state for the containers.

As you said, it's likely that such an error can't be recovered but to me, giving the caller the choice to decide what to do with an `err` is more flexible than being abruptly terminated by the callee. By the way, such a "low level" detail is not worth blocking an entire PR, so...

bbguimaraes:
(for the record, that was a `release-controller` failure, we don't execute `ci-operator` in integration tests)

PR #3255, opened 2023-01-23
More logging to identify deadlock

dgoodwin:


pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -189,23 +193,30 @@ func (o *jobLoaderOptions) Run(ctx context.Context) error {
 	for jobRunID := range jobRunProcessingCh {
```
deads2k, line 193: 30s skim, so apologies if it it's a non-sequitor, but where is this channel closed?  this loop will not exit until jobRunProcessingCh is closed.

PR #3257, opened 2023-01-24
Enable possibility to change cloud assignment during dispatch

jmguzik:
Tests are WIP
/hold
/cc @hongkailiu 
@hongkailiu please check what you think about this solution if you have something against it.

### DIFF
--- a/pkg/dispatcher/config.go
+++ b/pkg/dispatcher/config.go
@@ -22,6 +22,8 @@ import (
 type Config struct {
 	// the job will be run on the same cloud as the one for the e2e test
 	DetermineE2EByJob bool `json:"determineE2EByJob,omitempty"`
+	// the job will have exchanged cloud label if mapping defined
+	CloudMapping map[api.Cloud]api.Cloud `json:"cloudMapping,omitempty"`
 	// the cluster cluster name if no other condition matches
 	Default api.Cluster `json:"default"`
 	// the cluster name for ssh bastion jobs
@@ -128,7 +130,7 @@ func (config *Config) DetermineClusterForJob(jobBase prowconfig.JobBase, path st
 	}
 
 	if config.DetermineE2EByJob {
-		if cloud := DetermineCloud(jobBase); cloud != "" {
+		if cloud := config.DetermineCloudMapping(jobBase); cloud != "" {
 			if clusters, ok := config.BuildFarmCloud[api.Cloud(cloud)]; ok {
 				if len(clusters) > 0 {
 					return api.Cluster(clusters[len(filepath.Base(path))%len(clusters)]), false, nil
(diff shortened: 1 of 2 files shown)


jmguzik:
```
cloudMapping:
  openstack-vexxhost: aws
```
^ this will be added to cfg to make the requirements work. The same method can be used for other mappings.

pkg/dispatcher/config.go
```diff
@@ -196,6 +198,15 @@ func isSSHBastionJob(base prowconfig.JobBase) bool {
 	return false
 }
 
+// DetermineCloudMapping determines if for a given cloud there is a replacement to map, eg for cost saving reasons
+func (config *Config) DetermineCloudMapping(jobBase prowconfig.JobBase) string {
+	cloud := DetermineCloud(jobBase)
+	if mapping, ok := config.CloudMapping[api.Cloud(cloud)]; ok {
+		cloud = string(mapping)
+	}
+	return cloud
```
hongkailiu, line 207: NIT:

```suggestion
	if mapping, ok := config.CloudMapping[api.Cloud(cloud)]; ok {
		return string(mapping)
	}
	return DetermineCloud(jobBase)
```

pkg/dispatcher/config.go
```diff
@@ -128,7 +130,7 @@ func (config *Config) DetermineClusterForJob(jobBase prowconfig.JobBase, path st
 	}
 
 	if config.DetermineE2EByJob {
-		if cloud := DetermineCloud(jobBase); cloud != "" {
+		if cloud := config.DetermineCloudMapping(jobBase); cloud != "" {
```
hongkailiu, line 133: Why does it have effect only when `config.DetermineE2EByJob=true`?

jmguzik: and of course it's only place where I spotted we care about cloud determination which is a base for this change

hongkailiu:
Planned to take a look at this today but did not manage.
I will try next week.

pkg/dispatcher/config.go
hongkailiu, line 207: My bad. I was trying to avoid `cloud := DetermineCloud(jobBase)` but found it is used in the `if`.

pkg/dispatcher/config.go
```diff
@@ -22,6 +22,8 @@ import (
 type Config struct {
 	// the job will be run on the same cloud as the one for the e2e test
 	DetermineE2EByJob bool `json:"determineE2EByJob,omitempty"`
+	// the job will have exchanged cloud label if mapping defined
```
hongkailiu, line 25:
```suggestion
	// the job will be run on the target cloud if it otherwise runs on the source cloud.
	// The field has effect only when DetermineE2EByJob is true.
```

hongkailiu (commented):
Only a comment on the GoDoc. Otherwise, LGTM.

PR #3261, opened 2023-01-25
Add cluster-control-plane-machine-set-operator to sippy/testgrid

stbenjam:
[TRT-794](https://issues.redhat.com//browse/TRT-794)

This adds
periodic-ci-openshift-cluster-control-plane-machine-set-operator-* to testgrid and sippy for data collection. It also alphabetizes the list.


damdo (commented):
Thanks for putting this together.
/lgtm

PR #3263, opened 2023-01-26
Use Testplatform BigQuery for ProwJobs to upload alert/disruption/test data from

dgoodwin:
After on-going problems with deadlocks, we decided to rewrite the top layer of the importers for alert, disruption, and test run data that is being uploaded into bigquery. 

This PR replaces all the concurrency code, and switches from discovering job runs via scanning GCS buckets to instead querying the bigquery table testplatform uploads all prowjobs to. 

We limit to 10 concurrent goroutines, each of which can be pulling from GCS or inserting in bigquery. 

The change relies on job run end times now, we lookup the last job run end time we inserted at the start, then query the testplatform prowjobs for any that ended since then, plus a 30 minute buffer. The code additinally lists all the prowjobs we already uploaded since that 30 minute buffer time, to avoid duplicate detection. (we do believe there are duplicates going in today, and bigquery will not prevent it)

Once we list the potentially new prowjobs to import, we compare each to see if it's from a job we care about, or if it's a job run ID we already have. If both checks pass we proceed to import the job as we used to.

I attempted to replace the querying of prowjob.json from GCS, because we have all that data from the testplatform table, I got a long way with this in terms of the uploaders, but then ran into problems with the aggregation and analysis code using what I'd just changed. I abandoned this effort to cut out one of the files we read, but I do have it locally and may take another shot later.

[TRT-772](https://issues.redhat.com//browse/TRT-772)

### DIFF
--- a/pkg/branchcuts/bumper/release-controller-config-bumper.go
+++ b/pkg/branchcuts/bumper/release-controller-config-bumper.go
@@ -81,21 +81,22 @@ func (b *ReleaseControllerConfigBumper) BumpFilename(filename string, _ *Release
 }
 
 /*
-	Candidate bumping fields:
-		.message
-		.mirrorPrefix
-		.name
-		.overrideCLIImage
-		.check.*.consistentImages.parent
-		.publish.*
-			.verifyBugs.previousReleaseTag
-				.name
-				.tag
-			.imageStreamRef.name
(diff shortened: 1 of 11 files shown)


dgoodwin:
/hold

Need to get a secret updated.

PR #3270, opened 2023-01-31
Trigger first promotion automatically if isTag does not exist

hongkailiu:
https://issues.redhat.com/browse/DPTP-2509

/cc @openshift/test-platform 

/hold

Need to watch the rolling

### DIFF
--- a/cmd/pipeline-controller/reconciler.go
+++ b/cmd/pipeline-controller/reconciler.go
@@ -187,10 +187,10 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request) error
 		return err
 	}
 	if testContexts != "" {
-		comment += "\nScheduling the execution of all matched `pipeline_run_if_changed` tests:" + testContexts
+		comment += "\n\nScheduling tests matching the `pipeline_run_if_changed` parameter:" + testContexts
 	}
 	if overrideContexts != "" {
-		comment += "\nOverriding unmatched contexts:\n" + "\\override " + overrideContexts
+		comment += "\n\nOverriding unmatched contexts:\n" + "/override " + overrideContexts
 	}
 	if err := r.ghc.CreateComment(pj.Spec.Refs.Org, pj.Spec.Refs.Repo, pj.Spec.Refs.Pulls[0].Number, comment); err != nil {
 		r.ids.Delete(composeKey(pj.Spec.Refs))


pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -239,3 +260,59 @@ func configIndexFn(in cioperatorapi.ReleaseBuildConfiguration) []string {
...
+	log.Debug("Handling CIOpConfig change")
+	// We only care about new additions
+	if len(delta.Added) == 0 {
+		return nil
+	}
+	slashSplit := strings.Split(delta.IndexKey, "/")
+	if len(slashSplit) != 2 {
+		return fmt.Errorf("got an index delta event with a key that is not a valid namespace/name identifier: %s", delta.IndexKey)
+	}
+	namespace, name := slashSplit[0], slashSplit[1]
```
bbguimaraes, line 280: Isn't the key the `organization`/`repository` of the configuration?

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -239,3 +260,59 @@ func configIndexFn(in cioperatorapi.ReleaseBuildConfiguration) []string {
...
+	if err := registryClient.Get(context.Background(), types.NamespacedName{Namespace: namespace, Name: name}, &ist); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("failed to get imagestreamtag %s/%s: %w", namespace, name, err)
+		}
+		ist = imagev1.ImageStreamTag{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+			},
+		}
```
bbguimaraes, line 291: Can't we just enqueue this object and handle `IsNotFound` in `reconcile` instead of performing essentially the same actions here?

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -55,21 +58,31 @@ func AddToManager(mgr controllerruntime.Manager, opts Options) error {
...
 	if err != nil {
 		return fmt.Errorf("failed to construct prowjobreconciler: %w", err)
 	}
-
+	releaseBuildConfigs := func(identifier string) ([]*cioperatorapi.ReleaseBuildConfiguration, error) {
+		return opts.CIOperatorConfigAgent.GetFromIndex(configIndexName, identifier)
+	}
 	log := logrus.WithField("controller", ControllerName)
+	go func() {
+		for delta := range configChangeChannel {
```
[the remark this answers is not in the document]
hongkailiu: I asked the same question while reading/copying-from
https://github.com/openshift/ci-tools/blob/946fba8a7b4ef5c0366085eaaab7abaa52dfc5d1/pkg/controller/test-images-distributor/test_images_distributor.go#L170

And then found:
https://stackoverflow.com/questions/8593645/is-it-ok-to-leave-a-channel-open#:~:text=It's%20OK%20to%20leave%20a,that%20no%20more%20data%20follows.

pkg/controller/promotionreconciler/reconciler.go
hongkailiu: This one confused me too. I might be still wrong now.

The key is determined by the `indexFunc` when adding the index.

https://github.com/openshift/ci-tools/pull/3270/files#diff-584fd395d199b51f651a3b4528bc23bfd07f3190509d659a1c1ad53d5f5bd248R58

which is `istRef.ISTagName()`
https://github.com/openshift/ci-tools/pull/3270/files#diff-584fd395d199b51f651a3b4528bc23bfd07f3190509d659a1c1ad53d5f5bd248R255

pkg/controller/promotionreconciler/reconciler.go
hongkailiu: The PR handles ONLY `IsNotFound` cases, i.e., ci-op's defines a promotion tag but we cannot find it on the cluster. Then we trigger a image job for it.

pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -55,21 +58,31 @@ func AddToManager(mgr controllerruntime.Manager, opts Options) error {
 	if err := opts.CIOperatorConfigAgent.AddIndex(configIndexName, configIndexFn); err != nil {
 		return fmt.Errorf("failed to add indexer to config-agent: %w", err)
 	}
-
+	configChangeChannel, err := opts.CIOperatorConfigAgent.SubscribeToIndexChanges(configIndexName)
```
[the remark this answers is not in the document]
hongkailiu: It is a very good question. Here is my understanding:
The events in this channel is not K8S events. They are from watching the changes on the ci-op's configuration files on the file system.
If the system is restarted, it will treat all promoted tags as new ones.
Then we check all tags on the cluster, if any of them does not exist, we reconcile.

hongkailiu (commented):
Thanks for the review. @bbguimaraes 
You must have remembered that I said many times about reading Alvaro's code and I still have lots of things that I do not understand.
Here it comes: I borrow the code more or less from `test-images-distributor`.
The best way to know the code is to taste it myself.

pkg/controller/promotionreconciler/reconciler.go
hongkailiu: I guess I got your point now.
Enqueuing an object that does not exist feel odd: How do we tell if the tag is deleted or it has not been created yet when it is reconciled.

PR #3271, opened 2023-01-31
Skip removing non-existed files

bear-redhat:
Do not error out when a file pending removal is not existed


bbguimaraes:
Why is this necessary?  Shouldn't the files not be added to `toRemove` in the first place?

A comment in the main loop mentions the additions are based on a "heuristic", does it no longer apply?  And why?

PR #3272, opened 2023-01-31
Inject 0s for all known Alert combos when uploading to BigQuery

dgoodwin:
Origin's responsibility will be to observe alerts during a test run only, it doesn't know the full list of all possible alerts.

When we upload to bigquery using the tools in this repo, we will add 0s for all alerts we know are possible for that release. The list of all known alerts will be everything we've ever observed for that release.

This change compares the incoming observed alerts against the known list, and adds 0s for anything known but not in the incoming alerts.


pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -925,3 +927,33 @@ func GetUTCDay(in time.Time) time.Time {
 	year, month, day := in.UTC().Date()
 	return time.Date(year, month, day, 0, 0, 0, 0, time.UTC)
 }
+
+func (c *ciDataClient) ListAllKnownAlerts(ctx context.Context) ([]*jobrunaggregatorapi.KnownAlertRow, error) {
+	queryString := c.dataCoordinates.SubstituteDataSetLocation(
+		`SELECT *
+FROM DATA_SET_LOCATION.Alerts_AllKnown
```
xueqzhan, line 934: Who is updating this table? Is this a table that is automatically updated based on other table (e.g. Alerts) content?

dgoodwin: It is a view defined in BigQuery itself, here is the current query for it:

```sql
Select 
  AlertName, 
  AlertNamespace, 
  AlertLevel,
  Release, 
  min(JobRunStartTime) as FirstObserved, 
  max(JobRunStartTime) AS LastObserved, 
  count(*) AS Results
FROM `openshift-ci-data-analysis.ci_data.Alerts_UnifiedAlerts`
WHERE AlertSeconds > 0  
  AND (AlertNamespace LIKE 'openshift%' 
    OR AlertNamespace = "kube-system" 
    OR AlertNamespace = "default"
    OR AlertNamespace = "")
GROUP By AlertName, AlertLevel, AlertNamespace, Release
ORDER BY Release, AlertLevel, AlertName, AlertNamespace
```

danilo-gemoli (approved):

PR #3273, opened 2023-02-02
Include operating on manifests and create the corresponding images to the graph

droslean:
The tool now gets only the release repo path as a flag and hardcodes the required subdir that we need to operate on. We now read all YAML files in `clusters/app.ci` directory and load all `ImageStream` and `BuildConfigs`. 

/cc @danilo-gemoli @openshift/test-platform 

### DIFF
--- a/pkg/image-graph-generator/manifests.go
+++ b/pkg/image-graph-generator/manifests.go
@@ -1,0 +1,172 @@
+package imagegraphgenerator
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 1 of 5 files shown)


pkg/image-graph-generator/manifests.go
```diff
@@ -0,0 +1,172 @@
...
+				return nil
+			}
+
+			d, err := os.ReadFile(path)
+			if err != nil {
+				return err
+			}
+
+			// The YAML file could hold multiple objects separated by `---`
+			dataList := strings.Split(string(d), "---")
```
danilo-gemoli, line 48: Isn't a better way to unmarshal multi-document yaml file?

danilo-gemoli:
It looks good but the package `imagegraphgenerator` misses unit tests and we should also consider writing integration tests for the whole tool.

pkg/image-graph-generator/manifests.go
droslean: Nope :(

droslean:
> It looks good but the package `imagegraphgenerator` misses unit tests and we should also consider writing integration tests for the whole tool.

Let me work on adding some unittests and updating the PR. Or we can merge this one and I can add them in a follow-up

PR #3276, opened 2023-02-06
Fix issue DPTP-3312 Clean up leaked resources

bear-redhat:
Clean up HyperShift created but expired AWS resources.

### DIFF
--- a/images/dptp-3312-hypershift-leaks-cleaner/Dockerfile
+++ b/images/dptp-3312-hypershift-leaks-cleaner/Dockerfile
@@ -1,0 +1,13 @@
+FROM quay.io/centos/centos:stream8
+LABEL maintainer="bechen@redhat.com"
+
+RUN yum install -y unzip jq
+RUN cd /tmp && \
+    curl "https://awscli.amazonaws.com/awscli-exe-linux-x86_64.zip" -o "awscliv2.zip" && \
+    unzip awscliv2.zip && \
+    ./aws/install && \
+    rm -rf awscliv2.zip aws
+
+ADD cleanup-hypershift-leaks.sh /cleanup-hypershift-leaks.sh
+ENTRYPOINT ["sh"]
+CMD ["/cleanup-hypershift-leaks.sh"]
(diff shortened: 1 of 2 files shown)


hack/cleanup-hypershift-leaks.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
+export NOW=$(date +%s)
```
jmguzik, line 2: does it need to be export?

hack/cleanup-hypershift-leaks.sh
```diff
@@ -0,0 +1,59 @@
...
+echo Cleanning up resources older than $(date -u -d @$NOW +%Y-%m-%dT%H:%M:%SZ)
+
+echo 'Deleting expired OpenID Connect Providers...'
+# Get all OpenID Connect Providers with ARN
+arns=`aws iam list-open-id-connect-providers --query 'OpenIDConnectProviderList[*].Arn' --output json | jq -r '.[]'`
```
jmguzik, line 7: what other tools are needed as dependencies of this script? I would mark that somehow.

hack/cleanup-hypershift-leaks.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
jmguzik, line 1: I would generally make an image out of this script if we would like to run it as periodic. I would move it to images/ and compose a dockerfile

jmguzik:
See https://github.com/openshift/ci-tools/blob/master/images/quayio-pull-through-cache/Dockerfile in this repo. If you want to go this way, you need to make changes in the release as well, I can guide you through that.

bear-redhat:
@jmguzik  I converted the tool to be a docker image. I will add the job part once its merged.

images/dptp-3312-hypershift-leaks-cleaner/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM amazon/aws-cli:latest
```
jmguzik, line 1: If that is coming from dockerhub (as I think) we need to think about a different solution. It is rate-limited on our cluster. The image will not be downloaded.

jmguzik (changes requested):

jmguzik (approved):

PR #3278, opened 2023-02-06
Add more logs to the job primer and skip jobs without a Platform

DennisPeriquet:
[TRT-826](https://issues.redhat.com//browse/TRT-826)

When the `Jobs` table (for disruption testing) in Big Query is updated, the logs are terse so you cannot see what got added.

This PR:

* adds more logging so you can see what jobs are added
* prevents jobs from being added where the `Platform` column/attribute could not be determined (a warning is also emitted)

These changes prevent errors going forward, but the current `Jobs` table has entries that don't have a `Platform` value.  To fix this, I believe we must destroy and re-create the `Jobs` table (as big query tables are immutable).

### DIFF
--- a/pkg/jobrunaggregator/jobtableprimer/create_jobs.go
+++ b/pkg/jobrunaggregator/jobtableprimer/create_jobs.go
@@ -37,10 +37,17 @@ func (o *CreateJobsOptions) Run(ctx context.Context) error {
 			continue
 		}
 
+		if len(jobToCreate.Platform) == 0 {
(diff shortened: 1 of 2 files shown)


DennisPeriquet:
If we merged this, no one would notice unless we periodically inspected the logs -- this is not practical.  Since we can now easily delete/recreate the Jobs table in case we notice it has missing data (when looking at disruption) we can know to take action by correcting the data and deleting the Jobs table (it will then get re-created automatically since now creating the table is part of the job-table-updater CronJob).

PR #3282, opened 2023-02-07
Use Go 1.19

bbguimaraes:


bbguimaraes:
Some strange `git` errors, maybe because of:

```
# podman run --rm -it --entrypoint git registry.ci.openshift.org/ci/ci-tools-build-root:1.18 --version
git version 2.31.1
# podman run --rm -it --entrypoint git registry.ci.openshift.org/ci/ci-tools-build-root:1.19 --version
git version 2.39.0
```

I'll check tomorrow.

danilo-gemoli:
No, `ci/prow/unit` failing doesn't look good at all. That means we have to open another PR addressing what is missing whereas it should have really been part of the #3293 instead.

jmguzik:
> No, `ci/prow/unit` failing doesn't look good at all. That means we have to open another PR addressing what is missing whereas it should have really been part of #3293 instead.

Yeah, it seems so. Fortunately, it's only one job out of 7.

PR #3284, opened 2023-02-08
remove disrupt 100 run filter

dgoodwin:
We recently expanded the alert data file to include namespace and level. 
Origin is being updated to skip anything in the data file with less than 100 job runs, but previously we actually filtered them out entirely via the query, at least for disruption. 

This change makes disruption include them as well, to be consistent. It will be up to origin to decide what to enforce.

This also updates the PR message to show the new data for alerts, and hide rows from the tables with less than 100 runs, as we do not care to see whether they are up or down.

[TRT-769](https://issues.redhat.com//browse/TRT-769)


dgoodwin:
/hold

This needs https://github.com/openshift/origin/pull/27710 to land first or we'll start enforcing disruption on low job run counts.

dgoodwin:
/hold cancel

Origin is ready.

dgoodwin:
@DennisPeriquet had to rebase, can I get another lgtm plz.

PR #3291, opened 2023-02-10
`pod-scaler`: add priority scheduling class when CPU request is above a certain amount

smg247:
We need to add a `priorityClassName` of `openshift-user-critical` when a pod is either configured or determined to need above `8` CPU. I made the amount configurable in case we need to tune it in the future.

For: https://issues.redhat.com/browse/DPTP-3232


bbguimaraes:
Do we know:

- Is this the 30m pending timeout in `ci-operator`?
- Why is this happening?  Shouldn't the auto-scaler provision more nodes if there are pods which are prevented from being scheduled due to capacity?  Or is the problem that it does not do so in time?

But does a CPU request this high even make sense?  Isn't this what the caps in the admission controller are for?

- Our current allowed maximum is `10` (silently enforced by the admission controller), our build clusters have 8/16 cores.
- A test with enough unbounded CPU usage will just request increasing amounts indefinitely, eventually getting to the maximum.
- Since it's a compressible resource, requesting a higher amount only controls the scheduling of the different containers, after the pod is created on a node with enough capacity.

Given all this, why not just lower the limit to something which can be normally scheduled in our clusters?  This will make tests which really need that amount of CPU time execute more slowly, but I'm not sure letting a test take 50-100% of a node is preferable.

smg247:
@bbguimaraes I do believe this is the 30m pending timeout in `ci-operator` causing this. There is a [thread](https://redhat-internal.slack.com/archives/CBN38N3MW/p1669828305893639) where Justin proposed this as a solution to the problem.

I think it is a rare case that CPU requests are this high, but I still think we should accommodate them.

PR #3292, opened 2023-02-13
entrypoint-wrapper: create Git configuration

bbguimaraes:
Newer versions of Git refuse to interact with a repository owned by another
user, independently of file permissions.  This is a problem in CI since
source code is usually cloned by UID 0 and read by unprivileged users.

With this change, `entrypoint-wrapper` creates a Git configuration file in
the home directory if it does not already exist (n.b.: this directory is
already provided as a temporary `EmptyDir` volume to `entrypoint-wrapper`
in multi-stage tests).  This file will disable the ownership verification
using the following entry:

    [safe]
        directory = *

Tests which want to have better control over their configuration can
rewrite this file or use one of the other configuration methods in Git:

- https://git-scm.com/docs/git-config.html#FILES
- https://git-scm.com/docs/git-config.html#SCOPES

---

For review, testing this with our current CI will be… interesting.
Documentation pull request will come soon.

/hold

### DIFF
--- a/cmd/entrypoint-wrapper/main.go
+++ b/cmd/entrypoint-wrapper/main.go
@@ -6,6 +6,7 @@ import (
 	"flag"
 	"fmt"
 	"io"
+	"io/fs"
 	"io/ioutil"
 	"log"
 	"os"
@@ -277,7 +278,10 @@ func (o *options) execCmd() error {
 		// the inherited set instead of overwriting
 		proc.Env = os.Environ()
 	}
-	manageHome(proc)
+	home := manageHome(proc)
+	if err := manageGitConfig(home); err != nil {
+		return fmt.Errorf("failed to create Git configuration: %w", err)
+	}
 	manageCLI(proc)
 	if o.rwKubeconfig {
 		if err := manageKubeconfig(proc); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/entrypoint-wrapper/main.go
```diff
@@ -326,9 +330,34 @@ func manageHome(proc *exec.Cmd) {
 		}
 	}
 	if needHome {
+		home = "/alabama"
```
jmguzik, line 333: I am glad alabama is still here :)

cmd/entrypoint-wrapper/main.go
```diff
@@ -326,9 +330,34 @@ func manageHome(proc *exec.Cmd) {
...
+	}
+	return home
+}
+
+// manageGitConfig creates a default Git configuration file if necessary
+// This configuration will contain a `safe.directory` entry disabling the file
+// ownership verification (see `git-config(1)`).  If an existing configuration
+// file is found, no changes are made.
+func manageGitConfig(home string) error {
+	path := filepath.Join(home, ".gitconfig")
```
jmguzik, line 345: In this solution it does not matter, but do you think in entrypoint it could be just always `/etc/gitconfig`?

jmguzik: Not always, there is mysterious prefix before `/etc/gitconfig`

jmguzik (approved):

jmguzik:
/override  ci/prow/codegen ci/prow/format ci/prow/integration ci/prow/unit ci/prow/validate-vendor ci/prow/e2e ci/prow/frontend-checks

PR #3310, opened 2023-02-28
Consider jobRuns threshold when transitioning to use new release data

xueqzhan:
Count alert and disruption data from new releases only when the total count passes a certain threshold

[TRT-836](https://issues.redhat.com/browse/TRT-836)


dgoodwin:
/override ci/prow/e2e

Something going on with quay throttling, in any case not related to the trt area of this repo.

PR #3313, opened 2023-03-02
Operate on builds based on the list of architectures supported in the cluster

droslean:
/cc @openshift/test-platform @bbguimaraes 

1. Add the node architectures in the build client
2. We now add a nodeSelector to every build to specify the node with the desired architecture. E.x `kubernetes.io/arch: amd64` 
3. Spawn a build for each architecture for each image.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -832,8 +832,18 @@ func (o *options) Run() []error {
 
 	o.resolveConsoleHost()
 
+	client, err := coreclientset.NewForConfig(o.clusterConfig)
+	if err != nil {
+		return []error{fmt.Errorf("could not get core client for cluster config: %w", err)}
+	}
+
+	nodeArchitectures, err := resolveNodeArchitectures(ctx, client.Nodes())
+	if err != nil {
+		return []error{fmt.Errorf("could not resolve the node architectures: %w", err)}
+	}
+
 	// load the graph from the configuration
-	buildSteps, postSteps, err := defaults.FromConfig(ctx, o.configSpec, &o.graphConfig, o.jobSpec, o.templates, o.writeParams, o.promote, o.clusterConfig, o.podPendingTimeout, leaseClient, o.targets.values, o.cloneAuthConfig, o.pullSecret, o.pushSecret, o.censor, o.hiveKubeconfig, o.consoleHost, o.nodeName)
+	buildSteps, postSteps, err := defaults.FromConfig(ctx, o.configSpec, &o.graphConfig, o.jobSpec, o.templates, o.writeParams, o.promote, o.clusterConfig, o.podPendingTimeout, leaseClient, o.targets.values, o.cloneAuthConfig, o.pullSecret, o.pushSecret, o.censor, o.hiveKubeconfig, o.consoleHost, o.nodeName, nodeArchitectures)
 	if err != nil {
 		return []error{results.ForReason("defaulting_config").WithError(err).Errorf("failed to generate steps from config: %v", err)}
 	}
@@ -889,10 +899,6 @@ func (o *options) Run() []error {
 			if err := o.initializeLeaseClient(); err != nil {
 				return []error{fmt.Errorf("failed to create the lease client: %w", err)}
 			}
-		}
-		client, err := coreclientset.NewForConfig(o.clusterConfig)
-		if err != nil {
-			return []error{fmt.Errorf("could not get core client for cluster config: %w", err)}
 		}
 		go monitorNamespace(ctx, cancel, o.namespace, client.Namespaces())
 		authClient, err := authclientset.NewForConfig(o.clusterConfig)
@@ -2058,6 +2064,20 @@ func (o *options) loadConfig(info *api.Metadata) (*api.ReleaseBuildConfiguration
 		}
 	}
 	return &configSpec, nil
+}
+
+func resolveNodeArchitectures(ctx context.Context, client coreclientset.NodeInterface) ([]string, error) {
+	ret := sets.NewString()
+	nodeList, err := client.List(ctx, meta.ListOptions{})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine the node architectures: %w", err)
+	}
+
+	for _, node := range nodeList.Items {
+		ret.Insert(node.Status.NodeInfo.Architecture)
+	}
+	return ret.List(), nil
 }
 
 func monitorNamespace(ctx context.Context, cancel func(), namespace string, client coreclientset.NamespaceInterface) {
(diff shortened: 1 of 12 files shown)


pkg/steps/source.go
```diff
@@ -368,6 +370,52 @@ func isBuildPhaseTerminated(phase buildapi.BuildPhase) bool {
 	return true
 }
 
+func handleBuilds(ctx context.Context, buildClient BuildClient, build buildapi.Build) error {
+	var wg sync.WaitGroup
+	errChan := make(chan error)
+
+	for _, build := range constructMultiArchBuilds(build, buildClient.NodeArchitectures()) {
```
danilo-gemoli, line 377: What about handling context cancellation like the following?
```golang
for _, build := range constructMultiArchBuilds(build, buildClient.NodeArchitectures()) {
	select {
		case <-ctx.Done():
		default:
	}
	wg.Add(1)
	go func(b buildapi.Build) {
		defer wg.Done()

		if err := handleBuild(ctx, buildClient, b); err != nil {
			errChan <- fmt.Errorf("error occurred handling build %s: %w", b.Name, err)
		}

	}(build)
}
```
This would prevent some extra calls of `handleBuild` if  `ctx` was cancelled before the loop started or even in the middle of it.

droslean: This should be already handled in `handleBuild`. Also, once the context is canceled we need to cancel all builds and not continue.

danilo-gemoli: Yes the context is also handled downstream and, once canceled, every build cancels as well in both the implementation, but as an example:
```golang
for _, build := []string{"amd64", "x86", "arm64"} {
	go handleBuild(ctx, build)
}
```
it keeps processing `"x86"` and `"arm64"` even if the context get canceled after the first iteration `"amd64"`.
On the contrary:
```golang
for _, build := []string{"amd64", "x86", "arm64"} {
	select {
		case <-ctx.Done():
		    break
                default:
	}
	go handleBuild(ctx, build)
}
```
it stops right away after `"amd64"`.

droslean: That case will never happen either way. We need to spawn all of the builds simultaneously and if the context is canceled then each of the goroutines should handle the cancelation separately.

bbguimaraes: It can definitely happen, but it's unlikely that cancellation will happen between the creation of those threads.  In this particular case, it's fine to let each `handleBuild` return an error as soon as the cancellation happens.

But there is a deeper problem, of which this is not the first instance, but may be the worst.  `handleBuild` ultimately uses the `client-go` infrastructure, and each of these calls being completely independent means each will have its own informer, index, watch, etc.  This is already the case when we have multiple unrelated builds across steps, but here it is even more outrageous since they are are being created in a single place.

cmd/ci-operator/main.go
```diff
@@ -2060,6 +2064,17 @@ func (o *options) loadConfig(info *api.Metadata) (*api.ReleaseBuildConfiguration
 	return &configSpec, nil
 }
 
+func (o *options) resolveNodeArchitectures(ctx context.Context, client coreclientset.NodeInterface) {
+	nodeList, err := client.List(ctx, meta.ListOptions{})
+	if err != nil {
+		logrus.WithError(err).Warn("failed to determine the node architectures, defaulting to amd64...")
```
bbguimaraes, line 2070: Why not return the error?  Returning `AMD64Arch` will result in an unschedulable `Build` if the cluster has no nodes with that architecture.

bbguimaraes (changes requested):
Some comments on the implementation below, but this is missing some… architectural explanation.

For instance:

- How will this cluster (these clusters?) be incorporated in our collection of build clusters?  Will it/they be responsible for _all_ image build post-submit jobs?
- As mentioned in the issue ([DPTP-3319](https://issues.redhat.com/browse/DPTP-3319)) and in a previous discussion, using node information has the potential to miss architectures for which there are currently no nodes in the cluster.  Have we decided this is the desired behavior?  Why?
- Does a heterogeneous cluster not have the information of which architectures it contains?  Our build clusters get fairly big when stressed, and from my experience listing nodes is a very slow (so expensive, I suppose) operation.
- Do we want to always trust this list and not have a list of architectures for which we definitely want to produce images?

cmd/ci-operator/main.go
droslean: > Some comments on the implementation below, but this is missing some… architectural explanation.
> 
> For instance:
> 
> * How will this cluster (these clusters?) be incorporated in our collection of build clusters?  Will it/they be responsible for _all_ image build post-submit jobs?
> * As mentioned in the issue ([DPTP-3319](https://issues.redhat.com/browse/DPTP-3319)) and in a previous discussion, using node information has the potential to miss architectures for which there are currently no nodes in the cluster.  Have we decided this is the desired behavior?  Why?
> * Does a heterogeneous cluster not have the information of which architectures it contains?  Our build clusters get fairly big when stressed, and from my experience listing nodes is a very slow (so expensive, I suppose) operation.
> * Do we want to always trust this list and not have a list of architectures for which we definitely want to produce images?

- Once we start introducing the heterogenous clusters, I would expect test and image promotions should work the same way but with the multiple manifest lists included. We need to make sure that we build all the images for all architectures on tests as well because otherwise, we will have to nodeSelect every pod. If the architecture is available in the image, then the pod will pull the correct manifest despite the node that it's running. We will start by scheduling the postsubmits in the heterogenous clusters to build the image for all available architectures. (This requires app.ci to be 4.13 version)
- I had a discussion with the cloud team and it seems that there is no way to get the architecture from a machine set. The only case we need to cover is that if a machineset is scaled down to 0 but the implementation is still in progress. They suggested manually maintaining the machineset and applying a specific label that we maintain. I would prefer to start by gathering the available architectures from the active nodes, and as a follow-up we can cover the scaled-down-to-0 machinesets design...
- Nope, there is no way to get the architectures. I tried to find it but these kind of clusters are work-in-progress too.
- I would say to trust this list because it will still work once we add more architectures in any of the clusters.

bbguimaraes: That takes care of my initial questions, thanks.

cmd/ci-operator/main.go
```diff
@@ -2060,6 +2066,20 @@ func (o *options) loadConfig(info *api.Metadata) (*api.ReleaseBuildConfiguration
...
+func resolveNodeArchitectures(ctx context.Context, client coreclientset.NodeInterface) ([]string, error) {
+	ret := sets.NewString()
+	nodeList, err := client.List(ctx, meta.ListOptions{})
+
+	if err != nil {
+		return nil, fmt.Errorf("failed to determine the node architectures: %w", err)
+	}
+
+	for _, node := range nodeList.Items {
+		ret.Insert(node.Status.NodeInfo.Architecture)
```
droslean, line 2078: It's safe to grab the architecture from the status. https://redhat-internal.slack.com/archives/CK1AE4ZCK/p1678100046793209

PR #3314, opened 2023-03-02
DPTP-3326: Allow clusterpools to specify a release stream to select the release version from

bparees:
See https://github.com/openshift/release/pull/36890 for sample usage

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -279,17 +279,15 @@ type Prerelease struct {
 	ReleaseDescriptor
 	// VersionBounds describe the allowable version bounds to search in
 	VersionBounds VersionBounds `json:"version_bounds"`
-
-	// Stream allows for custom release streams to be selected,
-	// e.g. 4-dev-preview.  The architecture suffix should not be
-	// included.  Defaults to 4-stable.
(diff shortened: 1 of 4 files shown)


bparees:
@bear-redhat  here's a pass at modifying clusterimageset-updater to allow a clusterpool+clusterimageset to specify a release stream.  ptal.

pkg/api/types.go
```diff
@@ -279,17 +279,15 @@ type Prerelease struct {
...
 }
 
-// VersionBounds describe the upper and lower bounds on a version search
+// VersionBounds describe the upper and lower bounds and stream on a version search
 type VersionBounds struct {
 	Lower string `json:"lower"`
 	Upper string `json:"upper"`
+	// Stream dictates which stream to search for a version within the specified bounds
+	// defaults to 4-stable.
+	Stream string `json:"stream"`
```
bparees, line 290: I didn't see anywhere that this was being used/set so I folded it into VersionBounds which kept the code a little cleaner, but if you know of a reason this was useful(maybe vendored somewhere?), let me know.

bparees:
@bear-redhat bump

PR #3315, opened 2023-03-03
ClusterPool: Polulate the owner label

hongkailiu:
To follow up https://github.com/openshift/release/pull/36902#issuecomment-1453523369

/cc @openshift/test-platform @2uasimojo


hongkailiu:
/hold

Will watch the rolling.

cmd/clusterimageset-updater/main.go
```diff
@@ -65,6 +75,25 @@ func main() {
...
+	}
+
+	s := json.NewSerializerWithOptions(json.DefaultMetaFactory, scheme.Scheme,
+		scheme.Scheme, json.SerializerOptions{Yaml: true, Pretty: false, Strict: false})
+
+	if err := filepath.WalkDir(o.poolDir, func(path string, info fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.IsDir() || !strings.HasSuffix(info.Name(), "_clusterpool.yaml") {
```
2uasimojo, line 89: Is this really the best way to detect a clusterpool definition? as opposed to maybe some golang equivalent of `grep 'Kind: ClusterPool'`?

cmd/clusterimageset-updater/main.go
```diff
@@ -223,6 +252,47 @@ func main() {
...
+			pool.Spec.Labels = map[string]string{}
+		}
+		if pool.Spec.Labels["tp.openshift.io/owner"] != value {
+			modified = true
+			pool.Spec.Labels["tp.openshift.io/owner"] = value
+		}
+	} else {
+		logrus.WithField("namespace", pool.Namespace).WithField("name", pool.Name).
+			Warn("The cluster pool has no owner label")
+	}
```
2uasimojo, line 292: Would it be worth enforcing that this `owner` label is required?

2uasimojo (commented):
This seems sane.

2uasimojo:
The `integration` fail looks legit.

cmd/clusterimageset-updater/main.go
hongkailiu: It has been used in other places of this tool already and it turned to be OK.
The damage is very small even if it is not accurate.  :)

cmd/clusterimageset-updater/main.go
hongkailiu: I thought about this too and my logic is
- Once this is merged and proven working in our production, I will update https://docs.ci.openshift.org/docs/how-tos/cluster-claim/ to tell that the owner label will bring good impact for debugging which should be enough to convince the owners to do the right thing, hopefully. The log here is just to show which pools do not use it at the moment.
- There is a default value by Hive already.

Screenshot 2023-03-03 at 4 04 59 PM

hongkailiu: Even better: It is kind of implicitly enforced already, i.e., if a pool has no owner label, it will never be selected by any CI test.

https://github.com/openshift/ci-tools/blob/ecd2e4e9bb1f043ee926d99b8738b6969696a686/pkg/validation/test.go#L473-L475

2uasimojo:
awesome

/lgtm

hongkailiu:
Seems working.
https://github.com/openshift/release/pull/36985/commits/b55fdf18a703371713b94ed9592f439fc0717da9

PR #3320, opened 2023-03-09
Fall back to older releases disruption data during release transition

xueqzhan:
[TRT-891](https://issues.redhat.com/browse/TRT-891)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -203,9 +206,78 @@ func (a *weeklyAverageFromTenDays) getAggregatedTestRuns(ctx context.Context) (m
 	return a.aggregatedTestRunsByName, a.queryTestRunsErr
 }
 
+func getMajor(in string) int {
```
stbenjam, line 209: Hashicorp go-version is vendored here. Would it work for your needs?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -203,9 +206,78 @@ func (a *weeklyAverageFromTenDays) getAggregatedTestRuns(ctx context.Context) (m
...
+	major, err := strconv.ParseInt(strings.Split(in, ".")[0], 10, 32)
+	if err != nil {
+		panic(err)
```
stbenjam, line 212: Can we bubble up the error with some context (i.e. the version string we failed on) instead of panicking?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
xueqzhan: Good suggestion. I will look at this tomorrow to see if it works. I stole the code from origin. :)

stbenjam: It's not a big deal either way, if we use this code in other places it's fine too

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -203,9 +206,78 @@ func (a *weeklyAverageFromTenDays) getAggregatedTestRuns(ctx context.Context) (m
...
+			if j.Architecture == job.Architecture &&
+				j.Topology == job.Topology &&
+				j.Network == job.Network &&
+				j.Platform == job.Platform &&
+				j.FromRelease == targetFromRelease &&
+				j.Release == targetToRelease &&
+				j.IPMode == job.IPMode {
+				return j.JobName, nil
+			}
+		}
```
neisw, line 260: Just curious if we have tested this / know that the match covers everything like  RT

periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-gcp-ovn-rt-upgrade

Any chance we could swap an RT job with 

periodic-ci-openshift-release-master-ci-4.14-upgrade-from-stable-4.13-e2e-gcp-ovn-upgrade

xueqzhan: Good question. Job table does not have anything to differentiate RT vs non-RT jobs other than just the name. This mean, when we fall back, we might pick the other variant (RT fall back to non-RT or non-RT fall back to RT). Will that be a problem?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
xueqzhan: go-version does not manipulate version number. So it does not do what we want here.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
neisw: Technically the values could be different.  I was wondering if we could scrub the original & target release versions from the job names and then do a compare, everything should match at that point.

neisw: My only comment here is that is very specific to RT, don't know if that is the only case where a setting is captured in the job name but not the variants.  Just a thought.

xueqzhan: Long term plan is to have RT as a new variant. Name based matching is only an exception.

neisw: Looks to be more than just RT based on 

select count (distinct(JobName)), Platform, Network, IPMode, Topology, Release, FromRelease, IsUpgrade FROM `openshift-ci-data-analysis.ci_data.UnifiedJobRuns`  where Release = '4.14' group by Cluster, Platform, Network, IPMode, Topology, Release, FromRelease, IsUpgrade

and

select distinct(JobName) FROM `openshift-ci-data-analysis.ci_data.UnifiedJobRuns` where Platform = 'aws' and Network = 'sdn' and IPMode = 'ipv4' and Topology = 'ha' and Release = '4.14' and FromRelease = '' and IsUpgrade = false

Will review further with Ken

neisw: Patch to consider.

[ci-tools-fallback-normalize-jobname.patch.txt](https://github.com/openshift/ci-tools/files/11017851/ci-tools-fallback-normalize-jobname.patch.txt)

xueqzhan: Thanks! Took that.

PR #3323, opened 2023-03-13
`pod-scaler`: add priorityClassName when CPU above 8

smg247:
We need to add a priorityClassName of openshift-user-critical when a pod is either configured or determined to need above 8 CPU. I made the amount configurable in case we need to tune it in the future.

This reverts the revert that was made to kill a bug preventing pods from scheduling when `priority` was non-nil. We can only have one of `priority` and `priorityClassName` set. Since this class name is the highest priority, we should always allow it to override the previous `priority` value.

For: https://issues.redhat.com/browse/DPTP-3232


smg247:
/cc @bear-redhat 
Since you helped to merge the original PR

PR #3325, opened 2023-03-13
ci-operator: print pruned graph

bbguimaraes:
This allows printing a sub-graph based on the `--target` arguments.
Printing the entire graph is still possible by simply omitting the
argument.

```console
$ ci-operator … --unresolved-config openshift-ci-tools-master.yaml --print-graph --target e2e
…
INFO[2023-03-10T19:52:30Z] Running [input:root], src, test-bin, e2e
src [input:root]
test-bin src
e2e test-bin
INFO[2023-03-10T19:52:30Z] Ran for 3s
```


bbguimaraes:
```console
$ ci-operator … --print-graph --target e2e | hack/graphviz.sh > out.png
```

out

cmd/ci-operator/main.go
```diff
@@ -1823,14 +1822,12 @@ func nodeNames(nodes []*api.StepNode) []string {
...
-		for _, other := range steps {
-			if step == other {
-				continue
-			}
-			if api.HasAnyLinks(step.Requires(), other.Creates()) {
-				if _, err := fmt.Fprintf(w, "%s %s\n", step.Name(), other.Name()); err != nil {
+func printDigraph(w io.Writer, steps api.OrderedStepList) error {
+	for i, step := range steps {
+		req := step.Step.Requires()
+		for _, other := range steps[:i] {
```
danilo-gemoli, line 1828: Please add the following (or similar) comment here:
```golang
// steps have already been topologically sorted, 
// it means that if `a` depends on `b` then `b` will come before `a` on the list. 
// As a consequence, any dependencies of steps[i] will be found within steps[:i]
```

hack/graphviz.sh
```diff
@@ -0,0 +1,31 @@
...
+}
+
+END {
+    print "}"
+}
+EOF
+)"
+
+[[ "$#" -eq 0 ]] && set -- -T png
+sort | uniq | awk "$awk_prog" | dot "$@"
```
danilo-gemoli, line 31: Why do we need to `sort | uniq` ? Is the following duplication (as an example) possible?
```
bin src
...
bin src
```

danilo-gemoli (changes requested):

cmd/ci-operator/main.go
bbguimaraes: Added.

hack/graphviz.sh
bbguimaraes: You're right, step names are unique, so that is not possible.

Probably something for debugging which I forgot to remove, done now.

danilo-gemoli:
Thanks!
/lgtm

PR #3326, opened 2023-03-13
`pj-rehearse`: properly bubble up errors from loading configuration

smg247:
We were logging errors from loading the release configuration, but not actually bubbling them up so that the user could see them. This change throws errors instead of just logging them. The existing error commenting logic will take it from there.

I also cleaned up some of the areas where we were mistakenly wrapping a nil error.

For: https://issues.redhat.com/browse/DPTP-3311

### DIFF
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -9,6 +9,7 @@ import (
 	"github.com/mattn/go-zglob"
 	"github.com/sirupsen/logrus"
 
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/plugins"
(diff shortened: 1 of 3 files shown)


pkg/config/release.go
```diff
@@ -121,31 +121,31 @@ func NewLocalJobSpec(path string) (*pjdwapi.JobSpec, error) {
 // GetAllConfigs loads all configuration from the working copy of the release repo (usually openshift/release).
 // When an error occurs during some config loading, the error is not propagated, but the returned struct field will
 // have a nil value in the appropriate field. The error is only logged.
```
bbguimaraes, line 123: This documentation will have to be updated, but any idea why this was done this way?

The [pull request](https://github.com/openshift/ci-operator-prowgen/pull/99) linked in the Jira issue mentions this was an explicit decision, will we change existing behavior by returning earlier like this?

smg247: Ah I missed that somehow. I can retain the behavior of still returning the object with potentially nil fields. Other usages can just log the error. At least, for `pj-rehearse` we are better off failing entirely in this case.

smg247: I would like to clean this logic up a bit here. Essentially, the only difference is we will now fail if the master version of the `ci-operator` config fails to load, when previously we wouldn't have. I am going to search through the logs to see how often something like this happens, but I doubt this is a serious concern.

smg247: I checked the logs for the last 2 weeks, and it appears that this only fails to load the candidate configuration. We never encountered a failure to load the master configuration. I think it is worth taking the risk that cleaning this up in this manner presents.

PR #3327, opened 2023-03-16
oo-testing: opt out building index image

hongkailiu:
Requested by @qbarrand via [slack](https://redhat-internal.slack.com/archives/CBN38N3MW/p1678800245341399?thread_ts=1678118334.237179&cid=CBN38N3MW).

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1855,6 +1855,9 @@ type Bundle struct {
 	// UpdateGraph defines the update mode to use when adding the bundle to the base index.
 	// Can be: semver (default), semver-skippatch, or replaces
 	UpdateGraph IndexUpdate `json:"update_graph,omitempty"`
+	// Skip building the index image for this bundle. Default to false.
+	// This field works only for named bundles, i.e., "as" is not empty.
+	SkipBuildingIndex bool `json:"skip_building_index,omitempty"`
 }
 
 // IndexGeneratorStepConfiguration describes a step that creates an index database and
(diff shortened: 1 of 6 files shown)


pkg/api/types.go
```diff
@@ -1855,6 +1855,9 @@ type Bundle struct {
 	// UpdateGraph defines the update mode to use when adding the bundle to the base index.
 	// Can be: semver (default), semver-skippatch, or replaces
 	UpdateGraph IndexUpdate `json:"update_graph,omitempty"`
+	// Skip building the index image for this bundle. Default to false.
+	// This field works only for the name bundles, i.e., "as" is not empty.
```
smg247, line 1859:
```suggestion
	// This field works only for named bundles, i.e., "as" is not empty.
```

pkg/validation/config.go
```diff
@@ -298,6 +298,9 @@ func validateOperator(ctx *configContext, input *api.OperatorStepConfiguration,
 		if bundle.As == "" && bundle.BaseIndex != "" {
 			validationErrors = append(validationErrors, ctxN.AddField("base_index").errorf("base_index requires as to be set"))
```
smg247, line 299:
```suggestion
			validationErrors = append(validationErrors, ctxN.AddField("base_index").errorf("base_index requires 'as' to be set"))
```

pkg/validation/config.go
```diff
@@ -298,6 +298,9 @@ func validateOperator(ctx *configContext, input *api.OperatorStepConfiguration,
...
 		}
+		if bundle.As == "" && bundle.SkipBuildingIndex {
+			validationErrors = append(validationErrors, ctxN.AddField("skip_building_index").errorf("skip_building_index requires as to be set"))
```
smg247, line 302:
```suggestion
			validationErrors = append(validationErrors, ctxN.AddField("skip_building_index").errorf("skip_building_index requires 'as' to be set"))
```

since 'as' is such a common word, it reads a little funny if we don't add quotes to signify it is a field name

PR #3328, opened 2023-03-16
Remove useless environmentOverride

hongkailiu:
The test seems fine without them.


bbguimaraes:
The test will work but not go through the same code path.  These values are used here:
- https://github.com/openshift/ci-tools/blob/d6680d8ae1d1afbd1067e06b86b4ba2f30f90c51/pkg/defaults/defaults.go#L197
- https://github.com/openshift/ci-tools/blob/d6680d8ae1d1afbd1067e06b86b4ba2f30f90c51/pkg/defaults/defaults.go#L285

hongkailiu:
Thanks for the tips.

PR #3336, opened 2023-03-21
SPLAT-728: added AWS local zones profile

mtulio:
https://issues.redhat.com/browse/SPLAT-728

Added a dedicated Local Zones profile to run periodically to validate that feature.

The target account will be the one fully managed by SPLAT team.

PR #3339, opened 2023-03-22
SPLAT-728: fix/adding dedicated quota slice to the aws-local-zones

mtulio:
- Fix the dedicated quota slice for AWS Local Zones introduced on https://github.com/openshift/ci-tools/pull/3336 and https://github.com/openshift/release/pull/37543
- Remove the ConfigMap (not expected to be used)

Periodic that will be used in this profile: https://github.com/openshift/release/pull/36911

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1620,8 +1620,7 @@ func (p ClusterProfile) ConfigMap() string {
 		ClusterProfileGCPLoggingCRIO,
 		ClusterProfileGCPLoggingJSONFile,
 		ClusterProfileGCPLoggingJournald,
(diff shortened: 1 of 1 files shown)


mtulio:
ptal @smg247 ?

mtulio:
cc @rvanderp3

smg247:
@mtulio, it will take at least an hour post-merge for this to be available. It looks like that job was run too early.

mtulio:
> @mtulio, it will take at least an hour post-merge for this to be available. It looks like that job was run too early.

Thanks @smg247 ! Cheers!

PR #3345, opened 2023-03-27
prefer podman authentication file locations

atiratree:
we want to remove preference for docker auth files in favor of podman ones in https://github.com/openshift/oc/pull/1376. Since ci-operator depends on the cli (oc) we need to first merge the changes here before  the changes in oc can be merged. This change was announced in [4.10](https://docs.openshift.com/container-platform/4.10/release_notes/ocp-4-10-release-notes.html#ocp-4-10-oc-commands-obtain-credentials-from-podman-config) and a proper warning was shown when using oc commands that work with registries since then.

- podman ~/.docker/config.json is deprecated in favor of podman authentication file locations (default is ${XDG_RUNTIME_DIR}/containers/auth.json)
- oc registry login will try to write to this location so XDG_RUNTIME_DIR environment variable must be present and the XDG_RUNTIME_DIR directory created/accessible. Places that need to manipulate with ~/.docker/config.json for backwards compatibility reasons need to specify --to or --registry-config
- other oc commands that lookup credentials from registry authentication files will first try the podman locations and if the credentials are not found, oc will fallback and check ~/.docker/config.json

### DIFF
--- a/pkg/steps/clusterinstall/template.go
+++ b/pkg/steps/clusterinstall/template.go
@@ -143,6 +143,8 @@ objects:
         value: /tmp/artifacts
       - name: HOME
         value: /tmp/home
+      - name: XDG_RUNTIME_DIR
+        value: /tmp/home/run
       - name: IMAGE_FORMAT
         value: ${IMAGE_FORMAT}
       - name: KUBECONFIG
@@ -244,7 +246,7 @@ objects:
           done
         }
 
-        mkdir -p "${HOME}"
+        mkdir -p "${HOME}" "${XDG_RUNTIME_DIR}"
 
         # Share oc with other containers
         cp "$(command -v oc)" /tmp/shared
@@ -402,6 +404,8 @@ objects:
         value: test
       - name: HOME
         value: /tmp
+      - name: XDG_RUNTIME_DIR
+        value: /tmp/run
       - name: INSTALL_INITIAL_RELEASE
       - name: RELEASE_IMAGE_INITIAL
       command:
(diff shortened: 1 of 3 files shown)


pkg/steps/clusterinstall/template.go
```diff
@@ -433,7 +438,7 @@ objects:
 
           # mirror the release image and override the release image to point to the mirrored one
           mkdir /tmp/.docker && cp /etc/openshift-installer/pull-secret /tmp/.docker/config.json
-          oc registry login
+          oc registry login --to /tmp/.docker/config.json
```
atiratree, line 441: we need to mention the file explicitly to support the pre 4.10 jobs

jmguzik:
/cc @bbguimaraes 
Could have a look at this as well?

pkg/steps/clusterinstall/template.go
```diff
@@ -143,6 +143,8 @@ objects:
         value: /tmp/artifacts
       - name: HOME
         value: /tmp/home
+      - name: XDG_RUNTIME_DIR
+        value: /tmp/home/run
```
bbguimaraes, line 147: I don't see any `oc registry login` calls in either the `test` or `teardown` containers, are they done by `openshift-tests` / the installer?  Will those calls also need a `--to` argument?

bbguimaraes (commented):
LGTM other than a single question.  Cancel the hold whenever you like.

/lgtm
/hold

pkg/steps/clusterinstall/template.go
atiratree: @bbguimaraes These variables are not needed at this time. I am adding this so we that do not have to add them later in case someone needs to call the `oc registry login`, or have other functionality that depends on this common directory. We may also decide that we want to change/remove the `.docker/config.json` preferences in the code. Although I am not sure how likely that is. I can remove the support for `XDG_RUNTIME_DIR` to slim down the PR if you prefer.

atiratree: ok, thanks. Next one to go is this one: https://github.com/openshift/release/pull/37726

PR #3351, opened 2023-03-30
Update tide-config-manager to be in sync with current branching process

jmguzik:


cmd/branchingconfigmanagers/tide-config-manager/main.go
```diff
@@ -112,133 +111,63 @@ func newSharedDataDelegate() *sharedDataDelegate {
...
-	branches := sets.NewString(q.IncludedBranches...)
-	if branches.Intersection(ffe.mainMaster).Len() == 0 {
-		return
-	}
-	if requiredLabels.Intersection(ffe.excludedLabels).Len() > 0 {
-		return
-	}
-	requiredLabels = requiredLabels.Union(ffe.validBug)
-	q.Labels = requiredLabels.List()
+func (bde branchingDayEvent) GetDataFromProwConfig(pc *prowconfig.ProwConfig) {
```
danilo-gemoli, line 141: It seems that there is no more implementations of `GetDataFromProwConfig`, we could delete this method from `ShardProwConfigFunctors` then. 
Tip: I understand you like [functors](ShardProwConfigFunctors) but it would be nice to rename `ShardProwConfigFunctors` with a name closer to our [business domain](https://en.wikipedia.org/wiki/Domain-driven_design) (branch cuts, in this case) rather than the technical one: it would improve the readability.

danilo-gemoli (changes requested):

danilo-gemoli:
Both integration tests and golang docs are missing. I won't block this PR because of those but it would be a nice to have really.

cmd/branchingconfigmanagers/tide-config-manager/main.go
jmguzik: there is no more implementation in 2 out of 3 implementations. So, we still need that method.

jmguzik:
> Both integration tests and golang docs are missing. I won't block this PR because of those but it would be a nice to have really.

I do not see the reason for adding integration tests. As I do not see the use-case that we will use the functions outside of  main.go, I also vote for not adding godoc description to explanatory names.

PR #3353, opened 2023-03-30
promotionreconciler: ignore imagestreams

hongkailiu:
We found image jobs are triggered recently in openshift-priv without the counter job from openshift.
It could be `promotionreconciler` that triggers it.
We want it to ignore those cases.


droslean (approved):

hongkailiu:
> We want it to ignore those cases.

Giving another thought on this, I am not sure about it any more.
We could still make the flag from this PR available which might be useful for other cases.

PR #3354, opened 2023-03-30
`ci-operator`: allow the option to set a target-additional-suffix for aggregate tests

smg247:
When attempting to run every iteration of an aggregate payload job in a single namespace, we are seeing collisions when creating secrets for tests coming from: https://github.com/openshift/ci-tools/blob/66c77caff1dae103c14862fab46540d96775bfb5/pkg/steps/multi_stage/init.go#L30-L31

We need the ability to append an index to the end of the test's `As` field. This will allow us to differentiate the individual runs within the namespace.

There will be a follow up to this to make the `prpqr_resolver` utilize this option.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -414,6 +414,8 @@ type options struct {
 
 	multiStageParamOverrides stringSlice
 	dependencyOverrides      stringSlice
+
+	targetAdditionalSuffix string
 }
 
 func bindOptions(flag *flag.FlagSet) *options {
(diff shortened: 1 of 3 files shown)


cmd/ci-operator/main.go
```diff
@@ -712,6 +718,27 @@ func parseKeyValParams(input []string, paramType string) (map[string]string, err
...
 }
 
+func appendAdditionalSuffixToTarget(o *options) {
+	if o.targetAdditionalSuffix == "" {
+		return
+	}
+
+	for i, test := range o.configSpec.Tests {
+		for j, target := range o.targets.values {
+			if test.As == target {
```
smg247, line 728: It is unclear to me whether `o.configSpec.Tests` will contain tests that aren't one of the `targets` so this check is just for safety. If someone knows it will not, I can remove the double loop and the check, but it is low cost anyways.

jmguzik: @bbguimaraes please confirm, we will merge anyway as we want to test as soon as possible.

PR #3358, opened 2023-03-31
Add new Power VS leases to control zoning

jaypoulz:
Added two new leases for Power VS. Once they're present in openshift/release, we should be able to proceed with updating our jobs to use these, and then follow up with the removal of the old lease via https://github.com/openshift/ci-tools/pull/3352.


bear-redhat:
/hold
Feel free to cancel the hold once the leases are added

PR #3317, opened 2023-03-07
Skip disruption aggregation for azure jobs for now

xueqzhan:
[TRT-889](https://issues.redhat.com/browse/TRT-889)

PR #3361, opened 2023-04-03
Revert "Skip disruption aggregation for azure jobs for now"

neisw:
Reverts openshift/ci-tools#3317

We are past the point of needing to allow azure disruption to pass through unchecked.  Reverting.


dgoodwin:
/lgtm
/hold 

/payload 4.14 nightly blocking

Feel free to cancel hold if the jobs look healthy.

dgoodwin:
Oops can't /payload here I forgot :)

PR #3363, opened 2023-04-03
`ci-operator`: trim target-additional-suffix for input resolution

smg247:
As a follow up to https://github.com/openshift/ci-tools/pull/3354, we need to trim this suffix off for input resolution. Otherwise, it defeats the purpose as each aggregate run ends up in it's own namespace again. 

There are a couple of ways to go about this, but each of the others that I attempted resulted in complex and far-reaching changes. Simply trimming it off for input resolution should do the trick in a self-contained manner where we cannot accidentally break something else.

/cc @jmguzik @bbguimaraes 

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -1025,7 +1025,23 @@ func (o *options) resolveInputs(steps []api.Step) error {
 	}
 
 	// a change in the config for the build changes the output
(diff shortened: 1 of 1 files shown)


jmguzik (approved):
I am blindly lgtming it to enable testing as I do not have time for careful review.
/lgtm

PR #3371, opened 2023-04-09
Avoid config-files for feature branches

hongkailiu:
https://issues.redhat.com/browse/DPTP-3395

### DIFF
--- a/cmd/ci-operator-checkconfig/main.go
+++ b/cmd/ci-operator-checkconfig/main.go
@@ -5,7 +5,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"path/filepath"
 	"runtime"
 	"strings"
 
@@ -22,11 +21,11 @@ import (
 	"github.com/openshift/ci-tools/pkg/validation"
 )
 
-type tagSet map[api.ImageStreamTagReference][]*config.Info
+type tagSet map[api.ImageStreamTagReference][]*api.Metadata
 
 type promotedTag struct {
 	tag      api.ImageStreamTagReference
-	repoInfo *config.Info
+	metadata *api.Metadata
 }
 
 type options struct {
@@ -70,25 +69,13 @@ func (o *options) parse() error {
 }
 
 func (o *options) validate() (ret []error) {
-	type workItem struct {
-		configuration *api.ReleaseBuildConfiguration
-		repoInfo      *config.Info
-	}
-	inputCh := make(chan workItem)
+	inputCh := make(chan api.ReleaseBuildConfiguration)
 	produce := func() error {
 		defer close(inputCh)
 		for _, v := range o.ciOPConfigAgent.GetAll() {
 			for _, configs := range v {
 				for _, c := range configs {
-					c := c
-					configFilePath := filepath.Join(o.ConfigDir, c.Metadata.RelativePath())
-					configSpecDir := filepath.Dir(configFilePath)
-					inputCh <- workItem{configuration: &c, repoInfo: &config.Info{
-						Metadata: c.Metadata,
-						Filename: configFilePath,
-						OrgPath:  filepath.Dir(configSpecDir),
-						RepoPath: configSpecDir,
-					}}
+					inputCh <- c
 				}
 			}
 		}
@@ -98,9 +85,9 @@ func (o *options) validate() (ret []error) {
 	errCh := make(chan error)
 	map_ := func() error {
 		validator := validation.NewValidator()
-		for item := range inputCh {
-			if err := o.validateConfiguration(&validator, outputCh, item.configuration, item.repoInfo); err != nil {
-				errCh <- fmt.Errorf("failed to validate configuration %s: %w", item.repoInfo.Filename, err)
+		for c := range inputCh {
+			if err := o.validateConfiguration(&validator, outputCh, c); err != nil {
+				errCh <- fmt.Errorf("failed to validate configuration %s: %w", c.Metadata.RelativePath(), err)
 			}
 		}
 		return nil
@@ -108,7 +95,7 @@ func (o *options) validate() (ret []error) {
 	seen := tagSet{}
 	reduce := func() error {
 		for i := range outputCh {
-			seen[i.tag] = append(seen[i.tag], i.repoInfo)
+			seen[i.tag] = append(seen[i.tag], i.metadata)
 		}
 		return nil
 	}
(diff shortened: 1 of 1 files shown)


hongkailiu:
Follow up https://github.com/openshift/release/pull/38094#issuecomment-1501005713

/cc @bbguimaraes

hongkailiu:
```
ci-operator-checkconfig --config-dir ./ci-operator/config/ --registry ./ci-operator/step-registry/
INFO[0001] Configs reloaded                              duration=908.278644ms
ERRO[0004]                                               error="[failed to validate configuration ci-operator/config/ViaQ/vector/ViaQ-vector-v0.21-rh-release-5.6.yaml: found more than one matching config for branch v0.21-rh-release-5.6 on repo ViaQ/vector, failed to validate configuration ci-operator/config/redhat-developer/helm/redhat-developer-helm-release-next-playground.yaml: found more than one matching config for branch release-next-playground on repo redhat-developer/helm]"
FATA[0004] error validating configuration files         
```

What a nice surprise!

cmd/ci-operator-checkconfig/main.go
```diff
@@ -113,6 +118,15 @@ func (o *options) loadResolver(path string) error {
 	return nil
 }
 
+func (o *options) createCIOpConfigAgent() error {
+	ciOPConfigAgent, err := agents.NewConfigAgent(o.ConfigDir)
```
bbguimaraes, line 122: We're now loading configuration files twice.  This is by far the largest factor in execution time, which can be seen in the doubling caused by these changes.

Can we use `GetAll` in `validate` to avoid this?

https://github.com/openshift/ci-tools/blob/d79b51ab614ceb9c1a4f487b91d45f62b7457b9a/pkg/load/agents/configAgent.go#L161

bbguimaraes (changes requested):
This is very nice, but there's an implementation problem described below.

cmd/ci-operator-checkconfig/main.go
hongkailiu: Good point. My bash script wont suffer from this, for sure. :) 
Let me give it a try today.

bbguimaraes: `GetAll` gives you the configurations once they are loaded.  The agent loads them when it is created, so any errors will already have been reported.

`GetAll` would replace the `produce` part of the pipeline: you'd remove `OperateOnCIOperatorConfigDir` (removing the duplicated loading of the files) and instead send the values returned by the agent to `inputCh`.

This will remove the parallelism we have between the producer and the mappers, but it shouldn't be too bad (we can measure to be certain).

hongkailiu: Let me give another try tomorrow.

cmd/ci-operator-checkconfig/main.go
```diff
@@ -68,11 +74,20 @@ func (o *options) validate() (ret []error) {
...
 		defer close(inputCh)
-		if err := o.OperateOnCIOperatorConfigDir(o.ConfigDir, func(configuration *api.ReleaseBuildConfiguration, repoInfo *config.Info) error {
-			inputCh <- workItem{configuration, repoInfo}
-			return nil
-		}); err != nil {
-			return fmt.Errorf("error reading configuration files: %w", err)
+		for _, v := range o.ciOPConfigAgent.GetAll() {
+			for _, configs := range v {
+				for _, c := range configs {
+					c := c
```
hongkailiu, line 80: Oh boy!
I spent a whole afternoon on adding this line.

cmd/ci-operator-checkconfig/main.go
```diff
@@ -51,6 +54,9 @@ func (o *options) parse() error {
 	if err := o.loadResolver(registryDir); err != nil {
 		return fmt.Errorf("failed to load registry: %w", err)
 	}
+	if err := o.createCIOpConfigAgent(); err != nil {
+		return fmt.Errorf("failed to create CI Op config agent: %w", err)
+	}
```
bbguimaraes, line 59: We could just inline that function, it's more trouble to jump to its definition than to read this:

```suggestion
	if agent, err := agents.NewConfigAgent(o.ConfigDir); err != nil {
		return fmt.Errorf("failed to create CI Op config agent: %w", err)
	} else {
		o.ciOPConfigAgent = agent
	}
```

cmd/ci-operator-checkconfig/main.go
bbguimaraes: Don't you love how _simple_ Go is?

The `lint` program has some warnings for this problem, I'm guessing it does not recognize this particular case.

cmd/ci-operator-checkconfig/main.go
```diff
@@ -68,11 +74,20 @@ func (o *options) validate() (ret []error) {
...
+				for _, c := range configs {
+					c := c
+					configFilePath := filepath.Join(o.ConfigDir, c.Metadata.RelativePath())
+					configSpecDir := filepath.Dir(configFilePath)
+					inputCh <- workItem{configuration: &c, repoInfo: &config.Info{
+						Metadata: c.Metadata,
+						Filename: configFilePath,
+						OrgPath:  filepath.Dir(configSpecDir),
+						RepoPath: configSpecDir,
+					}}
```
bbguimaraes, line 88: This is not part of this pull request, but since we now have all this extra code…  `repoInfo` is only used by `validateTags`, and only the `Org`/`Repo`/`Branch`/`Variant` members are used there.  We can replace it by `api.Metadata` and avoid these path calculations.

cmd/ci-operator-checkconfig/main.go
hongkailiu: Where did you see the warning?

I cannot find with my local run:
```
➜  ci-tools git:(dbf3b6d40) golangci-lint version                                
golangci-lint has version 1.52.2 built with go1.20.2 from da04413a on 2023-03-25T18:11:28Z
➜  ci-tools git:(dbf3b6d40) golangci-lint run ./cmd/ci-operator-checkconfig/...  
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter. Replaced by unused. 

➜  ci-tools git:(dbf3b6d40) golint ./cmd/ci-operator-checkconfig/... 
cmd/ci-operator-checkconfig/main.go:96:2: don't use underscores in Go names; var map_ should be map
```
 or [ci run](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3371/pull-ci-openshift-ci-tools-master-lint/1664265385467187200/artifacts/test/build-log.txt).

hongkailiu: The line is removed because we are passing configuration to the channel, instead of its pointer.

bbguimaraes: I meant it has warnings for this type of problem — reuse of the address of a loop variable — e.g. it will warn if you start a Go routine which references it.  But it doesn't detect this case.

PR #3379, opened 2023-04-17
`pj-rehearse` use the synchronize event to auto refresh job list

smg247:
There is actually an event (synchronize) for a new push to a PR. We can utilize it to perform the refreshing of the affected job list automatically when a PR is pushed to. This will also remove the `rehearsals-ack` label if it is present and there is a new push (mimicking the `lgtm` functionality). Additionally, it will now delete any older job list comments so there is always only one on a PR.

There will be a follow up PR to remove the info about `/pj-rehearse refresh` from the docs, and I will announce this functionality change prior to merging.

For: [DPTP-3393](https://issues.redhat.com/browse/DPTP-3393)

### DIFF
--- a/cmd/pj-rehearse/server.go
+++ b/cmd/pj-rehearse/server.go
@@ -23,6 +23,7 @@ import (
 )
 
 const (
+	rehearsalNotifier  = "[REHEARSALNOTIFIER]"
 	pjRehearse         = "pj-rehearse"
 	needsOkToTestLabel = "needs-ok-to-test"
 	rehearseNormal     = "/pj-rehearse"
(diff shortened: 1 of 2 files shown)


cmd/pj-rehearse/server.go
```diff
@@ -201,6 +197,68 @@ func (s *server) respondToNewPR(pullRequest *github.PullRequest, logger *logrus.
...
+			logger.WithError(err).Error("failed to get pull request")
+			return
+		}
+
+		comments, err := s.ghc.ListIssueComments(org, repo, number)
+		if err != nil {
+			// This also shouldn't happen, but if it does just log and continue it doesn't affect the rest of the process
+			logger.WithError(err).Error("failed to get comments for pull request")
+		}
+		if len(comments) > 0 {
```
droslean, line 223: nit: This check is not necessary since you are doing the iteration below.

droslean (approved):

PR #3382, opened 2023-04-18
`ci-operator`: trim target-additional-suffix from profile secret name

smg247:
The additional suffix is currently part of the cluster profile secret name when present. This secret needs to be shared among all iterations of the test, and should not be duplicated. Therefore, the suffix should be trimmed off to resolve the secret. This will prevent errors like the following when using the `target-additional-suffix` param:
```
could not run steps: step e2e-aws-sdn-serial-1 failed: could not get cluster profile secret "e2e-aws-sdn-serial-1-cluster-profile": secrets "e2e-aws-sdn-serial-1-cluster-profile" not found
```

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -76,6 +76,7 @@ func FromConfig(
 	consoleHost string,
 	nodeName string,
 	nodeArchitectures []string,
+	targetAdditionalSuffix string,
 ) ([]api.Step, []api.Step, error) {
 	crclient, err := ctrlruntimeclient.NewWithWatch(clusterConfig, ctrlruntimeclient.Options{})
 	crclient = secretrecordingclient.Wrap(crclient, censor)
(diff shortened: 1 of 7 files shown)


pkg/defaults/defaults.go
```diff
@@ -112,7 +113,7 @@ func FromConfig(
 	httpClient := retryablehttp.NewClient()
 	httpClient.Logger = nil
 
-	return fromConfig(ctx, config, graphConf, jobSpec, templates, paramFile, promote, client, buildClient, templateClient, podClient, leaseClient, hiveClient, httpClient.StandardClient(), requiredTargets, cloneAuthConfig, pullSecret, pushSecret, api.NewDeferredParameters(nil), censor, consoleHost, nodeName)
+	return fromConfig(ctx, config, graphConf, jobSpec, templates, paramFile, promote, client, buildClient, templateClient, podClient, leaseClient, hiveClient, httpClient.StandardClient(), requiredTargets, cloneAuthConfig, pullSecret, pushSecret, api.NewDeferredParameters(nil), censor, consoleHost, nodeName, targetAdditionalSuffix)
```
danilo-gemoli, line 116: At this stage adding another parameter doesn't make any difference to this, already totally unreadable, function signature. It's something we have to refactor, not in this PR of course, but we have to.

smg247: yeah, agreed there is too much going on here...

jmguzik: unfortunately this is a common thing in our infra. we can change it as a follow-up.

PR #3388, opened 2023-04-21
Override the build labels only with non-empty values

hongkailiu:
https://issues.redhat.com/browse/OCPBUGS-3785

We found the cases where `labels("io.openshift.build.commit.id")=""`.
So we stop doing it by this PR.

We could also try to stop doing it in general for the labels that build-api supports.
But let us do the baby steps here.

@apoorvajagtap @bbguimaraes


hongkailiu:
/hold

Need more testing.

PR #3390, opened 2023-04-21
pkg/steps: add build pending timeout period

bbguimaraes:
https://github.com/openshift/ci-tools/pull/2875

Based on https://github.com/openshift/ci-tools/pull/3385.

/hold


pkg/kubernetes/client.go
```diff
@@ -71,7 +71,7 @@ func WaitForConditionOnObject(ctx context.Context, client ctrlruntimeclient.With
 
 type PodClient interface {
 	loggingclient.LoggingClient
-	PendingTimeout() time.Duration
+	GetPendingTimeout() time.Duration
```
danilo-gemoli, line 74: If I've understood it correctly this gets the value from `--pod-pending-timeout` and it controls a pod lifecycle regardless its nature: being it a test pod, a build, etc. Is it enough? Do we need it to be more fine-grained (I mean, having more than one value)?

danilo-gemoli:
Other than the scary commits history, which could be now simplified as #3385 has been merged, it looks fine overall.

pkg/kubernetes/client.go
bbguimaraes: Yes, that is correct, this is the same period as is used for pods in general.

The causes for scheduling delays are for the most part the same for all types of pods, so I don't see the need to add a separate value for builds.

bbguimaraes:
Rebased.

/hold cancel

bbguimaraes:
> ```
> INFO[2023-04-24T15:04:47Z] pod pending for more than 30s: container "ci-scheduling-dns-wait" has not started in 1m58.8714557s: 
> …
> ```

It may also be good to change the error message to include all pending containers, it's not immediately obvious that the scheduling requirements of one of the (non-init.) containers is the cause of the problem.

PR #3391, opened 2023-04-21
pkg/steps: add error reason for pending pods

bbguimaraes:
https://github.com/openshift/ci-tools/pull/2876


danilo-gemoli:
Why do I feel like I have reviewed (and merged) a PR that was exactly the same?

bbguimaraes:
There has been a previous attempt to make this change (see the description above), maybe that's what you are thinking of?

It was prior to the changes in the build/pod watching code, so this is a revised version.

/hold cancel

PR #3393, opened 2023-04-24
Change prowgen param name from api to remote_api

jmguzik:


jmguzik:
It was requested by the QE to have always as: first in config.

droslean (approved):

PR #3397, opened 2023-04-25
trt-945: support MasterNodesUpdated indicator

neisw:
Includes MasterNodesUpdated flag in BackendDisruptionRow uploaded to BigQuery to differentiate between jobs that triggered node updates and those that don't.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
@@ -5,6 +5,8 @@ import (
 	"math"
 	"strings"
 
+	"github.com/sirupsen/logrus"
+
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
@@ -49,12 +51,15 @@ type BackendDisruption struct {
 	DisruptionMessages []string
 }
 
-// defined in origin/platformidentification/types.go
-// minimized here
+// ClusterData is defined in origin/platformidentification/types.go
+// it is duplicated in a minimized form here
 type ClusterData struct {
 	MasterNodesUpdated string
 }
 
+// GetMasterNodesUpdatedStatusFromClusterData takes multiple file contents as a copy of the ClusterData
+// file is created for multiple test phases (upgrade / conformance) in the same manor that multiple disruption
+// files are created for the multiple phases
 func GetMasterNodesUpdatedStatusFromClusterData(clusterData map[string]string) string {
 	// default is unknown
 	masterNodesUpdated := ""
@@ -68,16 +73,17 @@ func GetMasterNodesUpdatedStatusFromClusterData(clusterData map[string]string) s
 
 		cd := &ClusterData{}
 		if err := json.Unmarshal([]byte(clusterdataResults), cd); err != nil {
+			logrus.WithError(err).Error("error unmarshalling clusterdataJson")
 			continue
 		}
 
-		// if the value is y then set it and quit scanning
+		// if the value is y then return it
+		// as it supersedes all other values
 		if strings.ToUpper(cd.MasterNodesUpdated) == "Y" {
-			masterNodesUpdated = cd.MasterNodesUpdated
-			break
+			return cd.MasterNodesUpdated
 		}
 
-		// if we don't have a value yet use whatever value we have
+		// if we don't have a value yet use whatever value we have coming in
 		if len(masterNodesUpdated) == 0 {
 			masterNodesUpdated = cd.MasterNodesUpdated
 			continue
@@ -96,6 +102,7 @@ func GetServerAvailabilityResultsFromDirectData(backendDisruptionData map[string
 		}
 		allDisruptions := &BackendDisruptionList{}
 		if err := json.Unmarshal([]byte(disruptionJSON), allDisruptions); err != nil {
+			logrus.WithError(err).Error("error unmarshalling disruptionJson")
 			continue
 		}
 
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
@@ -3,6 +3,7 @@ package jobrunaggregatorlib
 import (
 	"encoding/json"
 	"math"
+	"strings"
 
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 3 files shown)


neisw:
/hold

Need to update BigQuery schema before merging / after review.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -48,6 +49,44 @@ type BackendDisruption struct {
 	DisruptionMessages []string
 }
 
+// defined in origin/platformidentification/types.go
```
dgoodwin, line 52: just a nit but godoc convention would be to have the name of the thing at the start of the comment,  "ClusterData defined in..."

https://go.dev/blog/godoc

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -48,6 +49,44 @@ type BackendDisruption struct {
...
+
+	// there can be multiple files (upgrade / conformance) if any of them indicate the master nodes updated
+	// we indicate that for the entire run
+	for _, clusterdataResults := range clusterData {
+		if len(clusterdataResults) == 0 {
+			continue
+		}
+
+		cd := &ClusterData{}
+		if err := json.Unmarshal([]byte(clusterdataResults), cd); err != nil {
```
dgoodwin, line 70: Should log the error here.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -48,6 +49,44 @@ type BackendDisruption struct {
 	DisruptionMessages []string
 }
 
+// defined in origin/platformidentification/types.go
+// minimized here
+type ClusterData struct {
+	MasterNodesUpdated string
+}
+
+func GetMasterNodesUpdatedStatusFromClusterData(clusterData map[string]string) string {
```
dgoodwin, line 58: Can you godoc the a little here, i'm confused about why clusterData is a map of string to string. I'd expect we were working with a single cluster data file per job run here. Is this because GetOpenShiftTestsFilesWithPrefix was built to potentially return multiple files, but in this case we'd expect 1 or 0?

pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
```diff
@@ -163,17 +163,21 @@ func (o *disruptionUploader) uploadContent(ctx context.Context, jobRun jobrunagg
 		return nil
 	}
 
-	return o.uploadBackendDisruptionFromDirectData(ctx, jobRun.GetJobRunID(), backendDisruptionData, logger)
+	clusterData, _ := jobRun.GetOpenShiftTestsFilesWithPrefix(ctx, "cluster-data")
```
dgoodwin, line 166: Looks like another error that should be logged.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -48,6 +49,44 @@ type BackendDisruption struct {
 	DisruptionMessages []string
 }
 
+// defined in origin/platformidentification/types.go
+// minimized here
+type ClusterData struct {
+	MasterNodesUpdated string
```
dgoodwin, line 55: Wondering about a bool here, but the Y/N string does give us tri-state: yes, no, we have no idea.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
```diff
@@ -48,6 +49,44 @@ type BackendDisruption struct {
...
+
+		// if the value is y then set it and quit scanning
+		if strings.ToUpper(cd.MasterNodesUpdated) == "Y" {
+			masterNodesUpdated = cd.MasterNodesUpdated
+			break
+		}
+
+		// if we don't have a value yet use whatever value we have
+		if len(masterNodesUpdated) == 0 {
+			masterNodesUpdated = cd.MasterNodesUpdated
```
dgoodwin, line 82: Also got tripped up here, we start with empty value, but if we see an empty value we set our tracking var to empty value, which it either (a) already was, or (b) was N somehow and we just overwrote it? 

I almost think it would be better to handle multiple cluster data files explicitly (as that seems like a bug if we see it?), and then replace the loop with a more simple copy or leave undefined.

dgoodwin (commented):
This would have been an instant lgtm if it was a 2000 line diff. :)

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
neisw: So, I can but this is basically copy and paste from [right below](https://github.com/openshift/ci-tools/blob/6aa730873a19888644b4467373ec4ee13759f4d5/pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go#L98-L100 ) .  I could update both places if we want.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
neisw: Added, I also had a comment within the function covering the case.  It is because like disruption data a copy is created for both the upgrade and conformance portions of the job runs.  We can reevaluate that if we want.

pkg/jobrunaggregator/jobrunbigqueryloader/disruption.go
neisw: Added, don't want to fail the uploading of disruption though so it will proceed even on error.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
neisw: Yes so all the prior data will be empty (don't want to force it to false or true) and in the case that we can't determine the default is empty since we have no idea as you mentioned.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
neisw: This says:

if localVariable is empty, set it to the incomingVariable

Whatever is coming in will supersede our empty variable (even if it is empty itself).  The logic above it returns on the first 'Y' as any 'Y' value pertains to the entire run in the same way data from multiple disruption data files get collapsed into a single row.

The multiple files come from existing commands / code within origin like the disruption handling.  If we want to rework the logic that generates the actual cluster-data file we could could check for an existing file, merge any incoming results with the existing results and then write one file.  This was just built using the existing patterns.

pkg/jobrunaggregator/jobrunaggregatorlib/disruption.go
dgoodwin: Aha I see now, thanks!

PR #3399, opened 2023-04-26
one build for aggregate jobs (V7)

smg247:
With the additional-suffix changes, it is likely that we don't need to actually add the index to the job name itself. I confirmed this theory by manually triggering some jobs with the same name in the same namespace. They all succeeded as the pods were named differently due to the additional suffix on the target. Let's try this out for real now. If this works, there will be no need to change the aggregator itself.

/cc @jmguzik


smg247:
This one still needs work, but I wanted to get it up to illustrate the idea

PR #3416, opened 2023-05-05
`pr-reminder`: Replace --github-mapping-config-path

danilo-gemoli:
Following up what has been done in #3413 
Periodic definition [periodics.yaml#L149-L197](https://github.com/openshift/release/blob/2403901f72ad6a2aa3273edff3ca51d321c70221/ci-operator/jobs/infra-periodics.yaml#L149-L197) needs to be changed accordingly.

/cc @smg247 @hongkailiu


cmd/pr-reminder/main.go
```diff
@@ -283,10 +286,16 @@ func main() {
 		logrus.WithError(err).Fatal("failed to load config")
 	}
 
-	var gtk githubToKerberos
-	if err = loadConfig(o.githubMappingConfig, &gtk); err != nil {
+	roverUsers := []*rover.User{}
+	if err = loadConfig(o.githubUsers, &roverUsers); err != nil {
 		logrus.WithError(err).Fatal("failed to load rover groups config")
 	}
+	var gtk githubToKerberos = rover2github.MapGithubToKerberos(roverUsers,
```
hongkailiu, line 293: I would define the function in https://github.com/openshift/ci-tools/blob/master/pkg/rover/types.go
and avoid the new package.

The conflict function as an arg looks great but I would just push it into the `MapGithubToKerberos`.
I cannot see other usage as now. I could do it when we have another use case.

We can reuse the same function [here](https://github.com/openshift/ci-tools/blob/master/cmd/github-ldap-user-group-creator/main.go#L230) as well.

danilo-gemoli: Changed as you requested and just opened also [#3418](https://github.com/openshift/ci-tools/pull/3418)

hongkailiu (approved):

PR #3417, opened 2023-05-05
Fix issue DPTP-3440 Add server-side apply support

bear-redhat:
This PR introduce a flag to enable `--server-side` apply globally.
This PR also introduce a feature to use server-side apply for config files with file name starts with `SS_`.

Related to an HyperShift bug generating a large CRD which cannot be applied on client-side.
Ref: https://issues.redhat.com/browse/HOSTEDCP-985

cc. @hongkailiu

### DIFF
--- a/cmd/applyconfig/README.md
+++ b/cmd/applyconfig/README.md
@@ -20,6 +20,7 @@ that, it adds several lightweight features required by Test Platform team, its w
    resource that would be placed in that namespace are not yet present in the cluster (in a normal dry run, applying the
    resource to a namespace that does not exist in the cluster would fail). It does so by bookkeeping which namespaces
    would be applied if not in dry run, and creating them temporarily if they do not exist.
+7. Add `--server-side` to the `apply` command if a filename starts with `SS_`
 
 ## Why it exists
 
--- a/cmd/applyconfig/applyconfig.go
+++ b/cmd/applyconfig/applyconfig.go
@@ -31,6 +31,7 @@ import (
 
 type command string
 type dryRunMethod string
+type applyMethod string
 
 type options struct {
 	user        *nullableStringFlag
@@ -39,6 +40,7 @@ type options struct {
 	context     string
 	kubeConfig  string
 	dryRun      dryRunMethod
+	apply       applyMethod
 }
 
 const (
@@ -50,6 +52,9 @@ const (
 	dryAuto   dryRunMethod = "auto"
 	dryServer dryRunMethod = "server"
 	dryClient dryRunMethod = "client"
+
+	applyServer applyMethod = "server"
+	applyClient applyMethod = "client"
 )
 
 const defaultAdminUser = "system:admin"
@@ -73,7 +78,7 @@ func (n *nullableStringFlag) Set(val string) error {
 
 func gatherOptions() *options {
 	// nonempty dryRun is a safe default (empty means not a dry run)
-	opt := &options{user: &nullableStringFlag{}, dryRun: dryAuto}
+	opt := &options{user: &nullableStringFlag{}, dryRun: dryAuto, apply: applyClient}
 
 	var confirm bool
 	flag.BoolVar(&confirm, "confirm", false, "Set to true to make applyconfig commit the config to the cluster")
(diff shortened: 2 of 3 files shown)


cmd/applyconfig/README.md
```diff
@@ -20,6 +20,7 @@ that, it adds several lightweight features required by Test Platform team, its w
    resource that would be placed in that namespace are not yet present in the cluster (in a normal dry run, applying the
    resource to a namespace that does not exist in the cluster would fail). It does so by bookkeeping which namespaces
    would be applied if not in dry run, and creating them temporarily if they do not exist.
+7. Server-side apply is always enabled for file with name starts with `_SS`
```
hongkailiu, line 23:
```suggestion
7. Add `--server-side` to the `apply` command if a filename starts with `server-side_`
```

The file would be ignored if its filename starts with `_`.

cmd/applyconfig/applyconfig.go
```diff
@@ -50,6 +52,9 @@ const (
 	dryAuto   dryRunMethod = "auto"
 	dryServer dryRunMethod = "server"
 	dryClient dryRunMethod = "client"
+
+	applyServer applyMethod = "server"
+	applyClient applyMethod = "client"
```
hongkailiu, line 57: Why do we need it?

cmd/applyconfig/applyconfig.go
```diff
@@ -86,6 +91,11 @@ func gatherOptions() *options {
 	var dryMethod string
 	dryRunMethods := strings.Join(validDryRunMethods, ",")
 	flag.StringVar(&dryMethod, "dry-run-method", string(opt.dryRun), fmt.Sprintf("Method to use when running when --confirm is not set to true (valid values: %s)", dryRunMethods))
+
+	var applyMethod string
+	applyMethods := strings.Join([]string{string(applyServer), string(applyClient)}, ",")
+	flag.StringVar(&applyMethod, "apply-method", string(opt.apply), fmt.Sprintf("Method to use when applying the config (valid values: %s). Server-side apply is always enabled for file with names start with '_SS'.", applyMethods))
```
hongkailiu, line 97: Why do we need it?

cmd/applyconfig/applyconfig.go
```diff
@@ -180,6 +191,19 @@ func makeOcApply(kubeConfig, context, path, user string, dry dryRunMethod) *exec
 	case dryNone:
 		// No additional args needed
 	}
+
+	fileName := filepath.Base(path)
+	if strings.HasPrefix(fileName, "_SS") {
```
hongkailiu, line 196:
```suggestion
	if strings.HasPrefix(fileName, "server-side_") {
```

cmd/applyconfig/applyconfig.go
```diff
@@ -180,6 +191,19 @@ func makeOcApply(kubeConfig, context, path, user string, dry dryRunMethod) *exec
...
+		logrus.Info("Use server-side apply for ", fileName)
+		apply = applyServer
```
hongkailiu, line 198:
```suggestion
		cmd.Args = append(cmd.Args, "--server-side=true")
```

cmd/applyconfig/applyconfig.go
```diff
@@ -162,10 +172,11 @@ type configApplier struct {
 	path       string
 	user       string
 	dry        dryRunMethod
+	apply      applyMethod
 	censor     *secrets.DynamicCensor
 }
 
-func makeOcApply(kubeConfig, context, path, user string, dry dryRunMethod) *exec.Cmd {
+func makeOcApply(kubeConfig, context, path, user string, dry dryRunMethod, apply applyMethod) *exec.Cmd {
```
hongkailiu, line 179: Why do we need the extra flag?

cmd/applyconfig/applyconfig.go
bear-redhat: In case we switch to server-side apply globally.

cmd/applyconfig/applyconfig.go
bear-redhat: In case we switch to server-side apply globally.

cmd/applyconfig/applyconfig.go
bear-redhat: In case we switch to server-side apply globally.

cmd/applyconfig/applyconfig.go
hongkailiu: discussion over slack (private channel) about my concern:
https://redhat-internal.slack.com/archives/GB7NB0CUC/p1694790437683139?thread_ts=1694777342.094659&cid=GB7NB0CUC

bear-redhat: I think its better to have it

jmguzik:
I am not fully convinced we need this and I tend towards @hongkailiu opinion. Nevertheless its our repo so if @bear-redhat  is insisting to have something, and its not harmful, I do not want to stand in way.

/lgtm

PR #3422, opened 2023-05-09
[WIP] retester: s3 backoff cache

psalajova:
For https://issues.redhat.com/browse/DPTP-3375

Adds implementation for using the AWS s3 bucket to store retester's backoff cache file.


pkg/retester/s3BackOffCache.go
```diff
@@ -0,0 +1,138 @@
...
+		Body:   bytes.NewReader(content),
+	})
+	if err != nil {
+		return fmt.Errorf("failed to upload file %s into %s bucket: %w", b.file, retesterBucket, err)
+	}
+
+	return nil
+}
+
+func (b *s3BackOffCache) check(pr tide.PullRequest, baseSha string, policy RetesterPolicy) (retestBackoffAction, string) {
```
bbguimaraes, line 107: This can use the implementation in `fileBackOffCache.go` since they both use the same `map[string]*pullRequest` structure.

pkg/retester/s3BackOffCache.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return fmt.Errorf("couldn't get %s file from aws s3 bucket %s: %w", b.file, retesterBucket, err)
+	}
+
+	content, err := io.ReadAll(result.Body)
+	if err != nil {
+		return fmt.Errorf("failed to read file %s: %w", b.file, err)
+	}
+
+	cache := map[string]*pullRequest{}
```
hongkailiu, line 66: We could make a shared function (with `fileBackOffCache `) for this too.

loadAndDelete(content byte[], cache map[string]*pullRequest)

pkg/retester/s3BackOffCache.go
```diff
@@ -0,0 +1,108 @@
...
+		return fmt.Errorf("couldn't create new aws session")
+	}
+	b.awsSession = sess
+
+	_, err = sess.Config.Credentials.Get()
+	if err != nil {
+		return fmt.Errorf("credentials for aws not found")
+	}
+
+	svc := s3.New(b.awsSession)
```
hongkailiu, line 52: Could we pass `svc` with the constructor and init `session` in the `main` function?

pkg/retester/s3BackOffCache.go
psalajova: You mean **one** func for unmarshalling the []byte content into `cache` and also deleting the old records?

pkg/retester/s3BackOffCache.go
psalajova: We can init `session` in `retester/main.go` main function yes. And do I understand correctly you mean I should add `svc` as a field to the `s3BackOffCache struct`, so we can reuse that `svc` for each s3 operation (for loading in `loadFromAwsNow` func and also for uploading in `save` func)?

hongkailiu: Yes. Your understanding is correct.

cmd/retester/main.go
```diff
@@ -68,6 +71,7 @@ func gatherOptions() options {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Dry run for testing. Uses API tokens but does not mutate.")
 	fs.BoolVar(&o.runOnce, "run-once", false, "If true, run only once then quit.")
+	fs.BoolVar(&o.cacheFileOnS3, "cache-file-on-s3", false, "If true, use aws s3 bucket to store the cache file.")
```
hongkailiu, line 74: Could we invalidate if `cacheFileOnS3` and `cache-file` are both (or neither) set?
Also add the unit tests for those new cases?

pkg/retester/retester.go
```diff
@@ -202,6 +204,13 @@ func NewController(ghClient githubClient, cfg config.Getter, gitClient git.Clien
 	return ret
 }
 
+func newBackoffCache(awsSession *session.Session, cacheFile string, cacheRecordAge time.Duration, logger *logrus.Entry) backoffCache {
```
hongkailiu, line 207: NIT: we do not need a new function here. Moving it into the caller is easier to read the code.

pkg/retester/s3BackOffCache.go
```diff
@@ -0,0 +1,103 @@
...
+	cache, err := loadAndDelete(content, b.logger, now, b.cacheRecordAge)
+	if err != nil {
+		return err
+	}
+	b.cache = cache
+	return nil
+}
+
+// loadAndDelete loads content into cache and deletes old records from cache
+func loadAndDelete(content []byte, logger *logrus.Entry, now time.Time, cacheRecordAge time.Duration) (map[string]*pullRequest, error) {
```
hongkailiu, line 67: NIT: could we move this to `fileBackOffCache.go`?

cmd/retester/main.go
hongkailiu: After answering you in slack, I gave another thought.

>  we always need it (whether cacheFileOnS3 is true or false)

`cache-file` is not set if we want to disable the cache. It is unlikely we do it in our production but i think it is a valid case.

Before this PR, `cache-file` has two functions:

1. cache is enabled _iff_ it is not empty string.
2. it provides the name of the file.

After this PR, `cache-file`'s functions more or less the same:
1. cache is enabled _iff_ it is not empty string. (same as before)
2. it provides the name of the file: either local or on s3 according to `cache-file-on-s3`

So the only invalid case that I can think of is that `cache-file` is not set but `cache-file-on-s3` is `true`.
Let us error out in this case.

hongkailiu (commented):
/lgtm

/hold

whenever you are ready to watch the rolling ...

PR #3437, opened 2023-05-17
`ci-operator`: add new UNIQUE_HASH env var which includes jobName and targetAdditionalSuffix when available

smg247:
There were reported errors with aggregate jobs like the following, found in this [log](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/openshift-origin-27894-ci-4.14-upgrade-from-stable-4.13-e2e-azure-sdn-upgrade/1658692416556765184):
```
Error: A resource with the ID "/subscriptions/72e3a972-58b0-4afc-bd4f-da89b39ccebd/resourceGroups/os4-common/providers/Microsoft.Network/dnsZones/ci2.azure.devcluster.openshift.com/CNAME/api.ci-op-iz4zs9r6-12383" already exists - to be managed via Terraform this resource needs to be imported into the State. Please see the resource documentation for "azurerm_dns_cname_record" for more information.
```
These jobs install the cluster utilizing configuration like the following, found in this [step](https://steps.ci.openshift.org/reference/ipi-conf):
```
cluster_name=${NAMESPACE}-${JOB_NAME_HASH}
```
The `JOB_NAME_HASH` is the same for each aggregated job. This is due to not being able to change the actual job name for each aggregate job as that would mess the aggregator up.

This adds a new `UNIQUE_HASH` env var that will be used in these cases and includes the `targetAdditionalSuffix`.

/cc @bbguimaraes @jmguzik

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -140,7 +140,7 @@ of dynamic parameters that are inferred from previous steps. These parameters ar
     The job name in a form safe for use as a Kubernetes resource name.
 
   JOB_NAME_HASH
-    A short hash of the job name for making tasks unique.
+    A short hash of the job name, including target-additional-suffix if supplied, for making tasks unique.
 
   RPM_REPO_<org>_<repo>
     If the job creates RPMs this will be the public URL that can be used as the
@@ -694,7 +694,7 @@ func (o *options) Complete() error {
 		return err
 	}
 
-	appendAdditionalSuffixToTarget(o)
+	handleTargetAdditionalSuffix(o)
 
 	return overrideTestStepDependencyParams(o)
 }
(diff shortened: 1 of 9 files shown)


cmd/ci-operator/main.go
```diff
@@ -718,11 +718,12 @@ func parseKeyValParams(input []string, paramType string) (map[string]string, err
 	return params, nil
 }
 
-func appendAdditionalSuffixToTarget(o *options) {
+// TODO: method should be renamed
```
jmguzik, line 721: ??

bbguimaraes:
There's also `JOB_NAME` from Prow, and we provide `JOB_NAME_SAFE`.  They're all based on the same value, it would be very surprising for these to disagree.

smg247:
> There's also `JOB_NAME` from Prow, and we provide `JOB_NAME_SAFE`. They're all based on the same value, it would be very surprising for these to disagree.

@bbguimaraes,
Do you think we should provide another variable named something like `ADDITIONAL_SUFFIX` with this value and then update all of the usages of `JOB_NAME_HASH` to use that as well?

Alternatively, we could added the suffix to `JOB_NAME_SAFE`, but there isn't much we can do about `JOB_NAME` from prow...

smg247:
I added a new `UNIQUE_HASH` env var to account for this. I will add documentation and update the existing steps to use it in a follow up.

PR #3440, opened 2023-05-18
`sprint-automation`: remove secondary triage roles

smg247:
As discussed in the retro, we no longer require these roles and they aren't providing us value. Removing them from the daily digest. I will clean up the schedules in PagerDuty post-merge.


danilo-gemoli:
/lgtm
🎉

PR #3441, opened 2023-05-22
`prowgen`: promote on `multi01`

danilo-gemoli:
`prowgen` generates one additional promotion job on `multi01` when the configuration flag `multi_arch` is set.
~~This new configuration replaces the old `additional_architectures`, therefore the following PRs have been reverted:~~
- ~~[Prowgen: multiarch postsubmit images #3053](https://github.com/openshift/ci-tools/pull/3053)~~  
- ~~[[prowgen] merge additional_architectures when repo defines them #3153](https://github.com/openshift/ci-tools/pull/3153)~~  

More details here: [DPTP-3463](https://issues.redhat.com/browse/DPTP-3463)

/cc @droslean @deepsm007

### DIFF
--- a/pkg/prowgen/prowgen.go
+++ b/pkg/prowgen/prowgen.go
@@ -102,7 +102,7 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	newJobBaseBuilder := func() *prowJobBaseBuilder {
 		return NewProwJobBaseBuilder(configSpec, info, NewCiOperatorPodSpecGenerator())
 	}
-
+	newJobBaseBuilderForPromotion := NewProwJobBaseBuilderForPromotion(configSpec, info, NewCiOperatorPodSpecGenerator())
 	imageTargets := api.ImageTargets(configSpec)
 
 	if len(imageTargets) > 0 {
@@ -116,7 +116,7 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 		presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(jobBaseGen, "images", info))
 
 		if configSpec.PromotionConfiguration != nil {
-			postsubmitsForPromotion, err := generatePostsubmitsForPromotion(newJobBaseBuilder, info, func(options *generatePostsubmitOptions) {
+			postsubmitsForPromotion, err := generatePostsubmitsForPromotion(newJobBaseBuilderForPromotion, info, func(options *generatePostsubmitOptions) {
 				options.imageTargets = imageTargets
 			})
 			if err != nil {
@@ -158,6 +158,17 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 		PostsubmitsStatic: postsubmits,
 		Periodics:         periodics,
 	}, nil
+}
+
+func NewProwJobBaseBuilderForPromotion(configSpec *cioperatorapi.ReleaseBuildConfiguration,
+	info *ProwgenInfo, podSpecGenerator CiOperatorPodSpecGenerator) func() *prowJobBaseBuilder {
+	return func() *prowJobBaseBuilder {
+		builder := NewProwJobBaseBuilder(configSpec, info, podSpecGenerator)
+		if info.Config.MultiArch {
+			builder.Cluster(api.ClusterMulti01).WithLabel(api.ClusterLabel, string(api.ClusterMulti01))
+		}
+		return builder
+	}
 }
 
 func testContainsLease(test *cioperatorapi.TestStepConfiguration) bool {
--- a/pkg/prowgen/testdata/zz_fixture_TestGeneratePostSubmitForPromotion_spawn_on_multi01_cluster.yaml
+++ b/pkg/prowgen/testdata/zz_fixture_TestGeneratePostSubmitForPromotion_spawn_on_multi01_cluster.yaml
@@ -1,0 +1,13 @@
+- agent: kubernetes
+  always_run: true
+  branches:
+  - ^branch$
+  cluster: multi01
+  decorate: true
+  decoration_config:
+    skip_cloning: true
+  labels:
+    ci-operator.openshift.io/cluster: multi01
+    ci-operator.openshift.io/is-promotion: "true"
+  max_concurrency: 1
+  name: branch-ci-organization-repository-branch-images
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -14,7 +14,6 @@ import (
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 
-	"github.com/openshift/ci-tools/pkg/api"
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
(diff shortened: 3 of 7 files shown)


deepsm007:
This PR needs additional changes in release repo as there are some configuration running with `additional_architecture as arm01`. As we are removing that in this PR we should open up 1 against release repo to fix it and merge both together.

danilo-gemoli:
@deepsm007 yes I'm going to take care of that as well

pkg/prowgen/testdata/zz_fixture_TestGeneratePostSubmitForPromotion_spawn_on_multi01_cluster.yaml
```diff
@@ -0,0 +1,24 @@
...
+    skip_cloning: true
+  labels:
+    ci-operator.openshift.io/is-promotion: "true"
+  max_concurrency: 1
+  name: branch-ci-organization-repository-branch-images
+- agent: kubernetes
+  always_run: true
+  branches:
+  - ^branch$
+  cluster: multi01
```
bbguimaraes, line 16: I don't quite understand these changes.

The previous implementation created one post-submit job per architecture.  How are we covering the same number by scheduling a single extra job on a cluster with nodes of potentially different architectures?  Is there something magical in OpenShift builds in that type of cluster which will produce multiple images?  Otherwise, won't this produce a (single) job which will build each image in an unpredictable architecture?

bbguimaraes: I see, so we're keeping the replication based on node architectures.  Then that makes sense.

droslean:
/hold

We shouldn't deprecate the previous implementation in the same PR. I was expecting a new multi-arch implementation that points the only poststubmit to the `multi01` cluster.

pkg/prowgen/prowgen.go
```diff
@@ -236,6 +236,19 @@ func generatePostsubmitsForPromotion(jobBaseBuilderFactory func() *prowJobBaseBu
 	for _, opt := range options {
 		opt(opts)
 	}
+
+	generatePromotionPostsubmit := func(jobBaseGen *prowJobBaseBuilder) *prowconfig.Postsubmit {
```
droslean, line 240: I am not sure what this change is about. The only difference in the current implementation should be the cluster field.

danilo-gemoli: This factor out some code common to:
```golang
for _, arch := range architectures {
    ...
}
```
and
```golang
info.Config.MultiArch {
    ...
}
```
It's the same as before, just avoiding copy-paste.

droslean:
@danilo-gemoli note that the `breaking-changes` job shouldn't fail.

pkg/prowgen/prowgen.go
droslean: Let's don't refactor anything and focus only on the changes that are related to the PR's purpose.

pkg/prowgen/testdata/zz_fixture_TestGeneratePostSubmitForPromotion_spawn_on_multi01_cluster.yaml
```diff
@@ -0,0 +1,24 @@
+- agent: kubernetes
```
droslean, line 1: This test file proves that the implementation is wrong. The result should generate only one postsubmit.

danilo-gemoli: The current implementation of [`generatePostsubmitsForPromotion`](https://github.com/openshift/ci-tools/blob/09cdfa1aefc4a7d11a44e148f362fede533b5689/pkg/prowgen/prowgen.go#L234) creates at least one postsubmit even when neither `multi_arch` nor `additional_architectures` are set. We should either modify it entirely or create a new function like `generatePostsubmitsForPromotionMultiarch` then.

danilo-gemoli: We need to clarify the expected output for the following cases before keep modifying the code. What postsubmits and in which cluster do we expect them to be generated for each of the following prowgen configurations?  

1. 
```yaml
multi_arch: true
additional_architectures:
- arm64
```
2.
```yaml
multi_arch: true
```
3.
```yaml
additional_architectures:
- arm64
```
4.
```
{} // nothing at all
```

droslean: This PR has nothing to do with the `additional_architectures` and the changes are not affecting that feature. Currently, `prowgen` doesn't add any `cluster` field for the normal behavior. This is being done by the sanitize-jobs tool. However, we want to change this behavior when the `multi_arch` value is true.

danilo-gemoli: Yeah but a user can still set both `multi_arch` and `additional_architectures`. What postsubmits do we expect for the cases `1.` and `2.` then?
As the current implementation the case `3.` creates 2 postsubmits: one with no cluster defined and another one with `arm01`.
Case `4.` only one without any cluster set.
I agree on leaving `3.` and `4.` untouched but what about `1.` and `2.`?

droslean: The `additional_architectures` doesn't affect anything. The only thing we need to take care of in this PR is to add the `cluster: multi01` field in the generated postsubmit if the `multi_arch` is true in prowgen's configuration.

pkg/prowgen/prowgen.go
```diff
@@ -236,6 +236,24 @@ func generatePostsubmitsForPromotion(jobBaseBuilderFactory func() *prowJobBaseBu
 	for _, opt := range options {
 		opt(opts)
 	}
+
+	if info.Config.MultiArch {
+		jobBaseBuilder := jobBaseBuilderFactory()
```
droslean, line 241: What's the issue with the current implementation? I don't understand why we need to re-implement everything just to include a cluster.

droslean: @danilo-gemoli Check https://github.com/openshift/ci-tools/blob/master/pkg/prowgen/jobbase.go#L53-L100

pkg/prowgen/prowgen.go
```diff
@@ -160,6 +157,17 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	}, nil
 }
 
+func NewProwJobBaseBuilderForPromotion(configSpec *cioperatorapi.ReleaseBuildConfiguration,
```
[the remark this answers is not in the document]
droslean: Yea, it should be added there, but I wanted to point out, that we already have a mechanism to generate the base of the job. In your previous version of the code, you were doing that again.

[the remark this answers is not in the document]
danilo-gemoli: 👍🏻

PR #3457, opened 2023-06-01
ci-operator: remove claim release image from input

bbguimaraes:
These changes remove the pull-spec for the release payload image of the cluster
pool used by a test from the inputs used to determine the temporary test
namespace.

---

When a test requests a cluster from a pool (i.e. it has a `cluster_claim`), the
release payload configured for the pool, along with its images, is imported into
the test namespace (see https://github.com/openshift/ci-tools/pull/2074).  This
is done so that tests can use these images in the same way as they would other
types of releases.

---

The original implementation had a problem: since only tests with a
`cluster_claim` are given the Hive `kubeconfig`, other tests in the same
configuration would fail to obtain this information (see
https://github.com/openshift/ci-tools/pull/2092).

https://github.com/openshift/ci-tools/pull/2099 attempted to fix this by only
including the release import step when the client is present.  This has the
dubious effect of executing the test without claiming any cluster if the client
is not provided.  But, more importantly, it introduces another problem: since
the `ImportReleaseStep` declares the pull-spec of the release payload in its
`Inputs` --- as normally we want a separate test namespace if that pull-spec
has changed --- tests with an otherwise equal set of inputs are assigned
different temporary namespaces depending on whether they have a `cluster_claim`
or not.  E.g.:

- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hive/2029/pull-ci-openshift-hive-master-e2e/1661129286888722432
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hive/2029/pull-ci-openshift-hive-master-e2e-pool/1661129286913888256
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hive/2029/pull-ci-openshift-hive-master-images/1661129286939054080
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hive/2029/pull-ci-openshift-hive-master-unit/1661129286964219904
- https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_hive/2029/pull-ci-openshift-hive-master-verify/1661129286989385728

All of these jobs have the same set of inputs: they come from the same pull
request, have the same configuration, etc., but the first two are put into a
separate namespace.  While this is not a correctness problem (we could in
theory execute each test in its own namespace), this wastes resources since
none of the work (image builds, etc.) is shared between the two namespaces.

---

This behavior of `ImportReleaseStep` is desired in the case of regular release
payloads: we _want_ to import the new release if the pull-spec has changed, e.g.
if a new version is published.  But in the specific case of cluster pools, it is
not as important.  While some of them are regularly updated to use new payloads,
tests which use pools are not as concerned with using the absolute latest
versions.  Furthermore, it is already possible (in principle, at least) for a
test to import the release payload for a pool and later receive a lease for a
cluster provisioned with a different version.

For these reasons, these changes modify the `ImportReleaseStep` to only
conditionally include its pull-spec in the `Inputs`.  This continues to be done
for all cases, except now when the step is the result of a `cluster_claim`.

---

The release import refactoring in these changes is moderately over-abstracted
because it was part of the solution to an earlier conception I had of the
problem.  Still, it improves the mess that was the code in `pkg/defaults`, so
I'm including it here.

---

I want to monitor the deployment.

/hold

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -290,7 +290,7 @@ func fromConfig(
 					if err != nil {
 						return nil, nil, results.ForReason("reading_release").ForError(fmt.Errorf("failed to read input release pullSpec %s: %w", name, err))
 					}
-					logrus.Infof("Resolved release %s to %s", name, pullSpec)
+					logrus.Infof("Using explicitly provided pull-spec for release %s (%s)", name, pullSpec)
 					target := rawStep.ReleaseImagesTagStepConfiguration.TargetName(name)
 					releaseStep = releasesteps.ImportReleaseStep(name, nodeName, target, pullSpec, true, config.Resources, podClient, jobSpec, pullSecret, nil)
 				} else {
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -397,11 +397,10 @@ func stepForTest(
 		if len(leases) != 0 {
 			params = api.NewDeferredParameters(params)
 		}
-		var testSteps []api.Step
+		var ret []api.Step
 		step := multi_stage.MultiStageTestStep(*c, config, params, podClient, jobSpec, leases, nodeName, targetAdditionalSuffix)
 		if len(leases) != 0 {
 			step = steps.LeaseStep(leaseClient, leases, step, jobSpec.Namespace)
-			addProvidesForStep(step, params)
 		}
 		// hive client may not be present for jobs that execute non-claim based tests
 		if hiveClient != nil && c.ClusterClaim != nil {
@@ -414,13 +413,12 @@ func stepForTest(
 			claimRelease := c.ClusterClaim.ClaimRelease(c.As)
 			logrus.Infof("Resolved release %s to %s", claimRelease.ReleaseName, pullSpec)
 			target := api.ReleaseConfiguration{Name: claimRelease.ReleaseName}.TargetName()
-			importStep := releasesteps.ImportReleaseStep(claimRelease.ReleaseName, nodeName, target, pullSpec, false, config.Resources, podClient, jobSpec, pullSecret, nil)
-			testSteps = append(testSteps, importStep)
-			addProvidesForStep(step, params)
+			ret = append(ret, releasesteps.ImportReleaseStep(claimRelease.ReleaseName, nodeName, target, pullSpec, false, config.Resources, podClient, jobSpec, pullSecret, nil))
 		}
-		testSteps = append(testSteps, step)
-		newSteps := stepsForStepImages(client, jobSpec, inputImages, test, imageConfigs)
-		return append(testSteps, newSteps...), hasReleaseStep, nil
+		addProvidesForStep(step, params)
+		ret = append(ret, step)
+		ret = append(ret, stepsForStepImages(client, jobSpec, inputImages, test, imageConfigs)...)
+		return ret, hasReleaseStep, nil
 	}
 	if test := c.OpenshiftInstallerClusterTestConfiguration; test != nil {
 		if !test.Upgrade {
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -29,16 +29,13 @@ import (
 	templateapi "github.com/openshift/api/template/v1"
 	buildclientset "github.com/openshift/client-go/build/clientset/versioned/typed/build/v1"
 	templateclientset "github.com/openshift/client-go/template/clientset/versioned/typed/template/v1"
-	hivev1 "github.com/openshift/hive/apis/hive/v1"
 
 	"github.com/openshift/ci-tools/pkg/api"
 	testimagestreamtagimportv1 "github.com/openshift/ci-tools/pkg/api/testimagestreamtagimport/v1"
 	"github.com/openshift/ci-tools/pkg/kubernetes"
 	"github.com/openshift/ci-tools/pkg/lease"
 	"github.com/openshift/ci-tools/pkg/release"
-	"github.com/openshift/ci-tools/pkg/release/candidate"
 	"github.com/openshift/ci-tools/pkg/release/official"
-	"github.com/openshift/ci-tools/pkg/release/prerelease"
 	"github.com/openshift/ci-tools/pkg/results"
 	"github.com/openshift/ci-tools/pkg/secrets"
 	"github.com/openshift/ci-tools/pkg/steps"
@@ -162,7 +159,7 @@ func fromConfig(
 	rawSteps = append(graphConf.Steps, rawSteps...)
 	for _, rawStep := range rawSteps {
 		if testStep := rawStep.TestStepConfiguration; testStep != nil {
-			steps, testHasReleaseStep, err := stepForTest(ctx, config, params, podClient, leaseClient, templateClient, client, hiveClient, jobSpec, inputImages, testStep, &imageConfigs, pullSecret, censor, nodeName, targetAdditionalSuffix)
+			steps, testHasReleaseStep, err := stepForTest(config, params, podClient, leaseClient, templateClient, client, hiveClient, jobSpec, inputImages, testStep, &imageConfigs, pullSecret, censor, nodeName, targetAdditionalSuffix)
 			if err != nil {
 				return nil, nil, err
 			}
@@ -197,12 +194,14 @@ func fromConfig(
 			if overrideCLIResolveErr != nil {
 				return nil, nil, results.ForReason("resolving_cli_override").ForError(fmt.Errorf("failed to resolve override CLI image for release %s: %w", resolveConfig.Name, overrideCLIResolveErr))
 			}
+			var source releasesteps.ReleaseSource
 			if env := utils.ReleaseImageEnv(resolveConfig.Name); params.HasInput(env) {
 				value, err = params.Get(env)
 				if err != nil {
 					return nil, nil, results.ForReason("resolving_release").ForError(fmt.Errorf("failed to get %q parameter: %w", env, err))
 				}
 				logrus.Infof("Using explicitly provided pull-spec for release %s (%s)", resolveConfig.Name, value)
+				source = releasesteps.NewReleaseSourceFromPullSpec(value)
 			} else {
 				switch {
 				case resolveConfig.Integration != nil:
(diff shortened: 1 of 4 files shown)


bbguimaraes:
```
INFO[2023-06-02T13:20:33Z] pod pending for more than 30m0s: container "ci-scheduling-dns-wait" has not started in 30m0.007940494s:
```

/test e2e

pkg/defaults/defaults.go
```diff
@@ -218,23 +217,14 @@ func fromConfig(
...
-			}
-			if value != "" {
-				logrus.Infof("Resolved release %s to %s", resolveConfig.Name, value)
-				step := releasesteps.ImportReleaseStep(resolveConfig.Name, nodeName, resolveConfig.TargetName(), value, false, config.Resources, podClient, jobSpec, pullSecret, overrideCLIReleaseExtractImage)
-				buildSteps = append(buildSteps, step)
-				addProvidesForStep(step, params)
 			}
+			step := releasesteps.ImportReleaseStep(resolveConfig.Name, nodeName, resolveConfig.TargetName(), source, false, config.Resources, podClient, jobSpec, pullSecret, overrideCLIReleaseExtractImage)
+			buildSteps = append(buildSteps, step)
+			addProvidesForStep(step, params)
```
danilo-gemoli, line 227: `addProvidesForStep` has been moved out of the branch. Does it make any difference?

pkg/steps/release/import_release.go
```diff
@@ -67,7 +66,8 @@ type importReleaseStep struct {
 }
 
 func (s *importReleaseStep) Inputs() (api.InputDefinition, error) {
-	return api.InputDefinition{s.pullSpec}, nil
+	input, err := s.source.Input(context.Background())
```
danilo-gemoli, line 69: I'm not sure if just this single case is enough to justify a change of [this signature](https://github.com/openshift/ci-tools/blob/56fd6a6664337b8d989ea5940387abc7f08f5bfe/pkg/api/graph.go#LL22C2-L22C10) into:
```golang
Inputs(ctx context.Context) (InputDefinition, error)
````

File: pkg/steps/release/source.go [hunk 3]
```diff
@@ -0,0 +1,151 @@
+package release
```
danilo-gemoli, line 1: Ask ChatGPT to generate some unit tests :)

pkg/steps/release/import_release.go
```diff
@@ -103,8 +103,10 @@ func (s *importReleaseStep) run(ctx context.Context) error {
 	}
 
 	// tag the release image in and let it import
-	var pullSpec string
-
+	pullSpec, err := s.source.PullSpec(ctx)
+	if err != nil {
+		return err
```
danilo-gemoli, line 108:
```golang
return fmt.Errorf("unable to get pullspec for release %q: %w", s.name, err)
```

pkg/steps/release/source.go
```diff
@@ -0,0 +1,151 @@
...
+		}
+	}
+	return s.pullSpec, nil
+}
+
+func (s configurationReleaseSource) Input(ctx context.Context) (string, error) {
+	return s.PullSpec(ctx)
+}
+
+func (s *configurationReleaseSource) resolvePullSpec() (err error) {
```
danilo-gemoli, line 97: The fact the we have left `s.config.Integration` out makes me sleep bad at night... but I understand that we would need a totally different abstraction to include this whole [piece of code](https://github.com/openshift/ci-tools/blob/56fd6a6664337b8d989ea5940387abc7f08f5bfe/pkg/defaults/defaults.go#L207-L220)

danilo-gemoli (changes requested):

pkg/defaults/defaults.go
bbguimaraes: Good eye.  The condition (`value == ""`) which caused it and the rest of the block to not be executed was only ever `true` when:

- an integration stream is used as input; we now have a `continue` for that case
- a `RELEASE_*` environment variable was set but empty, in which case you get what you deserve (and an indication in the output logs, because of our unbounded mercy)
- an invalid `resolveConfig`; impossible by prior validation, but we now also detect that in `pkg/steps/release/source.go`

You can argue that the last two cases are a change in behavior, but they are edge cases that I didn't really care to preserve.

pkg/steps/release/import_release.go
bbguimaraes: https://github.com/openshift/ci-tools/pull/483/commits/aac2db8bbbdd5af929334c1c15d00feaabc6558c

I considered making this change simply as a matter of performance (to avoid creating useless `Context` objects), but it didn't seem like enough justification

pkg/steps/release/source.go
bbguimaraes: Integration streams are a completely different case and not handled at all in the same manner as all the other cases.  They are assembled from tags in an `ImageStream`, not imported as a single image.

bbguimaraes:
> Improving the section [Input resolution](https://github.com/openshift/ci-docs/blob/51406ee2bec90dfc1df99733572594fc37c9d772/content/en/docs/internals/_index.md?plain=1#L57-L101) by describing precisely what constitutes the finale hash value could be a potential TODO. Should I open a card @bbguimaraes?

It's on my list to add this to https://docs.ci.openshift.org/docs/internals/steps/ (the description of release steps is slightly outdated there, that's also on my list).  Each step already has an "input" section, but it does not describe the behavior of `Inputs`.

danilo-gemoli:
> I'll see if I have some time left to add some tests, but I wouldn't hold the pull request just for this.

We should, I won't. No regrets, I hope...

If everything is going to work as expected and if I have understood these changes properly, we should notice:
- a reduction of namespaces being used
- less resources consumption
- probably something else I'm not taking into the account (?)

in a matter of weeks.
 
I honestly don't know how many in terms of %. We can make a forecast but I guess it will simpler to just have a look at the metrics.

/lgtm

bbguimaraes:
We will be able to observe the effect immediately in jobs for repositories which have both types of tests.  I looked at https://prow.ci.openshift.org/?repo=openshift%2Fhive when I was diagnosing this problem; we can look at all jobs triggered for a given pull request and verify that they use the same namespace — which is currently not the case.

I don't know what the impact is, some information from the initial discussion of this problem:

> I imagine this results in a fair amount of wasted resources with the redundant builds etc.  But I don't know exactly what the impact is.  In trying to quantify it, we have 606 tests (out of ~41k) using claims, or 269 configuration files (out of ~8.6k).

bbguimaraes:
Hmm, the error seems legitimate, but that same test passed previously with the same code…

/hold

bbguimaraes:
As hypothesized, the test now passed.  Even though I think it's not related, I'm keeping the hold until I can explain the failure in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3457/pull-ci-openshift-ci-tools-master-e2e/1666361697566199808.

bbguimaraes:
Two differences from the previous version:

- 1e7495c0ef25c7d3b9a16d4ee221c4eb176d74f2 is new and adds the missing `JOB_SPEC` to some of the existing E2E tests.  This caused tests to still not use the same namespace.
- 3b0d33da31b98c80a2c55f796f5da50e6727fa67 corrects the receiver of `configurationReleaseSource::Input`.  It was mistakenly a value, meaning the `pullSpec` was not stored and later retrieved a second time — still correct, but inefficient.  `range-diff`:

```diff
 4:  f9926edf4 ! 13:  3b0d33da3 ci-operator: remove claim release image from input
    @@ pkg/steps/release/source.go: func (s configurationReleaseSource) PullSpec(
        return s.pullSpec, nil
      }
      
    -+func (s configurationReleaseSource) Input(ctx context.Context) (string, error) {
    ++func (s *configurationReleaseSource) Input(ctx context.Context) (string, error) {
     +  return s.PullSpec(ctx)
     +}
     +
```

bbguimaraes:
> As hypothesized, the test now passed. Even though I think it's not related, I'm keeping the hold until I can explain the failure in https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3457/pull-ci-openshift-ci-tools-master-e2e/1666361697566199808.

I really can't explain this one.  We imported the images here:

```
INFO[2023-06-07T09:00:35Z] Importing release image latest.
```

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3457/pull-ci-openshift-ci-tools-master-e2e/1666361697566199808#1:build-log.txt%3A58

30m later the pod is still pending:

```
INFO[2023-06-07T09:33:38Z] pod pending for more than 30m0s: containers have not started in 30m0.005754337s: release:
```

https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3457/pull-ci-openshift-ci-tools-master-e2e/1666361697566199808#1:build-log.txt%3A61

But it uses the correct image:

```
$ curl -sS https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3457/pull-ci-openshift-ci-tools-master-e2e/1666361697566199808/artifacts/e2e/e2e/artifacts/TestMultiStage_non-claim_test_in_config_with_a_claim/build-resources/pods.json | jq -r '.items[]|select(.metadata.name=="release-images-latest-cli").spec.containers[0].image'
image-registry.openshift-image-registry.svc:5000/ci-op-fzx4rlyq/release@sha256:d6efa3107f28b3eab7d44034e52a94296027ecf5d84826bc69ad6f91f006186c
```

And the image was imported almost immediately:

```
$ curl -sS https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_ci-tools/3457/pull-ci-openshift-ci-tools-master-e2e/1666361697566199808/artifacts/e2e/e2e/artifacts/TestMultiStage_non-claim_test_in_config_with_a_claim/build-resources/imagestreams.json | jq '.items[]|select(.metadata.name=="release").status'
{
  "dockerImageRepository": "image-registry.openshift-image-registry.svc:5000/ci-op-fzx4rlyq/release",
  "publicDockerImageRepository": "registry.build04.ci.openshift.org/ci-op-fzx4rlyq/release",
  "tags": [
    {
      "tag": "latest",
      "items": [
        {
          "created": "2023-06-07T09:00:35Z",
          "dockerImageReference": "quay.io/openshift-release-dev/ocp-release@sha256:d6efa3107f28b3eab7d44034e52a94296027ecf5d84826bc69ad6f91f006186c",
          "image": "sha256:d6efa3107f28b3eab7d44034e52a94296027ecf5d84826bc69ad6f91f006186c",
          "generation": 2
        }
      ]
    }
  ]
}
```

And the source has been there since 2023-05-18:

https://quay.io/repository/openshift-release-dev/ocp-release?tab=tags

So I have no idea why we got those `manifest unknown` errors for 30m and failed to pull the image.

bbguimaraes:
`TestObservers/observer_failure_must_not_affect_tests`

/retest
/hold cancel

bbguimaraes:
> - [1e7495c](https://github.com/openshift/ci-tools/commit/1e7495c0ef25c7d3b9a16d4ee221c4eb176d74f2) is new and adds the missing `JOB_SPEC` to some of the existing E2E tests.  This caused tests to still not use the same namespace.

This was fruitless since we [intentionally force a unique namespace](https://github.com/openshift/ci-tools/blob/8b9a2cbbb3a1380ec3d22eaa75ba7c85f8188dd5/test/e2e/framework/ci-operator.go#L89).  Leaving since it's still the correct thing to do.

bbguimaraes:
https://issues.redhat.com/browse/DPTP-3467

PR #3456, opened 2023-06-01
`e2e`: mitigate observer pods tests flakiness - 2nd attempt

danilo-gemoli:
Second attempt since the first one [#3025](https://github.com/openshift/ci-tools/pull/3025) didn't gain enough consensus.

Tests history of [pull-ci-openshift-ci-tools-master-e2e](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/pr-logs/directory/pull-ci-openshift-ci-tools-master-e2e) shows that a lot of flakes are due to the e2e test: `observer failure must not affect tests` (reference [here](https://github.com/openshift/ci-tools/blob/ff60d549afed41d1adf49843a33c1dd6ffdc9284/test/e2e/observer/e2e_test.go#L53-L69)).

The purpose of the aforementioned test is to show that a failure in an observer pod doesn't affect the result of the whole test. Execution breakdown follows:
- run an observer that it is going to fail
- run a test that it is going to succeed
- check that the whole test passed (despite the failing observer)
- check that the observer has failed

The last step is the problematic one as it happens that, from time to time, the cluster doesn't timely execute the observer therefore the run goes well but the `e2e` fails cause it won't find any traces of the observer having failed.

This is the output of a typical flake:
```
        INFO[2023-05-30T15:19:09Z] Step with-failing-observer-failing-observer failed after 41s. 
        INFO[2023-05-30T15:19:09Z] Ran for 47s                                  
    test.go:89: got diff between expected and actual result: 
        --- artifacts/failing-observer-junit_operator.xml
        +++ /logs/artifacts/TestObservers_observer_failure_must_not_affect_tests/junit_operator.xml
        @@ -1,5 +1,5 @@
6 skipped lines...
        @@ -11,9 +11,6 @@
             <testcase name="Run multi-stage test test phase" time="whatever">
               <system-out>The collected steps of multi-stage phase test.</system-out>
             </testcase>
        -    <testcase name="Run multi-stage test with-failing-observer - with-failing-observer-failing-observer container test" time="whatever">
[...]

PR #3465, opened 2023-06-06
`e2e`: mitigate observer pods tests flakiness - 3rd attempt

danilo-gemoli:
This is a lighter and cheaper alternative to #3456.
The problem is well described in #3456 and [DPTP-3140](https://issues.redhat.com/browse/DPTP-3140)


danilo-gemoli:
> While it is less likely to manifest here in a unit test, we could in principle go through the entirety of the pre/test/post steps before the observer has had a chance to be executed

Yes.

> Do we need to add thread synchronization to the FakePodExecutor?

If by "synchronization" you mean the guarantee that any observers is going to be executed then yes, we do need it.

These two inaccuracies were intentional as in the past I've tried several implementation and none of them have really convinced anyone, so I'll go ahead implementing something once we reach a consensus.

From my point of view there are two solutions:
- Extend the `FakePodExecutor` and add the capability to guarantee the execution of some pods, observers in our case. This would need a lot of synchronization code.
- > We can verify that the observers were created and not add them to names.
  
  What you have already stated would be the second solution, that is: leave everything as it is and just check for the 
  observers, whatever the outcome is, success or failure.
  It's way simpler than the former but nondeterministic.

bbguimaraes:
>> Do we need to add thread synchronization to the FakePodExecutor?
>
> If by "synchronization" you mean the guarantee that any observers is going to be executed then yes, we do need it.

No, I mean specifically its implementation.  It is now used by more than one thread — the one for regular steps and the one for observers — and it has no internal synchronization (mutual exclusion).

> Extend the `FakePodExecutor` and add the capability to guarantee the execution of some pods, observers in our case. This would need a lot of synchronization code.

We should certainly not do that as it would differ from the real `Pod` client.

>> We can verify that the observers were created and not add them to names.
>
> What you have already stated would be the second solution, that is: leave everything as it is and just check for the
> observers, whatever the outcome is, success or failure.
> It's way simpler than the former but nondeterministic.

I'm again talking very specifically about the implementation of this test.  Sorting _all_ names will not detect reordering of regular steps, which should not happen.  But also…

> > While it is less likely to manifest here in a unit test, we could in principle go through the entirety of the pre/test/post steps before the observer has had a chance to be executed
> 
> Yes.

Isn't this what we already have with the E2E test?

jmguzik (approved):

PR #3468, opened 2023-06-06
Add leases for s390x multiarch

werled:
These are analogous changes to the ones #3467 for ppc64le, and analogous to the changes in #3358


werled:
Replaced by #3479 for both ppc64le and s390x

PR #3471, opened 2023-06-07
`slack-bot`: update the auto response to use workflows to note the new location

smg247:
Slack moved the workflows, users are already confused. I updated the daily reminder already, but we need to note the new location in this response as well.


danilo-gemoli:
Seriously? What's going on here?
/retest

PR #3474, opened 2023-06-08
Use a new step for promotion to quay

hongkailiu:
Will rebase after https://github.com/openshift/ci-tools/pull/3473 is in.

This PR uses another step to promote images to quay.io: faster to run two pods in parallel.
The build time has been almost doubled after we started to promote to quay.io.

/cc @openshift/test-platform
/assign @bbguimaraes

### DIFF
--- a/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
+++ b/pkg/controller/prpqr_reconciler/prpqr_reconciler.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -443,14 +444,15 @@ type aggregatedOptions struct {
 func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration, defaulter periodicDefaulter, baseCiop *api.Metadata, prpqrName, prpqrNamespace string, pr *v1.PullRequestUnderTest, mimickedJob string, inject *api.MetadataWithTest, aggregatedOptions *aggregatedOptions) (*prowv1.ProwJob, error) {
 	fakeProwgenInfo := &prowgen.ProwgenInfo{Metadata: *baseCiop}
 
-	var annotations map[string]string
+	annotations := map[string]string{
+		releaseJobNameAnnotation: mimickedJob,
+	}
 	labels := map[string]string{
 		releaseJobNameLabel: jobNameHash(mimickedJob),
 	}
 
-	hashInput := prowgen.CustomHashInput(prpqrName)
+	var aggregateIndex *int
 	if aggregatedOptions != nil {
-		hashInput = prowgen.CustomHashInput(fmt.Sprintf("%s-%d", prpqrName, aggregatedOptions.aggregatedIndex))
 		if aggregatedOptions.labels != nil {
 			for k, v := range aggregatedOptions.labels {
 				labels[k] = v
@@ -459,13 +461,12 @@ func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration, defaulter period
 		annotations = map[string]string{
 			releaseJobNameAnnotation: aggregatedOptions.releaseJobName,
 		}
+		aggregateIndex = &aggregatedOptions.aggregatedIndex
 	} else {
 		labels[v1.PullRequestPayloadQualificationRunLabel] = prpqrName
-		annotations = map[string]string{
-			releaseJobNameAnnotation: mimickedJob,
-		}
 	}
 
+	hashInput := prowgen.CustomHashInput(prpqrName)
 	var periodic *prowconfig.Periodic
 	for i := range ciopConfig.Tests {
 		if ciopConfig.Tests[i].As != inject.Test {
@@ -473,6 +474,9 @@ func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration, defaulter period
 		}
 		jobBaseGen := prowgen.NewProwJobBaseBuilderForTest(ciopConfig, fakeProwgenInfo, prowgen.NewCiOperatorPodSpecGenerator(), ciopConfig.Tests[i])
 		jobBaseGen.PodSpec.Add(prowgen.InjectTestFrom(inject))
+		if aggregateIndex != nil {
+			jobBaseGen.PodSpec.Add(prowgen.TargetAdditionalSuffix(strconv.Itoa(*aggregateIndex)))
+		}
 
 		// Avoid sharing when we run the same job multiple times.
 		// PRPQR name should be safe to use as a discriminating input, because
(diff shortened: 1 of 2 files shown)


pkg/defaults/defaults.go
```diff
@@ -362,7 +362,8 @@ func fromConfig(
 		if config.PromotionConfiguration == nil {
 			return nil, nil, fmt.Errorf("cannot promote images, no promotion configuration defined")
 		}
-		postSteps = append(postSteps, releasesteps.PromotionStep(config, requiredNames, jobSpec, podClient, pushSecret))
+		postSteps = append(postSteps, []api.Step{releasesteps.PromotionStep(config, requiredNames, jobSpec, podClient, pushSecret, false),
+			releasesteps.PromotionStep(config, requiredNames, jobSpec, podClient, pushSecret, true)}...)
```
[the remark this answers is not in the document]
hongkailiu: That was exactly my first version of implementation and abandoned.
I will change it back.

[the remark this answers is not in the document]
hongkailiu: Done!

pkg/api/promotion.go
```diff
@@ -66,3 +75,53 @@ func ExtractPromotionName(configSpec *ReleaseBuildConfiguration) string {
 	}
 	return ""
 }
+
+// GetImageMirrorTargetQuay mutates a mirror mapping to the one for our repo in quay.io
+func GetImageMirrorTargetQuay(mirror map[string]string, params map[string]string) map[string]string {
+	today, ok := params[ParamKeyToday]
+	if !ok || today == "" {
```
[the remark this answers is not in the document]
bbguimaraes: But why not put `today` in a closure in `defaults.go` and keep the promotion step interface a simple `func(map[string]string) map[string]string`?

bbguimaraes (changes requested):

hongkailiu:
/hold

Need to test the image with a job.

pkg/steps/release/promote_test.go
```diff
@@ -571,18 +571,20 @@ func TestGetPromotionPod(t *testing.T) {
 			namespace: "ci-op-zyvwvffx",
 		},
 		{
-			name: "tag_by_commit is true",
```
hongkailiu, line 574: The bug (fixed by https://github.com/openshift/ci-tools/pull/3481/files) was hit when the same image (the same digest) is promoted to different tags (thus, it would lead to the same tag in quay.io because the tag in quay is determined by the original image's digest).
`tag_by_commit` is only its special case.

With the implementation of this PR, the problem is fixed implicitly because it just does the assignment of the map with the same key/value twice (i.e., no-op for the 2nd time) which is used to define the `oc image mirror` command.

pkg/api/promotion.go
```diff
@@ -66,3 +75,53 @@ func ExtractPromotionName(configSpec *ReleaseBuildConfiguration) string {
...
+		} else {
+			for _, quayTag := range quayTags {
+				ret[quayTag] = v
+			}
+		}
+	}
+	return ret
+}
+
+func tagsInQuay(image string, target string, date string) ([]string, error) {
```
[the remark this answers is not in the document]
hongkailiu: Modified.

pkg/steps/release/promote.go
```diff
@@ -176,7 +185,19 @@ func getImageMirrorTarget(tags map[string][]api.ImageStreamTagReference, pipelin
...
+				if quayTags, err := tagsInQuay(dockerImageReference, dst, date); err != nil {
+					logrus.WithField("key", k).WithError(err).
+						Warn("Failed to get the tag in quay.io and skipped the promotion to quay for this image")
+				} else {
+					for _, quayTag := range quayTags {
+						imageMirror[quayTag] = dockerImageReference
+					}
+				}
+			}
 			namespaces.Insert(dst.Namespace)
```
bbguimaraes, line 201: Specifically, these two branches are exactly what the body of the two parameter functions should be.

bbguimaraes (commented):
I don't understand why we are back to explicitly testing step and registry names instead of adding a parameter to the step.  These checks happen in these cases, all of which could be eliminated:

- To skip promotion when `registry_override` is set in the promotion configuration.  We should just detect this much earlier and not create the extra promotion step in the first place.
- To skip creating namespaces.  We can just return an empty list of namespaces from the parameter function, no extra checks necessary.
- To determine the destination tags.  This is what the parameter function should do.

pkg/steps/release/promote.go
hongkailiu: > I don't understand why we are back to explicitly testing step and registry names

I just feel it is easier this way than parsing function. But I will try another time the latter one.

pkg/api/promotion.go
```diff
@@ -66,3 +76,36 @@ func ExtractPromotionName(configSpec *ReleaseBuildConfiguration) string {
 	}
 	return ""
 }
+
+func tagsInQuay(image string, tag ImageStreamTagReference, date string) ([]string, error) {
+	if date == "" {
+		return nil, fmt.Errorf("date must not be empty")
+	}
```
bbguimaraes, line 83: Seems a bit strange to check a value that is completely under our control.

pkg/steps/release/promote.go
```diff
@@ -94,25 +94,32 @@ func (s *promotionStep) run(ctx context.Context) error {
 		return fmt.Errorf("could not resolve pipeline imagestream: %w", err)
 	}
 
-	imageMirrorTarget, namespaces := getImageMirrorTarget(tags, pipeline, registryDomain(s.configuration.PromotionConfiguration))
+	date := time.Now().Format("20060102")
+	imageMirrorTarget, namespaces := getImageMirrorTarget(tags, pipeline, registryDomain(s.configuration.PromotionConfiguration), date, s.mirrorFunc)
+	if s.name == api.PromotionQuayStepName {
+		imageMirrorTarget, _ = getImageMirrorTarget(tags, pipeline, api.QuayOpenShiftCIRepo, date, s.mirrorFunc)
```
bbguimaraes, line 100: This assignment will retain the value of `namespaces` from the previous call.

pkg/steps/release/promote.go
```diff
@@ -94,25 +94,32 @@ func (s *promotionStep) run(ctx context.Context) error {
 		return fmt.Errorf("could not resolve pipeline imagestream: %w", err)
 	}
 
-	imageMirrorTarget, namespaces := getImageMirrorTarget(tags, pipeline, registryDomain(s.configuration.PromotionConfiguration))
+	date := time.Now().Format("20060102")
+	imageMirrorTarget, namespaces := getImageMirrorTarget(tags, pipeline, registryDomain(s.configuration.PromotionConfiguration), date, s.mirrorFunc)
+	if s.name == api.PromotionQuayStepName {
```
bbguimaraes, line 99: The difference here is just the target registry, it should be another parameter of the step.

hongkailiu:
Oops, I pushed a wrong commit and the PR was closed. I will make a new one.

PR #3475, opened 2023-06-08
TRT-1071: Honor MasterNodesUpdated flag when present for aggregated d…

neisw:
Looks for the MasterNodesUpdated flag in the JobRun(s) cluster-data json file

If present includes the Y / N value in the where clause for aggregated disruption queries

If not present it is a noop and no additional where clause is added

Intention is to produce better aggregation disruption backstop data by comparing current run data to job runs that went through a similar path


xueqzhan:
/lgtm

It is great to see this differentiated. Will this be introduced to the per job disruption test as well? There is value there too.

PR #3479, opened 2023-06-12
Add leases for ppc64le and s390x multiarch

werled:
Analogous to the changes in #3358 (combines #3467, #3468)

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1390,10 +1390,12 @@ func (p ClusterProfile) ClusterType() string {
 		return string(CloudGCP)
 	case
 		ClusterProfileIBMCloud,
-		ClusterProfileIBMCloudQE,
-		ClusterProfileIBMCloudMultiPpc64le, // TODO: Check if the return value should be "ibmcloud-multi-ppc64le" instead of "ibmcloud"
-		ClusterProfileIBMCloudMultiS390x:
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -1386,7 +1390,9 @@ func (p ClusterProfile) ClusterType() string {
 		return string(CloudGCP)
 	case
 		ClusterProfileIBMCloud,
-		ClusterProfileIBMCloudQE:
+		ClusterProfileIBMCloudQE,
+		ClusterProfileIBMCloudMultiPpc64le, // TODO: Check if the return value should be "ibmcloud-multi-ppc64le" instead of "ibmcloud"
```
jaypoulz, line 1394: This is ultimately a billing question. Let's find out who owns the wallet for the IBM Cloud account(s)

jaypoulz: I suspect we will want "ibmcloud-multi-ppc64le" and "ibmcloud-multi-s390x" so that we can have our own cluster profiles. :)

jaypoulz: (which in turn will specify the leases).

werled: Thanks for the comments, updated :).

werled:
As far as I understand, the failing check has nothing to do with the changes in the PR, but currently doesn't pass for all PRs.

PR #1684, opened 2021-02-10
Grant test runner serviceaccount view role within namespace

jupierce:
Permits the serviceaccount to read non-secret elements of
the test namespace. More importantly, it can now grant that
permission to others. Cluster-bot uses pod status to monitor
cluster provisioning operations within a test namespace.

PR #3487, opened 2023-06-19
multi-stage: properly name role binding

bbguimaraes:
https://github.com/openshift/ci-tools/pull/1684 added an extra role binding to
the `ServiceAccount` used by each multi-stage test.  However, the name it uses
is not unique, meaning a single binding will be created (`util.CreateRBACs`
ignores already-existing bindings with the same name).

This problem went unnoticed for this long because the jobs which depend on this
functionality usually get their own namespace.  This was discovered while
investigating a test which abuses this (very abusable) mechanism to look at
parts of the temporary namespace it should have no business looking at:

https://redhat-internal.slack.com/archives/CBN38N3MW/p1687185233834439

While not meant to make that particular case work, this is a legitimate
correctness problem.

### DIFF
--- a/pkg/steps/multi_stage/init.go
+++ b/pkg/steps/multi_stage/init.go
@@ -141,9 +141,13 @@ func (s *multiStageTestStep) setupRBAC(ctx context.Context) error {
 			Subjects:   subj,
 		},
 		{
-			ObjectMeta: meta.ObjectMeta{Namespace: ns, Name: "test-runner-view-binding", Labels: labels},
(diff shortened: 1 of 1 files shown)


bbguimaraes:
```
kubelet: Failed to create pod sandbox: rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

/retest

bbguimaraes:
`ip-10-0-128-94.ec2.internal` is having a really bad day.

PR #3488, opened 2023-06-19
Work towards unit tests for the aggregator

dgoodwin:
After repeated failures to roll out improvements, it's clear we need some test coverage here, as there's substantial cost to breaking the aggregator in a long feedback loop to realize payloads are broken. This is a first step towards that, just a basic test that covers when all jobs are not finished yet and our time is up, and the refactors needed to make it possible. More testing will cover in followups, but we don't want too huge a PR.

- More consistent logging for analyzer
- Fix CIDataClient mock not visible outside it's package
- Add a mock for the GCSClient
- Refactor use of raw GCS client onto our interface
- Remove test from index, not ready yet
- logging
- Add mock impl for JobRunInfo


dgoodwin:
/override ci/prow/e2e 

Repo is broken again.

jmguzik (approved):

PR #3495, opened 2023-06-20
Update the importReleaseStep to use the PreserveOriginal import mode

aleskandro:
Since OCP 4.13, the internal registry supports the PreserveOriginal import mode for ImageStreamTags when the ReferencePolicy is set to local.

Using the PreserveOriginal import mode for the ImageStreamTags of the imported release payload will allow testing the non-amd64 control plane cluster scenarios when installing via the multi-arch payload.

This will only affect tests importing a multi-arch release payload and is harmless otherwise.

cc @tvardema @jaypoulz @Prashanth684


aleskandro:
Refers ARMOCP-526 MIXEDARCH-198

droslean (approved):

PR #3502, opened 2023-06-21
Update golint configuration

stbenjam:
- Use a container for running golangci-lint outside of CI, this fixes problems where a developer is running a different version of go or golangci-lint than CI (e.g. avoids problems like https://github.com/golangci/golangci-lint/issues/3565)

- Use errorlint built in linter instead of the direct vendored version


dgoodwin:
Using the wrong version of golangci-lint crashed my machine using all 64 GB of RAM. 

/lgtm

dgoodwin:
also thanks @stbenjam !

PR #3503, opened 2023-06-22
TRT-1091: Add second pass for disruption checks with fuzzy comparison

neisw:
Adds a second check for aggregated disruption.

Made the fuzzy matching optional and disabled for the `zero-disruption should not be worse` tests

Diffs the expected number of passes from the actual passes in the first pass.  Uses that as a multiplier, the close we were to passing the lower the multiplier.

Diffs the historical disruption value from the actual disruption and multiplies this by the multiplier to create a fuzz value

Creates a fuzzThreshold based on the historical disruption value,  currently 4 and under  = 1, 5-9 = 2, 10-14 = 3, etc.

Checks to see if the fuzz value is less than or equal to the fuzzThreshold and if so considers it a pass.

We should be able to adjust the calculation for the fuzzThreshold to increase / decrease as needed.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
@@ -442,14 +442,39 @@ func (a *weeklyAverageFromTenDays) CheckPercentileDisruption(ctx context.Context
 		return failureJobRunIDs, []string{}, testCaseSkipped, message, nil
 	}
 
-	failureJobRunIDs, successJobRunIDs, testCaseFailed, summary := a.checkPercentileDisruption(jobRunIDToAvailabilityResultForBackend, historicalDisruptionStatistic, percentile)
+	failureJobRunIDs, successJobRunIDs, testCaseFailed, summary := a.checkPercentileDisruptionWithGrace(jobRunIDToAvailabilityResultForBackend, historicalDisruptionStatistic, percentile)
 	return failureJobRunIDs, successJobRunIDs, testCaseFailed, messagePrefix + summary, nil
 }
 
-func (a *weeklyAverageFromTenDays) checkPercentileDisruption(jobRunIDToAvailabilityResultForBackend map[string]jobrunaggregatorlib.AvailabilityResult, historicalDisruptionStatistic backendDisruptionStats, thresholdPercentile int) ([]string, []string, testCaseStatus, string) {
+func (a *weeklyAverageFromTenDays) checkPercentileDisruptionWithGrace(jobRunIDToAvailabilityResultForBackend map[string]jobrunaggregatorlib.AvailabilityResult, historicalDisruptionStatistic backendDisruptionStats, thresholdPercentile int) ([]string, []string, testCaseStatus, string) {
+	historicalThreshold := historicalDisruptionStatistic.percentileByIndex[thresholdPercentile]
+	requiredNumberOfPasses, noGraceFailureJobRunIDs, noGraceSuccessJobRunIDs, noGraceTestCasePassed, noGraceSummary := a.innerCheckPercentileDisruptionWithGrace(jobRunIDToAvailabilityResultForBackend, historicalThreshold, thresholdPercentile, 0)
+	numberOfPasses := len(noGraceSuccessJobRunIDs)
+	if numberOfPasses >= requiredNumberOfPasses {
+		return noGraceFailureJobRunIDs, noGraceSuccessJobRunIDs, noGraceTestCasePassed, noGraceSummary
+	}
+
+	// the +1 is to have at least 1s, the 5 is arbitrary because it made forrest feel good and the denominator makes the grace smaller
+	// when there are more job runs that need to be better
+	graceSeconds := ((int(historicalThreshold) / 5) + 1) / (requiredNumberOfPasses - numberOfPasses)
+	if graceSeconds <= 0 {
+		return noGraceFailureJobRunIDs, noGraceSuccessJobRunIDs, noGraceTestCasePassed, noGraceSummary
+	}
+	thresholdWithGrace := historicalThreshold + float64(graceSeconds)
+
+	_, withGraceFailureJobRunIDs, withGraceSuccessJobRunIDs, withGraceTestCasePassed, withGraceSummary := a.innerCheckPercentileDisruptionWithGrace(jobRunIDToAvailabilityResultForBackend, thresholdWithGrace, thresholdPercentile, graceSeconds)
+	return withGraceFailureJobRunIDs, withGraceSuccessJobRunIDs, withGraceTestCasePassed, withGraceSummary
+}
+
+func (a *weeklyAverageFromTenDays) checkPercentileDisruptionWithoutGrace(jobRunIDToAvailabilityResultForBackend map[string]jobrunaggregatorlib.AvailabilityResult, historicalDisruptionStatistic backendDisruptionStats, thresholdPercentile int) ([]string, []string, testCaseStatus, string) {
+	historicalThreshold := historicalDisruptionStatistic.percentileByIndex[thresholdPercentile]
+	_, noGraceFailureJobRunIDs, noGraceSuccessJobRunIDs, noGraceTestCasePassed, noGraceSummary := a.innerCheckPercentileDisruptionWithGrace(jobRunIDToAvailabilityResultForBackend, historicalThreshold, thresholdPercentile, 0)
+	return noGraceFailureJobRunIDs, noGraceSuccessJobRunIDs, noGraceTestCasePassed, noGraceSummary
+}
+
+func (a *weeklyAverageFromTenDays) innerCheckPercentileDisruptionWithGrace(jobRunIDToAvailabilityResultForBackend map[string]jobrunaggregatorlib.AvailabilityResult, threshold float64, thresholdPercentile int, graceSeconds int) (int, []string, []string, testCaseStatus, string) {
 	failureJobRunIDs := []string{}
 	successJobRunIDs := []string{}
-	threshold := historicalDisruptionStatistic.percentileByIndex[thresholdPercentile]
 	successRuns := []string{} // each string example: jobRunID=5s
 	failureRuns := []string{} // each string example: jobRunID=5s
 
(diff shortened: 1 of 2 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -446,22 +446,45 @@ func (a *weeklyAverageFromTenDays) CheckPercentileDisruption(ctx context.Context
...
 }
 
-func (a *weeklyAverageFromTenDays) checkPercentileDisruption(jobRunIDToAvailabilityResultForBackend map[string]jobrunaggregatorlib.AvailabilityResult, historicalDisruptionStatistic backendDisruptionStats, thresholdPercentile int) ([]string, []string, testCaseStatus, string) {
-	failureJobRunIDs := []string{}
-	successJobRunIDs := []string{}
-	threshold := historicalDisruptionStatistic.percentileByIndex[thresholdPercentile]
-	successRuns := []string{} // each string example: jobRunID=5s
-	failureRuns := []string{} // each string example: jobRunID=5s
-
+// evaluatePercentileDisruption performs a straight disruption > historicalThreshold comparison when no fuzzThreshold or multiplier are provided
```
DennisPeriquet, line 449: Nit: can you reduce the wideness of the comments so they fit on one line in the git review window (I realize we probably have no standard for this and everyone has a wide screen).  I'm having trouble digesting the comment because there is no punctuation.

DennisPeriquet: Does this capture what you are trying to say?

```
// evaluatePercentileDisruption performs a disruption vs. historical threshold comparision.
// But when fuzzThreshold AND multiplier are present (i.e., > 0), we add fuzziness to the comparison.
// The multiplier is the difference between the successful runs and the required successes; the multiplier makes
// it so that the closer we were to having the required successes, the lower the bar for finding runs that are
// below the fuzzThreshold.
// The difference between the disruption and historicalThreshold is multiplied by the multiplier and the result
// compared against the fuzzThreshold.
// The fuzzThreshold allows the "fuzz factor" to increase based on the historicalThreshold, currently:
//   4 and under = 1, 5-9 = 2, 10-14 = 3, etc.
```

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -488,14 +511,30 @@ func (a *weeklyAverageFromTenDays) checkPercentileDisruption(jobRunIDToAvailabil
...
-		)
-		return failureJobRunIDs, successJobRunIDs, testCaseFailed, summary
+		// the further away we are from our required number of passes the greater the multiplier
+		// this rewards runs that were close to having the required number of passes and penalizes those that weren't (raising the bar for overriding the original failure)
+		multiplier := float64(requiredNumberOfPasses - numberOfPasses)
+		// disruptionValue = (disruptionValue - historicalThreshold) * multiplier
+		// this will effectively change the comparison from straight disruption to comparing the fuzziness value
+		// fuzz is meant to increase the tolerance more as the historical disruption is larger, less than 5 fuzz == 1, 5 - 9 fuzz == 2, 10- 14 fuzz == 3, etc.
+		// we can tweak the fuzz factor by lowering / increasing the divisor
+		fuzz := (int64(threshold) / 5) + 1
```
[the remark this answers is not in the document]
DennisPeriquet: nit: if `fuzz` is a `fuzzThreshold`, I'd name it the latter.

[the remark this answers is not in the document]
neisw: Deal on fuzzThreshold.  For the divisor, it isn't anything I expect to be changed a lot and if I make it a constant then you have to go search for what the value is (as I would put it at the top of the file) so I'm thinking leave it as is.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
```diff
@@ -0,0 +1,146 @@
...
+			status:               testCasePassed,
+			failedCount:          0,
+			successCount:         10,
+		},
+		{
+			// Required Passes for 95th percentile is 6
+			// 5 Natural Passes
+			// Multiplier (6 - 5) == 1
+			// Fuzz Threshold ((historicalDisruption / 5) + 1) = 2
+			// The Disruption value that == 7 gets flipped to pass
```
dgoodwin, line 49: These tests are excellent, and especially the comments, could you update the logging to include these details as much as possible so we can see exactly what happened when we examine the test output. It looks like we'll log the threshold but the fuzziness details may help in debugging.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
```diff
@@ -0,0 +1,146 @@
...
+			historicalDisruption: 6,
+			status:               testCaseFailed,
+			failedCount:          5,
+			successCount:         5,
+		},
+		{
+			// Required Passes for 80th percentile is 4
+			// 4 Natural Passes
+			name:                 "Test 80th Percentile Pass",
+			disruptions:          []int{0, 0, 1, 1, 2, 2, 2, 2, 2, 2},
```
dgoodwin, line 91: Were these tests based on real world data?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
```diff
@@ -0,0 +1,146 @@
...
+		{
+			// Required Passes for 80th percentile is 4
+			// 2 Natural Passes
+			// Multiplier (4 - 2) == 2
+			// Fuzz Threshold ((historicalDisruption / 5) + 1) = 1
+			// There are no disruption values that get flipped since (2-1)*2 = 2 and our Fuzz Threshold is 1
+			// But the 8s and above do not so we don't have enough passes
+			name:                 "Test 80th Percentile Fuzzy Fail",
+			disruptions:          []int{0, 0, 2, 2, 2, 2, 2, 2, 2, 2},
+			thresholdPercentile:  80,
```
dgoodwin, line 121: If I'm reading correctly, I think the thresholdPercentile number is not actually relevant to the test, would be one less thing for the next person to look at these tests  to think about.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
```diff
@@ -0,0 +1,146 @@
...
+			historicalDisruption: 1,
+			status:               testCaseFailed,
+			failedCount:          8,
+			successCount:         2,
+		},
```
dgoodwin, line 126: Could you add a few tests with BIG historicalDisruption. We've got some P95s over 100s consistently and it would be interesting to see how this behaves when dealing with values like that.

dgoodwin (commented):
This is great, very important PR. @xueqzhan you've thought about this area as well, how does this feel to you?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
neisw: There were based on cases where I saw us fail by 1 missing required job off by 1 second.  But not real data from one of those runs.  I can add that.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
neisw: thresholdPercentile is used to set the historical disruption for the test

historicalDisruptionStatistic.percentileByIndex[test.thresholdPercentile] = test.historicalDisruption

and passed into the function to look that data up:

weeklyAverageFromTenDays.checkPercentileDisruption(jobRunIDToAvailabilityResultForBackend, historicalDisruptionStatistic, test.thresholdPercentile)

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
neisw: Yep, Dennis asked for something similar but I didn't think we had 100s in the wild.  Will do.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
dgoodwin: Does using different percentiles change any behavior or could you just hard code say 50 or 95 in the test?

neisw: The requiredNumberOfPasses changes based on the percentiles

workingPercentage := thresholdPercentile // the percentile is our success percentage
requiredNumberOfPasses := requiredPassesByPassPercentageByNumberOfAttempts[numberOfAttempts][workingPercentage]

dgoodwin: Ahh I get it now, thanks!

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail_test.go
neisw: How would the following be?  If there is not fuzziness applied it won't be included as in `run_1=0s` but for those cases, pass or fail, where fuzziness was evaluated the values used will be included in the test output `un_6=2s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 0.70) `

`Passed: Passed 5 times, failed 4 times.  (P85=1.30s requiredPasses=4 successes=[run_1=0s run_6=2s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 0.70) run_0=1s run_2=0s run_5=2s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 0.70)] failures=[run_3=5s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 3.70) run_4=5s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 3.70) run_7=3s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 1.70) run_8=4s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 2.70)])`

dgoodwin: Could we log anything constant to all earlier on (looks like multiplier and fuzzThreshold are constant), then trim down to just the fuzzValue in brackets with each failure.

And am I reading it right that for a failure, run_4=5s means my actual disruption was 5s, and the fuzzValue=3.70 is our forgiving value for it, which was still higher than 1.3s, so we fail.

neisw: If we wanted the constants logged higher up only once AND only when the fuzziness logic was actually used I would have to return an additional flag, but that seems doable.

Almost, the 3.70 value is compared against the fuzzThreshold which is 1, not the historicalDisruptionValue which is the 1.30.

run_4=5s (Fuzzy Match: multiplier: 1.00, fuzzThreshold: 1.00, fuzzValue: 3.70)
Multiplier == 1,  We needed 4 passes but originally had only 3 so 4-3 = 1 == multiplier
fuzzThreshold = 1.00,  P85== 1.30, ((1.30 / 5) + 1) = 1 == fuzzThreshold) 
fuzzValue == 3.70, DisruptionValue - HistoricalValue * multiplier (( 5 - 1.30) * 1) == 3.70

3.70 > 1 so we still fail for this run.

neisw: So yeah that is definitely cleaner :-)  let me know if you see a better way to format it.

`Passed: Passed 5 times, failed 4 times, with fuzzy matching values: multiplier 1.00, fuzzThreshold 1.00.  (P85=1.30s requiredPasses=4 successes=[run_0=1s run_1=0s run_2=0s run_5=2s (fuzzValue=0.70) run_6=2s (fuzzValue=0.70)] failures=[run_3=5s (fuzzValue=3.70) run_4=5s (fuzzValue=3.70) run_7=3s (fuzzValue=1.70) run_8=4s (fuzzValue=2.70)])`

dgoodwin:
This is a lgtm from me, lets hold to see if Ken has any thoughts and/or decide if we merge on Monday.

/lgtm
/hold

xueqzhan:
Very clever implementation! 

I feel like another multiplier/margin might help things further. Let me clarify with an example:

Assuming we have historical threshold of 6 with 0s disruption, and assuming we have 4 0s job runs and 6 1s job runs. 

This will fail with the multiplier implementation. But I strongly feel that is a case we want pass. Maybe another evaluation can be implemented that will, instead of add multiplier to the disruption, we increase the historical threshold by a margin. Again, for the example above, we use the fixed fuzzy number (without the multiplier) to decide whether all jobs pass. In this case, historical is 0s and all ten jobs will pass after fuzz adjustment. But instead of requiring only 6 to pass based on historical threshold, we can add a margin (maybe 2) in this example to require more (8) to pass with the fuzz adjusted pass rate. 

Let me know what you think. We can sync on this in person if needed.

neisw:
@xueqzhan I definitely agree there is still some tweaking to be done here.  I'm interested in seeing how this performs and finding more cases where we want to open it up and model them.  I think we could easily change the calculation for the multiplier to something like ((requiredNumberOfPasses - numberOfPasses) % 2) + 1 or something if we wanted to open it up more.

I do want to be careful that if we have a case that says we need 6 0s disruption runs and we get 10 1s disruption runs I don't think we should pass that currently.

Happy to work on this more if we can identify our edge cases where flip from wanting to pass to not allowing a fuzzy pass.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/pass_fail.go
```diff
@@ -442,26 +442,59 @@ func (a *weeklyAverageFromTenDays) CheckPercentileDisruption(ctx context.Context
...
+				isFuzzy = true
+			}
+		}
+
+		fuzzyData := ""
+		if isFuzzy {
+			fuzzyData = fmt.Sprintf(" (fuzzValue=%.2f)", disruptionValue)
+		}
+
+		if disruptionValue > thresholdValue {
```
deads2k, line 481: `(disruption.SecondsUnavailable - historicalThreshold) * multiplier > fuzzThreshold`

`(disruption.SecondsUnavailable - historicalThreshold) > fuzzThreshold/multiplier`

`disruption.SecondsUnavailable >  fuzzThreshold / multiplier + historicalThreshold`

`historicalThresholdPlusGrace = fuzzThreshold / multiplier + historicalThreshold`

`historicalThresholdPlusGrace = historicalThreshold + ((historicalThreshold / 5) + 1) / (requiredNumberOfPasses - numberOfPasses)`

```
grace := ((historicalThreshold / 5) + 1) / (requiredNumberOfPasses - numberOfPasses)
threshold = threshold + grace
```

neisw:
/hold

need to clean up some comments after the refactor

neisw:
@xueqzhan - This is the refactor with a small update to capture the grace value in the test output

neisw:
Logging without grace addition:

`Passed: Passed 10 times, failed 0 times.  (P95=7.00s requiredPasses=6 successes=[run_4=0s run_9=0s run_3=0s run_1=0s run_2=0s run_5=0s run_6=0s run_7=0s run_8=0s run_0=0s] failures=[])`

With grace addition:
`Passed: Passed 6 times, failed 4 times.  (P95=8.00s (grace=2) requiredPasses=6 successes=[run_0=5s run_2=5s run_3=6s run_4=6s run_1=5s run_5=7s] failures=[run_6=9s run_9=9s run_7=9s run_8=9s])`

PR #3507, opened 2023-06-22
`pj-rehearse`: fix infrequent error when checking out work tree

smg247:
Infrequently, we are seeing errors like the following:
```
could not load configuration from base revision of release repo: could not checkout worktree: '[git checkout 3b4aad5de6bb9bcffc01bc3c845589b54280c5ea]' failed with out: fatal: reference is not a tree: 3b4aad5de6bb9bcffc01bc3c845589b54280c5ea
and error exit status 128
```
For example, in this comment: https://github.com/openshift/release/pull/39156#issuecomment-1601294684

These resolve themselves, and seem to happen when the plugin is under a heavy load. I researched the error and for this [post](https://stackoverflow.com/questions/54225047/git-error-on-checkout-fatal-reference-is-not-a-tree) suggesting a `git fetch` prior to checkout. I want to try this and see if it reduces the occurrences of this error.

### DIFF
--- a/pkg/config/release.go
+++ b/pkg/config/release.go
@@ -84,6 +84,16 @@ func revParse(repoPath string, args ...string) (string, error) {
 	return strings.TrimSpace(out), nil
 }
 
+func gitFetch(candidatePath string) error {
+	cmd := exec.Command("git", "fetch")
+	cmd.Dir = candidatePath
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("'%s' failed with out: %s and error %w", cmd.Args, stdoutStderr, err)
+	}
+	return nil
+}
+
 func gitCheckout(candidatePath, baseSHA string) error {
 	cmd := exec.Command("git", "checkout", baseSHA)
 	cmd.Dir = candidatePath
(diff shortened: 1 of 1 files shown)


pkg/config/release.go
```diff
@@ -159,6 +169,10 @@ func GetAllConfigsFromSHA(releaseRepoPath, sha string) (*ReleaseRepoConfig, erro
 	if restoreRev == "HEAD" {
 		restoreRev = currentSHA
 	}
+	// git fetch to be sure we are in sync
```
bbguimaraes, line 172: Shouldn't this be done in the `pj-rehearse` code when we process a command?  Based on a quick `grep`, this function is also used by `config-change-trigger`, and it is unexpected that it now suddenly potentially performs a network request.

bbguimaraes: In addition, we load the base configuration after the candidate, so I don't see how we could not have already fetched the base branch.

smg247: > Shouldn't this be done in the pj-rehearse code when we process a command? Based on a quick grep, this function is also used by config-change-trigger, and it is unexpected that it now suddenly potentially performs a network request.

That's a fair point. It looks like this issue hasn't occurred in the last couple of days anyways, so I am going to hold this for now. If we do see the problem again we can find a more specific way to do this.

> In addition, we load the base configuration after the candidate, so I don't see how we could not have already fetched the base branch.

The error is definitely puzzling. This was really just something to try based on that stack overflow post. There is no deterministic way that I have discovered to reproduce this, so there isn't much better I can do right now. If it occurs as infrequently as it is now, I don't think we need to worry about it too much.

smg247:
It looks like this error stopped occurring on its own.

PR #3509, opened 2023-06-23
Prepare for partitioned TestRuns table in BigQuery

dgoodwin:
This table is over 200GB dating back more than 2 years. It's costing us
roughly 20$ a day to query it for our test run summary table which powers
test aggregation.

We want to move to partitioned tables which must be done delicately.
Firstly however we need a time column to partition on, which was not
present on TestRows previously.

I've added the four new columns already to the table, they are empty for
all past rows.

I also have a query ready that builds a new partitioned table with all
the current data.

The plan is to:

- merge this change and start updating these columns when we upload
- when the test run uploader is NOT running:
  - build the new table with a different name
  - rename TestRuns to TestRuns_OLD
  - rename new table to TestRuns

All columns will then be populated, and new incoming rows will as well.

[TRT-1104](https://issues.redhat.com//browse/TRT-1104)

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_alert.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_alert.go
@@ -1,6 +1,8 @@
 package jobrunaggregatorapi
 
-import "time"
+import (
+	"cloud.google.com/go/bigquery"
+)
 
 const (
 	AlertsTableName = "Alerts"
@@ -70,10 +72,10 @@ type AlertRow struct {
 	Namespace       string
 	Level           string
 	AlertSeconds    int
-	JobName         string
+	JobName         bigquery.NullString
 	JobRunName      string
-	JobRunStartTime time.Time
-	JobRunEndTime   time.Time
-	Cluster         string
-	ReleaseTag      string
+	JobRunStartTime bigquery.NullTimestamp
+	JobRunEndTime   bigquery.NullTimestamp
+	Cluster         bigquery.NullString
+	ReleaseTag      bigquery.NullString
 }
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_alert.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_alert.go
@@ -1,5 +1,7 @@
 package jobrunaggregatorapi
 
+import "time"
+
 const (
 	AlertsTableName = "Alerts"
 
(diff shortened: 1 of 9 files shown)


dgoodwin:
/hold

Waiting till I'm in Monday to merge.

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_alert.go
```diff
@@ -64,9 +66,14 @@ INNER JOIN openshift-ci-data-analysis.ci_data.Jobs on JobRuns.JobName = Jobs.Job
...
+	Namespace       string
+	Level           string
+	AlertSeconds    int
+	JobName         string
+	JobRunName      string
+	JobRunStartTime time.Time
+	JobRunEndTime   time.Time
+	Cluster         string
+	ReleaseTag      string
 }
```
neisw, line 79: One thing that I have found when adding columns to existing tables is that they have to be Nulable if we are going to read existing values that don't have the new data.

https://pkg.go.dev/cloud.google.com/go/bigquery#InferSchema
https://stackoverflow.com/questions/51040470/exception-after-mapping-bigquerys-rows-with-null-to-struct

dgoodwin: Thanks, working on it.

dgoodwin: These schemas should all be deleted IMO, they're virtually all out of date and we can't kid ourselves into thinking we'll ever keep them up to date. They're unused, eternally out of date, buying us nothing but more work and confusion. If we ever need programatic schema management, we should do it properly, but today I think these should just be deleted in another PR personally.

DennisPeriquet: Looking at those, I have to agree; they are not used and if you know where the tables live in big query, that doc is kind of superfluous because it's that same info is right there in the schema section.  I also see we add some code just to make the linter happy.

dgoodwin: Ok the PR is updated with bigquery null types, tests are passing, and I've run an alert upload locally, you can now see values where (for example) cluster is populated in the db tables.

dgoodwin: hack/run-job-aggregator-analyzer.sh seems to run ok, that should be a good indicator of reading. A lot of those might use views though so it's hard to say how we'd expose this. 

Once we do the table swap there shouldn't be any more nulls, unless perhaps the source job run had nulls. 

I think we're ok to go.

PR #3516, opened 2023-06-27
Store MasterNodesUpdated and JobRunStatus with every result as well

dgoodwin:
Similar to yesterday, we're working to join less and deduplicate the job run tables entirely which all hold the same data


dgoodwin:
/hold 

So I can orchestrate rollout to avoid the jobs running.

PR #3517, opened 2023-06-27
TRT-1281: single command multiple uploaders 

neisw:
Updated to add an uploader registry that relies on a common JobRuns table.

Prior work done in [TRT-1120](https://issues.redhat.com//browse/TRT-1120) reduces GCS enumeration to a single pass

Still need to coordinate removing the existing alert and disruption loader and removing references to those jobs tables

Updated to reference [TRT-1281](https://issues.redhat.com//browse/TRT-1281).  Current plan is to get support in place with the uploader registry and work through reducing the commands down to a single process separately.

### DIFF
--- a/pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
+++ b/pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
@@ -100,15 +100,54 @@ func (f *BigQueryTestRunUploadFlags) ToOptions(ctx context.Context) (*allJobsLoa
 
 	var jobRunTableInserter jobrunaggregatorlib.BigQueryInserter
 	var testRunTableInserter jobrunaggregatorlib.BigQueryInserter
+
+	var backendAlertTableInserter jobrunaggregatorlib.BigQueryInserter
+	var backendDisruptionTableInserter jobrunaggregatorlib.BigQueryInserter
+
 	if !f.DryRun {
 		ciDataSet := bigQueryClient.Dataset(f.DataCoordinates.DataSetID)
 		jobRunTable := ciDataSet.Table(jobrunaggregatorapi.LegacyJobRunTableName)
 		testRunTable := ciDataSet.Table(jobrunaggregatorlib.TestRunTableName)
 		jobRunTableInserter = jobRunTable.Inserter()
 		testRunTableInserter = testRunTable.Inserter()
+
+		// could start with dry run for the new uploaders if we wanted
+		// backendAlertTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.AlertsTableName)
+		// backendDisruptionTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.BackendDisruptionTableName)
+
+		// backendAlertTable := ciDataSet.Table(jobrunaggregatorapi.AlertsTableName)
+		// backendAlertTableInserter = backendAlertTable.Inserter()
+		// backendDisruptionTable := ciDataSet.Table(jobrunaggregatorapi.BackendDisruptionTableName)
+		// backendDisruptionTableInserter = backendDisruptionTable.Inserter()
+
 	} else {
 		jobRunTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.LegacyJobRunTableName)
 		testRunTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorlib.TestRunTableName)
+
+		backendAlertTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.AlertsTableName)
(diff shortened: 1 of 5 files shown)


pkg/jobrunaggregator/jobrunbigqueryloader/uploader.go
```diff
@@ -248,6 +250,11 @@ func (o *jobRunLoaderOptions) uploadJobRun(ctx context.Context, jobRun jobrunagg
 	}
 	o.logger.Info("inserting job run row")
 
+	// get a regex from each jobRunUploader
+	// make one pass through the gcs artifacts for the jobrun
+	// return map[string][]string where the key is the regex and the value is a list of matching artifacts
+	// pass that list into the appropriate uploader so we don't have to scan over and over...
+
```
stbenjam, line 257: This is similar to what I was thinking of when I suggested a harvester interface.

WDYT about making the uploader interface exported and pushing off as much logic as possible to each implementation?  I think the design I used in [ci-test-mapping](https://github.com/openshift-eng/ci-test-mapping/blob/main/pkg/components/component.go#L21) is useful here, which has a [registry of components](https://github.com/openshift-eng/ci-test-mapping/blob/main/pkg/registry/registry.go#L187) and [each component](https://github.com/openshift-eng/ci-test-mapping/blob/main/pkg/components/clusterversionoperator/component.go) holds the logic to decide which test it owns and the various metadata it returns.

How many files are in a typical job run's folder? Would it be possible to pass the full list to each `Uploader` and then let it decide which ones (if any) to look at? A list of regexes may not be flexible enough - what if you're only interested in certain jobs, or some other condition that's in the job run info?

I think it might also be possible for some future uploader to need multiple files to collect the data it wants (although I am not entirely sure).

I was thinking of something like this:

design

neisw: I like that.  I don't know how many files we have per job run but I am a little worried about passing them all.  Wouldn't we be able to apply JobRun specific logic within the Uploader implementation whether we had all the files or just a subset?  With regex we might get duplicates between the regexes, with the full list we would have just the one but it might be big.  Latter is simpler, but means multiple passes over the full list...

stbenjam: It's fine either way, but I think you end up making the same amount of comparisons regardless if you're comparing against each uploader's regex, or passing in the list and letting the uploader decide what to do.   The benefit of the latter is a little more flexibility.

The upper limit of files in a job is about 10,000.  We're probably only going to have a handful of uploaders, but being generous say 20.  Iterating over the list of 10k 20x is less than 500ms, it's not significant.  In the regex  case you still probably have 20 sets of regexes right to know where to route the files?

neisw: I feel like it is more efficient to pass through the 10k file names once running the the 20 regexes and selecting out only the relevant files.  But then what if my regex says match everything and I dupe the 10k list 20 times or I have errors evaluating the regex...

I can see it would definitely be simpler to create one list of the 10k file names and pass that list into each Uploader for evaluation.  It would be fine with me to start with simple and then build in more logic if we see memory or processing impacts.

stbenjam: Just as a point of comparison, ci-test-mapping asks each of the ~200 components about each of the 20,000 tests and it finishes in less than a second with minimal memory usage.  It's really not that much :) 

But whatever you think works best for the uploaders

neisw: I created [TRT-1120](https://issues.redhat.com/browse/TRT-1120) for the gcs refactoring in ci-tools.  We can do that independent of any refactoring for uploaders.

pkg/jobrunaggregator/jobrunbigqueryloader/cmd.go
```diff
@@ -100,15 +100,54 @@ func (f *BigQueryTestRunUploadFlags) ToOptions(ctx context.Context) (*allJobsLoa
...
+
 	} else {
 		jobRunTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.LegacyJobRunTableName)
 		testRunTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorlib.TestRunTableName)
+
+		backendAlertTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.AlertsTableName)
+		backendDisruptionTableInserter = jobrunaggregatorlib.NewDryRunInserter(os.Stdout, jobrunaggregatorapi.BackendDisruptionTableName)
+	}
+
+	jobRunUploaderRegistry := JobRunUploaderRegistry{}
```
stbenjam, line 131: I see multiple instances of the registry, what's the reason for that instead of having one like this https://github.com/openshift-eng/ci-test-mapping/blob/main/pkg/registry/registry.go#L187 ?

stbenjam: Ah thanks, I didn't catch that they were different

dgoodwin (commented):
Just one nit but this looks great to me, very clever on the dry run usage.

PR #3518, opened 2023-06-28
Fix aggregated disruption tests being silently skipped

dgoodwin:
on ambiguous MasterNodesUpdated

This column was added to BackendDisruption yesterday, and now this query
is broken. To fix we specify the table to use, which is an alias defined
in both queries that call this function.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
@@ -347,7 +347,7 @@ func buildMasterNodesUpdatedSQL(masterNodesUpdated string) string {
 	masterNodesUpdatedSQL := ""
 
 	if len(masterNodesUpdated) > 0 {
-		masterNodesUpdatedSQL = fmt.Sprintf("AND MasterNodesUpdated = '%s'", masterNodesUpdated)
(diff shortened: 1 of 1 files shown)


pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -347,7 +347,7 @@ func buildMasterNodesUpdatedSQL(masterNodesUpdated string) string {
 	masterNodesUpdatedSQL := ""
 
 	if len(masterNodesUpdated) > 0 {
-		masterNodesUpdatedSQL = fmt.Sprintf("AND MasterNodesUpdated = '%s'", masterNodesUpdated)
+		masterNodesUpdatedSQL = fmt.Sprintf("AND JobRuns.MasterNodesUpdated = '%s'", masterNodesUpdated)
```
neisw, line 350: Future consideration, would we ever need to pass in the table name so it didn't always have to be JobRuns?

dgoodwin: I actually hope to kill all the JobRuns tables, so the column shouldn't be permanently ambiguous. We could also just not alias the one that was aliased, we're using the same table below technically.

PR #3521, opened 2023-06-29
Determine the type of index with the image in manifests

hongkailiu:
https://issues.redhat.com/browse/DPTP-3534

/cc @openshift/test-platform 
/assign @droslean

### DIFF
--- a/pkg/steps/index_generator.go
+++ b/pkg/steps/index_generator.go
@@ -42,15 +42,30 @@ func (*indexGeneratorStep) Validate() error { return nil }
 
 func databaseIndex(client ctrlruntimeclient.Client, name, namespace string) (bool, error) {
 	ist := &imagev1.ImageStreamTag{}
-	if err := client.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: namespace, Name: name}, ist); err != nil {
+	ctx := context.TODO()
+	if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: namespace, Name: name}, ist); err != nil {
 		return false, fmt.Errorf("could not fetch source ImageStreamTag: %w", err)
 	}
-	metadata := &docker10.DockerImage{}
-	if len(ist.Image.DockerImageMetadata.Raw) == 0 {
-		return false, fmt.Errorf("could not fetch Docker image metadata for ImageStreamTag %s", ist.Name)
+	dockerImageMetadata := ist.Image.DockerImageMetadata
+	for _, imageManifest := range ist.Image.DockerImageManifests {
+		// At the moment, we support only amd64
+		if imageManifest.Architecture == string(api.AMD64Arch) {
+			image := &imagev1.Image{}
+			// image is a cluster level CRD and thus no namespace should be provided to get the object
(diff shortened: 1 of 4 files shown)


hongkailiu:
/hold

Testing https://github.com/openshift/release/pull/40784

droslean:
@hongkailiu Can you elaborate more on why we need these changes?

pkg/steps/index_generator.go
```diff
@@ -42,15 +43,35 @@ func (*indexGeneratorStep) Validate() error { return nil }
...
 		return false, fmt.Errorf("could not fetch source ImageStreamTag: %w", err)
 	}
+	var dockerImageMetadata runtime.RawExtension
+	var foundInManifests bool
+	for _, imageManifest := range ist.Image.DockerImageManifests {
+		// At the moment, we support only amd64
+		if imageManifest.Architecture == string(api.AMD64Arch) {
+			image := &imagev1.Image{}
+			// image is a cluster level CRD and thus no namespace should be provided to get the object
+			if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Name: imageManifest.Digest}, image); err != nil {
```
droslean, line 57: This is being done automatically by Openshift.

hongkailiu:
> @hongkailiu Can you elaborate more on why we need these changes?

It is in the linked card of the PR description.

pkg/steps/index_generator.go
```diff
@@ -42,15 +43,35 @@ func (*indexGeneratorStep) Validate() error { return nil }
 
 func databaseIndex(client ctrlruntimeclient.Client, name, namespace string) (bool, error) {
 	ist := &imagev1.ImageStreamTag{}
-	if err := client.Get(context.TODO(), ctrlruntimeclient.ObjectKey{Namespace: namespace, Name: name}, ist); err != nil {
+	ctx := context.TODO()
+	if err := client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: namespace, Name: name}, ist); err != nil {
 		return false, fmt.Errorf("could not fetch source ImageStreamTag: %w", err)
 	}
+	var dockerImageMetadata runtime.RawExtension
+	var foundInManifests bool
```
droslean, line 51: No need to use the boolean. If the image doesn't have a manifest list, we will never end up iterating there.
You can just define:
`dockerImageMetadata := image.DockerImageMetadata` and then overwrite it in the iteration.

droslean (approved):

pkg/steps/index_generator.go
droslean: Adding more context here, I am not sure how exactly this is going to be used, but when we try to pull an image, we will get the manifest based on the architecture of the node. If that affects the idea, we can update it on a follow-up, if not, then ignore me.

hongkailiu: The isTag object is used to determine if the image underlying it is an db-based index (oo-testing has logic based on it) by checking the labels which was available in the isTag's metadata. However, after manifests are introduced, the metadata does not contain the labels any more. That is what we have to track it down into the image object now.
That is what we are doing in this PR.

PR #3522, opened 2023-06-29
Include registry in the target name

hongkailiu:
/cc @bbguimaraes

### DIFF
--- a/pkg/api/promotion.go
+++ b/pkg/api/promotion.go
@@ -108,4 +108,20 @@ var (
 			}
 		}
 	}
+
+	// DefaultTargetNameFunc is the default target name function
+	DefaultTargetNameFunc = func(registry string, config PromotionConfiguration) string {
+		if len(config.Name) > 0 {
+			return fmt.Sprintf("%s/%s/%s:${component}", registry, config.Namespace, config.Name)
+		}
+		return fmt.Sprintf("%s/%s/${component}:%s", registry, config.Namespace, config.Tag)
+	}
+
+	// QuayTargetNameFunc is the target name function for quay.io
+	QuayTargetNameFunc = func(_ string, config PromotionConfiguration) string {
+		if len(config.Name) > 0 {
+			return fmt.Sprintf("%s:%s_%s_${component}", QuayOpenShiftCIRepo, config.Namespace, config.Name)
+		}
+		return fmt.Sprintf("%s:%s_${component}_%s", QuayOpenShiftCIRepo, config.Namespace, config.Tag)
+	}
 )
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -363,12 +363,12 @@ func fromConfig(
 		if config.PromotionConfiguration == nil {
 			return nil, nil, fmt.Errorf("cannot promote images, no promotion configuration defined")
 		}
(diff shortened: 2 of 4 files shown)


pkg/steps/release/promote.go
```diff
@@ -43,11 +43,17 @@ type promotionStep struct {
...
 	if len(config.Name) > 0 {
-		return fmt.Sprintf("%s/%s:${component}", config.Namespace, config.Name)
+		if registry == api.QuayOpenShiftCIRepo {
+			return fmt.Sprintf("%s/%s_%s_${component}", registry, config.Namespace, config.Name)
+		}
+		return fmt.Sprintf("%s/%s/%s:${component}", registry, config.Namespace, config.Name)
+	}
+	if registry == api.QuayOpenShiftCIRepo {
+		return fmt.Sprintf("%s/%s_${component}_%s", registry, config.Namespace, config.Tag)
 	}
```
bbguimaraes, line 55: You know what I'm going to say about these `if`s =)  In general, "quay" appearing in this file is a layering violation.

We're also duplicating the logic already present in `pkg/api/promotion.go`.

hongkailiu: Let me try with another way.

pkg/api/promotion.go
```diff
@@ -109,3 +109,17 @@ var (
...
+		if registry == QuayOpenShiftCIRepo {
+			return fmt.Sprintf("%s/%s_%s_${component}", registry, config.Namespace, config.Name)
+		}
+		return fmt.Sprintf("%s/%s/%s:${component}", registry, config.Namespace, config.Name)
+	}
+	if registry == QuayOpenShiftCIRepo {
+		return fmt.Sprintf("%s/%s_${component}_%s", registry, config.Namespace, config.Tag)
+	}
+	return fmt.Sprintf("%s/%s/${component}:%s", registry, config.Namespace, config.Tag)
+}
```
bbguimaraes, line 125: We know in `defaults.go` which registry is used by each step, but loose that information when calling this function.  There is no need for these `if registry == …` conditionals.

pkg/defaults/defaults.go
```diff
@@ -363,18 +363,26 @@ func fromConfig(
...
 		} else {
-			postSteps = append(postSteps, releasesteps.PromotionStep(api.PromotionQuayStepName, config, requiredNames, jobSpec, podClient, pushSecret, api.QuayOpenShiftCIRepo, api.QuayMirrorFunc))
+			postSteps = append(postSteps, releasesteps.PromotionStep(api.PromotionQuayStepName, config, requiredNames, jobSpec, podClient, pushSecret, api.QuayOpenShiftCIRepo, targetName(api.QuayOpenShiftCIRepo, *config.PromotionConfiguration, ":", "_", "_"), api.QuayMirrorFunc))
 		}
 	}
 
 	return append(overridableSteps, buildSteps...), postSteps, nil
 }
 
+func targetName(registry string, config api.PromotionConfiguration, d1, d2, d3 string) string {
```
hongkailiu, line 379: d1, d2, d3 are ugly.
But I do not find another way provided you do not want me to do the following check

```go
if registry==... {
}
```

bbguimaraes: Just make it two functions.

PR #3523, opened 2023-06-29
replace more jobruns use

dgoodwin:
- Update historical disruption query
- Replace remaining use of BigQuery JobRuns tables


dgoodwin:
/hold

I'm about to go on PTO, not sure we want to merge this today.

PR #3525, opened 2023-07-03
`agents`: initialization failures do not result in an error

Prucek:
This PR was supposed to solve failures during the initialization and runtime of the `ci-operator-configresolver`. Instead, it was figured out, that the issue is mostly connected to the `agents` and touches also other services (`dptp-conotroller-manager`). In this PR is implemented channeled error forwarding for the `ConfigAgent` and `RegistryAgent`.

### DIFF
--- a/cmd/ci-operator-checkconfig/main.go
+++ b/cmd/ci-operator-checkconfig/main.go
@@ -49,7 +49,7 @@ func (o *options) parse() error {
 	if err := o.loadResolver(registryDir); err != nil {
 		return fmt.Errorf("failed to load registry: %w", err)
 	}
-	ciOPConfigAgent, err := agents.NewConfigAgent(o.ConfigDir, agents.WithOrg(o.Org), agents.WithRepo(o.Repo))
+	ciOPConfigAgent, err := agents.NewConfigAgent(o.ConfigDir, nil, agents.WithOrg(o.Org), agents.WithRepo(o.Repo))
 	if err != nil {
 		return fmt.Errorf("failed to create CI Op config agent: %w", err)
 	}
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -182,12 +182,14 @@ func main() {
 	if err != nil {
 		logrus.Fatalf("Failed to get config agent: %v", err)
 	}
+	go func() { logrus.Fatal(<-configErrCh) }()
 
 	registryErrCh := make(chan error)
 	registryAgent, err := agents.NewRegistryAgent(o.registryPath, registryErrCh, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry), registryAgentOption)
 	if err != nil {
 		logrus.Fatalf("Failed to get registry agent: %v", err)
 	}
+	go func() { logrus.Fatal(<-registryErrCh) }()
 
 	if o.validateOnly {
 		os.Exit(0)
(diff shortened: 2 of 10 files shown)


cmd/ci-operator-configresolver/main.go
```diff
@@ -177,16 +177,34 @@ func main() {
...
+		}
+	}()
+
+	registryErrCh := make(chan error)
+	registryAgent, err := agents.NewRegistryAgent(o.registryPath, registryErrCh, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry), registryAgentOption)
 	if err != nil {
 		logrus.Fatalf("Failed to get registry agent: %v", err)
 	}
 
+	go func() {
```
bbguimaraes, line 200: Same here.

cmd/ci-operator-configresolver/main.go
```diff
@@ -177,16 +177,34 @@ func main() {
...
 	}
 
-	registryAgent, err := agents.NewRegistryAgent(o.registryPath, agents.WithRegistryMetrics(configresolverMetrics.ErrorRate), agents.WithRegistryFlat(o.flatRegistry), registryAgentOption)
+	go func() {
+		err, ok := <-configErrCh
+		if ok {
+			close(configErrCh)
+			logrus.Fatalf("%v", err)
+		}
+	}()
```
bbguimaraes, line 192:
```suggestion
	go func() { logrus.Fatal(<-configErrCh) }()
```

bbguimaraes: No need to close the channel if we're exiting.

pkg/load/agents/utils.go
```diff
@@ -14,14 +14,17 @@ import (
...
 	var watchers []func(context.Context)
 
 	if universalSymlinkWatcher != nil {
 		watcher := func(ctx context.Context) {
 			errFunc := func(err error, msg string) {
 				recordErrorForMetric(metric, msg)
 				logrus.WithError(err).Error(msg)
+				go func() {
+					errCh <- err
+				}()
```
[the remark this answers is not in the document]
bbguimaraes: We'll also need to propagate errors from the other watchers (the `else` branch).  But it's fine if we implement just the universal watcher part in this pull request.

For now, we can just pass `nil` as the channel parameter everywhere we don't use this implementation, instead of creating unused channels.

cmd/ci-operator-checkconfig/main.go
```diff
@@ -49,7 +49,8 @@ func (o *options) parse() error {
 	if err := o.loadResolver(registryDir); err != nil {
 		return fmt.Errorf("failed to load registry: %w", err)
 	}
-	ciOPConfigAgent, err := agents.NewConfigAgent(o.ConfigDir, agents.WithOrg(o.Org), agents.WithRepo(o.Repo))
+	// TODO: add error handling
```
bbguimaraes, line 52: This program only loads the configuration once, we don't care if it's changed later.

Since the initial loading already validates that configuration files are valid, we don't need to do anything else here.  This `TODO` can be removed.

cmd/payload-testing-prow-plugin/filetestresolver_test.go
```diff
@@ -78,7 +78,8 @@ func TestFileTestResolver(t *testing.T) {
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			configAgent, err := agents.NewConfigAgent(tc.dir)
+			// TODO: add error handling
```
bbguimaraes, line 81: No need for this in a test file.

bbguimaraes (commented):
Looks good, and https://github.com/openshift/ci-tools/pull/3535 took care of the E2E failures.

We still need to add similar error handling to `dptp-controller-manager`, which is also using the universal watcher.  After that, this can be merged.

PR #3532, opened 2023-07-06
TRT-1120: Support single pass to collect gcs files

neisw:
- Simplifies finding / loading prowjob.json by looking specifically for the file and not a full scan.
- Delays the full scan until specifically called or artifacts are requested
- Caches artifact names so we don't have to perform multiple scans for different resources


neisw:
/hold

Hold for reviews

dgoodwin:
This looks good to me, but it's definitely in the ballpark of the type of changes that will only break once they go live. No conflicts on anything I'm working on.

/hold

Feel free to remove if Ken has a look and you're ok with it going in prior to PTO. If not we can wait till next week.

PR #3534, opened 2023-07-07
Add possibility to dely some of required jobs by marking them as always_run: false

jmguzik:


pkg/prowgen/prowgen.go
```diff
@@ -193,14 +193,18 @@ func generatePresubmitForTest(jobBaseBuilder *prowJobBaseBuilder, name string, i
 
 	shortName := info.TestName(name)
 	base := jobBaseBuilder.Rehearsable(!opts.disableRehearsal).Build(jc.PresubmitPrefix)
+	rerunCommand := prowconfig.DefaultRerunCommandFor(shortName)
+	if opts.defaultDisable {
```
smg247, line 197: don't we also want to verify that the job isn't `optional` here?

pkg/prowgen/prowgen.go
```diff
@@ -193,14 +193,18 @@ func generatePresubmitForTest(jobBaseBuilder *prowJobBaseBuilder, name string, i
 
 	shortName := info.TestName(name)
 	base := jobBaseBuilder.Rehearsable(!opts.disableRehearsal).Build(jc.PresubmitPrefix)
+	rerunCommand := prowconfig.DefaultRerunCommandFor(shortName)
```
smg247, line 196: I'm on the fence about this change, but you could also put the logic inside `prowconfig.DefaultRerunCommandFor`. This is the only usage.

jmguzik: Unfortunately `prowconfig.DefaultRerunCommandFor` is upstream

smg247: ah, it is. makes sense then.

jmguzik:
Interesting, after closing and reopening, old test result for e2e is qualified as pass while others are retested with the always run setting to false. It seems to be a little buggy and misleading solution.

pkg/prowgen/prowgen.go
jmguzik: conditions updated

jmguzik:
breaking changes can be fixed by periodic

PR #3536, opened 2023-07-12
Test: do not merge

jmguzik:


jmguzik:
@smg247 it seems unfortunately /retest-required does not work. It also seems that at least here these tests are not listed as required, which may be a potential bug.

smg247:
> @smg247 it seems unfortunately /retest-required does not work. 

Thats unfortunate.

> It also seems that at least here these tests are not listed as required, which may be a potential bug.

This is a known issue. GitHub won't mark the check as required if it isn't listed as `always_run: true`. Maybe there is some other way around it now, but there was nothing that could be done to change that previously. This will likely just be a downside to our solution here.

PR #3550, opened 2023-07-26
OKD-141 : Generate sippy config with the OKD jobs

bharath-b-rh:
PR is for updating the sippy-config-generator tool to include the OKD jobs along with existing OCP jobs in the generated sippy config. OKD jobs will added under a different release to segregate from the OCP jobs and the release name used for OKD jobs will be of the form `<release>-okd`.


stbenjam:
@bharath-b-rh FYI these ended up in the sippy config but they're not showing up in the DB.  I'll looking at it here https://issues.redhat.com/browse/TRT-1173

PR #3554, opened 2023-07-31
allow for promotion by tag in `openshift-priv` mirrored repos

smg247:
Auto-config-brancher has errors like the following when promotion by tag is configured:
```
time="2023-07-31T08:01:59Z" level=error msg="Failed to load CI Operator configuration" error="invalid ci-operator config: invalid configuration: promotion: both name and tag defined" source-file=ci-operator/config/openshift-priv/lvm-operator/openshift-priv-lvm-operator-main.yaml
```

We need to add the `-priv` suffix to the `tag` name when configured this way and **not** the `name` itself. We also should not mirror the `tag_by_commit` value as it is not useful and will result in many unnecessary images.

For: https://issues.redhat.com/browse/DPTP-3583


jmguzik:
/test e2e
as my pipeline is not yet working

droslean (approved):

PR #3560, opened 2023-08-02
`observer pods`: propagate the entrypoint's child exit code

Prucek:
This PR tries to solve [DPTP-3533](https://issues.redhat.com//browse/DPTP-3533). Using the new option from upstream [#30169](https://github.com/kubernetes/test-infra/pull/30169), we can propagate the exit code from observers to entrypoint and remove confusion to people when they analyze their job status.

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -286,7 +286,9 @@ func addPodUtils(
 	blobStorageOptions.SubDir = artifactDir
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, decorate.PlaceEntrypoint(decorationConfig, toolsMount))
 
-	wrapperOptions, err := decorate.InjectEntrypoint(&pod.Spec.Containers[0], decorationConfig.Timeout.Get(), decorationConfig.GracePeriod.Get(), "", "", false, false, logMount, toolsMount)
+	wrapperOptions, err := decorate.InjectEntrypoint(&pod.Spec.Containers[0],
+		decorationConfig.Timeout.Get(), decorationConfig.GracePeriod.Get(), "", "",
+		generatePodOptions.PropagateExitCode, false, logMount, toolsMount)
 	if err != nil {
 		return fmt.Errorf("could not inject entrypoint: %w", err)
 	}
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -271,7 +271,15 @@ func removeFile(podClient kubernetes.PodClient, ns, name, containerName string, 
 	return nil
 }
 
-func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.DecorationConfig, rawJobSpec string, secretsToCensor []coreapi.VolumeMount, clone bool, jobSpec *api.JobSpec) error {
+func addPodUtils(
+	pod *coreapi.Pod,
+	artifactDir string,
+	decorationConfig *prowv1.DecorationConfig,
+	rawJobSpec string,
+	secretsToCensor []coreapi.VolumeMount,
+	generatePodOptions *GeneratePodOptions,
+	jobSpec *api.JobSpec,
+) error {
 	logMount, logVolume := decorate.LogMountAndVolume()
 	toolsMount, toolsVolume := decorate.ToolsMountAndVolume()
 	blobStorageVolumes, blobStorageMounts, blobStorageOptions := decorate.BlobStorageOptions(*decorationConfig, false)
@@ -293,7 +301,7 @@ func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.
 	pod.Spec.Volumes = append(pod.Spec.Volumes, logVolume, toolsVolume)
 	pod.Spec.Volumes = append(pod.Spec.Volumes, blobStorageVolumes...)
 
-	if clone {
+	if generatePodOptions.Clone {
 		// Unless build_root.from_repository: true is set, the decorationConfig the ci-operator pod gets has cloning
 		// disabled.
 		decorationConfig := *decorationConfig
--- a/pkg/steps/artifacts_test.go
+++ b/pkg/steps/artifacts_test.go
@@ -661,7 +661,7 @@ func TestAddPodUtils(t *testing.T) {
 		},
 		GCSCredentialsSecret: func() *string { s := "gce-sa-credentials-gcs-publisher"; return &s }(),
 	}, "rawspec", []coreapi.VolumeMount{{Name: "secret", MountPath: "/secret"}},
-		&GeneratePodOptions{Clone: false}, nil); err != nil {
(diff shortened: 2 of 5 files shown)


pkg/steps/artifacts.go
```diff
@@ -271,14 +271,18 @@ func removeFile(podClient kubernetes.PodClient, ns, name, containerName string,
...
-func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.DecorationConfig, rawJobSpec string, secretsToCensor []coreapi.VolumeMount, clone bool, jobSpec *api.JobSpec) error {
+func addPodUtils(pod *coreapi.Pod, artifactDir string, decorationConfig *prowv1.DecorationConfig, rawJobSpec string, secretsToCensor []coreapi.VolumeMount, clone bool, jobSpec *api.JobSpec, genPodOpts *GeneratePodOptions) error {
 	logMount, logVolume := decorate.LogMountAndVolume()
 	toolsMount, toolsVolume := decorate.ToolsMountAndVolume()
 	blobStorageVolumes, blobStorageMounts, blobStorageOptions := decorate.BlobStorageOptions(*decorationConfig, false)
 	blobStorageOptions.SubDir = artifactDir
 	pod.Spec.InitContainers = append(pod.Spec.InitContainers, decorate.PlaceEntrypoint(decorationConfig, toolsMount))
 
-	wrapperOptions, err := decorate.InjectEntrypoint(&pod.Spec.Containers[0], decorationConfig.Timeout.Get(), decorationConfig.GracePeriod.Get(), "", "", false, false, logMount, toolsMount)
+	propagateWrappedError := false
```
danilo-gemoli, line 281:
```golang
var propagateWrappedError bool
```
looks better

pkg/steps/artifacts_test.go
```diff
@@ -660,7 +660,7 @@ func TestAddPodUtils(t *testing.T) {
 			PathStrategy: prowv1.PathStrategyExplicit,
```
danilo-gemoli, line 660: Can you introduce a new unit test to make sure the `entrypoint` has the proper argument set when it runs an observer?

danilo-gemoli (changes requested):

pkg/steps/artifacts.go
Prucek: I wanted to make sure false is there, if not set, but now I see Go has the default value false

pkg/steps/pod.go
```diff
@@ -55,6 +55,16 @@ type PodStepConfiguration struct {
...
+type GeneratePodOptions struct {
+	IsObserver bool
+}
+
+func DefaultGeneratePodOptions() *GeneratePodOptions {
+	return &GeneratePodOptions{
+		IsObserver: false,
+	}
+}
+
```
bbguimaraes, line 67: Isn't this functionally the same as passing `nil` to the functions?

Prucek: I see your point, and you are right. I moved it to `steps` because I need the information on whether the pod is an observer in `addPodUtils`. And this was the only structure that does that. If I had left it inside `multi_stage` I would have a cyclic dependency between those two packages. So what do you suggest? Create a new struct? I can't think of a good option.

bbguimaraes: It's fine here, but the option should be about the effect it has on the generated pod.

- `steps/pod.go`: provides the option to propagate the error, knows nothing about observers
- `steps/multi_stage/multi_stage.go`: knows it is generating an observer and requests that option

Prucek: Now it is more clear to me, but I cannot just rename the option, because the `pkg/steps/multi_stage/gen.go` still uses the `IsObserver` in `isKubeconfigNeeded()` and `addSecretWrapper()`

Prucek:
btw I updated a lot of deprecated functions and libraries, hope it's OK. My IDE just warned me everywhere and since I am working in the `steps` package and most of them were from there I thought it is a good idea.

bbguimaraes:
> btw I updated a lot of deprecated functions and libraries, hope it's OK. My IDE just warned me everywhere and since I am working in the `steps` package and most of them were from there I thought it is a good idea.

That's good, but please separate it into its own pull request.

pkg/steps/pod.go
bbguimaraes: I'd still rather have flags as parameters than `bool`s, especially since there is one already (`clone`).  But it should present the options that the function provides:

```go
type GeneratePodOptions struct {
    Clone             bool
    PropagateExitCode bool
}
```

The other functions can then set these according to their own internal criteria (`isObserver`, etc.).

Prucek: Okay, but I still think the `IsObserver` param is needed there. 
`func isKubeconfigNeeded(step *api.LiteralTestStep, opts *generatePodOptions) bool {
	needsKubeconfig := step.NoKubeconfig == nil || !*step.NoKubeconfig
	return needsKubeconfig || opts.IsObserver
}` 
For example here, the kubeconfig should be added only if it is an observer, with only `Clone` and `PropagateExitCode` present, it would become unclear.

bbguimaraes: It is needed, but it should not be in `pkg/steps`, since it's internal to `pkg/steps/multi_stage`.  We can translate from the private `generatePodOptions` to the public `GeneratePodOptions` inside the multi-stage module.

Prucek: That is a good idea! I hope I understood you correctly and addressed it with the latest changes.

pkg/steps/pod.go
```diff
@@ -285,7 +290,10 @@ func (s *podStep) generatePodForStep(image string, containerResources coreapi.Re
 	}
 
 	artifactDir := s.name
-	pod, err := GenerateBasePod(s.jobSpec, s.config.Labels, s.config.As, s.config.NodeName, s.name, []string{"/bin/bash", "-c", "#!/bin/bash\nset -eu\n" + s.config.Commands}, image, containerResources, artifactDir, s.jobSpec.DecorationConfig, s.jobSpec.RawSpec(), secretVolumeMounts, clone)
+	pod, err := GenerateBasePod(s.jobSpec, s.config.Labels, s.config.As,
+		s.config.NodeName, s.name, []string{"/bin/bash", "-c", "#!/bin/bash\nset -eu\n" + s.config.Commands},
+		image, containerResources, artifactDir, s.jobSpec.DecorationConfig, s.jobSpec.RawSpec(),
+		secretVolumeMounts, &GeneratePodOptions{Clone: clone, PropagateExitCode: false})
```
bbguimaraes, line 296: A minor note: here and in other places, `false` is the zero value of `bool`, so statements such as `PropagateExitCode: false` could have been omitted.

bbguimaraes (commented):
/lgtm

pkg/steps/pod.go
Prucek: Okay, I am coming from C world and I am more used to explicitly expressing, not used to that. I will try to remember!

PR #3563, opened 2023-08-02
Add a new tool: ci-images-mirror

hongkailiu:
This PR initiates the work for https://issues.redhat.com/browse/DPTP-3529
It basically introduces a new controller running in a new component as
- the current dptp-cm is too big already.
- it is useful only during the transition period, i.e., when we replace the app.ci's registry with the quay.io repo, this tool will be offline too.

The controller is doing nothing now. I can build images and make iterations with it.

/cc @openshift/test-platform @jupierce


cmd/ci-images-mirror/main.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func newOpts() (*options, error) {
+	opts := &options{}
+	opts.addDefaults()
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&opts.leaderElectionNamespace, "leader-election-namespace", "ci", "The namespace to use for leaderelection")
+	fs.StringVar(&opts.leaderElectionSuffix, "leader-election-suffix", "", "Suffix for the leader election lock. Useful for local testing. If set, --dry-run must be set as well")
+	fs.Var(&opts.enabledControllers, "enable-controller", fmt.Sprintf("Enabled controllers. Available controllers are: %v. Can be specified multiple times. Defaults to %v", allControllers.UnsortedList(), opts.enabledControllers.Strings()))
+	fs.BoolVar(&opts.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
```
smg247, line 54: I wouldn't expect `dry-run` to have `true` as a default

cmd/ci-images-mirror/main.go
```diff
@@ -0,0 +1,122 @@
...
+	fs.Var(&opts.quayIOCIImagesDistributorOptions.additionalImageStreamNamespacesRaw, "quayIOCIImagesDistributorOptions.additional-image-stream-namespace", "A namespace in which imagestreams will be distributed even if no test explicitly references them (e.G `ci`). Can be passed multiple times.")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		logrus.WithError(err).Fatal("could not parse args")
+	}
+
+	var errs []error
+	if opts.leaderElectionNamespace == "" {
+		errs = append(errs, errors.New("--leader-election-namespace must be set"))
+	}
```
smg247, line 63: This check doesn't seem necessary, there is already a default. It would have to be explicitly unset to get here.

cmd/ci-images-mirror/main.go
```diff
@@ -0,0 +1,122 @@
...
+	additionalImageStreamNamespacesRaw flagutil.Strings
+	additionalImageStreamNamespaces    sets.Set[string]
+}
+
+func newOpts() (*options, error) {
+	opts := &options{}
+	opts.addDefaults()
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&opts.leaderElectionNamespace, "leader-election-namespace", "ci", "The namespace to use for leaderelection")
+	fs.StringVar(&opts.leaderElectionSuffix, "leader-election-suffix", "", "Suffix for the leader election lock. Useful for local testing. If set, --dry-run must be set as well")
```
smg247, line 52: you mention `dry-run` must be set as well, but that isn't validated anywhere

cmd/ci-images-mirror/main.go
```diff
@@ -0,0 +1,122 @@
...
+	var errs []error
+	if opts.leaderElectionNamespace == "" {
+		errs = append(errs, errors.New("--leader-election-namespace must be set"))
+	}
+	if values := opts.enabledControllers.Strings(); len(values) > 0 {
+		opts.enabledControllersSet = sets.New[string](values...)
+		if diff := opts.enabledControllersSet.Difference(allControllers); diff.Len() > 0 {
+			errs = append(errs, fmt.Errorf("the following controllers are unknown: %v", diff.UnsortedList()))
+		}
+	}
```
smg247, line 69: Can we move all the validations into a separate function?

smg247: Also, what is the goal with this one? it seems like it is just checking that the single controller is enabled, right?

pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
```diff
@@ -0,0 +1,92 @@
...
+
+	controllerutil "github.com/openshift/ci-tools/pkg/controller/util"
+	"github.com/openshift/ci-tools/pkg/util/imagestreamtagmapper"
+	"github.com/openshift/ci-tools/pkg/util/imagestreamtagwrapper"
+)
+
+const ControllerName = "quay_io_ci_images_distributor"
+
+func AddToManager(manager manager.Manager,
+	additionalImageStreamNamespaces sets.Set[string]) error {
```
smg247, line 28: This can be one line

cmd/ci-images-mirror/main.go
hongkailiu: Done!

cmd/ci-images-mirror/main.go
hongkailiu: This would avoid explicit unsetting. But yes, I agree that it is unlikely to happen.

cmd/ci-images-mirror/main.go
hongkailiu: > Can we move all the validations into a separate function?

Done

hongkailiu: > Also, what is the goal with this one? it seems like it is just checking that the single controller is enabled, right?

It ensures that every enabled controller is known, i.e., predefined in the tool.
It is unlikely that there are more controllers to add into this tool but it is not harm to keep it here just in case.

smg247: I may be misunderstanding the `Difference` method. I think it would return results if there were controllers from `allControllers` that weren't in `enabledControllersSet` as well. Anyways, it isn't worth worrying about here, especially considering the lifecycle of this tool.

hongkailiu:
/override ci/prow/e2e

The fix to e2e is https://github.com/openshift/ci-tools/pull/3465 and this PR introducing a new tool that should not impact any e2e.

PR #3569, opened 2023-08-05
Initial implementation of the multi-arch builder controller

droslean:
- Introduces the `MultiArchBuildConfig` Custom Resource.

example:
```yaml
apiVersion: ci.openshift.io/v1
kind: MultiArchBuildConfig
metadata:
  name: test-mabc
  namespace: ci
spec:
  build_spec:
    output:
      to:
        kind: ImageStreamTag
        name: managed-clonerefs:latest-test
        namespace: ci
    resources:
      requests:
        cpu: 100m
        memory: 200Mi
    runPolicy: Serial
    source:
      dockerfile: |-
        FROM gcr.io/k8s-prow/git:latest
        ADD ./clonerefs /clonerefs
      images:
      - from:
          kind: ImageStreamTag
          name: clonerefs:latest
        paths:
        - destinationDir: .
          sourcePath: /ko-app/clonerefs
      type: Dockerfile
    strategy:
      dockerStrategy:
        imageOptimizationPolicy: SkipLayers
        noCache: true
      type: Docker

```

- Implements the `multi-arch-builder-controller`
The controller reconciles the `MultiArchBuildConfig` and generates multiple builds one for each architecture that exists on the cluster. Once all builds succeed, the controller uses the `manifest-tool` binary to create a new image based on the output configuration that includes the manifest list with all images that have been built per architecture correspondingly.

/cc @openshift/test-platform 
/cc @bbguimaraes @jmguzik @danilo-gemoli @deepsm007  


pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,350 @@
...
+)
+
+const (
+	controllerName        = "multiarchbuildconfig"
+	nodeArchitectureLabel = "kubernetes.io/arch"
+	noReasonCondition     = "NoReason"
+
+	// TODO: Now we hardcode the URL for the heterogeneous cluster.
+	// TODO: We need to push it to the app.ci registry or quay.io.
+	registryURL = "registry.multi-build01.arm-build.devcluster.openshift.com"
```
droslean, line 41: @hongkailiu Can we start pushing the images to quay.io or we should push them to `app.ci` registry in order to make sure that the test-image-distributor will sync the image across all clusters?

cmd/multi-arch-builder-controller/main.go
```diff
@@ -0,0 +1,111 @@
...
+	if err := mgr.Start(ctx); err != nil {
+		logrus.WithError(err).Fatal("Manager ended with error")
+	}
+
+	logrus.Info("Process ended gracefully")
+}
+
+func resolveNodeArchitectures(ctx context.Context, client coreclientset.NodeInterface) ([]string, error) {
+	ret := sets.New[string]()
+	nodeList, err := client.List(ctx, metav1.ListOptions{})
```
droslean, line 101: @deepsm007 This approach won't ensure to include an architecture from a scaled-down to 0 machineset. When we will start adding more architectures to the heterogeneous cluster, we will need to make sure that at least one node is running or change this approach to gather the architectures from the machinessets which is really hard now.

deepsm007: We need to add that in comment inside the code and keep track of it when we move from more than 2 archs

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,349 @@
...
+	w, err := r.buildClient.Builds(mabc.Namespace).Watch(ctx, metav1.ListOptions{
+		LabelSelector: labels.Set{v1.MultiArchBuildConfigNameLabel: mabc.Name}.String(),
+	})
+	if err != nil {
+		return err
+	}
+
+Loop:
+	for {
+		event, ok := <-w.ResultChan()
```
droslean, line 163: @bbguimaraes Do you think that it's better to create another controller that watches/reconciles the Builds instead of running a watcher here?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,361 @@
...
+
+	cmdOutput := &bytes.Buffer{}
+	cmdError := &bytes.Buffer{}
+	cmd.Stdout = cmdOutput
+	cmd.Stderr = cmdError
+
+	logrus.Infof("Running command: %s", cmd.String())
+	err := cmd.Run()
+	if err != nil {
+		logrus.WithError(err).WithField("output", cmdOutput.String()).WithField("error_output", cmdError.String()).Info("manifest-tool command failed...")
```
jmguzik, line 355:
```suggestion
		logrus.WithError(err).WithField("output", cmdOutput.String()).WithField("error_output", cmdError.String()).Info("manifest-tool command failed")
```

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,361 @@
...
+		return err
+	}
+
+	logrus.WithField("output", cmdOutput.String()).Info("manifest-tool command succeeded...")
```
jmguzik, line 359:
```suggestion
	logrus.WithField("output", cmdOutput.String()).Info("manifest-tool command succeeded")
```
... suggests that there is something more to it

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
jmguzik: Afaik @hongkailiu is syncing images to quay so you would need to treat still app.ci as the source of truth.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+
+	if err := r.handleMultiArchBuildConfig(ctx, mabc); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (r *reconciler) handleMultiArchBuildConfig(ctx context.Context, mabc *v1.MultiArchBuildConfig) error {
```
jmguzik, line 122: This function is not executed in any of tests.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,361 @@
...
+		}
+
+		if checkAllBuildsFinished(builds) {
+			r.logger.Info("All builds have finished...")
+			return nil
+		}
+	}
+}
+
+func (r *reconciler) createBuildsForArchitectures(ctx context.Context, mabc *v1.MultiArchBuildConfig) (map[string]*buildv1.Build, error) {
```
jmguzik, line 201: As a consequence, this one is also not covered.

jmguzik (approved):
I did a high level review of the code. Generally LGTM for me, found really small nits and left comments about some parts of functionality not covered in tests at all. As this is a new tool, not messing with implementation of any currently used tools we have I give a green stamp with hold if somebody else wants to check as well.
/hold

cmd/multi-arch-builder-controller/main.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if err := multiarchbuildconfigv1.AddToScheme(mgr.GetScheme()); err != nil {
+		logrus.WithError(err).Fatal("Failed to add multiarchbuildconfig to scheme")
+	}
+
+	if err := buildv1.AddToScheme(mgr.GetScheme()); err != nil {
+		logrus.WithError(err).Fatal("Failed to add multiarchbuildconfig to scheme")
+	}
+
+	nodeArchitectures, err := resolveNodeArchitectures(ctx, kubeClient.Nodes())
```
bbguimaraes, line 83: Do we want to do this only once at initialization time or also keep track of nodes as they enter and leave the cluster?

pkg/api/multiarchbuildconfig/v1/types.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type MultiArchBuildConfigState string
+
+const (
+	// SuccessState means all builds were completed without error (exit 0)
+	SuccessState MultiArchBuildConfigState = "success"
+	// FailureState means that all builds were completed with errors (exit non-zero)
+	FailureState MultiArchBuildConfigState = "failure"
+)
```
bbguimaraes, line 51: What if none of these are true?

pkg/controller/multiarchbuildconfig/buildclient.go
bbguimaraes, line 1: Why are these necessary?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: I will add tests for this as well.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,361 @@
...
+func (r *reconciler) handleMultiArchBuildConfig(ctx context.Context, mabc *v1.MultiArchBuildConfig) error {
+	targetImageRef := fmt.Sprintf("%s/%s", mabc.Spec.BuildSpec.CommonSpec.Output.To.Namespace, mabc.Spec.BuildSpec.CommonSpec.Output.To.Name)
+	r.logger.WithField("target_image", targetImageRef).Info("Start handling builds to generate image...")
+
+	builds, err := r.createBuildsForArchitectures(ctx, mabc)
+	if err != nil {
+		return fmt.Errorf("couldn't create builds for architectures: %s: %w", strings.Join(r.architectures, ","), err)
+	}
+
+	if err := r.watchAndHandleBuildEvents(ctx, mabc, builds); err != nil {
```
bbguimaraes, line 131: This will only give any indication of the progress when all builds are done, which I don't think will be appreciated by people watching their test results

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
bbguimaraes: Why not do it in a reconciliation loop?  See also the previous comment.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -0,0 +1,361 @@
...
+		}
+		return nil
+	})
+}
+
+func (r *reconciler) pruneBuilds(ctx context.Context, builds map[string]*buildv1.Build) error {
+	for _, build := range builds {
+		if err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			b := &buildv1.Build{}
+			if err := r.client.Get(ctx, types.NamespacedName{Namespace: build.Namespace, Name: build.Name}, b); err != nil {
```
bbguimaraes, line 253: Why is this `Get` necessary?

pkg/api/multiarchbuildconfig/v1/types.go
droslean: Now we assume that the resources are not yet processed. I think we should also add another state that indicates that the resource is in a Processing state.

cmd/multi-arch-builder-controller/main.go
droslean: Keeping track is a good idea but for now, let's do it at the initialization time. I am investigating also how to utilize the machine sets in order to keep track of the architectures in real-time. I am planning some follow-ups on that front.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: The error of each of the builds is being recorded in the resource conditions as well.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: To avoid conflicts. The resource could change in the meantime by another process that is not related to this controller. This is an idiomatic way to manage the resources in a kube/openshift cluster.

pkg/controller/multiarchbuildconfig/buildclient.go
droslean: For testing mostly. I need to write an adapter for the build client to use it in the tests.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: Yes, but all of the main operations inside are tested.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: We can add the builds with the multiarchbuildconfig label to be owned by the manager as well. We could potentially trigger the reconciliation process or write another controller for them. That will make things little bit more complicated though. I will dig into it.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
bbguimaraes: But only after all of them are done, since the resource is only modified after this point.  This is why I think we should treat `Build`s the same way we treat `MultiArchBuildConfig`s.  It's similar to what the payload plugin does with `ProwJob`s.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
bbguimaraes: But what are we gaining if we're just doing an arbitrary `Get` right before deleting it?  What conflict does this prevent?

If the intent is to delete the specific version created previously, shouldn't we use the values from `builds` when deleting?  Or do it this way:

https://github.com/openshift/ci-tools/blob/2dee011d0b7a2aa21cdf6dce77775a3c9bf4ace2/pkg/util/pods.go#L91

bbguimaraes: In any case,`ci-operator` doesn't delete `Build`s when they are done, we just let the namespace expire (our TTL is very short these days).  This is very useful if someone needs to investigate when something goes wrong.

droslean: Yea, the whole get/delete thing was to avoid conflicts if the build was mutated in the window between the get and the delete. The delete in that case will fail. However, I would agree that we shouldn't clean them up. I will check your suggestions. Definitely, the users should be able to investigate their build if it fails.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: Yea, when I wrote payload plugin controller, I though that it would be better to use another controller to watch prowjobs. However, I was hoping that the builds in this case, should be more trivial operation, and get away with it with just a watcher. I would prefer the controller though. I will make the changes.

pkg/api/multiarchbuildconfig/v1/types.go
```diff
@@ -0,0 +1,77 @@
...
+func UpdateMultiArchBuildConfig(ctx context.Context, logger *logrus.Entry, client ctrlruntimeclient.Client, namespacedName types.NamespacedName, mutateFn func(mabcToMutate *MultiArchBuildConfig)) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		multiArchBuildConfig := &MultiArchBuildConfig{}
+		if err := client.Get(ctx, namespacedName, multiArchBuildConfig); err != nil {
+			return fmt.Errorf("failed to get the MultiArchBuildConfig: %w", err)
+		}
+
+		mutateFn(multiArchBuildConfig)
+
+		logger.WithField("namespace", namespacedName.Namespace).WithField("name", namespacedName.Name).Info("Updating MultiArchBuildConfig...")
```
jmguzik, line 71:
```suggestion
		logger.WithField("namespace", namespacedName.Namespace).WithField("name", namespacedName.Name).Info("Updated MultiArchBuildConfig")
```

jmguzik (approved):
One small nit, but no reason to stop merge. I have no further remakrs.
/unhold

PR #3578, opened 2023-08-16
TRT-932: Adding alert query for known alerts

deepsm007:
Updating ci-tools BigQuery query to include first time alert nurp was observed and updates the query_results.json

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
@@ -2,6 +2,7 @@ package jobrunaggregatorapi
 
 import (
 	"fmt"
+	"time"
 
 	"cloud.google.com/go/bigquery"
 )
@@ -33,6 +34,8 @@ type AlertHistoricalDataRow struct {
 	AlertName      string
 	AlertNamespace string
 	AlertLevel     string
+	FirstObserved  time.Time
+	LastObserved   time.Time
 	HistoricalJobData
 	P95 string
 	P99 string
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
@@ -52,7 +52,7 @@ type JobLister interface {
 
 type HistoricalDataClient interface {
 	ListDisruptionHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalData, error)
-	ListAlertHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalData, error)
+	ListAlertHistoricalData(ctx context.Context) ([]*jobrunaggregatorapi.AlertHistoricalDataRow, error)
 }
 
 type CIDataClient interface {
@@ -168,7 +168,7 @@ ORDER BY
 	return jobrunaggregatorapi.ConvertToHistoricalData(disruptionDataSet), nil
 }
 
-func (c *ciDataClient) ListAlertHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalData, error) {
+func (c *ciDataClient) ListAlertHistoricalData(ctx context.Context) ([]*jobrunaggregatorapi.AlertHistoricalDataRow, error) {
 	queryString := c.dataCoordinates.SubstituteDataSetLocation(`
     SELECT AlertName, AlertNamespace, AlertLevel,
             Release, FromRelease, Platform, Architecture, Network, Topology,
@@ -211,7 +211,7 @@ func (c *ciDataClient) ListAlertHistoricalData(ctx context.Context) ([]jobrunagg
 		}
 		alertDataSet = append(alertDataSet, data)
 	}
-	return jobrunaggregatorapi.ConvertToHistoricalData(alertDataSet), nil
+	return alertDataSet, nil
 }
 
 func (c *ciDataClient) ListAllJobs(ctx context.Context) ([]jobrunaggregatorapi.JobRow, error) {
@@ -978,9 +978,9 @@ func GetUTCDay(in time.Time) time.Time {
 
 func (c *ciDataClient) ListAllKnownAlerts(ctx context.Context) ([]*jobrunaggregatorapi.KnownAlertRow, error) {
 	queryString := c.dataCoordinates.SubstituteDataSetLocation(
-		`SELECT *
-FROM DATA_SET_LOCATION.Alerts_AllKnown
-ORDER BY Release, AlertName, AlertNamespace ASC
+		`SELECT AlertName, AlertNamespace, AlertLevel, Release, FirstObserved, LastObserved, Results
+	FROM DATA_SET_LOCATION.Alerts_AllKnown
+	ORDER BY Release, AlertName, AlertNamespace, FirstObserved ASC
 `)
 
 	query := c.client.Query(queryString)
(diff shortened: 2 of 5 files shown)


pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
```diff
@@ -17,6 +18,13 @@ type HistoricalData interface {
 	GetJobRuns() int
 }
 
+type HistoricalDataWithSetters interface {
```
neisw, line 21: I don't think we need this

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
```diff
@@ -33,7 +41,10 @@ type AlertHistoricalDataRow struct {
 	AlertName      string
 	AlertNamespace string
 	AlertLevel     string
+	FirstObserved  time.Time
```
neisw, line 44: There were only needed for alerts so let's move there

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -52,7 +52,7 @@ type JobLister interface {
 
 type HistoricalDataClient interface {
 	ListDisruptionHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalData, error)
-	ListAlertHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalData, error)
+	ListAlertHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalDataWithSetters, error)
```
neisw, line 55: How about returning  []*jobrunaggregatorapi.AlertHistoricalDataRow

pkg/jobrunaggregator/jobrunaggregatorlib/ci_data_client.go
```diff
@@ -168,7 +168,7 @@ ORDER BY
 	return jobrunaggregatorapi.ConvertToHistoricalData(disruptionDataSet), nil
 }
 
-func (c *ciDataClient) ListAlertHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalData, error) {
+func (c *ciDataClient) ListAlertHistoricalData(ctx context.Context) ([]jobrunaggregatorapi.HistoricalDataWithSetters, error) {
```
neisw, line 171: Don't call jobrunaggregatorapi.ConvertToHistoricalData(alertDataSet), nil below just return  []*jobrunaggregatorapi.AlertHistoricalDataRow when done

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -91,6 +91,30 @@ func (o *JobRunHistoricalDataAnalyzerOptions) Run(ctx context.Context) error {
 	return nil
 }
 
+func (o *JobRunHistoricalDataAnalyzerOptions) getAlertData(ctx context.Context) []jobrunaggregatorapi.HistoricalData {
+	var allKnownAlerts []*jobrunaggregatorapi.KnownAlertRow
+	var newHistoricalData []jobrunaggregatorapi.HistoricalDataWithSetters
+
+	newHistoricalData, err := o.ciDataClient.ListAlertHistoricalData(ctx)
```
neisw, line 98: This should now be  []*jobrunaggregatorapi.AlertHistoricalDataRow and you can use it in your innner loop below

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -91,6 +91,30 @@ func (o *JobRunHistoricalDataAnalyzerOptions) Run(ctx context.Context) error {
...
+		return nil
+	}
+	allKnownAlerts, err = o.ciDataClient.ListAllKnownAlerts(ctx)
+	if err != nil {
+		return nil
+	}
+	for _, alerts := range allKnownAlerts {
+		for _, jobData := range newHistoricalData {
+			if alerts.AlertName == jobData.GetName() {
+				jobData.SetFirstObserved(alerts.FirstObserved)
```
neisw, line 109: you don't need the setters if you are using AlertHistoricalDataRow

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -91,6 +91,30 @@ func (o *JobRunHistoricalDataAnalyzerOptions) Run(ctx context.Context) error {
...
+				jobData.SetLastObserved(alerts.LastObserved)
```
neisw, line 110: break out of the inner loop if you found the one you want

pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go
```diff
@@ -91,6 +91,30 @@ func (o *JobRunHistoricalDataAnalyzerOptions) Run(ctx context.Context) error {
...
+			}
+		}
+	}
+
+	return newHistoricalData
```
neisw, line 115: return jobrunaggregatorapi.ConvertToHistoricalData(newHistoricalData),

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
deepsm007: they are in alerts

DennisPeriquet:
I see this in the query_resuilts.json uploaded to the Jira:

```
...
  "FirstObserved": "0001-01-01T00:00:00Z",
  "LastObserved": "0001-01-01T00:00:00Z",
...
```

DennisPeriquet:
the date looks like 0s
/hold

deepsm007:
Updated the code now it’s generating correct data in the file, new file is in the jira

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
```diff
@@ -44,7 +47,7 @@ func (a *AlertHistoricalDataRow) GetJobData() HistoricalJobData {
 	return a.HistoricalJobData
 }
 func (a *AlertHistoricalDataRow) GetName() string {
-	return fmt.Sprintf("%s %s %s", a.AlertName, a.AlertNamespace, a.AlertLevel)
+	return a.AlertName
```
neisw, line 50: Seems like we are loosing context here.  So if an alert with the same name but different namespace exists we can no longer tell the difference based on the name.  Looks like this is only used in a markdown table currently did you really need this change?

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
```diff
@@ -72,6 +75,8 @@ func (a *AlertHistoricalDataRow) GetKey() string {
 		a.Platform,
 		a.Network,
 		a.Topology,
+		a.FirstObserved,
+		a.LastObserved,
```
neisw, line 79: Do we want these in the keys?  LastObserved will be changing over time, if the keys in the maps that are generated change won't that break https://github.com/deepsm007/ci-tools/blob/33f9a92cb9fbd66e3759f084dbe7ef0f5074c57a/pkg/jobrunaggregator/jobrunhistoricaldataanalyzer/analyzer.go#L162

Have you tried running the update with an existing file and validated things still match up?

pkg/jobrunaggregator/jobrunaggregatorapi/types_row_historical_data.go
deepsm007: reverted back to original

PR #3580, opened 2023-08-23
simplified_pipeline_ctrl to support 2-staged pipelines

jmguzik:


cmd/simplified-pipeline-ctrl/reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+	for _, presubmit := range presubmits.alwaysRequired {
+		if pjob, ok := latestBatch[presubmit]; ok && pjob.Status.State != v1.SuccessState {
+			return false, nil
+		}
+	}
+	for _, presubmit := range presubmits.conditionallyRequired {
+		if pjob, ok := latestBatch[presubmit]; ok && pjob.Status.State != v1.SuccessState {
+			return false, nil
+		}
+	}
```
bbguimaraes, line 149: This distinction between always/conditionally required doesn't seem to be used anywhere?

bbguimaraes: If we merge them, naming them something to the effect of "first stage" and "second stage" would make it easier to understand the categorization here and in the provider.

cmd/simplified-pipeline-ctrl/reconciler.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return nil
+}
+
+func composeKey(refs *v1.Refs) string {
+	return fmt.Sprintf("%d:%s/%s@%s", refs.Pulls[0].Number, refs.Org, refs.Repo, refs.BaseRef)
+}
+
+func (r *reconciler) reportSuccessOnPR(ctx context.Context, pj *v1.ProwJob, presubmits presubmitTests) (bool, error) {
```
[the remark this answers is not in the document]
bbguimaraes: Or maybe make it clearer that this advances to the "second stage" of the pipeline, "success" is a bit ambiguous.

cmd/simplified-pipeline-ctrl/reconciler.go
jmguzik: There was a distinction but I removed it to clear logs. Thanks for the catch.

jmguzik:
\o/

jmguzik:
BTW this code is working on app.ci so we would see if it catches basic scenario o_0

PR #3581, opened 2023-08-23
TRT-1204: Support static list of JobRuns

neisw:
Allows a list of JobRunIdentifier objects to be passed in for analysis bypassing bigquery lookup to reduces costs and allow external tooling to designate runs to be analyzed.

JobRunIdentifier can be passed in via the command to specify a static list of identifiers for analysis.  Additionally JobRunIdentifier is now set when using BigQueryJobRunWaiter or ClusterJobRunWaiter to skip additional / unnecessary calls to bigquery for JobRun data.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
@@ -2,6 +2,7 @@ package jobrunaggregatorlib
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -44,11 +45,31 @@ type JobRunIdentifier struct {
 	JobRunID string
 }
 
+func GetStaticJobRunInfo(staticRunInfoJSON, staticRunInfoPath string) ([]JobRunIdentifier, error) {
+	var jsonBytes []byte
+	var jobRuns []JobRunIdentifier
+	var err error
+	if len(staticRunInfoJSON) == 0 {
+		jsonBytes, err = os.ReadFile(staticRunInfoPath)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		jsonBytes = []byte(staticRunInfoJSON)
+	}
+
+	if err = json.Unmarshal(jsonBytes, &jobRuns); err != nil {
+		return nil, err
+	}
+
+	return jobRuns, nil
+}
+
 type JobRunGetter interface {
 	// GetRelatedJobRuns gets all related job runs for analysis
 	GetRelatedJobRuns(ctx context.Context) ([]jobrunaggregatorapi.JobRunInfo, error)
 
-	// GetRelatedJobRunsFromIdentifiers passes along minimal information known about the jobs already so we can skip
+	// GetRelatedJobRunsFromIdentifiers passes along minimal information known about the jobs already so that we can skip
 	// querying and go directly to fetching the full job details when GetRelatedJobRuns is called
 	GetRelatedJobRunsFromIdentifiers(ctx context.Context, jobRunIdentifiers []JobRunIdentifier) ([]jobrunaggregatorapi.JobRunInfo, error)
 }
--- a/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
+++ b/pkg/jobrunaggregator/jobrunaggregatorlib/util.go
@@ -18,6 +18,7 @@ import (
 	prowv1 "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowjobclientset "k8s.io/test-infra/prow/client/clientset/versioned"
 	prowjobinformers "k8s.io/test-infra/prow/client/informers/externalversions"
+	v1 "k8s.io/test-infra/prow/client/informers/externalversions/prowjobs/v1"
 	"k8s.io/utils/clock"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
@@ -38,14 +39,23 @@ var (
 	KnownQuerySources = sets.Set[string]{JobStateQuerySourceBigQuery: sets.Empty{}, JobStateQuerySourceCluster: sets.Empty{}}
 )
 
+type JobRunIdentifier struct {
+	JobName  string
+	JobRunID string
+}
+
 type JobRunGetter interface {
 	// GetRelatedJobRuns gets all related job runs for analysis
 	GetRelatedJobRuns(ctx context.Context) ([]jobrunaggregatorapi.JobRunInfo, error)
+
+	// SetRelatedJobRuns sets minimal information known about the jobs already so we can skip
+	// querying and go directly to fetching the full job details when GetRelatedJobRuns is called
+	SetRelatedJobRuns([]JobRunIdentifier)
 }
 
 type JobRunWaiter interface {
 	// Wait waits until all job runs finish, or time out
-	Wait(ctx context.Context) error
+	Wait(ctx context.Context) ([]JobRunIdentifier, error)
 }
 
 // WaitUntilTime waits until readAt time has passed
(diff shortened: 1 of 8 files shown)


pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -157,32 +183,49 @@ type ClusterJobRunWaiter struct {
...
+func (w *ClusterJobRunWaiter) checkMatchedJobsForCompletion(prowJobInformer v1.ProwJobInformer) (bool, map[string]*prowv1.ProwJob, error) {
+	allItems, err := prowJobInformer.Lister().List(labels.Everything())
+	if err != nil {
+		logrus.Infof("Error listing prow jobs: %v", err)
+		// this is a change from the prior code, confirm if that was intentional
+		// if err != nil {
+		//				logrus.Infof("Error listing prow jobs: %v", err)
+		//				return false, nil
+		//			}
+		return false, nil, err
```
neisw, line 219: @xueqzhan, I wanted to double check on this.  In the original code the error was logged but not passed back up through PollUntilContextTimeout.  If that was intentional I can change this to match.

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
```diff
@@ -457,10 +465,57 @@ func (o *JobRunTestCaseAnalyzerOptions) findJobRunsWithRetry(ctx context.Context
...
+
 // GetRelatedJobRuns gets all related job runs for analysis
 func (o *JobRunTestCaseAnalyzerOptions) GetRelatedJobRuns(ctx context.Context) ([]jobrunaggregatorapi.JobRunInfo, error) {
 	var jobRunsToReturn []jobrunaggregatorapi.JobRunInfo
-	jobs, err := o.jobGetter.GetJobs(ctx)
+	var jobs []jobrunaggregatorapi.JobRow
+	var err error
+
+	// allow for the list of ids to be passed in
+	if len(o.staticJobRunIdentifiers) > 0 {
```
xueqzhan, line 513: You already have the job names and job ids. It is probably better to create a common library for both analyzers to GetRelatedJobRunsFromStaticIdentifiers.

pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go
```diff
@@ -134,6 +138,11 @@ func (f *JobRunsTestCaseAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
 	fs.StringArrayVar(&f.ExcludeJobNames, "exclude-job-names", f.ExcludeJobNames, "Applied only when --explicit-gcs-prefixes is not specified.  The flag can be specified multiple times to create a list of substrings used to filter JobNames from the analysis")
 	fs.StringArrayVar(&f.IncludeJobNames, "include-job-names", f.IncludeJobNames, "Applied only when --explicit-gcs-prefixes is not specified.  The flag can be specified multiple times to create a list of substrings to include in matching JobNames for analysis")
 	fs.StringVar(&f.JobStateQuerySource, "query-source", jobrunaggregatorlib.JobStateQuerySourceBigQuery, "The source from which job states are found. It is either bigquery or cluster")
+
+	// optional for local use or potentially gangway results
+	fs.StringVar(&f.StaticJobRunIdentifierPath, "staticRunInfoPath", f.StaticJobRunIdentifierPath, "The optional path to a file containing JSON formatted JobRunIdentifier array used for aggregated analysis")
```
xueqzhan, line 143: Should not use camel cases here for command prompts for consistent purposes.

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -208,22 +251,29 @@ func (w *ClusterJobRunWaiter) Wait(ctx context.Context) error {
...
 		},
 	)
 	if err != nil && err != context.DeadlineExceeded {
-		return fmt.Errorf("failed waiting for prowjobs to complete: %w", err)
+		return nil, fmt.Errorf("failed waiting for prowjobs to complete: %w", err)
 	}
-	return nil
+
+	// one more time to get the matched jobs
+	_, matchedJobs, err := w.checkMatchedJobsForCompletion(prowJobInformer)
```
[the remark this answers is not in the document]
neisw: ConditionWithContextFunc is: type ConditionWithContextFunc func(context.Context) (done bool, err error)

So we don't have a way to return the data back up from the call.  We could try to do something funky and set the data in the  ConditionWithContextFunc call to something outside but to me just doing the additional query is cleaner.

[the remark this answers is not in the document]
neisw: Ah yes, now I remember how I wound up looking at this...

pkg/jobrunaggregator/jobrunaggregatorlib/util.go:210:132: (*ClusterJobRunWaiter).checkMatchedJobsForCompletion - result 2 (error) is always nil (unparam)
func (w *ClusterJobRunWaiter) checkMatchedJobsForCompletion(prowJobInformer v1.ProwJobInformer) (bool, map[string]*prowv1.ProwJob, error) {

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -157,32 +183,45 @@ type ClusterJobRunWaiter struct {
...
+	logrus.Infof("%d/%d jobs completed, waiting for: [%v]", len(matchedJobMap)-len(uncompletedJobMap), len(matchedJobMap), strings.Join(sets.StringKeySet(uncompletedJobMap).List(), ", "))
+	return false, matchedJobMap
+}
+
+func (w *ClusterJobRunWaiter) checkMatchedJobsForCompletion(prowJobInformer v1.ProwJobInformer) (bool, map[string]*prowv1.ProwJob, error) {
+	allItems, err := prowJobInformer.Lister().List(labels.Everything())
+	if err != nil {
+		// log and suppress the error
+		logrus.Infof("Error listing prow jobs: %v", err)
+		return false, nil, nil
```
neisw, line 215: Ah yes, now I remember how I wound up looking at this...

pkg/jobrunaggregator/jobrunaggregatorlib/util.go:210:132: (*ClusterJobRunWaiter).checkMatchedJobsForCompletion - result 2 (error) is always nil (unparam)
func (w *ClusterJobRunWaiter) checkMatchedJobsForCompletion(prowJobInformer v1.ProwJobInformer) (bool, map[string]*prowv1.ProwJob, error) {

pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go
neisw: GetRelatedJobRuns is returning []jobrunaggregatorapi.JobRunInfo which we need jobLocators to fetch the JobRunInfo and those are constructed differently for jobrunaggregatoranalyzer and jobruntestcaseanalyzer.

These changes also allow us to pass in the list of jobs via command line, potentially for aggregating results from tools like gangway, so where we are reusing that code now we would have to duplicate some of it for the util.GetRelatedJobRunsFromStaticIdentifiers and pass in joblocator(s).

I see what you are saying but I think the implementation wouldn't be straightforward and might lead to even more duplicated code.

xueqzhan: I was thinking of skipping the locator and matchers. You already have the names and the corresponding IDs, we can do a straight read (maybe using ReadJobRunFromGCS, common for both analyzers) instead of gcs search (from startingJobRunID to endingJobRunID). That should further reduce our gcs usage. 

But we can also leave that for future enhancement.

neisw: I believe that is what it is, effectively, doing.  Just via the locator.

https://github.com/neisw/ci-tools/blob/e1d979b851238f356771617eaf1cd45e376df057/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go#L81

and 

https://github.com/neisw/ci-tools/blob/e1d979b851238f356771617eaf1cd45e376df057/pkg/jobrunaggregator/jobruntestcaseanalyzer/analyzer.go#L475

Call 

https://github.com/neisw/ci-tools/blob/e1d979b851238f356771617eaf1cd45e376df057/pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator.go#L88

neisw: I could rename it from FindJob to ReadJob or something if that makes it clearer.  I think I went with FindJob since the existing api was FindRelatedJobs but I'm not partial to the api name.

xueqzhan: Yes, as of now, this is used on the per job level. You still had to fit your code into the existing framework, like for test case analyzer, you had to get the list of job names before going to the per job level. But you already have the list of job names and the their corresponding job run IDs in both analyzers. If you have a common function to get related job runs from jobNames and their corresponding Run IDs, you could call that function and return from here, without using the logic that is unique to that particular analyzer. Let's sync on a huddle if that is easier.

neisw: Reviewed with Ken, the request was to see if we could create a generic locator that could handle multiple jobs.  The locator requires a prefix that is often generic in the form of  "logs/" + JobName.

However in the case of PR jobs it is nondeterministic and needs to be passed in 
`fs.Var(&jobGCSPrefixSlice{&f.JobGCSPrefixes}, "explicit-gcs-prefixes", "a list of gcs prefixes for jobs created for   payload. Only used by per PR payload promotion jobs. The format is comma-separated elements, each consisting of job name and gcs prefix separated by =, like openshift-machine-config-operator=3028-ci-4.11-e2e-aws-ovn-upgrade~logs/openshift-machine-config-operator-3028-ci-4.11-e2e-aws-ovn-upgrade")`

After review we decided it is not worth the effort to attempt the refactor currently.

PR #3582, opened 2023-08-24
Add prow job wait mechanism using informers from CI cluster

xueqzhan:
[TRT-1099](https://issues.redhat.com/browse/TRT-1099)

This PR adds a wait mechanism for prow job completion based on prow job CR in the CI cluster. Traditionally, bigquery DB is queried in a loop while waiting for jobs to finish. It is discovered that has cost lots of money when aggregators are run. Instead of querying bigquery, this PR uses informers and waits for related prow jobs to finish. Once the wait is over, original mechanism is used to fetch junit data.


pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -112,15 +132,110 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
...
+			continue
+		}
+
+		jobRunID := prowJob.Labels[prowJobJobRunIDLabel]
+		uncompletedJobMap[jobRunID] = prowJob
+	}
+	if len(uncompletedJobMap) == 0 {
+		return true
+	}
+	logrus.Infof("Current time %+v and next lgo time: [%v]", time.Now(), w.nextLogTime)
```
dgoodwin, line 167: Logrus includes the current time with your message. 

Coming into this loop every few minutes, I don't see why we wouldn't log every time we came through, it won't hurt anything and it helps us always see what the code is doing. I would simplify and ditch this concept.

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -112,15 +132,110 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
...
+			continue
+		}
+		if prowJob.Status.CompletionTime != nil {
+			continue
+		}
+
+		jobRunID := prowJob.Labels[prowJobJobRunIDLabel]
+		uncompletedJobMap[jobRunID] = prowJob
+	}
+	if len(uncompletedJobMap) == 0 {
```
dgoodwin, line 164: logrus.Info("all jobs completed")

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -112,15 +132,110 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
...
+		return true
+	}
+	logrus.Infof("Current time %+v and next lgo time: [%v]", time.Now(), w.nextLogTime)
+	if time.Now().After(w.nextLogTime) {
+		logrus.Infof("The following jobs are not complete: [%v]", strings.Join(sets.StringKeySet(uncompletedJobMap).List(), ", "))
```
dgoodwin, line 169: Please log # of completed jobs as well as not complete just in case there's ever a mismatch. Something like:

"x/y jobs completed, waiting for: [%v]"

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -112,15 +132,110 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
...
+}
+
+type ClusterJobRunWaiter struct {
+	ProwJobClient      *prowjobclientset.Clientset
+	TimeToStopWaiting  time.Time
+	ProwJobMatcherFunc ProwJobMatcherFunc
+	nextLogTime        time.Time
+}
+
+func (w *ClusterJobRunWaiter) allProwJobFinished(allItems []*prowv1.ProwJob) bool {
```
dgoodwin, line 151: Nit: allProwJobsFinished

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
```diff
@@ -112,15 +132,110 @@ func WaitAndGetAllFinishedJobRuns(ctx context.Context,
...
+	logrus.Infof("Going to wait until %+v with timeout value %+v", w.TimeToStopWaiting, timeout)
+
+	// wait for up to limit until we've finished
+	err := wait.PollUntilContextTimeout(
+		ctx,
+		5*time.Minute,
+		timeout,
+		true,
+		func(ctx context.Context) (bool, error) {
+			allItems, err := prowJobInformer.Lister().List(labels.Everything())
```
dgoodwin, line 203: What was your test process like? I'm curious if our aggregated pods have permissions to list all ProwJobs, and if not, how do we get it.

dgoodwin (commented):
Request for PRs especially of this complexity to include a good message, explain the feature, how it works, why it was selected and is being done. I was reasonably familiar with this but it always helps reviewers with dense PRs.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
```diff
@@ -1,19 +1,30 @@
...
+	// ProwJobAggregationIDLabel is the name of the label for the aggregation id in prow job
+	ProwJobAggregationIDLabel = "release.openshift.io/aggregation-id"
+	// ProwJobPayloadInvocationIDLabel is the name of the label for the payload invocation id in prow job
+	ProwJobPayloadInvocationIDLabel = "release.openshift.io/aggregation-id"
+	// prowJobReleaseJobNameAnnotation refers to the original periodic job name for PR based payload runs
+	prowJobReleaseJobNameAnnotation = "releaseJobName"
 )
 
+func NewProwJobMatcherFuncForPR(jobName, matchID, matchLabel string) ProwJobMatcherFunc {
+	return perPRProwJobMatcher{
```
stevekuznetsov, line 21: Could this just be a closure?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
```diff
@@ -51,6 +54,7 @@ func (f *JobRunsAnalyzerFlags) BindFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&f.ExplicitGCSPrefix, "explicit-gcs-prefix", f.ExplicitGCSPrefix, "only used by per PR payload promotion jobs.  This overrides the well-known mapping and becomes the required prefix for the GCS query")
 	fs.DurationVar(&f.Timeout, "timeout", f.Timeout, "Time to wait for aggregation to complete.")
 	fs.StringVar(&f.EstimatedJobStartTimeString, "job-start-time", f.EstimatedJobStartTimeString, fmt.Sprintf("Start time in RFC822Z: %s", kubeTimeSerializationLayout))
+	fs.StringVar(&f.JobStateQuerySource, "query-source", jobrunaggregatorlib.JobStateQuerySourceBigQuery, "The source from which job states are found. It is either bigquery or cluster")
```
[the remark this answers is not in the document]
xueqzhan: The two implementations are for different analyzers. One is for regular and PR aggregator, while the other is for test case analyzer (e.g. install). They are different enough to have different cmd.go.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
```diff
@@ -1,19 +1,30 @@
...
+		jobName:    jobName,
+		matchID:    matchID,
+		matchLabel: matchLabel,
+	}.shouldAggregateReleaseControllerJob
+}
```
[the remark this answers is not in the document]
xueqzhan: Yes, I thought of this at first. But it turned out to include more lower layer function refactoring that I am not ready to take with this PR.

pkg/jobrunaggregator/jobrunaggregatorlib/util.go
xueqzhan: This will be the next thing to figure out. In my test environment, just using my own token will be sufficient. I might need service account route for this.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
xueqzhan: @stevekuznetsov Can you elaborate on this? Thanks!

pkg/jobrunaggregator/jobrunaggregatoranalyzer/cmd.go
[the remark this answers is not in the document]
dgoodwin: I see, could you get that spelled out very clearly in the option and/or godoc.

Why does is test case analyzer not able to use a similar approach of watching prowjobs? I'm glad Payload and PR Payload both get the new though.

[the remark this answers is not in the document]
xueqzhan: I might have confused you. Test case analyzer IS supported in this PR. The jobrunaggregatoranalyzer/cmd.go is for both payload and PR aggregator since they both have very similar functionality. The jobruntestcaseanalyzer/cmd.go is for test case analyzer. The command options and the logic are different enough to warrant a separate implementation. 

In this case, both analyzers have the same command argument "query-source". Thus this is added in both cmd.go files.

pkg/jobrunaggregator/jobrunaggregatorlib/jobrun_locator_per_pr_jobs.go
stevekuznetsov: Right now you have:

```go
type perPRProwJobMatcher struct {
	// matchID is how we recognize per-PR payload jobs. It is set based on the matchLabel value
	// that the per-PR payload controller sets in the prowjobs it creates.
	matchID    string
	matchLabel string
}

func (a perPRProwJobMatcher) shouldAggregateReleaseControllerJob(prowJob *prowjobv1.ProwJob) bool {
	id := prowJob.Labels[a.matchLabel]
	jobName := prowJob.Labels["prow.k8s.io/job"]
	jobRunId := prowJob.Labels["prow.k8s.io/build-id"]
	fmt.Printf("  checking %v/%v for matchID match: looking for %q found %q.\n", jobName, jobRunId, a.matchID, id)
	idMatches := len(a.matchID) > 0 && id == a.matchID

	return idMatches
}
```


You could just as well have:

```go
func shouldAggregateReleaseControllerJob(matchId, matchLabel string) func(prowJob *prowjobv1.ProwJob) bool {
     return func(prowJob *prowjobv1.ProwJob) bool {
		id := prowJob.Labels[a.matchLabel]
		jobName := prowJob.Labels["prow.k8s.io/job"]
		jobRunId := prowJob.Labels["prow.k8s.io/build-id"]
		fmt.Printf("  checking %v/%v for matchID match: looking for %q found %q.\n", jobName, jobRunId, a.matchID, id)
		idMatches := len(a.matchID) > 0 && id == a.matchID

		return idMatches
	}
}
```

What is the point of the `perPRProwJobMatcher` type?

dgoodwin:
Will need to request testplatform review in their channel workflows.

jmguzik:
/hold

I need to better understand this change. 
Is it affecting TP tooling?

xueqzhan:
> /hold
> 
> I need to better understand this change. Is it affecting TP tooling?

No this should not affect TP tooling. This is an aggregator internal mechanism. Outside interface for aggregators stay the same.

jmguzik:
@xueqzhan yes, you can watch prowjobs in a similar way they are watched in prow component crier or in my WIP component called simplified-pipeline-ctrl

PR #3583, opened 2023-08-25
ci-images-mirror: reconcile on the tags of test input

hongkailiu:
The functions are copied (with a few modifications) from `test-images-distributor` and can be tested locally with the script `hack/run-ci-images-mirror.sh` from this PR.

I have not addressed the test coverage because
- most functions are tested already in `test-images-distributor`
- i plan to refactor to share the code between those 2 controllers AFTER PoC. If that does not happen, I will add tests by then.

/cc @openshift/test-platform 

https://issues.redhat.com/browse/DPTP-3529

### DIFF
--- a/cmd/ci-images-mirror/main.go
+++ b/cmd/ci-images-mirror/main.go
@@ -5,12 +5,13 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"path/filepath"
 
 	"github.com/sirupsen/logrus"
+	"gopkg.in/fsnotify.v1"
 
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
-	"k8s.io/client-go/rest"
 	"k8s.io/test-infra/prow/flagutil"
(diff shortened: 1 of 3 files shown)


cmd/ci-images-mirror/main.go
```diff
@@ -84,12 +96,41 @@ func main() {
...
+	registryErrCh := make(chan error)
+	registryAgentOption := func(opt *agents.RegistryAgentOptions) {
+		opt.UniversalSymlinkWatcher = universalSymlinkWatcher
+	}
+	stepConfigPath := filepath.Join(opts.releaseRepoGitSyncPath, config.RegistryPath)
+	registryConfigAgent, err := agents.NewRegistryAgent(stepConfigPath, registryErrCh, registryAgentOption)
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to construct registryAgent")
+	}
+	go func() { logrus.Fatal(<-registryErrCh) }()
```
droslean, line 132: I think it would be better to check both of the channels in the same go routine since the plan here is to exit on error. Otherwise, we need to make sure that the other go routine stops before exiting.

PR #3587, opened 2023-08-26
Include the manifest list on the output images on all builds

droslean:
This PR:

- Migrates the manifest pusher into its own package for generalized usage.
- Transform all ci-operator builds to be architecture-specific.  
- Start using the `manifest-tool` tool in ci-operator to generate the final image with the manifest list included.

I already did some testing with the `vega-project` images. See job https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-vega-project-ccb-operator-master-images/1695214240392024064

The final images that ended up in `apps.ci` cluster include the manifest list. 

/cc @openshift/test-platform @danilo-gemoli @jmguzik @deepsm007 @bbguimaraes 

### DIFF
--- a/pkg/manifestpusher/manifestpusher.go
+++ b/pkg/manifestpusher/manifestpusher.go
@@ -1,0 +1,107 @@
+package manifestpusher
+
+import (
+	"bytes"
+	"encoding/base64"
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/sirupsen/logrus"
+
+	buildv1 "github.com/openshift/api/build/v1"
+)
+
+const (
+	nodeArchitectureLabel = "kubernetes.io/arch"
+)
+
+type ManifestPusher interface {
+	PushImageWithManifest(builds map[string]*buildv1.Build, targetImageRef string) error
+}
+
+func NewManifestPushfer(logger *logrus.Entry, registryURL string, dockercfgPath string) ManifestPusher {
+	return &manifestPusher{
+		logger:        logger,
+		registryURL:   registryURL,
+		dockercfgPath: dockercfgPath,
+	}
+}
+
+type manifestPusher struct {
+	logger        *logrus.Entry
+	registryURL   string
+	dockercfgPath string
+}
+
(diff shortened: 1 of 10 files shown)


pkg/manifestpusher/manifestpusher.go
```diff
@@ -0,0 +1,107 @@
...
+		m.logger.WithError(err).WithField("output", cmdOutput.String()).WithField("error_output", cmdError.String()).Error("manifest-tool command failed")
+		return err
+	}
+	m.logger.WithField("output", cmdOutput.String()).Debug("manifest-tool command succeeded")
+
+	m.logger.Infof("Image %s created", targetImageRef)
+	return nil
+}
+
+func CreateDockerCfg(tokenPath, registryURL, dest string) error {
```
droslean, line 85: @hongkailiu I could use your insights on this one. The `manifest-tool` needs a dockercfg file to get permissions to inspect and push images into the local registry. To achieve that we could potentially mount an extra secret into all ci-operator jobs for each cluster correspondingly. However, the test pod is running under the `ci-operator`  service account and Openshift already mounts the `token` to `/var/run/secrets/kubernetes.io/serviceaccount/token`. My approach here is to construct a new `.dockercfg` file with that token included. This way we avoid configuring all jobs and pass extra secrets.

bbguimaraes:
I understood the new controller would be responsible for builds and for creating the final image with the manifest list.  Why is that being added to `ci-operator` now?

droslean:
> I understood the new controller would be responsible for builds and for creating the final image with the manifest list. Why is that being added to `ci-operator` now?

The controller will build the supplemental image. It is focusing on the images that are not being built by ci-operator.

The changes in this PR, will make sure that ci-operator builds the images with the manifest-list included.

However, to move forward with this PR, I need to get rid of the workaround of creating the dockercfg, and figure out what is happening with the e2e jobs.

droslean:
https://github.com/openshift/release/pull/43758
https://github.com/openshift/release/pull/43760
https://github.com/openshift/release/pull/43762

smg247 (commented):
/lgtm

PR #3590, opened 2023-08-29
OPRUN-1873: ci-operator: allow accessory promotions

stevekuznetsov:
There exist repositories which promote images to the OCP payload but would also like to be able to build utility images that are promoted to some other location, since inclusion into the core payload ImageStreams is tightly regulated.

The new `accesssory_promotion` stanza in `ci-operator` configuration allows users to opt images into other promotion schemes. This field is opt-in, so no changes are expected for any current configurations.

### DIFF
--- a/pkg/validation/config.go
+++ b/pkg/validation/config.go
@@ -154,13 +154,30 @@ func (v *Validator) validateConfiguration(ctx *configContext, config *api.Releas
 	// Validate promotion
 	if config.PromotionConfiguration != nil {
 		validationErrors = append(validationErrors,
-			validatePromotionConfiguration("promotion",
+			validatePromotionConfiguration(ctx.AddField("promotion"),
 				*config.PromotionConfiguration,
-				api.PromotesOfficialImages(config, api.WithOKD),
+				api.PromotesOfficialImages(config.PromotionConfiguration, api.WithOKD),
 				len(api.ImageTargets(config)) > 0,
-				api.ExtractPromotionNamespace(config),
+				api.ExtractPromotionNamespace(config.PromotionConfiguration),
 				config.ReleaseTagConfiguration,
 				config.Releases)...)
+	} else {
+		if len(config.AccessoryPromotionConfiguration) > 0 {
+			validationErrors = append(validationErrors, ctx.AddField("accessory_promotion").errorf("cannot specify accessory_promotion when main promotion is unset"))
+		}
+	}
+
+	for i, step := range config.AccessoryPromotionConfiguration {
+		validationErrors = append(validationErrors,
+			validatePromotionConfiguration(ctx.AddField("accessory_promotion").addIndex(i),
+				step,
+				api.PromotesOfficialImages(&step, api.WithOKD),
+				len(api.AccessoryImageTargets(&step)) > 0,
+				api.ExtractPromotionNamespace(&step),
+				config.ReleaseTagConfiguration,
+				config.Releases)...)
+		validateAccessoryPromotionConfiguration(ctx.AddField("accessory_promotion").addIndex(i),
+			step)
 	}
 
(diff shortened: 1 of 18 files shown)


stevekuznetsov:
Not really sure why e2e is not triggering and why bootstrapper-validation is failing ...

pkg/validation/config.go
```diff
@@ -395,33 +412,42 @@ var (
...
 	if promotesOfficialImages && imageTargets {
 		if _, ok := releases["latest"]; !ok && releaseTagConfiguration == nil {
-			validationErrors = append(validationErrors, fmt.Errorf("importing the release stream is required to ensure the promoted images to the namespace %s can be integrated properly. Although it can be achieved by tag_specification or releases[\"latest\"], adding an e2e test is strongly suggested", promotionNamespace))
+			validationErrors = append(validationErrors, ctx.errorf("importing the release stream is required to ensure the promoted images to the namespace %s can be integrated properly. Although it can be achieved by tag_specification or releases[\"latest\"], adding an e2e test is strongly suggested", promotionNamespace))
 		}
 	}
 	return validationErrors
 }
 
+func validateAccessoryPromotionConfiguration(ctx *configContext, input api.PromotionConfiguration) []error {
```
hongkailiu, line 442: Could we invalidate the cases where `AccessoryPromotionConfiguration` promotes to `ocp` or `okd` payload?
Those imagestreams have meaning for branching. Maybe excluding those cases to avoid questions such as "would it participate branching when it promotes to payload with `AccessoryPromotionConfiguration`"?
This should not be a really restriction for your cases the feature of this PR is all about promotion to other places than the payloads. ^_^

(If you agree, please include it in the GoDoc too. Thanks.)

Otherwise, the code looks good to me.

stevekuznetsov: Good call, done

hongkailiu:
/lgtm

/hold

Please give a day or two for others to take a look before canceling.
Thanks for the feature.

jmguzik:
@hongkailiu how it plays together with this OKD card that we discussed during last refinement? 
I am asking as preferably we do not want to end up having multiple unrelated configurations.

jmguzik:
So for the reference, the mentioned card is DPTP-3565. We will discuss the impact and asses conflicts before merge.

stevekuznetsov:
@jmguzik thanks for the cross-reference. It looks like the intent of the two things is different:

1. OKD wants the normal promotion flow (opt-out for images, assumes all built images are pushed) to go to two places
2. this PR wants an accessory promotion that's opt-in, has no bearing on release payloads, and is entirely separate from the main promotion

I'm happy to unify, but I am not certain I know exactly how. For my use-case, I want an image to be published *only* to the second stream.

danilo-gemoli:
With the following configuration:
```yaml
base_images:
  base:
    namespace: ci
    name: cucushift
    tag: latest
  golang:
    namespace: openshift
    name: release
    tag: rhel-9-release-golang-1.19-openshift-4.13
images:
  from: base
  to: verification-tests
promotion:
  namespace: ci
  name: whatever
  tag: latest
accesssory_promotion:
  namespace: ci
  name: whatever
  additional_images:
    latest: golang
```
are we going to promote both `pipeline:verification-tests` and `pipeline:golang` to `ci/whatever:latest` ?
Will the targeted registry race due to `oc image mirror ... --max-per-registry=20 ...` (see [here](https://github.com/openshift/ci-tools/blob/80e760661c2b495f7804426a625745e4c5bdf004/pkg/steps/release/promote.go#L215C32-L215C41)) ?
Not sure if it makes sense as I don't know the internals of either `oc` or the image registry (same for its backend storage) but I'm going to raise the question in any case as it (the registry) is being our major source of pain nowadays.

stevekuznetsov:
/close

Chatted with folks, taking a different approach to help OKD as well.

stevekuznetsov:
@danilo-gemoli we dedupe the list before using `oc image mirror`

PR #3602, opened 2023-09-05
OPRUN-1873: promotion: allow configuring additional targets

stevekuznetsov:


bbguimaraes (commented):
A couple of minor implementation details if we change that code in subsequent PRs.

/lgtm

PR #3606, opened 2023-09-07
Cluster profile changes

psalajova:
Implementation of [DPTP-3459](https://issues.redhat.com/browse/DPTP-3459), also taking into account the planned change to move cluster profiles into `openshift/release`. Depends on https://github.com/openshift/release/pull/43834


cmd/ci-operator-checkconfig/main.go
```diff
@@ -27,19 +28,32 @@ type promotedTag struct {
 	metadata *api.Metadata
 }
 
+type ClusterProfiles []ClusterProfileDetails
```
psalajova, line 31: This struct would probably not be defined here, but including it here for now

pkg/validation/config.go
```diff
@@ -24,6 +24,7 @@ type Validator struct {
 
 // NewValidator creates an object that optimizes bulk validations.
 func NewValidator() Validator {
+	// TODO: Changes here?
```
psalajova, line 27: If I understand it correctly, when we move cluster profiles from code in `ci-tools` into a config in the `release` repo, so we will need to do some changes in  `pkg/validation/config.go`

cmd/ci-operator-checkconfig/main.go
```diff
@@ -27,19 +28,32 @@ type promotedTag struct {
...
+
+type ClusterProfileDetails struct {
+	ProfileName string   `yaml:"profile"`
```
bbguimaraes, line 34:
```suggestion
	ProfileName api.ClusterProfile   `yaml:"profile"`
```

cmd/ci-operator-checkconfig/main.go
bbguimaraes: The validation code will need `ClusterProfileDetails`, so that will have to go to `pkg/api/types.go`.

jmguzik:
interesting why these tests are pending, do we have again some problems with one of cluster?

pkg/validation/config.go
```diff
@@ -24,13 +24,19 @@ type Validator struct {
...
 // NewValidator creates an object that optimizes bulk validations.
 func NewValidator() Validator {
-	profiles := api.ClusterProfiles()
+	return Validator{
+		hasTrapCache: make(map[string]bool),
+	}
+}
+
+// NewCPValidator creates an object that  and optimizes bulk validations.
+func NewCPValidator(profiles api.ClusterProfilesList) Validator {
```
psalajova, line 33: I basically split the function into 2, because the `validClusterProfiles` field of `Validator` is not filled anywhere besides when it is used for Cluster Profile checking.

pkg/api/types.go
```diff
@@ -2166,3 +2166,18 @@ func GetAvailableArchitectures() []string {
...
+
+type ClusterProfileDetails struct {
+	ProfileName ClusterProfile       `yaml:"profile"`
+	Owners      ClusterProfileOwners `yaml:"owners"`
+	ClusterType string               `yaml:"cluster_type,omitempty"`
+	LeaseType   string               `yaml:"lease_type,omitempty"`
+	Secret      string               `yaml:"secret,omitempty"`
+}
+
+type ClusterProfileOwners struct {
```
psalajova, line 2180: This will probably change depending on how the final config file looks like

pkg/validation/config.go
bbguimaraes: Are there any cases where we'd want to create a `Validator` that does not check for profile owners?

If not, we should have only one `New` function and modify all callers.

pkg/validation/config.go
```diff
@@ -24,13 +24,19 @@ type Validator struct {
...
 	ret := Validator{
-		validClusterProfiles: make(clusterProfileSet, len(profiles)),
+		validClusterProfiles: make(clusterProfileSet, len(api.ClusterProfiles())), // TODO: replace with 'len(profiles)' once the config file exists
 		hasTrapCache:         make(map[string]bool),
 	}
-	for _, x := range profiles {
-		ret.validClusterProfiles[x] = struct{}{}
+	for _, p := range api.ClusterProfiles() { // TODO: replace with 'profiles' once the config file exists
+		ret.validClusterProfiles[p] = api.ClusterProfileDetails{} // TODO: initialize with owner info once the config file exists
 	}
```
bbguimaraes, line 40: Why not keep the existing code (modulo changes to the value type of the `map`), since we're already requiring a profile list as a parameter?

pkg/validation/test_test.go
```diff
@@ -684,7 +684,7 @@ func TestValidateTests(t *testing.T) {
 	} {
 		t.Run(tc.id, func(t *testing.T) {
 			v := newSingleUseValidator()
-			errs := v.validateTestStepConfiguration(NewConfigContext(), "tests", tc.tests, tc.release, tc.releases, sets.New[string](), tc.resolved)
+			errs := v.validateTestStepConfiguration(NewConfigContext(), "tests", tc.tests, tc.release, nil, tc.releases, sets.New[string](), tc.resolved)
```
bbguimaraes, line 687: Don't leave commits that don't compile in the `git` history — unless you're going to use the `rebase` merge method, which I don't recommend.  They make `revert`, `bisect`, etc. more difficult to use.

c822c03d, 5f0c26de, and 38479a69 should be combined into one commit.

bbguimaraes (commented):
Looks good so far, a few comments.

pkg/validation/config.go
psalajova: Yes that's what I plan to do, this is just an intermediate step so all the existing tests pass, while this is still a WIP

pkg/load/load.go
```diff
@@ -232,3 +233,14 @@ func loadWorkflow(bytes []byte) (string, string, api.MultiStageTestConfiguration
 	}
 	return workflow.Workflow.As, workflow.Workflow.Documentation, workflow.Workflow.Steps, nil
 }
+
+func ClusterProfilesConfig(configPath string, profiles *api.ClusterProfilesList) error {
```
bbguimaraes, line 237: Let's merge the existing profiles from `pkg/api` with the loaded configuration here so that we just need to change this in the future.

bbguimaraes: Also, as much as I like the homage to C, the usual way to do this is to return multiple values.  So this function should receive just the path and return `(api.ClusterProfileList, error)`.

pkg/validation/config.go
```diff
@@ -13,7 +13,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/api"
 )
 
-type clusterProfileSet map[api.ClusterProfile]struct{}
+type clusterProfileSet map[api.ClusterProfile]api.ClusterProfileDetails
```
bbguimaraes, line 16: This makes it no longer a set, so it should be renamed.

pkg/validation/test.go
```diff
@@ -429,6 +433,32 @@ func (v *Validator) validateClusterProfile(fieldRoot string, p api.ClusterProfil
...
+}
+
+func contains(slice []string, item string) bool {
+	for _, s := range slice {
+		if s == item {
+			return true
+		}
+	}
+	return false
+}
```
bbguimaraes, line 460: https://github.com/openshift/ci-tools/blob/aa3322cb5573672ba80f1f70fe374781d0c96055/pkg/util/util.go#L44-L52

pkg/load/load.go
```diff
@@ -232,3 +233,34 @@ func loadWorkflow(bytes []byte) (string, string, api.MultiStageTestConfiguration
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read cluster profiles config: %w", err)
+	}
+
+	var profileOwnersList api.ClusterProfilesList
+	if err = yaml.Unmarshal(configContents, &profileOwnersList); err != nil {
+		return nil, fmt.Errorf("failed to unmarshall cluster profiles config: %w", err)
+	}
+
+	// The following code can be erased once profiles are completely moved
```
smg247, line 248: This comment should be prefixed with `TODO`. We should also make sure we have a card somewhere to followup on that

pkg/validation/config.go
smg247: @psalajova I am also interested to know if there are other valid usages without validating cluster profiles

smg247:
it looks like the `integration` test failures are actually caused by this PR, so that will need to be corrected.

pkg/load/load.go
psalajova: Good point. I have a card: https://issues.redhat.com/browse/DPTP-3642

psalajova:
I may be wrong but aren't the `integration` and `checkconfig` tests failing because of the missing arg that will be added in the PR for release repo (https://github.com/openshift/release/pull/43834)? The failure is caused by this line: https://github.com/openshift/ci-tools/blob/fffa18137ba3587f4ca06a1f66c6cc6af5a8ed37/pkg/load/load.go#L240

jmguzik:
@psalajova I am re-running tests here, it seems they are failing and time is approaching to focus on merging this PR.

PR #3610, opened 2023-09-08
Revert "Merge pull request #3608 from hongkailiu/revert-3602-skuznets/multiple-promotion-targets"

stevekuznetsov:
Revert "Merge pull request #3608 from hongkailiu/revert-3602-skuznets/multiple-promotion-targets"

This reverts commit 9a37c4b4c981d5c2ab97bfe41aebc9aa07fc2178, reversing
changes made to 3e5e4e03e41e0bbafddf48532a0fb4dd1910f3e9.

---

steps/promote: improve test coverage

---


pkg/steps/release/promote_test.go
```diff
@@ -287,16 +287,18 @@ func TestPromotedTags(t *testing.T) {
...
-		input    *api.ReleaseBuildConfiguration
-		options  []PromotedTagsOption
-		expected map[string][]api.ImageStreamTagReference
-		names    sets.Set[string]
+		name          string
+		input         *api.ReleaseBuildConfiguration
+		options       []PromotedTagsOption
+		expected      map[string][]api.ImageStreamTagReference
+		expectedNames sets.Set[string]
+		names         sets.Set[string]
```
hongkailiu, line 295: names seems not used at all and can be removed.

pkg/steps/release/promote_test.go
```diff
@@ -503,14 +515,62 @@ func TestPromotedTagsWithRequiredImages(t *testing.T) {
...
-			mapping, _ := PromotedTagsWithRequiredImages(testCase.input, testCase.options...)
-			if actual, expected := mapping, testCase.expected; !reflect.DeepEqual(actual, expected) {
-				t.Errorf("%s: got incorrect promoted tags: %v", testCase.name, diff.ObjectDiff(actual, expected))
+			mapping, names := PromotedTagsWithRequiredImages(testCase.input, testCase.options...)
+			actual, expected := mapping, testCase.expected
+			if diff := cmp.Diff(actual, expected); diff != "" {
+				t.Errorf("%s: got incorrect promoted tags: %v", testCase.name, diff)
+			}
+			if !names.Equal(testCase.expectedNames) {
+				actual, expected := names.UnsortedList(), testCase.expectedNames.UnsortedList()
```
hongkailiu, line 570: Do we need to sort them?

Set defines the `Equal` function like

https://github.com/openshift/ci-tools/blob/9a37c4b4c981d5c2ab97bfe41aebc9aa07fc2178/vendor/k8s.io/apimachinery/pkg/util/sets/set.go#L189-L194

Is it enough?

stevekuznetsov: We need to sort them in order to get stable diff output, but not for equality

hongkailiu: I tried with the following

```go
if !names.Equal(testCase.expectedNames) {
				t.Fatal("======")
			}
```

and then

```
go test -count 10000 -run TestPromotedTagsWithRequiredImages ./pkg/steps/release/...
```

It never failed. Isn't the diff stable enough?

hongkailiu: For my own understand, the sorting is only for pretty output on diff.
We could just fail the test as usual otherwise.

PR #3615, opened 2023-09-12
promotion: migrate to targets, tolerate mixed config

stevekuznetsov:
This commit updates the determinizer to migrate ci-operator configs to use the new promotion.to[] stanza instead of the previous inline configuration. For now, we will continue to honor the older configuration, a mixed configuration or only the new one.

### DIFF
--- a/cmd/config-brancher/main.go
+++ b/cmd/config-brancher/main.go
@@ -107,7 +107,7 @@ func generateBranchedConfigs(currentRelease, bumpRelease string, futureReleases 
 	devRelease := currentRelease
 	if bumpRelease != "" && promotion.IsBumpable(input.Info.Branch, currentRelease) {
 		devRelease = bumpRelease
-		updateRelease(&currentConfig, bumpRelease)
+		updateRelease(&currentConfig, currentRelease, bumpRelease)
 		updateImages(&currentConfig, currentRelease, bumpRelease)
 		// this config will continue to run for the dev branch but will be bumped
 		output = append(output, config.DataWithInfo{Configuration: currentConfig, Info: input.Info})
@@ -132,10 +132,14 @@ func generateBranchedConfigs(currentRelease, bumpRelease string, futureReleases 
 		}
 
 		// the new config will point to the future release
-		updateRelease(&futureConfig, futureRelease)
+		updateRelease(&futureConfig, devRelease, futureRelease)
 		// we cannot have two configs promoting to the same output, so
 		// we need to make sure the release branch config is disabled
-		futureConfig.PromotionConfiguration.Disabled = futureRelease == devRelease
+		for i := range futureConfig.PromotionConfiguration.Targets {
+			if futureConfig.PromotionConfiguration.Targets[i].Name == futureRelease {
+				futureConfig.PromotionConfiguration.Targets[i].Disabled = futureRelease == devRelease
+			}
+		}
 		// users can reference the release streams via build roots or
 		// input images, so we need to update those, too
 		updateImages(&futureConfig, devRelease, futureRelease)
@@ -162,9 +166,13 @@ func removePeriodics(tests *[]api.TestStepConfiguration) {
 
 // updateRelease updates the release that is promoted to and that
 // which is used to source the release payload for testing
-func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string) {
+func updateRelease(config *api.ReleaseBuildConfiguration, currentRelease, futureRelease string) {
 	if config.PromotionConfiguration != nil {
-		config.PromotionConfiguration.Name = futureRelease
+		for i := range config.PromotionConfiguration.Targets {
+			if config.PromotionConfiguration.Targets[i].Name == currentRelease {
+				config.PromotionConfiguration.Targets[i].Name = futureRelease
+			}
+		}
 	}
 	if config.ReleaseTagConfiguration != nil {
 		config.ReleaseTagConfiguration.Name = futureRelease
--- a/cmd/promoted-image-governor/main.go
+++ b/cmd/promoted-image-governor/main.go
@@ -441,7 +441,11 @@ func main() {
 			if _, ok := opts.explains[isTagRef]; ok {
 				opts.explains[isTagRef] = cfg.Metadata.AsString()
 			}
-			if cfg.PromotionConfiguration.TagByCommit {
+			var taggedByCommit bool
+			for _, target := range api.PromotionTargets(cfg.PromotionConfiguration) {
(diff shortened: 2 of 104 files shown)


stevekuznetsov:
Looking into integration

stevekuznetsov:
@bbguimaraes the `test/integration: determinize ci-operator configs` did not run integration with `UPDATE=true,` just the determinizer

stevekuznetsov:
@bbguimaraes the last commit ends up because `repo-init` does not write determinized prow config - would appreciate your help here, I have limited bandwidth to go digging into the repo right now :(

cmd/promoted-image-governor/main.go
```diff
@@ -441,7 +441,11 @@ func main() {
 			if _, ok := opts.explains[isTagRef]; ok {
 				opts.explains[isTagRef] = cfg.Metadata.AsString()
 			}
-			if cfg.PromotionConfiguration.TagByCommit {
+			var taggedByCommit bool
+			for _, target := range api.PromotionTargets(cfg.PromotionConfiguration) {
+				taggedByCommit = taggedByCommit || target.TagByCommit
+			}
```
bbguimaraes, line 447: This checks if _any_ tag is created by commit, not just `isTagRef`.

pkg/image-graph-generator/operator.go
```diff
@@ -80,12 +80,14 @@ func (o *Operator) callback(c *api.ReleaseBuildConfiguration, i *config.Info) er
...
-
-	for _, image := range c.Images {
-		if !excludedImages.Has(string(image.To)) {
-			if err := o.UpdateImage(image, c, branchID); err != nil {
-				errs = append(errs, err)
+	for _, target := range api.PromotionTargets(c.PromotionConfiguration) {
+		excludedImages := sets.New[string](target.ExcludedImages...)
+
+		for _, image := range c.Images {
+			if !excludedImages.Has(string(image.To)) {
```
bbguimaraes, line 87: This has to take into account the new meaning of `*`.

bbguimaraes: The existing code also doesn't seem to handle `additional_images`.

cmd/config-brancher/main.go
```diff
@@ -107,7 +107,7 @@ func generateBranchedConfigs(currentRelease, bumpRelease string, futureReleases
 	devRelease := currentRelease
 	if bumpRelease != "" && promotion.IsBumpable(input.Info.Branch, currentRelease) {
 		devRelease = bumpRelease
-		updateRelease(&currentConfig, bumpRelease)
+		updateRelease(&currentConfig, currentRelease, bumpRelease)
```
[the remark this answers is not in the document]
stevekuznetsov: Sorry, I don't quite understand the comment - isn't conditionally updating the specific `PromotionConfiguration.Targets` entry what my code does in `updateRelease` now?

cmd/promoted-image-governor/main.go
stevekuznetsov: Yes, I'm happy to make an issue for someone to make that better in the future but I am at the boundary of how much I can refactor here. The regex created will be a superset of the regex that's needed, and if for some reason images tagged by PR from another `Targets` stanza existed, they would not get cleaned up. The refactor here would require that `release.PromotedTags` return something other than a flat `isTagRef` list.

pkg/image-graph-generator/operator.go
stevekuznetsov: I have yet to code in the `*` bit - and as for `additional_images` here, I'm not sure exactly how to proceed, all I did in this diff was move the code into a loop over targets; so the previous code also did not handle `additional_images` - should it have?

stevekuznetsov:
/test e2e

Some ImagePullBackOff etc

cmd/promoted-image-governor/main.go
bbguimaraes: We can iterate over `PromotionTargets`, which is where all the information we need is.

There would still be some ways to create an unfortunate configuration which would not be processed correctly, but that's already the case for the current implementation (before `Targets`).

cmd/config-brancher/main.go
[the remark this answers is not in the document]
bbguimaraes: This is now a possible configuration:

```yaml
promotion:
  namespace: x
  name: y
  to:
  - namespace: ocp
    name: "4.14"
```

We cannot just modify the (top-level) `name` unconditionally anymore.

pkg/image-graph-generator/operator.go
bbguimaraes: I was hoping we could just use the same code that's used in other places to determine the images to be promoted and get rid of this partial (re-)implementation.

cmd/promoted-image-governor/main.go
stevekuznetsov: (Not impossible, just asking for this to be a follow-up that someone on DPTP could handle?)

pkg/image-graph-generator/operator.go
stevekuznetsov: (Not impossible, just asking for this to be a follow-up that someone on DPTP could handle?)

cmd/config-brancher/main.go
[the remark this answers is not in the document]
bbguimaraes: If we're not going to handle both, then this is fine.

cmd/promoted-image-governor/main.go
bbguimaraes: I see `PromotedTags` is quite messy already.  At least leave a comment here, then.

pkg/image-graph-generator/operator.go
bbguimaraes: Leave a comment here too.  There's no reason for this code duplication.

danilo-gemoli:
/retest
if it doesn't make it next run I'm going to `/override` as that's not an error

stevekuznetsov:
```
   * could not run steps: step src failed: error occurred handling build src: could not delete build src: the server is currently unable to handle the request (get builds.build.openshift.io src) 
```

/test e2e

stevekuznetsov:
```
 time="2023-09-19T11:16:33Z" level=error msg="failed to execute step" error="couldn't find SecretItem with item_name: build_farm name: sa.$(service_account).$(cluster).config containing cluster: build01" 
```

What is that in breaking changes job?

danilo-gemoli:
The test is failing due to https://github.com/openshift/release/pull/43448; that PR was opened to cope with `build01` outage.
/lgtm

PR #3627, opened 2023-09-16
quay_io_ci_images_distributor: consume mirroring tasks

hongkailiu:
https://issues.redhat.com/browse/DPTP-3529

/cc @openshift/test-platform

### DIFF
--- a/cmd/ci-images-mirror/main.go
+++ b/cmd/ci-images-mirror/main.go
@@ -67,7 +67,7 @@ func newOpts() *options {
 	fs.StringVar(&opts.leaderElectionNamespace, "leader-election-namespace", "ci", "The namespace to use for leader election")
 	fs.StringVar(&opts.leaderElectionSuffix, "leader-election-suffix", "", "Suffix for the leader election lock. Useful for local testing. If set, --dry-run must be set as well")
 	fs.Var(&opts.enabledControllers, "enable-controller", fmt.Sprintf("Enabled controllers. Available controllers are: %v. Can be specified multiple times. Defaults to %v", allControllers.UnsortedList(), opts.enabledControllers.Strings()))
-	fs.BoolVar(&opts.dryRun, "dry-run", false, "Whether to run the controller-manager with dry-run")
+	fs.BoolVar(&opts.dryRun, "dry-run", false, "Whether to run the controller-manager and the mirroring with dry-run")
 	fs.StringVar(&opts.releaseRepoGitSyncPath, "release-repo-git-sync-path", "", "Path to release repository dir")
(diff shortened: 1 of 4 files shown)


pkg/controller/quay_io_ci_images_distributor/mirror.go
```diff
@@ -80,3 +83,48 @@ func NewMirrorStore() MirrorStore {
...
+func (c *MirrorConsumerController) Run() error {
+	for {
+		mirrors, err := c.mirrorStore.Take(10)
+		if err != nil {
+			c.logger.WithError(err).Warn("Failed to take mirrors")
+			continue
+		}
+		if len(mirrors) == 0 {
+			c.logger.Debug("Waiting for mirror tasks ...")
+			time.Sleep(3 * time.Second)
```
danilo-gemoli, line 103: Have you already considered switching to an event base approach?

cmd/ci-images-mirror/main.go
```diff
@@ -267,3 +278,9 @@ func mirrors(action string, n int, ms quayiociimagesdistributor.MirrorStore) (an
 		return nil, fmt.Errorf("invalid action: %s", action)
 	}
 }
+
+func execute(c *quayiociimagesdistributor.MirrorConsumerController) {
+	if err := c.Run(); err != nil {
+		logrus.WithError(err).Error("Error running")
```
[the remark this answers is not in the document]
hongkailiu: > Is it enough to just log the error and move forward?

For now as a starter it should be enough. I do not know how the errors hit us yet.
I am thinking retries, and add `--force` etc there.
If none of them work, logging error for human debugging.
Plus, the tool is watching the imagestreams, when it starts, it will pipeline all the tags overall again. So that we do not miss.

> Will the distributor still be useful without a consumer?
The consumer is https://github.com/openshift/ci-tools/pull/3627/files#diff-d20fe4d5bdebdd0e898bb1b5a9cb21ce5d075f0285eb619022c277a9b68ca1b1R87

pkg/controller/quay_io_ci_images_distributor/mirror.go
hongkailiu: Yes. This is very good point.
The task producer is already event based.
In K8S world, typically I could generate a CRD in the producer and the consumer here watch its CR.
We do have code like this in other dptp-controllers as you might have seen.
For the simplicity, I did not go that direction for the consumer.
I want to feel how it goes in our production first and then optimize the tool for the issues there.

danilo-gemoli:
Please rebase on top of master, retest and then I'll `/lgtm`

hongkailiu:
> Please rebase on top of master, retest and then I'll `/lgtm`

Done. Thanks for the review.

PR #3630, opened 2023-09-18
TRT-1243: All changes to get 4.15 in sync as well as prior 4.14 && 4.…

neisw:
…13 changes

- 4.13 && 4.14 jobs were added prior then reverted
- revert did not remove the jobs from the jobs table after they had already been added
- revert was effectively a no-op since the jobs had already been added to the jobs table

If we want to match the state of 4.14 I *believe* this is a truer representation


neisw:
/hold

Need to review which update we want this or the min version https://github.com/openshift/ci-tools/pull/3631

PR #3633, opened 2023-09-20
testgrid-config-generator: allow OLM tests

stevekuznetsov:
RedHat manages quite a large list of TestGrid configurations that are *not* derived from the release controller configuration. It's a shame that every single one of those folks need to manually configure TestGrid for their use-case and cannot take advantage of this tool and the automatic PR updates.

This is a small commit that mirrors the OSD additions added last year; in the process of working through the wrinkles for this new TestGrid config we will know how to factor the code such that other config could also be added here.


danilo-gemoli:
Please rebase on top of the master tip

stevekuznetsov:
@danilo-gemoli done

perdasilva:
Hhhmmm:

```
ERROR: Running cluster-init in update mode in openshift/release results in changes ^^^
ERROR: To avoid breaking openshift/release for everyone you should regenerate the build clusters
ERROR: there and merge the changes ASAP after this change to cluster-init 
```

maybe there's a make target you need to run?

PR #3635, opened 2023-09-20
`ci-operator-prowgen`: fix image definition

danilo-gemoli:
This is a follow up of https://github.com/openshift/release/pull/43485, `sanitize-prow-jobs` was removed:
```diff
      paths:
      - destination_dir: .
        source_path: /go/bin/ci-operator-prowgen
-      - destination_dir: .
-        source_path: /go/bin/sanitize-prow-jobs
  to: ci-operator-prowgen
```

PR #3636, opened 2023-09-20
Revert #3615

danilo-gemoli:
Sorry @stevekuznetsov but the `config-brancher` tool doesn't work anymore and we have to carry on some branching activities.
We have not been hit yet only because the [promotion job](https://prow.ci.openshift.org/view/gs/origin-ci-test/logs/branch-ci-openshift-ci-tools-master-images/1704109540636626944) failed, lucky for us (well, if we can define "luck" as one of our most important jobs failing...) .


stevekuznetsov:
Linking to an issue on how, exactly, it broke the tool would help me look into it FWIW...

danilo-gemoli:
I'll handle `ci/prow/breaking-changes ` lately.

`images` are broken, the fix is either in https://github.com/openshift/ci-tools/pull/3635 or https://github.com/openshift/ci-tools/pull/3637
/override ci/prow/images 

> Linking to an issue on how, exactly, it broke the tool would help me look into it FWIW...

I'll take a look and open an issue if it will require a non trivial fix

PR #3638, opened 2023-09-21
`release`: Add observers support

danilo-gemoli:
Add support for the observers:
- the registry resolver gets the observers definitions
- the `registry` command is now able to properly resolve any observers

/cc @bbguimaraes

### DIFF
--- a/pkg/cmd/release/registry.go
+++ b/pkg/cmd/release/registry.go
@@ -32,6 +32,7 @@ resolution.`,
 	ret.AddCommand(newCmdRegistryStep(o))
 	ret.AddCommand(newCmdRegistryChain(o))
 	ret.AddCommand(newCmdRegistryWorkflow(o))
+	ret.AddCommand(newCmdRegistryObserver(o))
 	return &ret
 }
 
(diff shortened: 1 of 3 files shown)


pkg/cmd/release/registry.go
```diff
@@ -209,4 +238,6 @@ func printTreeWorkflow(o *options, name string, w test, level uint) {
 	printTreeSteps(o, w.Test, level+1)
 	fmt.Println("post:")
 	printTreeSteps(o, w.Post, level+1)
+	fmt.Println("observers:")
+	printTreeObservers(w.Observers, level+1)
```
bbguimaraes, line 242:
```
$ go run ./cmd/release -C ~/src/release/ registry workflow --tree
workflow: 3scale-ipi-aws
…
observers:
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x19dddfc]
```

This cannot be done unconditionally.

bbguimaraes: Also, tests and individual steps can declare observers (no workflow currently declares one in openshift/release@df4b5732d9d).  And I'd expect to get just the names normally and a full dump with `--resolve`, as we do for regular steps.

bbguimaraes: If you'd like, remove this and we can merge just the listing part now, which looks good.  Then we can add this and the rest later.

PR #3639, opened 2023-09-22
`observer pods`: add StepParameters

danilo-gemoli:
Introduce `[]StepParameter` on the observers and then allowing any environment variables to flow from a test definition.

/cc @bbguimaraes @jmguzik


jmguzik (approved):
LGTM
/hold
let's proceed on Mon with lifting hold label

PR #3641, opened 2023-09-23
ci-secret-generator: support disabled clusters

hongkailiu:
https://issues.redhat.com/browse/DPTP-3659

/cc @openshift/test-platform


hongkailiu:
/hold

We might not need this at all

PR #3644, opened 2023-09-25
`config-brancher`: skip promotion targets when branching

danilo-gemoli:
Skip `.promotion.to[]` targets when branching a configuration. 
As an example, the following configuration on the dev branch `$ORG-$REPO-master.yaml`:
```yaml
promotion:
  excluded_images:
  - ci-image
  - olm-bumper
  name: "4.15"
  namespace: ocp
  to:
  - additional_images:
      ci-image: ci-image
      olm-bumper: olm-bumper
    excluded_images:
    - '*'
    namespace: operator-lifecycle-manager
    tag: latest
```
gets two new branched configurations when `config-brancher --current-release=4.15 --future-release=4.16` is run:
```yaml
promotion:
  excluded_images:
  - ci-image
  - olm-bumper
  name: "4.15"
  namespace: ocp
  disabled: true
```
targeting current release `$ORG-$REPO-release-4.15.yaml` and:
```yaml
promotion:
  excluded_images:
  - ci-image
  - olm-bumper
  name: "4.16"
  namespace: ocp
```
targeting the next release `$ORG-$REPO-release-4.16.yaml`.
In both cases `.promotion.to[]` has been deleted.

### DIFF
--- a/cmd/config-brancher/main.go
+++ b/cmd/config-brancher/main.go
@@ -161,34 +161,39 @@ func removePeriodics(tests *[]api.TestStepConfiguration) {
 }
 
 func updatePromotion(currentConfig, futureConfig *api.ReleaseBuildConfiguration, currentRelease, futureRelease, devRelease string) {
-	if futureConfig.PromotionConfiguration == nil {
+	if currentConfig.PromotionConfiguration == nil {
 		return
 	}
 
-	// we cannot have two configs promoting to the same output, so
-	// we need to make sure the release branch config is disabled
-	futureConfig.PromotionConfiguration.Disabled = futureRelease == devRelease
+	currentPromotion := currentConfig.PromotionConfiguration
+	futurePromotion := futureConfig.PromotionConfiguration
 
-	if currentConfig.PromotionConfiguration.Targets == nil {
+	if currentPromotion.Name != "" || currentPromotion.Tag != "" {
+		// we cannot have two configs promoting to the same output, so
+		// we need to make sure the release branch config is disabled
+		futurePromotion.Disabled = futureRelease == devRelease
+	}
+
+	if currentPromotion.Targets == nil {
 		return
 	}
 
 	// filter and upgrade .promotion.to[] releases that promote to the current release
-	newTargets := make([]api.PromotionTarget, 0, len(currentConfig.PromotionConfiguration.Targets))
-	for _, target := range currentConfig.PromotionConfiguration.Targets {
+	newTargets := make([]api.PromotionTarget, 0, len(currentPromotion.Targets))
+	for _, target := range currentPromotion.Targets {
 		if target.Name == currentRelease {
 			target.Name = futureRelease
 			target.Disabled = futureRelease == devRelease
 			newTargets = append(newTargets, target)
 		}
 	}
-	futureConfig.PromotionConfiguration.Targets = newTargets
+	futurePromotion.Targets = newTargets
 }
 
 // updateRelease updates the release that is promoted to and that
 // which is used to source the release payload for testing
 func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string) {
-	if config.PromotionConfiguration != nil {
+	if config.PromotionConfiguration != nil && config.PromotionConfiguration.Name != "" {
 		config.PromotionConfiguration.Name = futureRelease
 	}
 	if config.ReleaseTagConfiguration != nil {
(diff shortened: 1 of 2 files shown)


cmd/config-brancher/main.go
```diff
@@ -133,9 +133,17 @@ func generateBranchedConfigs(currentRelease, bumpRelease string, futureReleases
...
-		// we cannot have two configs promoting to the same output, so
-		// we need to make sure the release branch config is disabled
-		futureConfig.PromotionConfiguration.Disabled = futureRelease == devRelease
+
+		if futureConfig.PromotionConfiguration != nil {
+			// we cannot have two configs promoting to the same output, so
+			// we need to make sure the release branch config is disabled
+			futureConfig.PromotionConfiguration.Disabled = futureRelease == devRelease
+
+			// Additional targets have nothing to do with the OCP release lifecycle, so
```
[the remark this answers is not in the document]
danilo-gemoli: So are we going to move from this schema:
```yaml
promotion:
  name: target-1
  to:
  - name: target-2
  - name: target-3
```
to something like the following?
```yaml
promotion:
- name: target-1
- name: target-2
- name: target-3
```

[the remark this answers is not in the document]
danilo-gemoli: Done, have a look please.

cmd/config-brancher/main.go
```diff
@@ -160,6 +160,31 @@ func removePeriodics(tests *[]api.TestStepConfiguration) {
...
 }
 
+func updatePromotion(currentConfig, futureConfig *api.ReleaseBuildConfiguration, currentRelease, futureRelease, devRelease string) {
+	if futureConfig.PromotionConfiguration == nil {
+		return
+	}
+
+	// we cannot have two configs promoting to the same output, so
+	// we need to make sure the release branch config is disabled
+	futureConfig.PromotionConfiguration.Disabled = futureRelease == devRelease
```
stevekuznetsov, line 170: You should check that the top-level fields (either `name` or `tag`) are set - if the `determinize-ci-operator` changes I originally suggested are live, no config in the repo should have them set, all should be using `to` - and, in that case, setting the top-level fields is not useful or wanted.

stevekuznetsov (commented):
Looks good, one edge case comment

cmd/config-brancher/main.go
danilo-gemoli: I think you might be referring to https://github.com/openshift/ci-tools/pull/3615/files#diff-69f51aa2fa3387ebe31fa93b59fcf08100eb94699b6e84715ecf09dc1f71e2f0 that is not live, not yet at least. I will anticipate the change then 👍🏻

cmd/config-brancher/main.go
```diff
@@ -160,10 +160,40 @@ func removePeriodics(tests *[]api.TestStepConfiguration) {
...
+		}
+	}
+	futurePromotion.Targets = newTargets
+}
+
 // updateRelease updates the release that is promoted to and that
 // which is used to source the release payload for testing
 func updateRelease(config *api.ReleaseBuildConfiguration, futureRelease string) {
-	if config.PromotionConfiguration != nil {
+	if config.PromotionConfiguration != nil && config.PromotionConfiguration.Name != "" {
```
danilo-gemoli, line 196: > You should check that the top-level fields (either name or tag) are set - if the determinize-ci-operator changes I originally suggested are live, no config in the repo should have them set, all should be using to - and, in that case, setting the top-level fields is not useful or wanted.

@stevekuznetsov  I guess it's the same here, right?

stevekuznetsov: Yes - and you'd need to check that both name *and* tag are `""`, since normal config only has to set one

danilo-gemoli: I don't quite get this one, as long as your changes to the `determinize-ci-operator` don't go live, either `.promotion.name` or `.promotion.tag` is set (exclude `.promotion.to[]` for now), doesn't it mean that I should set one of those two fields only if it's not empty?
```golang
currentPromotion := config.PromotionConfiguration
if currentPromotion != nil {
    if currentPromotion.Name != "" {
        currentPromotion.Name = futureRelease
    } else if currentPromotion.Tag != "" {
        currentPromotion.Tag = futureRelease
    }
}
```

stevekuznetsov: ACK, sorry - I misread your diff. I think the current PR is correct.

jmguzik (approved):

jmguzik (approved):

PR #3651, opened 2023-09-27
Ignore known infra-files in `prowgen` warnings

smg247:
We shouldn't output warnings for these files, as it confuses users and they think something is broken.

For: https://issues.redhat.com/browse/DPTP-3376

I will follow this up by adding the arguments to the call for `make jobs` in `openshift/release`

### DIFF
--- a/cmd/ci-operator-prowgen/main.go
+++ b/cmd/ci-operator-prowgen/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/prowgen"
 	"github.com/openshift/ci-tools/pkg/registry"
 	"github.com/openshift/ci-tools/pkg/util"
+	"k8s.io/test-infra/prow/flagutil"
 )
 
 type options struct {
@@ -33,6 +34,8 @@ type options struct {
 
 	registryPath string
 	resolver     registry.Resolver
+
+	knownInfraJobFiles flagutil.Strings
 
 	help bool
 }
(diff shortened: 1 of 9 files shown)


danilo-gemoli:
It's fine, but maybe we can see the problem from a different angle: `OperateOnJobConfigSubdirPaths(configDir, subDir string, callback func(*Info) error) error` is supposed to operate on a directory structure like `ci-operator/jobs/$ORG/$REPO/*.yaml`,  so when invoked as `OperateOnJobConfigSubdirPaths("$RELEASE_REPO/ci-operator/jobs", "", func(*Info) { ... } error)` I expect it to find every configuration that matches the pattern `$ORG/$REPO/*.yaml`, leaving out everything else. That because I set the  `subDir` parameter to be `""` and it somehow means the intention of me wanting to search for every ProwJob that follows the OrgRepo pattern.
It would be a breaking change as I'm modifying the default behavior so what about adding `opts ...JobSearchOptions` (type name is just an example here) on the signature and allowing the client to modify the searching behavior?

smg247:
@danilo-gemoli would that solution allow us to still throw an error/warn about misnamed files that are unexpected? For example, if a user adds a new file that doesn't follow the correct format we would need to throw an error there. This solution allows us to ignore the specific infra files we know about.

danilo-gemoli:
Do you mean throwing an error if a user place a file like so `ci-operator/jobs/misplaced.yaml` on purpose?

smg247:
Yes, exactly. Otherwise we would just completely remove this warning. The goal here is to omit the error/warning for known infra files.

PR #3654, opened 2023-09-27
cluster-display: cache page data

hongkailiu:


cmd/cluster-display/main.go
```diff
@@ -77,7 +78,8 @@ func addSchemes() error {
 	return nil
 }
 
-func getClusterPoolPage(ctx context.Context, hiveClient ctrlruntimeclient.Client) (*Page, error) {
+func getClusterPool(ctx context.Context, hiveClient ctrlruntimeclient.Client) ([]map[string]string, error) {
+	logrus.Info("Calling getClusterPool ...")
```
smg247, line 82: If you meant to leave this log in, I think it should be `debug` level

cmd/cluster-display/main.go
```diff
@@ -161,13 +162,62 @@ func (g *clusterInfoGetter) GetClusterDetails(ctx context.Context, cluster strin
...
+	ClusterPoolData              []map[string]string
+	ClusterPoolDataLastUpdatedAt time.Time
+}
+
+func (c *memoryCache) GetClusterPoolPage(ctx context.Context, client ctrlruntimeclient.Client) (*Page, error) {
+	c.ClusterPoolDataMutex.Lock()
+	defer c.ClusterPoolDataMutex.Unlock()
+	if c.ClusterPoolData == nil || time.Now().After(c.ClusterPoolDataLastUpdatedAt.Add(time.Hour)) {
+		data, err := getClusterPool(ctx, client)
+		if err != nil {
```
smg247, line 180: We return an error here, but some of the value I was thinking we could achieve by adding the cache is to continue to return the previous results in this state. That way we could still display the last available information about the cluster. We could maybe display a warning that there is a problem, and one or more cluster may be unreachable or outdated along with this info.

cmd/cluster-display/main.go
hongkailiu: Done

cmd/cluster-display/main.go
hongkailiu: > We could maybe display a warning that there is a problem

Added a log in the backend instead.

PR #3658, opened 2023-09-29
pipeline-controller to support 2-staged pipelines

jmguzik:


jmguzik:
sth about build09 maybe because my branch is not updated

cmd/pipeline-controller/config_data_provider.go
```diff
@@ -0,0 +1,94 @@
...
+	defer c.m.Unlock()
+	if presubmits, ok := c.updatedPresubmits[orgRepo]; ok {
+		return presubmits
+	}
+	return presubmitTests{}
+}
+
+func (c *ConfigDataProvider) Run() {
+	for {
+		time.Sleep(10 * time.Minute)
```
danilo-gemoli, line 43: What about using [Agent.Subscribe](https://github.com/kubernetes/test-infra/blob/1361dca58e78f2ac6ec6eda929d2c6f676e91b95/prow/config/agent.go#L385-L389) instead of polling each 10 mins?

cmd/pipeline-controller/reconciler.go
```diff
@@ -0,0 +1,220 @@
...
+			if existing, ok := latestBatch[pjob.Spec.Job]; !ok {
+				latestBatch[pjob.Spec.Job] = pjob
+			} else if pjob.CreationTimestamp.After(existing.CreationTimestamp.Time) {
+				latestBatch[pjob.Spec.Job] = pjob
+			}
+		}
+	}
+
+	for _, presubmit := range presubmits.protected {
+		if _, ok := latestBatch[presubmit]; ok {
```
danilo-gemoli, line 198: Does it mean that if I have already manually triggered one of the jobs this controller would have triggered then it won't do anything?
As an example: let's suppose `e2e-1` and `e2e-2` on a repository, both being `protected`. Then I open a PR and I manually `/test e2e-1` but  no `e2e-2`. What does happen next?

cmd/pipeline-controller/reconciler.go
```diff
@@ -0,0 +1,220 @@
...
+	return fmt.Sprintf("%s/%s/%d/%s/%s", refs.Org, refs.Repo, refs.Pulls[0].Number, refs.BaseRef, refs.Pulls[0].SHA)
+}
+
+type reconciler struct {
+	pjclientset        ctrlruntimeclient.Client
+	lister             ctrlruntimeclient.Reader
+	configDataProvider *ConfigDataProvider
+	ghc                minimalGhClient
+	closedPRsCache     closedPRsCache
+	ids                sync.Map
```
danilo-gemoli, line 86: Can you please add a comment describing what's the purpose of this field `ids`?

cmd/pipeline-controller/main.go
```diff
@@ -0,0 +1,122 @@
...
+	mgr, err := manager.New(restCfg, manager.Options{
+		Namespace:          cfg().ProwJobNamespace,
+		MetricsBindAddress: "0",
+	})
+	if err != nil {
+		logger.WithError(err).Fatal("failed to create manager")
+	}
+
+	if err := o.client.AddKubeconfigChangeCallback(func() {
+		logger.Info("kubeconfig changed, exiting to trigger a restart")
```
danilo-gemoli, line 90: This is nice 👍🏻  do we hit such a case by the way?

cmd/pipeline-controller/reconciler.go
jmguzik: yes, if you trigger manually one of the jobs this controller the given SHA won't do anything. I think this is kind of logical. If user wants to speed-up execution then he/she takes the responsibility for the pipeline. It will be part of documentation.

cmd/pipeline-controller/reconciler.go
jmguzik: I can change a name, as I am not a big fan of comments :)

cmd/pipeline-controller/main.go
jmguzik: rarely I guess, this is standard check we have in prow as well

cmd/pipeline-controller/config_data_provider.go
jmguzik: I understand the intention, but totally not needed here as we pool what is mounted to our local disk so no tokens of any sort are burned. In the initial version it was even 30 mins, but I guess anything above 5 mins is minimal load for the system.

jmguzik:
@danilo-gemoli I will change the name from `ids` to something like `prBatchIds` to be more descriptive. To rest of your comments I responded with polemics. Let me know if my responses are enough or you want some additional clarification, then I will update the PR.

cmd/pipeline-controller/config_data_provider.go
danilo-gemoli: > no tokens of any sort are burned

My concern is about being more responsive to the configuration change events rather than consuming less tokens. It's not something I strictly require for this PR to get merged by the way. We can come back to it as soon as we receive complaints by the users.

danilo-gemoli:
It would be nice to have some more comments in general, especially on `ids` as caching things is always hard. With that said, it looks fine to me.

PR #3661, opened 2023-09-29
cluster-display: support disabled clusters

hongkailiu:
https://issues.redhat.com/browse/DPTP-3659

@openshift/test-platform

### DIFF
--- a/pkg/prowconfigutils/config.go
+++ b/pkg/prowconfigutils/config.go
@@ -1,0 +1,60 @@
+package prowconfigutils
+
+import (
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/flagutil"
+	"sigs.k8s.io/yaml"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+// ProwDisabledClusters returns the disabled clusters in Prow and sets disable clusters on the given KubernetesOptions
+func ProwDisabledClusters(o *flagutil.KubernetesOptions) (ret []string, retErr error) {
+	prowDisabledClusters := sets.New[string]()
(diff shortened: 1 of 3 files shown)


danilo-gemoli:
Drop a comment when it's ready to be reviewed

hongkailiu:
> Drop a comment when it's ready to be reviewed

It is ready now. Thanks.

pkg/prowconfigutils/config.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/sirupsen/logrus"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/test-infra/prow/flagutil"
+	"sigs.k8s.io/yaml"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+// ProwDisabledClusters returns the disabled clusters in Prow and sets disable clusters on the given KubernetesOptions
```
danilo-gemoli, line 17: Why returning and setting on the options at the same time?

hongkailiu: A good one.
There is Getter from the API (upstream).
So the caller wants to know the disabled clusters for other purposes.

My initial version returns the list but it would create duplicated code for the setter on the caller part.

danilo-gemoli: I get your point but getting and setting things are two separate actions. IMHO it's up to the caller to decide what to do with the result (ex.: I could log it and nothing else).
Code duplication is not necessarily bad, not in every case at least.
I've also noticed that:
```golang
if o != nil {
    o.SetDisabledClusters(prowDisabledClusters)
}
```
set the result even when  `retErr != nil`

danilo-gemoli:
Other than a minor comment, it's fine to me.

pkg/prowconfigutils/config.go
hongkailiu: When retErr != nil, prowDisabledClusters is always empty.
I think it is the same effect with or without calling the setter when it is empty.
But let me do the error check since you prefer to it.

PR #3673, opened 2023-10-04
Add build09 to update_jobs

bear-redhat:
`build09` is OCP cluster, which needs the `oauth` to be set up manually.


hongkailiu (approved):

cmd/cluster-init/update_jobs.go
```diff
@@ -186,6 +186,10 @@ func generateContainer(image, clusterName string, extraArgs []string, extraVolum
 					},
 				},
 			},
+		}...)
+	}
+	if clusterName == string(api.ClusterBuild01) || clusterName == string(api.ClusterBuild02) {
```
smg247, line 191: Any reason why we don't want `ci-slack-api-url` in `build09`? I think we need this to get alerts sent to slack

bear-redhat: I am using robusta to relay the alerts to Slack

PR #3676, opened 2023-10-10
`MultiArchBuilderController`: push to external registries

danilo-gemoli:
Support pushing images to external registries.
The `.spec` stanza has now the new `.output` field through which specifying the external registries an imaged must be mirrored to.
The whole feature is implemented in `mirror.go`: the reconciler ~spawns a pod that `oc image mirror` to the locations set on the field `.spec.output.to`~ runs `oc image mirror` so mirroring to the locations set on the field `.spec.output.to`.
~A `mabc` owns those pods, the relationship is bound with the `ownerReferences` field.~


pkg/controller/multiarchbuildconfig/mirror.go
```diff
@@ -0,0 +1,282 @@
...
+	}
+	podName := types.NamespacedName{
+		Namespace: mabc.Namespace,
+		Name:      name,
+	}
+
+	// Either the pod doesn't exist yet so it has to be created or it does and we
+	// are going to check its state. Anything in between is treated as an error.
+	mirrorPod := corev1.Pod{}
+	if err := r.client.Get(ctx, podName, &mirrorPod, &client.GetOptions{}); err != nil {
```
droslean, line 103: We don't need to spawn a new pod to mirror. The controller can just run it directly.

danilo-gemoli: I've changed the implementation; it now mirrors running `oc image run` without making any new pods.

danilo-gemoli:
/hold
[openshift/release#44404](https://github.com/openshift/release/pull/44404) needed

pkg/controller/multiarchbuildconfig/mirror.go
```diff
@@ -0,0 +1,103 @@
...
+
+	// Condition reasons
+	ImageMirrorErrorReason   string = "ImageMirrorError"
+	ImageMirrorSuccessReason string = "ImageMirrorSuccess"
+)
+
+// handleMirrorImage pushes an image to the locations specified in .spec.output.to. The function requires an image
+// to have been already created and pushed to the local registry.
+func (r *reconciler) handleMirrorImage(srcImage string, mabc *v1.MultiArchBuildConfig) (func(mabcToMutate *v1.MultiArchBuildConfig), bool) {
+	if len(mabc.Spec.Output.To) == 0 {
```
droslean, line 28: This will never happen based on https://github.com/openshift/ci-tools/pull/3676/files#diff-a3312e2e48e4c26f431b062d5e8c3fd8b2c2abe781a6913135f17d2d4620afefR44-R45

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -61,32 +64,25 @@ func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath str
...
-	architectures  []string
-	manifestPusher manifestpusher.ManifestPusher
+	logger           *logrus.Entry
+	client           ctrlruntimeclient.Client
+	architectures    []string
+	manifestPusher   manifestpusher.ManifestPusher
+	dockerConfigPath string
+
+	// For testing purpose only
+	timeNowFn      func() time.Time
```
droslean, line 84: Why do we need this?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -61,32 +64,25 @@ func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath str
...
+	mirrorImagesFn func(log *logrus.Entry, registryConfig string, images []string) error
```
droslean, line 85: Also this.

droslean: Adding something in the reconciling struct just for the test cases is smelly. It can be done by another approach or the structure of the tested function should be improved.

danilo-gemoli: In the end I need a way to somehow mock [mirrorImages](https://github.com/openshift/ci-tools/pull/3676/files#diff-da2cf6dcc4c106cde64397ab1fa87a001d400aae2e5fd563afadefa98ac8cf43R63). I could pass the function to [handleMirrorImage](https://github.com/openshift/ci-tools/pull/3676/files#diff-da2cf6dcc4c106cde64397ab1fa87a001d400aae2e5fd563afadefa98ac8cf43R27) like so:
```golang
type mirrorFunc func(log *logrus.Entry, registryConfig string, images []string) error
// and then add it as a new arg here
func (r *reconciler) handleMirrorImage(srcImage string, mabc *v1.MultiArchBuildConfig, mirror mirrorFunc) (func(mabcToMutate *v1.MultiArchBuildConfig), bool) { ... }
```
but then the problem raises again when we unit test the `reconcile` function:
```golang
// again here...
func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logger *logrus.Entry, mirror mirrorFunc) error { ... }
```
I really would like to avoid this cascade effect as it's smelly as well.

What about an interface like the [manifestPusher](https://github.com/openshift/ci-tools/blob/fb0bf5d51faaff2760e5033ba3788314dd22c1af/pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go#L80)?

pkg/controller/multiarchbuildconfig/mirror.go
danilo-gemoli: Yeah, that's right thanks. But still I need to check `if mabc.Spec.Output.To != nil`

droslean: You don't need to do that, because it will never happen.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: Yea we can do a similar approach.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -45,10 +45,13 @@ func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath str
...
-			client:         mgr.GetClient(),
-			architectures:  architectures,
-			manifestPusher: manifestpusher.NewManifestPushfer(logger, registryURL, dockerCfgPath),
+			logger:           logger,
+			client:           mgr.GetClient(),
+			architectures:    architectures,
+			dockerConfigPath: dockerCfgPath,
+			manifestPusher:   manifestpusher.NewManifestPushfer(logger, registryURL, dockerCfgPath),
+			timeNowFn:        time.Now,
+			mirrorImagesFn:   mirrorImages,
```
droslean, line 54: This should exist here.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -45,10 +45,13 @@ func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath str
...
 		MaxConcurrentReconciles: 1,
 		Reconciler: &reconciler{
-			logger:         logger,
-			client:         mgr.GetClient(),
-			architectures:  architectures,
-			manifestPusher: manifestpusher.NewManifestPushfer(logger, registryURL, dockerCfgPath),
+			logger:           logger,
+			client:           mgr.GetClient(),
+			architectures:    architectures,
+			dockerConfigPath: dockerCfgPath,
```
droslean, line 51: Check the `manifestPusher`

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -61,32 +64,25 @@ func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath str
...
+
+	if err := c.Watch(source.Kind(mgr.GetCache(), &v1.MultiArchBuildConfig{}),
+		&handler.EnqueueRequestForObject{}, predicateFuncs); err != nil {
 		return fmt.Errorf("failed to create watch for MultiArchBuildConfig: %w", err)
 	}
 
 	return nil
 }
 
-func mabcHandler() handler.EventHandler {
```
droslean, line 71: Why this is deleted?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -133,22 +129,21 @@ func (r *reconciler) handleMultiArchBuildConfig(ctx context.Context, mabc *v1.Mu
...
-				Type:               "PushManifestError",
-				Status:             metav1.ConditionFalse,
-				LastTransitionTime: metav1.Time{Time: time.Now()},
-				Reason:             "PushManifestError",
-				Message:            err.Error(),
-			})
-			mabcToMutate.Status.State = v1.FailureState
-		}
-	}
+	mutateFn := r.runPostBuildActions(
```
droslean, line 132: This fn is just to update the actual resource. All of those operations should be done before that.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -185,6 +180,72 @@ func (r *reconciler) createBuildsForArchitectures(ctx context.Context, mabc *v1.
 	return nil
 }
 
+func (r *reconciler) handlePushImageWithManifest(mabc *v1.MultiArchBuildConfig, targetImageRef string) (func(mabcToMutate *v1.MultiArchBuildConfig), bool) {
+	if getCondition(mabc, PushImageManifestDone) != nil {
```
droslean, line 184: You can remove this as well.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -185,6 +180,72 @@ func (r *reconciler) createBuildsForArchitectures(ctx context.Context, mabc *v1.
...
+			LastTransitionTime: metav1.Time{Time: time.Now()},
+			Reason:             "PushManifestSuccess",
+		})
+	}, true
+}
+
+// runPostBuildActions ensure some actions are performed after the builds have been spawned.
+// Actions are performed in order, one after the other. If at least one fails then the chain breaks
+// and the overall status is set to failed.
+func (r *reconciler) runPostBuildActions(actions ...func() (func(mabcToMutate *v1.MultiArchBuildConfig), bool)) func(mabcToMutate *v1.MultiArchBuildConfig) {
```
droslean, line 214: You shouldn't be needed to care about this. You only need to run the mirroring after the manifestPusher succeeds.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: It's redundant since `source.Kind(mgr.GetCache(), &v1.MultiArchBuildConfig{})` already guarantees to filter objects of type `v1.MultiArchBuildConfig{}` and then `&handler.EnqueueRequestForObject{}` creates a new `reconcile.Request` with the proper object name and namespace. [Here](https://github.com/kubernetes-sigs/controller-runtime/blob/658c55227830b2d895b12fc1c86cbc731e36d291/examples/builtins/main.go#L62) an example from the `controller-runtime` repo.

pkg/api/multiarchbuildconfig/v1/types.go
```diff
@@ -33,6 +33,17 @@ type MultiArchBuildConfig struct {
 
 type MultiArchBuildConfigSpec struct {
 	BuildSpec buildv1.BuildConfigSpec `json:"build_spec"`
+	// output describes the container image the MultiArchBuildConfig should produce.
+	//+kubebuilder:validation:Optional
+	Output MultiArchBuildConfigOutput `json:"output,omitempty"`
```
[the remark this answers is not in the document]
danilo-gemoli: What about `mirrors`? Can you propose something?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: Yeah all those operations are indeed being done before we apply `mutateFn` . The func `runPostBuildActions` takes care of running them one after the other in a clean way (no unreadable list of `if {} else if {} ...`). `mutateFn` is the final outcome that is going to be applied **after**.

cmd/multi-arch-builder-controller/main.go
```diff
@@ -33,6 +36,9 @@ func gatherOptions() (*options, error) {
 
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
 	fs.StringVar(&o.dockerCfgPath, "docker-cfg", "/.docker/config.json", "Path of the registry credentials configuration file")
+	fs.Var(&o.namespaces, "namespace", "Namespace the controller watches on, default to all. Can be passed multiple times.")
```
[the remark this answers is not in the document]
danilo-gemoli: I was doing my test watching on a single namespace only but yeah, we have so few builds that is probably not even needed.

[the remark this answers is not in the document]
droslean: Yea, let's not add it.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: The way you are doing it now, its more complicated than the `if else`.

danilo-gemoli: This is how it would look like with no refactoring. Other than being more verbose and with some duplications, the worst part is that you have to pass every parameters you need as `registryURL, targetImageRef string` (list will grow as soon as we add more things to do). 
```golang
// runPostBuildActions ensure some actions are performed after the builds have been spawned.
func (r *reconciler) runPostBuildActions2(mabc *v1.MultiArchBuildConfig, registryURL, targetImageRef string) (mutate func(mabcToMutate *v1.MultiArchBuildConfig)) {
	mutateToFailure := func(mabcToMutate *v1.MultiArchBuildConfig) { mabcToMutate.Status.State = v1.FailureState }
	mutateFuncs := make([]func(mabcToMutate *v1.MultiArchBuildConfig), 0)

	defer func() {
		mutate = func(mabcToMutate *v1.MultiArchBuildConfig) {
			for _, f := range mutateFuncs {
				f(mabcToMutate)
			}
		}
	}()

	// Make sure the builds completed successfully
	if !checkAllBuildsSuccessful(mabc.Status.Builds) {
		mutateFuncs = append(mutateFuncs, mutateToFailure)
		return
	}

	// Assemble and push the manifest list
	mutateFn, success := r.handlePushImageWithManifest(mabc, targetImageRef)
	if mutateFn != nil {
		mutateFuncs = append(mutateFuncs, mutateFn)
	}
	if !success {
		mutateFuncs = append(mutateFuncs, mutateToFailure)
		return
	}

	// Mirror to external registries
	srcImage := fmt.Sprintf("%s/%s", registryURL, targetImageRef)
	mutateFn, success = r.handleMirrorImage(srcImage, mabc)
	if mutateFn != nil {
		mutateFuncs = append(mutateFuncs, mutateFn)
	}
	if !success {
		mutateFuncs = append(mutateFuncs, mutateToFailure)
	}
	return
}
```

droslean: The CR should be updated on every operation. I don't the point of gathering a list of functions and then mutating the object once.  For example, the `checkAllBuildsSuccessful` shouldn't happen inside the postActions. We need to follow the controller's flow. When a build finishes, we update the mabc, this will trigger the reconciliation and we will operate on the object once again. Since now we are adding an extra action, we need to run the manifestPusher first, update the object with the results, and last we should mirror the image to the external registries. This way you keep the flow of the controller. You have to keep in mind that the controller can run in multiple replicas or the single pod can be rescheduled at any time. If you wish to treat the manifest pusher and the mirroring as one step, still the code you shared above, is less complicated than what you tried to introduce in this PR.

danilo-gemoli: Definitely I like this approach of one operation at a time. In that case don't we need to make sure an operation has been performed already, like so https://github.com/openshift/ci-tools/pull/3676#discussion_r1371437766 ?

pkg/controller/multiarchbuildconfig/util.go
```diff
@@ -0,0 +1,28 @@
...
+		}
+	}
+	return nil
+}
+
+func setCondition(mabc *v1.MultiArchBuildConfig, cond *metav1.Condition) {
+	for i := range mabc.Status.Conditions {
+		current := &mabc.Status.Conditions[i]
+		if current.Type == cond.Type {
+			mabc.Status.Conditions[i] = *current
```
droslean, line 23: This is wrong. You shouldn't mutate existing conditions.

pkg/controller/multiarchbuildconfig/mirror.go
```diff
@@ -0,0 +1,122 @@
...
+	registryConfig string
+}
+
+// mirror invokes `oc image mirror` in a new process in order to mirror images
+// using credentials in registryConfig
+func (oci *ocImage) mirror(images []string) error {
+	args := append([]string{
+		"image",
+		"mirror",
+		// When the source is image-registry.openshift-image-registry.svc:5000 the oc client
```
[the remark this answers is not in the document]
danilo-gemoli: Destination is external, what about the source?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -185,6 +180,34 @@ func (r *reconciler) createBuildsForArchitectures(ctx context.Context, mabc *v1.
 	return nil
 }
 
+func (r *reconciler) handlePushImageWithManifest(mabc *v1.MultiArchBuildConfig, targetImageRef string) func(mabcToMutate *v1.MultiArchBuildConfig) {
+	if cond := getCondition(mabc, PushImageManifestDone); cond != nil {
```
droslean, line 184: This shouldn't be happening here.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -120,41 +113,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 	return nil
 }
 
-func (r *reconciler) handleMultiArchBuildConfig(ctx context.Context, mabc *v1.MultiArchBuildConfig) error {
+func (r *reconciler) handleMultiArchBuildConfig(ctx context.Context, mabc *v1.MultiArchBuildConfig) (resultErr error) {
+	var mutateFn func(mabcToMutate *v1.MultiArchBuildConfig)
```
droslean, line 117: We only have 2 functionalities.  You can just update the mabc accordingly on each of the operations. No need to return a mutate function.

pkg/controller/multiarchbuildconfig/mirror.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+	oci.log.WithField("output", cmdOutput.String()).Debug("oc command succeeded")
+	return nil
+}
+
+// handleMirrorImage pushes an image to the locations specified in .spec.external_registries. The image
+// required has to exist on local registry. Returns a function that knows how to properly update the current mabc
+// and a boolean to indicate whether it succeeded or not.
+func (r *reconciler) handleMirrorImage(srcImage string, mabc *v1.MultiArchBuildConfig) func(mabcToMutate *v1.MultiArchBuildConfig) {
```
droslean, line 74: Let's put this in the same file as the rest of the reconciler methods.

pkg/controller/multiarchbuildconfig/util.go
danilo-gemoli: I'm not sure it's wrong, see [here](https://github.com/kubernetes/kubernetes/blob/8f45b64c93563ef682f24d1f6300679d03d946f1/pkg/controller/replicaset/replica_set_utils.go#L153-L165) as an example.

pkg/controller/multiarchbuildconfig/mirror.go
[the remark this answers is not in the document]
droslean: Why do we need to pass the local registry URL for the source images? If this is a requirement from oc, then adding the insecure flag makes sense.

pkg/api/multiarchbuildconfig/v1/types.go
```diff
@@ -33,6 +33,11 @@ type MultiArchBuildConfig struct {
 
 type MultiArchBuildConfigSpec struct {
 	BuildSpec buildv1.BuildConfigSpec `json:"build_spec"`
+	// ExternalRegistries is a list of external registrie URLs the images are
+	// going to be pushed to. Private registries are allows as long as the
+	// mabc controller holds valid credentials.
+	// +kubebuilder:validation:MaxItems=5
```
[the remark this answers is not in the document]
danilo-gemoli: I was not convinced about allowing a user to set as many registries as he wants, so the upper bound.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: Where is should happen?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -120,41 +113,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+
+	defer func() {
```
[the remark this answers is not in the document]
droslean: Yea it accepts a mutate function to make it more generic. Since you are handling each step, you can just update the mabc accordingly by constructing the changes that you need to make.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: What's the point of the function `UpdateMultiArchBuildConfig` if we mutate the mabc in place without using a function?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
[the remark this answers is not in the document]
danilo-gemoli: So we can remove `UpdateMultiArchBuildConfig` entirely then?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: The reconciler should determine which of the operation should run based on the conditions.

pkg/controller/multiarchbuildconfig/util.go
droslean: The example you shared is a completely different case and its happening in the replicaSets for a reason. We don't need to update existing conditions because we need to follow the flow of what actually happened in the mabc.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: UpdateMultiArchBuildConfig is more generic to avoid using multiple functions for different kinds of changes.

pkg/api/multiarchbuildconfig/v1/types.go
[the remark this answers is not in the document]
droslean: Why not? The user should be able to add as many registries as it sees fit. Of course we need the credentials but the limitation is not based on logic. Either way, we will never have more than 2 registries in the current state...

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: By reconciler you mean the `handleMultiArchBuildConfig` function? Because `handlePushImageWithManifest` is a reconciler's method too.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: If we modify the `mabc` without using a mutate function then `UpdateMultiArchBuildConfig` would not be needed anymore. If that's fine, I'm going to remove it.

pkg/controller/multiarchbuildconfig/util.go
danilo-gemoli: > we need to follow the flow of what actually happened in the mabc
That's fine.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -120,41 +113,43 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+		if mutateFn == nil {
```
droslean, line 120: We should never have this case.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: This is not what I am saying. You can run the `UpdateMultiArchBuildConfig` with the mutateFn you need. There is no point of return the mutateFn on each of the two operations. Instead you can mutate the object directly and take care of the error on a higher level.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: See https://github.com/openshift/ci-tools/pull/3676#discussion_r1381690914

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: All I am saying is that you don't need to run the `handlePushImageWithManifest` method if there is no need to do it.

pkg/controller/multiarchbuildconfig/util.go
droslean: Btw, at some point, we need to start looking at how those objects will be maintained in the cluster. I am not sure if applying a new version of the mabc will keep the older conditions

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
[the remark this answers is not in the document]
droslean: Why? What I mean is that you can just return the UpdateMultiArchBuildConfig error. No need to do the defer tricks to append the existing `resultErr`. Either way, If you check my other comments, you can just get rid of the whole defer and update the mabc on each of the method. No need to return the mutateFn to pass it to the defer operation.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: > Instead you can mutate the object directly

I was following the existing approach like in `checkAllBuildsSuccessful`, that returns the mutating function. Do you mean using `UpdateMultiArchBuildConfig` directly inside each of the functions `checkAllBuildsSuccessful`, `handlePushImageWithManifest` and `handleMirrorImage` ? If you meant so, then the whole approach would be:
```golang
func (r *reconciler) handleMultiArchBuildConfig() {
  // ...
  if c := getCondition(PushImageManifestDone); c != nil {
    if err := r.handlePushImageWithManifest(mabc); err != nil {
      return err
    }
  }
  // ...
}
```

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: There is no other operation that updates the conditions. You can just open a switch statement for the latest condition and act accordingly. The `getCondition` now is useless since you don't need the condition for any reason rather than checking if the `PushManifestDone` is done.

pkg/controller/multiarchbuildconfig/mirror.go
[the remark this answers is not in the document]
danilo-gemoli: > Why do we need to pass the local registry URL for the source images?

Cause the local registry works regardless of the cluster the controller is running on. If we're planning to add `multi02` and we hardcode the external registry address `registry.multi-build01.arm-build.devcluster.openshift.com` then it won't work anymore.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: Why a switch statement then? I could just check the last one in that case.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: Yea. Something like that, but it would be better to check the last condition. And if all the builds are OK and there is no `PushImageManifestDone` condition, we run the manifest pusher. Let's just keep the logic simple.

pkg/controller/multiarchbuildconfig/util.go
danilo-gemoli: We can address this problem in a future PR.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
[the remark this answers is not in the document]
danilo-gemoli: See https://github.com/openshift/ci-tools/pull/3676#discussion_r1381726683

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: I mean that we will never have a nil mutateFn.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: Checking the last condition won't be sufficient since after mirroring the image a `MirrorImageManifestDone` condition is added too (we're adding it because we want to follow the flow of an mabc). In that case the reconciler would mirror and update the mabc, then it will be triggered again and, if we just look at the last condition, we would end up using the `manifestPusher` again. That's why we need `getCondition()`.

droslean: That's not entirely true. The only case where we want to run the mirroring is when the last condition is  `MirrorImageManifestDone`. So checking if the last condition makes absolute sense here.

danilo-gemoli: Why? Shouldn't we mirror only when the last condition is `PushImageManifestDone` as mirroring happens after we have assembled a manifest list? Then we are going to mirror and add `MirrorImageManifestDone`.

droslean: Yea exactly. We should mirror when the last condition is `PushImageManifestDone`. There is no other case.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -121,37 +120,44 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
 }
 
 func (r *reconciler) handleMultiArchBuildConfig(ctx context.Context, mabc *v1.MultiArchBuildConfig) error {
+	var mutateFn func(mabcToMutate *v1.MultiArchBuildConfig)
+
 	if mabc.Status.Builds == nil {
 		if err := r.createBuildsForArchitectures(ctx, mabc); err != nil {
 			return fmt.Errorf("couldn't create builds for architectures: %s: %w", strings.Join(r.architectures, ","), err)
 		}
+		if err := v1.UpdateMultiArchBuildConfig(ctx, r.logger, r.client, ctrlruntimeclient.ObjectKey{Namespace: mabc.Namespace, Name: mabc.Name}, mutateFn); err != nil {
```
droslean, line 129: This won't update anything. Also, keep in mind that there is another controller that watches the builds and will update the mabc.Status.Builds either way.

danilo-gemoli: Oh yeah, copy-paste mistake I guess.
> there is another controller that watches the builds

Yes I know but we can probably get rid of it and use the owner references instead. I'll open a follow up PR.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -61,32 +69,23 @@ func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath str
...
 type reconciler struct {
-	logger         *logrus.Entry
-	client         ctrlruntimeclient.Client
-	architectures  []string
-	manifestPusher manifestpusher.ManifestPusher
+	logger           *logrus.Entry
+	client           ctrlruntimeclient.Client
+	architectures    []string
+	manifestPusher   manifestpusher.ManifestPusher
+	dockerConfigPath string
```
droslean, line 86: Is this needed?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -205,3 +274,8 @@ func checkAllBuildsFinished(builds map[string]*buildv1.Build) bool {
 	}
 	return true
 }
+
+func checkPushImageManifestDone(mabc *v1.MultiArchBuildConfig) bool {
```
droslean, line 278:
```suggestion
func isPushImageManifestDone(mabc *v1.MultiArchBuildConfig) bool {
```

droslean (approved):

PR #3683, opened 2023-10-11
Override cluster allocation for agent jobs

andfasano:
Following the patch https://github.com/openshift/release-controller/pull/564, the current one is required to ensure that the agent-based installer jobs will be executed on `build05` cluster, since they are using ofcir


andfasano:
@droslean do you maybe have any idea about the jobs failures?

droslean (approved):

PR #3690, opened 2023-10-13
2-staged pipeline controller: fix branches bug

jmguzik:


jmguzik:
this change has nothing to do with e2e
/override ci/prow/e2e

PR #3700, opened 2023-10-16
Calm Aggregated Disruption Testing

dgoodwin:
- Drop the aggregated P95 should not get worse test
- Pass fixed graceseconds to aggregated percentile disruption tests
- Look at two weeks of disruption data instead of one for aggregated disruption percentile testing
- Remove the zero-disruption should not be worse aggregated test

More details on each commit and in the investigation doc shared with TRT.


dgoodwin:
Dropped the two weeks of data change, this generated a lot of discussion and we're not totally sure what the right play is there yet. Keeping in the commit list in case we ever want it back.

PR #3705, opened 2023-10-17
Add openshift-cherrypick-robot org membership check in check-gh-automation

Toma5OD:
Ensures openshift-cherrypick-robot is an org member for repos using the cherrypick plugin. Resolves [DPTP-3603](https://issues.redhat.com//browse/DPTP-3603).

[DPTP-3603](https://issues.redhat.com/browse/DPTP-3603)

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
@@ -45,10 +45,41 @@ type JobRunAggregatorAnalyzerOptions struct {
 	prowJobClient       *prowjobclientset.Clientset
 	jobStateQuerySource string
 	prowJobMatcherFunc  jobrunaggregatorlib.ProwJobMatcherFunc
+
+	staticJobRunIdentifiers []jobrunaggregatorlib.JobRunIdentifier
+}
+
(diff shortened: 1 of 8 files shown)


cmd/check-gh-automation/README.md
```diff
@@ -1,5 +1,6 @@
 # Check GH Automation
-A tool to check that our bots (`openshift-merge-robot` and `openshift-ci-robot`) have access to repositories that have CI configured.
+A tool to check that our bots (`openshift-merge-robot`, `openshift-ci-robot`, and `openshift-cherrypic
+k-robot`) have access to repositories that have CI configured.
```
smg247, line 3: I realize that when viewing the file on the browser this makes no difference, but there is a hard return in the middle of a word here.

cmd/check-gh-automation/main.go
```diff
@@ -157,6 +157,18 @@ func checkRepos(repos []string, bots []string, ignore sets.Set[string], client a
 			}
 		}
 
+		// Add a check for openshift-cherrypick-robot
+		isCherryPickRobotMember, err := client.IsMember(org, "openshift-cherrypick-robot")
```
smg247, line 161: The `openshift-cherrypick-robot` does not need to be a member in every org, only those with the `cherrypick` plugin enabled. You will need to check the prow configs for the org to see if the plugin is enabled prior to checking for membership

PR #3728, opened 2023-10-27
`ci-operator-configresolver`: merge additional configs for multi-pr payload testing

smg247:
The `base_rpm_images`, `canonical_go_repository`, `resources`, `images`, and `releases` all need to be merged as well for multi-pr payload testing. I have done so following the existing pattern. 

We also need to reference the org and repo with a `.` instead of a `/` as the slash is not a valid image name. This is the cleanest way to do this.


pkg/registry/server/server.go
```diff
@@ -315,6 +321,39 @@ func ResolveAndMergeConfigsAndInjectTest(configs Getter, resolver Resolver, reso
...
+			for step, resources := range config.Resources {
+				if step == "*" { // * is special, and we will only pull from one config to merge, this is "last in wins" we may need to do better in the future
+					mergedConfig.Resources[step] = resources
+				} else {
+					stepWithRef := fmt.Sprintf("%s-%s", step, ref)
+					mergedConfig.Resources[stepWithRef] = resources
+				}
+			}
+			if len(config.Releases) > 0 && len(mergedConfig.Releases) == 0 {
+				// Since the release configs "should" be identical, we can just use the first one we come across
```
danilo-gemoli, line 339: This is by design, right?

smg247: Yes, I am reasonably confident it will work as is. There is a chance we will have to implement smarter logic later though.

danilo-gemoli:
There are some edge cases like `releases:` and `*` on `resources:` but I'm happy with those because we at least have some comments for the future.
/lgtm

PR #3750, opened 2023-11-06
`ci-operator-configresolver`: handle a few simple `raw_step` ref substitutions

smg247:
We can't reasonably handle all `raw_steps` when merging configs for multi-pr payload testing, but there are a few scenarios that are simple substitutions that may get us some useful functionality.

For: https://issues.redhat.com/browse/DPTP-3685

/cc @danilo-gemoli

### DIFF
--- a/pkg/registry/server/server.go
+++ b/pkg/registry/server/server.go
@@ -354,6 +354,28 @@ func ResolveAndMergeConfigsAndInjectTest(configs Getter, resolver Resolver, reso
 				image.Ref = ref
 				mergedConfig.Images = append(mergedConfig.Images, image)
 			}
+
+			// Attempt to handle a few simple raw_step types on a best-effort basis
+			for i := range config.RawSteps {
+				rawStep := config.RawSteps[i]
+				modifiedStep := rawStep.DeepCopy()
+				if rawStep.RPMImageInjectionStepConfiguration != nil {
+					modifiedStep.RPMImageInjectionStepConfiguration.To = api.PipelineImageStreamTagReference(fmt.Sprintf("%s-%s", rawStep.RPMImageInjectionStepConfiguration.To, ref))
+					modifiedStep.RPMImageInjectionStepConfiguration.From = api.PipelineImageStreamTagReference(fmt.Sprintf("%s-%s", rawStep.RPMImageInjectionStepConfiguration.From, ref))
+				} else if rawStep.ProjectDirectoryImageBuildStepConfiguration != nil {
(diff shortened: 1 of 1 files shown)


pkg/registry/server/server.go
```diff
@@ -354,6 +354,28 @@ func ResolveAndMergeConfigsAndInjectTest(configs Getter, resolver Resolver, reso
...
 			}
+
+			// Attempt to handle a few simple raw_step types on a best-effort basis
+			for i := range config.RawSteps {
+				rawStep := config.RawSteps[i]
+				modifiedStep := rawStep.DeepCopy()
+				if rawStep.RPMImageInjectionStepConfiguration != nil {
+					modifiedStep.RPMImageInjectionStepConfiguration.To = api.PipelineImageStreamTagReference(fmt.Sprintf("%s-%s", rawStep.RPMImageInjectionStepConfiguration.To, ref))
+					modifiedStep.RPMImageInjectionStepConfiguration.From = api.PipelineImageStreamTagReference(fmt.Sprintf("%s-%s", rawStep.RPMImageInjectionStepConfiguration.From, ref))
+				} else if rawStep.ProjectDirectoryImageBuildStepConfiguration != nil {
```
danilo-gemoli, line 365: Shouldn't be `modifiedStep.ProjectDirectoryImageBuildStepConfiguration.Ref = ref` set too?

pkg/registry/server/server.go
```diff
@@ -354,6 +354,28 @@ func ResolveAndMergeConfigsAndInjectTest(configs Getter, resolver Resolver, reso
 				image.Ref = ref
 				mergedConfig.Images = append(mergedConfig.Images, image)
 			}
+
+			// Attempt to handle a few simple raw_step types on a best-effort basis
+			for i := range config.RawSteps {
```
danilo-gemoli, line 359: The lines are quite long and they don't even fit my screen. It would be nice if you could break them apart like so:
```golang
to := fmt.Sprintf("%s-%s", rawStep.RPMImageInjectionStepConfiguration.To, ref)
modifiedStep.RPMImageInjectionStepConfiguration.To = api.PipelineImageStreamTagReference(to)
from := fmt.Sprintf("%s-%s", rawStep.RPMImageInjectionStepConfiguration.From, ref)
modifiedStep.RPMImageInjectionStepConfiguration.From = api.PipelineImageStreamTagReference(from)
```

pkg/registry/server/server.go
smg247: yes, it isn't necessary since the step is already formed, but it is better to have it set. I will update that.

PR #3753, opened 2023-11-07
Add cluster_profile for redhat-openshift-ecosystem

mrhillsman:
Related to: openshift/release#45414

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1372,6 +1372,7 @@ const (
 	ClusterProfileQuayAWS               ClusterProfile = "quay-aws"
 	ClusterProfileAWSEdgeInfra          ClusterProfile = "aws-edge-infra"
 	ClusterProfileRosaAws               ClusterProfile = "rosa-aws"
+	ClusterProfileRHOpenShiftEcosystem  ClusterProfile = "rh-openshift-ecosystem"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -1249,128 +1249,129 @@ type ContainerTestConfiguration struct {
 type ClusterProfile string
 
 const (
-	ClusterProfileAWS                   ClusterProfile = "aws"
```
jmguzik, line 1252: What happened here so we have so many deletions and additions?

mrhillsman: It is because of the spacing/alignment. Due to the length of the name:

ClusterProfileRedHatOpenShiftEcosystem ClusterProfile = "redhat-openshift-ecosystem"

The other ClusterProfile above are pushed to the right to align.

jmguzik: I would suggest a shorter name, using shortcut then. It would be safer for us (TP) to know only several lines were changed and what PRs introduced what cluster profiles in the past from blame. Realigning that now will change this situation.

mrhillsman: @jmguzik fixed

acornett21: The linting in this project even enforces this change, and is failing bc of the rollback.

mrhillsman: Wrong push fixed

PR #3756, opened 2023-11-08
`multi-arch-builder-controller`: Handle deletion gracefully

danilo-gemoli:
Create a relationship between a `mabc` and the builds it spawns by setting the `ownerReferences` field.
In this way it is possible to clean everything up properly with `oc delete mabc/foo --cascade=foreground`.

/cc @droslean


pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -83,6 +86,7 @@ type reconciler struct {
 	architectures  []string
 	manifestPusher manifestpusher.ManifestPusher
 	imageMirrorer  imageMirrorer
+	scheme         *runtime.Scheme
```
droslean, line 89: No need to add this here.

danilo-gemoli: I need it for `SetControllerReference`

droslean: You need to make sure that the schema has the mabc and the builds. You don't need to grab the schema of the manager to pass it out. Instead handling schemas is related to the runtime. You can do it in an `init()` function.

droslean: For example: https://github.com/openshift/ci-tools/blob/master/pkg/steps/template.go#L34-L43

danilo-gemoli: > You need to make sure that the schema has the mabc and the builds

Yes, we already do that for the manager [main.go#L74-L80](https://github.com/openshift/ci-tools/blob/10cdb1f73183b7a4355ebd14865ea36dfa2a0d88/cmd/multi-arch-builder-controller/main.go#L74-L80).
I'm following one of the controller-runtime [examples](https://github.com/kubernetes-sigs/controller-runtime/blob/9f9f369134d82816648b691c168008c2d0aa9bfe/examples/crd/main.go#L125) about reusing the same schema of the manager.

droslean (approved):

PR #3761, opened 2023-11-10
`multi-arch-builder-controller`: Remove the builds reconciler

danilo-gemoli:
Remove the `builds` controller and handle the reconciliation with the `mabc` controller only.
The controller gets notified for children `builds` therefore it enqueues requests for the related parent `mabc`.
The `mabc` CRD gets simplified a lot as it won't store the builds status anymore.

/cc @droslean 


pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -263,17 +272,42 @@ func (r *reconciler) handleMirrorImage(ctx context.Context, targetImageRef strin
...
+		return nil, fmt.Errorf("failed to create requirement: %w", err)
+	}
+	listOpts := ctrlruntimeclient.ListOptions{LabelSelector: labels.NewSelector().Add(*requirement)}
+	if err := r.client.List(ctx, &builds, &listOpts); err != nil {
+		return nil, fmt.Errorf("failed to list builds: %w", err)
+	}
+	return &builds, nil
+}
+
+// missingArchitecture returns a set of architectures which no builds have been created for
```
droslean, line 288: This shouldn't be the case. When we will try to create the builds, if we won't create builds for all architectures we should fail.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -45,36 +47,35 @@ const (
...
+		CreateFunc:  func(e event.CreateEvent) bool { return true },
+		DeleteFunc:  func(event.DeleteEvent) bool { return false },
+		UpdateFunc:  func(e event.UpdateEvent) bool { return true },
+		GenericFunc: func(event.GenericEvent) bool { return false },
+	}
+
+	if err := ctrlruntime.NewControllerManagedBy(mgr).
+		WithOptions(controller.Options{MaxConcurrentReconciles: 1}).
+		For(&v1.MultiArchBuildConfig{}, builder.WithPredicates(mabcPredicateFuncs)).
+		Owns(&buildv1.Build{}, builder.WithPredicates(buildPredicateFuncs)).
```
droslean, line 69: We need only the builds that have been created by us with specific labels.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -126,18 +127,26 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
-			return fmt.Errorf("couldn't create builds for architectures: %s: %w", strings.Join(r.architectures, ","), err)
+	builds, err := r.listBuilds(ctx, mabc.Name)
+	if err != nil {
+		return fmt.Errorf("couldn't list builds: %w", err)
+	}
+
+	// Check what builds are missing and create them. createBuilds func may fails to create a build
+	// so we could end up having less builds than available architectures. The following code
+	// make sure the controller handle such a scenario.
+	if missingArchitectures := r.missingArchitectures(builds); missingArchitectures.Len() > 0 {
```
droslean, line 138: There shouldn't exist any missing architectures builds.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -263,17 +272,42 @@ func (r *reconciler) handleMirrorImage(ctx context.Context, targetImageRef strin
 	return nil
 }
 
-func checkAllBuildsSuccessful(builds map[string]*buildv1.Build) bool {
-	for _, build := range builds {
+func (r *reconciler) listBuilds(ctx context.Context, mabcName string) (*buildv1.BuildList, error) {
```
droslean, line 275: I don't think that we should do that. I added the `status.builds` field in `mabc` for exactly this reason. Otherwise we will start spamming this call on every reconciliation. When the build is being created/updated, the controller updates the mabc and that's it. Also, listing those builds everytime beats the whole purpose of this PR.

danilo-gemoli: I meant this PR as removing the `build` reconciler and rely only on the `mabc` one.
There won't be any spamming for two reasons:
- we don't operate at huge scale, we can't optimize without having issues or numbers at hands
- most importantly, the controller makes heavy use of caching (see [pkg/cache/cache.go](https://github.com/kubernetes-sigs/controller-runtime/blob/3de96247226760a0b7c4c6c36e688d1cb41609b8/pkg/cache/cache.go)), the client doesn't always talk with the API server directly, especially for `GET` calls

This mechanism is well consolidated it won't overload the API Server unless we go really crazy in terms of build counts.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: Should we fail and mark the `mabc` as failed?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: It's true but the point is the reconciler, using this configuration, is going to enqueue not any build but only those which belong to an `mabc` we created (builds having the label we want), that's thanks to the `ownerReferences` we set and `Owns()`. I could add the extra check on the label but it won't filter anything I guess.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: The problem is that we will trigger the reconciliation for all builds running in the cluster. Removing the build reconciler, to my eyes, will make things worse. Now the build reconciler will be triggered only for the builds that were created by the controller. We update the mabc and that's it. I am not convinced why we should go ahead with this PR yet.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: Yep.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
droslean: That is fine. Since the reconciler will be triggered, why do we need to list the builds since the event was triggered for a specific build instead?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
```diff
@@ -25,41 +26,115 @@ import (
...
+			Labels: map[string]string{
+				v1.MultiArchBuildConfigNameLabel: bb.mabcName,
+				v1.MultiArchBuildConfigArchLabel: bb.arch,
+			},
+		},
+		Status: buildv1.BuildStatus{Phase: bb.phase},
+	}
+}
+
+func TestMain(m *testing.M) {
```
droslean, line 85: I don't see any point in having this test.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
```diff
@@ -25,41 +26,115 @@ import (
...
+}
+
+type buildBuilder struct {
+	name     string
+	arch     string
+	phase    buildv1.BuildPhase
+	mabcName string
+}
+
+func NewBuildBuilder() *buildBuilder {
```
droslean, line 48: Why do we need this, where you can just pass the build directly to the test case?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: Because the `mabc` owns those `builds` and it needs to check their state to decide how to move ahead.
As an example:
- one of the builds controlled by any `mabc`s fails
- the controller gets notified and enqueue a new request for the parent `mabc` and it calls `r.reconcile()` 
- the reconciler lists the builds, notices one failed and mark the `mabc` as failed too

Again, if your concern is about spamming, I'd reassure you saying that `r.client.List(ctx, &builds, &listOpts)` it's likely to hit the manager cache rather than the API Server.
Furthermore, we are going to ignore any build not related to an `mabc`, so we won't get notified every time.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
danilo-gemoli: > The problem is that we will trigger the reconciliation for all builds running in the cluster

No that's not true, see comment above [here](https://github.com/openshift/ci-tools/pull/3761#discussion_r1389606092)

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
danilo-gemoli: Refactor build definition and save some lines. I can change if you prefer the build declaration to be inlined, I'm not opinionated here.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
danilo-gemoli: It defines and configure a `scheme` that is common to a couple of test. Again, I can change this too if you prefer it to be declared separately.

droslean: You can load the scheme in the init function. Can you try this instead?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
droslean: Let's utilize this not only in the tests and keep it.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
danilo-gemoli: `TestMain` should be used to perform extra setup that is common to more than one tests and it runs only in tests, whereas `init` would always run. If you don't like `TestMain` I would rather setup the scheme in each test I need it than creating `init` in `multiarchbuildconfig_test.go`. Does it sound reasonable to you?

droslean: There is no point of running a TestMain. If you need an higher layer of tests, you can create integration tests. Running the testMain just for load the schema is also no needed and its overkill, since you can do it with the init.

droslean (approved):

PR #3764, opened 2023-11-13
`ci-operator`: only determine the absolute path to `.ci-operator.yaml` when a ref has been defined

smg247:
`cluster-bot` sets the `path_alias` [here](https://github.com/openshift/ci-chat-bot/blob/af9a327dc8ceaef8cf17654894322eb5542159dc/pkg/manager/prow.go#L788), and it doesn't interact well with the standard codeMountPath that is defined. We don't need to generate an absolute path when we are only cloning one ref anyways, so there is no need to run through this logic in those scenarios.


bradmwilliams (approved):
/lgtm

PR #3776, opened 2023-11-16
WIP opt-out the registry and lease options and refactor the related code

droslean:
This PR just moves the code around. 

It opt-out the registry options and refactor the related code to avoid import cycles.


droslean:
/close

There is a chain dependency between all of those changes. I will open multiple small PRs.

PR #3782, opened 2023-11-17
`ci-operator-configresolver`: merge '*' resources when merging configs, and remove limits

smg247:
The naive way we were handling the '*' entry in the merged resources map was causing issues in certain scenarios. We cannot have `limits` set when merging as this can result in a configuration where the request is lower than the limit. We should also use the higher of the request values in order to always request enough of the resource.

### DIFF
--- a/pkg/registry/server/server.go
+++ b/pkg/registry/server/server.go
@@ -335,10 +335,12 @@ func ResolveAndMergeConfigsAndInjectTest(configs Getter, resolver Resolver, reso
 							existingValue, err := resource.ParseQuantity(existing.Requests[resourceType])
 							if err != nil {
 								logger.WithError(err).Warnf("couldn't parse existing '%s' resource quantity", resourceType)
+								return
 							}
 							value, err := resource.ParseQuantity(resources.Requests[resourceType])
 							if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/registry/server/server.go
```diff
@@ -328,8 +329,29 @@ func ResolveAndMergeConfigsAndInjectTest(configs Getter, resolver Resolver, reso
...
+						replaceIfGreater := func(resourceType string) {
+							existingValue, err := resource.ParseQuantity(existing.Requests[resourceType])
+							if err != nil {
+								logger.WithError(err).Warnf("couldn't parse existing '%s' resource quantity", resourceType)
+							}
+							value, err := resource.ParseQuantity(resources.Requests[resourceType])
+							if err != nil {
+								logger.WithError(err).Warnf("couldn't parse '%s' resource quantity", resourceType)
+							}
+							if existingValue.Cmp(value) < 0 { // This value is higher than existing
```
danilo-gemoli, line 343: What happens here if one of the previous two `ParseQuantity` fails?

smg247: Ah that would likely lead to a stack trace. It _shouldn't_ ever happen, but we know how that goes. I will add a check around that.

PR #3796, opened 2023-11-21
qci-appci: get token with the scope param

hongkailiu:
When the repo is private, we need the scope when applying for the token.

/cc @openshift/test-platform


hongkailiu:
/test e2e

Applied the built image and it works.

PR #3802, opened 2023-11-21
merge ci-operator configs for all payload tests

smg247:
This solves issues where multiple images share the same name as the endpoint to merge configs appends a suffix to each of them. With this change we will be able to run payload tests with `hypershift` and in the `installer` repo, which currently fail during graph resolution due to duplicate image names.

For: https://issues.redhat.com/browse/DPTP-3751


smg247:
This can wait until after I am back from leave in January.

PR #3807, opened 2023-11-24
Ci scheduling webhook

bear-redhat:
Worked done by @jupierce


redeploy.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
jupierce, line 1: This is just a hack script and should be removed. Actual build and deploy should be like other images.

cmd/ci-scheduling-webhook/cloudfront/generate_range_array.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
+
+from typing import List, Dict, Tuple
```
jupierce, line 3: Everything in cloudfront/ should be removed. This content is in https://github.com/openshift/ci-tools/pull/3429 .

cmd/ci-scheduling-webhook/testing/builds-admission-review-request.json
```diff
@@ -0,0 +1,61 @@
+{
```
jupierce, line 1: Everything under `testing/` are hack testing scripts used during development and can be removed. They are too custom to have value to commit them.

jupierce (changes requested):

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/openshift/ci-tools
 
-go 1.20
+go 1.21
+
+toolchain go1.21.3
```
hongkailiu, line 5: Could we do it without this line?

bear-redhat: no, tried.

bear-redhat: i manually removed the line as you suggested. lets see

PR #3812, opened 2023-11-29
`rehearsals`: Add image mirroring to quay

Prucek:
This PR implements the mirroring of images to quay.io introduced by a PR in the rehearsals. 

Resolves: [DPTP-3746](https://issues.redhat.com/browse/DPTP-3746)
Deployment change: https://github.com/openshift/release/pull/48638
/cc @hongkailiu

### DIFF
--- a/cmd/pj-rehearse/server.go
+++ b/cmd/pj-rehearse/server.go
@@ -17,6 +17,7 @@ import (
 	"k8s.io/test-infra/prow/pod-utils/gcs"
 
 	"github.com/openshift/ci-tools/pkg/config"
+	quayiociimagesdistributor "github.com/openshift/ci-tools/pkg/controller/quay_io_ci_images_distributor"
 	"github.com/openshift/ci-tools/pkg/rehearse"
 )
 
@@ -134,6 +135,19 @@ func serverFromOptions(o options) (*server, error) {
 	rehearsalConfig := rehearsalConfigFromOptions(o)
 	rehearsalConfig.ProwjobNamespace = c.ProwJobNamespace
 	rehearsalConfig.PodNamespace = c.PodNamespace
+	rehearsalConfig.MirrorOptions = quayiociimagesdistributor.OCImageMirrorOptions{
+		RegistryConfig:  o.registryConfig,
+		ContinueOnError: true,
+		MaxPerRegistry:  20,
+		BatchSize:       10,
+		DryRun:          false,
+	}
+	ocClientFactory := quayiociimagesdistributor.NewClientFactory()
+	quayIOImageHelper, err := ocClientFactory.NewClient()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to create QuayIOImageHelper")
+	}
+	rehearsalConfig.QuayIOImageHelper = quayIOImageHelper
 
 	return &server{
 		ghc:             ghc,
@@ -377,7 +391,7 @@ func (s *server) handlePotentialCommands(pullRequest *github.PullRequest, commen
 						continue
 					}
 
-					success, err := rc.RehearseJobs(candidate, candidatePath, prRefs, imageStreamTags, presubmitsToRehearse, changedTemplates, changedClusterProfiles, logger)
+					success, err := rc.RehearseJobs(candidate, candidatePath, prRefs, imageStreamTags, s.rehearsalConfig.MirrorOptions, s.rehearsalConfig.QuayIOImageHelper, presubmitsToRehearse, changedTemplates, changedClusterProfiles, logger)
 					if err != nil {
 						logger.WithError(err).Error("couldn't rehearse jobs")
 						s.reportFailure("failed to create rehearsal jobs", err, org, repo, user, number, true, false, logger)
--- a/images/pj-rehearse/Dockerfile
+++ b/images/pj-rehearse/Dockerfile
@@ -3,4 +3,5 @@ LABEL maintainer="nmoraiti@redhat.com"
 
 RUN yum install -y git && yum clean all
 ADD pj-rehearse /usr/bin/pj-rehearse
+ADD /usr/bin/oc /usr/bin/oc
 ENTRYPOINT ["/usr/bin/pj-rehearse"]
--- a/pkg/rehearse/rehearse.go
+++ b/pkg/rehearse/rehearse.go
@@ -20,7 +20,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
 	prowconfig "k8s.io/test-infra/prow/config"
 	"k8s.io/test-infra/prow/flagutil"
@@ -101,8 +100,8 @@ func RehearsalCandidateFromPullRequest(pullRequest *github.PullRequest, baseSHA 
 	}
 }
 
-func (rc RehearsalCandidate) createRefs() *pjapi.Refs {
-	return &pjapi.Refs{
+func (rc RehearsalCandidate) createRefs() *prowapi.Refs {
+	return &prowapi.Refs{
 		Org:     rc.org,
 		Repo:    rc.repo,
 		BaseRef: rc.base.ref,
@@ -192,7 +191,7 @@ func (r RehearsalConfig) DetermineAffectedJobs(candidate RehearsalCandidate, can
 	return filterPresubmits(presubmits, logger), filterPeriodics(periodics, logger), changedTemplates, changedClusterProfiles, nil
 }
 
-func (r RehearsalConfig) SetupJobs(candidate RehearsalCandidate, candidatePath string, presubmits config.Presubmits, periodics config.Periodics, rehearsalTemplates, rehearsalClusterProfiles *ConfigMaps, limit int, logger *logrus.Entry) (*config.ReleaseRepoConfig, *pjapi.Refs, apihelper.ImageStreamTagMap, []*prowconfig.Presubmit, error) {
+func (r RehearsalConfig) SetupJobs(candidate RehearsalCandidate, candidatePath string, presubmits config.Presubmits, periodics config.Periodics, rehearsalTemplates, rehearsalClusterProfiles *ConfigMaps, limit int, logger *logrus.Entry) (*config.ReleaseRepoConfig, *prowapi.Refs, apihelper.ImageStreamTagMap, []*prowconfig.Presubmit, error) {
 	resolver, err := r.createResolver(candidatePath)
 	if err != nil {
 		return nil, nil, nil, nil, err
@@ -270,7 +269,7 @@ func (r RehearsalConfig) AbortAllRehearsalJobs(org, repo string, number int, log
 	}
 
 	selector := labelSelectorForRehearsalJobs(org, repo, number)
-	jobs := &pjapi.ProwJobList{}
+	jobs := &prowapi.ProwJobList{}
 	err = pjclient.List(context.TODO(), jobs, selector, ctrlruntimeclient.InNamespace(r.ProwjobNamespace))
 	if err != nil {
 		logger.WithError(err).Error("failed to list prowjobs for pr")
@@ -308,7 +307,7 @@ func labelSelectorForRehearsalJobs(org, repo string, prNumber int) ctrlruntimecl
 }
 
 // RehearseJobs returns true if the jobs were triggered and succeed
-func (r RehearsalConfig) RehearseJobs(candidate RehearsalCandidate, candidatePath string, prRefs *pjapi.Refs, imageStreamTags apihelper.ImageStreamTagMap, presubmitsToRehearse []*prowconfig.Presubmit, rehearsalTemplates, rehearsalClusterProfiles *ConfigMaps, logger *logrus.Entry) (bool, error) {
+func (r RehearsalConfig) RehearseJobs(candidate RehearsalCandidate, candidatePath string, prRefs *prowapi.Refs, imageStreamTags apihelper.ImageStreamTagMap, presubmitsToRehearse []*prowconfig.Presubmit, rehearsalTemplates, rehearsalClusterProfiles *ConfigMaps, logger *logrus.Entry) (bool, error) {
 	buildClusterConfigs, prowJobConfig := r.getBuildClusterAndProwJobConfigs(logger)
 	pjclient, err := NewProwJobClient(prowJobConfig, r.DryRun)
 	if err != nil {
--- a/pkg/rehearse/rehearse.go
+++ b/pkg/rehearse/rehearse.go
@@ -15,6 +15,7 @@ import (
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
@@ -34,6 +35,7 @@ import (
 	apihelper "github.com/openshift/ci-tools/pkg/api/helper"
 	testimagestreamtagimportv1 "github.com/openshift/ci-tools/pkg/api/testimagestreamtagimport/v1"
 	"github.com/openshift/ci-tools/pkg/config"
+	quayiociimagesdistributor "github.com/openshift/ci-tools/pkg/controller/quay_io_ci_images_distributor"
 	"github.com/openshift/ci-tools/pkg/diffs"
 	"github.com/openshift/ci-tools/pkg/load"
 	"github.com/openshift/ci-tools/pkg/registry"
@@ -59,6 +61,9 @@ type RehearsalConfig struct {
 	NormalLimit int
 	MoreLimit   int
 	MaxLimit    int
+
+	MirrorOptions     quayiociimagesdistributor.OCImageMirrorOptions
+	QuayIOImageHelper quayiociimagesdistributor.OCClient
 
 	StickyLabelAuthors sets.Set[string]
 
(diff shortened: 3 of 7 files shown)


pkg/rehearse/rehearse.go
```diff
@@ -14,13 +14,12 @@ import (
 	"golang.org/x/sync/errgroup"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
-	pjapi "k8s.io/test-infra/prow/apis/prowjobs/v1"
```
hongkailiu, line 23: This is a good one. :)

pkg/rehearse/rehearse.go
```diff
@@ -581,11 +586,7 @@ func setupDependencies(
 // Allow manipulating the speed of time for tests
 var second = time.Second
 
-func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client, ists apihelper.ImageStreamTagMap, clusterName, namespace string, istImportClient ctrlruntimeclient.Client, log *logrus.Entry) error {
```
hongkailiu, line 584: Good work!
I believe that you find the right place to implement the feature.
However, what is in my mind is to do the mirroring in addition to the existing imports, instead of replacing it which is probably the next card.
I probably did not make this clear in the card and revised it a couple of mins ago.

In general, we might want to do another function `ensureImageInQCI()` where we did the similar thing:
- check if the corresponding image is there or not (and stale or not)
- if not up to date, do an oc-image-mirror

The logic is more or less the same as reconcile function in `ci-images-mirror`.

https://github.com/openshift/ci-tools/blob/09426d1c21edd095bac4308f9473e511a020c7a1/pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go#L132

pkg/rehearse/rehearse.go
```diff
@@ -581,11 +586,7 @@ func setupDependencies(
...
-	if clusterName == appCIContextName {
```
hongkailiu, line 585: This is the tricky part.
Let us verify this: Found out where the request is generated with `cluster` encoded. IIRC, it is `ci-operator`.
I believe we just need to do the mirror in one cluster because the target would be the same on QCI for the requests on different clusters.

pkg/rehearse/rehearse.go
Prucek: I was not sure if it is on purpose or not, but I think it shouldn't be like that

pkg/rehearse/rehearse.go
Prucek: So if I understand you correctly, we want to do the opposite? After the import is created, check if this image exists in quay and mirror it there. Not mirror from quay to the local registry, but from local registry to quay?

pkg/rehearse/rehearse.go
Prucek: I don't understand why only from one cluster. I think as it is right now the `ImageStreamImport` created in the `test_images_distributor.go` imports the IS only to one cluster, not all.

hongkailiu: You can compare those on the cluster: The difference is `spec.clusterName`.

```console
oc --context build01 get testimagestreamtagimports  -n ci
oc --context app.ci get testimagestreamtagimports  -n ci
```

pkg/rehearse/rehearse.go
```diff
@@ -630,13 +636,91 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+			}
+
 			return nil
 		})
 	}
 
 	return g.Wait()
 }
 
+func ensureImageInQCI(ctx context.Context, client ctrlruntimeclient.Client, requiredImageStreamTag types.NamespacedName, registryConfig string, log *logrus.Entry) error {
```
hongkailiu, line 650: I would also try to dry out a function shared by this tool and `ci-images-mirror`. They look quite similar to each other.

pkg/rehearse/rehearse.go
hongkailiu: Let us track down `testimagestreamtagimports` together.
Producer 1: the rehearsal tool

https://github.com/openshift/ci-tools/blob/7f1ce7102624689fe0a1d695232ea5e2266c7ef2/pkg/rehearse/rehearse.go#L608-L620

It sets up its `.spec.cluster` with a loop for each build-farm.

Producer 2: `ci-operator`

https://github.com/openshift/ci-tools/blob/7f1ce7102624689fe0a1d695232ea5e2266c7ef2/pkg/defaults/defaults.go#L1057-L1072

It does not set up its `.spec.cluster` and it creates only one object on the cluster where it is running.

Consumer: `test_images_distributor`. As displayed between, they have different handlers accordingly for the above two cases.

https://github.com/openshift/ci-tools/blob/7f1ce7102624689fe0a1d695232ea5e2266c7ef2/pkg/controller/test-images-distributor/test_images_distributor.go#L90-L104

One more thing (not relevant to this PR tho): There are two sources of the events to those handlers.

https://github.com/openshift/ci-tools/blob/7f1ce7102624689fe0a1d695232ea5e2266c7ef2/pkg/controller/test-images-distributor/test_images_distributor.go#L115-L132

I know this is quite complicated (i must admit I might have imcomplete/wrong understanding even though I have read the code many rounds, including the one for write this comment down).
So the general idea is:
When we need to create an import on a tag on a cluster, we create `testimagestreamtagimports` which is watched by `test_images_distributor` which makes the import takes place.
That is what the producer does, including the rehearsal tool.
Your task is ... make sure the tag is available on QCI by oc-mirror cmd.
You need to do it only once because the target image is the same image on QCI.

cmd/ci-images-mirror/main.go
```diff
@@ -177,21 +173,7 @@ func main() {
...
-
-	ocClientFactory := quayiociimagesdistributor.NewClientFactory()
-	quayIOImageHelper, err := ocClientFactory.NewClient()
-	if err != nil {
-		logrus.WithError(err).Fatal("failed to create QuayIOImageHelper")
-	}
-
-	mirrorConsumerController := quayiociimagesdistributor.NewMirrorConsumer(mirrorStore, quayIOImageHelper, opts.registryConfig, opts.dryRun)
-	interrupts.Run(func(ctx context.Context) { execute(ctx, mirrorConsumerController) })
+	mirrorStore, quayIOImageHelper := mirrorstoreserver.MirrorStoreServer(opts.port, opts.gracePeriod, opts.registryConfig, false, opts.dryRun)
```
Prucek, line 176: Here I wanted to use this server also for the `pj-rehearse` so I factored it out to a separate file/package, so I could use it there, maybe it can be done better.

cmd/pj-rehearse/server.go
```diff
@@ -135,10 +141,14 @@ func serverFromOptions(o options) (*server, error) {
...
+
 	return &server{
-		ghc:             ghc,
-		gc:              gc,
-		rehearsalConfig: rehearsalConfig,
+		ghc:               ghc,
+		gc:                gc,
+		rehearsalConfig:   rehearsalConfig,
+		mirrorStore:       mirrorStore,
+		quayIOImageHelper: quayIOImageHelper,
```
Prucek, line 151: Here I added the `MirrorStoreServer` to the already existing `pj-rehearse` server and propagated it to the `RehearseJobs` method

pkg/controller/quay_io_ci_images_distributor/mirror.go
```diff
@@ -100,7 +101,9 @@ func (c *MirrorConsumerController) Run(ctx context.Context) error {
 			continue
 		}
 		if len(mirrors) == 0 {
-			c.logger.Debug("Waiting for mirror tasks ...")
+			if !c.isPjrehearse {
```
Prucek, line 104: I didn't want to have the logs of  `Waiting for mirror tasks ...` in the `pj-rehearse` it is disturbing there

cmd/ci-images-mirror/main.go
hongkailiu: Couldnt this work?

```
        ocClientFactory := quayiociimagesdistributor.NewClientFactory()
	quayIOImageHelper, err := ocClientFactory.NewClient()
	if err != nil {
		logrus.WithError(err).Fatal("failed to create QuayIOImageHelper")
	}
        // we know the requiredImageStreamTags which is basically the pairs belew
        quayIOImageHelper.ImageMirror(pairs, c.options)
```

The MirrorStore in the existing tool is there because of many tags to mirror and we want to do it as efficient as possible.
Here, for the researsal tool, there are only a couple of images to mirror.
Let us do the mirror blindly without a MirrorStore.
The code would be much simpler and we do not need the mega refactoring.

images/pj-rehearse/Dockerfile
```diff
@@ -3,4 +3,5 @@ LABEL maintainer="nmoraiti@redhat.com"
 
 RUN yum install -y git && yum clean all
 ADD pj-rehearse /usr/bin/pj-rehearse
+ADD /usr/bin/oc /usr/bin/oc
```
hongkailiu, line 6: You need to make this source file available in the build.
Example:
https://github.com/openshift/release/blob/8c3c1dd5615f3c341781c3acd0dfe146748fa4eb/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml#L222-L223

pkg/rehearse/rehearse.go
```diff
@@ -630,14 +639,38 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+	ocClientFactory := quayiociimagesdistributor.NewClientFactory()
+	quayIOImageHelper, err := ocClientFactory.NewClient()
+	if err != nil {
+		logrus.WithError(err).Fatal("failed to create QuayIOImageHelper")
+	}
+
+	var errs []error
+	for _, ist := range ists {
+		requiredImageStreamTag := ist
+		if err := quayiociimagesdistributor.EnsureISTinQCI(ctx, client, requiredImageStreamTag, quayIOImageHelper,
```
hongkailiu, line 666: Could we do the mirror right away with `quayIOImageHelper`?
I think we do not even need to check if it exists on QCI already or is out of date.
Moreover, the nil case of store wont work at the moment, because we never consume the task in the store.

pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
```diff
@@ -130,23 +130,31 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, log *logrus.Entry) error {
 	*log = *log.WithField("namespace", req.Namespace).WithField("name", req.Name)
 	log.Info("Starting reconciliation")
-	colonSplit := strings.Split(req.Name, ":")
+	if err := EnsureISTinQCI(ctx, r.client, req.NamespacedName, r.quayIOImageHelper, r.ocImageInfoOptions, r.mirrorStore, r.onlyValidManifestV2Images, log); err != nil {
```
hongkailiu, line 133: In my mind, we should not need to refactor any code from this file.

pkg/rehearse/rehearse.go
```diff
@@ -569,6 +576,15 @@ func setupDependencies(
 			if err := ensureImageStreamTags(ctx, client, requiredImageStreamTags, buildCluster, prowJobNamespace, prowJobClient, log); err != nil {
 				return fmt.Errorf("failed to ensure imagestreamtags in cluster %s: %w", buildCluster, err)
 			}
+			ocClientFactory := quayiociimagesdistributor.NewClientFactory()
+			quayIOImageHelper, err := ocClientFactory.NewClient()
```
hongkailiu, line 580: Could we put those two in main() and pass `quayIOImageHelper` into the function?
It should be reusable.

pkg/rehearse/rehearse.go
```diff
@@ -637,6 +653,56 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
 	return g.Wait()
 }
 
+func ensureISTSInQCI(ctx context.Context, client ctrlruntimeclient.Client,
+	ists apihelper.ImageStreamTagMap, registryConfig string, quayIOImageHelper quayiociimagesdistributor.OCClient, log *logrus.Entry) error {
+
+	mirrorOptions := quayiociimagesdistributor.OCImageMirrorOptions{
```
hongkailiu, line 659: This one can be reused too and replacing `registryConfig` to pass it along.

pkg/rehearse/rehearse.go
```diff
@@ -637,6 +653,56 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+		if colonSplit[0] != "sha256" {
+			//should never happen
+			return fmt.Errorf("image name has no prefix `sha256:`: %s", imageName)
+		}
+
+		sourceImage := fmt.Sprintf("%s/%s/%s@%s", api.DomainForService(api.ServiceRegistry), tagRef.Namespace, tagRef.Name, sourceImageStreamTag.Image.ObjectMeta.Name)
+		targetImageWithDateAndDigest := api.QuayImageFromDateAndDigest(time.Now().Format("20060102"), colonSplit[1])
+
+		pairs := []string{sourceImage + "=" + quayImage, sourceImage + "=" + targetImageWithDateAndDigest}
+		if err := quayIOImageHelper.ImageMirror(pairs, mirrorOptions); err != nil {
```
hongkailiu, line 699: Could we dry out a function to compute `pairs`?
I would like to see some unit tests on it.

pkg/rehearse/rehearse.go
```diff
@@ -637,6 +653,56 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+		if n := len(colonSplit); n != 2 {
+			//should never happen
+			return fmt.Errorf("splitting %s by `:` didn't yield two but %d results", imageName, n)
+		}
+		if colonSplit[0] != "sha256" {
+			//should never happen
+			return fmt.Errorf("image name has no prefix `sha256:`: %s", imageName)
+		}
+
+		sourceImage := fmt.Sprintf("%s/%s/%s@%s", api.DomainForService(api.ServiceRegistry), tagRef.Namespace, tagRef.Name, sourceImageStreamTag.Image.ObjectMeta.Name)
```
hongkailiu, line 695: We have a function for this.
https://github.com/openshift/ci-tools/blob/7ea7d8d43463de4600dfce42664cf7fc71763e55/pkg/api/promotion.go#L109-L111

pkg/rehearse/rehearse.go
```diff
@@ -638,6 +651,56 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
 	return g.Wait()
 }
 
+func ensureISTSInQCI(ctx context.Context, client ctrlruntimeclient.Client, ists apihelper.ImageStreamTagMap,
+	mirrorOptions quayiociimagesdistributor.OCImageMirrorOptions, quayIOImageHelper quayiociimagesdistributor.OCClient, log *logrus.Entry) error {
+
+	var errs []error
+	for _, ist := range ists {
+		pairs, err := createPairs(ctx, ist, client, time.Now(), log)
```
hongkailiu, line 659: Could we collect all the pairs and then do the mirroring on all pairs?

pkg/rehearse/rehearse.go
```diff
@@ -638,6 +651,56 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+		if err == nil {
```
hongkailiu, line 660: Why eat up the err here?

pkg/rehearse/rehearse_test.go
```diff
@@ -405,3 +405,91 @@ func TestFilterJobsByRequested(t *testing.T) {
...
+					},
+					}},
+			).Build(),
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			timeForTest := time.Date(2000, time.January, 1, 0, 0, 0, 0, time.UTC)
+			actual, err := createPairs(context.TODO(), tc.namespacedName, tc.fakeClient, timeForTest, logrus.NewEntry(logrus.StandardLogger()))
+			if err != nil && (err == nil || !tc.expectedErr) {
```
hongkailiu, line 487: The condition does not look correct to me. 
There is a util function from `testhelper.EquateErrorMessage` to compare the error msg.

pkg/rehearse/rehearse.go
```diff
@@ -638,6 +651,56 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+			if err := quayIOImageHelper.ImageMirror(pairs, mirrorOptions); err != nil {
```
hongkailiu, line 661: Wrap the mirror into a retry like [this](https://github.com/openshift/ci-tools/blob/6359329eb9e2662b759474ba91e16eb3cd003067/pkg/controller/quay_io_ci_images_distributor/mirror.go#L111)?


We could do this later too in another PR.

pkg/rehearse/rehearse.go
```diff
@@ -308,7 +312,9 @@ func labelSelectorForRehearsalJobs(org, repo string, prNumber int) ctrlruntimecl
 }
 
 // RehearseJobs returns true if the jobs were triggered and succeed
-func (r RehearsalConfig) RehearseJobs(candidate RehearsalCandidate, candidatePath string, prRefs *pjapi.Refs, imageStreamTags apihelper.ImageStreamTagMap, presubmitsToRehearse []*prowconfig.Presubmit, rehearsalTemplates, rehearsalClusterProfiles *ConfigMaps, logger *logrus.Entry) (bool, error) {
+func (r RehearsalConfig) RehearseJobs(candidate RehearsalCandidate, candidatePath string, prRefs *prowapi.Refs,
```
hongkailiu, line 315: NIT: If you want to break the line, please do it for each arg?

pkg/rehearse/rehearse.go
```diff
@@ -638,6 +660,66 @@ func ensureImageStreamTags(ctx context.Context, client ctrlruntimeclient.Client,
...
+	var pairs []string
+	for _, ist := range ists {
+		istPairs, err := createPairs(ctx, ist, client, time.Now(), log)
+		if err != nil {
+			errs = append(errs, err)
+			continue
+		}
+		pairs = append(pairs, istPairs...)
+	}
+	if len(pairs) == 0 {
```
hongkailiu, line 674:
```suggestion
	if len(errs) > 0 {
```

hongkailiu (approved):

PR #3819, opened 2023-11-30
Replace vsphere with vsphere02 completely.

jcpowermac:
This PR removes all references to `vsphere`
and replaces it with vsphere02. The
only cluster that should be used for vsphere
ci jobs now is `vsphere02`.


jcpowermac:
After this is merged I intend to regenerate jobs

droslean:
/hold

@jcpowermac  Please open a follow-up PR to the release repo to regenerate the jobs.

jcpowermac:
cc: @hongkailiu @bear-redhat

PR #3855, opened 2023-12-18
chore(RHIDP-885): add cluster profile for Red Hat Developer Hub performance testing on AWS

jhutar:


jhutar:
Hello @psalajova. Same here. Any idea when `ci/prow/security` is failing?

psalajova:
@jhutar Hi, no, but it's optional so it won't block the merge

PR #3856, opened 2023-12-18
Panic if we unable to determine Release from a job name

DennisPeriquet:
[TRT-1414](https://issues.redhat.com//browse/TRT-1414)

Because of this, we didn't get 4.16 Job data which prevented us from getting 4.16 disruption/alerting data (as the Release was set to 'unknown'.  This PR makes it so that we panic to avoid this happening in the future in case we forgot to update the code with the branch after a new 4.n release is created.

### DIFF
--- a/pkg/jobrunaggregator/jobtableprimer/job_typer.go
+++ b/pkg/jobrunaggregator/jobtableprimer/job_typer.go
@@ -1,6 +1,7 @@
 package jobtableprimer
 
 import (
+	"fmt"
 	"strings"
 
 	"github.com/openshift/ci-tools/pkg/jobrunaggregator/jobrunaggregatorapi"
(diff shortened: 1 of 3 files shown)


pkg/jobrunaggregator/jobtableprimer/job_typer.go
```diff
@@ -80,6 +81,10 @@ func newJob(name string) *jobRowBuilder {
 	if len(versions) >= 1 {
 		currRelease = versions[0]
 	}
```
neisw, line 83: could be an else too, if we don't detect a version then panic.

DennisPeriquet: pushed

DennisPeriquet:
gonna fixup that unit test.

DennisPeriquet:
converted to ensure it panics for unknown release, ..., all green when run locally.

PR #3886, opened 2024-01-11
[WIP] Add cluster profiles to config resolver

psalajova:
A first step of https://issues.redhat.com/browse/DPTP-3642 -- making config resolver able to serve cluster profiles config file which lives in the release repo.

/cc @droslean


pkg/load/load.go
```diff
@@ -38,14 +38,18 @@ const (
 	RegistryDocumentation
 )
 
+// ClusterProfilesConfigPath is the cluster profiles config path from release repository
+const ClusterProfilesConfigPath = "core-services/cluster-profiles/_config.yaml"
+
 // Registry takes the path to a registry config directory and returns the full set of references, chains,
 // and workflows that the registry's Resolver needs to resolve a user's MultiStageTestConfiguration
-func Registry(root string, flags RegistryFlag) (registry.ReferenceByName, registry.ChainByName, registry.WorkflowByName, map[string]string, api.RegistryMetadata, registry.ObserverByName, error) {
+func Registry(root string, flags RegistryFlag) (registry.ReferenceByName, registry.ChainByName, registry.WorkflowByName, api.ClusterProfilesList, map[string]string, api.RegistryMetadata, registry.ObserverByName, error) {
```
droslean, line 46: Since you are doing that, I think it would be better refactor this function to return a struct with all properties included, instead of adding an extra return value.

droslean (approved):

PR #3892, opened 2024-01-16
TRT-1440: Allow prcreator to have comma separated assignee list

DennisPeriquet:
We want certain automated PRs to get assigned to multiple people instead of just one.  This allows the assignee to be a comma separated list; one assignee should still work.

TODO:

- [ ] Do some due diligence on this to avoid breaking automated PRs.

### DIFF
--- a/cmd/prcreator/main.go
+++ b/cmd/prcreator/main.go
@@ -28,7 +28,7 @@ func gatherOptions() (*options, error) {
 	flag.StringVar(&opts.prTitle, "pr-title", "", "The title of the PR to create")
 	flag.StringVar(&opts.prMessage, "pr-message", "", "The message of the PR to create")
 	flag.StringVar(&opts.gitCommitMessage, "git-message", "", "The git commit message of the PR to create. If not set, then its value will be composed of other flags")
-	flag.StringVar(&opts.prAssignee, "pr-assignee", "", "The assignee of the PR to create")
+	flag.StringVar(&opts.prAssignee, "pr-assignee", "", "Comma separated list of assignees for the PR to create")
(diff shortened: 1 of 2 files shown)


cmd/prcreator/main.go
```diff
@@ -28,7 +28,7 @@ func gatherOptions() (*options, error) {
 	flag.StringVar(&opts.prTitle, "pr-title", "", "The title of the PR to create")
 	flag.StringVar(&opts.prMessage, "pr-message", "", "The message of the PR to create")
 	flag.StringVar(&opts.gitCommitMessage, "git-message", "", "The git commit message of the PR to create. If not set, then its value will be composed of other flags")
-	flag.StringVar(&opts.prAssignee, "pr-assignee", "", "The assignee of the PR to create")
+	flag.StringVar(&opts.prAssignees, "pr-assignees", "", "Comma separated list of assignees for the PR to create")
```
neisw, line 31: Changing the name here will break existing code before we  update the periodic and any other code that might be using it that we aren't aware of.  How about changing prAssignee to []string and the flag to `flag.StringArrayVar()` like https://github.com/openshift/ci-tools/blob/17e9c52f4342964afe8ff762c878435f0e3da992/pkg/jobrunaggregator/jobruntestcaseanalyzer/cmd.go#L134

then the code can pass in `-pr-assignee dgoodwin -pr-assignee  neisw  -pr-assignee DennisPeriquet` and the existing single case code should continue to work as well

neisw:
/lgtm

I believe you said you ran this locally, just want to confirm the single assignee and the csv both work as expected.

deepsm007 (approved):

PR #3901, opened 2024-01-18
`payload-testing-prow-plugin`: add the ability to support multiple PRs

smg247:
This adds a new command to the plugin to support triggering a payload job with additional PRs built into the payload: `payload-job-with-prs`. As an example, the command `/payload-job-with-prs some-job-name openshift/kubernetes#1234 openshift/installer#999` would include the PR that the command was issued on as well as the 2 mentioned PRs in the generated `PRPQR` CR. I will follow this up with an update to the documentation.

For: https://issues.redhat.com/browse/DPTP-3689

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -38,6 +38,7 @@ type githubClient interface {
 var (
 	ocpPayloadTestsPattern              = regexp.MustCompile(`(?mi)^/payload\s+(?P<ocp>4\.\d+)\s+(?P<release>\w+)\s+(?P<jobs>\w+)\s*$`)
 	ocpPayloadJobTestsPattern           = regexp.MustCompile(`(?mi)^/payload-job\s+((?:[-\w.]+\s*?)+)\s*$`)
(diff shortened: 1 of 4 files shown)


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -154,6 +161,19 @@ func jobsFromComment(comment string) []config.Job {
 		})
 
 	}
+	for _, match := range ocpPayloadJobTestsWithPRsPattern.FindAllStringSubmatch(comment, -1) {
```
danilo-gemoli, line 164: Should we set an upper bound on how many PRs can be included?

smg247: hmm, that is an interesting thought. There is a 6 hour timeout on the job itself, so if you include too many than it will likely end up failing. As far as what "too many" is, I am unsure. I tried it with 5 and it was fine. I don't think we need to impose an arbitrary limit here. Let's wait and see how it is used and if we notice it running into problems we can adjust it. I think the majority of the time this will be run with 2 - 3 prs.

PR #3906, opened 2024-01-22
Add Unit Tests for manageHome + manageCLI

Toma5OD:
#### Files Changed:
- `cmd/entrypoint-wrapper/main_test.go`

#### Changes:
1. **TestManageCLI**: Added unit tests to verify the behavior of the `manageCLI` function. Tests include:
   - Appending CLI directory to the `PATH` environment variable when CLI directory is set.
   - Leaving `PATH` unchanged if CLI directory is not set.

2. **TestManageHome**: Implemented unit tests for `manageHome` function. These tests cover:
   - Ensuring no change when the home environment variable is set.
   - Assigning a default home directory if the home environment variable is not set.
   - Handling scenarios where the home environment is set but not writable.

https://issues.redhat.com/browse/DPTP-3847

### DIFF
--- a/cmd/entrypoint-wrapper/main_test.go
+++ b/cmd/entrypoint-wrapper/main_test.go
@@ -1,0 +1,124 @@
+package main
+
+import (
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+func TestManageCLI(t *testing.T) {
+	testCases := []struct {
+		testName string
+		cliDir   string
+		expected string
+	}{
+		{
+			testName: "CLI Directory Set",
+			cliDir:   "/usr/local/bin/oc",
+			expected: os.Getenv("PATH") + ":/usr/local/bin/oc",
+		},
+		{
+			testName: "CLI Directory Not Set",
+			expected: os.Getenv("PATH"),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.testName, func(t *testing.T) {
+			originalPath := os.Getenv("PATH")
+
+			defer os.Setenv("PATH", originalPath)
+
+			cmd := exec.Command("echo")
+			cmd.Env = os.Environ()
+
+			if testCase.cliDir != "" {
+				os.Setenv(api.CliEnv, testCase.cliDir)
+				defer os.Unsetenv(api.CliEnv)
+			}
+
+			manageCLI(cmd)
+
+			actualPath := ""
+			for _, env := range cmd.Env {
+				if strings.HasPrefix(env, "PATH=") {
+					actualPath = strings.TrimPrefix(env, "PATH=")
+				}
+			}
+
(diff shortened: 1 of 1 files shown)


cmd/entrypoint-wrapper/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+			homeSet:  false,
+			wantEnv:  []string{"HOME=/alabama"},
+			wantHome: "/alabama",
+		},
+		{
+			testName: "home env set but not writable",
+			homeSet:  true,
+			wantEnv:  []string{"HOME=/alabama"},
+			wantHome: "/alabama",
+			fileMode: fileModePtr(0555),
```
Prucek, line 84: I would make the 0555 a constant

cmd/entrypoint-wrapper/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+func TestManageHome(t *testing.T) {
+	fileModePtr := func(fm os.FileMode) *os.FileMode { return &fm }
+
+	for _, testCase := range []struct {
+		testName string
+		homeSet  bool
+		wantEnv  []string
+		wantHome string
```
Prucek, line 66: expectedEnv or expectedHome
I would delete one of it, they are basically the same

cmd/entrypoint-wrapper/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+			cmd.Env = os.Environ()
+
+			if testCase.cliDir != "" {
+				os.Setenv(api.CliEnv, testCase.cliDir)
+				defer os.Unsetenv(api.CliEnv)
+			}
+
+			manageCLI(cmd)
+
+			gotPath := ""
```
Prucek, line 45: nit: actualPath

cmd/entrypoint-wrapper/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+func TestManageCLI(t *testing.T) {
+	for _, testCase := range []struct {
+		testName string
+		cliDir   string
+		expected string
+	}{
+		{
+			testName: "If CLI directory is set, append it to the PATH environment variable",
```
Prucek, line 21: I would make the test names shorter, something like: CLI directory set
The code is self-explanatory

cmd/entrypoint-wrapper/main_test.go
```diff
@@ -0,0 +1,119 @@
...
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+func TestManageCLI(t *testing.T) {
+	for _, testCase := range []struct {
+		testName string
+		cliDir   string
+		expected string
+	}{
```
Prucek, line 19: I would separate the test cases and the run of the test, like we do in other tests, take a look here for example: https://github.com/openshift/ci-tools/blob/master/cmd/github-ldap-user-group-creator/main_test.go#L33-L42

cmd/entrypoint-wrapper/main_test.go
Toma5OD: changes made

PR #3907, opened 2024-01-22
`payload-testing-prow-plugin`: comment made on each additional PR included in payload test

smg247:
Each PR that the user requested to be included in the payload will be commented on with a link to the `PRPQR` run, and a reference to the originating PR. This will allow users to get a better understanding of all tests run for a PR.

For: https://issues.redhat.com/browse/DPTP-3689

### DIFF
--- a/cmd/payload-testing-prow-plugin/server.go
+++ b/cmd/payload-testing-prow-plugin/server.go
@@ -193,6 +193,7 @@ func (s *server) handleIssueComment(l *logrus.Entry, ic github.IssueCommentEvent
 			o, r, n, err := pr.GetOrgRepoAndNumber()
 			if err != nil {
 				l.WithError(err).Errorf("unable to determine PR from string: %s", pr)
+				continue
 			}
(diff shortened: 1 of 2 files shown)


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -182,12 +182,27 @@ const (
...
+	if comment, additionalPRs := s.handle(l, ic); comment != "" {
+		org := ic.Repo.Owner.Login
+		repo := ic.Repo.Name
+		number := ic.Issue.Number
+		user := ic.Comment.User.Name
+		s.createComment(org, repo, number, comment, user, l)
+		originalPRRef := fmt.Sprintf("%s/%s#%d", org, repo, number)
+		for _, pr := range additionalPRs {
+			o, r, n, err := pr.GetOrgRepoAndNumber()
+			if err != nil {
```
danilo-gemoli, line 194: When `err != nil` then `o, r, n` are usable. I would also `continue` to the next iteration other than logging

PR #3910, opened 2024-01-23
`multi-arch-builder-controller`: Better logging and tests

danilo-gemoli:
I had hard time trying to understand what was going wrong with a failing build and I have realized we could have done much better with logs.

/cc @droslean

### DIFF
--- a/pkg/api/multiarchbuildconfig/v1/types.go
+++ b/pkg/api/multiarchbuildconfig/v1/types.go
@@ -66,13 +66,13 @@ func UpdateMultiArchBuildConfig(ctx context.Context, logger *logrus.Entry, clien
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		mabc := &MultiArchBuildConfig{}
 		if err := client.Get(ctx, namespacedName, mabc); err != nil {
-			return fmt.Errorf("failed to get the MultiArchBuildConfig: %w", err)
+			return fmt.Errorf("failed to get the MultiArchBuildConfig during update: %w", err)
 		}
 
 		mabc = mabc.DeepCopy()
 		mutateFn(mabc)
 
-		logger.WithField("namespace", namespacedName.Namespace).WithField("name", namespacedName.Name).Info("Updating MultiArchBuildConfig...")
+		logger.Info("Updating MultiArchBuildConfig...")
 		if err := client.Update(ctx, mabc); err != nil {
 			return fmt.Errorf("failed to update MultiArchBuildConfig %s: %w", mabc.Name, err)
 		}
--- a/pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
+++ b/pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
@@ -43,6 +43,13 @@ const (
 	MirrorImageManifestDone  = "ImageMirrorDone"
 	ImageMirrorSuccessReason = "ImageMirrorSuccess"
 	ImageMirrorErrorReason   = "ImageMirrorError"
+
+	MABCNameLogField          = "multiarchbuildconfig_name"
+	PushTargetImageLogField   = "target_image"
+	MirrorTargetImageLogField = "target_image"
+	MirrorRegistriesLogField  = "registries"
+	BuildNameLogField         = "build_name"
+	BuildNamespaceLogField    = "build_namespace"
 )
 
 func AddToManager(mgr manager.Manager, architectures []string, dockerCfgPath string) error {
@@ -91,7 +98,7 @@ type reconciler struct {
 
 func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reconcile.Result, error) {
 	logger := r.logger.WithField("request", req.String())
-	err := r.reconcile(ctx, req, logger)
+	err := r.reconcile(ctx, logger, req)
 	if err != nil {
 		logger.WithError(err).Error("Reconciliation failed")
 	} else {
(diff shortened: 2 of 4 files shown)


pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
```diff
@@ -394,6 +430,13 @@ func TestReconcile(t *testing.T) {
 					},
 				},
 			},
+			wantLog: logOutput(func(e *logrus.Entry) {
```
droslean, line 433: You can just check the log output string, but I don't think that it makes any value in the tests.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
```diff
@@ -41,6 +42,29 @@ func init() {
 	}
 }
 
+type testLogger struct {
```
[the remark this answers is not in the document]
danilo-gemoli: Because of this solely
```golang
func (log *testLogger) Output() string {
	return log.Logger.Out.(*strings.Builder).String()
}
```

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
danilo-gemoli: I want to know how and what the reconciler is going to log because what it has been logging so far wasn't good.
`wantLog` is optional, if you're going to add a new test case and don't want to check any log, just leave it `nil`.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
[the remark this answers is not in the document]
droslean: Why do you need to check the log output on tests? I don't see the point. Also, the whole approach is wrong since you can use the logrus hooks to test this out.

pkg/api/multiarchbuildconfig/v1/types.go
```diff
@@ -66,14 +66,16 @@ func UpdateMultiArchBuildConfig(ctx context.Context, logger *logrus.Entry, clien
 	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		mabc := &MultiArchBuildConfig{}
 		if err := client.Get(ctx, namespacedName, mabc); err != nil {
+			logger.Error("Failed to get the MultiArchBuildConfig while updating")
```
droslean, line 69: But we are already returning the error. You need to log it on that layer.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig.go
```diff
@@ -235,9 +261,11 @@ func (r *reconciler) handlePushImageWithManifest(ctx context.Context, mabc *v1.M
 			})
 			mabcToMutate.Status.State = v1.FailureState
 		}
+	} else {
```
droslean, line 264: No need for the else statement here.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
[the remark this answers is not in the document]
danilo-gemoli: I'll assess the hooks and see how the code will look like when using them. To answer your question: https://github.com/openshift/ci-tools/pull/3910#discussion_r1463297795

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
droslean: I don't see any point. It would be better to check the returned error if a test case fail, instead of testing the actual logging.

pkg/api/multiarchbuildconfig/v1/types.go
danilo-gemoli: You mean outside the `RetryOnConflict` every time `UpdateMultiArchBuildConfig` is invoked?

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
danilo-gemoli: Well I might revert the tests but we need to improve logs anyway

pkg/api/multiarchbuildconfig/v1/types.go
droslean: This will retry only on conflicts. Technically we already return the error. You can add the log there.

pkg/controller/multiarchbuildconfig/multiarchbuildconfig_test.go
droslean: Yea, having better logging is always good, but there is no point in including the logs in the tests.

pkg/api/multiarchbuildconfig/v1/types.go
danilo-gemoli: Something like this?
```diff
if err := v1.UpdateMultiArchBuildConfig(ctx, logger, r.client, ctrlruntimeclient.ObjectKey{Namespace: mabc.Namespace, Name: mabc.Name}, mutateFn); err != nil {
+   logger.Errorf("Failed to update the mabc: %s", err)
    return fmt.Errorf("failed to update the MultiArchBuildConfig %s/%s: %w", mabc.Namespace, mabc.Name, err)
}
```

droslean: We are returning that error too.

danilo-gemoli:
It seems we have a flaky e2e, @droslean if you're happy with this PR I'm going to retest it again

droslean (approved):

PR #3921, opened 2024-01-30
TRT-1425: Automatically sync jobs with release table when priming job tables

xueqzhan:


pkg/jobrunaggregator/jobtableprimer/job_name_generator.go
```diff
@@ -0,0 +1,248 @@
...
+				!strings.Contains(curr.Name, "-4.13") &&
+				!strings.Contains(curr.Name, "-4.14") &&
+				!strings.Contains(curr.Name, "-4.15") &&
+				!strings.Contains(curr.Name, "-4.16") &&
+				!strings.Contains(curr.Name, "-4.17") &&
+				!strings.Contains(curr.Name, "-4.18") &&
+				!strings.Contains(curr.Name, "-4.19") &&
+				!strings.Contains(curr.Name, "-4.20") {
+				continue
+			}
```
neisw, line 221: Does this get changed so we don't have to update anymore?  Can we check for all releases > 4.10 and check contains against ReleaseRow.Release?

xueqzhan: Good catch! I guess I just copied the code without seeing the release string dealing there. :)

pkg/jobrunaggregator/jobtableprimer/job_name_generator.go
```diff
@@ -0,0 +1,263 @@
...
+	}
+)
+
+func newJobNameGenerator() *jobNameGenerator {
+	generator := &jobNameGenerator{
+		periodicURLs: []string{
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/release/openshift-release-master-periodics.yaml",
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/multiarch/openshift-multiarch-master-periodics.yaml",
+
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/release/openshift-release-release-4.10-periodics.yaml",
```
neisw, line 43: nit - do we even need this any longer or can we just use the templates for all the defined releases?

pkg/jobrunaggregator/jobtableprimer/job_name_generator.go
```diff
@@ -0,0 +1,263 @@
...
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/release/openshift-release-release-4.15-periodics.yaml",
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/release/openshift-release-release-4.16-periodics.yaml",
+
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/hypershift/openshift-hypershift-release-4.13-periodics.yaml",
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/hypershift/openshift-hypershift-release-4.14-periodics.yaml",
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/hypershift/openshift-hypershift-release-4.15-periodics.yaml",
+			"https://raw.githubusercontent.com/openshift/release/master/ci-operator/jobs/openshift/hypershift/openshift-hypershift-release-4.16-periodics.yaml",
+		},
+		releaseConfigURLs: []string{
+			"https://raw.githubusercontent.com/openshift/release/master/core-services/release-controller/_releases/release-ocp-4.10-arm64.json",
```
neisw, line 57: same nit - do we even need this any longer or can we just use the templates for all the defined releases?

pkg/jobrunaggregator/jobtableprimer/job_name_generator.go
```diff
@@ -0,0 +1,263 @@
...
+						if strings.Contains(curr.Name, "-"+release.Release) {
+							foundRelease = true
+							break
+						}
+					}
+				}
+				if !foundRelease {
+					continue
+				}
+			} else if !strings.Contains(curr.Name, "-4.10") &&
```
neisw, line 224: Is this an expected case where we wouldn't have any releases at this point?  Just wondering if it is in effect dead code that we don't expect to run and may as well clean up?

xueqzhan: These are kept here for ./job-run-aggregator generate-job-names command to continue working. Most likely there is no other users other than us. But I am just trying to be cautious. When we decide to remove the command, we will remove the job in the continuous-release-jobs repo and these code sections from ci-tools as well.  This applies to the two comments above as well.

neisw: Oh, interesting so you only call

jobNameGenerator.UpdateURLsForNewReleases(releases) for [createJobRowsFromReleases](https://github.com/xueqzhan/ci-tools/blob/2ccf98edeb22509ee82f535d6c70861f3b18a528/pkg/jobrunaggregator/jobtableprimer/create_jobs.go#L18) but not for [NewGenerateJobNamesCommand](https://github.com/xueqzhan/ci-tools/blob/2ccf98edeb22509ee82f535d6c70861f3b18a528/pkg/jobrunaggregator/jobtableprimer/generate_jobnames.go#L28).  Don't we need it there too?

neisw: Or is this a two step process and the expectation is that  generate-job-names  is obsoleted and removed later?

xueqzhan: generate-job-names will be removed, hopefully soon. It uses the default config as it is. So it is not calling jobNameGenerator.UpdateURLsForNewReleases(releases). There is no dependency on bigquery for that command to work. It is better to keep it that way. It will be removed later anyhow.

pkg/jobrunaggregator/jobtableprimer/jobs.go
```diff
@@ -47,36 +49,91 @@ const (
...
+			return false
+		}
+
+		return iMajor > jMajor || iMinor > jMinor
+	})
+	return reverseOrderedVersions
+}
 
-//go:embed generated_job_names.txt
-var jobNames string
```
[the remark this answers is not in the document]
xueqzhan: You are right. jobNames is used to prime the table. That is obsolete now with this PR. But generate-job-names cmd can still run. It just will have no effect on priming the table.

PR #3927, opened 2024-02-02
`gpu-scheduling-webhook`: First implementation

danilo-gemoli:
Add a mutating admission controller that adds a toleration on pods that request an Nvidia GPU.

Our clusters host some nodes that feature an Nvidia GPU. They are expensive to run workload on so
by using a mutating webhook we ensure that only the pods requesting a GPU actually run on those
nodes, leaving out everything else.

Check out `README.md` for a more detailed description.

/cc @bear-redhat @openshift/test-platform

### DIFF
--- a/cmd/gpu-scheduling-webhook/main.go
+++ b/cmd/gpu-scheduling-webhook/main.go
@@ -1,0 +1,169 @@
+package main
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/bombsimon/logrusr/v3"
+	"github.com/go-logr/logr"
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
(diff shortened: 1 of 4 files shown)


cmd/gpu-scheduling-webhook/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// Check if a GPU required on resources from containers and init containers.
+	var gpuNeeded bool
+	gpuNeeded = hasNvidaGPURequest(logger, pod.Spec.InitContainers)
+
+	if !gpuNeeded {
+		gpuNeeded = hasNvidaGPURequest(logger, pod.Spec.Containers)
+	}
+
+	if gpuNeeded {
```
bear-redhat, line 83: maybe something like
```
if hasNvidaGPURequest(logger, pod.Spec.InitContainers) || hasNvidaGPURequest(logger, pod.Spec.Containers)
```
?

bear-redhat (approved):

cmd/gpu-scheduling-webhook/main.go
danilo-gemoli: I'd like to save some iterations and skip the second loop if the first one has already found the request.

PR #3933, opened 2024-02-06
Add cluster-profiles-config arg to README

psalajova:
Since implementing the private cluster profile feature, the script needs a new arg to be run, as now it is also checking the cluster profiles config file to validate the configs in the release repository.


psalajova:
@deepsm007 could you lgtm? This PR is just updating the readme

PR #3937, opened 2024-02-07
DPTP-3826: Adding ODF cluster profile

Nikhil-Ladha:
Added a new cluster profile to be used for ODF, i.e, projects under `red-hat-storage` org.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1372,6 +1372,7 @@ const (
 	ClusterProfileQuayAWS               ClusterProfile = "quay-aws"
 	ClusterProfileAWSEdgeInfra          ClusterProfile = "aws-edge-infra"
 	ClusterProfileRosaAws               ClusterProfile = "rosa-aws"
+	ClusterProfileODFAWS                ClusterProfile = "odf-aws"
 )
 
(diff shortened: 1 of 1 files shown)


Nikhil-Ladha:
cc @malayparida2000 @iamniting

malayparida2000:
Looking at https://github.com/openshift/ci-tools/pull/1548/files & https://github.com/openshift/ci-tools/pull/1818/files, don't we need to make changes at other places also as made in these 2 similar PRs?

Nikhil-Ladha:
> Looking at #1548 & #1818, don't we need to make changes at other places like inside the prowgen folder also as made in these 2 similar PRs?

Those are very old PRs, I think the latest way to add profiles have been updated a bit and even the [example](https://github.com/openshift/ci-tools/pull/2808) PR given in the docs just mentions updating the `types.go` file.

malayparida2000:
Hm, I see they have moved around those functions so those changes might not be necessary anymore.

Nikhil-Ladha:
@jmguzik @droslean @psalajova can you please review the PR and approve if all looks good? So, that we can move ahead with making other required changes?

PR #3938, opened 2024-02-07
`ci-operator`: Add env var `OVERRIDE_IMAGE_*` to override component images with provided name

smg247:
This adds a new pattern for env vars to `ci-operator`: `OVERRIDE_IMAGE_*`. The remaining part of the var name relates to the image name, and the value is the tag to override it with in the `ocp` namespace. This will be used for payload-testing in a follow up PR.

For: https://issues.redhat.com/browse/DPTP-3688

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -164,7 +164,7 @@ func fromConfig(
 		return nil, nil, fmt.Errorf("failed to get steps from configuration: %w", err)
 	}
 	rawSteps = append(graphConf.Steps, rawSteps...)
-	rawSteps = append(rawSteps, stepsForImageOverrides()...)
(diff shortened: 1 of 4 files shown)


pkg/defaults/defaults.go
```diff
@@ -371,6 +372,32 @@ func fromConfig(
 	return append(overridableSteps, buildSteps...), postSteps, nil
 }
 
+func stepsForImageOverrides() []api.StepConfiguration {
```
danilo-gemoli, line 375:
```diff
- func stepsForImageOverrides() []api.StepConfiguration
+ func stepsForImageOverrides(overriddenImages map[string]string) []api.StepConfiguration
```
and then
```diff
- rawSteps = append(rawSteps, stepsForImageOverrides()...)
+ rawSteps = append(rawSteps, stepsForImageOverrides(GetOverriddenImages)...)
```

PR #3939, opened 2024-02-07
`payload-testing`: Add overriding image tag functionality to PRPQR

smg247:
Uses the new env var from https://github.com/openshift/ci-tools/pull/3938 to allow specific image tags to be overridden during payload testing.

For: https://issues.redhat.com/browse/DPTP-3688


pkg/prowgen/podspec.go
```diff
@@ -269,6 +270,21 @@ func ReleaseInitial(pullspec string) PodSpecMutator {
 	}
 }
 
+// OverrideImage sets the "OVERRIDE_IMAGE_*" env var in order to override the respective image with the provided tag
+func OverrideImage(name, tag string) PodSpecMutator {
+	return func(spec *corev1.PodSpec) error {
+		if name == "" || tag == "" {
+			return nil
```
danilo-gemoli, line 277: As this is an edge case, should we log before `return nil`?

smg247: The logging wouldn't do us much good as it would just end up in the `job-trigger-controller-manager` logs, but maybe returning an error would be useful. realistically we should never reach this code anyways.

PR #3943, opened 2024-02-08
Revert "added cluster-profile rosa-aws"

ritmun:
revert merge 

implementing using [sdcicd-1224](https://issues.redhat.com//browse/sdcicd-1224) instead


psalajova:
> > /override ci/prow/images
> 
> Doesn't look like this took effect

I think it did but this PR is really unlucky because then our bot retested everything again, which erased the override. After you solve the merge conficts just ping me or @smg247 and we'll override it again.

PR #3946, opened 2024-02-09
Handle `ocp-build-data` stream aliases

smg247:
Tools (like `registry-replacer`) need to be aware of the `aliases` in the streams in order to work properly now that the `ocp-build-data` contains these.

For: https://issues.redhat.com/browse/DPTP-3871

### DIFF
--- a/pkg/api/ocpbuilddata/types.go
+++ b/pkg/api/ocpbuilddata/types.go
@@ -143,9 +143,10 @@ func (oic *OCPImageConfig) setPublicOrgRepo(mappings []PublicPrivateMapping) {
 type StreamMap map[string]StreamElement
 
 type StreamElement struct {
-	Image         string `json:"image"`
-	UpstreamImage string `json:"upstream_image"`
-	Mirror        *bool  `json:"mirror"`
+	Image         string   `json:"image"`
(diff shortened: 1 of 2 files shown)


jupierce:
Technically `{VAR}` are variables that need to be resolved in order to ensure a match. For example `stream-{THIS}` and `stream-{THAT}` could match if [group.yaml](https://github.com/openshift-eng/ocp-build-data/blob/9577ab4a36ec9ec0a83db26fc7762983e2abf51d/group.yml#L3) `vars` had the same value for THIS and THAT. But this should fix us for now.

smg247:
> Technically `{VAR}` are variables that need to be resolved in order to ensure a match. For example `stream-{THIS}` and `stream-{THAT}` could match if [group.yaml](https://github.com/openshift-eng/ocp-build-data/blob/9577ab4a36ec9ec0a83db26fc7762983e2abf51d/group.yml#L3) `vars` had the same value for THIS and THAT. But this should fix us for now.

Does that just mean that the following logic needs to be updated to replace more (dynamic?) variables?
https://github.com/openshift/ci-tools/blob/10de6aeb4992eba7d67164de4bc4f95e5488acd3/pkg/api/ocpbuilddata/types.go#L388-L399

PR #3947, opened 2024-02-09
OpenStack: Add openstack-nerc-dev cluster-profile

MaysaMacedo:


MaysaMacedo:
lease was already added here https://github.com/openshift/release/pull/48606

EmilienM:
Don't we have a cluster profile for NERC already? Can you please explain the difference with what we have already and here?
Thanks

pierreprinetti:
/lgtm
@EmilienM I will let you remove the hold when/if you’re OK with this. Have you spotted a nerc configuration here already?

EmilienM:
> /lgtm @EmilienM I will let you remove the hold when/if you’re OK with this. Have you spotted a nerc configuration here already?

we have `ClusterProfileOpenStack` which I think is NERC?

pkg/api/types.go
```diff
@@ -1296,6 +1296,7 @@ const (
 	ClusterProfileOpenStackVexxhost     ClusterProfile = "openstack-vexxhost"
 	ClusterProfileOpenStackPpc64le      ClusterProfile = "openstack-ppc64le"
 	ClusterProfileOpenStackOpVexxhost   ClusterProfile = "openstack-operators-vexxhost"
+	ClusterProfileOpenStackNercDev      ClusterProfile = "openstack-nerc-dev"
```
EmilienM, line 1299: why do you call it `openstack-nerc-dev` btw, since it's for CI? Could we just name is `ClusterProfileOpenStackNerc` maybe? to avoid confusion and complexity.

pierreprinetti: We have two tenants on NERC and the one called CI… is the other one 😬 on which we don’t remotely have quota. We are setting up jobs on this tenant because @MaysaMacedo identified it as the one tenant where we happen to be able to run tests for  couple iterations of CAPI-installer (that is: only one external network).

I understand that the naming is suboptimal but at this stage I think it’s the one more likely to preserve our sanity

EmilienM: meh

pkg/api/types.go
```diff
@@ -1829,6 +1833,8 @@ func (p ClusterProfile) LeaseType() string {
 		return "openstack-vh-mecha-central-quota-slice"
 	case ClusterProfileOpenStackMechaAz0:
 		return "openstack-vh-mecha-az0-quota-slice"
+	case ClusterProfileOpenStackNercDev:
+		return "openstack-nerc-dev"
```
mandre, line 1837: Should have been
```suggestion
		return "openstack-nerc-dev-quota-slice"
```

mandre: Fixing in https://github.com/openshift/ci-tools/pull/3967/

PR #3948, opened 2024-02-09
remove ClusterProfileOpenStack (not used)

EmilienM:
We don't use that profile anymore, and want to get rid of it.


EmilienM:
can someone approve it please?

PR #3949, opened 2024-02-10
include the multi_arch field on images for the graph generation

droslean:
Adding the multiArch field to the graph scheme. Populate the field based on the prowgen's configuration

Follow-ups: I will update the scheme after this merges.

/cc @openshift/test-platform @danilo-gemoli 

### DIFF
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -75,6 +75,26 @@ func (p *Prowgen) MergeDefaults(defaults *Prowgen) {
 	p.Rehearsals.DisabledRehearsals = append(p.Rehearsals.DisabledRehearsals, defaults.Rehearsals.DisabledRehearsals...)
 }
 
+func LoadProwgenConfig(folder string) (*Prowgen, error) {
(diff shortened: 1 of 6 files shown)


pkg/config/load.go
```diff
@@ -75,6 +75,26 @@ func (p *Prowgen) MergeDefaults(defaults *Prowgen) {
 	p.Rehearsals.DisabledRehearsals = append(p.Rehearsals.DisabledRehearsals, defaults.Rehearsals.DisabledRehearsals...)
 }
 
+func LoadProwgenConfig(folder string) (*Prowgen, error) {
```
danilo-gemoli, line 78: Nice. As it's the same as [this one](https://github.com/openshift/ci-tools/blob/a3fd3261c807e0da2372cdd6a30f6e0087c578f5/cmd/ci-operator-prowgen/main.go#L104-L121), can you refactor them out?

droslean: Ah, yes. Nice catch.

PR #3954, opened 2024-02-12
`payload-testing`: allow multiple PRs from a single repo to be provided

smg247:
This is actually much simpler than I was anticipating to achieve. The `refs` allow multiple `Pulls` to be provided for the same base. When this happens, `clonerefs` already knows how to handle it and merges them all on top of the base. I confirmed this by looking through the logs when supplying multiple `pulls` in a `ref` for a prowjob. All we need to do to enable this functionality is to group the `pulls` by org and repo.

For: https://issues.redhat.com/browse/DPTP-3691


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -550,28 +540,50 @@ func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration,
...
 	}
 
-	var refs []prowv1.Refs
+	prsByRepo := make(map[string][]v1.PullRequestUnderTest)
 	for _, pr := range prs {
-		refs = append(refs, prowv1.Refs{
-			Org:  pr.Org,
-			Repo: pr.Repo,
+		orgRepo := fmt.Sprintf("%s/%s", pr.Org, pr.Repo)
+		prsByRepo[orgRepo] = append(prsByRepo[orgRepo], pr)
```
danilo-gemoli, line 546: Do we assume that the user has been careful enough not to specify the same PR more than once, right?

smg247: `clonerefs` is actually okay with that. It does do a second merge of the changes, but it ends up resulting in the same source as if it were only merged once. In the end, it just wastes a couple of seconds.

PR #3959, opened 2024-02-13
`ci-operator-configresolver`: guarantee unique metadata when merging configs from multiple sources

smg247:
Now that we allow PRs from the same repo to be tested together, we have situations where we are merging multiple of the same configs together. In practice, this appears to be okay with `ci-operator`, but it is a waste of time and resources. Instead, we should not merge duplicated metadata.

For: https://issues.redhat.com/browse/DPTP-3691

### DIFF
--- a/pkg/registry/server/server.go
+++ b/pkg/registry/server/server.go
@@ -10,6 +10,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	"k8s.io/apimachinery/pkg/api/resource"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/test-infra/prow/metrics"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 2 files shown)


pkg/registry/server/server.go
```diff
@@ -461,8 +462,8 @@ func MetadataEntriesFromQuery(w http.ResponseWriter, r *http.Request) ([]api.Met
 		if variantsExist {
 			element.Variant = variants[i]
 		}
-		metadata = append(metadata, element)
+		metadata.Insert(element)
 	}
 
-	return metadata, nil
+	return metadata.UnsortedList(), nil
```
danilo-gemoli, line 468: Ok, it's not sorted anymore but it doesn't matter.

smg247: yes, the order doesn't matter. We don't have any unit tests here anyways. If we did, we would probably need to guarantee the order.

smg247:
@danilo-gemoli can you retag this?

PR #3961, opened 2024-02-13
Pipeline controller: add info about repo being managed by pipeline ctrl

jmguzik:
I was thinking about creating some unit tests for that, but in the end this is one API call which is pretty straightforward I hope.


jmguzik:
?
/approve

PR #3966, opened 2024-02-14
pj-rehearse: tell the user when you're ignoring them

stevekuznetsov:
When tools don't do anything and don't tell the user why they explicitly choose not to do something, the UX is frustrating.

### DIFF
--- a/cmd/pj-rehearse/server.go
+++ b/cmd/pj-rehearse/server.go
@@ -250,7 +250,7 @@ func (s *server) handleNewPush(l *logrus.Entry, event github.PullRequestEvent) {
 			jobTableLines = append(jobTableLines, fmt.Sprintf("A full list of affected jobs can be found [here](%s%s)", s.rehearsalConfig.GCSBrowserPrefix, fileLocation))
 		}
 		jobTableLines = append(jobTableLines, s.getUsageDetailsLines()...)
-		if err = s.ghc.CreateComment(org, repo, number, strings.Join(jobTableLines, "\n")); err != nil {
+		if err := s.ghc.CreateComment(org, repo, number, strings.Join(jobTableLines, "\n")); err != nil {
 			logger.WithError(err).Error("failed to create comment")
 		}
 	}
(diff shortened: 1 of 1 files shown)


cmd/pj-rehearse/server.go
```diff
@@ -250,7 +250,7 @@ func (s *server) handleNewPush(l *logrus.Entry, event github.PullRequestEvent) {
 			jobTableLines = append(jobTableLines, fmt.Sprintf("A full list of affected jobs can be found [here](%s%s)", s.rehearsalConfig.GCSBrowserPrefix, fileLocation))
 		}
 		jobTableLines = append(jobTableLines, s.getUsageDetailsLines()...)
-		if err = s.ghc.CreateComment(org, repo, number, strings.Join(jobTableLines, "\n")); err != nil {
+		if err := s.ghc.CreateComment(org, repo, number, strings.Join(jobTableLines, "\n")); err != nil {
```
stevekuznetsov, line 253: Making it look like you're handling the error produced by CreateComment but implicitly shadowing the error from the larger scope is super pernicious

cmd/pj-rehearse/server.go
```diff
@@ -309,7 +309,10 @@ func (s *server) handlePotentialCommands(pullRequest *github.PullRequest, commen
 				s.rehearsalConfig.AbortAllRehearsalJobs(org, repo, number, logger)
 			default:
 				if rehearsalsTriggered {
-					// We don't want to trigger rehearsals more than once per comment
+					message := fmt.Sprintf("@%s: requesting more than one rehearsal in one comment is not supported", user)
```
smg247, line 312: I don't know that this is a necessary thing to surface to the user. We are still triggering their rehearsals here, we are just not going to trigger them multiple times redundantly. Did you notice some case where this resulted in a completely ignored command?

stevekuznetsov: Yes. I asked for N rehearsals and I did not get them. I had no idea why and started wondering if my configuration in the PR was broken. `pj-rehearse` is commonly not telling the user what's going on, which makes people think it's broken or has flakes - anecdotally I have met many people who just try something many times over because it's so opaque.

smg247: Maybe if you tried to rehearse single job in multiple different lines it would ignore all, but the first one?

smg247: Can you link to the specific comment? I would like to see if there is a gap that we missed here.

Prow plugins are complicated for sure, so this is a hard thing to do perfectly, but I would like to improve the communication if possible.

stevekuznetsov: There were likely other issues involved, since there were no jobs triggered at all for a while, with no comment as to why - perhaps error cases in handling the comment that don't lead to comment responses. Then, we saw the case where only the first got triggered. 

https://github.com/openshift/release/pull/48608#issuecomment-1942036110

PR #3967, opened 2024-02-15
Fix lease type for openstack-nerc-dev

mandre:
The lease type was wrong and jobs would fail with:

    failed to acquire lease for "openstack-nerc-dev": resources not found


pierreprinetti:
/lgtm
for the OpenShift-on-OpenStack team

mandre:
Looks like the security job is broken: https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-ci-tools-master-security

smg247:
> Looks like the security job is broken: https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-ci-tools-master-security

It is, and its optional so it won't need to pass.

PR #3975, opened 2024-02-16
`prpqr-controller`: Handle deletion gracefully

danilo-gemoli:
This relates to [DPTP-3819](https://issues.redhat.com/browse/DPTP-3819).

As of today the `prpqr-reconciler` isn't able to deal with the deletion of a `prpqr` resource, this PR is an attempt to fill the gap.

The main idea behind these changes is about putting a finalizer on a `prpqr` as long as at least one not completed ProwJob exists. In this way the reconciler has the time to intercept a deletion request, change the state of each ProwJob to `aborted` and then wait for `prpqr` status to reconcile.
Once the status gets reconciled, the controller removes the finalizer and then kubernetes deletes the resource.

In more detail, consider the following example:
1. a `prpqr-test` resource exists
1. `prpqr-reconciler` creates the ProwJobs `pj1` and `pj2`, each one bound to `prpqr-test` and puts a finalizer on it
1. an operator executes `oc delete prpqr/prpqr-test`
1. `prpqr-reconciler` receives a deletion event and aborts `pj1` and `pj2`
1. `pjstatussyncer-reconciler` reconciles the status of `prpqr-test`: each job has been properly aborted
1. `prpqr-reconciler` removes the finalizer
1. kubernetes recollects the resource

/cc @DennisPeriquet @droslean @smg247 @openshift/test-platform


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -732,3 +827,29 @@ func generateJobNameToSubmit(inject *api.MetadataWithTest, prs []v1.PullRequestU
 
 	return fmt.Sprintf("%s%s-%s", refs, variant, inject.Test)
 }
+
+// manageDependantProwJobsFinalizer adds a finalizer if the prpqr has at least one running job,
+// remove otherwise.
+func manageDependantProwJobsFinalizer(atLeastOneJobRunning bool, objMeta *metav1.ObjectMeta) {
```
smg247, line 833:
```suggestion
func manageDependentProwJobsFinalizer(atLeastOneJobRunning bool, objMeta *metav1.ObjectMeta) {
```

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -96,8 +98,10 @@ func AddToManager(mgr manager.Manager, ns string, rc injectingResolverClient, pr
 		CreateFunc: func(e event.CreateEvent) bool {
 			return e.Object.GetNamespace() == ns
 		},
-		DeleteFunc:  func(event.DeleteEvent) bool { return false },
-		UpdateFunc:  func(event.UpdateEvent) bool { return false },
+		DeleteFunc: func(event.DeleteEvent) bool { return false },
```
droslean, line 101: Who will add the deletationTime stamp? I guess you need to include those changes in this PR as well, right?

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -315,31 +359,77 @@ func (r *reconciler) reconcile(ctx context.Context, req reconcile.Request, logge
...
+		return label, exists
+	}
 
-		oldStatus := prpqr.Status.DeepCopy()
-		reconcileStatus(prpqr, statuses, allJobsTriggeredCondition)
-		if reflect.DeepEqual(*oldStatus, prpqr.Status) {
-			logger.Info("PullRequestPayloadQualificationRun status is up to date, no updates necessary")
-			return nil
+	abort := func(ctx context.Context, logger *logrus.Entry, job *prowv1.ProwJob) {
+		if !job.Complete() || (job.Status.State != prowv1.TriggeredState && job.Status.State != prowv1.PendingState) {
```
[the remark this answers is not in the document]
danilo-gemoli: No you're right, my bad.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
danilo-gemoli: Good point. It is actually set by kubernetes when one requests a deletion, as an example: `kubectl delete pj/foobar`.
See [here](https://kubernetes.io/docs/reference/kubernetes-api/common-definitions/object-meta/).

danilo-gemoli: Aborting feature is [there](https://github.com/openshift/ci-tools/blob/9f82e2ac5526d7901623b6167fa97991f4b7b4eb/cmd/payload-testing-prow-plugin/server.go#L91-L96), a user can already abort the jobs. I think that @DennisPeriquet was also concerned about `prpqr` cleanup in his comment on https://issues.redhat.com/browse/DPTP-3819. That's the reason of this PR.

PR #3980, opened 2024-02-20
Private cluster pools

jmguzik:


psalajova (commented):
- the new `cluster-claim-owners-config` arg also needs to be added here https://github.com/openshift/ci-tools/blob/master/test/validate-checkconfig.sh#L31 
- the codegen test complains that you need to run `make generate`

jmguzik:
@psalajova could you re-add lgtm?

PR #3991, opened 2024-02-26
Private cluster pools: fix config in integration tests

psalajova:
This was missed in https://github.com/openshift/ci-tools/pull/3980


test/integration/pj-rehearse/master/core-services/cluster-pools/_config.yaml
```diff
@@ -0,0 +1 @@
+
```
danilo-gemoli, line 1: Autogenerated file?

psalajova: no, it has to exist so the pj-rehearse.sh test has something to work with

PR #3998, opened 2024-02-29
qci-appci: generate jwt token

hongkailiu:
This is to avoid bugging app.ci too often.

/cc @openshift/test-platform 
/assign @bear-redhat


hongkailiu:
/hold

Need to watch

PR #4005, opened 2024-03-03
Use aws as the CLUSTER_TYPE

mrhillsman:
https://steps.ci.openshift.org/reference/ipi-install-install step fails for preflight tests using rh-openshift-ecosystem as the CLUSTER_TYPE when building the install-config.yaml and attempting to create find the cloud credentials. It is my understanding that resources will still be created using the rh-openshift-ecosystem-quota-slice


mrhillsman:
breaking-changes failure is ok/expected we are working through setting up the cluster profile. The job(s)/config(s) referenced will be updated.

skattoju (approved):

droslean (approved):

PR #4009, opened 2024-03-04
Updating image to be arch agnostic

deepsm007:
It is failing on build10 as when it runs on arm64 node
/cc @bear-redhat @openshift/test-platform


deepsm007:
https://github.com/openshift/ci-tools/pull/4010

PR #4011, opened 2024-03-05
deprecate the additional architectures in prowgen config, refactor multi_arch

droslean:
Removes the `additional_architectures` configuration in prowgen config. Force the images presubmits to run on `build09`when the `multi_arch` is enabled.

/cc @openshift/test-platform @danilo-gemoli @deepsm007 


droslean:
/hold

Need a follow-up PR to the release repo. @deepsm007 Feel free to tag this if there are no suggestions. I will hold cancel and follow-up in the release repo.

pkg/prowgen/prowgen.go
```diff
@@ -100,17 +99,25 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 			presubmitTargets = append(presubmitTargets, "[release:latest]")
 		}
 		jobBaseGen := newJobBaseBuilder().TestName("images")
+		if info.Config.MultiArch {
```
danilo-gemoli, line 102: You could add this logic inside the builder once and for all

pkg/prowgen/prowgen.go
```diff
@@ -100,17 +99,25 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
...
 		presubmits[orgrepo] = append(presubmits[orgrepo], *generatePresubmitForTest(jobBaseGen, "images", info))
 
 		if configSpec.PromotionConfiguration != nil {
-			postsubmitsForPromotion, err := generatePostsubmitsForPromotion(newJobBaseBuilderForPromotion, info, func(options *generatePostsubmitOptions) {
-				options.imageTargets = imageTargets
-			})
-			if err != nil {
-				return nil, fmt.Errorf("error generating postsubmits for promotion: %w", err)
+			jobBaseGen = newJobBaseBuilder().TestName("images")
+			if info.Config.MultiArch {
```
danilo-gemoli, line 110: Same as above here

pkg/prowgen/prowgen.go
droslean: We don't want that. We need to mutate only the images presubmits and postsubmit.

danilo-gemoli: Got it

PR #4018, opened 2024-03-07
`check-gh-automation`: fix branch-protection determination logic

smg247:
Lots of failures in this [run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-check-gh-automation/1764485958339661824) due to this case.

Branch protection is enabled by default unless there is configuration stating unmanaged=true.

The unit tests were also incorrectly passing without actually looking at branch protection. I fixed those as well.

### DIFF
--- a/cmd/check-gh-automation/main.go
+++ b/cmd/check-gh-automation/main.go
@@ -191,7 +191,11 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
 		orgConfig := prowAgent.Config().BranchProtection.GetOrg(org)
 		var branchProtectionEnabled bool
 		if orgConfig != nil {
-			_, branchProtectionEnabled = orgConfig.Repos[repo]
+			branchProtection, exists := orgConfig.Repos[repo]
+			if exists {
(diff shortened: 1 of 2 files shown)


cmd/check-gh-automation/main.go
```diff
@@ -189,9 +189,16 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
 		}
 
 		orgConfig := prowAgent.Config().BranchProtection.GetOrg(org)
-		var branchProtectionEnabled bool
+		branchProtectionEnabled := true // By default, it is turned on in absence of configuration stating otherwise
```
danilo-gemoli, line 192: Changing to default to `true`, I'm wondering if we are going to hit any error. Has this tool a `dry-run` mode? Alternatively we can merge and see what happens.

smg247: It's always dry as it doesn't mutate anything. I tested it out with a few repos to make sure the behavior was correct. The default should have always been true, as that matches with the documentation on branch protection. This feature was never working as intended.

PR #4024, opened 2024-03-11
Fix TestInputImageStreamTagsFromResolvedConfig function

hongkailiu:
Follow up https://redhat-internal.slack.com/archives/CBN38N3MW/p1710169725703149?thread_ts=1710168229.675879&cid=CBN38N3MW

I believe this bug has been there since the beginning.
As it shown in the PR, `TestInputImageStreamTagsFromResolvedConfig` is used by other tools too.

However, it is not outstanding in other tools is because of the difference on `testInputImageStreamTagFilterFactory`.
https://github.com/openshift/ci-tools/blob/0a9b4d7a2992f110a4569f92a9c5690e9e3d7406/pkg/controller/test-images-distributor/test_images_distributor.go#L583
For example, the tag `openshift/release:rhel-8-release-golang-1.20-openshift-4.16` is not mirrored to QCI.
But the dptp-controller still distritubtes the image to build farms (the component has logs on the request).

/cc @openshift/test-platform

### DIFF
--- a/pkg/api/helper/imageextraction_test.go
+++ b/pkg/api/helper/imageextraction_test.go
@@ -96,7 +96,7 @@ func TestTestInputImageStreamTagsFromConfigParsing(t *testing.T) {
 			expectedResult: "foo/Baz:Bar",
 		},
 		{
-			name: "boot image from repo",
+			name: "root image from repo",
 			config: api.ReleaseBuildConfiguration{
 				InputConfiguration: api.InputConfiguration{
 					BuildRootImage: &api.BuildRootImageConfiguration{FromRepository: true}},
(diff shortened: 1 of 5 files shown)


pkg/api/helper/imageextraction_test.go
```diff
@@ -71,26 +72,42 @@ func TestTestInputImageStreamTagsFromResolvedConfigReturnsAllImageStreamTags(t *
...
-			name:           "happy path",
-			istr:           api.ImageStreamTagReference{Namespace: "foo", Name: "Baz", Tag: "Bar"},
+			name: "happy path",
+			config: api.ReleaseBuildConfiguration{
+				InputConfiguration: api.InputConfiguration{BaseImages: map[string]api.ImageStreamTagReference{"": {Namespace: "foo", Name: "Baz", Tag: "Bar"}}},
+			},
 			expectedResult: "foo/Baz:Bar",
 		},
+		{
+			name: "boot image from repo",
```
smg247, line 99:
```suggestion
			name: "root image from repo",
```

PR #4026, opened 2024-03-11
Revert "Skip periodics for the future and current releases"

jluhrsen:
Reverts openshift/ci-tools#3208

not sure if we want to revert this or if we have a different way to prevent jobs intentionally configured as periodics to not be
removed.


jluhrsen:
not something we want to do:
https://redhat-internal.slack.com/archives/CBN38N3MW/p1710232383017649?thread_ts=1709849262.455669&cid=CBN38N3MW

PR #4027, opened 2024-03-12
Handle error better in TestInputImageStreamTagsFromResolvedConfig

hongkailiu:
This is needed because `FileGetterFactory` could return `nil, nil` (which would lead zero value of tag getting inserted).

https://github.com/openshift/ci-tools/blob/a6f24219a5526f810560be36770896d77023d299/pkg/github/github.go#L56-L58

Also added more unit tests to cover those cases.

/cc @openshift/test-platform

### DIFF
--- a/pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
+++ b/pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
@@ -287,6 +287,15 @@ func imageStreamNameFromImageStreamTagName(nn types.NamespacedName) (types.Names
 	return types.NamespacedName{Namespace: nn.Namespace, Name: colonSplit[0]}, nil
 }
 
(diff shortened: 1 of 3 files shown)


pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
```diff
@@ -287,6 +287,15 @@ func imageStreamNameFromImageStreamTagName(nn types.NamespacedName) (types.Names
 	return types.NamespacedName{Namespace: nn.Namespace, Name: colonSplit[0]}, nil
 }
 
+func repoFileGetterWithIgnore(metadata cioperatorapi.Metadata) func(_, _, _ string, _ ...github.Opt) github.FileGetter {
+	// We ignore openshift-priv because it is a private org and thus would lead to 404 without providing authentication information
+	// Moreover, the .ci-operator.yaml file there is a duplication of the one from the public repo
+	if metadata.Org == "openshift-priv" {
```
smg247, line 293: Is a similar issue going to occur for repos in `qe-private-deck` like, for example, `openshift/openshift-tests-private`?

PR #4048, opened 2024-03-18
`helpdesk-faq`: server that provides an endpoint with a list of all faq items

smg247:
This is the first step in a 3-ish part processes. This PR contains the `helpdesk-faq` server that simply responds with the FAQ items in the ConfigMap. There are currently some example items in the [CM](https://console-openshift-console.apps.ci.l2s4.p1.openshiftapps.com/k8s/ns/ci/configmaps/helpdesk-faq).
For: [DPTP-3521](https://issues.redhat.com/browse/DPTP-3521)

### DIFF
--- a/cmd/helpdesk-faq/main.go
+++ b/cmd/helpdesk-faq/main.go
@@ -1,0 +1,169 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"html/template"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/sirupsen/logrus"
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/test-infra/prow/flagutil"
+	"k8s.io/test-infra/prow/interrupts"
+	"k8s.io/test-infra/prow/logrusutil"
+
+	"github.com/openshift/ci-tools/pkg/api"
+)
+
+const (
+	appCIContextName = string(api.ClusterAPPCI)
+	faqConfigMap     = "helpdesk-faq"
+	ci               = "ci"
+)
+
+type options struct {
+	logLevel          string
+	port              int
+	gracePeriod       time.Duration
+	kubernetesOptions flagutil.KubernetesOptions
+}
+
+type Page struct {
+	Data []FaqItem `json:"data"`
+}
+
+// TODO(sgoeddel): these structs will be placed in a common package somewhere so slack-bot can use them too
+type FaqItem struct {
+	Question  string   `json:"question"`
+	Timestamp string   `json:"timestamp"`
+	Author    string   `json:"author"`
+	Answers   []Answer `json:"answers"`
+}
+
+type Answer struct {
+	Author    string `json:"author"`
+	Timestamp string `json:"timestamp"`
+	Body      string `json:"body"`
+}
+
+func gatherOptions() (options, error) {
+	o := options{}
+	fs := flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	fs.StringVar(&o.logLevel, "log-level", "info", "Level at which to log output.")
+	fs.IntVar(&o.port, "port", 8080, "Port to run the server on")
+	o.kubernetesOptions.AddFlags(fs)
+	fs.DurationVar(&o.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
+	if err := fs.Parse(os.Args[1:]); err != nil {
+		return o, fmt.Errorf("failed to parse flags: %w", err)
+	}
+	return o, nil
+}
+
+func validateOptions(o options) error {
+	_, err := logrus.ParseLevel(o.logLevel)
+	if err != nil {
+		return fmt.Errorf("invalid --log-level: %w", err)
+	}
+	return o.kubernetesOptions.Validate(false)
+}
+
+func router(kubeClient kubernetes.Interface) *http.ServeMux {
+	handler := http.NewServeMux()
+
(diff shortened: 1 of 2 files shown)


cmd/helpdesk-faq/main.go
```diff
@@ -0,0 +1,169 @@
...
+	if err != nil {
+		logrus.WithError(err).Fatal("failed go gather options")
+	}
+	if err := validateOptions(o); err != nil {
+		logrus.WithError(err).Fatal("invalid options")
+	}
+	level, _ := logrus.ParseLevel(o.logLevel)
+	logrus.SetLevel(level)
+
+	kubeClient, err := o.kubernetesOptions.ClusterClientForContext(appCIContextName, false)
```
hongkailiu, line 157: nit:
Make a flag of the tool for `dry-run` and use it as the 2nd param of the function?

cmd/helpdesk-faq/main.go
```diff
@@ -0,0 +1,169 @@
...
+		page := map[string]bool{"ok": true}
+		if err := json.NewEncoder(w).Encode(page); err != nil {
+			logrus.WithError(err).WithField("page", page).Error("failed to encode page")
+		}
+	})
+
+	handler.HandleFunc("/api/v1/faq-items", func(w http.ResponseWriter, r *http.Request) {
+		logrus.WithField("path", "/api/v1/faq-items").Info("serving")
+
+		configMap, err := getConfigMap(kubeClient)
```
hongkailiu, line 92: Each endpoint leads to a call on the cluster?
Do you want to cache it?

cmd/helpdesk-faq/main.go
```diff
@@ -0,0 +1,169 @@
...
+			faqItem := &FaqItem{}
+			if err := json.Unmarshal([]byte(item), faqItem); err != nil {
+				logrus.WithError(err).Fatal("unable to unmarshall faq item")
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			page.Data = append(page.Data, *faqItem)
+		}
+
+		if callbackName := r.URL.Query().Get("callback"); callbackName != "" {
```
hongkailiu, line 110: The price to pay when reusing ci-docs frontend.
Will it keep it as simple as one table?
Otherwise, maybe start with a new frontend project?
I have no preference but I guess we have to do it soon or later to avoid blowing up the ci-docs site.

cmd/helpdesk-faq/main.go
```diff
@@ -0,0 +1,169 @@
...
+
+		configMap, err := getConfigMap(kubeClient)
+		if err != nil {
+			logrus.WithError(err).Fatal("unable to get helpdesk-faq config map")
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		page := Page{}
+		for _, item := range configMap.Data {
```
hongkailiu, line 100: I would dry out a service (as interface?) to provide faqItems so that the switching in the future to MongoDB is easier.

cmd/helpdesk-faq/main.go
```diff
@@ -0,0 +1,169 @@
...
+		}
+	})
+
+	return handler
+}
+
+func getConfigMap(kubeClient kubernetes.Interface) (*v1.ConfigMap, error) {
+	configMap, err := kubeClient.CoreV1().ConfigMaps(ci).Get(context.TODO(), faqConfigMap, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
```
hongkailiu, line 135:
```suggestion
		return nil, fmt.Errorf("failed to get configMap %s: %w", faqConfigMap, err)
```

cmd/helpdesk-faq/main.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	if configMap.Data == nil {
+		configMap.Data = make(map[string]string)
+	}
```
hongkailiu, line 140: It looks to me like the logic from the caller of the function.

cmd/helpdesk-faq/main.go
smg247: That may be a good idea. It would lead to a non-instantaneous update experience, but outside of a demo scenario, I think that is okay.

cmd/helpdesk-faq/main.go
smg247: Yeah, I think this will eventually move. The `ci-docs` frontend is a first pass at this. We will see how useful it becomes and plan a better experience based on that.

cmd/helpdesk-faq/main.go
hongkailiu: Ack. I hope we will keep it readonly as the tool grows. 😄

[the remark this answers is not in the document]
smg247: Yes, it only needs access to `app.ci`. This is a good suggestion, I will consider implementing it as a follow up. I have a few more PRs here to get the MVP deployed, and I am hoping to finish up during shift week :)

cmd/helpdesk-faq/main.go
```diff
@@ -29,6 +29,47 @@ const (
...
+	lastReload  time.Time
+}
+
+func (c *configMapClient) GetFAQItems() ([]string, error) {
+	fifteenMinutesFromLastCacheReload := c.lastReload.Add(time.Minute * 15)
+	if len(c.cachedItems) > 0 && time.Now().Before(fifteenMinutesFromLastCacheReload) {
+		logrus.Debugf("returning faq items from cache")
+		return c.cachedItems, nil
+	}
+	logrus.Debugf("reloading faq items from configmap")
```
hongkailiu, line 48: Debugf --> Debug?

hongkailiu (commented):
/hold
Free to cancel if you want to remove the `F` in logrus with another PR.

/lgtm

smg247:
/hold cancel
I will fix the nit in a future PR

PR #4057, opened 2024-03-22
pkg/default,steps: fix bundle support for multiple refs

AlexNPavel:
This PR fixes an issue where operator bundle builds fail if a job has multiple refs set in `extra_refs`. It attempts to identify which ref matches provided ci-operator config based on the config metadata.


AlexNPavel:
This still needs some more tweaking and testing, but I am using this PR to help built the ci-operator images to use in the cluster

AlexNPavel:
/close

I've found a different and significantly simpler way to fix this issue with ci-chat-bot.

PR #4069, opened 2024-04-08
[testgrid-config-generator] Allow arbitrary testgrid dashboards and multiple allow lists

pierDipi:
- Allow arbitrary testgrid dashboard roles
- Allow specifying multiple allow list files so that teams can manage their own allow list files without requiring core team reviews


pierDipi:
Closing for now until team decides what to do

PR #4071, opened 2024-04-08
ci-scheduling-webhook: don't overwrite `.spec.nodeSelector`

danilo-gemoli:
`ci-scheduling-webhook` should be gentle and add only its own stuff into the map instead of replacing it entirely.

/cc @bear-redhat @droslean


droslean:
Can we add a test for the logic we want to resolve in this PR?

PR #4074, opened 2024-04-09
Revert "add cluster profile for konflux workspaces"

sadlerap:
This cluster profile is no longer needed.  It has no current users, and it will not be used in the future.

This reverts commit 18b957af294ec2cf4a532188131ae6658c581f0c.


sadlerap:
For context: after discussions with the konflux QE team, I was informed that we will not need our own cluster profile use on https://github.com/konflux-workspaces/workspaces.

cc: @ascerra

PR #4079, opened 2024-04-15
Include ComplianceAsCode periodic jobs into Sippy

rhmdnd:
The periodic jobs from ComplianceAsCode test different OpenShift versions
against various compliance benchmarks. Let's pull those jobs into Sippy
so we can start viewing the information in dashboards.


rhmdnd:
This is being done in https://github.com/openshift/ci-tools/pull/4078

PR #4101, opened 2024-05-02
releaseSnapshotStep: no dependencies on isTags of the local cluster

hongkailiu:
Follow up https://redhat-internal.slack.com/archives/CHY2E1BL4/p1714644855129709?thread_ts=1714643460.138379&cid=CHY2E1BL4

For the cluster-bot jobs, we get the tags from the stream in the local cluster, as we did before.

/cc @openshift/test-platform 

/hold

Will verify with the built image

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -30,6 +30,7 @@ import (
 	imagev1 "github.com/openshift/api/image/v1"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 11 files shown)


hongkailiu:
/hold

```console
curl -s https://config.ci.openshift.org/integratedStream\?namespace\=ocp-private\&name\=4.15-priv
failed to get image stream ocp-private/4.15-priv: imagestreams.image.openshift.io "4.15-priv" is forbidden: User "system:serviceaccount:ci:default" cannot get resource "imagestreams" in API group "image.openshift.io" in the namespace "ocp-private"

```

Need to fix this first.

PR #4103, opened 2024-05-02
Reuse utils.WaitForImportingISTag in import_release

hongkailiu:
The replaced implementation uses different API from the one in `utils`.
But they have the same feature that is "wait until all tags are imported on an image stream or timeout".

/cc @openshift/test-platform


hongkailiu:
/hold

Will watch the rolling

PR #4108, opened 2024-05-07
pj-rehearse: No dependencies of istags on the local cluster

hongkailiu:
This is the 2nd case we got from the user.
The cause is that the mirroring of external images to CI is targeting QCI now.
However `pj-rehearse` still mirrors image from app.ci (to build farms) and QCI.
Then we hit the error because of the missing source image on app.ci.

I think this is the time that we stop
- checking images on build farms: `ensureImageStreamTags`, and
- mirroring images to QCI (introduced by the PR and used by rehearsal): https://issues.redhat.com/browse/DPTP-3906

(Only the first item above depends on the istags on the local cluster. But both of them share the same goal: making the images available for CI tests,  for either the local cluster or QCI. When QCI becomes CI authentic registry, neither of them would be necessary. We could remove them with 2 PRs.) 

If either of the above failed, that would imply that respectively
- jobs still use istags from local cluster, (and then we still need `test-images-distributor` to distribute the image from app.ci to build-farms), or
- QCI is still not the source of the truth of CI images.

If this PR goes well in production, we will remove the flags

```
         registryConfig           string
	ciImagesMirrorConfigPath string
```

This is a warm-up exercise before decommissioning [test-images-distributor](https://issues.redhat.com/browse/DPTP-3998). 

/cc @openshift/test-platform 
/assign @Prucek


hongkailiu:
> is QCI the source of truth already?

That is my assumption.

hongkailiu:
/hold

Found an interesting thing about this.
https://redhat-internal.slack.com/archives/C01C8502FMM/p1715261844358519?thread_ts=1713208112.920919&cid=C01C8502FMM

PR #4112, opened 2024-05-09
Update how ci-operator gets cluster profile secrets

psalajova:
Use config-resolver to get the name of cluster profile secret we need, instead of mounting it to the job and then reading it locally along with the other user-defined secrets.

Next steps:
- create a presubmit job to check the cluster profile secret (defined in the test config) exists in the ci namespace
- edit prowgen so it doesn't add the cluster profile volume to the job yaml, as it won't be needed anymore & regenerate all jobs

/cc @droslean

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -388,7 +388,7 @@ type options struct {
 	leaseServerCredentialsFile string
 	leaseAcquireTimeout        time.Duration
 	leaseClient                lease.Client
-	clusterProfileNames        []string
+	clusterProfiles            []clusterProfileForTarget
 
 	givePrAuthorAccessToNamespace bool
 	impersonateUser               string
@@ -1453,10 +1453,10 @@ func (o *options) initializeNamespace() error {
 
 	// adds the appropriate cluster profile secrets to o.secrets,
 	// so they can be created by ctrlruntime client in the for cycle below this one
-	for _, cp := range o.clusterProfileNames {
+	for _, cp := range o.clusterProfiles {
 		cpSecret, err := getClusterProfileSecret(cp, ctrlClient, o.resolverClient, ctx)
 		if err != nil {
-			return fmt.Errorf("failed to create cluster profile  secret %s: %w", cp, err)
+			return fmt.Errorf("failed to create cluster profile secret %s: %w", cp, err)
 		}
 		cpSecret.Namespace = o.namespace
 		o.secrets = append(o.secrets, cpSecret)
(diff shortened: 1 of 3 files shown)


cmd/ci-operator/main.go
```diff
@@ -2243,3 +2257,38 @@ func addSchemes() error {
...
+	err = client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: "ci", Name: cpDetails.Secret}, ciSecret)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get secret '%s' from ci namespace: %w", cpDetails.Secret, err)
+	}
+
+	return ciSecret, nil
+}
+
+// getClusterProfileNamesFromTargets extracts the needed cluster profile name(s) from the target arg(s)
+func (o *options) getClusterProfileNamesFromTargets() {
```
droslean, line 2280: We could use unittests for this.

cmd/ci-operator/main_test.go
```diff
@@ -1612,3 +1614,109 @@ func TestHandleTargetAdditionalSuffix(t *testing.T) {
...
+
+func (m *mockResolverClient) Resolve(data []byte) (*api.ReleaseBuildConfiguration, error) {
+	return nil, nil
+}
+
+func (m *mockResolverClient) IntegratedStream(namespace string, name string) (*configresolver.IntegratedStream, error) {
+	return nil, nil
+}
+
+func TestGetClusterProfileSecret(t *testing.T) {
```
droslean, line 1645: Technically you are not testing anything. The main operation that is being done to retrieve the cluster profiles is being tested in https://github.com/psalajova/ci-tools/blob/35c5f8f232ef977adf77ce20c237560df3e4eea8/pkg/registry/server/client_test.go#L88

cmd/ci-operator/main.go
```diff
@@ -2243,3 +2257,38 @@ func addSchemes() error {
...
+	for _, targetName := range o.targets.values {
+		for _, test := range o.configSpec.Tests {
```
droslean, line 2282: Consider having a helper method on the Tests object to retrieve the cluster profiles.

cmd/ci-operator/main_test.go
psalajova: hm, right

cmd/ci-operator/main.go
```diff
@@ -632,6 +633,11 @@ func (o *options) Complete() error {
 	}
 
 	for _, path := range o.secretDirectories.values {
+		// TODO: Delete this after regenerating all jobs with the new cluster profile system.
+		if strings.Contains(path, "-cluster-profile") {
```
droslean, line 637: Since this is a suffix, we should check that directly.. `strings.HasSuffix`?

cmd/ci-operator/main.go
droslean: You can add a method to the `Tests` corresponding struct to operator on the cluster profiles

pkg/api/types.go
```diff
@@ -805,6 +805,19 @@ func (config TestStepConfiguration) IsPeriodic() bool {
...
+func (config TestStepConfiguration) GetClusterProfileName() string {
+	switch {
+	case config.MultiStageTestConfigurationLiteral != nil:
+		return config.MultiStageTestConfigurationLiteral.ClusterProfile.Name()
+	case config.MultiStageTestConfiguration != nil:
+		return config.MultiStageTestConfiguration.ClusterProfile.Name()
+	case config.OpenshiftInstallerClusterTestConfiguration != nil:
+		return config.OpenshiftInstallerClusterTestConfiguration.ClusterProfile.Name()
+	}
+	return ""
```
droslean, line 818: you can add the `default` in the switch and you don\t have to return.. However, is the default empty string a valid case?

psalajova: Ok. And yeah empty string is valid, I deal with it in another function. Only jobs which have cluster profiles defined have the value something other than the empty string.

cmd/ci-operator/main.go
droslean: check also https://github.com/psalajova/ci-tools/blob/2a6fab78c976f5188ae62af7a5e7ec51ddf5de78/pkg/api/types.go#L800-L806 for example

psalajova:
/hold 
I want to wait until tomorrow morning so I'm present to resolve any issues

PR #4128, opened 2024-05-16
Handle known bad versions to suppress panic

neisw:
Adds a check for version strings 414,415,416 due to errors we were seeing from https://github.com/openshift/release/pull/51560/


neisw:
/hold
not seeing the errors currently, may not be needed

PR #4130, opened 2024-05-17
`ci-operator`: Allow for the release of unused BYOIP leases

smg247:
BYOIP automatically leases `13` ip-pool addresses in the appropriate region when using one of the (for now, hardcoded) relevant cluster profiles. In many cases, less than `13` addresses are actually needed. In order to maximize the savings from these ip pools we must provide a way to release unused leases ASAP. There aren't really many ways to do this without a significant re-architecture of `ci-operator`, but this represents the least intrusive way I could imagine. We look for the presence of an `UNUSED_IP_COUNT` item in the shared-dir secret, and if it is populated by an int, we release that number of leases immediately. If this item isn't found within fifteen minutes, then we stop checking for it. The workflow step will have to be updated to store the proper number in a file with the aforementioned name in the shared-dir.

Verified functionality with a [run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-ci-openshift-release-master-nightly-4.17-e2e-aws-ovn-upi-sgoeddel/1791498360352411648) of a special job using my, locally-built, `ci-operator` image.

For: https://issues.redhat.com/browse/DPTP-3929

### DIFF
--- a/pkg/steps/ip_pool.go
+++ b/pkg/steps/ip_pool.go
@@ -5,9 +5,12 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/sirupsen/logrus"
 
+	coreapi "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	"github.com/openshift/ci-tools/pkg/api"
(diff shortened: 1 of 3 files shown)


smg247:
Looks like I have a race in my unit test. I am looking into a way to avoid it. I think the rest of the code is good to review in the meantime if anyone is looking.

pkg/steps/ip_pool.go
```diff
@@ -105,9 +110,70 @@ func (s *ipPoolStep) run(ctx context.Context) error {
...
 	return aggregateWrappedErrorAndReleaseError(wrappedErr, releaseErr)
 }
+
+const UnusedIpCount = "UNUSED_IP_COUNT"
+
+// checkAndReleaseUnusedLeases periodically checks for a positive value in the
+// UNUSED_IP_COUNT data in the shared secret which signals that there are leases that can be released.
+// If/when this is discovered it will release that number of leases, and stop checking.
+// minute is provided as an argument to assist with unit testing.
+func (s *ipPoolStep) checkAndReleaseUnusedLeases(ctx context.Context, minute time.Duration) {
```
deepsm007, line 130: Can we add extra debug log to understand the average wait time for an ipPool Lease? This will help us calibrate it better and avoid longer wait times.

smg247: We don't wait at all, if they aren't available immediately we move on. The 15 minute timeout is just for checking if the `UNUSED_IP_COUNT` file is present, this will be set by the workflow step, and will likely be present in the first minute or two. This just provides a buffer in case the install step takes longer than anticipated.

I am still working through the fix to the race here, but the idea with this function is that it is async and lightweight. It won't affect test times at all.

smg247:
The `e2e` failure is a flake, I will let it retest when this has an `lgtm` on it.

PR #4136, opened 2024-05-22
`BYOIP`: simplify configurations for other cluster-profiles to utilize

smg247:
Originally, this was hardcoded to only work on the `aws` cluster-profile. This configuration change will make it possible for others to be added as desired. I will be working on adding documentation for this in the cluster-profile area. There is an, in progress, change to cluster-profile configuration that we will be reworking these configs into once it lands.

Also fixed a flaky race condition with the, previously added, unit tests.

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1946,6 +1946,16 @@ func (p ClusterProfile) IPPoolLeaseType() string {
 	}
 }
 
+// IPPoolLeaseShouldValidateBranch declares whether the ip-pool leases should only be applied to branches matching a
+// specific OpenShift validation model. returns true by default, but should return false for any cluster-profiles
+// that don't want this validation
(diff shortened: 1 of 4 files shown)


pkg/api/types.go
```diff
@@ -1946,6 +1946,16 @@ func (p ClusterProfile) IPPoolLeaseType() string {
 	}
 }
 
+// IPPoolLeaseShouldValidateBranch declares whether the ip-pool leases should only be applied to branches matching a
+// specific OpenShift validation model. returns true by default, but should return false for any cluster-profiles
```
deepsm007, line 1950: Ideal condition would be to set it false by default and opt in? (based on only the cloud platforms are using it in this case for now aws)

smg247: I thought about that, but I think the vast majority of use cases are going to want true. So I made it opt out.

PR #4138, opened 2024-05-23
trt-1627: fix up table references

neisw:
Seeing failures like
```
error getting historical disruption data, skipping: googleapi: Error 400: Unrecognized name: StartTime at [6:5], invalidQuery 
```


neisw:
/assign dgoodwin

You might have a better fix in mind

PR #4143, opened 2024-05-29
pr-reminder: improve signal to noise ratio

stevekuznetsov:


hongkailiu (approved):

hongkailiu (approved):

smg247:
This looks like good functionality to me with only the caveat that it will definitely use more gh tokens than the current design. I think the `ghproxy` cache will help us out a lot here, though, so we might be okay. Since this tool is only run once daily during non-core hours, we can get this merged and see if it ends up being a problem.

Only other note, were you not able to find any ways to unit test this?

stevekuznetsov:
@smg247 the first commit does the draft checking (by proxy, since it looks at the WIP labels)

I can add a unit test here for the new function.

PR #4156, opened 2024-06-04
pr-reminder: allow omitting bots

stevekuznetsov:
The current design of the configuration for this tool doesn't validate that the same user/repo combination is only provided by one configuration item, so it's not really possible to filter DMs for bot activity, since we would need a different factoring to know which configuration we were using.

/assign @smg247

### DIFF
--- a/cmd/pr-reminder/main.go
+++ b/cmd/pr-reminder/main.go
@@ -164,14 +164,17 @@ func (c *config) createUsers(gtk githubToKerberos, slackClient slackClient) (map
 	return users, kerrors.NewAggregate(errors)
 }
 
-func (c *config) channels() map[string]sets.Set[string] {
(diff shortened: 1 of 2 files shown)


smg247 (commented):
/lgtm

stevekuznetsov:
```
STEP 3/8: RUN dnf install -y git diffutils && dnf clean all && rm -rf /var/cache/dnf
CentOS Stream 8 - AppStream                     194  B/s |  38  B     00:00    
Error: Failed to download metadata for repo 'appstream': Cannot prepare internal mirrorlist: No URLs in mirrorlist
error: build error: building at STEP "RUN dnf install -y g...m -rf /var/cache/dnf": while running runtime: exit status 1
```

/retest

PR #4164, opened 2024-06-07
Adding new image for ci-operator with ubi base image

deepsm007:
Adding a new image for smooth transition from centos:stream9 to ubi-minimal

/cc @danilo-gemoli @openshift/test-platform

### DIFF
--- a/images/ci-operator-minimal/Dockerfile
+++ b/images/ci-operator-minimal/Dockerfile
@@ -1,0 +1,8 @@
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+
+RUN microdnf install -y git python3
+RUN alternatives --install /usr/bin/python python /usr/bin/python3.9 1
+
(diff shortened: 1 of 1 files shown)


images/ci-operator-minimal/Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
+
+RUN microdnf install -y git python3
```
droslean, line 3: Can you check if we can deprecate that? IIRC the git dependency exists to allow users to pass their own git revisions to test stuff. I thing we long passed this idea. Also, python3, why?

danilo-gemoli: I can't tell anything about git but python is required, see [this](https://redhat-internal.slack.com/archives/CBN38N3MW/p1717595757867709)

PR #4174, opened 2024-06-11
`helpdesk-faq`: remove special slack formatting from links

smg247:
Slack [adds](https://api.slack.com/reference/surfaces/formatting#escaping) a "<" and ">" around links from their api. We don't want this formatting, and must remove it prior to saving down the question or answer.


smg247:
For: https://issues.redhat.com/browse/DPTP-3521

PR #4175, opened 2024-06-12
Fix issue DPTP-4032 Add flag to restrict tests accessing RH intranet

bear-redhat:
cc. @danilo-gemoli


cmd/ci-operator/main.go
```diff
@@ -1285,6 +1289,47 @@ func (o *options) initializeNamespace() error {
...
+						Type: "Allow",
+					},
+				},
+			},
+		}
+		if err := client.Create(ctx, egressFirewall); err != nil && !kerrors.IsAlreadyExists(err) {
+			return fmt.Errorf("could not create egress firewall: %w", err)
+		}
+	} else {
+		logrus.Debugf("Not creating egress firewall in namespace %s", o.namespace)
```
danilo-gemoli, line 1330: The EgressFirewall is not being created by default. It's ok to log and inform the user when `--restrict-network-access=true` but I believe we can avoid logging when no firewall rules have been created.

pkg/api/egressfirewall/NOTE.md
```diff
@@ -0,0 +1,10 @@
+This directory is generated by
```
droslean, line 1: Why can't we use their library directly?

danilo-gemoli: https://issues.redhat.com/browse/DPTP-4064 Cause our repo is in a really bad state and I'm fighting against it since Friday.

droslean: I am against that idea. If you are fighting with the vendoring, which most probably you need to force a specific version, coping-pasting manually the library into the pkg folder is a really bad idea.

danilo-gemoli: I agree but it's a temporary workaround that we are going to eliminate as soon as I manage to solve the dependency issue we have (I'm on it already). Bear delayed his work for several weeks and at some point we had to do something.

PR #4182, opened 2024-06-14
OSASINFRA-3477: openstack: Add new location for OpenStack periodics to Sippy

pierreprinetti:
Jobs will be added in https://github.com/openshift/release/pull/53236

We will remove the old locations as a follow-up.


mandre (commented):
/lgtm

PR #4185, opened 2024-06-17
install git for job-run-aggregator for weekly disruption PR

DennisPeriquet:
The latest [weekly PR run](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/periodic-update-origin-disruption-alert-data/1801419585887408128) shows:

```
failed to check for changes: running command git [status --porcelain]: exec: \"git\": executable file not found in $PATH
```

indicating the `git` executable needs to be installed.


DennisPeriquet:
Let me try to run the whole procedure using:

```bash
$ podman run -it --rm --entrypoint /bin/bash registry.ci.openshift.org/ci/job-run-aggregator:latest
```

before we merge this in case there are other things missing

PR #4201, opened 2024-06-25
`autoowners`: changes the functionality of the 'ignore-repo' argument

smg247:
It used to be a misnomer, and actually expect a directory path to ignore. I believe this was to allow the possibility for files to be synced for some dirs, but not others. In practice, this wasn't being used this way and each of the directory paths were passed for each repo. The one exception was a 'template' path, which is irrelevant and deprecated anyway.

Follow up to https://github.com/openshift/ci-tools/pull/4199

I will change the job config file in a follow up shortly.

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -93,6 +93,11 @@ func loadRepos(configRootDir string, blocked blocklist, configSubDirs, extraDirs
 				if org == githubOrg && repo == githubRepo {
 					continue
 				}
+				orgRepoName := fmt.Sprintf("%s/%s", org, repo)
+				if blocked.repos.Has(orgRepoName) {
+					logrus.WithField("repository", orgRepoName).Info("repository is on repository blocklist, skipping")
+					continue
+				}
 				var repoConfigDirs []string
 				for _, sourceSubDir := range configSubDirectories {
 					repoConfigDirs = append(repoConfigDirs, filepath.Join(sourceSubDir, org, repo))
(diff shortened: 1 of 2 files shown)


hongkailiu:
> In practice, this wasn't being used this way and each of the directory paths were passed for each repo. 

I assume that you have checked each of the existing repo/dir already.
What if a team comes in the future and wants to sync in config dir but not job dir? Unlikely but possible. :)

> The one exception was a 'template' path, which is irrelevant and deprecated anyway.

IIUC, `--ignore-repo=ci-operator/templates/openshift/installer` becomes `--ignore-repo=openshift/installer` after merging of this PR. The owners file will be ignored in the config dir and the job dir. Is it what the team want?

I would correct the name of the flag and keep the granularity, e.g., to something like `--ignore-dir`.

smg247:
> I assume that you have checked each of the existing repo/dir already.

yes, https://github.com/openshift/release/pull/53646 details the planned changes to the job's config.

> What if a team comes in the future and wants to sync in config dir but not job dir? Unlikely but possible. :)
I don't think that is a use case that we need to support.

> IIUC, --ignore-repo=ci-operator/templates/openshift/installer becomes --ignore-repo=openshift/installer after merging of this PR. The owners file will be ignored in the config dir and the job dir. Is it what the team want?
I am actually completely removing that ignore, because we don't care about templates anymore. I doubt they will even notice, but if they do somehow we can workaround the problem.

> I would correct the name of the flag and keep the granularity, e.g., to something like --ignore-dir.
The goal here wasn't to make the flag name more clear, but to make an easier to use flag. Each repo would actually need to be added 3 times to use the current flag, otherwise it will still load the config for the `prow config` dir. Even still, it appears that the config was being loaded, and would result in errors if it was improperly formatted. As was the case with `openshift/velero`. I am willing to abandon this effort if you believe the current structure is preferred as you have greater historical context on this tool.

hongkailiu:
It seems that you have plans for all the coming cases. :) 

/lgtm

PR #4206, opened 2024-06-28
Add default secret logic for cluster profiles

psalajova:
Implemented new logic; when loading cluster profiles from their [config file](https://github.com/openshift/release/blob/master/ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml), we autogenerate the secret name for each cluster profile. If a profile had the secret field defined in the config, we will use that one (this is mainly for keeping backwards compatibility with a few  derivative cluster profiles that don't follow the naming convention; I added those to the config file in https://github.com/openshift/release/pull/53787). 

The result of this is that `config-resolver` will be able to provide the secret for each cluster profile. 
Now:
```
$ curl "https://config.ci.openshift.org/clusterProfile?name=packet"
{
  "profile": "packet"
}
```
After this PR merges:
```
$ curl "https://config.ci.openshift.org/clusterProfile?name=packet"
{
  "profile": "packet",
  "secret": "cluster-secrets-packet"
}
```
/cc @droslean @smg247

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1997,6 +1997,11 @@ func (p ClusterProfile) Secret() string {
 	return fmt.Sprintf("cluster-secrets-%s", name)
 }
 
+// GetClusterProfileSecret returns the default secret name for the profile
+func GetClusterProfileSecret(profile ClusterProfile) string {
+	return fmt.Sprintf("cluster-secrets-%s", string(profile))
+}
+
 // LeaseTypeFromClusterType maps cluster types to lease types
 func LeaseTypeFromClusterType(t string) (string, error) {
 	switch t {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1997,8 +1997,8 @@ func (p ClusterProfile) Secret() string {
 	return fmt.Sprintf("cluster-secrets-%s", name)
 }
 
-// GetClusterProfileSecret returns the default secret name for the profile
-func GetClusterProfileSecret(profile ClusterProfile) string {
+// GetDefaultClusterProfileSecretName returns the default secret name for the profile
+func GetDefaultClusterProfileSecretName(profile ClusterProfile) string {
 	return fmt.Sprintf("cluster-secrets-%s", string(profile))
 }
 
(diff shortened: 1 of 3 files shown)


pkg/api/types.go
```diff
@@ -1997,6 +1997,11 @@ func (p ClusterProfile) Secret() string {
 	return fmt.Sprintf("cluster-secrets-%s", name)
 }
 
+// GetClusterProfileSecret returns the default secret name for the profile
+func GetClusterProfileSecret(profile ClusterProfile) string {
```
smg247, line 2001: Since this isn't looking at the configuration, it should be named `GetDefaultClusterProfileSecretName` to be more explicit.

pkg/load/load.go
```diff
@@ -255,32 +255,38 @@ func ClusterProfilesConfig(configPath string) (api.ClusterProfilesMap, error) {
...
+		profilesFromConfigMap[p.Profile] = p
 	}
 
 	mergedMap := make(api.ClusterProfilesMap)
 	for _, profileName := range api.ClusterProfiles() {
-		profile, found := profileOwnersMap[profileName]
+		profile, found := profilesFromConfigMap[profileName]
 		if found {
+			if profile.Secret == "" {
+				profile.Secret = api.GetClusterProfileSecret(profileName) // get the default secret name
```
smg247, line 275: Then, you won't need this comment :)

pkg/load/load.go
```diff
@@ -255,32 +255,38 @@ func ClusterProfilesConfig(configPath string) (api.ClusterProfilesMap, error) {
...
+				Secret:  api.GetClusterProfileSecret(profileName),
+			}
 		}
 	}
 
 	return mergedMap, nil
 }
 
-// ClusterClaimsOwnersConfig loads cluster claim owners information from its config in the release repository
+// ClusterClaimOwnersConfig ClusterClaimsOwnersConfig loads cluster claim owners information from its config in the release repository
```
smg247, line 289:
```suggestion
// ClusterClaimOwnersConfig loads cluster claim owners information from its config in the release repository
```

PR #4208, opened 2024-07-01
github-ldap-user-group-creator: include a group with all rover usernames

droslean:
This change will allow the tool to create the group `redhat-rover-users` to all clusters that will hold ALL rover users. 

/cc @openshift/test-platform @hongkailiu 


hongkailiu:
It is a subset of all users.
See our discussion https://redhat-internal.slack.com/archives/GB7NB0CUC/p1719579819933669?thread_ts=1719573614.675169&cid=GB7NB0CUC

PR #4203, opened 2024-06-27
Sort out dependencies

danilo-gemoli:
This PR deals with:
- upgrade the build root to use go v1.22.x
- use the new import paths for Prow [sigs.k8s.io/prow](http://sigs.k8s.io/prow)
- bump some important modules to a reasonably recent version: controller-runtime , [k8s.io/client-go](http://k8s.io/client-go), [k8s.io/api](http://k8s.io/api) etc.
- remove every replace in go.mod with the exception of `github.com/google/go-cmp => github.com/alvaroaleman/go-cmp v0.5.7-0.20210615160450-f8688cd5aaa0`

PR #4211, opened 2024-07-03
Change test-infra to kubernetes-sigs/prow

psalajova:
> This script checks that the vendored version of the test-infra repo is not newer than those vendored into the release controller and ci chat bot.

Since prow changed repositories from kubernetes/test-infra to kubernetes-sigs/prow, and we are updating our dependencies to work with this new import path, this script (used in the `pull-ci-openshift-ci-tools-master-validate-test-infra` presubmit job) should reflect this change as well. 

Should probably be merged after https://github.com/openshift/ci-tools/pull/4203 merges.

/cc @danilo-gemoli @jmguzik


hack/check-test-infra-version.sh
```diff
@@ -29,14 +29,14 @@ ci_chat_bot_vendored_commit="$( determine_vendored_commit ./../ci-chat-bot )"
 
 failures=0
 
-pushd ./../../kubernetes/test-infra
+pushd ./../../kubernetes-sigs/prow
```
psalajova, line 32: I'm not sure about this line..

danilo-gemoli: It should be fine

danilo-gemoli:
Can you also change the script filename from into `hack/check-prow-version.sh`?

PR #4212, opened 2024-07-03
fix high severity findings from the snyk scan

smg247:
`repo-init-ui` and `job-runtime-analyzer` are effectively unused, so this is pretty low priority, but I am tired of seeing this job fail. `hack/jira.py` is entirely unused, and I don't think beneficial for our team any longer. I am hoping with all the high sev alerts fixed it will succeed, but we may have to fix the others as well.


smg247:
Despite the attempt to sanitize and escape the input, Snyk still doesn't like it. These tools are low-priority anyways, and spending anymore time on this to make the job green isn't worth it at this point. There will still be 20+ Medium severity issues keeping this from being green.
/close

PR #4214, opened 2024-07-09
tiny post-GA tool to reduce exec frequency of periodics

jmguzik:
UTs are not needed as rules will be adjusted quite frequently if needed and also exec on repo can be treated as test.


psalajova:
Can you add README?

PR #4217, opened 2024-07-10
`prowgen`: allow the ability to configure the slack reporter config

smg247:
This also allows the ability to use the slack reporter functionality for postsubmits and periodics. It leaves the possibility for existing slack reporter configs in periodics to behave as they currently do, with the option to remove that later once configs are migrated.

The new integration test case showcases an example of how this can be used, but I will be following this up with a change to the docs and an announcement.
Example usage in `.prowgen.config`, this would generate slack_reporter configs in the `unit`, `upload-results`, and `lint` jobs.
```
slack_reporter:
- channel: "#slack-channel"
  job_states_to_report:
  - success
  - failure
  - error
  report_template: '{{if eq .Status.State "success"}} :rainbow: Job *{{.Spec.Job}}*
                           ended with *{{.Status.State}}*. <{{.Status.URL}}|View logs> :rainbow: {{else}}
                           :volcano: Job *{{.Spec.Job}}* ended with *{{.Status.State}}*. <{{.Status.URL}}|View
                           logs> :volcano: {{end}}'
  job_names:
  - unit
  - upload-results
  - lint
```

Fixes https://issues.redhat.com/browse/DPTP-3366, and completes the enhancement in https://issues.redhat.com/browse/DPTP-2875

### DIFF
--- a/pkg/config/load.go
+++ b/pkg/config/load.go
@@ -6,10 +6,15 @@ import (
 	"os"
 	"path"
 	"path/filepath"
+	"slices"
 	"strings"
 
 	"github.com/ghodss/yaml"
 	"github.com/sirupsen/logrus"
+
+	utilerrors "k8s.io/apimachinery/pkg/util/errors"
+	"k8s.io/apimachinery/pkg/util/sets"
+	prowv1 "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 
 	cioperatorapi "github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/util"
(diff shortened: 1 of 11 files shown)


pkg/config/load.go
```diff
@@ -37,6 +42,27 @@ type Prowgen struct {
...
+	SlackReporterConfigs []SlackReporterConfig `json:"slack_reporter,omitempty"`
+}
+
+// SlackReporterConfig groups test names to a channel to report; mimicking Prow's version, with some unnecessary fields removed
+type SlackReporterConfig struct {
+	Channel           string                `json:"channel,omitempty"`
+	JobStatesToReport []prowv1.ProwJobState `json:"job_states_to_report,omitempty"`
+	ReportTemplate    string                `json:"report_template,omitempty"`
+	JobNames          []string              `json:"job_names,omitempty"`
+	//TODO(sgoeddel): if desired, we could add a list of excluded or included variants here to limit which jobs
```
droslean, line 55: Filtering the variants makes sense. Let's include this change in this PR as well.

pkg/config/load.go
```diff
@@ -87,9 +113,32 @@ func LoadProwgenConfig(folder string) (*Prowgen, error) {
...
+
 	return pConfig, nil
 }
 
+func validateProwgenConfig(pConfig *Prowgen) error {
+	var errs []error
+	if len(pConfig.SlackReporterConfigs) > 1 { // There is no reason to validate if we only have one slack_reporter_config
+		jobsSeen := sets.NewString()
+		for _, sc := range pConfig.SlackReporterConfigs {
+			for _, job := range sc.JobNames {
```
droslean, line 130: Using a continue and getting rid of the else statement feels more idiomatic.

```go
for _, sc := range pConfig.SlackReporterConfigs {
    for _, job := range sc.JobNames {
        if jobsSeen.Has(job) {
            errs = append(errs, fmt.Errorf("job: %s exists in multiple slack_reporter_configs, it should only be in one", job))
            continue
        }
        jobsSeen.Insert(job)
    }
}
```

pkg/config/load.go
smg247: Good to hear it is reasonable. I would like to do that as a follow up though, if that is okay with you.

droslean: Follow-up SGTM

droslean (approved):

PR #4220, opened 2024-07-15
`ci-operator`: promote to all registries in parallel

smg247:
This is some low-hanging fruit for `images` jobs' performance improvements. Publishing images to `quay.io` and the internal registry can be done in parallel with no concerns.

The changes to `defaults.go` are just to make it more clear that the `postSteps` we are referring to here are specifically `promotion` steps.

I have done a few [runs](https://prow.ci.openshift.org/job-history/gs/test-platform-results/logs/branch-ci-openshift-ci-tools-master-images-sgoeddel) of this now (with an image built on my machine), and it is showing a speed-up of _at least_ 20 mins over the best times in the production image.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -894,7 +894,7 @@ func (o *options) Run() []error {
 
 	mergedConfig := o.injectTest != "" || len(o.jobSpec.ExtraRefs) > 1
 	// load the graph from the configuration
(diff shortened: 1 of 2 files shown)


smg247:
@deepsm007 this is ready if you want to give it another look

deepsm007:
> @deepsm007 this is ready if you want to give it another look

Results are promising.
/lgtm

PR #4225, opened 2024-07-19
remove Alvaro and Steve from ownership of tools

smg247:
These don't really need to have a `maintainer` listed in the image


bear-redhat:
/;gtm

PR #4231, opened 2024-07-24
[DPTP-4093] [ci-operator] introduce a configuration to control the multi arch builds for single images

droslean:
Add the `multi_arch` configuration to the `images` stanza. It defaults to false which is a breaking change.

[DPTP-4093](https://issues.redhat.com/browse/DPTP-4093)

/cc @openshift/test-platform @smg247 


droslean:
/hold

A follow-up PR is needed in the release repo to update the existing repos using multi-arch builds.

PR #4236, opened 2024-07-24
DPTP-4046: Allow more registry domains for OSD clusters

deepsm007:
/cc @openshift/test-platform

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -621,27 +621,35 @@ func mutateGlobalPullSecret(original, secret *coreapi.Secret) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("failed to parse the constructed secret: %w", err)
 	}
-	registryDomain := api.DomainForService(api.ServiceRegistry)
-	if dockerConfig.Auths == nil || dockerConfig.Auths[api.DomainForService(api.ServiceRegistry)].Auth == "" {
-		return false, fmt.Errorf("failed to get token for %s", registryDomain)
+	if dockerConfig.Auths == nil {
+		return false, fmt.Errorf("failed to get any token")
 	}
-	token := dockerConfig.Auths[api.DomainForService(api.ServiceRegistry)].Auth
-	dockerConfig, err = dockerConfigJSON(original)
+	originalDockerConfig, err := dockerConfigJSON(original)
 	if err != nil {
 		return false, fmt.Errorf("failed to parse the original secret: %w", err)
 	}
-	if dockerConfig.Auths[api.DomainForService(api.ServiceRegistry)].Auth == token {
-		return false, nil
+	mutatedSecret := false
+	domains := []string{api.DomainForService(api.ServiceRegistry), api.QCIAPPCIDomain, api.QuayOpenShiftCIRepo, api.QCICacheDomain}
+	for _, domain := range domains {
+		if dockerConfig.Auths[domain].Auth == "" {
+			return false, fmt.Errorf("failed to get token for %s", domain)
+		}
+		originalToken := originalDockerConfig.Auths[domain].Auth
+		originalDockerConfig.Auths[domain] = secretbootstrap.DockerAuth{
+			Auth: dockerConfig.Auths[domain].Auth,
+		}
+		if !mutatedSecret && originalToken != originalDockerConfig.Auths[domain].Auth {
+			mutatedSecret = true
+		}
 	}
-	dockerConfig.Auths[api.DomainForService(api.ServiceRegistry)] = secretbootstrap.DockerAuth{
-		Auth: token,
+	if mutatedSecret {
+		data, err := json.Marshal(originalDockerConfig)
+		if err != nil {
+			return false, fmt.Errorf("failed to marshal the docker config: %w", err)
+		}
+		original.Data[coreapi.DockerConfigJsonKey] = data
 	}
-	data, err := json.Marshal(dockerConfig)
-	if err != nil {
-		return false, fmt.Errorf("failed to marshal the docker config: %w", err)
-	}
-	original.Data[coreapi.DockerConfigJsonKey] = data
-	return true, nil
+	return mutatedSecret, nil
 }
 
 func dockerConfigJSON(secret *coreapi.Secret) (*secretbootstrap.DockerConfigJSON, error) {
--- a/cmd/ci-secret-bootstrap/main_test.go
+++ b/cmd/ci-secret-bootstrap/main_test.go
@@ -3101,7 +3101,7 @@ func TestMutateGlobalPullSecret(t *testing.T) {
 					".dockerconfigjson": []byte(`{"key":"value"}`),
 				},
 			},
-			expectedErr: fmt.Errorf("failed to get token for registry.ci.openshift.org"),
+			expectedErr: fmt.Errorf("failed to get any token"),
 		},
 		{
 			name: "bad original",
@@ -3139,6 +3139,15 @@ func TestMutateGlobalPullSecret(t *testing.T) {
 		"registry.ci.openshift.org": {
 			"auth": "cool"
 		},
+		"quay-proxy.ci.openshift.org": {
+			"auth": "supercool"
+		},
+		"quay.io/openshift/ci": {
+			"auth": "supercool"
+		},
+		"qci-pull-through-cache-us-east-1-ci.apps.ci.l2s4.p1.openshiftapps.com": {
+			"auth": "supercool"
+		},
 		"c": {
 			"auth": "bar",
 			"email": "g"
@@ -3157,6 +3166,15 @@ func TestMutateGlobalPullSecret(t *testing.T) {
 		},
 		"registry.ci.openshift.org": {
 			"auth": "expired"
+		},
+		"quay-proxy.ci.openshift.org": {
+			"auth": "supercool"
+		},
+		"quay.io/openshift/ci": {
+			"auth": "supercool"
+		},
+		"qci-pull-through-cache-us-east-1-ci.apps.ci.l2s4.p1.openshiftapps.com": {
+			"auth": "supercool"
 		}
 	}
 }`),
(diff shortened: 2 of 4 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -636,6 +636,55 @@ func mutateGlobalPullSecret(original, secret *coreapi.Secret) (bool, error) {
...
+		return false, fmt.Errorf("failed to parse the original secret: %w", err)
+	}
+	if dockerConfig.Auths[api.DomainForService(api.QuayProxy)].Auth == qciToken {
+		return false, nil
+	}
+	dockerConfig.Auths[api.DomainForService(api.QuayProxy)] = secretbootstrap.DockerAuth{
+		Auth: qciToken,
+	}
+	// token is same as QCI
+	quayRepoDomain := api.DomainForService(api.QuayRepo)
```
deepsm007, line 656: This might conflict with other secret need to confirm

cmd/ci-secret-bootstrap/main.go
```diff
@@ -641,6 +641,29 @@ func mutateGlobalPullSecret(original, secret *coreapi.Secret) (bool, error) {
 		return false, fmt.Errorf("failed to marshal the docker config: %w", err)
 	}
 	original.Data[coreapi.DockerConfigJsonKey] = data
```
smg247, line 643: This ends up getting overwritten below, I assume the goal is to append to it instead?

deepsm007: yeah, that is the ideal scenario to append those secrets.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -541,18 +541,31 @@ func updateSecrets(getters map[string]Getter, secretsMap map[string][]*coreapi.S
...
+				}
+				mutateQuayOpenshiftCI, err := mutateGlobalPullSecret(existingSecret, secret, api.QuayOpenShiftCIRepo)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to mutate secret %s:%s/%s: %w", cluster, secret.Namespace, secret.Name, err))
+				}
+				mutateQCICache, err := mutateGlobalPullSecret(existingSecret, secret, api.QCICache)
+				if err != nil {
+					errs = append(errs, fmt.Errorf("failed to mutate secret %s:%s/%s: %w", cluster, secret.Namespace, secret.Name, err))
+				}
+				if mutated || mutateQCIAPPCIDomain || mutateQuayOpenshiftCI || mutateQCICache {
```
deepsm007, line 560: @smg247 I am skeptical about this statement, other than that everything else should be good

cmd/ci-secret-bootstrap/main.go
```diff
@@ -621,27 +621,35 @@ func mutateGlobalPullSecret(original, secret *coreapi.Secret) (bool, error) {
 	if err != nil {
 		return false, fmt.Errorf("failed to parse the constructed secret: %w", err)
 	}
-	registryDomain := api.DomainForService(api.ServiceRegistry)
-	if dockerConfig.Auths == nil || dockerConfig.Auths[api.DomainForService(api.ServiceRegistry)].Auth == "" {
-		return false, fmt.Errorf("failed to get token for %s", registryDomain)
+	if len(dockerConfig.Auths) == 0 {
```
hongkailiu, line 624: NIT:

```suggestion
	if dockerConfig.Auths == nil {
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -621,27 +621,35 @@ func mutateGlobalPullSecret(original, secret *coreapi.Secret) (bool, error) {
...
+	domains := []string{api.DomainForService(api.ServiceRegistry), api.QCIAPPCIDomain, api.QuayOpenShiftCIRepo, api.QCICache}
+	for _, domain := range domains {
+		if dockerConfig.Auths[domain].Auth == "" {
+			return false, fmt.Errorf("failed to get token for %s", domain)
+		}
+		originalToken := originalDockerConfig.Auths[domain].Auth
+		originalDockerConfig.Auths[domain] = secretbootstrap.DockerAuth{
+			Auth: dockerConfig.Auths[domain].Auth,
+		}
+		if originalToken != originalDockerConfig.Auths[domain].Auth {
```
hongkailiu, line 641: NIT:

```suggestion
		if !mutatedSecret && originalToken != originalDockerConfig.Auths[domain].Auth {
```

cmd/ci-secret-bootstrap/main.go
```diff
@@ -621,27 +621,35 @@ func mutateGlobalPullSecret(original, secret *coreapi.Secret) (bool, error) {
...
+			mutatedSecret = true
+		}
 	}
-	data, err := json.Marshal(dockerConfig)
-	if err != nil {
-		return false, fmt.Errorf("failed to marshal the docker config: %w", err)
+	if mutatedSecret {
+		data, err := json.Marshal(dockerConfig)
```
hongkailiu, line 646:
```suggestion
		data, err := json.Marshal(originalDockerConfig)
```

This is an important one i guess.
The key here is that we would like to keep the change minimal, i.e., we modify the secret ONLY when it is necessary and we modify only the parts that belong to us.

pkg/api/domain.go
```diff
@@ -26,6 +26,8 @@ const (
 	QuayOpenShiftCIRepo = "quay.io/openshift/ci"
 
 	QCIAPPCIDomain = "quay-proxy.ci.openshift.org"
+
+	QCICache = "qci-pull-through-cache-us-east-1-ci.apps.ci.l2s4.p1.openshiftapps.com/openshift/ci"
```
hongkailiu, line 30:
```suggestion
	QCICacheDomain = "qci-pull-through-cache-us-east-1-ci.apps.ci.l2s4.p1.openshiftapps.com"
```

We will modify https://github.com/openshift/release/blob/master/core-services/ci-secret-bootstrap/_config.yaml#L692 accordingly.

I missed it in https://github.com/openshift/release/pull/52732/files (or I cannot remember the reason now)

hongkailiu (commented):
Besides the comments from the PR,
I would like to keep the original unit tests unless you can justify the removal.
If you want to add more, it is totally fine with me.

deepsm007:
> Besides the comments from the PR, I would like to keep the original unit tests unless you can justify the removal. If you want to add more, it is totally fine with me.

I have added few more tests and renamed 1 test

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -3101,168 +3103,509 @@ func TestMutateGlobalPullSecret(t *testing.T) {
...
 			},
-			expectedErr: fmt.Errorf("failed to get token for registry.ci.openshift.org"),
+			expectedErr: fmt.Errorf("failed to get any token"),
 		},
 		{
 			name: "bad original",
 			secret: &coreapi.Secret{
 				Data: map[string][]byte{
 					".dockerconfigjson": []byte(`{
-	"auths": {
```
hongkailiu, line 3111: I do not see the change in this section, except the reformatting?
Could we revert them?

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -3101,168 +3103,509 @@ func TestMutateGlobalPullSecret(t *testing.T) {
...
+								"email": "g"
+							}
+						}
+					}`),
 				},
 			},
 			expectedErr: fmt.Errorf("failed to parse the original secret: failed to get content from nil secret"),
 		},
 		{
-			name: "basic case: expired auth is replaced",
```
hongkailiu, line 3130: Could we keep this case and rename it to "basic case: auth for app.ci is expired auth is replaced"

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -3101,168 +3103,509 @@ func TestMutateGlobalPullSecret(t *testing.T) {
...
+						}
+					}`),
+				},
+			},
+		},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			actual, actualErr := mutateGlobalPullSecret(tc.original, tc.secret)
-			if diff := cmp.Diff(actualErr, tc.expectedErr, testhelper.EquateErrorMessage); diff != "" {
```
hongkailiu, line 3258: Why do we change this section?

deepsm007: when the slices of registries are not sorted the diff provided is different. Using `cmpopts.SortedSlices()` was also an option but assert seemed more convenient

hongkailiu: Do you mind doing it with another PR?
I would like to keep the change minimal so that the review is easier and the chance of screwing thing up is smaller, in my opinion.

cmd/ci-secret-bootstrap/main_test.go
hongkailiu: I do not see this one addresse.

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -3127,142 +3129,457 @@ func TestMutateGlobalPullSecret(t *testing.T) {
 			expectedErr: fmt.Errorf("failed to parse the original secret: failed to get content from nil secret"),
 		},
 		{
-			name: "basic case: expired auth is replaced",
+			name: "not mutated if the auth is still valid",
+			original: &coreapi.Secret{
```
hongkailiu, line 3133: Could we switch the order back?
If you want you may change it with another PR.

deepsm007: I ordered them as they were before, new tests are at the end

hongkailiu: The previous order was secret and then original and now it is the other way around.

hongkailiu:
/lgtm

/hold

Whenever you are ready to watch the rolling AFTER the backup of the secrets is in place.

deepsm007:
/unhold 

Will watching the rolling

PR #4237, opened 2024-07-26
NO-JIRA: Fix JobRuns reference

neisw:
JobRuns table has been removed, need to fix the reference to it.


neisw:
/override ci/prow/e2e

permafailing outside of the pr

PR #4245, opened 2024-08-01
ci-images-mirror: add --validate-config-only option

Prucek:
This PR adds the option to validate the ci-images-mirror config: [core-services/image-mirroring/_config.yaml](https://github.com/openshift/release/blob/master/core-services/image-mirroring/_config.yaml)
Issue: [DPTP-3915](https://issues.redhat.com/browse/DPTP-3915)
/cc @hongkailiu

### DIFF
--- a/cmd/ci-images-mirror/main.go
+++ b/cmd/ci-images-mirror/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/config"
 	quayiociimagesdistributor "github.com/openshift/ci-tools/pkg/controller/quay_io_ci_images_distributor"
 	"github.com/openshift/ci-tools/pkg/load/agents"
+	"github.com/openshift/ci-tools/pkg/steps/release"
 	"github.com/openshift/ci-tools/pkg/util"
 )
 
@@ -57,6 +58,7 @@ type options struct {
 	config                           *quayiociimagesdistributor.CIImagesMirrorConfig
 	configMutex                      sync.Mutex
 	configLastModifiedAt             time.Time
+	validateConfigOnly               bool
 }
 
 type quayIOCIImagesDistributorOptions struct {
@@ -84,6 +86,7 @@ func newOpts() *options {
 	fs.IntVar(&opts.port, "port", 8090, "Port to run the server on")
 	fs.DurationVar(&opts.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
 	fs.BoolVar(&opts.onlyValidManifestV2Images, "only-valid-manifest-v2-images", true, "If set, source images with invalidate manifests of v2 will not be mirrored")
+	fs.BoolVar(&opts.validateConfigOnly, "validate-config-only", false, "If set, only validate the config file and exit")
 	if err := fs.Parse(os.Args[1:]); err != nil {
 		logrus.WithError(err).Fatal("could not parse args")
 	}
@@ -102,7 +105,7 @@ func (o *options) loadConfig(caller string) error {
 		return nil
 	}
 	logrus.WithField("caller", caller).Info("Loading config ...")
-	c, err := quayiociimagesdistributor.LoadConfig(o.configFile)
+	c, err := quayiociimagesdistributor.LoadConfigFromFile(o.configFile)
 	if err != nil {
 		return fmt.Errorf("failed to load config file %s: %w", o.configFile, err)
 	}
@@ -113,7 +116,7 @@ func (o *options) loadConfig(caller string) error {
 
 func (o *options) validate() error {
 	var errs []error
-	if o.leaderElectionNamespace == "" {
+	if o.leaderElectionNamespace == "" && !o.validateConfigOnly {
 		errs = append(errs, errors.New("--leader-election-namespace must be set"))
 	}
 	if o.leaderElectionSuffix != "" && !o.dryRun {
@@ -151,6 +154,10 @@ type supplementalCIImagesServiceWithMirrorStore struct {
 	logger      *logrus.Entry
 }
 
+func targetToQuayImage(target string) string {
+	return fmt.Sprintf("%s:%s", api.QuayOpenShiftCIRepo, strings.Replace(strings.Replace(target, "/", "_", 1), ":", "_", 1))
+}
+
 func (s *supplementalCIImagesServiceWithMirrorStore) Mirror(m map[string]quayiociimagesdistributor.Source) error {
 	s.logger.Info("Mirroring supplemental CI images ...")
 	for k, v := range m {
@@ -160,7 +167,7 @@ func (s *supplementalCIImagesServiceWithMirrorStore) Mirror(m map[string]quayioc
 		}
 		if err := s.mirrorStore.Put(quayiociimagesdistributor.MirrorTask{
 			Source:      source,
-			Destination: fmt.Sprintf("%s:%s", api.QuayOpenShiftCIRepo, strings.Replace(strings.Replace(k, "/", "_", 1), ":", "_", 1)),
+			Destination: targetToQuayImage(k),
 			Owner:       "supplementalCIImagesService",
 		}); err != nil {
 			return fmt.Errorf("failed to put mirror task: %w", err)
@@ -181,6 +188,16 @@ func main() {
 	opts := newOpts()
 	if err := opts.validate(); err != nil {
 		logrus.WithError(err).Fatal("Failed to validate options")
+	}
+
+	ciOperatorConfigPath := filepath.Join(opts.releaseRepoGitSyncPath, config.CiopConfigInRepoPath)
+	if opts.validateConfigOnly {
+		configFilePathInReleaseRepo := strings.Replace(opts.configFile, opts.releaseRepoGitSyncPath, "", 1)
+		err := validateConfig(opts.config, opts.registryConfig, ciOperatorConfigPath, configFilePathInReleaseRepo)
+		if err != nil {
+			logrus.WithError(err).Fatal("Failed to validate config")
+		}
+		os.Exit(0)
 	}
 
 	ctx := controllerruntime.SetupSignalHandler()
@@ -276,7 +293,6 @@ func main() {
 		configAgentOption := func(opt *agents.ConfigAgentOptions) {
 			opt.UniversalSymlinkWatcher = universalSymlinkWatcher
 		}
-		ciOperatorConfigPath := filepath.Join(opts.releaseRepoGitSyncPath, config.CiopConfigInRepoPath)
 
 		ciOPConfigAgent, err := agents.NewConfigAgent(ciOperatorConfigPath, errCh, configAgentOption)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/ci-images-mirror/main.go
```diff
@@ -83,6 +87,8 @@ func newOpts() *options {
 	fs.IntVar(&opts.port, "port", 8090, "Port to run the server on")
 	fs.DurationVar(&opts.gracePeriod, "gracePeriod", time.Second*10, "Grace period for server shutdown")
 	fs.BoolVar(&opts.onlyValidManifestV2Images, "only-valid-manifest-v2-images", true, "If set, source images with invalidate manifests of v2 will not be mirrored")
+	fs.BoolVar(&opts.validateConfigOnly, "validate-config-only", false, "If set, only validate the config file and exit")
+	fs.StringVar(&opts.ciOperatorconfigPath, "ci-operator-config-path", "", "Path to the ci operator config")
```
hongkailiu, line 91: This is not really needed.
It is calculated the code already.
https://github.com/openshift/ci-tools/blob/1420774f51896ba02c21ce2c6dab35f39fa8db89/cmd/ci-images-mirror/main.go#L277

cmd/ci-images-mirror/main.go
```diff
@@ -181,6 +190,15 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to validate options")
 	}
 
+	if opts.validateConfigOnly {
+		err := validateConfig(opts.config, opts.registryConfig, opts.ciOperatorconfigPath)
+		if err != nil {
+			logrus.WithError(err).Fatal("Failed to validate config")
+			os.Exit(1)
```
hongkailiu, line 197:
```suggestion
			
```

cmd/ci-images-mirror/main.go
```diff
@@ -320,6 +338,69 @@ func main() {
 	logrus.Info("Process ended gracefully")
 }
 
+func validateConfig(config *quayiociimagesdistributor.CIImagesMirrorConfig, registryConfig string, ciOperatorconfigPath string) error {
+	r, err := regexp.Compile("docker.io")
```
hongkailiu, line 342: Wouldnt something like this do the job?

```go
strings.HasPrefix(image, "docker.io/")
```

cmd/ci-images-mirror/main.go
```diff
@@ -320,6 +338,69 @@ func main() {
 	logrus.Info("Process ended gracefully")
 }
 
+func validateConfig(config *quayiociimagesdistributor.CIImagesMirrorConfig, registryConfig string, ciOperatorconfigPath string) error {
```
hongkailiu, line 341: I would collect all the errors and then return instead of return right after the first error.

cmd/ci-images-mirror/main.go
```diff
@@ -320,6 +338,69 @@ func main() {
...
+			if sourceISTagRef == isTagRef.ISTagName() {
+				return fmt.Errorf("source image %s is promoted", sourceISTagRef)
+			}
+		}
+		return nil
+	})
+	return err != nil
+}
+
+func isDuplicated(nsWithName string, config *quayiociimagesdistributor.CIImagesMirrorConfig) bool {
```
hongkailiu, line 381: I believe you could not even do this today?
IIRC, we have a yaml lint to avoid this.
Could you please make a PR to the release repo and see if any failures show up?

cmd/ci-images-mirror/main.go
```diff
@@ -320,6 +338,69 @@ func main() {
...
+	var name_count int
+	for name := range config.SupplementalCIImages {
+		if name == nsWithName {
+			name_count++
+		}
+	}
+	return name_count > 1
+}
+
+func isAccessible(image string, registryConfig string) bool {
```
hongkailiu, line 391: This would check all the images in the config.
Ideally, we check only newly added images.

cmd/ci-images-mirror/main.go
hongkailiu: "A new tag is added to override an existing tag."
By this I mean:
We check if the target exists in QCI for a new tag.
So you need to know somehow the list of new tags coming with a PR.

To achieve that, we may parse the config file from the PR and the config file in the current release repo and see the difference.

cmd/ci-images-mirror/main.go
```diff
@@ -183,6 +192,16 @@ func main() {
 		logrus.WithError(err).Fatal("Failed to validate options")
 	}
 
+	if opts.validateConfigOnly {
+		ciOperatorConfigPath := filepath.Join(opts.releaseRepoGitSyncPath, config.CiopConfigInRepoPath)
```
hongkailiu, line 196: NIT:
We could defined it only once for 
https://github.com/openshift/ci-tools/blob/fbee3dfe3b0757b98cd2d6f49838621e3e69f757/cmd/ci-images-mirror/main.go#L279 as well.

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +341,89 @@ func main() {
...
+	diff := []string{}
+	for k := range config.SupplementalCIImages {
+		if _, ok := masterConfig.SupplementalCIImages[k]; !ok {
+			diff = append(diff, k)
+		}
+	}
+	return diff, nil
+}
+
+func isPromoted(sourceISTagRef string, ciOperatorConfigPath string) bool {
```
hongkailiu, line 396: I would collect the promoted tags once and reuse them for all new tags.

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +341,89 @@ func main() {
...
+			errs = append(errs, fmt.Errorf("target %s would overwrite a promoted tag", target))
+		}
+		if isAccessible(targetToQuayImage(target), registryConfig) {
+			errs = append(errs, fmt.Errorf("target %s already exists in quay", target))
+		}
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func getConfigDiff(configFilePathInReleaseRepo string, config *quayiociimagesdistributor.CIImagesMirrorConfig) ([]string, error) {
```
hongkailiu, line 370: I would do it this way: Add a new function `quayiociimagesdistributor.LoadConfigFromReleaseRepo()` to do the job where we use the existing GH client API to get the content of the config file. Moreover, I would imagine that we could take some sharing code from `quayiociimagesdistributor.LoadConfig(o.configFile)`.

Prucek: This was a great idea! I didn't know such a thing existed.

pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
```diff
@@ -374,6 +374,20 @@ func sourceForConfigChangeChannel(registryClient ctrlruntimeclient.Client, chang
...
+func LoadConfigFromReleaseRepo(configPath string) ([]byte, error) {
+	metadata := cioperatorapi.Metadata{
+		Org:    "openshift",
+		Repo:   "release",
+		Branch: "master",
+	}
+	fileGetter := repoFileGetterWithIgnore(metadata)
+	file, err := fileGetter(metadata.Org, metadata.Repo, metadata.Branch)(configPath)
+	if err != nil {
+		return []byte{}, err
```
hongkailiu, line 386:
```suggestion
		return []byte{}, fmt.Errorf("failed to get the file %s in repo openshift/release for branch master: %w", configPath, err)
```

pkg/controller/quay_io_ci_images_distributor/quay_io_ci_images_distributor.go
```diff
@@ -374,6 +374,20 @@ func sourceForConfigChangeChannel(registryClient ctrlruntimeclient.Client, chang
 	return channelSource
 }
 
+func LoadConfigFromReleaseRepo(configPath string) ([]byte, error) {
+	metadata := cioperatorapi.Metadata{
```
hongkailiu, line 378: We do not need this variable.

cmd/ci-images-mirror/main.go
```diff
@@ -102,7 +106,11 @@ func (o *options) loadConfig(caller string) error {
 		return nil
 	}
 	logrus.WithField("caller", caller).Info("Loading config ...")
-	c, err := quayiociimagesdistributor.LoadConfig(o.configFile)
+	bytes, err := gzip.ReadFileMaybeGZIP(o.configFile)
```
hongkailiu, line 109: NIT: I would define `quayiociimagesdistributor.LoadConfigFromFile(file string)` and `LoadConfig(bytes)` might become private.

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +343,94 @@ func main() {
...
+	bytes, err := quayiociimagesdistributor.LoadConfigFromReleaseRepo(configFilePathInReleaseRepo)
+	if err != nil {
+		return []string{}, fmt.Errorf("failed to load config from release repo: %w", err)
+	}
+	masterConfig, err := quayiociimagesdistributor.LoadConfig(bytes)
+	if err != nil {
+		return []string{}, fmt.Errorf("failed to load config file %s: %w", configFilePathInReleaseRepo, err)
+	}
+
+	diff := []string{}
```
hongkailiu, line 382: I would make return type a set of strings which is more friendly with unit testing.

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +343,94 @@ func main() {
...
+			errs = append(errs, fmt.Errorf("target %s would overwrite a promoted tag", target))
+		}
+		if isAccessible(targetToQuayImage(target), registryConfig) {
+			errs = append(errs, fmt.Errorf("target %s already exists in quay", target))
+		}
+	}
+	return utilerrors.NewAggregate(errs)
+}
+
+func getConfigDiff(configFilePathInReleaseRepo string, config *quayiociimagesdistributor.CIImagesMirrorConfig) ([]string, error) {
```
hongkailiu, line 372:
```suggestion
func getNewSupplementalCIImageTargets(configFilePathInReleaseRepo string, config *quayiociimagesdistributor.CIImagesMirrorConfig) ([]string, error) {
```

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +343,94 @@ func main() {
...
+	promotedTags := getPromotedTags(ciOperatorConfigPath)
+	for _, promotedTag := range promotedTags {
+		if sourceISTagRef == promotedTag {
+			return true
+		}
+	}
+	return false
+}
+
+func getPromotedTags(ciOperatorConfigPath string) []string {
```
hongkailiu, line 401: Return a set of strings.

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +343,94 @@ func main() {
...
+	for k := range config.SupplementalCIImages {
+		if _, ok := masterConfig.SupplementalCIImages[k]; !ok {
+			diff = append(diff, k)
+		}
+	}
+	return diff, nil
+}
+
+func isPromoted(sourceISTagRef string, ciOperatorConfigPath string) bool {
+	promotedTags := getPromotedTags(ciOperatorConfigPath)
```
hongkailiu, line 392: `getPromotedTags` should be done long before `isPromoted`.

cmd/ci-images-mirror/main.go
```diff
@@ -322,6 +338,94 @@ func main() {
...
+		source := config.SupplementalCIImages[target]
+		if source.Image != "" {
+			if strings.HasPrefix(source.Image, "docker.io/") {
+				errs = append(errs, fmt.Errorf("source image %s is from docker.io", source.Image))
+			}
+			if !isAccessible(source.Image, registryConfig) {
+				errs = append(errs, fmt.Errorf("source image %s is not accessible", source.Image))
+			}
+		}
+		if isPromoted(target, promotedTags) {
```
hongkailiu, line 358: Would `promotedTags.has(target)` do the job?

Prucek: Professional deformation 🤦🏼

PR #4257, opened 2024-08-09
NO-JIRA: feat: update html for aggregated results

eggfoobar:
changed to using go template for the html
updated ui to help with glancable information

Ex:

image

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/aggregation-testrun-summary.gohtml
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/aggregation-testrun-summary.gohtml
@@ -1,0 +1,234 @@
+{{define "printTable"}}
+    <table>
+            <tr>
+                <th>Name</th>
+                <th>Job Runs</th>
+            </tr>
+        {{ range . }}
+        {{ $jobRuns := .JobRuns }}
+            {{ $formatedSummary := (.Summary | formatSummary)}}
+            {{ $failedQuantileData := ( "" | parseQuantileValues) }}
+            {{with $formatedSummary.failures}}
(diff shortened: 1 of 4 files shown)


stbenjam:
What do the purple squares represent next to the passed tests? We usually reserve purple for flakes

eggfoobar:
> What do the purple squares represent next to the passed tests? We usually reserve purple for flakes

Yup, thats exactly what they represent, so it's being driven off of the data using the same color hex values as testgrid to keep it familiar. 

I only focused on the UI here, so I kept the logic that generates the data the same, that is currently setting them as flakes. I just looked at the other aggregated jobs and it seems we label them all as flakes. Maybe we should revisit why we consider them flakes? 

This was off of a recent aggregated job, https://prow.ci.openshift.org/view/gs/test-platform-results/logs/aggregated-aws-ovn-upgrade-4.17-micro-release-openshift-release-analysis-aggregator/1821748494696714240. 

This logic is what considers them flakes, https://github.com/openshift/ci-tools/pull/4257/files#diff-6bfd3d2b536f5dad8a8f79cc26eba771afb36a131b8b83222393cc0fc6715d53R246

neisw:
I ran it through https://prow.ci.openshift.org/view/gs/test-platform-results/logs/aggregated-aws-ovn-single-node-upgrade-4.17-micro-release-openshift-release-analysis-aggregator/1820723738107187200 for comparison.  Need to remove the .txt to view locally and compare 

[aggregation-testrun-summary.html.txt](https://github.com/user-attachments/files/16564847/aggregation-testrun-summary.html.txt)

eggfoobar:
Oh nice, thanks @neisw !

neisw:
lgtm

I'll give Stephen a chance to tag it / add any followup comments

PR #4259, opened 2024-08-09
DPTP-4138: Step-registry supports arm64 nodeSelector

deepsm007:
This PR adds the functionality of `node_architecture` for individual steps, Tests, workflows in the step-registry. This leverages the use of multiple arch nodes available in build farm. 

`workflow > Test > steps`

If `node_architecture` field is specified at workflow level it allows all the steps in the workflow to run on the same node arch. The default is set to `amd64`.

**Note: This feature is currently supported on build10 cluster and we will introduce it to other clusters in the future**


pkg/api/types.go
```diff
@@ -1014,6 +1014,9 @@ type LiteralTestStep struct {
 	// RunAsScript defines if this step should be executed as a script mounted
 	// in the test container instead of being executed directly via bash
 	RunAsScript *bool `json:"run_as_script,omitempty"`
+	// NodeArchitecture is the architecture for the node where the test will run.
+	// If set, the generated test pod will include a nodeSelector for this architecture.
+	NodeArchitecture NodeArchitecture `json:"node_architecture,omitempty"`
```
smg247, line 1019: It looks like most of the fields here are pointers, is there a reason why you chose not to use a pointer here?

pkg/steps/multi_stage/gen.go
```diff
@@ -163,6 +163,13 @@ func (s *multiStageTestStep) generatePods(
 				pod.Spec.DNSPolicy = coreapi.DNSNone
 			}
 		}
+		if step.NodeArchitecture != "" {
+			if pod.Spec.NodeSelector == nil {
+				// if nodeArchitecture is not set, default to amd64 node selector
+				pod.Spec.NodeSelector = map[string]string{"kubernetes.io/arch": string(api.NodeArchitectureAMD64)}
+			}
```
smg247, line 170: What is the point of this block? It looks like it is always replaced on line 171

pkg/validation/test.go
```diff
@@ -879,6 +879,14 @@ func validateDNSConfig(fieldRoot string, dnsConfig []api.StepDNSConfig) (ret []e
 	return errs
 }
 
+func validateNodeArchitecture(fieldRoot string, nodeArchitecture api.NodeArchitecture) (ret []error) {
```
smg247, line 882: It looks like this will only ever return, at most, one error. Can the function sig reflect that?

pkg/validation/test.go
```diff
@@ -879,6 +879,14 @@ func validateDNSConfig(fieldRoot string, dnsConfig []api.StepDNSConfig) (ret []e
...
+	var errs []error
+	if nodeArchitecture != api.NodeArchitectureAMD64 && nodeArchitecture != api.NodeArchitectureARM64 {
+		errs = append(errs, fmt.Errorf("%s.nodeArchitecture cannot be empty", fieldRoot))
```
smg247, line 885: Is this error message accurate? You aren't checking for emptiness, but rather one of two possible values.

pkg/api/types.go
deepsm007: `NodeArchitecture` was defined on line 782 and I chose to use as it is.

pkg/steps/multi_stage/gen.go
deepsm007: missed the else 🤦🏻

pkg/api/types.go
deepsm007: @smg247 do you want me to make it a pointer?

smg247: yeah, I think it would be simpler if you did:
```suggestion
	NodeArchitecture *NodeArchitecture `json:"node_architecture,omitempty"`
```

pkg/validation/test.go
```diff
@@ -879,6 +879,14 @@ func validateDNSConfig(fieldRoot string, dnsConfig []api.StepDNSConfig) (ret []e
 	return errs
 }
 
+func validateNodeArchitecture(fieldRoot string, nodeArchitecture api.NodeArchitecture) error {
+	var errs error
+	if nodeArchitecture != api.NodeArchitectureAMD64 && nodeArchitecture != api.NodeArchitectureARM64 {
+		errs = fmt.Errorf("%s.nodeArchitecture expected one of %v or %v", fieldRoot, api.NodeArchitectureAMD64, api.NodeArchitectureARM64)
+	}
+	return errs
```
smg247, line 887:
```suggestion
	if nodeArchitecture != api.NodeArchitectureAMD64 && nodeArchitecture != api.NodeArchitectureARM64 {
		return fmt.Errorf("%s.nodeArchitecture expected one of %v or %v", fieldRoot, api.NodeArchitectureAMD64, api.NodeArchitectureARM64)
	}
	return nil
```

pkg/validation/test.go
```diff
@@ -879,6 +879,13 @@ func validateDNSConfig(fieldRoot string, dnsConfig []api.StepDNSConfig) (ret []e
 	return errs
 }
 
+func validateNodeArchitecture(fieldRoot string, nodeArchitecture api.NodeArchitecture) error {
+	if nodeArchitecture != api.NodeArchitectureAMD64 && nodeArchitecture != api.NodeArchitectureARM64 {
+		return fmt.Errorf("%s.nodeArchitecture expected one of %v or %v", fieldRoot, api.NodeArchitectureAMD64, api.NodeArchitectureARM64)
```
droslean, line 884: But we are the ones that give value to the empty nodeSelectors. Can you elaborate more on this validation?

deepsm007: This validation is to avoid someone accidentally using different or wrong architecture

test/integration/ci-operator-configresolver/expected/openshift-installer-release-4.2.json
```diff
@@ -112,6 +112,23 @@
               }
             }
           }
+        ],
+        "observers": [
```
droslean, line 116: I don't see how this is related to the PRs changes...

pkg/steps/multi_stage/gen.go
```diff
@@ -163,6 +163,14 @@ func (s *multiStageTestStep) generatePods(
 				pod.Spec.DNSPolicy = coreapi.DNSNone
 			}
 		}
+		if step.NodeArchitecture != nil {
+			if pod.Spec.NodeSelector == nil {
```
droslean, line 167: This seems smelly. How the NodeSelector can be nil or not? We are creating the pod. It seems that we are using the pod from [steps/pod.go#L248-L250](https://github.com/deepsm007/ci-tools/blob/c963d9b5984fbdfb2652e67d25ce336c0f4535c9/pkg/steps/pod.go#L248-L250).

droslean: You need to handle the nodeSelector in https://github.com/openshift/ci-tools/blob/master/pkg/steps/multi_stage/gen.go#L137-L139 through the generator options.

deepsm007: Addressed in the commit

pkg/validation/test.go
```diff
@@ -879,6 +885,13 @@ func validateDNSConfig(fieldRoot string, dnsConfig []api.StepDNSConfig) (ret []e
 	return errs
 }
 
+func validateNodeArchitecture(fieldRoot string, nodeArchitecture api.NodeArchitecture) (ret []error) {
+	if nodeArchitecture != api.NodeArchitectureAMD64 && nodeArchitecture != api.NodeArchitectureARM64 {
+		ret = append(ret, fmt.Errorf("%s.nodeArchitecture expected one of %v or %v", fieldRoot, api.NodeArchitectureAMD64, api.NodeArchitectureARM64))
```
droslean, line 890: Why do we need to return a list of errors?

PR #4260, opened 2024-08-09
`pod-scaler`: update consumers and e2e tests to use v2 data

smg247:
After some time of v2 data gathering, it will be time to utilize that data with the consumers. This is a fairly simple change as we just need to swap out `v1` for `v2` in many places. The only significant difference here is accessing the `fingerprint` within the `fingerprintTime` struct. Once we confirm this is working correctly, all `v1` logic will be removed entirely and the data will be purged. 

For: https://issues.redhat.com/browse/DPTP-4069

/hold to merge when we have sufficient v2 data available


smg247:
Note that this will require a coordinated update to point the deployments to the v2 GSC bucket

deepsm007 (approved):

PR #4262, opened 2024-08-13
jobdisruption: exclude internal-lb

vrutkovs:
There is no way for us to differentiate between "real" internal LB in the cloud or fake one in SNO, so this disruption is not being alerted upon.

Ref: https://github.com/openshift/origin/pull/28982


neisw:
@dgoodwin was going to look at excluding SNO from aggregated disruption.  If we do that would we still want/need this?

vrutkovs:
Yes, we'd always have "internal-lb" disruption on SNO due to having one node only, so it would be equivalent to localhost disruption

PR #4263, opened 2024-08-13
TRT-1770: Disable aggregated disruption testing on single-node

dgoodwin:
- **Add function to list variants for a specific job**
- **Manually update mock ci data client**
- **Fix hack/update-mocks.sh which does not like kube sets.Set**
- **Fix job variants query**
- **Skip aggregated disruption testing on single-node**

Single node routinely experiences 10 minutes of disruption, our tolerances in testing are not applicable here and thus these jobs can fail all the time. Turn off aggregated disruption testing for the new single node aggregated job, we'll lean on our alerts and dashboard review instead.


dgoodwin:
/close

Seems jammed.

PR #4268, opened 2024-08-15
Fix validation for ocp-build-data from.builder.image

deepsm007:
/cc @openshift/test-platform 

`image configs from ocp-build-data: error validating images/okd-only-upi-installer.yml: [.from.5 failed validation: both stream and member were unset, .from.6 failed validation: both stream and member were unset, .from.7 failed validation: both stream and member were unset, .from.8 failed validation: both stream and member were unset]"
`

### DIFF
--- a/pkg/api/ocpbuilddata/types.go
+++ b/pkg/api/ocpbuilddata/types.go
@@ -17,6 +17,7 @@ import (
 )
 
 type OCPImageConfig struct {
+	Mode           *Mode                  `json:"mode"`
 	Content        *OCPImageConfigContent `json:"content"`
 	From           OCPImageConfigFrom     `json:"from"`
 	Push           OCPImageConfigPush     `json:"push"`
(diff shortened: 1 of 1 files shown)


pkg/api/ocpbuilddata/types.go
```diff
@@ -85,14 +85,15 @@ type OCPImageConfigFrom struct {
...
 
 func (icfs OCPImageConfigFromStream) validate() error {
-	if icfs.Stream == "" && icfs.Member == "" {
-		return errors.New("both stream and member were unset")
+	if icfs.Stream == "" && icfs.Member == "" && icfs.Image == "" {
+		return errors.New("all stream, member and image were unset")
 	}
-	if icfs.Stream != "" && icfs.Member != "" {
-		return fmt.Errorf("both stream(%s) and member(%s) were set", icfs.Stream, icfs.Member)
+	if icfs.Stream != "" && icfs.Member != "" && icfs.Image != "" {
```
smg247, line 95: This means it is okay to have 2 of the 3 set. Are we sure that it isn't necessary to validate that exactly 1 is set?

PR #4270, opened 2024-08-19
dispatcher: changing rest req name to job

jmguzik:


jmguzik:
/override ci/prow/e2e 
no need to e2e test for that tiny change

PR #4275, opened 2024-08-20
Dispatcher: various small fixes after testing on cluster + sanitizer to pkg

jmguzik:


cmd/prow-job-dispatcher/main.go
```diff
@@ -754,5 +785,23 @@ func main() {
 		logrus.Info("Finished dispatching and create no PR, exiting ...")
 		os.Exit(0)
 	}
-	createPR(o)
+	logrus.WithField("targetDir", o.targetDir).Info("Changing working directory ...")
```
jmguzik, line 788: This part about PR creation is re-added. It was here before. Dispatcher as daemon diverged too much in PR creation and I had to re-introduce the old way for non-daemonized run.

cmd/prow-job-dispatcher/main.go
```diff
@@ -656,29 +685,31 @@ func runAsDaemon(o options, promVolumes *prometheusVolumes) {
 	}
 	defer watcher.Close()
 
-	go func() {
+	go func(config string) {
```
jmguzik, line 688: Unfortunately, watchers do not work good with git-sync, they completely do not catch events as git sync is removing and re-adding everything. Simple loop with 30 secs delay should handle this.

cmd/prow-job-dispatcher/main.go
```diff
@@ -529,25 +529,59 @@ func clustersMapToSet(clusterMap ClusterMap) sets.Set[string] {
 	return clusterSet
 }
 
-func createPR(o options) {
+func gitCloneRelease() error {
+	cmd := exec.Command("git", "clone", "--depth", "1", "--single-branch", "git@github.com:openshift/release.git")
```
smg247, line 533: We could probably use `prow's` `gitClient` here to do this, but if this would be the only usage, I think this is fine also

cmd/prow-job-dispatcher/main.go
```diff
@@ -529,25 +529,59 @@ func clustersMapToSet(clusterMap ClusterMap) sets.Set[string] {
...
+	err := os.RemoveAll(directory)
+	if err != nil {
+		logrus.WithField("directory", directory).WithError(err).Error("failed to remove directory")
+	}
+	logrus.WithField("directory", directory).Info("Successfully removed directory")
+}
+
+func createPR(o options, config *dispatcher.Config) {
+	targetDirWithRelease := filepath.Join(o.targetDir, "/release")
+	cleanup(targetDirWithRelease)
```
smg247, line 553: Did you mean to call this right away? I would have expected it only in the next line

cmd/prow-job-dispatcher/main.go
```diff
@@ -529,25 +529,59 @@ func clustersMapToSet(clusterMap ClusterMap) sets.Set[string] {
...
 	command := "/usr/bin/sanitize-prow-jobs"
-	arguments := []string{"--prow-jobs-dir", "./ci-operator/jobs", "--config-path", "./core-services/sanitize-prow-jobs/_config.yaml"}
+	arguments := []string{"--prow-jobs-dir", "./release/ci-operator/jobs", "--config-path", "./release/core-services/sanitize-prow-jobs/_config.yaml"}
 	fullCommand := fmt.Sprintf("%s %s", filepath.Base(command), strings.Join(arguments, " "))
 	logrus.WithField("fullCommand", fullCommand).Infof("Running the command ...")
 
 	combinedOutput, err := exec.Command(command, arguments...).CombinedOutput()
 	if err != nil {
-		logrus.WithError(err).WithField("combinedOutput", string(combinedOutput)).Fatal("failed to run the command")
+		logrus.WithError(err).WithField("combinedOutput", string(combinedOutput)).Error("failed to run the command")
```
smg247, line 579: Why switch from Fatal? Is the PR creation still valid when this fails?

cmd/prow-job-dispatcher/main.go
smg247: This comment is probably worthy of being immortalized in the code itself :)

cmd/prow-job-dispatcher/main.go
jmguzik: only and actually working pretty well, so I would keep it

cmd/prow-job-dispatcher/main.go
jmguzik: if garbare from other times/restarts etc

cmd/prow-job-dispatcher/main.go
jmguzik: in this place it should be followed by return, good catch

jmguzik:
This will go after https://github.com/openshift/ci-tools/pull/4283

PR #4277, opened 2024-08-20
cluster-init: adapt to use cobra commands

danilo-gemoli:
`cluster-init` tool is going to undergo a series of changes that will allow us to provision, onboard and deprovision a cluster.
In order to achieve that I'm planning to add several new commands, at least two for now:
```sh
$ cluster-init provision ...
$ cluster-init onboard ...
```
there may be another one `deprovision`, but is not as critical as the previous ones.

This PR is about a refactoring to introduce the [spf13/cobra](https://github.com/spf13/cobra) framework as it makes the management of commands and flags very convenient.

Notice also that we are introducing a breaking change as [spf13/cobra](https://github.com/spf13/cobra) conceives posix compliant only flags, in contrast to the default Golang behaviour.
That means a "long argument" with a single leading dash is not allowed anymore:
```
$ cluster-init -release-repo=/src/openshift/release
``` 
and it mush be fixed as follows:
```
$ cluster-init --release-repo=/src/openshift/release
```


jmguzik (approved):

jmguzik:
@danilo-gemoli you have the failure here

PR #4289, opened 2024-08-26
Changes to reflect inproper modification info while doing shallow git-sync

jmguzik:


cmd/prow-job-dispatcher/main.go
```diff
@@ -718,32 +719,32 @@ func runAsDaemon(o options, promVolumes *prometheusVolumes) {
 	}
 	c.Start()
 
-	// instead using fsnotify watcher falling back to periodic checks as
```
jmguzik, line 721: unfortunately with git-sync and depth 1 master info is lost and fallback to new checkout time which is inconvenient

PR #4307, opened 2024-09-04
ci-operator: treat bundle images as non multi-arch without manifest list

droslean:
https://issues.redhat.com/browse/DPTP-4143

Bundle images by design are not multi-arch. In the meantime, there is currently a supported way to allow users to enable multi-arch builds for bundles. So ci-operator was building the image with a manifest-list for amd64 only. That caused a problem using the image on arm nodes since the desired manifest from the list was missing. Also in the meantime, the bundle doesn't care about any architecture... 

This change will use the old building way. Ci-operator will just create a single build forced on amd64 node for consistency and the output image won't hold a manifest list.

/cc @deepsm007 @openshift/test-platform @aleskandro 

### DIFF
--- a/pkg/steps/bundle_source.go
+++ b/pkg/steps/bundle_source.go
@@ -26,7 +26,6 @@ type bundleSourceStep struct {
 	podClient          kubernetes.PodClient
 	jobSpec            *api.JobSpec
 	pullSecret         *coreapi.Secret
-	multiArch          bool
 }
 
 func (s *bundleSourceStep) Inputs() (api.InputDefinition, error) {
@@ -80,7 +79,10 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
 		"",
 	)
 
-	return handleBuilds(ctx, s.client, s.podClient, *build, newImageBuildOptions(s.multiArch))
+	// Bundle images are not multi-arch by design. Here we build it without creating a manifest-listed image.
+	// Note that we are not configuring a node selector here, so the build will be scheduled on any available
+	// node no matter the architecture.
+	return handleBuild(ctx, s.client, s.podClient, *build)
 }
 
 func replaceCommand(pullSpec, with string) string {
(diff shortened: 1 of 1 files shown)


pkg/steps/bundle_source.go
```diff
@@ -80,7 +79,13 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
 		"",
 	)
 
-	return handleBuilds(ctx, s.client, s.podClient, *build, newImageBuildOptions(s.multiArch))
+	// It doesn't matter where the bundle image will be built, but to keep the consistency
+	// let's force it to run on an amd64 node.
+	build.Spec.NodeSelector = map[string]string{"kubernetes.io/arch": "amd64"}
```
aleskandro, line 84: Removing this at all would allow to schedule on arm64 whenever arm64 nodes are available and also avoid amd64 nodes scale out when arm64 resources are available, but amd64 are not.

Why don't we just set it to build on amd64? What's the 'consistency' we are trying to guarantee here?

If it's about avoiding errors we are not aware of, should we consider a story to remove this node selector at one point?

aleskandro:
/hold

(just to wait for a response on the above comment)

pkg/steps/bundle_source.go
droslean: I would prefer forcing an amd64 build as a default behaviour to keep the consistency and allow the user to override it by specifying something like `allowAnyArchitecture: true` in the configuration.

aleskandro: I would refrain from exposing this specificity to the users, especially in a case like this, where it is unneeded. Users are not even meant to know about this specific topic and also struggle to get it if they don't know all the nuances of multi-arch/manifest-list images, multi-arch clusters, etc... 

We also discussed this PR and the broader issue of 'default node selectors' in CI a few minutes ago with other folks during the multiarch office hour. 

We could add this topic to the agenda of the next sync meeting on Monday. Wdyt?

cc @tvardema @jaypoulz @jeffdyoung

droslean: For now, I removed the amd64 defaulting and just left a comment for future observations if needed.

pkg/steps/bundle_source.go
```diff
@@ -80,7 +79,11 @@ func (s *bundleSourceStep) run(ctx context.Context) error {
 		"",
 	)
 
-	return handleBuilds(ctx, s.client, s.podClient, *build, newImageBuildOptions(s.multiArch))
+	// Bundle images are not multi-arch by design. Here we build it without creating a manifest-listed image.
+	// The output image will be with no manifest list and it will be build on an amd64 node.
```
smg247, line 83: This line is outdated now, right?

smg247 (commented):
/lgtm

pkg/steps/bundle_source.go
aleskandro: Thanks @droslean 

/lgtm

aleskandro:
Did this fix land on the build10 cluster already?

PR #4308, opened 2024-09-06
Clean-up dispatcher to run only as daemon & fix sanitizer inconsistency with dispatcher

jmguzik:
sanitizer will now respect blocked clusters the same way dispatcher does


jmguzik:
Sorry, that was supposed to be 2 PRs but mixed together due to my mistake

jmguzik:
first part was mostly removal though, which I guess is nice :)

jmguzik:
/hold
I would like to be here when this merges to observe everything

PR #4318, opened 2024-09-12
Add new cluster profile aro-redhat-tenant

slintes:
new cluster profile for the ARO (managed Openshift on Azure) team

also see https://github.com/openshift/release/pull/56644

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1355,6 +1355,7 @@ const (
 	ClusterProfileAWSStackrox           ClusterProfile = "aws-stackrox"
 	ClusterProfileAWSSDCICD             ClusterProfile = "aws-sd-cicd"
 	ClusterProfileGCPSDCICD             ClusterProfile = "gcp-sd-cicd"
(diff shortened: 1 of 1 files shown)


pkg/api/types.go
```diff
@@ -1355,6 +1355,7 @@ const (
 	ClusterProfileAWSStackrox           ClusterProfile = "aws-stackrox"
 	ClusterProfileAWSSDCICD             ClusterProfile = "aws-sd-cicd"
 	ClusterProfileGCPSDCICD             ClusterProfile = "gcp-sd-cicd"
+	ClusterProfileAroRH                                = "aro-redhat-tenant"
```
smg247, line 1358:
```suggestion
	ClusterProfileAroRH                   ClusterProfile = "aro-redhat-tenant"
```

smg247: You'll have to format that properly as well

slintes: oh, thx, done!

PR #4328, opened 2024-09-17
Add a comment to auto-generated file

stephenfin:
Useful breadcrumb for people who do not touch openshift/release as often as they probably should.

This is half-baked. I'm mainly pushing to see if it's a good idea. If it is, I'll flesh it out more/do it better (TM).


smg247:
@stephenfin, I appreciate the attempt to make it more clear, but this isn't something that we will want to do. The file actually can be edited by hand, and there are many usecases for doing so. Individual jobs are generated by `prowgen`, but even those are tolerant of specific manual changes. This comment will be more confusing than helpful, I think

stephenfin:
@smg247 Apologies, I missed your response. I wasn't aware that you could modify these by hand and assumed the `generated-config` job (i.e. the `hack/validate-generated-config.sh` script) and other related jobs like `ordered-prow-config` would prevent that. Is that possible in `openshift/release` or are you talking about other places that `prowgen` is used (are there any)?

Happy to close this of course, but before I do, I'll ask would it make sense to add an alternative note? Ideally something like the below that would provide helpful breadcrumbs to newer contributors?

```
# This file was initially auto-generated by prowgen from the following sources:
# 
# <path/to/config/file/a>
# <path/to/config/file/b>
# ...
#
# You can modify this by hand, but you may need to run prowgen after the fact to
# ensure conflicts are not introduced.
```

(assuming we even have the context at this stage of the pipeline to populate this list, of course).

It's a totally different problem and totally different toolkit, but the storage folks do something like this in [`openshift/csi-operator`](https://github.com/openshift/csi-operator/blob/7a185cd28d13c01f84af914e5904c36b941c0948/assets/overlays/openstack-cinder/generated/standalone/controller.yaml#L1-L22) and it is indeed very helpful at the outset.

smg247:
@stephenfin, I don't see any downsides to a comment like that, but I think that DPTP will want to craft the exact verbiage of the comment. Can you file a card for us in our DPTP jira project, and link this PR as a draft?

stephenfin:
> @stephenfin, I don't see any downsides to a comment like that, but I think that DPTP will want to craft the exact verbiage of the comment. Can you file a card for us in our DPTP jira project, and link this PR as a draft?

Done https://issues.redhat.com/browse/DPTP-4361

PR #4331, opened 2024-09-18
Fix issue DPTP-4032 Delete firewall when RH intranet access is desired

bear-redhat:
We will use template to create the firewall object when NS creates. `ci-operator` just need to delete those created object on-demand.


bear-redhat:
/hold
Hold until the `prowgen` changes are reversed.

smg247 (commented):
/lgtm

PR #4339, opened 2024-09-20
bump `prow` dependency

smg247:
Needed to grab: https://github.com/kubernetes-sigs/prow/pull/275


psalajova:
The failing `validate-prow` job was an issue when we were bumping the ci-tools repo with Danilo a while ago; it demands that release-controller and ci-chat-bot repos need to have their prow versions bumped as well

smg247:
> The failing `validate-prow` job was an issue when we were bumping the ci-tools repo with Danilo a while ago; it demands that release-controller and ci-chat-bot repos need to have their prow versions bumped as well

yes, I took care of that already

PR #4343, opened 2024-09-23
`prowjob-dispatcher`: refactor dispatcher and create a client for use by other tools

smg247:
This moves the server and necessary logic into the `pkg/dispatcher`, and adds a `client`. This is necessary for tools like `payload-testing's` `prpqr_reconciller` and `multi-pr-prow-plugin` to be able to know which `cluster` to set on job's they create.

/cc @jmguzik @openshift/test-platform


Prucek:
/lgtm
/hold for @jmguzik to review
but if you are blocked with this, feel free to unhold, it's nicely done

PR #4360, opened 2024-10-04
`multi-pr-prow-plugin`: Create a prow plugin to handle multi-pr presubmit testing

smg247:
The initial MVP for multi-pr presubmit testing. There are a couple of TODOS left in here intentionally, and a couple of open questions that I have that will only be answered once I start trying this out on real scenarios. Expect a few follow-ups here.

For: https://issues.redhat.com/browse/DPTP-4148

### DIFF
--- a/images/multi-pr-prow-plugin/Dockerfile
+++ b/images/multi-pr-prow-plugin/Dockerfile
@@ -1,0 +1,3 @@
+FROM quay.io/centos/centos:stream8
+ADD multi-pr-prow-plugin /usr/bin/multi-pr-prow-plugin
+ENTRYPOINT ["/usr/bin/multi-pr-prow-plugin"]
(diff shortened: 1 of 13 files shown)


images/multi-pr-prow-plugin/Dockerfile
Prucek, line 1: I thought we don't use this anymore and publish the images here: [ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml](https://github.com/openshift/release/blob/master/ci-operator/config/openshift/ci-tools/openshift-ci-tools-master.yaml#L163)

smg247: Defining a Dockerfile is necessary for the `build-and-publish-image` make target to work. This allows us to dev test locally made changes to the image. Defining one has always been optional, but I prefer to do so for this reason.

PR #4369, opened 2024-10-14
Extend cluster config: capacity, capabilities

jmguzik:
Only extend - no functionality integrated yet.

### DIFF
--- a/pkg/sanitizer/helpers.go
+++ b/pkg/sanitizer/helpers.go
@@ -10,34 +10,59 @@ import (
 
 const blocked = "blocked"
 
-type ClusterMap map[string]string
+// ClusterInfo holds the provider, capacity, and capabilities.
+type ClusterInfo struct {
+	Provider     string
+	Capacity     int
+	Capabilities []string
+}
 
+// ClusterMap maps a cluster name to its corresponding ClusterInfo.
+type ClusterMap map[string]ClusterInfo
+
+func loadClusterConfigFromBytes(data []byte) (ClusterMap, sets.Set[string], error) {
+	var clusters map[string][]struct {
+		Name         string   `yaml:"name"`
+		Capacity     int      `yaml:"capacity"`
+		Capabilities []string `yaml:"capabilities"`
+	}
+	if err := yaml.Unmarshal(data, &clusters); err != nil {
+		return nil, nil, err
+	}
+	blockedClusters := sets.New[string]()
+	clusterMap := make(ClusterMap)
+
+	for provider, clusterList := range clusters {
+		if provider != blocked {
+			for _, cluster := range clusterList {
+				if cluster.Capacity == 0 {
+					cluster.Capacity = 100
+				}
+				clusterMap[cluster.Name] = ClusterInfo{
+					Provider:     provider,
+					Capacity:     cluster.Capacity,
+					Capabilities: cluster.Capabilities,
+				}
+			}
+		}
+		if provider == blocked {
+			for _, cluster := range clusterList {
+				blockedClusters.Insert(cluster.Name)
+			}
+		}
+	}
+
+	return clusterMap, blockedClusters, nil
+}
+
+// LoadClusterConfig loads cluster configuration from a YAML file, returning a ClusterMap and a set of blocked clusters.
 func LoadClusterConfig(filePath string) (ClusterMap, sets.Set[string], error) {
 	data, err := os.ReadFile(filePath)
 	if err != nil {
 		return nil, nil, err
 	}
+	return loadClusterConfigFromBytes(data)
 
-	var clusters map[string][]string
-	err = yaml.Unmarshal(data, &clusters)
-	if err != nil {
-		return nil, nil, err
-	}
-
-	blockedClusters := sets.New[string]()
-	clusterMap := make(ClusterMap)
-	for provider, clusterList := range clusters {
-		if provider != blocked {
-			for _, cluster := range clusterList {
-				clusterMap[cluster] = provider
-			}
-		}
-		if provider == blocked {
-			blockedClusters.Insert(clusterList...)
-		}
-	}
-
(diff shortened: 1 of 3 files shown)


pkg/sanitizer/helpers.go
```diff
@@ -10,36 +10,61 @@ import (
 
 const blocked = "blocked"
 
-type ClusterMap map[string]string
+// ClusterInfo holds the provider, capacity, and capabilities.
+type ClusterInfo struct {
+	Provider     string
+	Capacity     int
+	Capabilities []string
```
Prucek, line 17: Do we maybe want to create a type Capabilty? So we don't compare it with a string value?
Could we maybe add `capabilities` stanza to the config so we know what capabilities are available?
```
capabilities:
   - vpn
   - aarch64
   - arm64
aws:
...
blocked:
...
```

pkg/sanitizer/helpers.go
```diff
@@ -10,36 +10,61 @@ import (
 
 const blocked = "blocked"
 
-type ClusterMap map[string]string
+// ClusterInfo holds the provider, capacity, and capabilities.
+type ClusterInfo struct {
+	Provider     string
+	Capacity     int
```
Prucek, line 16: What is capacity? Capacity of what?

pkg/sanitizer/helpers.go
Prucek: Just ideas....

pkg/sanitizer/helpers.go
jmguzik: cluster capacity, how many % of load can be accepted. Will be in documentation

pkg/sanitizer/helpers.go
jmguzik: string is fine because labels in prowjob will be strings as well

pkg/sanitizer/helpers.go
```diff
@@ -8,38 +8,61 @@ import (
...
-type ClusterMap map[string]string
+// ClusterMap maps a cluster name to its corresponding ClusterInfo.
+type ClusterMap map[string]ClusterInfo
 
-func LoadClusterConfig(filePath string) (ClusterMap, sets.Set[string], error) {
-	data, err := os.ReadFile(filePath)
-	if err != nil {
-		return nil, nil, err
+func loadClusterConfigFromBytes(data []byte) (ClusterMap, sets.Set[string], error) {
+	var clusters map[string][]struct {
```
danilo-gemoli, line 22: Why an anonymous struct instead of enhancing the `ClusterMap` type?
As an example:
```golang
type ClusterMap struct {
	AWS     map[string]ClusterProvider `yaml:"aws,omitempty"`
	GCP     map[string]ClusterProvider `yaml:"gcp,omitempty"`
	Blocked []string                   `yaml:"blocked,omitempty"`
}

func (cm *ClusterMap) BlockedClusters() sets.Set[string] {
	return sets.New[string](cm.BlockedClusters().UnsortedList()...)
}

type ClusterProvider struct {
	Capacity     int      `yaml:"capacity"`
	Capabilities []string `yaml:"capabilities"`
}
```

pkg/sanitizer/helpers.go
```diff
@@ -8,38 +8,61 @@ import (
...
 }
 
+// LoadClusterConfig loads cluster configuration from a YAML file, returning a ClusterMap and a set of blocked clusters.
+func LoadClusterConfig(filePath string) (ClusterMap, sets.Set[string], error) {
+	data, err := os.ReadFile(filePath)
+	if err != nil {
+		return nil, nil, err
+	}
+	return loadClusterConfigFromBytes(data)
+
```
danilo-gemoli, line 63: Remove this blank line.

pkg/sanitizer/helpers.go
jmguzik: Not sure why we should limit ourselves to gcp and aws since we discuss next cluster providers, I guess we should not create static types like this. and as for anonymous, it is used only there (it differs from what is later used and translated in the code) so no need to take it out of this function.

pkg/sanitizer/helpers.go
jmguzik: If that will be the only discussed change, I will do it in the follow-up PR.

pkg/sanitizer/helpers.go
droslean: I don't think that capacity makes any sense. There can be more jobs that use less resources. Are there any plans to calculate the workload of the cluster based on the resource usage?

droslean (approved):

pkg/sanitizer/helpers.go
jmguzik: yes, based on prometheus data. It is needed for the special events if you want to run with degradated load. It was discussed on arch call

jmguzik:
/hold 
till next week because of breaking changes

pkg/sanitizer/helpers.go
danilo-gemoli: > Not sure why we should limit ourselves to gcp and aws since we discuss next cluster providers

There are not so many out there that we might support, and could be added when needed.
But if you don't like "fixed" providers:

```golang
type ClusterMap struct {
	Clusters map[string]ClusterProvider `json:"providers,omitempty"`
	Blocked  []string                   `json:"blocked,omitempty"`
}

type ClusterProvider struct {
	Capacity     int      `json:"capacity"`
	Capabilities []string `json:"capabilities"`
}
```
Consider introducing something like:
```golang
func (cm *ClusterMap) BlockedClusters() sets.Set[string] {
	return sets.New[string](cm.BlockedClusters().UnsortedList()...)
}
```
because `func` returning three parameters are quite unidiomatic in Go:
```golang
LoadClusterConfig(filePath string) (ClusterMap, sets.Set[string], error)
```

PR #4375, opened 2024-10-17
Update OWNERS

pruan-rht:
added pruan-rht as part of onboarding.


pruan-rht:
outdated PR...closing.

PR #4377, opened 2024-10-17
update prow dependency

smg247:
Need to update so that https://github.com/openshift/release/pull/57951 can be merged. The `determinize-prow-config` test cannot succeed when using the new field without this.


smg247:
https://github.com/openshift/ci-chat-bot/pull/519
https://github.com/openshift/release-controller/pull/638

smg247:
/hold cancel
Nevermind, It shouldn't matter for our purposes, we can just grab it the next time

PR #4381, opened 2024-10-22
clean up a couple of TODOs that are no longer relevant, and add plugin description

smg247:
The TODOs have already been taken care of, and the description was just missed in the original PR


deepsm007:
No point in running e2e, we can override

PR #4385, opened 2024-10-25
WIP: Capacity feature for dispatcher

jmguzik:
Allows to reduce cluster capacity


jmguzik:
Some things still missing: how to detect capacity change with current impl.

PR #4389, opened 2024-10-29
Add retry during extraction

deepsm007:
/cc @openshift/test-platform

Related to the `unexpected EOF`  errors  
https://redhat-internal.slack.com/archives/CBN38N3MW/p1730198582759599

### DIFF
--- a/pkg/steps/release/create_release.go
+++ b/pkg/steps/release/create_release.go
@@ -204,8 +204,21 @@ done
 if [[ "$exit_code" != "0" ]]; then
 	exit $exit_code
 fi
-oc adm release extract --from=%q --to=${ARTIFACT_DIR}/release-payload-%s
-`, s.jobSpec.Namespace(), streamName, cvo, destination, version, destination, s.name),
+for ((i=1; i<=5; i++)); do
+	rm -rf ${ARTIFACT_DIR}/release-payload-%s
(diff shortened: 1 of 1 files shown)


pkg/steps/release/create_release.go
```diff
@@ -204,7 +204,19 @@ done
 if [[ "$exit_code" != "0" ]]; then
 	exit $exit_code
 fi
-oc adm release extract --from=%q --to=${ARTIFACT_DIR}/release-payload-%s
+for ((i=1; i<=5; i++)); do
+	if oc adm release extract --from=%q --to=${ARTIFACT_DIR}/release-payload-%s; then
```
stbenjam, line 208: Do you know if it's a problem on retry if `${ARTIFACT_DIR}/release-payload-%s` still has contents in it?

stbenjam (commented):
Looks useful to get some debug data. Not the only place we run `oc adm release extract`, though.

pkg/steps/release/create_release.go
deepsm007: I tried running it locally with partial contents and there was no error

deepsm007:
/test e2e

breaking-changes test is failing because of [this](https://github.com/openshift/ci-tools/pull/4261)

PR #4391, opened 2024-11-01
Remove code related to `cluster/test-deploy`

psalajova:
With the deprecation `cluster/test-deploy` in the release repo, removing related code also in ci-tools repo.

### DIFF
--- a/pkg/rehearse/rehearse.go
+++ b/pkg/rehearse/rehearse.go
@@ -47,8 +47,7 @@ type RehearsalConfig struct {
 	ProwjobNamespace string
 	PodNamespace     string
 
-	NoRegistry        bool
-	NoClusterProfiles bool
+	NoRegistry bool
 
 	NormalLimit int
 	MoreLimit   int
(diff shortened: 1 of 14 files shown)


pkg/rehearse/rehearse.go
```diff
@@ -399,25 +382,6 @@ func determineChangedRegistrySteps(candidate, baseSHA string, logger *logrus.Ent
 	return changedRegistrySteps, nil
 }
 
-func determineChangedClusterProfiles(candidate, baseSHA, headSHA string, prNumber int, configUpdaterCfg prowplugins.ConfigUpdater, logger *logrus.Entry) (*ConfigMaps, error) {
```
psalajova, line 402: I think we can just delete this function? It is called 'cluster profiles' but it looks like it just deals with config maps, which we no longer need

smg247:
Overall, looks good. Can you also remove the `no-cluster-profiles` option and related code?

psalajova:
I'm not sure if the failure in the `breaking-changes` test doesn't seem related to this PR?

PR #4396, opened 2024-11-08
[PERFSCALE-3337] New jetlag-ci cluster profile

josecastillolema:


josecastillolema:
cc @rsevilla87 @jtaleric @krishvoor  besides the new jetlag profile correcting the naming of metal-perfscale-cpt, we will need to update the config files accordingly

PR #4402, opened 2024-11-12
`ci-operator`: allow for the possibility of fetching the config from the `CONFIG_SPEC_GCS_URL`

smg247:
The first part of allowing rehearsals to retrieve their ci-operator config spec from a GCS bucket is to amend ci-operator to get it from GCS when the proper env var is found.

For: https://issues.redhat.com/browse/DPTP-4241

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -57,6 +57,7 @@ import (
 	prowapi "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	_ "sigs.k8s.io/prow/pkg/cache"
 	"sigs.k8s.io/prow/pkg/config/secret"
+	prowio "sigs.k8s.io/prow/pkg/io"
 	"sigs.k8s.io/prow/pkg/logrusutil"
 	"sigs.k8s.io/prow/pkg/pod-utils/downwardapi"
 	"sigs.k8s.io/prow/pkg/version"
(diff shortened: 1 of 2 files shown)


cmd/ci-operator/main.go
```diff
@@ -2176,13 +2184,46 @@ func (o *options) getInjectTest() (*api.MetadataWithTest, error) {
...
-	configSpecEnv, configSpecSet := os.LookupEnv("CONFIG_SPEC")
-	unresolvedConfigEnv, unresolvedConfigSet := os.LookupEnv("UNRESOLVED_CONFIG")
+	configSpecEnv, configSpecSet := os.LookupEnv(configSpecVar)
+	configSpecGCSEnv, configSpecGCSSet := os.LookupEnv(configSpecGcsUrlVar)
+	unresolvedConfigEnv, unresolvedConfigSet := os.LookupEnv(unresolvedConfigVar)
 
+	decodeAndUnzip := func(rawConfig string) (string, error) {
+		// if being run by pj-rehearse, config spec may be base64 and gzipped
+		if decoded, err := base64.StdEncoding.DecodeString(rawConfig); err != nil {
+			return rawConfig, nil
```
danilo-gemoli, line 2218: Why don't we propagate `err` here?

danilo-gemoli:
The size of `ci-operator/main.go` is becoming a threat for my mental sanity. Is it ok if we start splitting it up and add a `load.go` or something?

cmd/ci-operator/main.go
smg247: I was just maintaining the original behavior. I believe that the intention is to assume the rawConfig is not encoded or zipped in that case. I guess that logging or failing would not be useful.

smg247:
> The size of `ci-operator/main.go` is becoming a threat for my mental sanity. Is it ok if we start splitting it up and add a `load.go` or something?

I agree with you, but I don't think that now is the time. This is really adding a very small drop to the bucket of file size. We should have a refactoring effort though. I will file a card and link it.

PR #4410, opened 2024-11-14
`pj-rehearse` upload the ci-operator config-spec to GCS 

smg247:
Following up on https://github.com/openshift/ci-tools/pull/4402, we now upload the config-spec to GCS and add the _full_ URL to the `CONFIG_SPEC_GCS_URL` env var.

The `prNumber` argument that was previously passed around everywhere is now replaced by looking in the `refs` where it will always be located anyways.

This also includes a change to `ci-operator` to expect the full GCS location, rather than just the file name. This is to make it easier to move it to a new bucket if desired in the future.

The existing `CONFIG_SPEC` env var logic still exists as it is simpler for the integration test, so I have put it behind the `dryRun` flag. I also filed https://issues.redhat.com/browse/DPTP-4243 to redesign the integration test entirely, as it is getting to be outdated.

For: https://issues.redhat.com/browse/DPTP-4241

### DIFF
--- a/pkg/rehearse/jobs.go
+++ b/pkg/rehearse/jobs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"slices"
 	"sort"
 	"strconv"
@@ -28,6 +29,7 @@ import (
 	pjapi "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	prowconfig "sigs.k8s.io/prow/pkg/config"
 	"sigs.k8s.io/prow/pkg/pjutil"
+	"sigs.k8s.io/prow/pkg/pod-utils/gcs"
 
 	"github.com/openshift/ci-tools/pkg/api"
 	apihelper "github.com/openshift/ci-tools/pkg/api/helper"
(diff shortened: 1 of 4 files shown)


pkg/rehearse/jobs.go
```diff
@@ -372,39 +374,87 @@ func inlineCiOpConfig(container *v1.Container, ciopConfigs config.DataByFilename
...
 	}
 
-	ciOpConfigContent, imageStreamTags, err := getResolvedConfigForTest(ciopConfig, resolver, testname)
+	ciOpConfigContent, imageStreamTags, err := getResolvedConfigForTest(ciopConfig, resolver, testName)
 	if err != nil {
 		logger.WithError(err).Error("Failed to get resolved config for test")
 		return nil, err
 	}
 	apihelper.MergeImageStreamTagMaps(allImageStreamTags, imageStreamTags)
+	if !jc.dryRun {
```
danilo-gemoli, line 383: Upload the config on `!jc.dryRun` makes sense, but I'd expect nothing more than logging when `dryRun == true`. Is perhaps an `enum` more appropriate in this case?

smg247: The dryRun functionality essentially just means "integration test mode". It is very bad, and that is why I have filed https://issues.redhat.com/browse/DPTP-4243 to completely redesign it.

danilo-gemoli: It's misleading but I don't want to delay this PR further just to add one comment. So, it's ok.

PR #4411, opened 2024-11-15
Capabilities in the dispatcher algorithm

jmguzik:
prowgen is coming soon as separate PR

### DIFF
--- a/pkg/dispatcher/config.go
+++ b/pkg/dispatcher/config.go
@@ -18,6 +18,16 @@ import (
 	"github.com/openshift/ci-tools/pkg/util/gzip"
 )
 
+// ClusterInfo holds the provider, capacity, and capabilities.
+type ClusterInfo struct {
+	Provider     string
+	Capacity     int
+	Capabilities []string
+}
+
+// ClusterMap maps a cluster name to its corresponding ClusterInfo.
+type ClusterMap map[string]ClusterInfo
+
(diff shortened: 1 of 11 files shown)


pkg/dispatcher/config.go
```diff
@@ -120,6 +154,38 @@ func (config *Config) DetermineClusterForJob(jobBase prowconfig.JobBase, path st
...
+					provider := clusterInfo.Provider
+					matchingClustersByProvider[provider] = append(matchingClustersByProvider[provider], clusterName)
+				}
+			}
+			if config.DetermineE2EByJob {
+				if cloud := config.DetermineCloudMapping(jobBase); cloud != "" {
+					if clusters, ok := matchingClustersByProvider[cloud]; ok {
+						if len(clusters) > 0 {
+							// as in other places in this file, use this method to have basic deterministic distribution
+							return api.Cluster(clusters[len(filepath.Base(path))%len(clusters)]), false, nil
```
smg247, line 175: Can we extract the "basic deterministic distribution" logic to a function?

danilo-gemoli:
mostly ok to me

pkg/dispatcher/config.go
jmguzik: We could but it would be very strange one liner as every time basically we are diving against different table len

jmguzik:
e2e are failing on infra failure on b10
/override ci/prow/e2e

PR #4412, opened 2024-11-15
`multi-pr`: fix mergedConfig validation

smg247:
Creates a new `mergedConfig` bool based on the presence of an injected test (which **always** means we are merging the configs. This bool is used to trigger specific validation rather than relying on the absence of metadata. This improves our validation, while still allowing for the proper logic to occur when we are running multi-pr presubmit tests for multiple PRs on the same repo.

For: https://issues.redhat.com/browse/DPTP-4246

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -584,7 +584,8 @@ func (o *options) Complete() error {
 	}
 	o.configSpec = config
 	o.jobSpec.Metadata = config.Metadata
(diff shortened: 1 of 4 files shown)


deepsm007 (approved):

EmilienM:
> /hold for manual testing

how do you do manual testing?

smg247:
> how do you do manual testing?
Build the `ci-operator-configresolver` locally and deploy an 'alpha' version of the tool using that image. I can then use that on my forked repo to test out the full workflow. I did find an issue that I need to fix real quick, unfortunately.

PR #4416, opened 2024-11-18
`payload testing`: use the `prowjob-dispatcher` client to determine the cluster for jobs

smg247:
Mimics the logic created for the `multi-pr-prow-plugin` to use the `prowjob-dispatcher` to dynamically determine the cluster for the submitted payload testing jobs. Includes a `15` minute cache to limit duplicate requests.

For: https://issues.redhat.com/browse/DPTP-3676


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -693,16 +703,34 @@ func generateProwjob(ciopConfig *api.ReleaseBuildConfiguration,
...
 
-	pj := pjutil.NewProwJob(pjutil.PeriodicSpec(*periodic), labels, annotations, pjutil.RequireScheduling(getCfg.Config().Scheduler.Enabled))
+	pj := pjutil.NewProwJob(pjutil.PeriodicSpec(*periodic), labels, annotations, pjutil.RequireScheduling(r.prowConfigGetter.Config().Scheduler.Enabled))
 	pj.Namespace = prpqrNamespace
 
 	return &pj, nil
 }
 
+func (r *reconciler) clusterForJob(jobName string) (string, error) {
+	if time.Now().Add(time.Minute * -15).After(r.jobClusterCache.lastCleared) {
```
droslean, line 717: Do we need to use a cache? The dispatcher is already caching the job configurations.

cmd/job-trigger-controller-manager/main.go
```diff
@@ -42,6 +43,7 @@ func gatherOptions() (*options, error) {
 	fs.BoolVar(&o.dryRun, "dry-run", true, "Whether to run the controller-manager with dry-run")
 	fs.StringVar(&o.namespace, "namespace", "ci", "In which namespace the operation will take place")
 	fs.Int64Var(&o.jobTriggerWaitInSeconds, "job-trigger-wait-seconds", 60, "Amount of seconds to wait for job to trigger in order to update status")
+	fs.StringVar(&o.dispatcherAddress, "dispatcher-address", "http://prowjob-dispatcher.ci.svc.cluster.local:8080", "Address of prowjob-dispatcher server.")
```
droslean, line 46: It's just food for thought for the future before expanding more. Since we are in the same cluster, and the services need to communicate, I would prefer using gRPC.

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
smg247: Yes, Jakub recommended we cache the responses at the client level

cmd/job-trigger-controller-manager/main.go
smg247: Do you know of any place where we are doing that already? outside of gangway, of course

droslean: I am not sure, but in theory we could have a proto repository holding all protobuf configuration that any component can just use based on the service it needs. We could have a list of services with a list of rpc calls that satisfy the needs of our infra. Not directly related to that PR, but in some point I would prefer using gRPC instead of REST API calls.

smg247: I do think that is a good direction.

PR #4422, opened 2024-11-22
cluster-display: fetch disabled clusters on demand

danilo-gemoli:
Fetch disabled clusters from Prow on demand instead of getting the list at startup.
This handles the problem of a new cluster being disabled and `cluster-display` not being able to notice it unless restarted.

### DIFF
--- a/cmd/cluster-display/main.go
+++ b/cmd/cluster-display/main.go
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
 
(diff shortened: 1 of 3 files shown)


cmd/cluster-display/main.go
```diff
@@ -14,6 +14,7 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
+	"golang.org/x/time/rate"
```
liangxia, line 17: Looks like we also need to update `go.mod` ?

danilo-gemoli: Yes, updated. Thanks!

PR #4434, opened 2024-12-02
adding the telcov10n ci/cd cluster profile

kononovn:


pkg/api/types.go
```diff
@@ -1346,6 +1346,7 @@ const (
 	ClusterProfileMetalPerfscaleCPT     ClusterProfile = "metal-perscale-cpt"
 	ClusterProfileMetalPerfscaleJetlag  ClusterProfile = "metal-perfscale-jetlag"
 	ClusterProfileMetalTelco5G          ClusterProfile = "metal-telco5g"
+	ClusterProfileMetalTelcoV10N		ClusterProfile = "metal-telcov10n"
```
pruan-rht, line 1349: @kononovn, there seems to be a text alignment issue here...minor nitpick.

kononovn: +1 fixed.

PR #4442, opened 2024-12-06
[DPTP-4239] Verifying branch protection on public repos or paid plans

hector-vido:
This modifications verify if branch protection is enable for private repositories using a free plan, if this is the case, the verification should return an error.

### DIFF
--- a/cmd/check-gh-automation/main.go
+++ b/cmd/check-gh-automation/main.go
@@ -13,6 +13,7 @@ import (
 	"sigs.k8s.io/prow/pkg/flagutil"
 	configflagutil "sigs.k8s.io/prow/pkg/flagutil/config"
 	prowpluginconfig "sigs.k8s.io/prow/pkg/flagutil/plugins"
+	"sigs.k8s.io/prow/pkg/github"
 	"sigs.k8s.io/prow/pkg/logrusutil"
 	"sigs.k8s.io/prow/pkg/plugins"
 	"sigs.k8s.io/prow/pkg/pod-utils/downwardapi"
@@ -82,6 +83,8 @@ type automationClient interface {
 	IsCollaborator(org, repo, user string) (bool, error)
 	IsAppInstalled(org, repo string) (bool, error)
 	HasPermission(org, repo, user string, permissions ...string) (bool, error)
+	GetRepo(owner, name string) (github.FullRepo, error)
+	GetOrg(name string) (*github.Organization, error)
 }
 
 func main() {
@@ -155,6 +158,12 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
 			"repo": repo,
 		})
 
+		fullRepo, err := client.GetRepo(org, repo)
+		if err != nil {
+			logger.Errorf("Error obtaining repository from github: %s/%s: %v", org, repo, err)
+			return nil, fmt.Errorf("error obtaining repository from github: %s/%s: %w", org, repo, err)
+		}
+
 		if ignore.Has(org) || ignore.Has(orgRepo) {
 			repoLogger.Infof("skipping ignored repo")
 			continue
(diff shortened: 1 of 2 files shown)


cmd/check-gh-automation/main.go
```diff
@@ -191,6 +200,11 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
 			repoLogger.Info("no bots provided to check")
 		}
 
+		fullOrg, err := client.GetOrg(org)
+		if err != nil {
+			logger.Errorf("Error checking organization %s: %v", org, err)
+			return nil, fmt.Errorf("error checking organization %s: %w", org, err)
```
smg247, line 206:
```suggestion
			logger.Errorf("Error obtaining org from github: %s: %v", org, err)
			return nil, fmt.Errorf("error obtaining org from github: %s: %w", org, err)
```

cmd/check-gh-automation/main.go
```diff
@@ -206,8 +220,20 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
...
 
 			// If branch protection is configured, verify admin access for the hardcoded admin bot
+			// also branch protection should be used only on public repos or paid plans
 			if branchProtectionEnabled {
 				logger.Infof("Branch protection is enabled for %s/%s", org, repo)
+
+				if err != nil {
+					logger.Errorf("Error gathering repository data in %s/%s: %v", org, repo, err)
+					return nil, fmt.Errorf("error gathering repository data in %s/%s: %w", org, repo, err)
+				}
```
smg247, line 230: This error doesn't make sense...what is throwing an error here?

cmd/check-gh-automation/main.go
```diff
@@ -206,8 +220,20 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
...
+				if fullRepo.Private && fullOrg.Plan.Name == "free" {
```
smg247, line 231: Just thought of this, but is it possible for a repo to have its own "plan"? We should verify this prior to merging...

cmd/check-gh-automation/main.go
```diff
@@ -206,8 +220,20 @@ func checkRepos(repos []string, bots []string, appName string, ignore sets.Set[s
...
+					failing.Insert(orgRepo)
+					logger.Errorf("Branch protection is enabled, the org %s has a free plan, the repository %s must be public", org, repo)
+					return nil, fmt.Errorf("branch protection is enabled, the org %s has a free plan, the repository %s must be public", org, repo)
```
smg247, line 234: We shouldn't just return here. Just log and add the repo to the `failing` set.

cmd/check-gh-automation/main_test.go
```diff
@@ -329,6 +369,24 @@ func TestCheckRepos(t *testing.T) {
...
+			expected: []string{},
+		},
+		{
+			name:     "public repository has branch protection enabled and is a free plan",
+			repos:    []string{"org-5/repo-a"},
+			mode:     standard,
+			expected: []string{},
+		},
+		{
+			name:        "private repository has branch protection enabled and is not a free plan",
```
smg247, line 385: Is this test named correctly? Seems like it **is** the free plan. Either way, you should have tests for both cases.

cmd/check-gh-automation/main.go
hector-vido: This is a left over of the other modifications I was doing, thanks for noticing!

cmd/check-gh-automation/main.go
hector-vido: There is no "plan" property on "get a repository" api endpoint:
https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#get-a-repository

Also, the [Github documentation](https://docs.github.com/en/get-started/learning-about-github/githubs-plans) says that is only available for accounts:
```GitHub offers free and paid plans for storing and collaborating on code. Some plans are available only to personal accounts, while other plans are available only to organization and enterprise accounts.```

File: cmd/check-gh-automation/main_test.go (hunk 5 above)
Comment (hector-vido) [reply cmd/check-gh-automation/main_test.go:385]: No is not, thanks for noticing!

File: cmd/check-gh-automation/main.go (hunk 4 above)
Comment (hector-vido) [reply cmd/check-gh-automation/main.go:234]: True, this was given false positive on one of the tests!

File: cmd/check-gh-automation/main_test.go [hunk 6]
```diff
@@ -329,6 +372,24 @@ func TestCheckRepos(t *testing.T) {
...
+			mode:     standard,
+			expected: []string{},
+		},
+		{
+			name:     "private repository has branch protection enabled and is a free plan",
+			repos:    []string{"org-5/repo-d"},
+			mode:     standard,
+			expected: []string{"org-5/repo-d"},
+		},
 	}
```
smg247, line 393: Can we add a "private repository has branch protection enabled and is a pro plan" case too?

PR #4457, opened 2024-12-31
Some security tweaks

hector-vido:
This update is to solve some problems in our security test.

Despite all my modifications, Snyk still consider a possible Server-Side Request Forgery (SSRF) in `cmd/job-runtime-analyzer/main.go:32` - **I am accepting ideas s2**

About javascript, basically everything that is `innerHTML` but only renders text can be replaced with `innerText`.
When HTML is needed, in simple cases we can use `appendChild` with `document.createTextNode` and `document.createElement`.
To erase elements inside another one we use the following code to avoid `innerHTML`, manipulating DOM directly:
```
    while(element.firstChild)
      element.removeChild(element.firstChild);
```

Since this is `TypeScript` there is better ways to render HTML.

Some errors was fixed with `filepath.Clean`, but @smg247 noticed that `filepath.Join` already executes a `filepath.Clean` internally, I removed it and start to talk with Snyk people to see if they can fix that.
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/path/filepath/path.go;l=130 
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/path/filepath/path_unix.go;l=36

### DIFF
--- a/cmd/autoowners/main.go
+++ b/cmd/autoowners/main.go
@@ -237,7 +237,7 @@ func addHeader(path string, header string) error {
 func writeOwners(orgRepo orgRepo, httpResult httpResult, cleaner ownersCleaner, header string) error {
 	for _, directory := range orgRepo.Directories {
 		path := filepath.Join(directory, "OWNERS")
-		err := os.Remove(path)
+		err := os.Remove(filepath.Clean(path))
 		if err != nil && !os.IsNotExist(err) {
 			return err
 		}
--- a/cmd/job-runtime-analyzer/main.go
+++ b/cmd/job-runtime-analyzer/main.go
@@ -2,7 +2,6 @@ package main
 
 import (
 	"flag"
-	"html"
 
 	"github.com/sirupsen/logrus"
 
@@ -15,7 +14,7 @@ func main() {
 	jobURL := flag.String("job-url", defaultJobURL, "url to a job")
 	flag.Parse()
 
-	if err := jobruntimeanalyzer.Run(html.EscapeString(*jobURL)); err != nil {
+	if err := jobruntimeanalyzer.Run(*jobURL); err != nil {
 		logrus.WithError(err).Fatal("Failed")
 	}
 }
--- a/cmd/job-runtime-analyzer/main.go
+++ b/cmd/job-runtime-analyzer/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"html"
 
 	"github.com/sirupsen/logrus"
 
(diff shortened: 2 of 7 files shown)


cmd/autoowners/main.go
```diff
@@ -81,7 +81,7 @@ func loadRepos(configRootDir string, blocked blocklist, configSubDirs, extraDirs
 				logrus.WithField("org", org).Info("org is on organization blocklist, skipping")
 				continue
 			}
-			repoDirs, err := os.ReadDir(filepath.Join(subdirectory, orgDir.Name()))
+			repoDirs, err := os.ReadDir(filepath.Clean(filepath.Join(subdirectory, orgDir.Name())))
```
smg247, line 84: Doesn't `Join` already perform a `Clean` internally? I guess the snyk is unaware of that?

cmd/job-runtime-analyzer/main.go
```diff
@@ -1,20 +1,39 @@
 package main
```
smg247, line 1: I don't think anyone uses this tool anymore. I believe that all this info is already in BigQuery anyways. I guess I would rather see this removed than changed so heavily, but we could get confirmation from the rest of the team first.

cmd/vault-secret-collection-manager/index.ts
```diff
@@ -144,7 +144,12 @@ function getSelectValues(select: HTMLSelectElement): string[] {
 function deleteColectionEventHandler(collectionName: string) {
   return function () {
     const deleteConfirmation = document.getElementById('deleteConfirmation') as HTMLDivElement;
-    deleteConfirmation.innerHTML = `Are you sure you want to irreversibly delete the secret collection ${collectionName} and all its content?<br><br>`;
+    // clear div before adding elements, this avoids innerHTML issues
+    while(deleteConfirmation.firstChild)
+      deleteConfirmation.removeChild(deleteConfirmation.firstChild);
```
smg247, line 149:
```suggestion
    while(deleteConfirmation.firstChild) {
      deleteConfirmation.removeChild(deleteConfirmation.firstChild);
    }
```

cmd/job-runtime-analyzer/main.go
hector-vido: Nice, let's dig into it and if nobody is using, we remove.

cmd/autoowners/main.go
hector-vido: Right! I will remove it, talk with Snyk guys and see what happens.

https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/path/filepath/path.go;l=130 
https://cs.opensource.google/go/go/+/refs/tags/go1.23.4:src/path/filepath/path_unix.go;l=36

smg247 (commented):
Please revert changes to `job-runtime-analyzer`, and then this is good to merge

PR #4463, opened 2025-01-07
Security fixes - Second Round

hector-vido:
The only change here was the "upgrade" from md5 to sha224 - instead of sha256, avoiding Kubernetes 63 characters limitation for names.

https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names

This was done to silence one of the Snyk alerts and probably will be useful in the future to avoid hash collisions.


pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -547,10 +547,11 @@ func constructCondition(statuses map[string]*v1.PullRequestPayloadJobStatus) met
...
 
 func jobNameHash(name string) string {
-	hasher := md5.New()
-	// MD5 Write never returns error
+	hasher := sha256.New()
+	// SHA256 Write never returns error
 	_, _ = hasher.Write([]byte(name))
-	return hex.EncodeToString(hasher.Sum(nil))
+	// Simulate the same output as MD5 but with a newer hash algorithm
+	return hex.EncodeToString(hasher.Sum(nil))[:32]
```
danilo-gemoli, line 554: I'm wondering whether it's safe to remove `[:32]`

hector-vido: Well, this is only to trick the scanner to think that we are using a modern hash algorithm and keep the same size.
What do you think, should we use the entire hash?

danilo-gemoli: I guess so but I'd like to have another feedback on the matter @smg247

smg247: I don't think we need to simulate the same exact hash. As long as we have a hash it should be fine for our purposes. Using a different algorithm will result in actively running PRPQRs to likely have odd results on the UI, but that will be fairly low impact. I think we can omit the `[:32]` here.

hector-vido: Since k8s limits the names up to 63 characters, I will use sha224.

danilo-gemoli:
Rebase on top of master and then `braking-changes` should be fine

smg247 (commented):
/lgtm

PR #4466, opened 2025-01-10
config-brancher: respect prefix when bumping to future-release

Prucek:
This should fix the issue with variants of Openshift during branching and avoid PRs like this: https://github.com/openshift/release/pull/56051

Resolves: [DPTP-4150](https://issues.redhat.com/browse/DPTP-4150)


deepsm007:
Test e2e

PR #4467, opened 2025-01-14
Added the error on every return

hector-vido:
This code fix a problem that return the following message:

```
```

The local `err` variable was replacing the original on the function signature.

### DIFF
--- a/pkg/clusterinit/onboard/certmanager/certmanager.go
+++ b/pkg/clusterinit/onboard/certmanager/certmanager.go
@@ -165,18 +165,19 @@ loop:
 }
 
 // queryRedHatCatalog pull information regarding the cart-manager package from the Red Hat catalog.
-func queryRedHatCatalog(ctx context.Context, clientConnFactory GRPCClientConnFactory, port string) (pack *Package, err error) {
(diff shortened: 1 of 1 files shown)


pkg/clusterinit/onboard/certmanager/certmanager.go
```diff
@@ -165,18 +165,19 @@ loop:
 }
 
 // queryRedHatCatalog pull information regarding the cart-manager package from the Red Hat catalog.
-func queryRedHatCatalog(ctx context.Context, clientConnFactory GRPCClientConnFactory, port string) (pack *Package, err error) {
+func queryRedHatCatalog(ctx context.Context, clientConnFactory GRPCClientConnFactory, port string) (pack *Package, retErr error) {
+	pack = &Package{}
```
danilo-gemoli, line 169: Sorry for the back and forth but, is this required?

hector-vido: This is the "real" change on this PR, after this that `invalid memory address or nil pointer dereference` was solved.
The `err/retErr` is just to make the code more clear.

PR #4468, opened 2025-01-14
Added a verification to avoid duplicated clusters

hector-vido:
When someone execute `cluster-init onboard config generate` more than once, the cluster name will be appended to the list despite already being there.

I added a simple verification to see if that value is already present before `append`.


smg247 (commented):
Can these be made into sets instead and then output as lists at the end?

danilo-gemoli:
Thanks!
/lgtm

danilo-gemoli:
Steve asked something

PR #4479, opened 2025-01-30
`multi-pr-plugin`: clarify error message that is bubbled up to the user

smg247:
This message ends up as a comment on the PR when this is hit. This only happens when the user has entered an invalid command, so we can provide more info about what was incorrect.


jlebon:
Much better, thanks!
/lgtm

PR #4481, opened 2025-02-01
Avoid registry.access.redhat.com outage

jupierce:
registry.access.redhat.com is returning 503s
inside and outside the clusters.


jupierce:
This outage has passed. So no longer needed.

PR #4498, opened 2025-02-17
Pass pod_spec_options to the step.

LiangquanLi930:
Currently, when we use the Gangway API with the following structure:
```
"pod_spec_options": {
  "envs": {
    "xxxx": "xxxx"
  }
}
```
The variables are not being received in the step. However, in some cases, such as with the Hypershift team, this is required. For example, when performing multi-version testing, we need to change the MCE version via the Gangway API. Therefore, I'm submitting this PR to support passing pod_spec_options to the step.

### DIFF
--- a/cmd/ci-operator/main.go
+++ b/cmd/ci-operator/main.go
@@ -708,6 +708,8 @@ func (o *options) Complete() error {
 		o.hiveKubeconfig = kubeConfig
 	}
 
+	applyEnvOverrides(o)
+
 	if err := overrideMultiStageParams(o); err != nil {
 		return err
 	}
@@ -784,6 +786,29 @@ func overrideMultiStageParams(o *options) error {
 	}
 
 	return nil
+}
+
+// applyEnvOverrides processes environment variables with override prefixes and applies them to the test configurations.
+// It checks for environment variables that start with "MULTISTAGE_PARAM_OVERRIDE_" and applies them to the environment settings of each test.
+func applyEnvOverrides(o *options) {
+	for _, envVar := range os.Environ() {
+		if !strings.HasPrefix(envVar, "MULTISTAGE_PARAM_OVERRIDE_") {
+			continue
+		}
+		parts := strings.SplitN(envVar, "=", 2)
+		if len(parts) != 2 {
+			continue
+		}
+		key, value := parts[0], parts[1]
+		for _, test := range o.configSpec.Tests {
+			if test.MultiStageTestConfigurationLiteral != nil {
+				if test.MultiStageTestConfigurationLiteral.Environment == nil {
+					test.MultiStageTestConfigurationLiteral.Environment = make(api.TestEnvironment)
+				}
+				test.MultiStageTestConfigurationLiteral.Environment[key] = value
+			}
+		}
+	}
 }
 
 func overrideTestStepDependencyParams(o *options) error {
--- a/cmd/ci-operator/main_test.go
+++ b/cmd/ci-operator/main_test.go
@@ -1308,6 +1308,110 @@ func TestMultiStageParams(t *testing.T) {
 	}
 }
 
+func TestApplyEnvOverrides(t *testing.T) {
+	testCases := []struct {
+		id             string
+		envVars        map[string]string
+		expectedParams map[string]string
+		testConfig     []api.TestStepConfiguration
+		expectedErrs   []string
+	}{
+		{
+			id: "Apply overrides",
+			envVars: map[string]string{
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM1": "VAL1",
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM2": "VAL2",
+				"XXX_OVERRIDE_PARAM2":              "VAL3",
+			},
+			expectedParams: map[string]string{
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM1": "VAL1",
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM2": "VAL2",
+			},
+			testConfig: []api.TestStepConfiguration{
+				{
+					MultiStageTestConfigurationLiteral: &api.MultiStageTestConfigurationLiteral{
+						Environment: map[string]string{
+							"MULTISTAGE_PARAM_OVERRIDE_PARAM1": "VAL1",
+							"MULTISTAGE_PARAM_OVERRIDE_PARAM2": "VAL2",
+						},
+					},
+				},
+			},
+		},
+		{
+			id: "No OVERRIDE_ prefix",
+			envVars: map[string]string{
+				"PARAM1": "VAL1",
+			},
+			expectedParams: map[string]string{},
+			testConfig: []api.TestStepConfiguration{
+				{
+					MultiStageTestConfigurationLiteral: &api.MultiStageTestConfigurationLiteral{
+						Environment: map[string]string{},
+					},
+				},
+			},
+		},
+		{
+			id: "test environment variable contains an equal sign",
+			envVars: map[string]string{
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM1": "VAL2",
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM2": "VAL=2",
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM3": "VAL2",
+			},
+			expectedParams: map[string]string{
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM1": "VAL2",
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM2": "VAL=2",
+				"MULTISTAGE_PARAM_OVERRIDE_PARAM3": "VAL2",
+			},
+			testConfig: []api.TestStepConfiguration{
+				{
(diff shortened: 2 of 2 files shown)


psalajova:
Hi, can you please add a description of what is the context and reason for this change?

LiangquanLi930:
> Hi, can you please add a description of what is the context and reason for this change?

Thanks for your reminder, it's been updated.

pkg/steps/pod.go
```diff
@@ -217,6 +218,14 @@ func GenerateBasePod(
 	if err != nil {
 		return nil, err
 	}
+	var pj v1.ProwJob
+	if err = client.Get(CleanupCtx, ctrlruntimeclient.ObjectKey{Namespace: jobSpec.Namespace(), Name: jobSpec.ProwJobID}, &pj); err == nil {
```
LiangquanLi930, line 222: I'm not sure if the namespec is written correctly, or if we should write it as 
```
Namespace: "ci"
```
I noticed that the ns in the prowjob CRs in the job are usually "ci". For example, see this link: [prowjob.json](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_release/61690/rehearse-61690-periodic-ci-openshift-hypershift-release-4.16-periodics-mce-e2e-aws-ovn-conformance/1891033971593056256/prowjob.json). Or maybe it should be written in another way?

danilo-gemoli:
/hold
we are discussing [here](https://redhat-internal.slack.com/archives/CBN38N3MW/p1739802685433299?thread_ts=1739544357.642539&cid=CBN38N3MW)

cmd/ci-operator/main.go
```diff
@@ -786,6 +789,34 @@ func overrideMultiStageParams(o *options) error {
...
+				if test.MultiStageTestConfigurationLiteral.Environment == nil {
+					test.MultiStageTestConfigurationLiteral.Environment = make(api.TestEnvironment)
+				}
+
+				test.MultiStageTestConfigurationLiteral.Environment[env] = envValue
+			}
+		}
+
+	}
+}
```
jmguzik, line 818:
```suggestion
func applyEnvOverrides(o *options) {
	viper.AutomaticEnv()

	for _, envVar := range os.Environ() {
		parts := strings.SplitN(envVar, "=", 2)
		if len(parts) != 2 {
			continue
		}
		key, value := parts[0], parts[1]

		if !strings.HasPrefix(key, "OVERRIDE_") {
			continue
		}

		for _, test := range o.configSpec.Tests {
			if test.MultiStageTestConfigurationLiteral != nil {
				if test.MultiStageTestConfigurationLiteral.Environment == nil {
					test.MultiStageTestConfigurationLiteral.Environment = make(api.TestEnvironment)
				}
				test.MultiStageTestConfigurationLiteral.Environment[key] = value
			}
		}
	}
}
```

jmguzik: I would change the logic of this function to pass everything starting with OVERRIDE_ prefix to multistage test.

jmguzik (changes requested):
This is added logic and I would be in favor of testing it, although it's not public interface. One way would be to do viper outside of function and later take it as param. That way function could be UTested.

jmguzik:
/cc @danilo-gemoli 

Danilo, PTAL as well.

LiangquanLi930:
> This is added logic and I would be in favor of testing it, although it's not public interface. One way would be to do viper outside of function and later take it as param. That way function could be UTested.

Thanks for your review. I have made the adjustments, removed the viper, and added the UTest.

cmd/ci-operator/main.go
```diff
@@ -786,6 +788,29 @@ func overrideMultiStageParams(o *options) error {
...
+		key, value := parts[0], parts[1]
+		if !strings.HasPrefix(key, "OVERRIDE_") {
+			continue
+		}
+		for _, test := range o.configSpec.Tests {
+			if test.MultiStageTestConfigurationLiteral != nil {
+				if test.MultiStageTestConfigurationLiteral.Environment == nil {
+					test.MultiStageTestConfigurationLiteral.Environment = make(api.TestEnvironment)
+				}
+				test.MultiStageTestConfigurationLiteral.Environment[key] = value
```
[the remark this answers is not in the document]
LiangquanLi930: Theoretically, this is the case. We added the if just to ensure that no issues occur, in order to prevent errors when `applyEnvOverrides()` is moved in the future.

cmd/ci-operator/main.go
```diff
@@ -712,6 +712,8 @@ func (o *options) Complete() error {
 		return err
 	}
 
+	applyEnvOverrides(o)
```
danilo-gemoli, line 715: This should be placed before:
```golang
if err := overrideMultiStageParams(o); err != nil {
    return err
}
```
I guess cmd arguments take precedence over the env. variables.
One drawback of keeping `overrideMultiStageParams` and `applyEnvOverrides` separate is that now is quite hard to write a unit test to makes sure that cmd args override env. variables.
Unfortunately `func (o *options) Complete() error` is quite a mess to test.

cmd/ci-operator/main.go
```diff
@@ -786,6 +788,29 @@ func overrideMultiStageParams(o *options) error {
 	return nil
 }
 
+// applyEnvOverrides processes environment variables with override prefixes and applies them to the test configurations.
```
danilo-gemoli, line 791: Comment is nice but a bit outdated:
```golang
It checks for environment variables that start with "OVERRIDE_"
```
not the case anymore, it's `MULTISTAGE_PARAM_OVERRIDE_`

cmd/ci-operator/main.go
```diff
@@ -786,6 +788,29 @@ func overrideMultiStageParams(o *options) error {
...
+// It checks for environment variables that start with "OVERRIDE_" and applies them to the environment settings of each test.
+func applyEnvOverrides(o *options) {
+	for _, envVar := range os.Environ() {
+		parts := strings.SplitN(envVar, "=", 2)
+		if len(parts) != 2 {
+			continue
+		}
+		key, value := parts[0], parts[1]
+		if !strings.HasPrefix(key, "MULTISTAGE_PARAM_OVERRIDE_") {
```
danilo-gemoli, line 800: Move `strings.HasPrefix` at the top of this loop, before `strings.SplitN`

cmd/ci-operator/main_test.go
```diff
@@ -1308,6 +1309,125 @@ func TestMultiStageParams(t *testing.T) {
...
+
+	t.Parallel()
+
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.id, func(t *testing.T) {
+			for _, envVar := range tc.envVars {
+				parts := strings.SplitN(envVar, "=", 2)
+				if len(parts) == 2 {
+					_ = os.Setenv(parts[0], parts[1])
```
danilo-gemoli, line 1390: Please use `t.Setenv` rather than then `os` package. The the `defer` clause shouldn't be needed anymore.

cmd/ci-operator/main_test.go
```diff
@@ -1308,6 +1309,125 @@ func TestMultiStageParams(t *testing.T) {
 	}
 }
 
+func TestApplyEnvOverrides(t *testing.T) {
+	testCases := []struct {
+		id             string
+		envVars        []string
```
danilo-gemoli, line 1315: Turn `envVars` into a `map[string]string` and then you can just:
```golang
for k, v := range tc.envVars {
    t.Setevn(k, v)
}
```

cmd/ci-operator/main_test.go
```diff
@@ -1308,6 +1309,125 @@ func TestMultiStageParams(t *testing.T) {
...
+
+			o := &options{
+				configSpec: &configSpec,
+			}
+
+			applyEnvOverrides(o)
+
+			actualParams := make(map[string]string)
+
+			for _, test := range o.configSpec.Tests {
```
danilo-gemoli, line 1416: I would rather introduce a the field `wantConfig api.ReleaseBuildConfiguration` on the testCase struct and use that instead of extracting the env. variables we expect:
```golang
cmp.Diff(o.configSpec, tc.wantConfig)
```

danilo-gemoli (changes requested):

cmd/ci-operator/main_test.go
LiangquanLi930: The reason I did this is that I referred to TestMultiStageParams:
https://github.com/openshift/ci-tools/blob/33124116cffe24e0877b6fce4b593b6bf30d5f88/cmd/ci-operator/main_test.go#L1204-L1309

cmd/ci-operator/main.go
LiangquanLi930: An error occurs when the value of an environment variable contains MULTISTAGE_PARAM_OVERRIDE_ but the key does not contain MULTISTAGE_PARAM_OVERRIDE_.

cmd/ci-operator/main.go
LiangquanLi930: I adjusted the order of applyEnvOverrides and handleTargetAdditionalSuffix yesterday, but the E2E tests failed, so I reverted the order of these two functions.

PR #4499, opened 2025-02-17
Add retry feature to prowgen

jmguzik:


pkg/api/types.go
```diff
@@ -781,6 +782,9 @@ type TestStepConfiguration struct {
 	// AlwaysRun can be set to false to disable running the job on every PR
 	AlwaysRun *bool `json:"always_run,omitempty"`
 
+	// Retry is a configuration entry for re-retrying periodic prowjobs
```
smg247, line 785:
```suggestion
	// Retry is a configuration entry for retrying periodic prowjobs
```

pkg/api/types.go
```diff
@@ -781,6 +782,9 @@ type TestStepConfiguration struct {
...
+	Retry *prowconfig.Retry
```
smg247, line 786: We probably want an `omitempty` here

pkg/api/types.go
```diff
@@ -781,6 +782,9 @@ type TestStepConfiguration struct {
 	// AlwaysRun can be set to false to disable running the job on every PR
 	AlwaysRun *bool `json:"always_run,omitempty"`
 
+	// Retry is a configuration entry for reretrying periodic prowjobs
```
smg247, line 785:
```suggestion
	// Retry is a configuration entry for retrying periodic prowjobs
```

PR #4506, opened 2025-02-26
ci-operator: external_images stanza

Prucek:
This PR adds a new stanza to ci-operator `external_images`:
```
external_images:
  konflux:
    registry: quay.io
    namespace: rh-ee-prucek-tenant
    name: gnu-go-sort
    tag: latest
    pull_secret: my-pull-secret # must be in test-credetnials NS with the field .dockerconfigjson containing the auth
```

and images outside of our CI can be referenced like this. The image will be tagged as `pipeline:konflux`

```
./ci-operator --namespace=prucek-ci-op-debug --delete-when-idle=0 --delete-after=0 \                                                                            1 ✘  16s   16:16:56  
--config=/home/prucek/work/ci-tools/.vscode/test.yaml \
--lease-server-credentials-file=/tmp/ci-operator/secrets/boskos/credentials \
--gcs-upload-secret=/tmp/ci-operator/secrets/gcs/service-account.json \
--secret-dir=/tmp/ci-operator/secrets/ci-pull-credentials \
--image-import-pull-secret=/tmp/ci-operator/secrets/pull-secret/.dockerconfigjson \
--local-registry-dns=registry.build01.ci.openshift.org \
--manifest-tool-dockercfg=/tmp/ci-operator/secrets/manifest-tool-local-pusher/.dockerconfigjson \
--target=test
INFO[2025-03-06T14:49:51+01:00] unset version 0                              
INFO[2025-03-06T14:49:51+01:00] Resolved source https://github.com/openshift/ci-tools to master@fbbc82d4, merging: #1 fbbc82d4 @fake 
INFO[2025-03-06T14:49:53+01:00] Using namespace https://console.build01.ci.openshift.org/k8s/cluster/projects/prucek-debug 
INFO[2025-03-06T14:49:53+01:00] Running [input:konflux], test                
INFO[2025-03-06T14:49:55+01:00] Tagging quay.io/redhat-user-workloads/rh-ee-prucek-tenant/gnu-go-sort:on-pr-90ce48325d59958614cc221086c533e247d17354 into pipeline:konflux. 
INFO[2025-03-06T14:50:02+01:00] Running multi-stage test test                
INFO[2025-03-06T14:50:04+01:00] Running multi-stage phase pre                
INFO[2025-03-06T14:50:04+01:00] Step phase pre succeeded after 0s.           
INFO[2025-03-06T14:50:04+01:00] Running multi-stage phase test               
INFO[2025-03-06T14:50:04+01:00] Running step test-noop-test.                 
INFO[2025-03-06T14:50:12+01:00] Step test-noop-test succeeded after 8s.      
INFO[2025-03-06T14:50:12+01:00] Step phase test succeeded after 8s.          
INFO[2025-03-06T14:50:12+01:00] Running multi-stage phase post               
INFO[2025-03-06T14:50:12+01:00] Step phase post succeeded after 0s.          
INFO[2025-03-06T14:50:13+01:00] Ran for 21s      
```


pkg/validation/config.go
```diff
@@ -229,19 +229,21 @@ func (v *Validator) ValidateTestStepConfiguration(ctx *configContext, config *ap
...
 	if input == nil {
 		if hasImages {
-			return []error{errors.New("when 'images' are specified 'build_root' is required and must have image_stream_tag, project_image or from_repository set")}
+			return []error{errors.New("when 'images' are specified 'build_root' is required and must have image_stream_tag, project_image, from_repository or external_image set")}
 		}
 		return nil
 	}
 
-	if input.ProjectImageBuild != nil && input.ImageStreamTagReference != nil {
+	if input.ExternalImage != "" && (input.ProjectImageBuild != nil || input.ImageStreamTagReference != nil || input.FromRepository) {
```
danilo-gemoli, line 237: At a first glance it seems that these four branches:
```golang
if input.ExternalImage != "" && (input.ProjectImageBuild != nil || input.ImageStreamTagReference != nil || input.FromRepository) {
	ret = append(ret, ctx.errorf("external_image and project_image, image_stream_tag, from_repository are mutually exclusive"))
} else if input.ProjectImageBuild != nil && input.ImageStreamTagReference != nil {
	ret = append(ret, ctx.errorf("image_stream_tag and project_image are mutually exclusive"))
} else if input.ProjectImageBuild != nil && input.FromRepository {
	ret = append(ret, ctx.errorf("project_image and from_repository are mutually exclusive"))
} else if input.FromRepository && input.ImageStreamTagReference != nil {
	ret = append(ret, ctx.errorf("from_repository and image_stream_tag are mutually exclusive"))
```
can be merged into a single one:
```golang
mutuallyExclusive := []bool{input.ExternalImage != "", input.ProjectImageBuild != nil, input.ImageStreamTagReference != nil, input.FromRepository}
if len(slices.DeleteFunc(mutuallyExclusive, func(e bool) bool { return !e })) > 1 {
	ret = append(ret, ctx.errorf("external_image and project_image, image_stream_tag, from_repository are mutually exclusive"))
}
```

pkg/api/types.go
```diff
@@ -453,7 +453,8 @@ type BuildRootImageConfiguration struct {
 	ImageStreamTagReference *ImageStreamTagReference          `json:"image_stream_tag,omitempty"`
 	ProjectImageBuild       *ProjectDirectoryImageBuildInputs `json:"project_image,omitempty"`
 	// If the BuildRoot images pullspec should be read from a file in the repository (BuildRootImageFileName).
-	FromRepository bool `json:"from_repository,omitempty"`
+	FromRepository bool   `json:"from_repository,omitempty"`
+	ExternalImage  string `json:"external_image,omitempty"`
```
danilo-gemoli, line 457: In case of a private registry, where is the `ci-operator` going to get the credentials from?

pkg/steps/input_image_tag.go
```diff
@@ -72,14 +72,21 @@ func (s *inputImageTagStep) Run(ctx context.Context) error {
 }
 
 func (s *inputImageTagStep) run(ctx context.Context) error {
-	logrus.Infof("Tagging %s into %s:%s.", s.config.BaseImage.ISTagName(), api.PipelineImageStream, s.config.To)
-
+	var from *coreapi.ObjectReference
```
danilo-gemoli, line 75: Wouldn't it be better moving this whole block of new code right after this check?
```golang
if _, err := s.Inputs(); err != nil {
...
}
```

pkg/api/types.go
Prucek: well, for now, it will only work with public images, but we can add this, I was thinking about it, and I am not sure how tough. Is it a good idea a field `pullSecret: my-secret-in-vault` ?

droslean: Why not expand the ImageStreamTagReference to include an external registry instead?  If not, I also don't like having this entry as a string.

Prucek: I thought `ImageStreamTagReference` is for image stream tags and this is just an image, what else than a `string` would you like to see?

droslean: We can have a different object that wraps ImageStreamTagReference or something similar. Notice also that we also need to validate the values and ci-operator will try to tag it in. A string value is not that consistent. Let's consider having a proper object with the proper validation.

droslean: One more thing. Why we are doing this in the build_root? IIRC the whole idea is to make a usage of the Konflux root image that already have the source included. What am I missing here?

Prucek: You want to have it in the `images` stanza? this way it is easier to stop building anything. If I put the external image in the image stanza then I have to tell somehow not to build the build_root

droslean: We need to play with the ci-operator targets. The graph needs to be resolved based on that. You need to check first that behavior. 

What you actually want is to skip the root+src dependency and use an external image directly. This is more complicated since its not currently supported by ci-operator.

Prucek: WIth the external image beign in the build root, I have skipped them both, look at the log ^^^ it just tag the image and run the test

pkg/steps/input_image_tag.go
```diff
@@ -72,14 +75,44 @@ func (s *inputImageTagStep) Run(ctx context.Context) error {
...
+
+	var objectReferenceName string
+	if s.config.ExternalImage != nil {
+		logrus.Infof("Tagging %s into %s:%s.", externalImageReference(s.config), api.PipelineImageStream, s.config.To)
+		objectReferenceName = externalImageReference(s.config)
+
+		if s.config.ExternalImage.PullSecret != "" {
+			// Get the secret from the test-credentials namespace. We expect it exists
+			ciSecret := &coreapi.Secret{}
+			err := s.client.Get(ctx, ctrlruntimeclient.ObjectKey{Namespace: "test-credentials", Name: s.config.ExternalImage.PullSecret}, ciSecret)
```
droslean, line 90: Is this the right place to grab the secret? ci-operator has already the mechanisms of gathering any secret is required for the runtime. Perhaps it would be better to do it in one place and pass the secret to the steps.

pkg/api/types.go
```diff
@@ -171,17 +171,14 @@ func (config ReleaseBuildConfiguration) BuildsImage(name string) bool {
...
-			return true
-		}
-	}
-	return false
+	return hasKey(config.BaseImages, name) ||
+		hasKey(config.BaseRPMImages, name) ||
+		hasKey(config.ExternalImages, name)
+}
+
+func hasKey[T any](m map[string]T, key string) bool {
```
droslean, line 179: No need for that. You can use the sets library directly.

Prucek: I am not sure how. `ExternalImages` is of a different type (`map[string]ExternalImage`) compared to `BaseImages` (`map[string]ImageStreamTagReference`)

PR #4508, opened 2025-02-26
`payload-testing`: add the ability to skip jobs when running payload tests from a PR

smg247:
This ability is based on a couple of new environment variables of the form: "SKIP_JOB_REGEX_xyz" and "SKIP_JOB_EXPIRE_xyz". The jobs with names matching the provided regex will be skipped on *all* PRPQR runs until the provided expire time.

This functionality will allow us to skip arbitrary jobs matched by any regex in the future, and will not be enabled at all until a corresponding change is made to add the proper env vars to the plugin deployment. I will add that in a followup PR.

For: https://issues.redhat.com/browse/DPTP-4335

### DIFF
--- a/cmd/payload-testing-prow-plugin/rcjobresolver.go
+++ b/cmd/payload-testing-prow-plugin/rcjobresolver.go
@@ -45,7 +45,7 @@ func (r *releaseControllerJobResolver) resolve(ocp string, releaseType api.Relea
 	}
 	jobSkips, err := determineJobSkips(time.Now())
 	if err != nil {
-		return nil, fmt.Errorf("could not determine job skips: %v", err)
+		return nil, fmt.Errorf("could not determine job skips: %w", err)
 	}
 	if len(jobSkips) == 0 {
 		return jobs, nil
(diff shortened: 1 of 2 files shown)


cmd/payload-testing-prow-plugin/rcjobresolver.go
```diff
@@ -24,12 +32,99 @@ func (r *releaseControllerJobResolver) resolve(ocp string, releaseType api.Relea
...
 		},
 		Stream:  releaseType,
 		Version: ocp,
 	}, jobType)
+	if err != nil {
+		return nil, err
+	}
+	jobSkips, err := determineJobSkips(time.Now())
+	if err != nil {
+		return nil, fmt.Errorf("could not determine job skips: %w", err)
```
Prucek, line 48: Do you want to return `nil` if there is an error with the `jobSkips`? I don't know what happens with the resolved `jobs`, but the resolving was successful, so I would return `jobs` with the error.

smg247: I thought about that, but if we do that we will probably never notice that there is an error. The user won't complain because the jobs are running still. If we fail "hard" like this, we will be alerted fairly quickly.

Prucek:
/test e2e
should now run on b11

PR #4514, opened 2025-03-04
`prowgen`: Add enable_secrets_store_csi_driver field

psalajova:
Related to https://github.com/openshift/ci-tools/pull/4507

`enable_secrets_store_csi_driver` is a temporary option for ci-operator to enable us to test out the new CSI Secrets mechanism. This PR allows us to toggle it using the `.config.prowgen` files (which are in the release repo ci-operator/config dirs)

### DIFF
--- a/pkg/prowgen/jobbase.go
+++ b/pkg/prowgen/jobbase.go
@@ -160,6 +160,9 @@ func NewProwJobBaseBuilderForTest(configSpec *cioperatorapi.ReleaseBuildConfigur
 		if configSpec.Releases != nil {
 			p.PodSpec.Add(CIPullSecret())
 		}
+		if info.Config.EnableSecretsStoreCSIDriver {
+			p.PodSpec.Add(Arg("enable-secrets-store-csi-driver", "true"))
+		}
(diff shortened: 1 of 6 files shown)


pkg/prowgen/jobbase.go
```diff
@@ -160,6 +160,9 @@ func NewProwJobBaseBuilderForTest(configSpec *cioperatorapi.ReleaseBuildConfigur
 		if configSpec.Releases != nil {
 			p.PodSpec.Add(CIPullSecret())
 		}
+		if info.Config.EnableSecretsStoreCSIDriver {
+			p.PodSpec.Add(Arg("enable_secrets_store_csi_driver", "true"))
```
danilo-gemoli, line 164: `enable_secrets_store_csi_driver` it's ok as a JSON field, but as an argument name I'd rename it to `enable-secrets-store-csi-driver`

danilo-gemoli (changes requested):

pkg/prowgen/jobbase.go
psalajova: Yeah I just realized this while writing tests 😄 Fixed now

PR #4517, opened 2025-03-05
rework payload testing image tag overrides to allow for an aritrary pullspec override

smg247:
> It is uncommon for users to have the permission necessary to push test images to the ocp namespace or established imagestreams. Instead, permit payload tags to reference arbitrary images.

```
spec:
  payload:
    tags:
      - name: "machine-os-content"
        image: quay.io/openshift-release-dev/.... (arbitrary pullspec)
```

Note that this PR includes both changes for `ci-operator` and `prpqr_reconciller` as they need to merge together, and they are relatively simple on their own.

For: https://issues.redhat.com/browse/DPTP-4278

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -303,9 +303,14 @@ type InputConfiguration struct {
 
 // ExternalImage describes the external image that is imported into the pipeline
 type ExternalImage struct {
-	Registry                string `json:"registry"` // Registry is the registry to pull images from (e.g. quay.io)
+	// Registry is the registry to pull images from (e.g. quay.io)
+	Registry                string `json:"registry"`
 	ImageStreamTagReference `json:",inline"`
-	PullSecret              string `json:"pull_secret,omitempty"` // PullSecret is the name of the secret to use to pull the image
+	// PullSecret is the name of the secret to use to pull the image
+	PullSecret string `json:"pull_secret,omitempty"`
+	// PullSpec is the full pullSpec of the external image, only to be set programmatically,
+	// and takes precedent over the other fields in ExternalImage
+	PullSpec string `json:"pull_spec,omitempty"`
 }
 
 // UnresolvedRelease describes a semantic release payload
@@ -607,11 +612,6 @@ type StepConfiguration struct {
 type InputImageTagStepConfiguration struct {
 	InputImage `json:",inline"`
 	Sources    []ImageStreamSource `json:"-"`
-
-	// ExternalPullSpec allows for an arbitrary pullspec that does not have to come from QCI to be used as the image source.
-	// This functionality was created to allow for specific image tag content to be overridden during payload testing.
-	// This field should *not* be used to circumvent the standard mirroring process for QCI.
-	ExternalPullSpec string `json:"external_pull_spec,omitempty"`
 }
 
 func (config InputImageTagStepConfiguration) TargetName() string {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -607,6 +607,11 @@ type StepConfiguration struct {
 type InputImageTagStepConfiguration struct {
 	InputImage `json:",inline"`
 	Sources    []ImageStreamSource `json:"-"`
+
+	// ExternalPullSpec allows for an arbitrary pullspec that does not have to come from QCI to be used as the image source.
+	// This functionality was created to allow for specific image tag content to be overridden during payload testing.
+	// This field should *not* be used to circumvent the standard mirroring process for QCI.
+	ExternalPullSpec string `json:"external_pull_spec,omitempty"`
 }
 
 func (config InputImageTagStepConfiguration) TargetName() string {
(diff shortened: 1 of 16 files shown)


pkg/steps/input_image_tag.go
```diff
@@ -88,6 +88,14 @@ func (s *inputImageTagStep) run(ctx context.Context) error {
 			Namespace: s.config.BaseImage.Namespace,
 		}
 	}
+	if s.config.ExternalPullSpec != "" {
+		from = &coreapi.ObjectReference{
+			Namespace: s.jobSpec.Namespace(),
```
Prucek, line 93: why ns here? it is from an external source 🤔

pkg/defaults/defaults.go
```diff
@@ -380,27 +374,17 @@ func fromConfig(
 	return append(overridableSteps, buildSteps...), promotionSteps, nil
 }
 
-func stepsForImageOverrides(ctx context.Context, overriddenImages map[string]string, consoleHost string, client ctrlruntimeclient.Client, second time.Duration) []api.StepConfiguration {
+func stepsForImageOverrides(overriddenImages map[string]string) []api.StepConfiguration {
 	var overrideSteps []api.StepConfiguration
 	for tag, value := range overriddenImages {
-		istRef := api.ImageStreamTagReference{
```
Prucek, line 386: this is no longer used?

pkg/api/types.go
```diff
@@ -603,6 +603,11 @@ type StepConfiguration struct {
 type InputImageTagStepConfiguration struct {
 	InputImage `json:",inline"`
 	Sources    []ImageStreamSource `json:"-"`
+
+	// ExternalPullSpec allows for an arbitrary pullspec that does not have to come from QCI to be used as the image source.
+	// This functionality was created to allow for specific image tag content to be overridden during payload testing.
+	// This field should *not* be used to circumvent the standard mirroring process for QCI.
+	ExternalPullSpec string `json:"external_pull_spec,omitempty"`
```
Prucek, line 610: Have you considered adding `ExternalPullSpec` into `InputImage`? It is basically an `InputImage` just external... I am doing something very similar and I am also changing this step/step-config in a similar way

pkg/steps/input_image_tag.go
smg247: That is leftover from when I was doing this slightly differently. I guess it can be removed, it doesn't seem to have any effect either way.

pkg/defaults/defaults.go
smg247: Correct, we are no longer looking for images from the "ocp" NS for overrides.

pkg/api/types.go
smg247: No, because I specifically did not want this to be a field that someone would try to set in their config. It is not relevant to the functionality I am trying to achieve here. I am aware of the similarity of your ongoing work, and we should certainly use as much common functionality as we can between the two features. If you are planning on adding it to `InputImage`, you can refactor this functionality to put it there at that time.

pkg/defaults/defaults.go
```diff
@@ -410,7 +403,31 @@ func stepsForImageOverrides(ctx context.Context, overriddenImages map[string]str
...
 	}
-	return overrideSteps
+	return overrideSteps, nil
+}
+
+// externalImageFromPullSpec creates an ExternalImage from a pullSpec (like: "quay.io/openshift-release-dev/ocp-v4.0-art-dev@sha256:9a49368aad56c984302c3cfd7d3dfd3186687381ca9a94501960b0d6a8fb7f98")
+// This is intended to be used for image tag override purposes.
+func externalImageFromPullSpec(pullSpec string) (*api.ExternalImage, error) {
+	split := strings.Split(pullSpec, "/")
+	if len(split) != 3 {
```
Prucek, line 413: Will the images be only from this registry: `quay.io/openshift-release-dev/ocp-v4.0-art-dev:latest`? The konflux registry is ` quay.io/redhat-user-workloads/multiarch-tuning-ope-tenant/multiarch-tuning-operator/multiarch-tuning-operator:latest ` I wanted to say that splitting in 3 must not always be correct as the NS can have additional "/"

pkg/steps/input_image_tag.go
```diff
@@ -98,6 +98,7 @@ func (s *inputImageTagStep) run(ctx context.Context) error {
 			Namespace: s.config.BaseImage.Namespace,
 		}
 	}
+
```
Prucek, line 101: Is this on purpose?

pkg/defaults/defaults.go
smg247: It is supposed to be from an arbitrary registry, but in practice it will likely always be art images. 

It is kind of unfortunate that I am decomposing a pullspec just so that it can be re-composed later. I wonder if it makes more sense to add a new field to ExternalImage that represents a pullspec and can be set mutually exclusively to the existing fields. Thoughts?

Prucek: Yes, it didn't make sense to me as well. My original idea was to have only a string `name:` but Nikos was not a fan of it, and proposed to have this for better validation https://github.com/openshift/ci-tools/pull/4506#discussion_r1973574547

Prucek (approved):

PR #4518, opened 2025-03-06
`pod-scaler`: temporarily use 120% of what we have determined necessary during admission

smg247:
This is a temporary experiment to see what effect setting values that are 120% of what has been determined has on the rate of OOMKilled and similar termination of workloads. There have been many complaints about workloads being killed recently on our forum. As a precursory measure before https://issues.redhat.com/browse/DPTP-4225 We will see what empirical effect this has on these issues.


smg247:
overrode e2e due to unrelated flake

PR #4523, opened 2025-03-11
Updated branches references from "master" to "main"

hector-vido:
Various references of `master` branch was renamed to `main`, this include:

- The minimum tests possible;
- References to `openshift/release@master` and;
- Default branch names used by some tools.


Prucek:
This change needs to be announced to the public and it should be merged right after you rename the branches on o/release and o/ci-tools. I would even announce it beforehand, like branching, on XX if march we will rename branches, expect delay/outage for X hours

jmguzik (changes requested):
I think this PR should be the focus first.
I also think it is not as straightforward as it looks like in this PR.

Whenever "master" is missing in code, it should not be replaced with "main", rather "main" should be added as an option to treat "main" and "master" equally. It is the only path towards smooth transition.

cmd/autoperibolossync/main.go
```diff
@@ -144,9 +144,9 @@ func main() {
 	}
 
 	// TODO fix the bumper in upstream to retrieve a default branch
-	defaultBranch := "master"
+	defaultBranch := "main"
```
jmguzik, line 147: This as an example, main should be equal option in the code, not just replacement

hector-vido: That was my bad, now I just swap to check for `main` first and them try `master`.

PR #4531, opened 2025-03-19
Optional promotion periodic

jmguzik:
forum followup, not to give away gangway tokens for some hacky solution which will run postsubmits periodically in external loops


Prashanth684:
any reason why this is on hold?

danilo-gemoli:
@Prashanth684 it's fine to me and @jmguzik can unhold it whenever he feels confident.

PR #4549, opened 2025-04-08
Konflux: Fetch EphemeralCluster kubeconfig

danilo-gemoli:
The `EphemeralCluster` has been requested, and the controller spawned a `ci-operator` instance that is provisioning an ephemeral cluster.

As soon as the cluster is ready, `ci-operator` stores an admin kubeconfig into `secret/wait-test-complete`, as per design.
This reconciler will then retrieve it by following these steps in the `func fetchKubeconfig`:
- Identify the cluster in which `ci-operator` is running by looking at the `.spec.cluster` stanza on the ProwJob.
- Identify the `ci-op-xxxx` test NS by matching any namespace against this label: `ci.openshift.io/jobid == <PJ_ID_HERE>`.
- Get the `kubeconfig` from the `wait-test-complete` and store it into the `EphemeralCluster`'s `.status.kubeconfig` field.

### DIFF
--- a/pkg/controller/ephemeralcluster/reconciler.go
+++ b/pkg/controller/ephemeralcluster/reconciler.go
@@ -242,7 +242,7 @@ func (r *reconciler) uploadCIOperatorConfig(ctx context.Context, config *api.Rel
 func (r *reconciler) fetchKubeconfig(ctx context.Context, log *logrus.Entry, ec *ephemeralclusterv1.EphemeralCluster, pj *prowv1.ProwJob) (reconcile.Result, error) {
 	buildClient, ok := r.buildClients[pj.Spec.Cluster]
 	if !ok {
-		err := fmt.Errorf("unkown cluster %s", pj.Spec.Cluster)
+		err := fmt.Errorf("uknown cluster %s", pj.Spec.Cluster)
 		r.upsertCondition(ec, ephemeralclusterv1.ClusterReady, ephemeralclusterv1.ConditionFalse, ephemeralclusterv1.KubeconfigFetchFailureReason, err.Error())
 		return reconcile.Result{}, reconcile.TerminalError(err)
 	}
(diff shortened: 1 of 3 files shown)


pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -232,6 +239,56 @@ func (r *reconciler) uploadCIOperatorConfig(ctx context.Context, config *api.Rel
 	return r.configUploader.UploadConfigSpec(ctx, gcsPath, string(configBytes))
 }
 
+func (r *reconciler) fetchKubeconfig(ctx context.Context, log *logrus.Entry, ec *ephemeralclusterv1.EphemeralCluster, pj *prowv1.ProwJob) (reconcile.Result, error) {
+	buildClient, ok := r.buildClients[pj.Spec.Cluster]
+	if !ok {
+		err := fmt.Errorf("unkown cluster %s", pj.Spec.Cluster)
```
liangxia, line 245: lint fails,
```
pkg/controller/ephemeralcluster/reconciler.go:245:22: `unkown` is a misspelling of `unknown` (misspell)
		err := fmt.Errorf("unkown cluster %s", pj.Spec.Cluster)
		                   ^
```

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -51,7 +53,8 @@ type reconciler struct {
 	configUploader ConfigSpecUploader
 
 	// Mock for testing
-	now func() time.Time
+	now     func() time.Time
+	polling func() time.Duration
```
Prucek, line 57: Do we want to set a value for polling in production?

danilo-gemoli: I have yet to understand if this is going to be a fixed value, in this case a plain `time.Duration` is enough, or computed on the fly, therefore `func`. The controller might have to check, potentially, any secret from any `ci-op-xxxx` NS from any cluster, so there are two solutions for this coming to my mind:
1. Leverage `client.Watch` and setup a persistent HTTP connection to each cluster
2. Polling

The first approach is quite resource intensive as the `Manager` receives and stores a lot of objects into its cache, therefore I have opted for the second one. Nothing it's written into the stone, if this turns out not working well I'll go another direction.

Prucek (approved):

danilo-gemoli:
/override ci/prow/e2e
flake

PR #4550, opened 2025-04-08
release: add balance profile cmd (again)

danilo-gemoli:
The original PR https://github.com/openshift/ci-tools/pull/4406 has been closed and I don't remember the reason why anymore.

The release tool acquires a new feature, that is the ability to redistribute cluster profiles used by ci-operator tests.
Quoting what is already stated on a comment:
```
 The following incantation:

	--from=gcp --scaling-factor=0.5 --to=gcp2,gcp3 --balancing-factors=0.75,0.25 --exclude-orgs=openshift-priv

 gather every ci-operator config, excluding the ones under openshift-priv/, having at least
 one test that has `cluster_profile: gcp` set.
 Assuming the selected tests to be N = 100, it then redistributes R = N*0.5 = 50 of them.

 The balancing algorithm follows this schema:

	tests that will target 'gcp2': R * 0.75 = 37 => adjusted to 38 to compensate truncation errors
	tests that will target 'gcp3': R * 0.25 = 12

 After the distribution has taken place:

	'gcp'  tests: 50
	'gcp2' tests: +38
	'gcp3' tests: +12
```

### DIFF
--- a/cmd/release/main.go
+++ b/cmd/release/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"flag"
+	"fmt"
 	"os"
 
 	"github.com/spf13/pflag"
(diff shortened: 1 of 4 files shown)


deepsm007:
We need to consider the cluster-profiles based on https://github.com/openshift/release/blob/master/ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml

PR #4551, opened 2025-04-08
Adjust the version naming scheme to comply with the semver specification

LiangquanLi930:


droslean:
@LiangquanLi930 Can you share some details about this change? When the prefix was changed? Was it always wrong?

LiangquanLi930:
> @LiangquanLi930 Can you share some details about this change? When the prefix was changed? Was it always wrong?

I think it has always been incorrect — the image version used by the integration build should be greater than the other versions, e.g.
```
  initial:
    candidate:
      product: ocp
      stream: ci
      version: "4.19"
  latest:
    integration:
      include_built_images: true
      name: 4.19-priv
      namespace: ocp-private
```
When comparing using semver.Parse(), it turns out that release:initial (4.19.0-0.ci-2025-03-30-173758)  > release:latest (4.19.0-0.ci.test-2025-03-31-075924-ci-op-stt4mr8g-latest)
In semantic version comparisons, numeric identifiers always take precedence over alphabetic ones, which is why I submitted this PR.

droslean:
overriding the e2e job since the multi-stage test case passed and the pod-scaler flaked which is not related to the change.

/override ci/prow/e2e

PR #4552, opened 2025-04-10
OCPEDGE-1741: feat: add edge enablement equinix profile

eggfoobar:
added edge enablement equinix cluster profile for the time being to move metal sno workloads off of other teams accounts, this is a short term solution for now until we fully migrate all jobs off of equinix later this year.


deepsm007:

the failure is unrelated to the change

PR #4553, opened 2025-04-10
Konflux: Support EphemeralCluster Deletion

danilo-gemoli:
Introducing some features:
1. Report the status of a `ProwJob` as a condition to the `EphemeralCluster`. As an example: if a PJ is in a `success` state, the following condition will be added to the `EphemeralCluster`:
```yaml
conditions:
- type: ProwJobCompleted,
  status: True
  reason: ProwJobCompleted
  message: "prowjob state: success",
  lastTransitionTime: "2025-04-10T10:23:00Z"
```
3. Add a finalizer to the `EphemeralCluster` when its related `ProwJob` has been created, so to make sure it won't be deleted as long as the PJ exists.
4. Abort `ProwJob` when the `EphemeralCluster` deletion has been requested.
5. Add a new reconciler to handle a corner case: abort a `ProwJob` when its related `EphemeralCluster` doesn't exist anymore.

### DIFF
--- a/pkg/controller/ephemeralcluster/reconciler.go
+++ b/pkg/controller/ephemeralcluster/reconciler.go
@@ -5,6 +5,7 @@ import (
 	_ "embed"
 	"errors"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/ghodss/yaml"
@@ -31,8 +32,10 @@ import (
 )
 
 const (
-	WaitTestStepName = "wait-test-complete"
-	ProwJobNamespace = "ci"
+	WaitTestStepName          = "wait-test-complete"
+	ProwJobNamespace          = "ci"
+	EphemeralClusterNameLabel = "ci.openshift.io/ephemeral-cluster-name"
+	EphemeralClusterNamespace = "konflux-ephemeral-cluster"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -118,10 +131,14 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 	return reconcile.Result{}, nil
 }
 
-func (r *reconciler) handleGetProwJobError(err error) (reconcile.Result, error) {
+func (r *reconciler) handleGetProwJobError(ctx context.Context, ec *ephemeralclusterv1.EphemeralCluster, err error) (reconcile.Result, error) {
 	if kerrors.IsNotFound(err) {
-		// TODO
-		return reconcile.Result{}, reconcile.TerminalError(errors.New("unimplemented"))
+		finalizers, removed := removeElement(ec.Finalizers, DependentProwJobFinalizer)
```
jmguzik, line 136: probably `DependentProwJobFinalizer` defined in this file?

jmguzik (commented):
It should have more logging. Logger is passed but barely used.

jmguzik:
I'll lgtm but think about the comments

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: Yes, correct. I'll address it in an upcoming PR as I already have an LGTM on this.

danilo-gemoli:
Sure I can log more, no problem with that, but any important event is also reported as a condition to the `EphemeralCluster` object. Ideally, it would be nice to understand what's going on by running this:
```sh
$ oc get ephemeralcluster/foobar -o yaml | yq .status.conditions
```

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: Nevermind, I had to push something else anyway.

PR #4563, opened 2025-04-24
`prowgen`: add `shard_count` field to the test step configuration

smg247:
If provided, this configuration will result in multiple jobs being generated for the ci-op config test. Each job will receive unique SHARD_ARGS values that will flow to the workflow.

This will simplify the sharding of jobs. Currently, we must do something like https://github.com/openshift/release/pull/64180, and duplicate the test configuration.

There will be follow-up PR(s) to `release` that will begin to utilize the `SHARD_ARGS` in the necessary workflow(s).

For: https://issues.redhat.com/browse/TRT-2086

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -826,6 +826,11 @@ type TestStepConfiguration struct {
 	// RestrictNetworkAccess restricts network access to RedHat intranet.
 	RestrictNetworkAccess *bool `json:"restrict_network_access,omitempty"`
 
+	// ShardCount describes the number of jobs that should be generated as shards for this test
+	// Each generated job will be a duplication, but contain a suffix and the necessary SHARD_ARGS will be passed to the steps
+	// Only applicable to presubmits and periodics
+	ShardCount *int `json:"shard_count,omitempty"`
+
 	// Only one of the following can be not-null.
 	ContainerTestConfiguration                                *ContainerTestConfiguration                                `json:"container,omitempty"`
 	MultiStageTestConfiguration                               *MultiStageTestConfiguration                               `json:"steps,omitempty"`
--- a/pkg/api/zz_generated.deepcopy.go
+++ b/pkg/api/zz_generated.deepcopy.go
@@ -2184,6 +2184,11 @@ func (in *TestStepConfiguration) DeepCopyInto(out *TestStepConfiguration) {
 		*out = new(bool)
 		**out = **in
 	}
+	if in.ShardCount != nil {
+		in, out := &in.ShardCount, &out.ShardCount
+		*out = new(int)
+		**out = **in
+	}
 	if in.ContainerTestConfiguration != nil {
 		in, out := &in.ContainerTestConfiguration, &out.ContainerTestConfiguration
 		*out = new(ContainerTestConfiguration)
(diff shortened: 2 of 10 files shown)


deepsm007:
- Does it make sense to have the sharded jobs use same cluster and run tests against it rather than spinning up new ones?  
- What if a test is dependent on each other? Assuming they are handled in origin repo correctly to be mapped as 1 batch? - Maybe 
- I believe this is meant mainly for serial tests which takes a lot of time? - Maybe/Yes
-  Would this be increasing lease count and resource as well for spinning up new clusters for testing? - Yes

stbenjam:
> Does it make sense to have the sharded jobs use same cluster and run tests against it rather than spinning up new ones?

origin can already run tests in parallel against a cluster.  There are serial and disruptive suites that can only execute 1 test at a time, and some of those jobs take far too long. 

https://docs.google.com/document/d/14uXl_JF-n7WvJj7xQOmn7wRm71LM7Fbuw-2QnP_-bTs/edit?tab=t.0#heading=h.66y4kqbj468a

and https://github.com/openshift/origin/pull/29696 add some more context

> What if a test is dependent on each other? Assuming they are handled in origin repo correctly to be mapped as 1 batch? 

This isn't something we support, a test has to be fully independent

> I believe this is meant mainly for serial tests which takes a lot of time? - Maybe/Yes

Largely, yes but parallel suites could also be sharded so you could go from 30 -> 60 tests at a time.

> Would this be increasing lease count and resource as well for spinning up new clusters for testing? 

Yes, it will

stbenjam:
Would it make sense to have a `shard_strategy` option? The only sharder currently is hashing, and remarkably the serial jobs are time balanced, but as we integrate QE tests we might find we'll need to implement the time-balanced sharder

stbenjam:
Yea, it's a command-line argument `--shard-strategy`.  There's only one now so seems OK to wait until its needed

pkg/api/types.go
```diff
@@ -826,6 +826,11 @@ type TestStepConfiguration struct {
 	// RestrictNetworkAccess restricts network access to RedHat intranet.
 	RestrictNetworkAccess *bool `json:"restrict_network_access,omitempty"`
 
+	// ShardCount describes the number of jobs that should be generated as shards for this test
+	// Each generated job will be a duplication, but contain a suffix and the necessary SHARD_ARGS will be passed to the steps
+	// Only applicable to presubmits and periodics
+	ShardCount *int `json:"shard_count,omitempty"`
```
droslean, line 832: I don't think that this is the right place. The shard count, is a prowgen configuration. Since ci-operator is not related here, it should be better to allow the user to control the extra prowgen behavior for the job generation.
Perhaps in the `.config.prowgen`:
```yaml
jobs:
  test-job:
     shard_count: 2
     other_config: ...
```

smg247: I was expecting you would say that, but many of the features that we have added recently have been implemented in the same manner. [Retry](https://github.com/openshift/ci-tools/blob/1960063e9e900f4ca3665742fbbad18529b7bf97/pkg/api/types.go#L801), for instance. I considered adding this to the prowgen config, but the implementation would be considerably more complex, and more difficult to use.

smg247: I think that if the prowgen config were redesigned to be more "configurer"-friendly, it would be a better selling point to move this there. That is definitely a larger effort though. IMO, leaving it here until that takes place is the best move.

PR #4569, opened 2025-04-30
Bumping e2e ocp versions

hector-vido:


deepsm007:
You missed in couple of other tests

PR #4570, opened 2025-04-30
`ci-operator`: replace underscore in build name 

smg247:
When building from a ref that contains `_` in the repo name, we must replace it in order to name the build resource. The name of the build itself doesn't affect anything that relies on it, as it still has the same `Output` configured.

I have tested this by firing off a manual [job](https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-node_exporter-165-ci-4.19-upgrade-from-stable-4.18-e2e-aws-ovn-upgrade/1917627205832151040) using a locally built image.


deepsm007:
/lgtm
/override ci/prow/e2e 
e2e failure is being worked on in https://github.com/openshift/ci-tools/pull/4569

PR #4599, opened 2025-06-10
Update credentials `mount_path` verification

psalajova:
This change allows for the same mount path if credentials are in the same collection and have different names. With the new GCP mechanism, this is an allowed use case:
```
credentials:
  - collection: team1
    name: name
    mount_path: /var/run
  - collection: team1
    name: diff-name
    mount_path: /var/run
```
We still enforce that mount paths be different for different collections, to avoid possible collisions of file names.

### DIFF
--- a/pkg/validation/test.go
+++ b/pkg/validation/test.go
@@ -808,10 +808,10 @@ func validateCredentials(fieldRoot string, credentials []api.CredentialReference
 		for j, other := range credentials[i+1:] {
 			index := i + j + 1
 			if credential.MountPath == other.MountPath {
-				if credential.Collection == other.Collection && credential.Name != other.Name {
+				if credential.Name != other.Name {
 					continue
 				}
-				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s), but are in different collections", fieldRoot, i, index, credential.MountPath))
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s) and have the same name, which would result in a collision", fieldRoot, i, index, credential.MountPath))
 				continue
 			}
 			// we can make a couple of assumptions here to improve our check:
--- a/pkg/validation/test.go
+++ b/pkg/validation/test.go
@@ -808,7 +808,10 @@ func validateCredentials(fieldRoot string, credentials []api.CredentialReference
 		for j, other := range credentials[i+1:] {
 			index := i + j + 1
 			if credential.MountPath == other.MountPath {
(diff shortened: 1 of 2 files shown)


pkg/validation/test.go
```diff
@@ -808,7 +808,10 @@ func validateCredentials(fieldRoot string, credentials []api.CredentialReference
 		for j, other := range credentials[i+1:] {
 			index := i + j + 1
 			if credential.MountPath == other.MountPath {
-				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s)", fieldRoot, i, index, credential.MountPath))
+				if credential.Collection == other.Collection && credential.Name != other.Name {
+					continue
+				}
+				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s), but are in different collections", fieldRoot, i, index, credential.MountPath))
```
danilo-gemoli, line 814:
```yaml
credentials:
  - collection: collection-a
    name: a1
    mount_path: /mnt/secrets
  - collection: collection-b
    name: b1
    mount_path: /mnt/secrets
```
This is not passing the validation, right? What's wrong with this? Secrets are not colliding, do we want just to enforce a 1-to-1 mapping between the collection and the mount point?

pkg/validation/test.go
```diff
@@ -808,7 +808,10 @@ func validateCredentials(fieldRoot string, credentials []api.CredentialReference
 		for j, other := range credentials[i+1:] {
 			index := i + j + 1
 			if credential.MountPath == other.MountPath {
-				errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s)", fieldRoot, i, index, credential.MountPath))
+				if credential.Name != other.Name {
```
danilo-gemoli, line 811:
```golang
if credential.MountPath == other.MountPath && credential.Name == other.Name {
    errs = append(errs, fmt.Errorf("%s.credentials[%d] and credentials[%d] mount to the same location (%s) and have the same name, which would result in a collision", fieldRoot, i, index, credential.MountPath))
    continue
}
// ...
// validate even more here
```
Does this make sense? Store any error and continue to the next iteration or keep going with the validation.

psalajova: This case would fail:
```
credentials:
  - collection: collection-a
    name: a1
    mount_path: /mnt/secrets
  - collection: collection-a
    name: b1
    mount_path: /mnt/secrets
```
bacause the later code checks if the the `mount_path`s aren't subdirectories of each other, specifically see [this test](https://github.com/openshift/ci-tools/pull/4599/files#diff-2c04428953e5c63ac9f66bfd5fffa4a113f659cd08214c734ab116c05c8c3880R1295-R1301)

PR #4611, opened 2025-06-25
[WIP] - Add code to execute pre-submits by tag

kasturinarra:


kasturinarra:
> @kasturinarra Can you elaborate more? On the first look I don't think that this feature makes any sense for the rehearse plugin. This tool is not supposed to rehearse jobs provided by the user. Let's try to figure out what is needed first and how we can implement it.

@droslean here is what i am trying to achieve using the PR here,  Improve this by introducing a command in github like /pj-rehearse {tag-name}  where {tag-name} could be something like TNF. That command would look up the relevant available TNF tests and try to schedule rehearsals of the ones that match.

kasturinarra:
> > @kasturinarra Can you elaborate more? On the first look I don't think that this feature makes any sense for the rehearse plugin. This tool is not supposed to rehearse jobs provided by the user. Let's try to figure out what is needed first and how we can implement it.
> 
> @droslean here is what i am trying to achieve using the PR here, Improve this by introducing a command in github like /pj-rehearse {tag-name} where {tag-name} could be something like TNF. That command would look up the relevant available TNF tests and try to schedule rehearsals of the ones that match.

@droslean I think there might be a misunderstanding about what this feature does. This is NOT about allowing users to rehearse arbitrary jobs. 

The tag-based selection works WITHIN the existing rehearsal system:

1. **Existing flow**: PR changes → Rehearsal intelligence determines affected jobs → User runs `/pj-rehearse {specific-job-names}`
2. **Enhanced flow**: PR changes → Rehearsal intelligence determines affected jobs → User runs `/pj-rehearse-by-tag {tag}` → System filters affected jobs by tag

The key point is that we're only adding tag-based filtering to jobs that the rehearsal system has already determined are affected by the PR changes. We're not bypassing any existing safety mechanisms.

**Problem we're solving**: With 81k+ presubmit jobs, developers can't discover which tests are relevant to their changes. Tags like `tnf`, `telco`, `storage` help group related tests for better discoverability.

**Example scenario**: 
- Developer makes changes to telco-related code
- Rehearsal system identifies 50 affected jobs
- Instead of needing to know specific job names, developer can run `/pj-rehearse-by-tag telco`
- System runs only the telco-tagged jobs from those 50 affected jobs

This maintains all existing safety while improving usability. Does this clarify the intent?

cmd/pj-rehearse/server.go
```diff
@@ -655,6 +676,84 @@ func (s *server) acknowledgeRehearsals(org, repo string, number int, logger *log
...
+	}
+
+	var periodicSlice []prowconfig.Periodic
+	for _, p := range periodics {
+		periodicSlice = append(periodicSlice, p)
+	}
+	presubmits = rehearse.FilterPresubmitsByTag(presubmits, periodicSlice, s.tagConfig, tag)
+
+	if len(presubmits) > 0 {
+		prConfig, prRefs, presubmitsToRehearse, err := rc.SetupJobs(candidate, candidatePath, presubmits, nil, math.MaxInt, logger)
```
deepsm007, line 729: `math.MaxInt` concerns us

PR #4614, opened 2025-07-01
Do not skip current release periodics when config-branching on o/release

jmguzik:


jmguzik:
/override ci/prow/e2e

we do not need that for this pr

PR #4619, opened 2025-07-03
Fix CSI credentials naming mechanism

psalajova:
As of now, the new CSI mutli-stage step credential system creates one SecretProviderClass (SPC) per credential, mounted to the test pod in as a CSI `volumeMount`. Since we need to shard the existing credential secrets, like this:
```
credentials:
  - collection: team1
    name: name
    mount_path: /var/run
  - collection: team1
    name: diff-name
    mount_path: /var/run
```
the existing system resulted in Kubernetes's `Invalid value: "<mount_path>": must be unique` errors (see e.g. [this test run](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/65850/rehearse-65850-periodic-ci-stackrox-stackrox-master-ocp-4-12-merge-qa-e2e-tests-psalajov-test/1941045013337083904#1:build-log.txt%3A202)), because even if this results in a volumeMount with different name, the `mount_path`s cannot be the same.
Also, because the credential's name can be quite long, we were hitting character limit exceeded errors.

**Solution:**
- Credentials with the same (collection, mountPath) are now grouped into a single SPC 
- SPC names are based on a hash of collection, mount path, and credential names to prevent overwrites when different steps use the same collection/mount combination
- Truncation logic -- to stay within Kubernetes' 63-character DNS label limit -- was used for SPC and CSI volumes naming
- For sidecar censoring, all credentials are created as individual SPCs, with different names and mount paths

### DIFF
--- a/pkg/steps/multi_stage/init.go
+++ b/pkg/steps/multi_stage/init.go
@@ -7,7 +7,6 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/GoogleCloudPlatform/secrets-store-csi-driver-provider-gcp/config"
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
@@ -16,7 +15,6 @@ import (
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	csiapi "sigs.k8s.io/secrets-store-csi-driver/apis/v1"
-	"sigs.k8s.io/yaml"
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/kubernetes"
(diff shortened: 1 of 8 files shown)


pkg/steps/multi_stage/init.go
```diff
@@ -88,58 +84,53 @@ func (s *multiStageTestStep) createCredentials(ctx context.Context) error {
...
+	}
+
+	// Create SPCs for sidecar censoring; each credential gets its own SPC.
+	logrus.Infof("Creating SPCs for sidecar censoring...")
+	seenCredentials := make(map[string]bool)
 	for _, step := range append(s.pre, append(s.test, s.post...)...) {
 		for _, credential := range step.Credentials {
-			name := fmt.Sprintf("%s-%s-spc", s.jobSpec.Namespace(), credential.Name)
-			if _, exists := toCreate[name]; exists {
+			if seenCredentials[credential.Name] {
```
danilo-gemoli, line 109: I've noticed that this check happens in several places. If we forbid two credentials with the same name, wouldn't it be appropriate to add this policy to the code that validates a ci-operator configuration? [pkg/validation/config.go](https://github.com/openshift/ci-tools/blob/c291f916d7c3da878c0fc59ea3574d82acf7b81b/pkg/validation/config.go)

psalajova: This check is for the scenario when there's more steps in a workflow which use the same credentials. E.g. this [workflow](https://steps.ci.openshift.org/workflow/stackrox-automation-flavors-aks-e2e) -- has steps [stackrox-stackrox-begin](https://steps.ci.openshift.org/reference/stackrox-stackrox-begin) and [stackrox-stackrox-end](https://steps.ci.openshift.org/reference/stackrox-stackrox-end), and both have the `stackrox-automation-flavors` credential in their definitions.

PR #4623, opened 2025-07-08
Konflux: Improve reconcile loop

danilo-gemoli:
Several small improvements:
- Replace the `ClusterProvisioning` condition with the more explicit `ProwJobCreating`. Furthermore, the `ProwJobCreating` is "stable" in a sense that the condition stays there even after the ProwJob has been created. It wasn't the case for `ClusterProvisioning` and this situation wasn't handled properly by Crossplane, leading to the following error on the XR:
```
- lastTransitionTime: "2025-07-08T09:35:55Z"
  message: A fatal error occurred before the status of this condition could be determined.
  reason: FatalError
  status: Unknown
  type: ClusterProvisioning
```
- Do not use `RequeueAfter` together with an error
- Calling `DeepCopy` isn't required anymore, see [UnsafeDisableDeepCopy](https://github.com/kubernetes-sigs/controller-runtime/blob/bfd1cf925238323fd53b372780a3d0a1a750a85c/pkg/cache/cache.go#L321)

### DIFF
--- a/pkg/controller/ephemeralcluster/prowjobreconciler.go
+++ b/pkg/controller/ephemeralcluster/prowjobreconciler.go
@@ -72,8 +72,6 @@ func (r *prowJobReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, reconcile.TerminalError(fmt.Errorf("%s doesn't have the EC label", pj.Name))
 	}
 
-	pj = pj.DeepCopy()
-
 	ec := ephemeralclusterv1.EphemeralCluster{}
 	if err := r.client.Get(ctx, types.NamespacedName{Name: ecName, Namespace: EphemeralClusterNamespace}, &ec); err != nil {
 		if kerrors.IsNotFound(err) {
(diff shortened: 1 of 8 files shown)


pkg/controller/ephemeralcluster/prowjobreconciler.go
```diff
@@ -72,8 +72,6 @@ func (r *prowJobReconciler) Reconcile(ctx context.Context, req reconcile.Request
 		return reconcile.Result{}, reconcile.TerminalError(fmt.Errorf("%s doesn't have the EC label", pj.Name))
 	}
 
-	pj = pj.DeepCopy()
```
jmguzik, line 75: why this is removed?

danilo-gemoli: The controller runtime internal cache is already making a copy of the object (see [this](https://github.com/kubernetes-sigs/controller-runtime/blob/bbcaf6a9d9ff0ba256e18800f194d387b96dca33/pkg/cache/internal/cache_reader.go#L84-L90)). This wasn't the case in the older versions but we use a recent one now, therefore we can rely on that behavior.

PR #4644, opened 2025-08-08
[ci-operator] include the --keep-manifest-list option on release payload generation

droslean:
Adds the `--keep-manifest-list` flag to the `oc adm release new` command. 
Create a helper function to generate the command and unittests added.

The `--keep-manifest-list` flag will be added only for releases higher than 4.11.

The version was updated on the e2e tests as well. 

I also removed all template e2e test cases instead of just fixing the release version.

/cc @openshift/test-platform @bradmwilliams


bradmwilliams (approved):
/lgtm

droslean:
/hold

This change will break all <4.14 jobs that are using a release stanza. I will make some investigation to see which jobs are still using them, and update the code to distinguish this case.

PR #4649, opened 2025-08-15
Allow special pods to get NET_ caps for diagnostic purposes

jupierce:
TRT is working on understanding network outages detected during e2e tests. To facilitate this, we want to do network capture when the condition is detected.

It is intended to be paired with: https://github.com/openshift/release/pull/68174

### DIFF
--- a/cmd/ci-scheduling-webhook/mutation.go
+++ b/cmd/ci-scheduling-webhook/mutation.go
@@ -185,6 +185,73 @@ func mutatePod(w http.ResponseWriter, r *http.Request) {
 	}
 
 	klog.Infof("Pod %s in namespace %s is classified as %s", podName, namespace, podClass)
+
+	// Add NET_ADMIN and NET_RAW capabilities to test containers that require them
+	if podClass == PodClassTests {
+		// Check each container for the environment variable
+		for i, container := range pod.Spec.Containers {
+			if container.Name == "test" {
+				// Check if this container has TEST_REQUIRES_BUILDFARM_NET_ADMIN=true
+				requiresNetAdmin := false
+				for _, env := range container.Env {
+					if env.Name == "TEST_REQUIRES_BUILDFARM_NET_ADMIN" && env.Value == "true" {
+						requiresNetAdmin = true
+						break
+					}
+				}
+
+				if requiresNetAdmin {
+					// Build the correct patch based on existing securityContext
+					if container.SecurityContext == nil {
+						// No securityContext exists, create one with capabilities
+						securityContext := map[string]interface{}{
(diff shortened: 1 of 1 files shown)


cmd/ci-scheduling-webhook/mutation.go
```diff
@@ -185,6 +185,73 @@ func mutatePod(w http.ResponseWriter, r *http.Request) {
...
+						} else {
+							addPatchEntry("add", fmt.Sprintf("/spec/containers/%d/securityContext/capabilities/add", i), existingCaps)
+						}
+					}
+
+					klog.Infof("Added NET_ADMIN and NET_RAW capabilities to test container in pod %s in namespace %s due to TEST_REQUIRES_BUILDFARM_NET_ADMIN=true", podName, namespace)
+				}
+				break
+			}
+		}
```
hector-vido, line 252: I think we can use `k8s.io/api/core/v1`  `Pod` definition and the child properties to avoid generic interfaces and shrink the code a little:

```go
for i, container := range pod.Spec.Containers {
    if container.Name == "test" {
        // Check if this container has TEST_REQUIRES_BUILDFARM_NET_ADMIN=true
        requiresNetAdmin := false
        for _, env := range container.Env {
            if env.Name == "TEST_REQUIRES_BUILDFARM_NET_ADMIN" && env.Value == "true" {
                requiresNetAdmin = true
                break
            }   
        }   

        if requiresNetAdmin {
            // Build the correct patch based on existing securityContext
            if container.SecurityContext == nil {
                // No securityContext exists, create one with capabilities
                container.SecurityContext = &corev1.SecurityContext{
                    Capabilities: &corev1.Capabilities{
                        Add: []corev1.Capability{"NET_ADMIN", "NET_RAW"},
                    },  
                }   
            } else if container.SecurityContext.Capabilities == nil {
                // securityContext exists but no capabilities, add capabilities
                container.SecurityContext.Capabilities = &corev1.Capabilities{
                    Add: []corev1.Capability{"NET_ADMIN", "NET_RAW"},
                }   
            } else {
                // Both securityContext and capabilities exist, merge the "add" array
                container.SecurityContext.Capabilities.Add = append(container.SecurityContext.Capabilities.Add, "NET_ADMIN", "NET_RAW")
            }   

            addPatchEntry("replace", fmt.Sprintf("/spec/containers/%d/securityContext/capabilities", i), container.SecurityContext)
            fmt.Printf("Added NET_ADMIN and NET_RAW capabilities to test container in pod %s in namespace %s due to TEST_REQUIRES_BUILDFARM_NET_ADMIN=true\n", podName, namespace)
        }   
        fmt.Println(container)
        break
    }   
}
```

PR #4667, opened 2025-08-27
`gsm-secret-sync`: Adjust logging

psalajova:
Cleaning up logging

### DIFF
--- a/cmd/gsm-secret-sync/main.go
+++ b/cmd/gsm-secret-sync/main.go
@@ -119,15 +119,10 @@ func main() {
 func logChangeSummary(actions gsm.Actions) {
 	totalChanges := len(actions.SAsToCreate) + len(actions.SAsToDelete) + len(actions.SecretsToCreate) + len(actions.SecretsToDelete)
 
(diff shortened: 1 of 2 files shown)


cmd/gsm-secret-sync/main.go
```diff
@@ -56,13 +81,23 @@ func main() {
 	if err := o.Validate(); err != nil {
 		logrus.WithError(err).Fatal("Failed to validate options")
 	}
-	if err := o.setupLogger(); err != nil {
+
+	censor := secrets.NewDynamicCensor()
```
droslean, line 85: In some point you should add the sensitive data into the censor via https://github.com/openshift/ci-tools/blob/main/pkg/secrets/censor.go#L30

psalajova: Adding them here https://github.com/openshift/ci-tools/pull/4667/files#diff-f76efbacdffe8d72345f140e5e9362365918a9932e8e3085cf20b2ad3d618771R90

droslean (approved):

PR #4671, opened 2025-08-28
Add retry mechanism to generateServiceAccountKey

psalajova:
GCP sometimes takes a while to catch up while listing freshly created/deleted resources... sometimes, after a service account was freshly created, the call to create a key for it failed with NotFound error, as it couldn't find the SA (even though it was created just moments ago). This retry mechanism tries to remedy that. 3 retries should be enough, but maybe further tweaking will be necessary in the future as we'll see how it behaves in practice.

### DIFF
--- a/pkg/gsm-secrets/execution.go
+++ b/pkg/gsm-secrets/execution.go
@@ -3,16 +3,22 @@ package gsmsecrets
 import (
 	"context"
 	"fmt"
+	"net/http"
+	"time"
 
 	iamadmin "cloud.google.com/go/iam/admin/apiv1"
 	"cloud.google.com/go/iam/admin/apiv1/adminpb"
 	"cloud.google.com/go/iam/apiv1/iampb"
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
-	secretmanagerpb "cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
 	"github.com/googleapis/gax-go/v2"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+
+	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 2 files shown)


pkg/gsm-secrets/execution.go
```diff
@@ -98,19 +104,67 @@ func (a *Actions) CreateServiceAccounts(ctx context.Context, client IAMClient) {
...
+	return generateServiceAccountKeyWithBackoff(ctx, client, saEmail, projectID, gcpServiceAccountBackoff)
+}
+
+func generateServiceAccountKeyWithBackoff(ctx context.Context, client IAMClient, saEmail string, projectID string, backoff wait.Backoff) ([]byte, error) {
 	name := fmt.Sprintf("%s/serviceAccounts/%s", GetProjectResourceString(projectID), saEmail)
-	key, err := client.CreateServiceAccountKey(ctx, &adminpb.CreateServiceAccountKeyRequest{
-		Name: name,
+	attemptCount := 0
+	var key *adminpb.ServiceAccountKey
+	err := retry.OnError(backoff, isServiceAccountNotFoundError, func() error {
```
danilo-gemoli, line 144: From my understanding we need to introduce this retry because GCP takes a while to propagate changes across its infra.

Instead of doing this:
```golang
retry.OnError(backoff, isServiceAccountNotFoundError, func() error {
    key, innerErr = client.CreateServiceAccountKey(...)
    if innerErr != nil {
        ...
    }
}
```

Why don't we create the SA and then loop until it shows up?
```golang
client.CreateServiceAccountKey(...)
retry.OnError(backoff, isServiceAccountNotFoundError, func() error {
    // Not sure this call exists but you get the idea
    err = client.GetServiceAccount(...)
    if err != nil {
        ...
    }
}
```

psalajova: I don't understand what's the difference in your suggested approach and what the code's already doing 😅 as it is right now, it tries to create the service account (https://github.com/openshift/ci-tools/pull/4671/files#diff-9909cf7eebaa1d45dbbc42493a83dd1da1e6e4db11e8eed32c3d2f4567945139R148-R150), and if the creation fails, it tries again...

PR #4674, opened 2025-09-01
Konflux: Safeguard from creating several PJs for the same request

danilo-gemoli:
This check makes sure that no more than one ProwJob is created for any EphemeralCluster request.

When a new EphemeralCluster request arrives, the controller checks whether a ProwJob has been already created or not, by trying to get a ProwJob whose `metadata.name` matches what found within the EphemeralCluster's `status.prowJobId` stanza.

If it finds nothing, it goes ahead and try to list any ProwJobs whose `ci.openshift.io/ephemeral-cluster` label matches the aforementioned id, making sure that there are really no ProwJobs associated to this EphemeralCluster already.

### DIFF
--- a/pkg/controller/ephemeralcluster/reconciler.go
+++ b/pkg/controller/ephemeralcluster/reconciler.go
@@ -156,17 +156,7 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
 			return r.handleGetProwJobError(ctx, log, ec, err)
 		}
 	} else {
-		log.Info("ProwJob not found, creating")
-		err := r.createProwJob(ctx, log, ec)
-		if updateErr := r.updateEphemeralCluster(ctx, ec); updateErr != nil {
-			msg := utilerrors.NewAggregate([]error{updateErr, err}).Error()
-			return reconcile.Result{}, errors.New(msg)
-		}
-		requeueAfter := r.polling()
-		if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -258,6 +248,50 @@ func (r *reconciler) generateCIOperatorConfig(log *logrus.Entry, ec *ephemeralcl
...
+	}
+	if err := r.masterClient.List(ctx, &pjsForEC, listOpts...); err != nil {
+		log.WithError(err).Error("Failed to list ProwJobs to check whether the request has one associated already")
+		return reconcile.Result{}, fmt.Errorf("list pjs to create ec: %w", err)
+	}
+
+	// This is a case that should never happen but if it does it means that either someone
+	// is creating PJs manually or there is a bug in this controller.
+	if len(pjsForEC.Items) > 1 {
+		log.Error("Too many ProwJobs bound, this is a bug")
```
hector-vido, line 265: A typo? Should it be `found` instead of `bound` or `bounded`?

danilo-gemoli: `bound` is what I mean here. If we branch into that `if`, it means there is more than a single ProwJob associated (bound) to an EphemeralCluster, and this is an invalid condition, hence a bug in this controller.

PR #4677, opened 2025-09-02
`gsm-secret-sync`: Add e2e test

psalajova:
- This adds the e2e test which tests the functionality of `cmd/gsm-secret-sync` tool.
- It validates the complete lifecycle of Google Secret Manager resource management in a real GCP environment (`openshift-ci-secrets-dev` account created for this purpose). 
- It's expected to be run as a presubmit in the release repo, only if the source code of `gsm-secret-sync` tool changes.

This test suite runs the `gsm-secret-sync` binary against actual GCP APIs to verify it correctly manages:
- Service Accounts with proper key generation
- Secret Manager secrets (both service account keys and index secrets)
- IAM policy bindings with managed resource conditions

Test Coverage:
- Resource Creation: Creates resources from scratch
- Idempotency: Ensures repeated runs converge to a desired state
- Updates: Adds/modifies configurations without breaking existing resources
- Cleanup: Properly removes obsolete resources
 
Runs against staging GCP project using:
- `GCP_PROJECT_ID` and `GCP_PROJECT_NUMBER` environment variables
- `GOOGLE_APPLICATION_CREDENTIALS` for authentication
- Cleanup of resources before/after test runs so that each run of this e2e has a clean slate to work on

Release repo PR: https://github.com/openshift/release/pull/68661

### DIFF
--- a/test/e2e/gsm-secret-sync/gsm-e2e_test.go
+++ b/test/e2e/gsm-secret-sync/gsm-e2e_test.go
@@ -32,6 +32,12 @@ import (
 	"github.com/openshift/ci-tools/pkg/secrets"
 )
 
+const (
+	credentialsEnvVar   = "GCP_SECRETS_DEV_CREDENTIALS_FILE"
+	projectIdEnvVar     = "GCP_DEV_PROJECT_ID"
+	projectNumberEnvVar = "GCP_DEV_PROJECT_NUMBER"
+)
+
 type GCPState struct {
 	Secrets         map[string]gsm.GCPSecret
 	IAMBindings     []*iampb.Binding
@@ -61,14 +67,14 @@ var (
 )
 
 func getProjectConfigFromEnv() gsm.Config {
-	projectID := os.Getenv("GCP_PROJECT_ID")
+	projectID := os.Getenv(projectIdEnvVar)
 	if projectID == "" {
-		logrus.Fatal("GCP_PROJECT_ID not set")
+		logrus.Fatalf("%s not set", projectIdEnvVar)
 	}
 
-	projectNumber := os.Getenv("GCP_PROJECT_NUMBER")
+	projectNumber := os.Getenv(projectNumberEnvVar)
 	if projectNumber == "" {
-		logrus.Fatal("GCP_PROJECT_NUMBER not set")
+		logrus.Fatalf("%s not set", projectNumberEnvVar)
 	}
 
 	return gsm.Config{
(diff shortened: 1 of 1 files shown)


test/e2e/gsm-secret-sync/gsm-e2e_test.go
```diff
@@ -0,0 +1,495 @@
...
+		binaryPath:            "/go/bin/gsm-secret-sync",
+		secretsClient:         secretsClient,
+		iamAdminClient:        iamAdminClient,
+		resourceManagerClient: resourceManagerClient,
+	}
+
+	tr.cleanup()
+	if !tr.verifyProjectIsClean() {
+		logrus.Error("gcp project is not clean; skipping tests")
+		tr.cleanup()
```
danilo-gemoli, line 236: If the first call didn't make it, why would you try a second time here?

test/e2e/gsm-secret-sync/gsm-e2e_test.go
```diff
@@ -0,0 +1,495 @@
...
+
+	return GCPState{
+		Secrets:         secrets,
+		IAMBindings:     iamBindings,
+		ServiceAccounts: serviceAccounts,
+	}
+}
+
+// getActualGCPStateWithRetry gets the actual state with retry for eventual consistency
+func (tr *testRunner) getActualGCPStateWithRetry(expectedState GCPState) GCPState {
```
danilo-gemoli, line 380: Not sure whether should I raise this concern or not as it might not be a big deal. I would expect `getActualGCPStateWithRetry` to read the state from GCP and nothing else, instead of reading the state and asserting something (the number of SA and secrets in this case).

danilo-gemoli (changes requested):

test/e2e/gsm-secret-sync/gsm-e2e_test.go
psalajova: Yeah, I see your point. My reasoning for this is that because GCP sometimes takes a while to propagate changes, so if I see that I'm missing some resources from the expected state, I do a few retries to make sure the difference is not because of GCP's slowness, otherwise I'd get a lot of false negatives.

psalajova:
The new test [passed](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ci-tools/4677/pull-ci-openshift-ci-tools-main-e2e/1967945798280810496#1:build-log.txt%3A34)
/override ci/prow/e2e

PR #4680, opened 2025-09-03
`gsm-secrets`: Rename updater SA suffix

psalajova:
Refactoring: GCP has a limit of service account IDs being between 6 to 30 characters. Renaming from `-updater` suffix for all service accounts this tool manages to `-sa`, so that we get a little bit more space in naming. Also fixing some tests so that the string isn't hardcoded.


pkg/gsm-secrets/secrets.go
```diff
@@ -8,9 +8,10 @@ import (
 	"github.com/openshift/ci-tools/pkg/group"
 )
 
-// ValidateSecretName validates if a secret name matches the allowed pattern
+// ValidateSecretName validates if a secret name matches the allowed pattern.
+// The secret name must be between 1 and 255 characters long (GCP limit).
 func ValidateSecretName(secretName string) bool {
-	return regexp.MustCompile(SecretNameRegex).MatchString(secretName)
+	return regexp.MustCompile(SecretNameRegex).MatchString(secretName) && len(secretName) <= 255
```
pruan-rht, line 14: I would recommend replacing `255` with a constant name like `GCP_MAX_NAME_LENGTH` or something to that effect.

PR #4688, opened 2025-09-09
Added mabc crd to cluster-init

hector-vido:


pkg/api/multiarchbuildconfig/v1/types.go
```diff
@@ -14,6 +15,9 @@ import (
 	buildv1 "github.com/openshift/api/build/v1"
 )
 
+//go:embed ci.openshift.io_multiarchbuildconfigs.yaml
```
danilo-gemoli, line 18: I would move this into its own file at `/pkg/api/clusterinit.go`. IMHO this is a good move that would anticipate future needs when we might want to add more CRDs or stuff. By having a `//go:embed` directive rooted at `/pkg/api` means we can reference everything in `/pkg/api/**`.

pkg/clusterinit/onboard/passthrough.go
```diff
@@ -72,21 +74,24 @@ func (s *passthroughGenerator) Generate(ctx context.Context, log *logrus.Entry)
...
 		}
-
 		pathToManifests[path.Join(clusterRoot, p)] = data
 
 		return nil
 	}); err != nil {
 		return nil, fmt.Errorf("create manifests: %w", err)
 	}
 
+	data := make([]interface{}, 1)
```
danilo-gemoli, line 84: I guess that we can handle it better than this. Instead of:
```golang
//go:embed ci.openshift.io_multiarchbuildconfigs.yaml
var MultiArchBuildConfigManifest []byte
```
we could declare another `embed.FS` (see my comment above about moving this declaration into its own file):
```golang
//go:embed multiarchbuildconfig/v1/ci.openshift.io_multiarchbuildconfigs.yaml
var MultiArchBuildConfigManifest embed.FS
```
Then we change `passthroughGenerator` such that it would be able to handle multiple `fs.FS`:
```golang
type passthroughGenerator struct {
	manifests []fs.FS // or something similar, this is just to give you an idea
}
```

danilo-gemoli (changes requested):

pkg/clusterinit/onboard/passthrough.go
```diff
@@ -49,44 +50,43 @@ func (s *passthroughGenerator) Patches() []cinitmanifest.Patch {
...
 	pathToManifests := make(map[string][]interface{})
 
-	if err := fs.WalkDir(subFS, ".", func(p string, d fs.DirEntry, _ error) error {
-		if p == "." || d.IsDir() {
-			return nil
-		}
-
-		bytes, err := s.readFile(subFS, p)
+	for _, v := range s.manifests {
+		subFS, err := fs.Sub(v, passthroughRoot)
```
danilo-gemoli, line 56: I'm not sure that `passthroughRoot` (which value is `manifests`) is right for each `FS`. We have one filesystem rooted at `manifests` but the new one is `multiarchbuildconfig/v1/ci.openshift.io_multiarchbuildconfigs.yaml`.

PR #4690, opened 2025-09-10
Rename dummy index secret creation on selfservice

hector-vido:
This feature is creating secrets with . as name and . as value.

Since we are doing efforts on removing secrets with less than 3 characters, renaming this ones will allow us to see less noise on our investigations.


deepsm007 (approved):

pruan-rht (commented):
typo should be `length` instead of of `lenght`

PR #4708, opened 2025-09-22
Konflux: Allow provisioning clusters from Hive

danilo-gemoli:
Introducing the following changes:
1. Provision clusters from Hive:
    - The `.spec.ciOperator.test.clusterClaim` stanza has been added. It maps to the `cluster_claim` stanza of a regular `ci-operator` configuration.
    - The EphemeralCluster controller waits until the secrets that hold the hive cluster kubeconfig and admin password exist within the `ci-op-xxxx` NS, then copying them on the `EphemeralCluster` `.status` stanza.
2. Add Pull Request metadata. The idea is that the [tasks/provision-ephemeral-cluster/0.1/provision-ephemeral-cluster.yaml](https://github.com/openshift/konflux-tasks/blob/main/tasks/provision-ephemeral-cluster/0.1/provision-ephemeral-cluster.yaml) task would expand the following variables into the `annotations` of an Ephemeral Cluster object:
```yaml
apiVersion: ci.openshift.org/v1alpha1
kind: TestPlatformCluster
metadata:
  annotations:
    "ephemeralcluster.ci.openshift.io/pr-event-payload": |
      {{ body }}
    "ephemeralcluster.ci.openshift.io/pr-event-headers": |
      {{ headers }}
```
that hold, respectively, the [Pull Request event payload](https://docs.github.com/en/rest/using-the-rest-api/github-event-types?apiVersion=2022-11-28#pullrequestevent) coming from GitHub and the HTTP headers of the same request.
Those are needed in order to properly create a `ci-operator` configuration and a ProwJob, without missing any information. Some of the pieces we need are: `organization`, `repository` and `branch` of the PR from which the Ephemeral Cluster request originated from.

3. Add the following new stanzas:
    - `.spec.ciOperator.buildRoot` mapping to the `build_root` of a regular `ci-operator` configuration.
    - `.spec.ciOperator.baseImages` mapping to the `base_images` of a regular `ci-operator` configuration.
    - `.spec.ciOperator.externalImages` mapping to the `external_images` of a regular `ci-operator` configuration.

### DIFF
--- a/pkg/controller/ephemeralcluster/reconciler.go
+++ b/pkg/controller/ephemeralcluster/reconciler.go
@@ -7,6 +7,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"regexp"
+	"slices"
 	"strings"
 	"time"
 
@@ -41,11 +43,15 @@ import (
 const (
 	ControllerName            = "ephemeral_cluster_provisioner"
 	WaitTestStepName          = "wait-test-complete"
+	HiveKubeconfigSecret      = EphemeralClusterTestName + "-" + api.HiveAdminKubeconfigSecret
+	HiveAdminPasswdSecret     = EphemeralClusterTestName + "-" + api.HiveAdminPasswordSecret
 	EphemeralClusterTestName  = "cluster-provisioning"
 	EphemeralClusterLabel     = "ci.openshift.io/ephemeral-cluster"
 	EphemeralClusterNamespace = "ephemeral-cluster"
 	AbortProwJobDeleteEC      = "Ephemeral Cluster deleted"
 	DependentProwJobFinalizer = "ephemeralcluster.ci.openshift.io/dependent-prowjob"
+	PREventPayload            = "ephemeralcluster.ci.openshift.io/pr-event-payload"
+	PREventHeaders            = "ephemeralcluster.ci.openshift.io/pr-event-headers"
 	UnresolvedConfigVar       = "UNRESOLVED_CONFIG"
 	ProwJobCreatingDoneReason = "ProwJob has been properly created"
 	ProwJobNamePrefix         = "ephemeralcluster"
@@ -66,10 +72,13 @@ var (
 	defaultReconcilerOpts = reconcilerOptions{
 		polling: 3 * time.Second,
 	}
+
+	isTagRegexp = regexp.MustCompile(`(?P<Namespace>.+)/(?P<Name>.+)\:(?P<Tag>.+)`)
 )
 
 type reconcilerOptions struct {
-	polling time.Duration
+	polling     time.Duration
+	cliISTagRef string
 }
 
 type ReconcilerOption func(*reconcilerOptions)
@@ -77,6 +86,13 @@ type ReconcilerOption func(*reconcilerOptions)
 func WithPolling(polling time.Duration) ReconcilerOption {
 	return func(o *reconcilerOptions) {
 		o.polling = polling
+	}
+}
+
+// Set the image stream tag reference for the `cli` image in the form of `ocp/4.22:cli`.
+func WithCLIISTagRef(isTagRef string) ReconcilerOption {
+	return func(o *reconcilerOptions) {
+		o.cliISTagRef = isTagRef
 	}
 }
 
@@ -105,8 +121,9 @@ type reconciler struct {
 	prowConfigAgent *prowconfig.Agent
 
 	// Mock for testing
-	now     func() time.Time
-	polling func() time.Duration
+	now         func() time.Time
+	polling     func() time.Duration
+	cliISTagRef api.ImageStreamTagReference
 }
 
 func ECPredicateFilter(object ctrlruntimeclient.Object) bool {
@@ -124,6 +141,11 @@ func AddToManager(log *logrus.Entry, mgr manager.Manager, allManagers map[string
 		opt(&defaultReconcilerOpts)
 	}
 
+	cliISTagRef, err := parseCLIISTagRef(defaultReconcilerOpts.cliISTagRef)
+	if err != nil {
+		return err
+	}
+
 	r := reconciler{
 		logger:          log.WithField("controller", ControllerName),
 		masterClient:    mgr.GetClient(),
@@ -132,6 +154,7 @@ func AddToManager(log *logrus.Entry, mgr manager.Manager, allManagers map[string
 		newPresubmit:    pjutil.NewPresubmit,
 		now:             time.Now,
 		polling:         func() time.Duration { return defaultReconcilerOpts.polling },
+		cliISTagRef:     cliISTagRef,
 	}
 
 	if err := ctrlbldr.ControllerManagedBy(mgr).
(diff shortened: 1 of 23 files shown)


pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -265,7 +346,10 @@ func (r *reconciler) handleCreateProwJob(ctx context.Context, log *logrus.Entry,
 	// This is a case that should never happen but if it does it means that either someone
 	// is creating PJs manually or there is a bug in this controller.
 	if len(pjsForEC.Items) > 1 {
-		log.Error("Too many ProwJobs bound, this is a bug")
+		const errMsg = "Too many ProwJobs bound, this is a bug"
```
droslean, line 349: Let's make this a global const.

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -300,7 +384,21 @@ func (r *reconciler) createProwJob(ctx context.Context, log *logrus.Entry, ec *e
...
-	ciOperatorConfig, err := r.generateCIOperatorConfig(log, ec)
+	annotations := ec.Annotations
+	if annotations == nil {
+		annotations = map[string]string{}
+	}
+
+	prMeta, err := parsePRMeta(annotations[PREventPayload], annotations[PREventHeaders])
+	if err != nil {
+		log.WithError(err).Error("parse pull request meta")
+		err = fmt.Errorf("parse pull request meta: %w", err)
```
droslean, line 395: Here you are mutating the existing err.. In the end you will return the wrong error... Let's just put this in the global const of errors to ensure consistency.

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -392,43 +490,103 @@ func (r *reconciler) makeProwJob(ciOperatorConfig *api.ReleaseBuildConfiguration
...
+	for _, secret := range []struct {
+		name string
+		key  string
+		data *[]byte
+	}{
+		{name: HiveKubeconfigSecret, key: api.HiveAdminKubeconfigSecretKey, data: &kubeconfig},
+		{name: HiveAdminPasswdSecret, key: api.HiveAdminPasswordSecretKey, data: &passwd},
+	} {
+		s := corev1.Secret{}
+		if err = buildClient.Get(ctx, types.NamespacedName{Name: secret.name, Namespace: ns}, &s); err != nil {
```
droslean, line 533: The whole point of encapsulating the err in the if statement is to maintain a temporary variable.. It would be better to handle the error case directly instead of mutating the existing err variable.

droslean:
```go
if err := buildClient.Get(ctx, types.NamespacedName{Name: secret.name, Namespace: ns}, &s); err != nil && !kerrors.IsNotFound(err){
    log.WithField("secret", secret.name).WithError(err).Error("Failed to read secret")
    break
}

File: pkg/controller/ephemeralcluster/reconciler.go [hunk 4]
```diff
@@ -392,43 +490,103 @@ func (r *reconciler) makeProwJob(ciOperatorConfig *api.ReleaseBuildConfiguration
...
+			break
+		}
+
+		data, ok := s.Data[secret.key]
+		if !ok {
+			secretsReady = false
+			break
+		}
+
+		*secret.data = slices.Clone(data)
```
droslean, line 550: This looks dangerous.

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: Sure

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: Here this is what I'm doing:
```golang
result, err := foobar(...)
if err != nil {
    log.WithError(err).Error("foobar failed")
    err = fmt.Errorf("foobar failed: %w", err)
    return err
}
```
The `err` I return is the original one plus some context about the `foobar` function that has failed. Not returning the original error is intentional because I want the extra information to be there.

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: I want the error to be reported into the `status.conditions[]` stanza as well, to let the user know what went wrong.
Yeah I understand that shadowing `err` is not super clear. I would consider adding another `err` variable at wider scope then, like so:
```golang
var readSecretErr error
for ... {
    if err := buildClient.Get(ctx, types.NamespacedName{Name: secret.name, Namespace: ns}, &s); err != nil {
        if !kerrors.IsNotFound(err) {
            readSecretErr = fmt.Errorf("read secret %s/%s: %w", secret.name, ns, err)
            log.WithField("secret", secret.name).WithError(err).Error("Failed to read secret")
        }
        break
    }
}

if err != nil {
    // Report readSecretErr into `status.conditions[]`
    return
}
```

pkg/controller/ephemeralcluster/reconciler.go
droslean: But as the secret is just an element of the anonymous struct, you are trying to mutate it here and you are not using it anywhere else.

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -392,43 +490,103 @@ func (r *reconciler) makeProwJob(ciOperatorConfig *api.ReleaseBuildConfiguration
...
+	}
+
+	if err != nil {
```
droslean, line 553: This is confusing as well. You need to handle the error at the place you create it.

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -392,43 +490,103 @@ func (r *reconciler) makeProwJob(ciOperatorConfig *api.ReleaseBuildConfiguration
...
+		r.fetchHiveSecrets(ctx, log, buildClient, ns, ecStatus)
+	} else {
+		r.fetchClusterKubeconfig(ctx, log, buildClient, ns, ecStatus)
+	}
+
+	return nil
+}
+
+func (r *reconciler) fetchHiveSecrets(ctx context.Context, log *logrus.Entry, buildClient ctrlruntimeclient.Client, ns string, ecStatus *ephemeralclusterv1.EphemeralClusterStatus) {
+	secretsReady := true
```
droslean, line 520: This should be false by default.

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: Alright I'll do this way then, updating the conditions and `return`-ing as soon as I hit an error:
```golang
if err := buildClient.Get(ctx, types.NamespacedName{Name: secret.name, Namespace: ns}, &s); err != nil {
      if !kerrors.IsNotFound(err) {
          err = fmt.Errorf("read secret %s/%s: %w", secret.name, ns, err)
          log.WithField("secret", secret.name).WithError(err).Error("Failed to read secret")
          upsertCondition(ecStatus, ephemeralclusterv1.ClusterReady, ephemeralclusterv1.ConditionFalse, r.now(), ephemeralclusterv1.SecretsFetchFailureReason, err.Error())
          return
      }
      break
}
```

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: `data` is a pointer to `kubeconfig` on the first iteration and to `passwd` on the second. Doing so on the first iteration:
```golang
*secret.data = slices.Clone(data)
```
is like:
```
kubeconfig = slices.Clone(data)
```
Test this out:
```golang
func TestPointerToSlice(t *testing.T) {
	var kubeconfig []byte
	var adminPasswd []byte

	for i, testCase := range []struct {
		data *[]byte
	}{
		{data: &kubeconfig},
		{data: &adminPasswd},
	} {
		*testCase.data = []byte(fmt.Sprintf("data-%d", i))
	}

	if string(kubeconfig) != "data-0" {
		t.Fatalf("want data-1 but got %s", kubeconfig)
	}

	if string(adminPasswd) != "data-1" {
		t.Fatalf("want data-2 but got %s", adminPasswd)
	}
}
```

PR #4710, opened 2025-09-22
ci-operator: use QCI managed-clonerefs image

Prucek:
After this PR is merged, `ci-operator` will no longer require the `managed-clonerefs` IS to be present on buildfarms; it will use the image from QCI instead.
The build accepts [`DockerImage`](https://github.com/openshift/ci-tools/blob/main/vendor/github.com/openshift/api/build/v1/types.go#L474-L476) as an ImageSource so this is the biggest change (everything else changes the tests):
```
clonerefsRef := corev1.ObjectReference{Kind: "DockerImage", Name: s.config.ClonerefsPullSpec}
```


pkg/defaults/defaults.go
```diff
@@ -50,6 +50,8 @@ import (
 
 type inputImageSet map[api.InputImage]struct{}
 
+const ClonerefsPullSpec string = "quay-proxy.ci.openshift.org/openshift/ci/ci:ci_managed-clonerefs_latest"
```
danilo-gemoli, line 53: I'd put this into `pkg/api/types.go`

danilo-gemoli (changes requested):

PR #4712, opened 2025-09-23
Modified generic interfaces to k8s.io/api/core/v1

hector-vido:
This modifies the generic interfaces that are in use by defined kubernetes ones.

### DIFF
--- a/cmd/ci-scheduling-webhook/mutation.go
+++ b/cmd/ci-scheduling-webhook/mutation.go
@@ -221,7 +221,7 @@ func mutatePod(w http.ResponseWriter, r *http.Request) {
 					container.SecurityContext.RunAsUser = int64(0)
 					container.SecurityContext.RunAsNonRoot = false
 
-					addPatchEntry("replace", fmt.Sprintf("/spec/containers/%d/securityContext/capabilities", i), container.SecurityContext)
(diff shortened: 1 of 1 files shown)


cmd/ci-scheduling-webhook/mutation.go
```diff
@@ -203,74 +203,24 @@ func mutatePod(w http.ResponseWriter, r *http.Request) {
...
 					// Build the correct patch based on existing securityContext
 					if container.SecurityContext == nil {
-						// No securityContext exists, create one with capabilities and runAsUser
-						securityContext := map[string]interface{}{
-							"runAsUser":    int64(0),
-							"runAsNonRoot": false,
-							"capabilities": map[string]interface{}{
-								"add": []string{"NET_ADMIN", "NET_RAW"},
+						// No securityContext exists, create one with capabilities
+						container.SecurityContext = &corev1.SecurityContext{
```
bear-redhat, line 207: i would just skip if the SCC does not exist... creating SCC requires certain privileges and I'd rather create them manually/via apply-config

bear-redhat (changes requested):

cmd/ci-scheduling-webhook/mutation.go
hector-vido: Justin's answer:
It is not creating an SCC, it is configuring a pod's security context.
The necessary SCC to support the request is applied via gitops

PR #4724, opened 2025-09-29
CSI secrets: Handle dots in credential names

psalajova:


psalajova:
> @psalajova: Added dot‑replacement logic for GCP Secret Manager. Introduced `DotReplacementString` and `replaceDotInCredentialName()` to convert `--dot--` back to `.` so secrets mount with original names. Updated secret build to use the new function and added unit tests.

The changes in this PR assume that during the migration, secrets will have been renamed, e.g. what would before be 'collection__.awscreds' will be 'collection__--dot--awscreds' in GSM.

PR #4735, opened 2025-10-06
Revert "Allow escalation when tcpdump needs to run"

Prucek:
Reverts openshift/ci-tools#4733


jupierce:
/hold
Back online and have the real fix incoming.

PR #4737, opened 2025-10-06
Add GCS lock to prevent concurrent test runs

psalajova:
This PR adds distributed locking using GCS to prevent concurrent gsm-secret-sync e2e test runs from conflicting when modifying shared gcp resources (on our staging project). Uses atomic GCS conditional writes (GenerationMatch: 0) with exponential backoff retry (5 attempts over ~13 minutes: 1m, 1.5m, 2.25m, 3.4m, 5.1m) to ensure only one test instance can acquire the lock at a time, preventing race conditions and resource conflicts. Lock object is also automatically released when tests complete, with a 1-day bucket lifecycle policy (after 1 day, all objects in the bucket are deleted) as fail-safe cleanup.

For https://issues.redhat.com/browse/DPTP-4555

/hold
https://github.com/openshift/release/pull/69998 needs to be merged first

### DIFF
--- a/test/e2e/gsm-secret-sync/gsm-e2e_test.go
+++ b/test/e2e/gsm-secret-sync/gsm-e2e_test.go
@@ -5,8 +5,10 @@ package gsm_secret_sync
 
 import (
 	"context"
+	"errors"
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"os/exec"
 	"slices"
@@ -19,7 +21,9 @@ import (
 	resourcemanager "cloud.google.com/go/resourcemanager/apiv3"
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
 	"cloud.google.com/go/secretmanager/apiv1/secretmanagerpb"
+	"cloud.google.com/go/storage"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/api/googleapi"
 	"google.golang.org/api/option"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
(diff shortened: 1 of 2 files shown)


psalajova:
If the linter compains again I'm going to cry

test/e2e/gsm-secret-sync/gsm-e2e_test.go
```diff
@@ -239,15 +271,74 @@ func TestMain(m *testing.M) {
...
 		logrus.Error("gcp project is not clean; skipping tests")
 		tr.cleanup()
+		releaseLock()
 		os.Exit(1)
 	}
 
 	code := m.Run()
 
 	tr.cleanup()
+	releaseLock()
```
Prucek, line 281: I am wondering if these are needed when you used defer in the beginning.

test/e2e/gsm-secret-sync/gsm-e2e_test.go
```diff
@@ -36,8 +40,19 @@ const (
 	credentialsEnvVar   = "GCP_SECRETS_DEV_CREDENTIALS_FILE"
 	projectIdEnvVar     = "GCP_DEV_PROJECT_ID"
 	projectNumberEnvVar = "GCP_DEV_PROJECT_NUMBER"
+
+	// GCS lock bucket configuration
+	gcsBucketEnvVar = "GCS_E2E_LOCK_BUCKET"
+	lockObjectName  = "gsm-secret-sync-e2e-lock"
```
Prucek, line 46: I am wondering if you couldn't just use a `sync.Mutex` or `sync.RWMutex` with  `m.Lock()` and `m.Unlock()` instead of writing into a bucket. I think it would simplify it. Go has a great multithreaded model, and you are using something else to synchronize the threads.
I was just reviewing this PR of Jakub: https://github.com/openshift/ci-tools/pull/4744, and it's used there.

droslean:
Why do we need to run this tool in our e2e tests?

PR #4764, opened 2025-10-16
Fix to exclude creating slackreportconfig for excluded pattern

deepsm007:
Follow up of https://github.com/openshift/ci-tools/pull/4758 
/cc @openshift/test-platform


pkg/prowgen/prowgen.go
```diff
@@ -266,6 +257,19 @@ func generatePresubmitForTest(jobBaseBuilder *prowJobBaseBuilder, name string, i
 
 	shortName := info.TestName(name)
 	base := jobBaseBuilder.Rehearsable(!opts.disableRehearsal).Build(jc.PresubmitPrefix)
+
+	// Set slack reporter config using full job name for proper excluded_job_patterns matching
+	fullJobName := info.JobName(jc.PresubmitPrefix, name)
+	if slackReporter := info.Config.GetSlackReporterConfigForJobName(fullJobName, name, info.Metadata.Variant); slackReporter != nil {
```
danilo-gemoli: This code has been duplicated three times. You might want to consider extracting it into its own function:
```golang
func addSlackReporterConfig(base *prowconfig.JobBase, jobName, testName string, info *ProwgenInfo) {
	if slackReporter := info.Config.GetSlackReporterConfigForJobName(jobName, testName, info.Metadata.Variant); slackReporter != nil {
		if base.ReporterConfig == nil {
			base.ReporterConfig = &prowv1.ReporterConfig{}
		}
		base.ReporterConfig.Slack = &prowv1.SlackReporterConfig{
			Channel:           slackReporter.Channel,
			JobStatesToReport: slackReporter.JobStatesToReport,
			ReportTemplate:    slackReporter.ReportTemplate,
		}
	}
}
```

pkg/config/load.go
```diff
@@ -96,6 +96,38 @@ func (p *Prowgen) GetSlackReporterConfigForTest(test, variant string) *SlackRepo
 	return nil
 }
 
+// GetSlackReporterConfigForJobName checks against full job names, allowing excluded_job_patterns
+// to work with prefixes like "pull-", "periodic-", etc.
+func (p *Prowgen) GetSlackReporterConfigForJobName(fullJobName, testName, variant string) *SlackReporterConfig {
```
danilo-gemoli: I have two considerations regarding this function.
In general I prefer less nesting levels, like so:
```golang
func (p *Prowgen) GetSlackReporterConfigForJobName(fullJobName, testName, variant string) *SlackReporterConfig {
nextSlackReporterConfig:
	for _, s := range p.SlackReporterConfigs {
		if slices.Contains(s.ExcludedVariants, variant) {
			continue
		}

		// Check if job is excluded by pattern (using full job name)
		for _, excludePattern := range s.ExcludedJobPatterns {
			if matched, err := regexp.MatchString(excludePattern, fullJobName); err == nil && matched {
				continue nextSlackReporterConfig
			}
		}

		// Check exact job name matches first (against test name for backward compatibility)
		if slices.Contains(s.JobNames, testName) {
			return &s
		}

		// Check regex pattern matches (against test name for backward compatibility)
		for _, pattern := range s.JobNamePatterns {
			if matched, err := regexp.MatchString(pattern, testName); err == nil && matched {
				return &s
			}
		}
	}
	return nil
}
```

Then I notice that we use the PJ name to match against `excluded_job_patterns` and the ci-operator test name to match against `job_names` and `job_name_patterns`. Is it intended? Why do we differentiate this way?

PR #4841, opened 2025-11-20
aggregation: remove `Cluster should remain functional during upgrade` from always passing

stbenjam:
This test was previously excluded from aggregation because it would fail alongside other results, so we didn't need to aggregate this one (I think that was the reasoning).

https://prow.ci.openshift.org/view/gs/test-platform-results/logs/aggregated-aws-ovn-upgrade-4.21-micro-fips-release-openshift-release-analysis-aggregator/1991442012515078144 is an example where it failed 8x alone and we didn't aggregate it or catch the problem.

The only reason we caught the problem was openshift/origin#30509, the ClusterOperators test does not always run so we failed on missing results, rather than the actual failing test.


stbenjam:
Closing — this PR has gone stale and is no longer being actively pursued. (Automated cleanup by agent)

PR #4849, opened 2025-11-24
Only apply 5-minute sleep workaround for multi-arch builds

deepsm007:
/cc @openshift/test-platform

### DIFF
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -467,12 +467,30 @@ func isBuildPhaseTerminated(phase buildapi.BuildPhase) bool {
 }
 
 type ImageBuildOptions struct {
-	Architectures []string
+	Architectures            []string
+	NeedsMultiArchWorkaround func() bool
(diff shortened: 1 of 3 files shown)


Prucek (approved):

pkg/steps/source.go
```diff
@@ -482,6 +500,17 @@ func handleBuilds(ctx context.Context, buildClient BuildClient, podClient kubern
 		o = opts[0]
 	}
 
+	// Create closure that checks if this build is for a multi-arch repo
+	needsMultiArchWorkaround := o.NeedsMultiArchWorkaround
+	if needsMultiArchWorkaround == nil {
+		needsMultiArchWorkaround = func() bool {
```
stevekuznetsov: If this closure doesn't take any dynamic data as input, I don't understand why calculating the build's labels later would help? What's the reasoning behind making this a closure?

PR #4866, opened 2025-12-05
DPTP-4340: Remove direct mirroring to registry.ci.openshift.org, consolidate to single promotion step

deepsm007:
Removes the  direct mirroring step to `registry.ci.openshift.org` and consolidates promotion into a single step that handles both quay.io mirroring and tagging.

The promotion now: 
(1) mirrors to `quay.io/openshift/ci` using `oc image mirror`, then
(2) tags in `registry.ci.openshift.org` from `quay-proxy.ci.openshift.org` using `oc tag`.
(3) Release assembly: `oc adm release new` now tries `--from-image-stream-file` (from oc get imagestream … -o yaml) and falls back to `--from-image-stream` if that fails (fromFile || fromStream when export succeeds; stream-only if oc get fails).

/hold
/cc @Prucek @openshift/test-platform 

## Summary by CodeRabbit

* **Bug Fixes**
  * Quay promotions now run only when no registry override is configured; otherwise quay promotion is skipped.
  * Promotion targets no longer include the "-quay" infix, yielding shorter tag names and digest-based mirror targets when available.

* **Refactor**
  * Promotion logic adjusted to better decide tag vs. mirror operations.
  * Release creation now prefers an imagestream-file approach with a safe fallback to stream-based creation.

* **Tests**
  * Updated tests and fixtures to match the new naming, tagging, and mirror behaviors.


deepsm007:
@coderabbitai resume

pkg/steps/release/create_release_test.go
```diff
@@ -82,22 +82,47 @@ func TestBuildOcAdmReleaseNewCommand(t *testing.T) {
...
+		{"uses from-image-stream-file", "--from-image-stream-file=${RELEASE_IS_FILE}"},
+		{"file then stream with or", " || "},
+		{"from-image-stream fallback", "--from-image-stream stable"},
+	}
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			if !strings.Contains(got, testCase.want) {
+				t.Errorf("%s: got does not contain %q:\n%s", testCase.name, testCase.want, got)
+			}
+		})
```
Prucek: I don't understand what this is testing

pkg/steps/release/create_release.go
```diff
@@ -331,5 +333,24 @@ func buildOcAdmReleaseNewCommand(config *api.ReleaseTagConfiguration, namespace,
...
+RELEASE_IS_FILE="${ARTIFACT_DIR:-/tmp}/release-input-is.yaml"
+GET_ERR="${ARTIFACT_DIR:-/tmp}/release-is-get.err"
+if oc get imagestream -n %q %q -o yaml > "${RELEASE_IS_FILE}" 2>"${GET_ERR}"; then
+	%s || %s
+else
+	echo "$(date -Is) oc get imagestream -n %q %q -o yaml failed, stderr:" >&2
+	cat "${GET_ERR}" >&2
+	%s
+fi
+)`, namespace, streamName, fromFile, fromStream, namespace, streamName, fromStream)
```
Prucek: Could we do something with this? I don't like the logic in a shell script, but I guess not.

pkg/steps/release/promote.go
```diff
@@ -262,12 +262,15 @@ func getPromotionPod(imageMirrorTarget map[string]string, timeStr string, namesp
...
+			if strings.HasPrefix(k, api.QuayOpenShiftCIRepo+":") {
+				// Images promoted into quay.io/openshift/ci always use oc image mirror (tag or digest sources).
+				images = append(images, fmt.Sprintf("%s=%s", src, k))
+			} else if strings.Contains(k, api.ComponentFormatReplacement) || !strings.Contains(src, "@sha256:") || strings.Contains(src, api.QCIAPPCIDomain) {
+				// Cluster ImageStream tags: oc tag for non-digest sources, quay-proxy imports, or ${component} templates.
+				tags = append(tags, fmt.Sprintf("%s %s", src, k))
 			} else {
-				// Default to mirroring for quay.io targets and other registries
-				images = append(images, fmt.Sprintf("%s=%s", imageMirrorTarget[k], k))
+				images = append(images, fmt.Sprintf("%s=%s", src, k))
```
Prucek, line 364: Could you explain this to me in an example, maybe?

deepsm007:
> when we merge this only the X.Y-quay IS will stay, is that what we want? Or do we want the current X.Y IS to use QCI images https://github.com/openshift/ci-tools/blob/main/pkg/api/promotion.go#L114

We only want 4.x imagestream and the mechanism which was used to tag from quay-proxy will be used on 4.X.Y imagestream. The $.X.Y-quay was created for testing.

pkg/steps/release/create_release_test.go
deepsm007: it ensures the big generated shell string still has the right pieces after refactors, instead of testing behavior by running a shell in Go.

pkg/steps/release/create_release.go
deepsm007: Stuck with shell in the string for this design; we can still make the Go side clearer, safer, and better tested. If you want, I can do a small refactor in-repo: extract the template + add a one-shot golden expected string for the default args (no behavior change)

pkg/steps/release/promote.go
deepsm007: Destinations under `quay.io/openshift/ci` always use `oc image mirror` (tag or digest sources). All other destinations use `oc tag` when the key has `${component}`, the source has no `@sha256:`, or the source is quay-proxy; otherwise they use `oc image mirror` (e.g. pipeline digest → `registry.ci` / similar). This fixes mis-routing that came from classifying only on quay-proxy instead of destination + digest/pull spec.

pkg/steps/release/create_release_test.go
```diff
@@ -101,3 +102,22 @@ func TestBuildOcAdmReleaseNewCommand(t *testing.T) {
 		}
 	}
 }
+
+func TestBuildOcAdmReleaseNewShell(t *testing.T) {
```
droslean: You are not testing anything here. The `buildOcAdmReleaseNewShell()` returns a string. Instead, you are trying to simulate the same behavior of the original function and try to count or check if there are parts of string in there.. The whole concept is wrong. Try to compare exactly the `string` you get from the buildOcAdmReleaseNewShell fn.

pkg/steps/release/create_release.go
```diff
@@ -311,11 +311,17 @@ func hasMinimumVersion(config *api.ReleaseTagConfiguration, majorVersion, minorV
 	return major > majorVersion || (major == majorVersion && minor >= minorVersion)
 }
 
+// buildOcAdmReleaseNewCommand returns the `oc adm release new` invocation using only
+// --from-image-stream. Tests and callers that need the stream-only form use this.
 func buildOcAdmReleaseNewCommand(config *api.ReleaseTagConfiguration, namespace, streamName, cvo, destination, version string) string {
+	return buildOcAdmReleaseNewCommandLine(config, namespace, streamName, cvo, destination, version, "--from-image-stream", streamName)
+}
+
+func buildOcAdmReleaseNewCommandLine(config *api.ReleaseTagConfiguration, namespace, streamName, cvo, destination, version, fromFlag, fromValue string) string {
```
droslean: I don't see any point having this function. Why this has to be splitted away from buildOcAdmReleaseNewCommand?

PR #4886, opened 2026-01-06
DPTP-2938: optional authoritative decreases in pod-scaler admission

deepsm007:
Pod-scaler looks at past runs in Prometheus and adjusts CPU/memory on new pods. Today it only **bumps requests up** when history says you asked for too little. It never lowers them, even when a job has been over-provisioned for weeks.

This PR adds two opt-in flags so we *can* lower requests later — but **both default to off**, so nothing changes until someone turns them on in release.

### Flags

- `--authoritative-cpu` (default: `false`)
- `--authoritative-memory` (default: `false`)

### Example (CPU request)

A step always requests **4 CPUs**. Histogram data says it usually needs **1 CPU** (after the usual 1.2× safety factor that's about **1.2 CPUs**).

| Mode | What admission does |
|------|---------------------|
| Today (flag off) | Leaves **4 CPU** — only increases, never decreases |
| `--authoritative-cpu=true` | Can lower toward **~1.2 CPU**, but not more than **25% per pod admission** (4 → 3, not 4 → 1 in one shot) |

Authoritative mode is "right-size down slowly," not "slash requests in one go."

### Other guardrails

- Change under **5%** → skip (avoid noise)
- **Measured pods** (`pod-scaler.openshift.io/measured=true`) → still no decreases
- No recommendation for that container → skip

### Why default off

On busy build nodes, CPU metrics can look *low* when the node is starved — lowering requests from bad data makes packing worse.

/cc @openshift/test-platform

### DIFF
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -37,7 +37,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/steps"
 )
 
-func admit(port, healthPort int, certDir string, client buildclientv1.BuildV1Interface, kubeClient kubernetes.Interface, loaders map[string][]*cacheReloader, mutateResourceLimits bool, cpuCap int64, memoryCap string, cpuPriorityScheduling int64, percentageMeasured float64, measuredPodCPUIncrease float64, systemReservedCPU int64, reporter results.PodScalerReporter) {
+func admit(port, healthPort int, certDir string, client buildclientv1.BuildV1Interface, kubeClient kubernetes.Interface, loaders map[string][]*cacheReloader, mutateResourceLimits bool, cpuCap int64, memoryCap string, cpuPriorityScheduling int64, percentageMeasured float64, measuredPodCPUIncrease float64, systemReservedCPU int64, authoritativeCPU, authoritativeMemory bool, reporter results.PodScalerReporter) {
 	logger := logrus.WithField("component", "pod-scaler admission")
 	logger.Infof("Initializing admission webhook server with %d loaders.", len(loaders))
 	health := pjutil.NewHealthOnPort(healthPort)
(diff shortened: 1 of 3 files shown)


cmd/pod-scaler/admission.go
```diff
@@ -44,23 +44,25 @@ func admit(port, healthPort int, certDir string, client buildclientv1.BuildV1Int
...
+	logger                      *logrus.Entry
+	client                      buildclientv1.BuildV1Interface
+	resources                   *resourceServer
+	mutateResourceLimits        bool
+	decoder                     admission.Decoder
+	cpuCap                      int64
+	memoryCap                   string
+	cpuPriorityScheduling       int64
+	authoritativeCPURequests    bool
+	authoritativeMemoryRequests bool
```
Prucek: I think this is not needed. I think this was meant for easy swithcing back to the current state/version, right? Which can be done with the script to revert to previous version.
Podscaler is the authoritative resource requester and the flags just make the code a bit more complicated, IMO.

cmd/pod-scaler/admission.go
```diff
@@ -231,13 +243,40 @@ func useOursIfLarger(allOfOurs, allOfTheirs *corev1.ResourceRequirements, worklo
...
+						} else {
+							fieldLogger.Debug("difference less than 5%, skipping micro-adjustment")
+						}
+					} else {
+						fieldLogger.Debug("theirs is zero, applying recommendation")
+						(*pair.theirs)[field] = our
+					}
+				} else {
+					fieldLogger.Debug("authoritative mode disabled, keeping existing value")
+				}
```
Prucek: I think there are too many indentations here, and they could be resolved by flipping the condition or altering the condition slightly.

Prucek (commented):
Left some small comments, otherwise LGTM.

jupierce:
/hold
This defaults to "on" right now for authoritativeCPU. We know there is a significant issue here until measurements do not pods running on CPU pressured worker nodes.

deepsm007:
> /hold This defaults to "on" right now for authoritativeCPU. We know there is a significant issue here until measurements do not pods running on CPU pressured worker nodes.

Updated default to false "off"

deepsm007:
> /hold This defaults to "on" right now for authoritativeCPU. We know there is a significant issue here until measurements do not pods running on CPU pressured worker nodes.

/unhold
now the authoritative flag will be set false as default

Prucek (approved):

PR #4887, opened 2026-01-06
DPTP-4613: pod-scaler implement measured pods for accurate resource measurement 

deepsm007:
Implements measured pods feature to address CPU bottleneck issues from node contention. Pods are automatically classified as "normal" or "measured" based on measurement freshness (>10 days), with measured pods running on isolated nodes via pod anti-affinity rules. Integrates BigQuery to query and cache measured pod data from `ci_operator_metrics` table, and adds data collection component that queries Prometheus for completed measured pods and writes max CPU/memory usage to BigQuery. Applies measured resource recommendations to longest-running containers using actual measured utilization instead of Prometheus data skewed by node contention. This enables accurate resource recommendations and addresses the CPU bottleneck issue ([DPTP-4613](https://issues.redhat.com//browse/DPTP-4613)) blocking authoritative CPU mode.

/cc @openshift/test-platform

### DIFF
--- a/cmd/pod-scaler/measured.go
+++ b/cmd/pod-scaler/measured.go
@@ -1,0 +1,410 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"time"
+
+	"cloud.google.com/go/bigquery"
+	"github.com/sirupsen/logrus"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/prow/pkg/interrupts"
+
+	"github.com/openshift/ci-tools/pkg/api"
+	podscaler "github.com/openshift/ci-tools/pkg/pod-scaler"
+)
+
+const (
+	// PodScalerLabelKey is the label we use to mark pods as either "normal" or "measured".
+	// This helps the scheduler know which pods can share nodes and which need isolation.
+	PodScalerLabelKey = "pod-scaler"
+	// PodScalerLabelValueNormal means this pod can run alongside other CI workloads.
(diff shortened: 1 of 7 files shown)


cmd/pod-scaler/measured.go
```diff
@@ -0,0 +1,443 @@
...
+						Values:   []string{PodScalerLabelValueMeasured},
+					},
+				},
+			},
+			TopologyKey: "kubernetes.io/hostname",
+		})
+		logger.Debug("Added podAntiAffinity for normal pod - will avoid measured pods")
+	}
+
+	if len(requiredTerms) > 0 {
```
Prucek: This if can be omitted, I don't see a sitution where it is not true

cmd/pod-scaler/measured.go
```diff
@@ -0,0 +1,443 @@
...
+// We check each container - if any container needs measuring, the whole pod gets the "measured" label.
+// This label tells the scheduler to keep it away from other CI workloads.
+func ClassifyPod(pod *corev1.Pod, bqClient *BigQueryClient, logger *logrus.Entry) {
+	if pod.Labels == nil {
+		pod.Labels = make(map[string]string)
+	}
+
+	// Check each container to see if we need fresh measurement data for it.
+	// If any container needs measuring, we mark the whole pod as "measured".
+	shouldBeMeasured := false
```
Prucek: this can be omited, if you would move the 
```
pod.Labels[PodScalerLabelKey] = PodScalerLabelValueMeasured
logger.Debugf("Classified pod as measured - will run on isolated node")
```
directly into the `if bqClient.ShouldBeMeasured(fullMeta)`

deepsm007:
/close

This branch implemented the original [DPTP-4613](https://redhat.atlassian.net/browse/DPTP-4613) design (BigQuery + ci_operator_metrics + large producer changes). Measured pods landed on main a different way: Prometheus histograms, pod anti-affinity, and the flags in release (#4912, openshift/release#74535). 

Closing to avoid confusion. Further 4613 work should build on what is already deployed, not this branch.

PR #4926, opened 2026-02-03
DPTP-4632: use merge-base for accurate change detection in pj-rehearse when base branch has moved forward

deepsm007:
/cc @openshift/test-platform @Prucek


Prucek (commented):
So the difference from the last PR is that we use merge-base for everything except changes in the step registry?

Prucek (approved):

PR #4990, opened 2026-03-06
prowgen: add sparse checkout for .ci-operator.yaml and Dockerfiles

Prucek:
## Summary
- Adds sparse checkout support to prowgen-generated jobs, reducing clone size by only fetching files that ci-operator actually needs
- When `from_repository` is set, `.ci-operator.yaml` is included in the sparse checkout
- Dockerfiles from the `images` stanza are included (with deduplication), respecting `context_dir` and `dockerfile_path`
- Images using `dockerfile_literal` or `ref` are skipped (no file to clone)
- Uses `DecorationConfig.SparseCheckoutFiles` which prow applies to the primary ref for presubmits/postsubmits

## Summary by CodeRabbit

* **Improvements**
  * Replace blanket skip-cloning with sparse-checkout: only required files (Dockerfiles, .ci-operator.yaml) are fetched, with path derivation, ordering preservation and deduplication to minimize checkouts. Private-job decoration now retains token configuration while relying on sparse-checkout emptiness to decide cloning behavior.

* **Tests**
  * Added unit tests and many updated fixtures validating sparse-checkout generation across build roots, image inputs, private jobs, and periodic job scenarios.

### DIFF
--- a/pkg/prowgen/jobbase.go
+++ b/pkg/prowgen/jobbase.go
@@ -1,9 +1,11 @@
 package prowgen
 
 import (
+	"path"
 	"time"
 
-	utilpointer "k8s.io/utils/pointer"
+	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/utils/ptr"
 	prowv1 "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 	prowconfig "sigs.k8s.io/prow/pkg/config"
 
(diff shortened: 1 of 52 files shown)


pkg/prowgen/jobbase.go
```diff
@@ -20,22 +21,51 @@ type prowJobBaseBuilder struct {
...
 
-	return true
+// sparseCheckoutFiles returns the list of files to include in a sparse checkout
+// when the configuration requires cloning. This includes the in-repo ci-operator
+// configuration file and any Dockerfiles that need to be built.
+// Returns an empty slice if no sparse checkout is needed.
+func sparseCheckoutFiles(configSpec *cioperatorapi.ReleaseBuildConfiguration) []string {
+	seen := map[string]bool{}
+	var result []string
+	add := func(path string) {
```
droslean: You can just use a set and just Insert it. Now you are using a closure to combine a map that won't keep duplicates to append to a slice with deduplication...

droslean (approved):

PR #5021, opened 2026-03-17
determinize-ci-operator: add --validate-main-promotion for main/master promotion validation

deepsm007:
## What

Add main-promotion validation to `determinize-ci-operator` so the same rules enforced in openshift/release (via a standalone script) run inside the existing toolchain. No new script in release once release repo is updated to call determinize with the new flag.

## How

- **New package** `pkg/mainpromotion`: `Validate()` enforces (1) main/master configs promote only to the current release (`ocp/{current}`, `ocp-private/{current}-priv`), and (2) release-X/openshift-X configs have promotion disabled when the repo has main/master and Prow has that release in a tide query’s `includedBranches`. Uses the same ignore list as the release script (gatekeeper, offline_migration, etc.; cri-o is not ignored).
- **New flags** on `determinize-ci-operator`: `--validate-main-promotion`, `--current-release`, `--prow-config-dir`, `--infra-periodics`, `--main-promotion-ignore`. When `--validate-main-promotion` is set, validation runs first and the process exits 1 on violations. Current release can come from `--current-release` or from the `periodic-prow-auto-config-brancher` job in `--infra-periodics`.

## Release repo follow-up

After this merges, release will switch `check-validate-main-promotion` to run determinize-ci-operator with `--validate-main-promotion` and the appropriate mounts (`--config-dir`, `--infra-periodics`, `--prow-config-dir`) and remove `hack/validate-main-promotion-guard.py`.

/cc @danilo-gemoli

### DIFF
--- a/cmd/determinize-ci-operator/main.go
+++ b/cmd/determinize-ci-operator/main.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/validation/promotion"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/mainpromotion/validate.go
```diff
@@ -0,0 +1,229 @@
...
+
+type tideQuery struct {
+	IncludedBranches []string `json:"includedBranches"`
+}
+
+type prowTide struct {
+	Queries []tideQuery `json:"queries"`
+}
+
+type prowConfig struct {
```
danilo-gemoli: You can reuse the `ProwConfig` type defined in the package `sigs.k8s.io/prow/pkg/config`.

danilo-gemoli (changes requested):

PR #5042, opened 2026-03-23
multi-pr-prow-plugin: keep origin base ref for cross-repo /testwith

deepsm007:
## What / why

The operand PR’s `Base.Ref` was always cleared for `/testwith`. For **cross-repo** runs (e.g. comment on `openshift/oc` while the job targets `openshift/cluster-version-operator`), the origin repo never gets a branch from job metadata, so `createRefsForPullRequests` called configresolver with an **empty** branch and failed with:

`resolve config openshift/oc@: got unexpected http 400 status code from configresolver: branch query missing or incorrect`

Only clear `Base.Ref` when the operand PR is on the **same** org/repo as the job in the command; otherwise keep the GitHub API base ref.

##
Ref:  [openshift/oc#2170 (comment)](https://github.com/openshift/oc/pull/2170#issuecomment-4109566062) — `testwith: could not generate prow job` with the `branch query missing or incorrect` error.

/cc @openshift/test-platform

### DIFF
--- a/cmd/multi-pr-prow-plugin/server.go
+++ b/cmd/multi-pr-prow-plugin/server.go
@@ -291,11 +291,14 @@ func (s *server) determineJobRuns(comment string, originPR github.PullRequest) (
 				additionalPRs = append(additionalPRs, *pr)
 			}
 
-			// Store the OriginPR with its base branch cleared. The branch is captured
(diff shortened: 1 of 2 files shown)


joepvd (approved):
/lgtm

PR #5044, opened 2026-03-24
DPTP-4739: payload-job expand sharded base names to all shards

deepsm007:
## What

When a periodic is **sharded** (`-1ofN`, …, `-NofN`), `/payload-job` with the **base** job name (no shard suffix) now submits **all shards** in one run. Passing the **full** name including `-XofY` still runs **only that shard**.

## Why

Matches pre-sharding UX: one command with the usual job name runs the full job; optional `-XofY` selects a single shard.

## Behavior

| Command | Result |
|--------|--------|
| `/payload-job <base-periodic-name>` | All shards if `shard_count` > 1 in ci-operator |
| `/payload-job <name>-XofY` | Shard X only |

`/payload-aggregate` is unchanged: it does not expand a base name to all shards (still one target job per invocation).

## How it works

`fileTestResolver` returns the test’s configured shard count from ci-operator. If the comment has no `-XofY` suffix, shard count > 1, and the job is not an aggregate command, the handler emits one `ReleaseJobSpec` per shard.

/cc @openshift/test-platform

### DIFF
--- a/cmd/payload-testing-prow-plugin/filetestresolver.go
+++ b/cmd/payload-testing-prow-plugin/filetestresolver.go
@@ -12,7 +12,7 @@ type fileTestResolver struct {
 	configAgent agents.ConfigAgent
 }
 
-func (r *fileTestResolver) resolve(job string) (api.MetadataWithTest, error) {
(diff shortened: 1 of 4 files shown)


joepvd (approved):
/lgtm

PR #5075, opened 2026-04-01
feat: align lease acquire timeout with test timeout

raelga:
### Problem

The lease acquire timeout is fixed and independent of the test's own configured timeout. When a test sets a long timeout (e.g. 10h), the lease can still time out well before the job's overall timeout budget is exhausted, causing jobs to fail to acquire a lease even though there's plenty of time left to wait.

### Goal

Scale the lease acquire timeout with the test's configured timeout, so long-running tests don't fail to acquire a lease while the overall job timeout would still allow waiting.

### What changes

The lease acquire timeout now automatically increases to match the test timeout when it's longer than the current default.

### Validation

`/retest` run passed all tests.

### Follow-ups

Jira: [AROSLSRE-1922](https://redhat.atlassian.net/browse/AROSLSRE-1922)


joepvd (approved):
/lgtm
/test images

PR #5101, opened 2026-04-10
auto-config-brancher: fix mirror args parsing and remove rebalancer

jmguzik:
Ensure private-prow-configs-mirror receives auth flags before other args and use --dry-run=false to avoid bool parsing pitfalls. Remove the rebalancer flag/step and related cron window logic from auto-config-brancher.

### DIFF
--- a/cmd/auto-config-brancher/main.go
+++ b/cmd/auto-config-brancher/main.go
@@ -11,7 +11,6 @@ import (
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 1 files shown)


cmd/auto-config-brancher/main.go
```diff
@@ -232,17 +226,6 @@ func main() {
 		},
 	}
 
-	if rebalance {
```
jmguzik, line 235: not needed anymore

PR #5104, opened 2026-04-14
Konflux: Use Periodics rather than Presubmits

danilo-gemoli:
It turns out that Konflux users can request an ephemeral cluster from different "contexts", and not each of them has a PR associated with it.
This means that we should not rely on any data coming from a PR, like the HTTP Request payload or headers.

For that reason, the reconciler now uses a Periodic rather than a Presubmit in order to setup and run `ci-operator`. That seems more appropriate because a Presubmit, by definition, is strictly tied to a PR, whereas a Periodic is not.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed GitHub pull request event handling from ephemeral cluster provisioning.
  * Ephemeral cluster jobs now use periodic scheduling instead of pull request–based scheduling.
  * Simplified job creation workflow by eliminating PR metadata parsing and injection steps.

* **Tests**
  * Updated test fixtures and unit tests to reflect the new periodic job creation behavior.

### DIFF
--- a/cmd/dptp-controller-manager/main.go
+++ b/cmd/dptp-controller-manager/main.go
@@ -543,15 +543,10 @@ func main() {
 	}
 
 	if opts.enabledControllersSet.Has(ephemeralcluster.ControllerName) {
-		gitHubClient, err := opts.GitHubClient(opts.dryRun)
-		if err != nil {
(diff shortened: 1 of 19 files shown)


jmguzik (approved):
Using periodics makes sense

PR #5106, opened 2026-04-15
refactor(prowgen): move Prowgen config to pkg/api and add ProwgenExtras

Prucek:
## Summary                                                                                                                                                                                                                           
  - Moves `Prowgen` struct from `pkg/config` to `pkg/api`                                                                                                                                                                              
  - Introduces `ProwgenExtras` with `NewProwgenExtras` constructor that merges                                                                                                                                                         
    `.config.prowgen` and ci-operator config into a single object                                                                                                                                                                      
  - Decouples `ProwgenInfo`  so it only holds .config.prowgen data.                                                                                                                                                                                                
  - First step of incremental migration away from `.config.prowgen`

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added configuration options to disable test rehearsals at both per-test and repository levels
  * Enhanced Slack reporter configuration with pattern matching and variant exclusion capabilities
  * Added ability to skip presubmits based on branch and variant conditions
  * Extended release configuration schema to support new Prow job generation settings
<!-- end of auto-generated comment: release notes by coderabbit.ai -->


pkg/api/types.go
```diff
@@ -859,6 +869,11 @@ type TestStepConfiguration struct {
 	// RestrictNetworkAccess restricts network access to RedHat intranet.
 	RestrictNetworkAccess *bool `json:"restrict_network_access,omitempty"`
 
+	// DisableRehearsal prevents this specific test from being picked up for rehearsals.
+	// Note: this cannot re-enable rehearsals if they are globally disabled via
```
droslean, line 873: Well, the note here is irrelevant since you will depreciate that. Also, we need to override it. So if its globaly disabled, then enabling it here should override the global setting. So the comment is wrong too.

pkg/config/load.go
```diff
@@ -102,6 +102,17 @@ nextSlackReporterConfig:
 	return nil
 }
 
+// ApplyOverrides applies overrides from ci-operator config's prowgen stanza.
+// Fields set in the ci-operator config take precedence over .config.prowgen values.
```
droslean, line 106: The comments are irrelevant. Same for the other. Just remove them all.

pkg/prowgen/jobbase.go
```diff
@@ -3,7 +3,7 @@ package prowgen
 import (
 	"time"
 
-	utilpointer "k8s.io/utils/pointer"
+	"k8s.io/utils/ptr"
```
droslean, line 6: all changes in this file are not related to what you are introducing in this PR.

pkg/prowgen/prowgen.go
```diff
@@ -43,7 +43,7 @@ func GenerateJobs(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *Pro
 	postsubmits := map[string][]prowconfig.Postsubmit{}
 	var periodics []prowconfig.Periodic
 	rehearsals := info.Config.Rehearsals
-	disabledRehearsals := sets.New[string](rehearsals.DisabledRehearsals...)
+	disabledRehearsals := sets.New(rehearsals.DisabledRehearsals...)
```
droslean, line 46: same here. Its not that important but it would be better to open another PR that fixes that in the whole ci-tools repo.

pkg/prowgen/prowgen_test.go
```diff
@@ -841,6 +888,9 @@ func TestGenerateJobs(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.id, func(t *testing.T) {
+			// Mimic the behavior of ci-operator config loading where the repo config is overridden
```
droslean, line 891: No need for that comment.

pkg/prowgen/prowgen_test.go
```diff
@@ -841,6 +888,9 @@ func TestGenerateJobs(t *testing.T) {
...
+			// by prowgen config and then generate jobs based on the merged config
+			tc.repoInfo.Config.ApplyOverrides(tc.config.Prowgen)
```
droslean, line 893: This is wrong here. This is proof that the whole test cases are wrong now. You shouldn't need to apply the override before running the function you want to test. It seems that you are trying to hack the test to work instead.

pkg/api/types.go
```diff
@@ -24,6 +26,100 @@ func IsPromotionJob(jobLabels map[string]string) bool {
 	return ok
 }
 
+// ProwgenExtras holds fields that control Prow job generation behavior.
+// TODO: After migration, remove the Prowgen embedding.
+type ProwgenExtras struct {
```
droslean, line 31: But you don't even need it now. You can just remove it and use the Prowgen struct directly, right?

pkg/api/types.go
```diff
@@ -24,6 +26,100 @@ func IsPromotionJob(jobLabels map[string]string) bool {
...
+	if defaults.EnableSecretsStoreCSIDriver {
+		p.EnableSecretsStoreCSIDriver = true
+	}
+	if defaults.Rehearsals.DisableAll {
+		p.Rehearsals.DisableAll = true
+	}
+	p.Rehearsals.DisabledRehearsals = append(p.Rehearsals.DisabledRehearsals, defaults.Rehearsals.DisabledRehearsals...)
+}
+
+func (p *Prowgen) GetSlackReporterConfigForJobName(fullJobName, testName, variant string) *SlackReporterConfig {
```
droslean, line 117: This is the old logic and belongs to the config.prowgen mechanism.

pkg/api/types.go
```diff
@@ -24,6 +26,100 @@ func IsPromotionJob(jobLabels map[string]string) bool {
...
+		for _, pattern := range s.JobNamePatterns {
+			if matched, err := regexp.MatchString(pattern, testName); err == nil && matched {
+				return &s
+			}
+		}
+	}
+	return nil
+}
+
+func (p *Prowgen) SkipPresubmits(branch string, variant string) bool {
```
droslean, line 140: Same here. This logic belongs to the .config.prowgen.

pkg/api/types.go
Prucek: still is

pkg/api/types.go
Prucek: still is, just moved

pkg/api/types.go
Prucek: here new fields will be added
like `DisableRehearsals` adn in this struct `ProwgenExtras` I will have info of both and I could merge it and then the deprecation would just be to remove the Prowgen from ProwgenExtras and everyhting will work

PR #5110, opened 2026-04-16
Added the capacity to use AWS Codebuild on ci-operator

hector-vido:
The purpose of this changes is to have another build mechanism, se we can relay on other alternatives than Openshift Builds, in this case we choose [AWS CodeBuild](https://docs.aws.amazon.com/codebuild/latest/userguide/welcome.html).

A new key was added to `pkg/api/types.go` named `BuildType`, right now it can be `openshift` or `aws`.

Most of the logic are inside `pkg/steps/source.go`.

All builds happens on AWS and the resulting image is pushed inside the internal registry of the build cluster.

### build.openshift.io/v1

Actually, most of the info used to build the image inside AWS are extracted from the `build` object, this made the integration easier - since almost all info are inside it - but this object is passed along the way only for that purpose, this could be overkill and confuse.

### TO-DOS:

- Add all builds inside a single project (if this is possible, will be easier to understand, clear and possibly faster)
- Clean aws codebuild project and cloudwatch logs after the build succeed.
- Test more targets

## Summary by CodeRabbit

* **New Features**
  * Configurable BuildType option supporting AWS CodeBuild and OpenShift execution
  * Defaults to OpenShift when unspecified
  * BuildType is propagated to all relevant build steps
  * AWS CodeBuild path with project provisioning, buildspec generation, registry credential handling, and CloudWatch log collection
  * Added utility to shell-escape commands used in generated build specs

* **Chores**
  * Updated AWS SDK and related dependencies to newer versions


pkg/api/types.go
```diff
@@ -39,6 +39,8 @@ type ReleaseBuildConfiguration struct {
 
 	InputConfiguration `json:",inline"`
 
+	BuildType string `json:"build_type,omitempty"`
```
droslean, line 42: Since this is a top level configuration, and we have only one new way of using the build which is the aws builds. Perhaps it would be just better to have a boolean here like `use_aws_builds` or something like that.

pkg/defaults/defaults.go
```diff
@@ -201,13 +201,13 @@ func fromConfig(ctx context.Context, cfg *Config) ([]api.Step, []api.Step, error
 			inputImages[conf.InputImage] = struct{}{}
 		} else if rawStep.PipelineImageCacheStepConfiguration != nil {
 			skippedBinaries := filterRequiredBinariesFromSkipped(cfg.CIConfig.Images.Items, cfg.SkippedImages)
-			step = steps.PipelineImageCacheStep(*rawStep.PipelineImageCacheStepConfiguration, cfg.CIConfig.Resources, cfg.buildClient, cfg.podClient, cfg.JobSpec, cfg.PullSecret, cfg.MetricsAgent, skippedBinaries)
+			step = steps.PipelineImageCacheStep(*rawStep.PipelineImageCacheStepConfiguration, cfg.CIConfig.Resources, cfg.buildClient, cfg.podClient, cfg.JobSpec, cfg.PullSecret, cfg.MetricsAgent, skippedBinaries, cfg.CIConfig.BuildType)
```
droslean, line 204: You will avoid all of those changes if you encapsulate the build method inside the buildClient. Since all builds will use the same build type.

pkg/steps/source.go
```diff
@@ -908,3 +906,302 @@ func addLabelsToBuild(refs *prowv1.Refs, build *buildapi.Build, contextDir strin
...
+
+func awsBuild(ctx context.Context, buildClient BuildClient, build buildapi.Build, errs *[]error) (bool, error) {
+	projectName := fmt.Sprintf("%s-%s", build.Namespace, build.Name)
+	sdkConfig, err := config.LoadDefaultConfig(ctx)
+	if err != nil {
+		*errs = append(*errs, err)
+		return false, fmt.Errorf("could not load aws credentials: %w", err)
+	}
+
+	cbClient := codebuild.NewFromConfig(sdkConfig)
```
droslean, line 943: Instead of initializing the same client all over again for each build, you should be able to initialize it once inside the buildClient interface.

pkg/steps/source.go
```diff
@@ -908,3 +906,302 @@ func addLabelsToBuild(refs *prowv1.Refs, build *buildapi.Build, contextDir strin
...
+	pList, err := cbClient.ListProjects(ctx, nil)
```
droslean, line 944: Do we really need to create a new project for each build?

pkg/steps/source.go
```diff
@@ -908,3 +906,302 @@ func addLabelsToBuild(refs *prowv1.Refs, build *buildapi.Build, contextDir strin
...
+		return false, handleFailedBuild(ctx, buildClient, ns, name, err)
+	}
+	if err := gatherSuccessfulBuildLog(buildClient, ns, name); err != nil {
+		// log error but do not fail successful build
+		logrus.WithError(err).Warnf("Failed gathering successful build %s logs into artifacts.", name)
+	}
+	return true, nil
+}
+
+func awsBuild(ctx context.Context, buildClient BuildClient, build buildapi.Build, errs *[]error) (bool, error) {
```
droslean, line 935: There is no point on passing the errs and mutate it. The function will return one error in the end, you should append it out of this function.

pkg/steps/source.go
```diff
@@ -908,3 +906,302 @@ func addLabelsToBuild(refs *prowv1.Refs, build *buildapi.Build, contextDir strin
...
+		return false, fmt.Errorf("could not list project: %w", err)
+	}
+
+	if !slices.Contains(pList.Projects, projectName) {
+		is := &imagev1.ImageStreamTag{}
+		credentials, err := awsBuildAssemblyCredentials(ctx, buildClient, build, is, errs)
+		if err != nil {
+			return false, err
+		}
+		buildSpec, err := awsBuildSpec(build, string(credentials), is.Image.DockerImageReference, buildClient.LocalRegistryDNS())
```
droslean, line 956: Can you elaborate more on where this build will run? Is this going to still be an Openshift object that uses AWS builds natively, or will we watch the resource directly in AWS? What are those credentials? How will the build push back in the local registry in the end? I would really need this kind of information to review the code more appropriately.

pkg/steps/source.go
```diff
@@ -908,3 +906,302 @@ func addLabelsToBuild(refs *prowv1.Refs, build *buildapi.Build, contextDir strin
...
+}
+
+func awsBuildSpec(build buildapi.Build, credentials string, imageStreamReference, localRegistryDNS string) (buildspec.BuildSpec, error) {
+	commands := []string{
+		`mkdir -p ~/.docker`,
+		`echo "$credentials" > ~/.docker/config.json`,
+		`echo "$dockerfile" > Dockerfile`,
+	}
+	for i, image := range build.Spec.CommonSpec.Source.Images {
+		commands = append(commands, fmt.Sprintf(`docker create --name c%d %s`, i, image.From.Name))
```
droslean, line 993: If this runs in an OpenShift pod, it will fail.

pkg/steps/source.go
hector-vido: I probably can condensate all builds inside a single project and push everything from there, but there will be at least one project per "job".

pkg/steps/source.go
hector-vido: The build will run inside AWS.
We will watch the resource directly on AWS.
These credentials are used to pull/push from/to AWS, Build Clusters (and/or eventually inside Quay).
After the build, the image is automatically uploaded to a registry, this is an example:
```docker buildx build --build-arg BUILD_LOGLEVEL='0' --platform linux/amd64 --tag registry.build11.ci.openshift.org/hvidosil-ci-operator/pipeline:bin-amd64 --output type=registry .```

PR #5119, opened 2026-04-21
prowgen: allow ci-operator config to override .config.prowgen

Prucek:
## Summary
- Add `ProwgenOverrides` to ci-operator config (`prowgen` field on `ReleaseBuildConfiguration`) so that ci-operator config takes precedence over all other prowgen configuration layers (org-level and repo-level `.config.prowgen` files)
- Add per-test `disable_rehearsal` on `TestStepConfiguration` for fine-grained rehearsal control
- Add `ApplyOverrides` method as the single merge point where ci-operator config overrides `.config.prowgen` values
- First step toward deprecating `.config.prowgen` by allowing its features to be controlled directly from ci-operator config

## Test plan
- [x] Unit tests for global rehearsal disable via ci-operator config
- [x] Unit test for ci-operator config taking precedence over `.config.prowgen`
- [x] Unit test for per-test `disable_rehearsal`
- [x] All existing tests pass (`go test ./pkg/prowgen/...`)

## Summary by CodeRabbit

* **New Features**
  * Added release-level and per-test controls to disable rehearsals, allowing teams to turn off rehearsals for an entire release or for individual tests for finer-grained execution control.

* **Tests**
  * Added unit tests and test fixtures to validate rehearsal-disable behavior, precedence of release-level overrides, and per-test disablement scenarios.


pkg/api/types.go
```diff
@@ -24,6 +24,12 @@ func IsPromotionJob(jobLabels map[string]string) bool {
 	return ok
 }
 
+// ProwgenOverrides holds prowgen-related settings from the ci-operator config
+// that override the corresponding .config.prowgen values.
```
droslean: The comment is irrelevant.

pkg/api/types.go
```diff
@@ -24,6 +24,12 @@ func IsPromotionJob(jobLabels map[string]string) bool {
...
+type ProwgenOverrides struct {
```
droslean, line 27: No need to introduce this as an override option.

pkg/api/types.go
```diff
@@ -37,6 +43,10 @@ func IsPromotionJob(jobLabels map[string]string) bool {
 type ReleaseBuildConfiguration struct {
 	Metadata Metadata `json:"zz_generated_metadata"`
 
+	// Prowgen holds prowgen-related overrides from the ci-operator config that
+	// take precedence over .config.prowgen values.
```
droslean: The comment is irrelevant as well. This stanza should include the prowgen options either way. The config.prowgen is not related now, and also it will be deprecated.

pkg/api/types.go
```diff
@@ -856,6 +866,9 @@ type TestStepConfiguration struct {
 	// If set, the generated test pod will include a nodeSelector for this architecture.
 	NodeArchitecture NodeArchitecture `json:"node_architecture,omitempty"`
 
+	// DisableRehearsal prevents this specific test from being picked up for rehearsals.
+	DisableRehearsal *bool `json:"disable_rehearsal,omitempty"`
```
droslean: Any particular reason why this is a pointer?

pkg/config/load.go
```diff
@@ -102,6 +102,17 @@ nextSlackReporterConfig:
 	return nil
 }
 
+// ApplyOverrides applies ci-operator config overrides on top of .config.prowgen values.
+// Fields set in the ci-operator config take precedence.
+func (p *Prowgen) ApplyOverrides(overrides *cioperatorapi.ProwgenOverrides) {
```
droslean: There are no overrides. The logic already exists, and it already overrides different layers. You should just mutate that logic and add the ci-op config in the equation.

droslean (approved):

PR #5126, opened 2026-04-23
Ephemeral Cluster Controller: Terminate ci-operator gracefully

danilo-gemoli:
Up until now, when the `EphemeralCluster` object is gone, the `ProwJob` running `ci-operator` is simply aborted.
This might lead to an unpleasant situation in which `ci-operator` is getting killed in the middle of provisioning an ephemeral-cluster, without giving it the chance to clean up the resources that have been created in the cloud so far.

With the new approach, the controller lets `ci-operator` finish the provisioning procedure and then triggers the deprovision right away, without waiting for the Konflux's test to complete.

In this way we let `ci-operator` to gracefully terminate its job.

In order to ease the review, I will provide a real use case:
1. The `EphemeralCluster` object in `app.ci` gets delete. This might result as a consequence of the Konflux's pipeling being canceled, as an example.
1. The `prowjobreconciler` kicks in and get the `ProwJob` object tied to the `EphemeralCluster`. We don't know what is the status of this `ProwJob`, it might be running, pending, etc.
1. The `prowjobreconciler` looks for the `ci-op-xxx` NS that should have been created by the `ci-operator` instance run by the `ProwJob`
    1. If the **NS exists**, the reconciler create the `test-done-signal` secret in order to inform `ci-operator` to start the cluster deprovisioning procedures as soon as possible.
    1. If the **NS doesn't exist**, the reconciler simply aborts the `ProwJob`, as it currently does.

## Summary by CodeRabbit

* **Bug Fixes**
  * Graceful handling when a referenced ephemeral cluster is missing: triggers deprovisioning via a test-namespace signal secret or cleanly aborts the job; aborts now reliably record completion time.

* **Improvements**
  * Better multi-cluster behavior: reconciler selects the correct cluster-specific client per job and standardizes metadata/time handling and error checks.
  * Simplified test-complete signaling to treat existing signals as success.

* **Tests**
  * Added/updated tests for graceful termination, signal secret creation/preservation, and unknown-cluster error cases.

### DIFF
--- a/pkg/controller/ephemeralcluster/prowjobreconciler.go
+++ b/pkg/controller/ephemeralcluster/prowjobreconciler.go
@@ -8,19 +8,22 @@ import (
 
 	"github.com/sirupsen/logrus"
 
-	kerrors "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	ctrlbldr "sigs.k8s.io/controller-runtime/pkg/builder"
-	ctrlclient "sigs.k8s.io/controller-runtime/pkg/client"
+	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	prowv1 "sigs.k8s.io/prow/pkg/apis/prowjobs/v1"
 
+	"github.com/openshift/ci-tools/pkg/api"
 	ephemeralclusterv1 "github.com/openshift/ci-tools/pkg/api/ephemeralcluster/v1"
+	"github.com/openshift/ci-tools/pkg/steps"
 )
 
 const (
@@ -28,22 +31,24 @@ const (
 )
 
 type prowJobReconciler struct {
-	logger *logrus.Entry
-	client ctrlclient.Client
+	logger       *logrus.Entry
+	masterClient ctrlruntimeclient.Client
+	buildClients buildClients
 
 	now func() time.Time
 }
 
-func ProwJobFilter(object ctrlclient.Object) bool {
+func ProwJobFilter(object ctrlruntimeclient.Object) bool {
 	_, ok := object.GetLabels()[EphemeralClusterLabel]
 	return ok
 }
 
-func addPJReconcilerToManager(logger *logrus.Entry, mgr manager.Manager) error {
+func addPJReconcilerToManager(logger *logrus.Entry, mgr manager.Manager, buildClients buildClients) error {
 	r := prowJobReconciler{
-		logger: logger.WithField("controller", "ephemeral_cluster_provisioner_pj"),
-		client: mgr.GetClient(),
-		now:    time.Now,
+		logger:       logger.WithField("controller", "ephemeral_cluster_provisioner_pj"),
+		masterClient: mgr.GetClient(),
+		buildClients: buildClients,
+		now:          time.Now,
 	}
 
 	if err := ctrlbldr.ControllerManagedBy(mgr).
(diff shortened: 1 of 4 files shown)


pkg/controller/ephemeralcluster/prowjobreconciler.go
```diff
@@ -72,13 +80,71 @@ func (r *prowJobReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
+	}
+
+	log = log.WithField("secret", api.EphemeralClusterTestDoneSignalSecretName)
+	if createSecret {
+		if err := buildClient.Create(ctx, &corev1.Secret{ObjectMeta: metav1.ObjectMeta{
+			Name:      api.EphemeralClusterTestDoneSignalSecretName,
+			Namespace: ns,
+		}}); err != nil {
+			log.WithError(err).Warn("Failed to create the secret")
 			return reconcile.Result{}, err
```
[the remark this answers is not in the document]
danilo-gemoli: This should never happen in practice because the function `gracefullyTerminateClusterProvisioning` is executed only when the `EphemeralCluster` object doesn't exist anymore, therefore there is no way for the reconciler in `pkg/controller/ephemeralcluster/reconciler.go` to create the `test-done-signal`.

In `notifyTestComplete` requests are handled sequentially, therefore there is no way to interleave `Get`/`Create` sequence.

Am I missing anything?

pkg/controller/ephemeralcluster/prowjobreconciler.go
```diff
@@ -72,13 +80,71 @@ func (r *prowJobReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
+		if apierrors.IsNotFound(err) {
+			createSecret = true
 		} else {
+			log.WithError(err).Warn("Failed to fetch the secret")
+			return reconcile.Result{}, err
+		}
+	}
+
+	log = log.WithField("secret", api.EphemeralClusterTestDoneSignalSecretName)
+	if createSecret {
```
droslean: You can just create the secret directly, instead of turning a boolean to true and checking later. Also, can you elaborate more on this, IIUC based on the description of the PR, it sounds like a workaround because ci-operator doesn't support graceful shutdowns in the way we want it.

pkg/controller/ephemeralcluster/prowjobreconciler.go
```diff
@@ -91,11 +157,24 @@ func (r *prowJobReconciler) abortProwJob(ctx context.Context, pj *prowv1.ProwJob
...
+	if err := r.masterClient.Update(ctx, pj); err != nil {
 		return reconcile.Result{}, fmt.Errorf("abort prowjob: %w", err)
 	}
 
 	return reconcile.Result{}, nil
 }
+
+func (r *prowJobReconciler) findCIOperatorTestNS(ctx context.Context, buildClient ctrlruntimeclient.Client, pj *prowv1.ProwJob) (string, error) {
+	nss := corev1.NamespaceList{}
+	if err := buildClient.List(ctx, &nss, ctrlruntimeclient.MatchingLabels{steps.LabelJobID: pj.Name}); err != nil {
```
droslean, line 161: There will always be only one namespace. Since you already know the label selector, you can just Get it instead of Listing it and getting the first element by default.

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -830,3 +823,11 @@ func parseCLIISTagRef(isTag string) (api.ImageStreamTagReference, error) {
 
 	return isTagRef, nil
 }
+
+func buildClientFor(pj *prowv1.ProwJob, buildClients map[string]ctrlruntimeclient.Client) (ctrlruntimeclient.Client, error) {
```
droslean: It would more idiomatic if you make this function a method of the buildClients instead.

pkg/controller/ephemeralcluster/prowjobreconciler.go
danilo-gemoli: Unfortunately you can't a GET a single secret by using a label, the pattern is: LIST all the secrets that match a label. See [options.go](https://github.com/kubernetes-sigs/controller-runtime/blob/626b2f36e1b6ef8daa1bec519a282f49618f0d71/pkg/client/options.go#L430-L434)

pkg/controller/ephemeralcluster/reconciler.go
danilo-gemoli: Will do, thanks.

pkg/controller/ephemeralcluster/prowjobreconciler.go
danilo-gemoli: > You can just create the secret directly, instead of turning a boolean to true and checking later.

Done, thanks.

> Also, can you elaborate more on this, IIUC based on the description of the PR, it sounds like a workaround because ci-operator doesn't support graceful shutdowns in the way we want it.

I don't think this is a workaround but rather an optimization. Aborting a Prowjob while `ci-operator` is in the middle of provisioning a cluster might be more disrupting that letting `ci-operator` provisioning the cluster and the trigger the deprovisioning procedures right away.

droslean (approved):

PR #5127, opened 2026-04-23
prowgen: allow ci-operator config to set private and expose

Prucek:
## Summary
- Add `private` and `expose` fields to the `prowgen` section of ci-operator config
- Jobs can now control visibility (hidden from deck) and git credential mounting directly in ci-operator config
- Continues the incremental deprecation of `.config.prowgen` (follows #5119 rehearsals, #5120 CSI, #5121 skip-presubmits)

## Test plan
- [x] New test: "private job via ci-operator config" — fixture identical to existing `.config.prowgen`-based private test
- [x] New test: "private job with expose via ci-operator config" — verifies `hidden` is not set when expose is true
- [x] All existing tests pass (`go test ./pkg/prowgen/...`)

## Summary by CodeRabbit

* **New Features**
  * Added a configurable "private" flag to mark generated jobs as hidden from deck
  * Added an "expose" flag to allow showing jobs that are otherwise private
  * Private jobs now mount appropriate git credentials automatically when needed

* **Tests**
  * Added test coverage and fixtures for private jobs with and without expose enabled

### DIFF
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -28,6 +28,13 @@ type ProwgenOverrides struct {
 	DisableRehearsals           bool `json:"disable_rehearsals,omitempty"`
 	SkipOperatorPresubmits      bool `json:"skip_operator_presubmits,omitempty"`
 	EnableSecretsStoreCSIDriver bool `json:"enable_secrets_store_csi_driver,omitempty"`
+	// Private indicates that generated jobs should be marked as hidden
+	// from display in deck and that they should mount appropriate git credentials
+	// to clone the repository under test.
+	Private bool `json:"private,omitempty"`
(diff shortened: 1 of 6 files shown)


pkg/api/types.go
```diff
@@ -28,6 +28,13 @@ type ProwgenOverrides struct {
 	DisableRehearsals           bool `json:"disable_rehearsals,omitempty"`
 	SkipOperatorPresubmits      bool `json:"skip_operator_presubmits,omitempty"`
 	EnableSecretsStoreCSIDriver bool `json:"enable_secrets_store_csi_driver,omitempty"`
+	// Private indicates that generated jobs should be marked as hidden
```
droslean, line 31: This not entirely true. We need this configuration to inject the oauth settings in the decoration config.

PR #5129, opened 2026-04-23
Always include kubeAdminPassword in EphemeralCluster credentials Secret

amisstea:
The credentials Secret created by the EphemeralCluster controller was missing the kubeAdminPassword key when provisioned via the non-Hive code path. Ensure both credential keys are always present so downstream consumers never encounter a missing key.

## Summary by CodeRabbit

* **Refactor**
  * Credentials secrets now consistently include both kubeconfig and admin password fields (admin password may be empty), standardizing secret structure.

* **Tests**
  * Updated test expectations to validate the revised credentials secret format, ensuring both fields are present.

### DIFF
--- a/pkg/controller/ephemeralcluster/reconciler.go
+++ b/pkg/controller/ephemeralcluster/reconciler.go
@@ -532,7 +532,10 @@ func (r *reconciler) fetchHiveSecrets(
 		return
 	}
 
-	secretData := map[string][]byte{"kubeconfig": kubeconfig, "kubeAdminPassword": passwd}
+	secretData := map[string][]byte{
+		"kubeconfig":        kubeconfig,
+		"kubeAdminPassword": passwd,
+	}
 	if err := r.createCredentialsSecret(ctx, log, ec, secretData); err != nil {
 		log.WithError(err).Error("Failed to create credentials secret")
 		upsertCondition(ecStatus, ephemeralclusterv1.ClusterReady, ephemeralclusterv1.ConditionFalse, r.now(), ephemeralclusterv1.SecretsFetchFailureReason, err.Error())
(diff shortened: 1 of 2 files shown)


pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -568,7 +568,7 @@ func (r *reconciler) fetchClusterKubeconfig(
 		return
 	}
 
-	secretData := map[string][]byte{"kubeconfig": kubeconfig}
+	secretData := credentialsSecretData(kubeconfig, nil)
 	if err := r.createCredentialsSecret(ctx, log, ec, secretData); err != nil {
```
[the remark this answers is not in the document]
amisstea: This seems unnecessary since the secret is ephemeral and there is no current usage.

pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -600,6 +600,16 @@ func credentialsSecretName(ec *ephemeralclusterv1.EphemeralCluster) string {
 	return ec.Name + "-credentials"
 }
 
+func credentialsSecretData(kubeconfig, kubeAdminPassword []byte) map[string][]byte {
```
danilo-gemoli: Just inline this function.
In one place replace with:
```golang
map[string][]byte{
  "kubeconfig":        kubeconfig,
  "kubeAdminPassword": kubeAdminPassword,
}
```
in another:
```golang
map[string][]byte{
  "kubeconfig":        kubeconfig,
  "kubeAdminPassword": []byte{},
}
```

amisstea: Done. The AI review might flag this though.

PR #5139, opened 2026-04-28
fix: use LocalTagReferencePolicy for input image tags

Prucek:
## Summary
- Switch input image tag IST reference policy from `Source` to `Local`
- Builds now pull base images from the internal registry instead of quay-proxy
- Avoids "manifest unknown" failures when quay.io garbage-collects a digest after import
- Add "manifest unknown" to `hintsAtInfraReason` so it's classified as an infrastructure error
- Add `isNonRetriableInfraError` check to skip futile retries when a manifest is permanently gone

## Problem
When quay.io GCs a base image digest (e.g., `ubi-minimal`), builds fail with `PullBuilderImageFailed: manifest unknown` even though the image was already imported into the internal registry. With `SourceTagReferencePolicy`, builds are directed back to quay-proxy to pull, which fails. ci-operator then retries 5 times futilely.

## Fix
1. **`LocalTagReferencePolicy`** for input image tags — builds pull from the internal registry where the image is already cached from the import step. External registry GC no longer causes build failures.
2. **`hintsAtInfraReason`** — classify "manifest unknown" as an infrastructure error so it's properly reported.
3. **`isNonRetriableInfraError`** — detect "manifest unknown" as a permanent error and skip retries, saving minutes of futile attempts per affected image.

## Summary by CodeRabbit

* **Updates**
  * Image tag reference policy now uses local references when creating tags, affecting how images are resolved and pulled.
* **Bug Fixes**
  * Builds reporting a "manifest unknown" error are no longer retried automatically; these failures are logged and handled without recreate/delete cycles.

### DIFF
--- a/pkg/steps/input_image_tag.go
+++ b/pkg/steps/input_image_tag.go
@@ -108,7 +108,7 @@ func (s *inputImageTagStep) run(ctx context.Context) error {
 		},
 		Tag: &imagev1.TagReference{
 			ReferencePolicy: imagev1.TagReferencePolicy{
-				Type: imagev1.SourceTagReferencePolicy,
+				Type: imagev1.LocalTagReferencePolicy,
 			},
 			From: from,
 			ImportPolicy: imagev1.TagImportPolicy{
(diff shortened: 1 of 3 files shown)


deepsm007:
/lgtm
Would be nice to have some rehearsal or runs to demonstrate the imagestreams.json from artifacts

deepsm007:
> Line 125 of pkg/steps/input_image_tag.go discards error from waitForTagInSpec() instead of wrapping with %w, violating the pattern used elsewhere in the file. 	Add error parameter to fmt.Errorf call on line 125 to include %w and preserve error context as shown in the resolution.

Worth addressing

PR #5149, opened 2026-04-30
prowgen: allow per-test slack reporter config in ci-operator config

Prucek:
## Summary
- Adds a `reporter_config` field to `TestStepConfiguration` so ci-operator config files can specify Slack notifications directly on individual tests
- Per-test config takes precedence over `.config.prowgen` `slack_reporter`, which remains as fallback during the migration
- Applies defaults when not specified: `job_states_to_report: [success, failure, error]` and a standard report template matching the most common pattern in the release repo
- `slackReporterConfig()` returns `*prowv1.ReporterConfig` instead of mutating the job base
- For periodic tests with `presubmit: true`, the presubmit job does not get the slack config by default — set `report_presubmit: true` on the reporter config to include it
- Adds validation: `channel` is required, `report_presubmit` can only be set when `presubmit: true`
- Part of the incremental `.config.prowgen` deprecation

## Summary by CodeRabbit

* **New Features**
  * Slack notifications for CI jobs: Report job states (success, failure, error) to Slack channels with customizable message templates.
  * Per-job Slack configuration: Enable reporting independently for different job types (periodic, presubmit, postsubmit).

* **Tests**
  * Added integration test configurations demonstrating Slack reporting across multiple job scenarios.


Prucek:
/override ci/prow/images
/override ci/prow/e2e
We don't have to run them, they are going to fail anyway -_-
/skip

jmguzik (approved):

PR #5160, opened 2026-05-06
release-import optional reference_policy, payload tags, Source digest wait

deepsm007:
cc @Prucek @openshift/test-platform


pkg/steps/release/import_release.go
```diff
@@ -232,26 +233,24 @@ oc create configmap release-%s --from-file=%s.yaml=${ARTIFACT_DIR}/%s
...
-		}
 		existing := sets.New[string]()
 		tags := make([]imagev1.TagReference, 0, len(releaseIS.Spec.Tags)+len(stable.Spec.Tags))
 		for _, tag := range releaseIS.Spec.Tags {
 			existing.Insert(tag.Name)
-			tag.ReferencePolicy.Type = referencePolicy
-			tag.ImportPolicy.ImportMode = imagev1.ImportModePreserveOriginal
-			tags = append(tags, tag)
+			t := tag
+			mergePayloadReferencePolicy(&t, s.referencePolicy)
```
Prucek: python style object handling, Nikos complains about it all the time. Curious if coderabbit catches it, this time. Rather return the object, instead of mutating it inside

pkg/steps/release/import_release.go
```diff
@@ -361,6 +392,23 @@ func ImportReleaseStep(
...
+}
+
+// overrideCLIReferencePolicy matches legacy behavior for the optional non-amd64 CLI override stream.
+func (s *importReleaseStep) overrideCLIReferencePolicy() imagev1.TagReferencePolicyType {
+	if s.referencePolicy != nil {
+		return *s.referencePolicy
+	}
+	return imagev1.SourceTagReferencePolicy
+}
+
```
Prucek: I'd rename these to RefPolicyLocalIfUnset RefPolicySourceIfUnset, or somethig similar, becasue it has nothing to do with the CLI image. And I think they could be used in defaults as well. WDYT?

pkg/steps/utils/image.go
```diff
@@ -87,6 +93,10 @@ func FindStatusTag(is *imagev1.ImageStream, tag string) (*coreapi.ObjectReferenc
 
 const DefaultImageImportTimeout = 45 * time.Minute
 
+// ReleasePayloadImportTimeout bounds how long we wait for every tag of an imported release
+// payload (often 100+ images) to populate ImageStream status.
+const ReleasePayloadImportTimeout = 90 * time.Minute
```
Prucek: Isn't this too long?

pkg/api/reference_policy.go
```diff
@@ -0,0 +1,17 @@
+package api
```
droslean: It doesn't seem that the whole file is needed, since both of the functions do nothing. The logic can just be used directly.

pkg/defaults/defaults.go
```diff
@@ -149,9 +149,12 @@ func fromConfig(ctx context.Context, cfg *Config) ([]api.Step, []api.Step, error
 			if overrideCLIResolveErr != nil {
 				return nil, nil, results.ForReason("resolving_cli_override").ForError(fmt.Errorf("failed to resolve override CLI image for release %s: %w", resolveConfig.Name, overrideCLIResolveErr))
 			}
-			referencePolicy := imagev1.SourceTagReferencePolicy
-			if resolveConfig.Integration != nil && resolveConfig.Integration.ReferencePolicy != nil {
-				referencePolicy = *resolveConfig.Integration.ReferencePolicy
+			var importReferencePolicy *imagev1.TagReferencePolicyType
+			switch {
```
droslean, line 153: This will allow importReferencePolicy to be nil.

pkg/steps/multi_stage/gen.go
```diff
@@ -283,6 +287,23 @@ func isKubeconfigNeeded(step *api.LiteralTestStep, opts *generatePodOptions) boo
 	return needsKubeconfig || opts.IsObserver
 }
 
+func (s *multiStageTestStep) resolveStepImage(stream, tag string) string {
```
droslean: You are eating all the errors in here.

pkg/steps/multi_stage/image_resolution_test.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/openshift/ci-tools/pkg/steps/loggingclient"
+)
+
+func init() {
+	if err := imagev1.AddToScheme(scheme.Scheme); err != nil {
+		panic(err)
+	}
+}
+
+func TestResolveStepImage(t *testing.T) {
```
droslean: You literally don't test anything here. The resolveStepImage() uses another helper to resolve the name. And since you are passing the fake client, there is no point of the test. Which means that the function itself has a bad logic included.

pkg/steps/release/import_release.go
```diff
@@ -333,10 +329,34 @@ func (s *importReleaseStep) Objects() []ctrlruntimeclient.Object {
...
 }
 
+func payloadTagImpliesSource(tag *imagev1.TagReference) bool {
+	if tag.Reference {
+		return true
+	}
+	return tag.ImportPolicy.ImportMode == imagev1.ImportModePreserveOriginal
+}
+
+func mergedPayloadReferencePolicy(tag imagev1.TagReference, config *imagev1.TagReferencePolicyType) imagev1.TagReference {
```
droslean: This shouldn't be happening here. By the time you construct this step, the import policy should be clear.

pkg/steps/utils/image.go
```diff
@@ -87,6 +90,8 @@ func FindStatusTag(is *imagev1.ImageStream, tag string) (*coreapi.ObjectReferenc
 
 const DefaultImageImportTimeout = 45 * time.Minute
 
+const ReleasePayloadImportTimeout = 45 * time.Minute
```
droslean: This is duplicated.

pkg/steps/utils/image_test.go
```diff
@@ -27,6 +27,77 @@ func init() {
 	}
 }
 
+func TestImageDigestFor_SourceUsesSpecDigestWhenStatusIncomplete(t *testing.T) {
```
droslean: All tests here are wrong. Can you reformat them in `TestImageDigestFor()`. There is no point of adding the test case name as the name of the function. I guess I am replying to your AI now, which I suggest you to add some extra rules in your agent.

pkg/util/imagestream.go
```diff
@@ -29,8 +30,80 @@ func ParseImageStreamTagReference(s string) (api.ImageStreamTagReference, error)
 	return ret, nil
 }
 
-// ResolvePullSpec if a tag of an imagestream is resolved
+func findSpecTagReference(is *imageapi.ImageStream, tag string) *imageapi.TagReference {
+	for i := range is.Spec.Tags {
```
droslean: I don't know why AI is doing this......In Go, the iterations expose the element directly. You don't need to use the index.

pkg/util/imagestream.go
```diff
@@ -29,8 +30,80 @@ func ParseImageStreamTagReference(s string) (api.ImageStreamTagReference, error)
...
+		if is.Spec.Tags[i].Name == tag {
+			return &is.Spec.Tags[i]
+		}
+	}
+	return nil
+}
+
+func sourcePolicyAllowsDigestOnlyStatus(policy imageapi.TagReferencePolicyType) bool {
```
droslean: No reason at all of having this function. (only in Python I guess :P)

pkg/util/imagestream.go
```diff
@@ -29,8 +30,80 @@ func ParseImageStreamTagReference(s string) (api.ImageStreamTagReference, error)
...
+	return policy == imageapi.SourceTagReferencePolicy
+}
+
+func specTagImpliesSourceHints(tag *imageapi.TagReference) bool {
```
droslean: and this too..

pkg/util/imagestream.go
```diff
@@ -29,8 +30,80 @@ func ParseImageStreamTagReference(s string) (api.ImageStreamTagReference, error)
...
+func tagImportExplicitSuccess(conditions []imageapi.TagEventCondition) bool {
+	for _, c := range conditions {
+		if c.Type == imageapi.ImportSuccess && c.Status == corev1.ConditionTrue {
+			return true
+		}
+	}
+	return false
+}
+
+func dockerDigestPullSpecFromSpecTag(is *imageapi.ImageStream, tag string) string {
```
droslean: I really don't get why ci-operator should care about this. Can you try to elaborate more? Because this looks like a hack.

pkg/util/imagestream_test.go
```diff
@@ -0,0 +1,230 @@
+package util
```
droslean, line 1: Same here. All tests are wrong.

pkg/api/types.go
```diff
@@ -362,6 +362,8 @@ type UnresolvedRelease struct {
 	Prerelease *Prerelease `json:"prerelease,omitempty"`
 	// Release describes a released payload
 	Release *Release `json:"release,omitempty"`
+	// ReferencePolicy sets payload tag policy.
+	ReferencePolicy *imagev1.TagReferencePolicyType `json:"reference_policy,omitempty"`
```
droslean: This looks like its in the wrong place based on your comment.

droslean (changes requested):

PR #5163, opened 2026-05-06
🤖 TRT-2622: test result aggregation: properties

sosiouxme:
The aggregator now propagates properties of the testcases aggregated, choosing the properties of the first testcase instance as representative in the aggregated testcase.

🤖 Assisted by Claude Code

- What changed (practical effect): The test-result aggregator now preserves testcase-level metadata when combining multiple JUnit test runs. For each aggregated testcase it will adopt the lifecycle attribute and the Properties list from the first observed testcase instance and will not overwrite them on subsequent aggregations. Censoring was extended so property names and values are redacted alongside suite/case names, messages and stdout/stderr.

- Component(s) affected: jobrunaggregator (jobrunaggregatoranalyzer) and the junit package used to parse/manipulate JUnit XML.

- Behavioral impact for CI users/operators:
  - Aggregated JUnit emitted by the aggregator will now include testcase-level lifecycle attributes and properties (e.g., lifecycle, owner) taken from the first seen testcase. This metadata becomes visible to downstream consumers (reporting, BigQuery, dashboards).
  - Property values are subject to the same secret-censoring as other fields, so secrets in property names/values will be redacted when CensorTestSuite is applied.
  - Because the aggregator treats the first-seen testcase as authoritative for lifecycle and properties, differing property values across runs will not be merged — later values are ignored. Operators whose tooling expects merged or last-seen property semantics should update their consumers or aggregation workflow.

- Key implementation notes (what changed in code/tests):
  - Introduced junit.Property and added TestCase.Properties ([]*Property) and TestCase.Lifecycle fields; TestSuite.Properties now uses []*Property.
  - aggregateTestCase updated to copy Lifecycle into combined if combined.Lifecycle is empty and to copy Properties into combined if combined has no properties yet.
  - CensorTestSuite updated to censor Property.Name and Property.Value and replaced the previous per-value helper with censorStr to redact multiple string pointers in-place.
  - Unit tests and fixtures updated: new test TestAggregateTestCasePropagatesProperties verifies propagation and non-overwrite behavior; junit censor tests and YAML fixtures updated to include case-level properties; example e2e JUnit artifacts were adjusted to include explicit <properties> blocks.

- Operator considerations:
  - Expect testcase <properties> elements and lifecycle attributes to appear in aggregated JUnit outputs. Update any downstream parsing or dashboards that assume properties are absent or that properties should be merged/updated across runs.
  - The change introduced a new exported type name (Property); downstream consumers importing junit types should review compatibility if they relied on the prior TestSuiteProperty type name.

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
@@ -496,14 +496,13 @@ func TestAggregateTestCasePropagatesProperties(t *testing.T) {
 	assert.Equal(t, 2, len(combined.Properties))
 	assert.Equal(t, "lifecycle", combined.Properties[0].Name)
 	assert.Equal(t, "informing", combined.Properties[0].Value)
-	assert.Equal(t, "owner", combined.Properties[1].Name)
-	assert.Equal(t, "team-platform", combined.Properties[1].Value)
 
-	// Second aggregation should not overwrite the properties
+	// A different second aggregation should not overwrite the properties
+	// (Usually the important properties would be the same anyway, but where properties differ, just one gets picked)
 	source2 := &junit.TestCase{
 		Name: "test-with-properties",
 		Properties: []*junit.Property{
-			{Name: "lifecycle", Value: "informing"},
+			{Name: "lifecycle", Value: "blocking"},
 		},
 	}
 	err = aggregateTestCase("suite", combined, "logs/job", "run2", source2)
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
@@ -480,6 +480,40 @@ func TestAggregateTestCasePropagatesLifecycle(t *testing.T) {
 	assert.Equal(t, "informing", combined.Lifecycle)
 }
 
+func TestAggregateTestCasePropagatesProperties(t *testing.T) {
+	combined := &junit.TestCase{Name: "test-with-properties"}
+
+	source := &junit.TestCase{
+		Name: "test-with-properties",
+		Properties: []*junit.Property{
+			{Name: "lifecycle", Value: "informing"},
+			{Name: "owner", Value: "team-platform"},
+		},
+	}
+
+	err := aggregateTestCase("suite", combined, "logs/job", "run1", source)
+	assert.NoError(t, err)
+	assert.Equal(t, 2, len(combined.Properties))
+	assert.Equal(t, "lifecycle", combined.Properties[0].Name)
+	assert.Equal(t, "informing", combined.Properties[0].Value)
+	assert.Equal(t, "owner", combined.Properties[1].Name)
+	assert.Equal(t, "team-platform", combined.Properties[1].Value)
+
+	// Second aggregation should not overwrite the properties
+	source2 := &junit.TestCase{
+		Name: "test-with-properties",
+		Properties: []*junit.Property{
+			{Name: "lifecycle", Value: "informing"},
+		},
+	}
+	err = aggregateTestCase("suite", combined, "logs/job", "run2", source2)
+	assert.NoError(t, err)
+	// Properties should remain unchanged from first aggregation
+	assert.Equal(t, 2, len(combined.Properties))
+	assert.Equal(t, "lifecycle", combined.Properties[0].Name)
+	assert.Equal(t, "informing", combined.Properties[0].Value)
+}
+
 func TestInformingTestFailureMessage(t *testing.T) {
 	suite := &junit.TestSuites{
 		Suites: []*junit.TestSuite{
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
@@ -235,5 +235,16 @@ func aggregateTestCase(testSuiteName string, combined *junit.TestCase, jobGCSBuc
 		combined.Lifecycle = toAdd.Lifecycle
 	}
 
+	// Propagate properties from source test case to the combined one
+	if len(combined.Properties) == 0 && len(toAdd.Properties) > 0 {
+		combined.Properties = make([]*junit.Property, len(toAdd.Properties))
+		for i, prop := range toAdd.Properties {
+			combined.Properties[i] = &junit.Property{
+				Name:  prop.Name,
+				Value: prop.Value,
+			}
+		}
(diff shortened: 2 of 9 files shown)


pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
```diff
@@ -235,5 +235,16 @@ func aggregateTestCase(testSuiteName string, combined *junit.TestCase, jobGCSBuc
...
+	// Propagate properties from source test case to the combined one
+	if len(combined.Properties) == 0 && len(toAdd.Properties) > 0 {
+		combined.Properties = make([]*junit.Property, len(toAdd.Properties))
+		for i, prop := range toAdd.Properties {
+			combined.Properties[i] = &junit.Property{
+				Name:  prop.Name,
+				Value: prop.Value,
+			}
+		}
+	}
```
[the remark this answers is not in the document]
sosiouxme: It is not important that it be the first testcase, I just needed to point out that it's a representative, not some kind of amalgamation of all the test cases.

pkg/junit/censor_test.go
```diff
@@ -13,7 +13,7 @@ func TestCensorTestSuite(t *testing.T) {
 	censorer.Refresh("secret")
 	input := TestSuite{
 		Name: "some secret",
-		Properties: []*TestSuiteProperty{
+		Properties: []*Property{
 			{Name: "secret things", Value: "secret values"},
 			{Name: "also secret things", Value: "really secret values"},
```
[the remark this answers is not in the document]
sosiouxme: yes, while i don't expect any secret data in there, it's good to apply the censor just in case that assumption is wrong in the future. i added a commit to censor the values - the name should be coming from code not environment.

pkg/junit/censor.go
```diff
@@ -14,6 +14,9 @@ func CensorTestSuite(censor secretutil.Censorer, testSuite *TestSuite) {
 	}
 	for i := range testSuite.TestCases {
 		testSuite.TestCases[i].Name = censored(censor, testSuite.TestCases[i].Name)
+		for j := range testSuite.TestCases[i].Properties {
+			testSuite.TestCases[i].Properties[j].Value = censored(censor, testSuite.TestCases[i].Properties[j].Value)
+		}
```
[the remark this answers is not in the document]
sosiouxme: there's no way this will panic, but the way the loop is written offends me so i'll rewrite it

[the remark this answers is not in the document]
sosiouxme: censoring the name is excessive and might even lead to weird bugs if someone has a "secret" that matches a name, but this will probably be clearer if it's consistent, so we'll let those hypothetical bugs be a problem for future devs.

sosiouxme:
/override ci/prow/images

it's not me breaking it

sosiouxme:
/test e2e
according to https://redhat.atlassian.net/browse/DPTP-4861 from ship-help it's just a matter of which build cluster the job lands on (broken on build-05,6,7). would be nice to see it succeed.

sosiouxme:
note https://redhat.atlassian.net/browse/DPTP-4861 and https://redhat.atlassian.net/browse/DPTP-4865

sosiouxme:
> **Actionable comments posted: 2**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > pkg/junit/types.go (1)> `41-55`: _⚠️ Potential issue_ | _🟠 Major_ | _⚡ Quick win_
> > > **Keep a compatibility alias for `TestSuiteProperty`.**
> > > `pkg/junit` is already consumed from `openshift/release`, and those callers still construct `[]*junit.TestSuiteProperty`. Replacing the exported type name here turns this into a cross-repo compile break even though the XML shape is effectively the same. A temporary alias keeps this PR safe while downstream code migrates.

`openshift/release` is not consuming this `TestSuiteProperty`; it makes its own in two different places and uses those. It's a pretty trivial type, it would be silly to import the module just for that. If anybody really is, they'll notice next time they bump the module and it'll be an easy fix...

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
```diff
@@ -480,6 +480,40 @@ func TestAggregateTestCasePropagatesLifecycle(t *testing.T) {
...
+		Properties: []*junit.Property{
+			{Name: "lifecycle", Value: "informing"},
+		},
+	}
+	err = aggregateTestCase("suite", combined, "logs/job", "run2", source2)
+	assert.NoError(t, err)
+	// Properties should remain unchanged from first aggregation
+	assert.Equal(t, 2, len(combined.Properties))
+	assert.Equal(t, "lifecycle", combined.Properties[0].Name)
+	assert.Equal(t, "informing", combined.Properties[0].Value)
```
smg247, line 513: Why are we not checking the `owner` still? I also don't see how we verified this didn't overwrite since `lifecycle` is set to informing in both cases.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
```diff
@@ -235,5 +235,16 @@ func aggregateTestCase(testSuiteName string, combined *junit.TestCase, jobGCSBuc
 		combined.Lifecycle = toAdd.Lifecycle
 	}
```
smg247, line 236: is this going to result in duplicated lifecycle, or is that just a poor example from the unit test?

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
sosiouxme: Nope, you're right. It's not testing what it's supposed to.
TBH I let Claude write the test cases and didn't much examine them.
Also TBH this part of the unit test seems like overkill. Claude seems to have gotten overly focused on a minor detail. I don't really care which representative test case supplies the properties, they should all be the same anyway.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/junit.go
sosiouxme: that's just setting the attribute, which is technically independent of the properties.

pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer_test.go
smg247: Makes sense to just remove this one then

petr-muller:
fwiw lgtm

sosiouxme:
this is ridiculous, it succeeded already.
/override ci/prow/images

PR #5171, opened 2026-05-11
ci-operator-checkconfig: Add cluster profile sets allowlist

danilo-gemoli:
Revisit the Cluster Profile Sets policy that has been introduced with https://github.com/openshift/ci-tools/pull/5087 by adding a whitelist that allows some tests to use the underlying cluster profile rather than the cluster profile set.

With the actual policy, the following test in `ci-operator/config/openshift/openshift-tests-private/openshift-openshift-tests-private-release-4.22__multi-nightly.yaml` doesn't pass the validation since it is using `aws` rather than `openshift-org-aws`:
```yaml
tests:
- as: aws-ipi-public-ipv4-pool-byo-subnet-mini-perm-arm-f14
  cron: 33 8 4,18 * *
  steps:
    cluster_profile: aws
```

After we merge this PR along with another one in `openshift/release` that would produce an allowlist like the following one, the test will pass the validation:
```json
{
  "cluster_profile_sets": {
    "openshift-org-aws": [
      "aws",
      "aws-2",
      "aws-3",
      "aws-4",
      "aws-5"
    ],
  "tests_allowlist": {
    "openshift/openshift-tests-private": {
      "release-4.22": {
        "multi-nightly": [
          "aws-ipi-public-ipv4-pool-byo-subnet-mini-perm-arm-f14"
        ]
      }
    }
  }
}
```

The `tests_allowlist` consists of:
```json
{
  "tests_allowlist" : {
     "${ORG}/${REPO}": {   // Regex allowed
       "${BRANCH}": {      // Regex allowed
         "${VARIANT}": [   // Regex allowed
           "${TEST_NAME}"  // Plain text, NO regex allowed
         ]
       }
     }
  }
}
```

The type `ClusterProfileSetDetailsNew` is temporary and will be deleted once we merge a PR in `openshift/release` that will modify the `cluster-profile-set-details.json` schema to accomodate the new `tests_whitelist` stanza.
The combination of both `ClusterProfileSetDetails` and `ClusterProfileSetDetailsNew` makes sure we are backward compatible until the migration is complete.

## Summary by CodeRabbit

* **New Features**
  * Added a regex-based allowlist for cluster profile set usage so specific tests can continue using legacy cluster profiles.
  * Backwards-compatible loading of both legacy and new allowlist JSON shapes.
* **Validation**
  * Validation now enforces per-test allowlisting, suggests appropriate cluster profile sets, and skips allowlisted tests.
* **Tests**
  * Expanded unit tests covering legacy/new schemas, allowlist behavior, and regexp utilities.

### DIFF
--- a/cmd/ci-operator-checkconfig/main.go
+++ b/cmd/ci-operator-checkconfig/main.go
@@ -1,7 +1,6 @@
 package main
 
 import (
-	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
@@ -35,7 +34,7 @@ type options struct {
 	ciOPConfigAgent          agents.ConfigAgent
 	clusterProfiles          api.ClusterProfilesMap
 	clusterClaimOwners       api.ClusterClaimOwnersMap
-	clusterProfileSetDetails validation.ClusterProfileSetDetails
+	clusterProfileSetDetails api.ClusterProfileSetDetails
 }
 
 func (o *options) parse() error {
@@ -79,9 +78,11 @@ func (o *options) parse() error {
 	o.ciOPConfigAgent = ciOPConfigAgent
 
 	if clusterProfileSetDetailsPath != "" {
-		if err := o.loadClusterProfileDetails(clusterProfileSetDetailsPath); err != nil {
+		cpsd, err := load.ClusterProfileSetDetails(clusterProfileSetDetailsPath)
+		if err != nil {
 			return fmt.Errorf("load cluster profile set details: %w", err)
 		}
+		o.clusterProfileSetDetails = cpsd
 	}
 
 	if err := o.Options.Validate(); err != nil {
@@ -130,20 +131,6 @@ func (o *options) validate() (ret []error) {
 		ret = append(ret, err)
 	}
 	return append(ret, validateTags(seen)...)
-}
-
-func (o *options) loadClusterProfileDetails(p string) error {
-	cpsDetailsBytes, err := os.ReadFile(p)
-	if err != nil {
-		return fmt.Errorf("read file %s: %w", p, err)
-	}
-
-	o.clusterProfileSetDetails = make(map[api.ClusterProfile][]string)
-	if err := json.Unmarshal(cpsDetailsBytes, &o.clusterProfileSetDetails); err != nil {
-		return fmt.Errorf("unmarshal cluster profile set details: %w", err)
-	}
-
-	return nil
 }
 
 func (o *options) loadResolver(path string) error {
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -1,6 +1,7 @@
 package api
 
 import (
+	"encoding/json"
 	"fmt"
 	"slices"
 	"strings"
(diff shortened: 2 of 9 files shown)


Prucek (approved):

petr-muller:
1. Can we call it "allowlist" instead of "whitelist"? 
2. I'm worried about the maintenance of a separate allowlist long-term, e.g. in events such as branch cuts but also copies in `openshift-priv`. Do we plan to extend the tools that propagate jobs through the system to manage the allowlist too, or is it expected to be maintained manually?

danilo-gemoli:
> Can we call it "allowlist" instead of "whitelist"?  

Sure.

> I'm worried about the maintenance of a separate allowlist long-term, e.g. in events such as branch cuts but also copies in openshift-priv. Do we plan to extend the tools that propagate jobs through the system to manage the allowlist too, or is it expected to be maintained manually?  

Good point. The `allowlist` is manually maintained in and generated by `core-services/prow/02_config/generate-boskos.py`. The output looks like what follows:
```json
{
  "cluster_profile_sets": {
    "openshift-org-aws": [
      "aws",
      "aws-2",
      "aws-3",
      "aws-4",
      "aws-5"
    ],
    "openshift-org-azure": [
      "azure-2",
      "azure4"
    ],
    "openshift-org-gcp": [
      "gcp",
      "gcp-arm64",
      "gcp-openshift-gce-devel-ci-2",
      "gcp-3"
    ]
  },
  "tests_whitelist": {
    "openshift/openshift-tests-private": {
      "release-4.22": {
        "multi-nightly": [
          "aws-ipi-public-ipv4-pool-byo-subnet-mini-perm-arm-f14"
        ]
      }
    }
  }
}
```

I can think of several approaches to address your, very valid, concerns:

1. Regex:
```json
"tests_whitelist": {
  "openshift(-priv)?/openshift-tests-private": { // openshift or openshift-priv
    ".+": { // Any branch
      ".*": [ // Any or no variant
        "aws-ipi-public-ipv4-pool-byo-subnet-mini-perm-arm-f14"
      ]
    }
  }
}
```

2. Extend the `cluster_profile` stanza to something like this:
```yaml
tests:
- as: aws-ipi-public-ipv4-pool-byo-subnet-mini-perm-arm-f14
  cron: 33 8 4,18 * *
  steps:
    cluster_profile:
      name: aws
      enforce_cluster_profile_set_policy: false
```

3. When performing the validation, raise a warning to the user rather than a failure.

4. Define both Cluster Profile Sets and the allowlist outside of `core-services/prow/02_config/generate-boskos.py` and modify our tools such that they will keep the new config up to date.

To me, the least complex that keeps enforcing the policy and requires little to no long term, manual, maintenance is (1).
WDYT?

petr-muller:
I'm not a greatest fan of regexp in configuration files but here it seems like a pragmatic choice. I think *eventually* someone will ask you for a regex in the job names too (the list items) ;)

/hold cancel

danilo-gemoli:
> I'm not a greatest fan of regexp

Neither am I.

> someone will ask you for a regex in the job names too

and I won't do anything until then because:

> I'm not a greatest fan of regexp

Prucek (approved):

PR #5182, opened 2026-05-13
DPTP-2938: pod-scaler add --max-data-age flag to cap cache retention and query range

deepsm007:
/cc @openshift/test-platform 

## pod-scaler: Add configurable --max-data-age to cap cache retention and query range

This PR adds a new --max-data-age flag to the pod-scaler producer so operators can cap how far back Prometheus is queried and how long cached Prometheus data is retained.

Practical changes for CI operators
- New flag: --max-data-age (type duration). Default is 90d (90*24*time.Hour) when not specified.
- Query behavior: When >0, Prometheus query ranges are clamped to not go further back than now - maxDataAge (i.e., the producer will not request data older than the configured age). If max-data-age is set to 0 (explicitly), the clamping code is skipped; the producer relies on Prometheus's native retention.
- Cache pruning: Cached results are pruned using the supplied max-data-age via a new PruneWithMaxAge(maxAge time.Duration) API on the cached query type; storeCache calls this method before persisting cache data.

Component and impact
- Affects the pod-scaler producer and its caching: controls both the Prometheus query range and the cache prune window.
- Default behavior unchanged for typical installs (90 days). Operators can reduce the data footprint and query/storage load by setting a smaller max-data-age, or keep longer history by increasing it.
- Note: the code paths do not implement a separate fallback when max-data-age <= 0 (the default prevents that). Operators setting max-data-age to 0 should be aware that query clamping is disabled and the prune call receives 0 (i.e., pruning behavior is determined entirely by the new PruneWithMaxAge implementation).

Why this matters
- Provides explicit, operator-controlled tradeoff between historical coverage and resource usage (query volume, cache size, and storage costs).

### DIFF
--- a/pkg/pod-scaler/types.go
+++ b/pkg/pod-scaler/types.go
@@ -335,14 +335,9 @@ func coalesceOnce(input []TimeRange) []TimeRange {
 	return input
 }
 
-// Prune ensures that no identifying set of labels contains more than twenty-five entries,
-// as well as removing any data that was added more than 90 days ago.
-// We know that an entry fingerprint can only exist for one fully-qualified label set,
-// but if the label set contains a multi-stage step, it will also be referenced in
-// the additional per-step index.
-func (q *CachedQuery) Prune() {
-	ninetyDaysAgo := time.Now().Add(-90 * 24 * time.Hour)
-	q.prune(ninetyDaysAgo)
+// PruneWithMaxAge removes data older than maxAge and caps each label set to 25 entries.
+func (q *CachedQuery) PruneWithMaxAge(maxAge time.Duration) {
+	q.prune(time.Now().Add(-maxAge))
 }
 
 func (q *CachedQuery) prune(pruneBefore time.Time) {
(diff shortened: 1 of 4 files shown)


pkg/pod-scaler/types.go
```diff
@@ -335,14 +335,11 @@ func coalesceOnce(input []TimeRange) []TimeRange {
...
 
-// Prune ensures that no identifying set of labels contains more than twenty-five entries,
-// as well as removing any data that was added more than 90 days ago.
-// We know that an entry fingerprint can only exist for one fully-qualified label set,
-// but if the label set contains a multi-stage step, it will also be referenced in
-// the additional per-step index.
-func (q *CachedQuery) Prune() {
-	ninetyDaysAgo := time.Now().Add(-90 * 24 * time.Hour)
-	q.prune(ninetyDaysAgo)
+const DefaultMaxDataAge = 90 * 24 * time.Hour
```
droslean: Since this is being controlled by a flag, the default value should go there.

deepsm007: fixed

droslean (approved):

deepsm007:
/override ci/prow/e2e
failures not related to PR changes

PR #5189, opened 2026-05-19
Added labels support for ci-secret-bootstrap

hector-vido:
Since we are using ci-secret-bootstrap to manage ArgoCD cluster secrets we need a way specify labels inside the secret stored on vault to avoid manual work.

ArgoCD identify cluster secrets based on the label `argocd.argoproj.io/secret-type: cluster` added to the secrets that maps a `ServiceAccount` and a **token** from other clusters inside `openshift-gitops`.

This PR add this possibility.

## Labels support for ci-secret-bootstrap

This PR enables ci-secret-bootstrap to apply Kubernetes labels to Secrets it syncs from Vault so operators can manage label-dependent workflows (notably ArgoCD cluster-secret discovery) automatically.

What changed (practical impact)
- ci-secret-bootstrap (cmd/ci-secret-bootstrap)
  - Recognizes a new Vault metadata key (pkg/api/vault.SecretSyncTargetLabelsKey = "secretsync/target-labels") that contains comma-separated "key: value" label pairs.
  - Parses, trims and validates label keys and values against Kubernetes label rules; invalid entries are skipped and logged.
  - Merges validated labels with the existing ci-secret-bootstrap requester label and writes them to Secret.ObjectMeta.Labels when creating or updating Secrets.
  - Treats the secretsync/target-labels Vault key as reserved metadata so it is not stored in Secret.Data.
  - When deciding whether to update an existing Secret, compares desired vs existing ObjectMeta.Labels in addition to Secret.Data and performs in-place updates when labels differ, honoring the existing requester-label guard.

- vault-subpath-proxy (cmd/vault-subpath-proxy)
  - KV update request validation now treats the new secretsync/target-labels key as metadata (like secretsync/target-clusters) so it bypasses normal secret-key regex validation and is accepted as a metadata field.

Why this matters for CI operators
- Operators can store desired Kubernetes labels in Vault and have ci-secret-bootstrap apply them during sync. This eliminates manual post-sync labeling for workflows that depend on labels (for example adding argocd.argoproj.io/secret-type: cluster so ArgoCD discovers cluster secrets automatically).

Technical notes
- New exported constant: pkg/api/vault.SecretSyncTargetLabelsKey = "secretsync/target-labels".
- No public function signatures changed; go.mod dependency versions were updated as part of the commit.

### DIFF
--- a/cmd/ci-secret-bootstrap/main.go
+++ b/cmd/ci-secret-bootstrap/main.go
@@ -671,19 +671,29 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 				secretsMap[cluster] = map[types.NamespacedName]coreapi.Secret{}
 			}
 			entry, alreadyExists := secretsMap[cluster][secretName]
+			labels, err := generateUserSecretLabels(secretKeys)
+			if err != nil {
+				errs = append(errs, err)
+				continue
+			}
+			labels[api.DPTPRequesterLabel] = api.CISecretBootstrapName
 			if !alreadyExists {
 				entry = coreapi.Secret{
-					ObjectMeta: metav1.ObjectMeta{Namespace: secretName.Namespace, Name: secretName.Name, Labels: map[string]string{api.DPTPRequesterLabel: "ci-secret-bootstrap"}},
-					Data:       map[string][]byte{},
-					Type:       coreapi.SecretTypeOpaque,
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: secretName.Namespace,
+						Name:      secretName.Name,
+					},
+					Data: map[string][]byte{},
+					Type: coreapi.SecretTypeOpaque,
 				}
 			}
+			entry.ObjectMeta.Labels = labels
 			if entry.Type != coreapi.SecretTypeOpaque {
 				errs = append(errs, fmt.Errorf("secret %s in cluster %s has ci-secret-bootstrap config as non-opaque type and is targeted by user sync from key %s", secretName.String(), cluster, secretKeys[vaultapi.VaultSourceKey]))
 				continue
 			}
 			for vaultKey, vaultValue := range secretKeys {
-				if vaultKey == vaultapi.SecretSyncTargetClusterKey {
+				if vaultKey == vaultapi.SecretSyncTargetClusterKey || vaultKey == vaultapi.SecretSyncTargetLabelsKey {
 					continue
 				}
 				if _, alreadyExists := entry.Data[vaultKey]; alreadyExists {
@@ -698,6 +708,26 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 	}
 
 	return secretsMap, utilerrors.NewAggregate(errs)
+}
+
+func generateUserSecretLabels(secretKeys map[string]string) (map[string]string, error) {
+	labels := map[string]string{}
+	raw, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]
+	if !ok {
+		return labels, nil
+	}
+	for _, label := range strings.Split(raw, ",") {
+		label = strings.TrimSpace(label)
+		if label == "" {
+			continue
+		}
+		key, value, ok := strings.Cut(label, ":")
+		if !ok {
+			return nil, fmt.Errorf("invalid label %q: expected key:value", label)
+		}
+		labels[key] = value
+	}
+	return labels, nil
 }
 
 type Getter interface {
@@ -802,6 +832,7 @@ func updateSecrets(getters map[string]Getter, secretsMap map[string][]*coreapi.S
 
 				if !shouldCreate {
 					differentData := !equality.Semantic.DeepEqual(secret.Data, existingSecret.Data)
+					differentLabels := !equality.Semantic.DeepEqual(secret.ObjectMeta.Labels, existingSecret.ObjectMeta.Labels)
 					var addedKeys, changedKeys, removedKeys []string
 					for k, value := range secret.Data {
 						if existingValue, ok := existingSecret.Data[k]; !ok {
@@ -821,7 +852,7 @@ func updateSecrets(getters map[string]Getter, secretsMap map[string][]*coreapi.S
 						errs = append(errs, fmt.Errorf("secret %s:%s/%s needs updating in place (%s), use --force to do so", cluster, secret.Namespace, secret.Name, change))
 						continue
 					}
-					if existingSecret.Labels == nil || existingSecret.Labels[api.DPTPRequesterLabel] != "ci-secret-bootstrap" || differentData {
+					if existingSecret.Labels == nil || existingSecret.Labels[api.DPTPRequesterLabel] != "ci-secret-bootstrap" || differentData || differentLabels {
 						if _, err := secretClient.Update(context.TODO(), secret, metav1.UpdateOptions{DryRun: dryRunOptions}); err != nil {
 							errs = append(errs, fmt.Errorf("error updating secret %s:%s/%s: %w", cluster, secret.Namespace, secret.Name, err))
 							continue
--- a/cmd/ci-secret-bootstrap/main_test.go
+++ b/cmd/ci-secret-bootstrap/main_test.go
@@ -31,6 +31,7 @@ import (
 	"github.com/openshift/ci-tools/pkg/api"
 	"github.com/openshift/ci-tools/pkg/api/secretbootstrap"
 	"github.com/openshift/ci-tools/pkg/api/secretgenerator"
+	vaultapi "github.com/openshift/ci-tools/pkg/api/vault"
 	gsm "github.com/openshift/ci-tools/pkg/gsm-secrets"
 	"github.com/openshift/ci-tools/pkg/secrets"
 	"github.com/openshift/ci-tools/pkg/testhelper"
@@ -1017,6 +1018,7 @@ Code: 404. Errors:
 					Data: map[string]string{
 						"secretsync/target-namespace": "some-namespace",
 						"secretsync/target-name":      "some-name",
+						"secretsync/target-labels":    "label:value",
 						"some-data-key":               "a-secret",
 					},
 				},
@@ -1024,7 +1026,7 @@ Code: 404. Errors:
 			config: secretbootstrap.Config{UserSecretsTargetClusters: []string{"a", "b"}},
 			expected: map[string][]*coreapi.Secret{
 				"a": {{
-					ObjectMeta: metav1.ObjectMeta{Namespace: "some-namespace", Name: "some-name", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap"}},
+					ObjectMeta: metav1.ObjectMeta{Namespace: "some-namespace", Name: "some-name", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap", "label": "value"}},
 					Type:       coreapi.SecretTypeOpaque,
 					Data: map[string][]byte{
 						"some-data-key":                []byte("a-secret"),
@@ -1032,7 +1034,7 @@ Code: 404. Errors:
 					},
 				}},
 				"b": {{
-					ObjectMeta: metav1.ObjectMeta{Namespace: "some-namespace", Name: "some-name", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap"}},
+					ObjectMeta: metav1.ObjectMeta{Namespace: "some-namespace", Name: "some-name", Labels: map[string]string{"dptp.openshift.io/requester": "ci-secret-bootstrap", "label": "value"}},
 					Type:       coreapi.SecretTypeOpaque,
 					Data: map[string][]byte{
 						"some-data-key":                []byte("a-secret"),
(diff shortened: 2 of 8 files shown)


cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +671,34 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 				secretsMap[cluster] = map[types.NamespacedName]coreapi.Secret{}
 			}
 			entry, alreadyExists := secretsMap[cluster][secretName]
+			labels := map[string]string{api.DPTPRequesterLabel: "ci-secret-bootstrap"}
+			if vaultValue, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]; ok {
+				if vaultValue != "" {
+					for _, label := range strings.Split(vaultValue, ",") {
+						label := strings.Split(label, ":")
+						labels[strings.Trim(label[0], " ")] = strings.Trim(label[1], " ")
+					}
```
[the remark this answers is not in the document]
hector-vido: This is an edge case and even if a user overwrite this label, this should not be a problem for ci-secret-bootstrap.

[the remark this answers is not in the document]
hector-vido: Using regex now

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -1336,13 +1338,41 @@ Code: 404. Errors:
...
+		},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			client := vaultClientFromTestItems(tc.items)
 
 			var actualErrorMsg string
+			if tc.name == "Usersecret with invalid label key" {
+				print("ok")
```
droslean: debuging leftover?

cmd/vault-subpath-proxy/kv_update_transport.go
```diff
@@ -122,7 +122,7 @@ func (k *kvUpdateTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 			}
 			continue
 		}
-		if key == vault.SecretSyncTargetClusterKey {
+		if key == vault.SecretSyncTargetClusterKey || key == vault.SecretSyncTargetLabelsKey {
 			continue
```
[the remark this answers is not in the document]
hector-vido: Done!

cmd/ci-secret-bootstrap/main.go
```diff
@@ -660,6 +661,8 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 	}
 
 	var errs []error
+	labelValueRegex := regexp.MustCompile("^(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?$")
```
droslean: Both regexes are smelly here. Is this an assumption on what is permitted and what is not? Also, since those labels are meant to be used in a Kubernetes object, then I am would bet that there is already a validation in place somewhere in the kube libraries.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +674,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 				secretsMap[cluster] = map[types.NamespacedName]coreapi.Secret{}
 			}
 			entry, alreadyExists := secretsMap[cluster][secretName]
+			labels := map[string]string{api.DPTPRequesterLabel: "ci-secret-bootstrap"}
+			if vaultValue, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]; ok {
```
droslean:
```suggestion
			if vaultValue, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]; ok &&  vaultValue != "" {
```

To avoid the extended nesting. Also, this is an assumption that the value could be empty. If the value can't be empty at all, we should have validated that in the first place.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +674,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
+				if vaultValue != "" {
+					for _, label := range strings.Split(vaultValue, ",") {
```
droslean: Here, you are assuming a specific format of the value. If this is how you want to deliver the feature, you need to:
- Validate the formatting when the user is applying it.
- Encapsulate the whole block to each own logic and testing. 
- It is Go idiomatic to wrap the whole thing under its own struct with the methods you require to operate on the values.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +674,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
+						label := strings.Split(label, ":")
+						if labelKeyRegex.MatchString(label[0]) && labelValueRegex.MatchString(label[1]) {
+							labels[strings.Trim(label[0], " ")] = strings.Trim(label[1], " ")
```
droslean: Trimming the value, is also an assumption and unnecessary correction of the wrong input of the user. This is a wrong approach.

hector-vido: Using only `validation.IsQualifiedName` and `validation.IsValidLabelValue` now.

cmd/ci-secret-bootstrap/main.go
hector-vido: Using `validation.IsQualifiedName` and `validation.IsValidLabelValue` inside the proxy.

cmd/ci-secret-bootstrap/main.go
hector-vido: Adding a `if len(strings.TrimSpace(value)) == 0 {}` on **kv_update_transport.go**.

cmd/ci-secret-bootstrap/main.go
hector-vido: Regex removed, using k8s lib validation.

cmd/vault-subpath-proxy/kv_update_transport.go
```diff
@@ -122,6 +122,14 @@ func (k *kvUpdateTransport) RoundTrip(r *http.Request) (*http.Response, error) {
 			}
 			continue
 		}
+		if key == vault.SecretSyncTargetLabelsKey {
+			if len(strings.TrimSpace(value)) == 0 {
```
droslean: You are mutating the real value here.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +672,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 				secretsMap[cluster] = map[types.NamespacedName]coreapi.Secret{}
 			}
 			entry, alreadyExists := secretsMap[cluster][secretName]
+			labels := map[string]string{api.DPTPRequesterLabel: "ci-secret-bootstrap"}
+			if vaultValue, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]; ok {
+				if vaultValue != "" {
```
droslean: This is already now validated, and also you don't want to check that.. If the value is empty, then we go with the empty value and fail in the next validation either way.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +672,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
+					for _, label := range strings.Split(vaultValue, ",") {
```
droslean: Let's get this logic out of here. Put it on its own function, or even better a method of the secretKeys if its easy to refactor. Either way, you need to have a separate unittests for that logic as well.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +672,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
+						label := strings.Split(label, ":")
+						if validation.IsQualifiedName(label[0]) == nil && validation.IsValidLabelValue(label[1]) == nil {
+							labels[strings.Trim(label[0], " ")] = strings.Trim(label[1], " ")
+						} else {
+							errs = append(errs, fmt.Errorf("ignoring label for secret %s, it is invalid: key '%s' value '%s'", secretName, label[0], label[1]))
```
droslean: I guess you are not ignoring anything now, the error message is wrong.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +672,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
-					Type:       coreapi.SecretTypeOpaque,
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: secretName.Namespace,
+						Name:      secretName.Name,
+						Labels:    labels,
+					},
+					Data: map[string][]byte{},
+					Type: coreapi.SecretTypeOpaque,
 				}
+			} else {
```
droslean: What you do inside this `else` is not related to what you are checking on if. Also, the existance check has to do with the objectMeta namespace and name, since we will get the existing secret. Either way, you will override the labels in the end.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +672,38 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
+				entry.ObjectMeta.Labels = labels
 			}
 			if entry.Type != coreapi.SecretTypeOpaque {
 				errs = append(errs, fmt.Errorf("secret %s in cluster %s has ci-secret-bootstrap config as non-opaque type and is targeted by user sync from key %s", secretName.String(), cluster, secretKeys[vaultapi.VaultSourceKey]))
 				continue
 			}
 			for vaultKey, vaultValue := range secretKeys {
-				if vaultKey == vaultapi.SecretSyncTargetClusterKey {
+				if vaultKey == vaultapi.SecretSyncTargetClusterKey || vaultKey == vaultapi.SecretSyncTargetLabelsKey {
```
droslean, line 696: I don't understand why you changed that. Can you elaborate?

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -1017,22 +1017,23 @@ Code: 404. Errors:
 					Data: map[string]string{
 						"secretsync/target-namespace": "some-namespace",
 						"secretsync/target-name":      "some-name",
+						"secretsync/target-labels":    "label:value",
```
droslean, line 1021: Since you want to support multiple labels, you need to test that too.

cmd/ci-secret-bootstrap/main.go
hector-vido: Since this key is not used to create an entry inside the `data` secret stanza, we need to skip it the same way we skip the target cluster vault data.

cmd/ci-secret-bootstrap/main.go
hector-vido: I can see a relation, if the secret exists I just overwrite the labels because someone could changed it for another value.

Should I just overwrite it, declaring an ObjectMeta without labels and after the if define it?

```go
if !alreadyExists {
    entry = coreapi.Secret{
        ObjectMeta: metav1.ObjectMeta{
            Namespace: secretName.Namespace,
            Name:      secretName.Name,
        },
        Data: map[string][]byte{},
        Type: coreapi.SecretTypeOpaque,
    }
}
entry.ObjectMeta.Labels = labels
```

cmd/vault-subpath-proxy/kv_update_transport.go
hector-vido: This happens during the put/patch, I am mutating, indeed, but this is the only way to check if this value is not empty or some blank string.
Why is this method a problem?

cmd/ci-secret-bootstrap/main.go
hector-vido: Ok I will avoid testing if the value is empty during secrets synchronization, but when this "next validation" will happen? Are you talking about `IsQualifiedName` and `IsValidLabelValue`?

cmd/ci-secret-bootstrap/main.go
```diff
@@ -700,6 +705,17 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
 	return secretsMap, utilerrors.NewAggregate(errs)
 }
 
+func generateUserSecretLabels(secretKeys map[string]string) map[string]string {
+	labels := map[string]string{api.DPTPRequesterLabel: "ci-secret-bootstrap"}
+	if vaultValue, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]; ok {
+		for _, label := range strings.Split(vaultValue, ",") {
+			parts := strings.Split(label, ":")
+			labels[parts[0]] = parts[1]
```
droslean: Well, we already validated the formating but this can panic.. Perhaps you can improve it like:

```go
func generateUserSecretLabels(secretKeys map[string]string) (map[string]string, error) {
	labels := map[string]string{api.DPTPRequesterLabel: "ci-secret-bootstrap"}
	raw, ok := secretKeys[vaultapi.SecretSyncTargetLabelsKey]
	if !ok {
		return labels, nil
	}
	for label := range strings.SplitSeq(raw, ",") {
		label = strings.TrimSpace(label)
		if label == "" {
			continue
		}
		key, value, ok := strings.Cut(label, ":")
		if !ok {
			return nil, fmt.Errorf("invalid label %q: expected key:value", label)
		}
		labels[key] = value
	}
	return labels, nil
}

```

Please also create a specific unittest for this.

cmd/ci-secret-bootstrap/main.go
```diff
@@ -671,19 +671,29 @@ func fetchUserSecrets(secretsMap map[string]map[types.NamespacedName]coreapi.Sec
...
-					Type:       coreapi.SecretTypeOpaque,
+					ObjectMeta: metav1.ObjectMeta{
+						Namespace: secretName.Namespace,
+						Name:      secretName.Name,
+					},
+					Data: map[string][]byte{},
+					Type: coreapi.SecretTypeOpaque,
 				}
 			}
+			entry.ObjectMeta.Labels = labels
```
droslean, line 690: You just put it when you define the struct in 682/683

cmd/ci-secret-bootstrap/main_test.go
```diff
@@ -4534,3 +4609,54 @@ func TestMergeSecretMaps(t *testing.T) {
...
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			label, err := generateUserSecretLabels(tc.secretKeys)
+
+			if err != nil && err.Error() != tc.expectedError {
+				t.Errorf("expected error: %q, got: %q", tc.expectedError, err.Error())
+			} else if tc.expectedError == "" {
+				equal(t, "labels", tc.expected, label)
```
droslean, line 4658: Just use cmp.Diff directly. Also there is no point on passing the testing.T in there.

pkg/api/constant.go
```diff
@@ -33,8 +33,9 @@ const (
 
 	ReleaseAnnotationSoftDelete = "release.openshift.io/soft-delete"
 
-	// DPTPRequesterLabel is the label on a Kubernates CR whose value indicates the automated tool that requests the CR
-	DPTPRequesterLabel = "dptp.openshift.io/requester"
+	// DPTPRequesterLabelKey is the label on a Kubernates CR whose value indicates the automated tool that requests the CR
+	DPTPRequesterLabelKey   = "dptp.openshift.io/requester"
```
droslean: Let's avoid this change. Just keep it as it was, to avoid changing a quadrillion references.

PR #5194, opened 2026-05-20
Add destroy previous secret versions functionality when updating a secret

psalajova:
I identified that GSM costs were getting high by accumulating secret versions (~150 per secret across ~550 secrets in the `test-platform-infra` collection), all using automatic replication. I ran a one-time cleanup script to destroy all non-latest versions, and modified CreateOrUpdateSecret to accept a destroyPreviousVersions flag — enabled for ci-secret-generator — so old versions are automatically destroyed after each update, preventing future cost buildup.

This PR adds an opt-in GSM secret-update path that destroys previous secret versions after a successful update and enables that path for the ci-secret-generator so old versions will not re-accumulate.

What it does (practical terms)
- New helper: CreateOrUpdateSecretDestroyingPreviousVersions
  - Performs the same create-or-update work as CreateOrUpdateSecret, then lists ENABLED secret versions for the secret, keeps the newest version (by CreateTime) and destroys every other enabled version via DestroySecretVersion.
  - Errors from listing or destroying are returned (cleanup is not best-effort).
  - Requires the GSM client to implement ListSecretVersions and DestroySecretVersion.

- Where it's used
  - The ci-secret-generator code path now calls CreateOrUpdateSecretDestroyingPreviousVersions from pkg/secrets/gsm.go (SetFieldOnItem and UpdateIndexSecret). As a result, routine periodic updates performed by ci-secret-generator will automatically remove prior enabled secret versions after each update.
  - Existing callers that still call CreateOrUpdateSecret are unaffected.

- Tests & mocks
  - Test fakes and GoMock-generated mocks were extended with ListSecretVersions and DestroySecretVersion so unit tests compile and can exercise the new behavior.
  - Some test fakes return nil iterators or a not-implemented error consistent with other test stubs.

Operational impact
- Prevents future accumulation of enabled GSM secret versions for secrets managed by ci-secret-generator, reducing GSM storage usage and cost (the author ran a one-time cleanup to remove existing non-latest versions).
- Because listing or destruction failures are surfaced as errors, callers that opt into this destructive cleanup (currently ci-secret-generator) may see failed GSM updates when cleanup fails and should handle those errors accordingly.

Compatibility
- Backwards-compatible for callers that do not opt in; CreateOrUpdateSecret remains available and unchanged for non-destructive updates.

### DIFF
--- a/pkg/gsm-secrets/execution.go
+++ b/pkg/gsm-secrets/execution.go
@@ -28,8 +28,7 @@ import (
 
 // CreateOrUpdateSecret creates a new secret in Google Secret Manager or updates an existing one with a new version.
 // If labels or annotations are nil, they won't be set on the secret.
-// destroyPreviousVersions controls whether old enabled versions are cleaned up after adding the new one.
-func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, projectIdNumber, secretName string, payload []byte, labels, annotations map[string]string, destroyPreviousVersions bool) error {
+func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, projectIdNumber, secretName string, payload []byte, labels, annotations map[string]string) error {
 	parent := GetProjectResourceIdNumber(projectIdNumber)
 	secretPath := fmt.Sprintf("%s/secrets/%s", parent, secretName)
 
@@ -53,25 +52,31 @@ func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, proje
 		}
 	}
 
-	newVersion, err := client.AddSecretVersion(ctx, &secretmanagerpb.AddSecretVersionRequest{
+	_, err = client.AddSecretVersion(ctx, &secretmanagerpb.AddSecretVersionRequest{
 		Parent:  secretPath,
 		Payload: &secretmanagerpb.SecretPayload{Data: payload},
 	})
 	if err != nil {
 		return fmt.Errorf("failed to add version to secret %s: %w", secretName, err)
 	}
-
-	if destroyPreviousVersions {
-		if err := destroyPreviousSecretVersions(ctx, client, secretPath, newVersion.Name); err != nil {
-			return fmt.Errorf("failed to destroy previous versions of %s: %w", secretName, err)
-		}
-	}
 	return nil
 }
 
-// destroyPreviousSecretVersions destroys all enabled secret versions except the current one
+// CreateOrUpdateSecretDestroyingPreviousVersions creates or updates a secret and then destroys
+// all previously enabled versions to prevent version accumulation and cost growth.
+// Use this for periodic updates (e.g. ci-secret-generator). For initial creation where there
+// are no previous versions, use CreateOrUpdateSecret instead.
+func CreateOrUpdateSecretDestroyingPreviousVersions(ctx context.Context, client SecretManagerClient, projectIdNumber, secretName string, payload []byte, labels, annotations map[string]string) error {
+	if err := CreateOrUpdateSecret(ctx, client, projectIdNumber, secretName, payload, labels, annotations); err != nil {
+		return err
+	}
+	secretPath := fmt.Sprintf("%s/secrets/%s", GetProjectResourceIdNumber(projectIdNumber), secretName)
+	return destroyPreviousSecretVersions(ctx, client, secretPath)
+}
+
+// destroyPreviousSecretVersions destroys all enabled secret versions except the latest one,
 // to avoid accumulating billable versions in GSM.
-func destroyPreviousSecretVersions(ctx context.Context, client SecretManagerClient, secretPath, currentVersionName string) error {
+func destroyPreviousSecretVersions(ctx context.Context, client SecretManagerClient, secretPath string) error {
 	it := client.ListSecretVersions(ctx, &secretmanagerpb.ListSecretVersionsRequest{
 		Parent: secretPath,
 		Filter: "state:ENABLED",
@@ -79,6 +84,8 @@ func destroyPreviousSecretVersions(ctx context.Context, client SecretManagerClie
 	if it == nil {
 		return nil
 	}
+
+	var versions []*secretmanagerpb.SecretVersion
 	for {
 		v, err := it.Next()
 		if errors.Is(err, iterator.Done) {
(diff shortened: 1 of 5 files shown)


pkg/gsm-secrets/execution.go
```diff
@@ -52,24 +53,59 @@ func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, proje
...
+		if v.Name == currentVersionName {
+			continue
+		}
+		if _, err := client.DestroySecretVersion(ctx, &secretmanagerpb.DestroySecretVersionRequest{
+			Name: v.Name,
+		}); err != nil {
+			logrus.WithError(err).Warnf("Failed to destroy version %s", v.Name)
+		}
+	}
+}
```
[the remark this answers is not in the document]
psalajova: but we have no other tool that would concurrently edit/update these secrets, apart from ci-secret-generator itself

Prucek (approved):

pkg/gsm-secrets/execution.go
```diff
@@ -28,7 +28,8 @@ import (
 
 // CreateOrUpdateSecret creates a new secret in Google Secret Manager or updates an existing one with a new version.
 // If labels or annotations are nil, they won't be set on the secret.
-func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, projectIdNumber, secretName string, payload []byte, labels, annotations map[string]string) error {
+// If destroyPreviousVersions is true, all previously enabled versions are destroyed after adding the new one.
+func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, projectIdNumber, secretName string, payload []byte, labels, annotations map[string]string, destroyPreviousVersions bool) error {
```
droslean: It seems that you always want to destroy the previous versions, so the bool doesn't make any difference here.

pkg/gsm-secrets/execution.go
```diff
@@ -52,24 +53,59 @@ func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, proje
...
+	if it == nil {
+		return
+	}
+	for {
+		v, err := it.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil {
+			logrus.WithError(err).Warnf("Failed to list versions for secret %s", secretPath)
```
droslean: This means that the secret will keep the older versions, and we will eat the error completely. In the end, we could still end up with older versions in the current secrets. Therefore, the tool needs to fail here.

pkg/gsm-secrets/execution.go
```diff
@@ -52,24 +53,59 @@ func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, proje
...
+			return
+		}
+		if v.Name == currentVersionName {
+			continue
+		}
+		if _, err := client.DestroySecretVersion(ctx, &secretmanagerpb.DestroySecretVersionRequest{
+			Name: v.Name,
+		}); err != nil {
+			logrus.WithError(err).Warnf("Failed to destroy version %s", v.Name)
```
droslean: Same as well. We need to fail here.

psalajova:
We are discussing if this is the right approach in https://redhat-internal.slack.com/archives/GB7NB0CUC/p1779274337330229 , waiting for the conclusion there before I move forward with this

pkg/gsm-secrets/execution.go
psalajova: There are 3 call sites:
 - gsm.go:68 (SetFieldOnItem): passes `true` -- ci-secret-generator updating existing secrets
 - gsm.go:81 (UpdateIndexSecret): passes `true` -- ci-secret-generator updating index
 - execution.go:299 (CreateSecrets): passes `false` -- initial secret creation by gsm-secret-sync

The false at execution.go:299 is intentional: when gsm-secret-sync creates secrets for the first time, there are no previous versions to destroy. Additionally, the migration code I'm working on also calls CreateOrUpdateSecret with false (initial migration, no previous versions). I agree that maybe we could set it to always destroy previous versions, but it feels better to be able to control it.

droslean: So, you have two explicit behaviors. You shouldn't flip the switch with a boolean. Try to split the logic correctly and use the corresponding function in the correct places.

pkg/gsm-secrets/execution_mock.go
```diff
@@ -125,6 +125,26 @@ func (mr *MockSecretManagerClientMockRecorder) DeleteSecret(ctx, req any, opts .
 	return mr.mock.ctrl.RecordCallWithMethodType(mr.mock, "DeleteSecret", reflect.TypeOf((*MockSecretManagerClient)(nil).DeleteSecret), varargs...)
 }
 
+// DestroySecretVersion mocks base method.
+func (m *MockSecretManagerClient) DestroySecretVersion(ctx context.Context, req *secretmanagerpb.DestroySecretVersionRequest, opts ...gax.CallOption) (*secretmanagerpb.SecretVersion, error) {
```
droslean, line 129: I don't see the point of the whole MockSecretManagerClient.  In the end, your tests are testing the mocked mechanism. The whole point of the unittests is to actually test the logic that you introduced and use a fake client.

pkg/gsm-secrets/execution.go
```diff
@@ -59,17 +59,83 @@ func CreateOrUpdateSecret(ctx context.Context, client SecretManagerClient, proje
...
 		return fmt.Errorf("failed to add version to secret %s: %w", secretName, err)
 	}
+	return nil
+}
+
+// CreateOrUpdateSecretDestroyingPreviousVersions creates or updates a secret and then destroys
+// all previously enabled versions to prevent version accumulation and cost growth.
+// Use this for periodic updates (e.g. ci-secret-generator). For initial creation where there
+// are no previous versions, use CreateOrUpdateSecret instead.
+func CreateOrUpdateSecretDestroyingPreviousVersions(ctx context.Context, client SecretManagerClient, projectIdNumber, secretName string, payload []byte, labels, annotations map[string]string) error {
```
droslean, line 69: The name of the function is bigger than the function actually does :D. You might need to add some extra rules on your agent to generate Go idiomatic names instead of python ones.

droslean:
/lgtm
/approve

@psalajova Try to check my comment about the mocked stuff, I guess you can just remove them and refactor the tests in a follow up.

PR #5203, opened 2026-05-26
qci-appci: Add multi cluster token service

danilo-gemoli:
Allow qci-appci to authenticate SA coming from a supplemental cluster other than the one it is deployed to.
This should ease the app.ci to core-ci transition.

## Multi-Cluster Token Validation Support

The qci-appci reverse proxy now supports validating service account tokens against multiple Kubernetes clusters. Previously, it could only authenticate tokens from the cluster it was deployed on. This change enables it to accept tokens from a supplemental cluster, easing the transition of CI infrastructure from app.ci to core-ci.

## Changes

**Token Service Architecture**
- Refactored token validation into a `ClusterTokenService` interface with `Validate(token string) (bool, error)` and `Host() string` methods
- Introduced a new `multiClusterTokenService` that attempts token validation across multiple clusters in sequence
- The per-cluster token service (`simpleClusterTokenService`) now tracks which cluster it's validating against via the `Host()` method

**Configuration**
- Added `--supplemental-kubeconfig` flag to optionally specify a secondary cluster's kubeconfig
- When provided, qci-appci creates token validators for both the in-cluster (primary) and supplemental clusters
- Token validation succeeds if any cluster validates the token; if all fail, an aggregate error containing validation failures from all clusters is returned

**Behavior Impact**
- Service accounts from the supplemental cluster can now authenticate against qci-appci without requiring a separate token service instance
- CI operators can gradually migrate workloads between clusters during infrastructure transitions without breaking authentication
- Error diagnostics now include per-cluster validation failures, making it easier to troubleshoot multi-cluster authentication issues

The changes maintain backward compatibility—when no supplemental kubeconfig is provided, qci-appci operates as a single-cluster validator as before.

### DIFF
--- a/cmd/qci-appci/main.go
+++ b/cmd/qci-appci/main.go
@@ -22,7 +22,9 @@ import (
 
 	authenticationv1 "k8s.io/api/authentication/v1"
 	authorizationv1 "k8s.io/api/authorization/v1"
+	aggerrs "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/clientcmd"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/prow/pkg/config/secret"
 	"sigs.k8s.io/prow/pkg/interrupts"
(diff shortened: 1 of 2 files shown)


deepsm007 (approved):
Temp workaround and should be reverted once we migrate

PR #5207, opened 2026-05-27
prowgen: add disable_sparse_checkout option

Prucek:
## Summary
- Adds `disable_sparse_checkout` field to `ProwgenOverrides` in ci-operator config
- When set, prowgen generates jobs that perform a full clone instead of sparse checkout
- This is needed for repositories that use git submodules, as sparse checkout with blobless fetches (`--filter=blob:none`) fails when submodule gitlink entries are modified but the submodule path is excluded from the sparse working tree

Repos that need this can add to their ci-operator config:
```yaml
prowgen:
  disable_sparse_checkout: true
```

This is intended as a workaround until kubernetes-sigs/prow#723 lands with a proper fallback in clonerefs.

## Adds disable_sparse_checkout option for repository cloning

This PR adds a new ci-operator prowgen configuration option to opt out of the sparse-checkout optimization used during Prow job repository cloning.

### Context
Prowgen previously generated jobs that use sparse checkout with blobless fetches (--filter=blob:none) to fetch only files required for CI (e.g., Dockerfiles, ci-operator configs). That optimization can fail for repositories that use git submodules when submodule gitlink entries are modified but the submodule path is excluded from the sparse working tree. This change is a workaround until clonerefs gains a proper fallback (kubernetes-sigs/prow#723).

### Changes (user-visible)
- New config field: prowgen: disable_sparse_checkout: true — added as `disable_sparse_checkout` on ProwgenOverrides (pkg/api/types.go). When set, prowgen will generate jobs that perform a full git clone instead of using sparse checkout / blobless fetches.
- Job generation: prowgen's job base builder now respects this flag and leaves SparseCheckoutFiles unset when disable_sparse_checkout is true, preserving existing SkipCloning and OAuth token behavior for private jobs (pkg/prowgen/jobbase.go).
- Tests and fixtures: Added unit tests and YAML fixtures exercising public and private image-build jobs with sparse checkout disabled (pkg/prowgen/jobbase_test.go and testdata fixtures).

### Impact for CI users/operators
Repository owners can opt out of sparse-checkout on a per-repo basis by setting prowgen.disable_sparse_checkout: true in ci-operator configuration. This forces full repository checkouts for generated Prow jobs, avoiding failures for repos that rely on submodules or other cases where sparse checkout breaks, while other repos retain the performance benefits of sparse checkout.

### DIFF
--- a/pkg/prowgen/jobbase.go
+++ b/pkg/prowgen/jobbase.go
@@ -97,7 +97,10 @@ func NewProwJobBaseBuilder(configSpec *cioperatorapi.ReleaseBuildConfiguration, 
 	if shouldSkipCloning {
 		b.base.UtilityConfig.DecorationConfig.SkipCloning = ptr.To(true)
 	} else {
-		b.base.UtilityConfig.DecorationConfig.SparseCheckoutFiles = sparseFiles
+		disableSparseCheckout := configSpec.Prowgen != nil && configSpec.Prowgen.DisableSparseCheckout
+		if !disableSparseCheckout {
+			b.base.UtilityConfig.DecorationConfig.SparseCheckoutFiles = sparseFiles
+		}
 		if private {
 			b.base.UtilityConfig.DecorationConfig.OauthTokenSecret = &prowv1.OauthTokenSecret{Key: cioperatorapi.OauthTokenSecretKey, Name: cioperatorapi.OauthTokenSecretName}
 		}
--- a/pkg/api/types.go
+++ b/pkg/api/types.go
@@ -53,6 +53,9 @@ type ProwgenOverrides struct {
 	// Expose declares that jobs should not be hidden from view in deck if they
 	// are private. This field has no effect if private is not set.
 	Expose bool `json:"expose,omitempty"`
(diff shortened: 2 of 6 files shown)


pkg/prowgen/jobbase.go
```diff
@@ -91,11 +91,16 @@ func NewProwJobBaseBuilder(configSpec *cioperatorapi.ReleaseBuildConfiguration,
 
 	private := configSpec.Prowgen != nil && configSpec.Prowgen.Private
 	expose := configSpec.Prowgen != nil && configSpec.Prowgen.Expose
+	disableSparseCheckout := configSpec.Prowgen != nil && configSpec.Prowgen.DisableSparseCheckout
 
 	sparseFiles := sparseCheckoutFiles(configSpec)
 	shouldSkipCloning := len(sparseFiles) == 0
 	if shouldSkipCloning {
 		b.base.UtilityConfig.DecorationConfig.SkipCloning = ptr.To(true)
+	} else if disableSparseCheckout {
```
droslean: This needs to go in the else statement in 104-R109

droslean (approved):

PR #5215, opened 2026-05-29
fix: dockerfile registry.ci input detection with baseImages

deepsm007:
Skip Dockerfile registry.ci inputs already covered by `from` + `baseImages`; stop dropping the last unrelated `FROM` when `from` names a different base.

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_continuous-release-jobs/1797/pull-ci-openshift-continuous-release-jobs-master-images/2060040001508347904

## Summary

This PR fixes Dockerfile registry reference detection in OpenShift CI's pipeline configuration processing, addressing issues with duplicate input detection when base images are explicitly specified.

### Problem Fixed

The CI infrastructure was incorrectly detecting and processing Docker registry references from Dockerfile `FROM` instructions that were already declared through the `baseImages` configuration. This caused:
- Duplicate input detection when a base image was both explicitly configured and referenced in the Dockerfile
- Loss of the last `FROM` instruction in multi-stage builds when the primary `from` configuration pointed to a different base image
- Overly aggressive removal of `FROM` directives that were unrelated to the primary build stage

### Changes

**Core Detection Logic (`pkg/dockerfile/inputs.go`)**:
- Extended `DetectInputsFromDockerfile()` with a new `baseImages` parameter to enable skipping of Dockerfile registry references that are already covered by explicitly configured base images
- Added `matchesFromBaseImage()` helper function that compares Dockerfile references against the base images map by namespace, repository, and tag to determine if a reference should be skipped

**Simplified Registry Extraction (`pkg/dockerfile/extract.go`)**:
- Removed the `from` parameter from `ExtractRegistryReferences()` 
- The function now unconditionally extracts all registry references without special handling for dropping references based on the `from` configuration

**Updated Callers**:
- `pkg/defaults/defaults.go`: Now passes the `baseImages` map when detecting inputs during pipeline configuration processing
- `cmd/registry-replacer/main.go`: Simplified to use the new signature of `ExtractRegistryReferences()`

### Impact

CI operators and users defining multi-stage container builds will see more accurate image input detection. The fix ensures that:
- Base images explicitly declared in configuration are not redundantly detected from Dockerfile instructions
- All `FROM` instructions in multi-stage builds are properly preserved, regardless of which stages are used as the primary build base
- Registry reference detection correctly handles scenarios where the primary build base differs from secondary image references

### DIFF
--- a/cmd/registry-replacer/main.go
+++ b/cmd/registry-replacer/main.go
@@ -383,7 +383,7 @@ func replacer(
 }
 
 func ensureReplacement(image *api.ProjectDirectoryImageBuildStepConfiguration, dockerfile []byte) ([]cidockerfile.OrgRepoTag, error) {
-	toReplace := cidockerfile.ExtractRegistryReferences(dockerfile, "")
+	toReplace := cidockerfile.ExtractRegistryReferences(dockerfile)
 	var result []cidockerfile.OrgRepoTag
 	for _, toReplace := range toReplace {
 		orgRepoTag, err := cidockerfile.OrgRepoTagFromPullString(toReplace)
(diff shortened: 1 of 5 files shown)


deepsm007:
/override ci/prow/images

not related and this PR is critical for CRT

PR #5220, opened 2026-06-01
DPTP-3787: fix tolerate missing ocp source IS in release snapshot

deepsm007:
https://search.dptools.openshift.org/?search=Reporting+job+state.*with+reason.*executing_graph:step_failed:creating_release_images&maxAge=6h&context=1&type=build-log&name=&excludeName=&maxMatches=5&maxBytes=20971520&groupBy=job

Fixes issue introduced by #5217 on Saturday

/cc @openshift/test-platform 

## Fix for Release Snapshot Source ImageStream Recovery

This PR restores tolerant behavior in the release snapshot step so CI jobs no longer hard-fail when the official OCP source ImageStream (e.g., ocp/4.22) is missing from build clusters.

What changed (practical effect)
- Component affected: release snapshot logic used by the release-inputs step in ci-operator (pkg/steps/release).
- When the source ImageStream cannot be GETted and the error is a Kubernetes NotFound, snapshotStream now treats the source as absent (source = nil) and continues instead of failing the step. Other error types still fail.
- snapshotImportSource was reorganized so consolidated Quay promotion paths are evaluated independently of the presence of the source ImageStream and the spec tag lookup is only performed when a source exists. As a result, when the source ImageStream is absent, tag resolution falls back to a Quay-derived DockerImage reference (api.QuayImageReference).
- A unit test was added to verify behavior for consolidated streams when the source ImageStream is missing, asserting fallback to the Quay proxy reference.

Impact for CI users/operators
- Jobs that rely on official OCP ImageStreams that only exist on app.ci (and are not mirrored to build clusters) will now proceed: the [release-inputs:initial] step will gracefully fall back to Quay proxy references instead of hard-failing. This restores prior behavior for tag resolution and allows periodic and PR jobs that depend on unpinned float tags and Quay promotion fallbacks to run.

### DIFF
--- a/pkg/steps/release/snapshot.go
+++ b/pkg/steps/release/snapshot.go
@@ -8,6 +8,7 @@ import (
 	"github.com/sirupsen/logrus"
 
 	coreapi "k8s.io/api/core/v1"
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
 	ctrlruntimeclient "sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


deepsm007:
/test lint images

No nodes available infra failure

deepsm007:
/override ci/prow/e2e

Failure unrelated

PR #5227, opened 2026-06-02
DPTP-2938: fix pod-scaler quantile int64 overflow at digest

deepsm007:
/cc @openshift/test-platform 

## Pod-scaler: skip invalid negative cache recommendations

This PR hardens the pod-scaler component (the controller that recommends container CPU/memory requests and limits for CI workloads based on Prometheus metrics) by validating quantile-derived values and skipping any invalid recommendations so they are not cached or propagated to containers.

### What changed (practical effect)
- pod-scaler now validates quantile values derived from metrics and will not write invalid conversions into the recommendation cache. Invalid inputs include +/-Inf, NaN, negative values, and extreme/overflow-like values.
- When an invalid quantile is encountered for a given metadata/resource pair, the code logs a debug message and skips updating that recommendation instead of storing an unusable quantity.
- Tests were consolidated and expanded into a table-driven test that verifies correct conversions for normal CPU and memory inputs and confirms invalid inputs are rejected.

### Components affected
- cmd/pod-scaler (recommender/cache logic): runtime behavior changed to avoid caching invalid resource.Quantity values.
- pod-scaler tests: updated to cover the new validation logic.

### Impact for CI users/operators
Pod-scaler will be more robust against anomalous Prometheus data and will no longer produce or cache unusable resource requests/limits derived from invalid metric quantiles. This reduces the risk that CI pods receive invalid resource recommendations due to metric corruption or edge-case values.

### DIFF
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -47,7 +47,7 @@ func admit(port, healthPort int, certDir string, client buildclientv1.BuildV1Int
 		}).Info("authoritative decrease dry-run enabled")
 	}
 	health := pjutil.NewHealthOnPort(healthPort)
-	resources := newResourceServer(loaders, health)
+	resources := newResourceServer(loaders, health, cpuCap, memoryCap)
 	decoder := admission.NewDecoder(scheme.Scheme)
 
 	// Initialize node allocatable CPU cache
@@ -378,7 +378,15 @@ func reconcileLimits(resources *corev1.ResourceRequirements) {
 	}
 }
 
+func capDigestRequests(resources *corev1.ResourceRequirements, cpuCap, memoryCap resource.Quantity, logger *logrus.Entry) {
+	preventUnschedulableWithCaps(resources, cpuCap, memoryCap, logger.WithField("stage", "digest"))
+}
+
 func preventUnschedulable(resources *corev1.ResourceRequirements, cpuCap int64, memoryCap string, logger *logrus.Entry) {
+	preventUnschedulableWithCaps(resources, *resource.NewQuantity(cpuCap, resource.DecimalSI), resource.MustParse(memoryCap), logger)
+}
+
+func preventUnschedulableWithCaps(resources *corev1.ResourceRequirements, cpuCap, memoryCap resource.Quantity, logger *logrus.Entry) {
 	if resources.Requests == nil {
 		logger.Debug("no requests, skipping")
 		return
@@ -386,18 +394,16 @@ func preventUnschedulable(resources *corev1.ResourceRequirements, cpuCap int64, 
 
 	if _, ok := resources.Requests[corev1.ResourceCPU]; ok {
 		// TODO(DPTP-2525): Make cluster-specific?
-		cpuRequestCap := *resource.NewQuantity(cpuCap, resource.DecimalSI)
-		if resources.Requests.Cpu().Cmp(cpuRequestCap) == 1 {
+		if resources.Requests.Cpu().Cmp(cpuCap) == 1 {
 			logger.Debugf("setting original CPU request of: %s to cap", resources.Requests.Cpu())
-			resources.Requests[corev1.ResourceCPU] = cpuRequestCap
+			resources.Requests[corev1.ResourceCPU] = cpuCap
 		}
 	}
 
 	if _, ok := resources.Requests[corev1.ResourceMemory]; ok {
-		memoryRequestCap := resource.MustParse(memoryCap)
-		if resources.Requests.Memory().Cmp(memoryRequestCap) == 1 {
+		if resources.Requests.Memory().Cmp(memoryCap) == 1 {
 			logger.Debugf("setting original memory request of: %s to cap", resources.Requests.Memory())
-			resources.Requests[corev1.ResourceMemory] = memoryRequestCap
+			resources.Requests[corev1.ResourceMemory] = memoryCap
 		}
 	}
 }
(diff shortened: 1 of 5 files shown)


Prucek (approved):

cmd/pod-scaler/resources.go
```diff
@@ -81,7 +119,30 @@ func (s *resourceServer) digestData(data *podscaler.CachedQuery, quantile float6
 			overall.Merge(data.Data[fingerprintTime.Fingerprint].Histogram())
 		}
 		metaLogger.Trace("merged all fingerprints")
-		valueAtQuantile := overall.ValueAtQuantile(quantile)
+		recommendation, recOK := recommendationFromHistogram(overall, quantile)
```
droslean: To avoid maintaining the unnecessary bool here, it is idiomatic in Go to work with pointers and errors if needed. Instead, you could just return a *float64 and and return nil instead of just `0`, and rely on another return var to check if there is any recommendation.

cmd/pod-scaler/resources.go
```diff
@@ -81,7 +119,30 @@ func (s *resourceServer) digestData(data *podscaler.CachedQuery, quantile float6
...
+		if !recOK {
+			metaLogger.Debug("skipping recommendation with no usable histogram data")
+			continue
+		}
+		var (
```
droslean: The whole block here is needed as long with the switch below, because you are splitting the same logic into different functions. Instead you need to return a correct struct from a `quantityFromQuantile()` that includes the desired result based on the input you need to give which is technically the corev1.ResourceCPU and corev1.ResourceMemory that you are checking below. So in the end your function should just return the `corev1.ResourceList{}`

cmd/pod-scaler/resources.go
```diff
@@ -34,43 +37,102 @@ type resourceServer struct {
...
+	q, ok := quantityFromUsage(*usage, request)
+	if !ok {
+		return nil
+	}
+	requests := corev1.ResourceList{request: q}
+	capRequests(requests, cpuCap, memoryCap)
+	return requests
+}
+
+func quantityFromUsage(usage float64, request corev1.ResourceName) (resource.Quantity, bool) {
```
droslean: I don't see why you need the bool here.

cmd/pod-scaler/resources.go
```diff
@@ -34,43 +37,102 @@ type resourceServer struct {
...
+		return resource.Quantity{}, false
 	}
 }
 
+func capRequests(requests corev1.ResourceList, cpuCap, memoryCap resource.Quantity) {
+	reqs := &corev1.ResourceRequirements{Requests: requests}
+	capDigestRequests(reqs, cpuCap, memoryCap, nil)
+}
+
+func (s *resourceServer) digestCPU(data *podscaler.CachedQuery) {
```
droslean: This method is completely unnecessary.

cmd/pod-scaler/resources.go
```diff
@@ -34,43 +37,102 @@ type resourceServer struct {
...
+		if usage > float64(math.MaxInt64) {
+			return resource.Quantity{}, false
+		}
+		return *resource.NewQuantity(int64(usage), resource.BinarySI), true
+	default:
+		return resource.Quantity{}, false
 	}
 }
 
+func capRequests(requests corev1.ResourceList, cpuCap, memoryCap resource.Quantity) {
```
droslean: Same. No need for this function here.

cmd/pod-scaler/resources.go
```diff
@@ -34,43 +37,102 @@ type resourceServer struct {
...
+	if !ok {
+		return nil
+	}
+	requests := corev1.ResourceList{request: q}
+	capRequests(requests, cpuCap, memoryCap)
+	return requests
+}
+
+func quantityFromUsage(usage float64, request corev1.ResourceName) (resource.Quantity, bool) {
+	switch request {
```
droslean: The fact that you need to switch the request here means that the logic is already broken in a previous layer.

deepsm007:
/override ci/prow/e2e
unrelated to the changes

cmd/pod-scaler/admission.go
```diff
@@ -378,26 +378,41 @@ func reconcileLimits(resources *corev1.ResourceRequirements) {
...
 func preventUnschedulable(resources *corev1.ResourceRequirements, cpuCap int64, memoryCap string, logger *logrus.Entry) {
+	preventUnschedulableWithCaps(resources, *resource.NewQuantity(cpuCap, resource.DecimalSI), resource.MustParse(memoryCap), logger)
+}
+
+func preventUnschedulableWithCaps(resources *corev1.ResourceRequirements, cpuCap, memoryCap resource.Quantity, logger *logrus.Entry) {
 	if resources.Requests == nil {
-		logger.Debug("no requests, skipping")
+		if logger != nil {
+			logger.Debug("no requests, skipping")
+		}
```
smg247: nil logger? why would this ever be?

cmd/pod-scaler/resources.go
```diff
@@ -34,44 +45,105 @@ type resourceServer struct {
...
+	usage := recommendationValue(hist, quantile)
+	if usage == nil {
+		return nil
+	}
+	q := memoryQuantityFromBytes(*usage)
+	if q == nil {
+		return nil
+	}
+	reqs := &corev1.ResourceRequirements{Requests: corev1.ResourceList{corev1.ResourceMemory: *q}}
+	capDigestRequests(reqs, resource.Quantity{}, memoryCap, nil)
```
smg247: passing in nil for the logger explains the above. we shouldn't do this though. always supply a logger

deepsm007:
/override ci/prow/e2e
Unrelated to PR changes

PR #5244, opened 2026-06-10
fix(ci-operator): keep registry.ci builder FROM when final stage is external

deepsm007:
/cc @openshift/test-platform 

## Dockerfile Input Detection with External Final Stage

**Component**: ci-operator (Dockerfile parsing for build input detection)

**Problem**: ci-operator's Dockerfile scanner previously missed some registry.ci.openshift.org references in real-world Dockerfiles. It also incorrectly excluded all registry.ci references when the final stage used an external base image (via `from`), which broke multi-stage builds that relied on registry.ci builder stages.

**What changed**:
- ExtractRegistryReferences (pkg/dockerfile/extract.go) now:
  - Normalizes Dockerfile line continuations (replaces backslash-newline with spaces) so matches can span continued lines.
  - Scans RUN instructions in addition to FROM and COPY, so image references in commands like `podman pull` are detected.
  - Deduplicates matches with a seen set.
  - Tracks the registry captured from the final FROM line only, and excludes only that registry when a final external `from` is specified — earlier builder-stage registry.ci references are preserved.

**Behavioral impact**: Multi-stage Dockerfiles that use registry.ci builder images (e.g., OCP builders) in intermediate stages and an external final base image (e.g., Red Hat UBI) will now have the registry.ci builder references correctly detected and wired as build inputs. Detection is also more robust across line continuations and image pulls inside RUN commands.

**Tests**: Added table-driven tests (pkg/dockerfile/inputs_test.go) covering:
- detection of registry.ci references in `podman pull` (single-line and continued-line forms),
- a case where a builder-stage FROM uses registry.ci while the final stage is an external UBI image, asserting the builder registry is retained.

**Practical benefit**: Reduces manual configuration and prevents accidental omission of required builder images from CI build inputs, improving reliability for multi-stage builds that mix internal builder images and external base images.

### DIFF
--- a/pkg/dockerfile/extract.go
+++ b/pkg/dockerfile/extract.go
@@ -14,6 +14,9 @@ import (
 // RegistryRegex matches registry references to registry.ci.openshift.org or quay-proxy.ci.openshift.org
 var RegistryRegex = regexp.MustCompile(`(registry\.(?:svc\.)?ci\.openshift\.org|quay-proxy\.ci\.openshift\.org)/[^\s\\]+`)
 
+// dockerfileLineContinuation matches Dockerfile backslash line continuations.
+var dockerfileLineContinuation = regexp.MustCompile(`\\\r?\n[ \t]*`)
+
 // OrgRepoTag represents a parsed image reference
 type OrgRepoTag struct {
 	Org, Repo, Tag string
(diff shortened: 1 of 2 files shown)


deepsm007:
/override ci/prow/integration

unrelated failures to the PR

PR #5252, opened 2026-06-16
Register libvirt-s390x-vpn-ssp cluster profile

Davo911:
Add a new cluster profile `libvirt-s390x-vpn-ssp` for running SSP e2e functional tests on s390x architecture in the orange zone VPN environment

This provides a dedicated profile for the SSP operator CI lane, using its own host and Boskos leases, will be added in a follow-up PR
Signed-off-by: Thomas-David Griedel griedel911@gmail.com

## Summary

This pull request extends the OpenShift CI tools cluster profile API to add a new s390x “orange zone” VPN libvirt cluster profile for SSP e2e functional testing. It introduces the `libvirt-s390x-vpn-virt` profile so CI jobs and infrastructure resolution can consistently derive the expected cluster type and Boskos quota-slice lease name for this SSP-focused lane.

## Changes

**pkg/api/clusterprofile.go**
- Added a new exported `ClusterProfile` constant:
  - `ClusterProfileLibvirtS390xVPNVirt ClusterProfile = "libvirt-s390x-vpn-virt"`
- Registered the new profile in `ClusterProfiles()` so it’s accepted as a valid input.
- Updated the conversion helpers to support this profile end-to-end:
  - `ClusterType()` now maps `ClusterProfileLibvirtS390xVPNVirt` → `"libvirt-s390x-vpn-virt"`.
  - `LeaseType()` now maps `ClusterProfileLibvirtS390xVPNVirt` → `"libvirt-s390x-vpn-virt-quota-slice"`.
  - `LeaseTypeFromClusterType()` now accepts `"libvirt-s390x-vpn-virt"` and converts it to the corresponding `-quota-slice` lease type.

## Impact

CI configuration can reference `libvirt-s390x-vpn-virt` to ensure SSP e2e jobs resolve the correct cluster type and Boskos lease/quota-slice naming for the orange zone VPN environment (with host and Boskos lease resource details planned in a follow-up).

### DIFF
--- a/pkg/api/clusterprofile.go
+++ b/pkg/api/clusterprofile.go
@@ -123,6 +123,7 @@ const (
 	ClusterProfileLibvirtS390xAmd64          ClusterProfile = "libvirt-s390x-amd64"
 	ClusterProfileLibvirtS390xVPN            ClusterProfile = "libvirt-s390x-vpn"
 	ClusterProfileLibvirtS390xVPNOZ          ClusterProfile = "libvirt-s390x-vpn-oz"
+	ClusterProfileLibvirtS390xVPNVirt        ClusterProfile = "libvirt-s390x-vpn-virt"
 	ClusterProfileMetalPerfscaleBMCPT        ClusterProfile = "metal-perfscale-cpt"
 	ClusterProfileMetalPerfscaleJetlag       ClusterProfile = "metal-perfscale-jetlag"
 	ClusterProfileMetalPerfscaleOSP          ClusterProfile = "metal-perfscale-osp"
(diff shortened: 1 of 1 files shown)


pkg/api/clusterprofile.go
```diff
@@ -123,6 +123,7 @@ const (
 	ClusterProfileLibvirtS390xAmd64          ClusterProfile = "libvirt-s390x-amd64"
 	ClusterProfileLibvirtS390xVPN            ClusterProfile = "libvirt-s390x-vpn"
 	ClusterProfileLibvirtS390xVPNOZ          ClusterProfile = "libvirt-s390x-vpn-oz"
+	ClusterProfileLibvirtS390xVPNSSP         ClusterProfile = "libvirt-s390x-vpn-ssp"
```
nestoracunablanco: @Davo911 thanks for the PR. The idea is to use the profile not only in the ssp operator, so in another components. What about another naming convention, like ClusterProfileLibvirtS390xVPNOcpVirt?

nestoracunablanco: @bear-redhat, to avoid any confusion: this is for the end-to-end tests. For unit tests, we'll need the approach we discussed regarding the compute node.

nestoracunablanco (commented):
Thanks for the PR, @Davo911. Some comments only regarding naming convention.

pkg/api/clusterprofile.go
Davo911: in that case I agree, the naming should be more inclusive
I'm fine with OcpVirt as suffix

deepsm007 (approved):

PR #5253, opened 2026-06-16
ci-operator-configresolver: Add upstream config resolver

danilo-gemoli:
`e2e` tests use a real `ci-operator-configresolver` that is expected to find IS `ocp/X.Y` within the cluster the test is running on, but these streams exist on `app.ci` only.

As a consequence of that, we experience failures like [this](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_ci-tools/5247/pull-ci-openshift-ci-tools-main-e2e/2066457872044134400).
This PR overcomes the problem by adding a new parameter on `ci-operator-configresolver`:
```
--upstream-resolver-address=...
```
that `e2e`s set to the production endpoint `https://config.ci.openshift.org/`.

The upstream resolver is used to retrieve information about the integrated streams.

## Upstream Resolver Support for `ci-operator-configresolver`

This PR adds an **optional upstream resolver** path for resolving OpenShift “integrated streams” so **e2e tests in isolated clusters** don’t fail when the test cluster lacks `ocp/X.Y` ImageStreams (those exist only on `app.ci`). When configured, `ci-operator-configresolver` delegates integrated-stream lookups to the **production config service**.

### What changed

**`ci-operator-configresolver` (`cmd/ci-operator-configresolver/main.go`)**
- Added a new CLI flag: `--upstream-resolver-address`.
- Implemented a new TTL-backed cache for integrated stream resolution:
  - Uses **local** resolution when no upstream is configured.
  - Uses an **upstream resolver client** when `--upstream-resolver-address` is set.
- Updated the server wiring so the `/integratedStream` endpoint serves from the new integration cache (replacing the prior in-memory/mutex cache approach).

**Unit tests (`cmd/ci-operator-configresolver/main_test.go`)**
- Refactored integrated-stream tests to exercise the TTL cache for both:
  - **Local** integrated stream resolution
  - **Upstream** integrated stream resolution, including an **upstream miss** case with the expected HTTP 400 error.

**E2E framework + tests**
- Extended `test/e2e/framework/framework.go`:
  - Added `ConfigResolverOptions.UpstreamResolverAddress`
  - Passed it through to `ci-operator-configresolver` as `--upstream-resolver-address`.
- Updated e2e callers to set the upstream resolver address using the CI API config service endpoint (`api.URLForService("config")`):
  - `test/e2e/multi-stage/e2e_test.go`
  - `test/e2e/observer/e2e_test.go`
- Also removed per-iteration loop-variable shadowing in the e2e tests to ensure the intended `testCase` is used.
- `test/e2e/multi-stage/integration-releases.yaml`: adjusted the **4.18 “latest”** checks to read `/release-manifests/release-metadata` instead of grepping `cluster-version-operator version` / `oc version`.

### Practical impact
- **Default behavior is unchanged**: if the flag is not set, integrated streams are resolved locally as before.
- **Opt-in behavior**: e2e environments can point `ci-operator-configresolver` to `https://config.ci.openshift.org/` (via the framework wiring) to resolve integrated streams even when `ocp/X.Y` ImageStreams are missing from the test cluster.

### Test/validation notes
- The PR author requested e2e runs and reruns via PR commands (`/test e2e`, `/hold`, `/retest-required`) while validating the change.

### DIFF
--- a/cmd/ci-operator-configresolver/main.go
+++ b/cmd/ci-operator-configresolver/main.go
@@ -40,18 +40,19 @@ import (
 )
 
 type options struct {
-	configPath             string
-	registryPath           string
-	logLevel               string
-	address                string
-	releaseRepoGitSyncPath string
-	port                   int
-	uiAddress              string
-	uiPort                 int
-	gracePeriod            time.Duration
-	validateOnly           bool
-	flatRegistry           bool
-	instrumentationOptions flagutil.InstrumentationOptions
+	configPath              string
+	registryPath            string
+	logLevel                string
+	address                 string
+	releaseRepoGitSyncPath  string
+	port                    int
+	uiAddress               string
+	uiPort                  int
+	gracePeriod             time.Duration
+	validateOnly            bool
+	flatRegistry            bool
+	instrumentationOptions  flagutil.InstrumentationOptions
+	upstreamResolverAddress string
 }
 
 var (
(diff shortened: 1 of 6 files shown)


droslean (approved):

danilo-gemoli:
I'm fixing this with https://github.com/openshift/release/pull/81013

droslean (approved):

PR #5258, opened 2026-06-17
fix: read dockerfiles from extra-ref clone paths

deepsm007:
PRPQR builds hello-openshift from extra ref openshift.origin, but Dockerfile auto-detection reads from wrong working directory and silently skips input wiring and the build pulls hardcoded [registry.ci](http://registry.ci/) URLs that are now reference only.

https://redhat-internal.slack.com/archives/CBN38N3MW/p1781697111353389?thread_ts=1781621737.997829&cid=CBN38N3MW

https://prow.ci.openshift.org/view/gs/test-platform-results/logs/openshift-origin-31303-openshift-cluster-cloud-controller-manager-operator-471-nightly-4.22-e2e-vsphere-ovn-hybrid-env-techpreview/2067199635323097088

## Summary

This PR fixes Dockerfile auto-detection for ci-operator builds that fetch sources from additional reference clone paths (PRPQR builds). Previously, Dockerfile detection could look in the wrong working directory for ref-specific image builds, causing the ci-operator to silently skip wiring detected inputs and fall back to hardcoded `registry.ci.openshift.org` references (which can now be reference-only and fail). The change is motivated by a Slack discussion and corroborated by a failing Prow build log.

### Reference-aware workdir/Dockerfile resolution
- `pkg/defaults/defaults.go`
  - Makes Dockerfile input detection ref-aware by threading `JobSpec` through:
    - `runtimeStepConfigsForBuild(...)` now passes `jobSpec` into `detectDockerfileInputs(...)`.
    - `detectDockerfileInputs(...)` now calls `readDockerfileForImage(image, readFile, jobSpec)` while iterating project images.
    - `readDockerfileForImage(...)` now:
      - Computes which `prowapi.Refs` apply to the image via a new `matchingRefsForImage(image.Ref, jobSpec)` helper (matches `org.repo` against `jobSpec.Refs` and `jobSpec.ExtraRefs`).
      - Uses `decorate.DetermineWorkDir(codeMountPath, matchingRefs)` to resolve the correct ref clone working directory.
      - Builds the final Dockerfile path by joining the resolved workdir with `image.ContextDir` (if set) and the Dockerfile path.

### Related behavior for build-root workdir selection
- `runtimeStepConfigsForBuild(...)` also uses `matchingRefsForImage(...)` when determining the workdir for ref-specific build roots, falling back to `determinePrimaryRef(...)` when no matching refs are found.

## Tests

- `pkg/defaults/defaults_test.go`
  - Updates `TestReadDockerfileForImage` to pass `jobSpec` into `readDockerfileForImage`.
  - Adds coverage for the “extra ref dockerfile path” case, validating Dockerfile reads from the computed filesystem path based on `jobSpec.ExtraRefs` (including `PathAlias`).

## Impact for CI Operators

CI jobs that build with multiple repositories (primary + `ExtraRefs`) will now reliably detect and wire Dockerfile inputs from the correct cloned ref directory, avoiding the silent skip that previously led to hardcoded registry fallbacks.

## Validation note

The author requested running end-to-end testing using `/test e2e` to validate the fix in PRPQR contexts.

### DIFF
--- a/pkg/defaults/defaults.go
+++ b/pkg/defaults/defaults.go
@@ -872,14 +872,11 @@ func runtimeStepConfigsForBuild(
 			if root.FromRepository {
 				path := "."        // By default, the path will be the working directory
 				if len(refs) > 1 { // If we are getting the build root image for a specific ref we must determine the absolute path
-					var matchingRefs []prowapi.Refs
-					for _, r := range refs {
-						if ref == fmt.Sprintf("%s.%s", r.Org, r.Repo) {
-							matchingRefs = append(matchingRefs, r)
+					matchingRefs := matchingRefsForImage(ref, jobSpec)
+					if len(matchingRefs) == 0 {
+						if primaryRef := determinePrimaryRef(jobSpec, injectedTest); primaryRef != nil {
+							matchingRefs = append(matchingRefs, *primaryRef)
 						}
-					}
-					if len(matchingRefs) == 0 { // If we didn't find anything, use the primary refs
-						matchingRefs = append(matchingRefs, *determinePrimaryRef(jobSpec, injectedTest))
 					}
 					path = decorate.DetermineWorkDir(codeMountPath, matchingRefs)
 				}
@@ -920,7 +917,7 @@ func runtimeStepConfigsForBuild(
 	buildSteps = append(buildSteps, getSourceStepsForJobSpec(jobSpec, injectedTest)...)
 
 	// Detect Dockerfile inputs for project images and add InputImageTagStepConfiguration entries
-	dockerfileInputSteps, images := detectDockerfileInputs(config.Images.Items, config.BaseImages, readFile)
+	dockerfileInputSteps, images := detectDockerfileInputs(config.Images.Items, config.BaseImages, readFile, jobSpec)
 	config.Images.Items = images
 	buildSteps = append(buildSteps, dockerfileInputSteps...)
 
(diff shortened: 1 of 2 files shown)


pkg/defaults/defaults.go
```diff
@@ -952,9 +952,25 @@ func detectDockerfileInputs(images []api.ProjectDirectoryImageBuildStepConfigura
 	return steps, images
 }
 
+func matchingRefsForImage(imageRef string, jobSpec *api.JobSpec) []prowapi.Refs {
+	if imageRef == "" || jobSpec == nil {
+		return nil
+	}
+	var refs []prowapi.Refs
+	if jobSpec.Refs != nil && imageRef == fmt.Sprintf("%s.%s", jobSpec.Refs.Org, jobSpec.Refs.Repo) {
```
smg247, line 955: are `jobSpec.Refs` ever nil?

pkg/defaults/defaults.go
```diff
@@ -952,9 +952,25 @@ func detectDockerfileInputs(images []api.ProjectDirectoryImageBuildStepConfigura
 	return steps, images
 }
 
+func matchingRefsForImage(imageRef string, jobSpec *api.JobSpec) []prowapi.Refs {
+	if imageRef == "" || jobSpec == nil {
```
smg247: `jobSpec` will never be nil, and I expect `imageRef` will never be empty, I think we can omit this early return

pkg/defaults/defaults.go
smg247, line 881: isn't this logic pretty similar to what we are doing here? can this be consolidated?

pkg/defaults/defaults.go
deepsm007: it is intentional; `jobSpec.Refs` is nil whenever ProwJob has `extra_refs` only — no top-level refs.

smg247 (commented):
/lgtm

PR #5268, opened 2026-06-24
ci-operator: Add `ip_pool_lease_type` field

danilo-gemoli:
This is part of some preparatory work to move cluster profiles definition out of the code base.

Up until now the `IPPoolLeaseType` exists in code only, this PR adds `ip_pool_lease_type` in the cluster profile configuration file to fill this gap.

This updates ci-operator’s cluster profile handling to support `ip_pool_lease_type` in profile configuration, closing the gap between config files and the in-code `IPPoolLeaseType` setting. In practice, CI cluster profiles can now carry this value through loading and merging just like the existing lease/secret/owner fields, so operators can define IP pool lease behavior in the cluster profile config instead of relying on code-only defaults.

The loader now fills in missing `ip_pool_lease_type` values from the profile name’s mapped default when needed, and the tests were expanded to cover profiles that include this field.

### DIFF
--- a/pkg/api/clusterprofile.go
+++ b/pkg/api/clusterprofile.go
@@ -1270,12 +1270,13 @@ func (cpl *ClusterProfilesList) Resolve() error {
 type ClusterProfilesMap map[ClusterProfile]ClusterProfileDetails
 
 type ClusterProfileDetails struct {
-	Name        ClusterProfile         `yaml:"name,omitempty" json:"name,omitempty"`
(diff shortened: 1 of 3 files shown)


Prucek (approved):

danilo-gemoli:
Unrelated to this PR, but this is a mess anyway
/override ci/prow/images
/override ci/prow/integration
/override ci/prow/e2e
/override ci/prow/validate-prow
/override ci/prow/validate-vendor

PR #5269, opened 2026-06-24
ci-operator: Rename the param key that holds the cluster profile details

danilo-gemoli:
`ci-operator` stores the `ClusterProfileDetails` object into the `api.Parameters` container under the  `CLUSTER_PROFILE` key.  This causes troubles when `CLUSTER_PROFILE` is already defined at ProwJob level as an environment variable, reporting this error:
```
{  get cluster profile from parameters: param types mismatch: type *api.ClusterProfileDetails expected but got string}
```
That's because the `api.ClusterProfileFromParams` function looks for a parameter with key `CLUSTER_PROFILE` of type `*api.ClusterProfileDetails` and, for legacy reasons, the `api.Parameters` looks into the env variables first. In this case, when a match is found, there is a type mismatch issue since the env variables are stored as `string` but we expect a `*api.ClusterProfileDetails`.

This updates ci-operator’s cluster profile plumbing to stop using the generic `CLUSTER_PROFILE` parameter key and switch to a dedicated `CLUSTER_PROFILE_DETAILS` key instead.

Practically, this prevents `api.ClusterProfileFromParams` from accidentally reading a ProwJob environment variable named `CLUSTER_PROFILE` and failing with a type mismatch. The affected behavior is in ci-operator’s parameter handling and lease/pod generation paths, which now consistently provide and consume the renamed cluster profile details parameter. Test expectations were updated to match the new key.


Prucek (approved):

danilo-gemoli:
I'm fixing this with https://github.com/openshift/release/pull/81013

PR #5272, opened 2026-06-25
Register aws-konflux-stg cluster profile

amisstea:
This new cluster profile is needed to support cluster provisioning on AWS using the OpenShift-CI integration in Red Hat's Konflux staging environments.

Adds support for a new `aws-konflux-stg` cluster profile in the CI API layer.

For CI operators, this means cluster provisioning can now target the AWS Konflux staging environment through OpenShift-CI. The new profile is recognized as an AWS cluster type, is returned with the available cluster profiles, and uses its own lease/quota slice type (`aws-konflux-stg-quota-slice`) for resource management.


amisstea:
No longer needed. The cluster profile definition was setup by https://github.com/openshift/release/pull/81097 and https://github.com/openshift/release/pull/81143 instead.

PR #5277, opened 2026-06-26
fix(tool-detector): skip build-tagged packages in changed file detection

deepsm007:
When a PR changes both real tool code (like cmd/pod-scaler) and e2e test files tagged` //go:build e2e`, the build detector used to crash on the test files and give up, so it built every image instead of just the one that changed.

Now it ignores those build-tagged test packages and still spots which `cmd/*` tools actually changed, so only the right images get built.

/cc @droslean 

Updated the tool detector so CI image selection no longer breaks when a changed file belongs to a `//go:build e2e` package.

In practical terms, when OpenShift CI changes both real tool code and e2e-only test files, the detector now skips the build-tagged test package instead of treating it as a fatal package-load failure. This lets `pkg/tool-detector` continue identifying which `cmd/*` tools actually changed and build only the relevant images, rather than falling back to rebuilding everything.

Added test coverage to verify that build-tagged packages are ignored during changed-package detection.

### DIFF
--- a/pkg/tool-detector/detector.go
+++ b/pkg/tool-detector/detector.go
@@ -235,6 +235,10 @@ func (d *Detector) loadChangedPackages(files []string) (sets.Set[string], error)
 	for _, pkg := range pkgs {
 		if len(pkg.Errors) > 0 {
 			for _, pkgErr := range pkg.Errors {
+				if strings.Contains(pkgErr.Error(), "build constraints exclude all Go files") {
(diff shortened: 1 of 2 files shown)


droslean:
/hold

I don't think this is a solution. You are just hacking around it. We need to ensure that when it fails, that it fallbacks to build all images, I think this is in place. Can you check to confirm?

PR #5284, opened 2026-06-30
Remove Cluster Profiles

danilo-gemoli:
Cluster profiles are entirely defined in this config file [ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml](https://github.com/openshift/release/blob/99f360dc627a2b5a79608cd7e8e2cd5a06699746/ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml), therefore we can now remove the old definitions embedded in the code.

Removed legacy, code-defined cluster profile types and logic across ci-operator tooling, switching all cluster profile handling to plain string identifiers sourced from `ci-operator/step-registry/cluster-profiles/cluster-profiles-config.yaml` (the new source of truth).

Practical impact:
- CI cluster profile resolution, validation, and multi-stage test/job generation now operate directly on profile names from config (no enum-like `api.ClusterProfile` conversions).
- `release profile` now lists/validates cluster profiles by loading the YAML config at runtime, ensuring the CLI output matches what CI uses.
- Load-time “merge/default” behavior was removed/simplified, so profile details are taken as defined in the config rather than being backfilled from built-in defaults.
- Rebalancing, repo-init prompts, registry resolution, and Prow jobbase creation now track profile weights/membership and set `cloud.cluster_profile` labels using string profile names.

Additional notes:
- Updated APIs and structs to replace `ClusterProfile`-typed fields with `string` (including JSON/unmarshal behavior for test configuration).
- Updated tests/fixtures and fuzzing inputs to use string-based cluster profile identifiers and the new config-driven expectations.


danilo-gemoli:
/override ci/prow/integration
not related

PR #5287, opened 2026-06-30
[DPTP-5001] feat(prpqr-reconciler): inject credentials and hide jobs for openshift-priv PRs

smith-xyz:
Adds cred injection for prpqr usage in openshift-priv repositories

This update teaches the PR payload qualification flow to recognize private repositories and handle them like CI needs to for OpenShift private repos.

In practice, `payload-testing-prow-plugin` now marks a PRPQR as private when any participating repository is private, and the `prpqr-reconciler` uses that signal to inject GitHub credentials and hide the generated ProwJobs. That applies to both regular and aggregated jobs, so private PR processing can run without exposing those jobs publicly.

The API/CRD was extended with a new `spec.private` flag to carry this state through the PRPQR object, and tests were added to cover private-repo behavior end to end.


cmd/payload-testing-prow-plugin/server.go
```diff
@@ -878,3 +878,18 @@ Please contact an administrator to resolve this issue.`,
...
+
+const openShiftPrivOrg = "openshift-priv"
+
+func isConfigResolverNotFound(err error) bool {
+	return err != nil && strings.Contains(err.Error(), "got unexpected http 404 status code from configresolver")
+}
+
+func resolveCIOpConfig(resolver ciOpConfigResolver, org, repo, branch string, logger *logrus.Entry) (*api.ReleaseBuildConfiguration, error) {
+	ciOpConfig, err := resolver.Config(&api.Metadata{Org: org, Repo: repo, Branch: branch})
+	if err != nil && org == openShiftPrivOrg && isConfigResolverNotFound(err) {
```
droslean: I don't see how the resolution and the fallback is related to what you need to introduce in this PR.

cmd/payload-testing-prow-plugin/server.go
```diff
@@ -878,3 +878,18 @@ Please contact an administrator to resolve this issue.`,
...
+		logger.WithField("fallback", "openshift").Info("falling back to openshift org config for openshift-priv repo")
```
droslean: The fallback is wrong. There are dedicated configuration for the openshift private. Consider them as a different repos.

droslean:
I don't think that this is the right approach and it seems like a hack/workaround. The plugin should already know if the repo is private or not, which is not directly related to openshift private. Also the decoration config which needs to be different on private repos, is only related to Prow and not ci-operator.

cmd/payload-testing-prow-plugin/server.go
smith-xyz: yeah looking at this again I was missing some bigger context where metadata already existed, so all these fallbacks aren't necessary. Reshaped the PR with a cleaner approach, I hope

cmd/payload-testing-prow-plugin/server.go
smith-xyz: see above - same issues

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -326,6 +325,9 @@ func (r *reconciler) triggerJobs(ctx context.Context,
 				}
 				continue
 			}
+			if ciopConfig.Prowgen != nil && ciopConfig.Prowgen.Private {
+				aggregatorJob.Spec.Hidden = true
+			}
```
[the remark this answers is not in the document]
smith-xyz: fixed in commit 5e842a0

pkg/controller/prpqr_reconciler/prpqr_reconciler.go
```diff
@@ -326,6 +325,9 @@ func (r *reconciler) triggerJobs(ctx context.Context,
 				}
 				continue
 			}
+			if ciopConfig.Prowgen != nil && ciopConfig.Prowgen.Private && !ciopConfig.Prowgen.Expose {
```
droslean: The prowgen stanza in the ci-operator config is no being used anywhere but only from prowgen. The controller doesn't care about it. If you make the check here, which makes sense, we will just have another dependency of a configuration that is being used by a specific tool only.  When the plugin receives the github comment event, it should know if its private or not. Perhaps it would be better to pass that information through the prpqr resource.

smith-xyz: I see it more clearly now I think, I've modified this so that it becomes part of the prpqr resource.

droslean (approved):

PR #5307, opened 2026-07-10
DPTP-2938: add configurable skip for authoritative decrease by workload type

deepsm007:
Follow-up to [DPTP-2938](https://redhat.atlassian.net/browse/DPTP-2938) 
Hardens pod-scaler against corrupt GCS cache: skip zero-config and >10x increases, reject sub-512Ki recommendations, and replace per-resource cache snapshots on reload so stale RAM data clears without pod restarts.
Adds configurable skip flags for authoritative decrease by workload type/class so OOM'd workloads can be adjusted via escalation and adaptive decrease instead of bad admission mutations.

https://github.com/openshift/release/pull/81763

/cc @openshift/test-platform 

## Summary

Hardens the OpenShift CI pod-scaler admission/recommendation pipeline and makes authoritative decrease behavior safer and more flexible.

### Component behavior changes (pod-scaler)
- **Skips unsafe cache-driven recommendations**: rejects recommendations that don’t meet the pod-scaler “usable minimum” (e.g., memory below **512Ki**), avoids **zero-config** and **over-10x** increases, and sanitizes previously-stamped/corrupt configured resource values before applying increases. Increases are clamped to both cluster caps and configured thresholds (including a max increase ratio), with warnings emitted when recommendation-based increases are capped.
- **Makes GCS cache reload resilient to corruption/staleness**: repairs cached query data on both load and store (removing orphan metadata, corrupt histograms/values, and stale references), and improves reload delivery semantics with retries, freshness checks, and “pending delivery” so early subscribers immediately see the correct cached snapshot.
- **Prevents unschedulable mutations more robustly**: refactors CPU/memory capping logic (with nil-guards) to cap both **requests and limits** consistently.

### Authoritative decrease controls for CI workloads
- Adds **configurable skip flags** to avoid authoritative CPU/memory *decreases* for selected **workload types/classes**. For those workloads, the pipeline uses escalation/adaptive decrease instead of admission-driven authoritative decrease mutations.
- Updates authoritative decrease logic (including usage-basis selection) and changes the apply/dry-run branching and logging to reflect the new configuration model.

### Tests/fixtures
- Updates and expands unit tests for cache repair, reload/pending delivery, recommendation sanitization/capping, authoritative basis/skip parsing, and unschedulable-prevention edge cases.
- Updates pod mutation fixtures and e2e expectations (including authoritative dry-run flag behavior).

### DIFF
--- a/cmd/pod-scaler/admission.go
+++ b/cmd/pod-scaler/admission.go
@@ -19,6 +19,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -41,7 +42,6 @@ import (
 
 type authoritativePair struct {
 	apply        bool
-	dryRun       bool
 	maxReduction float64
 }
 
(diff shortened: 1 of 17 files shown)


deepsm007:
@coderabbitai review

PR #5311, opened 2026-07-13
bug fix: Don't drop secrets from index that don't come from ci-secret-generator config

psalajova:
`ci-secret-generator` runs every 12h and rebuilds the collection index from `_config.yaml`. Secrets created manually (e.g. `secret-manager create cherrypick-signing-key/id_ed25519` using the CLI tool) were dropped from the index each run — GSM kept the secret but the index didn't list it, so `secret-manager list` missed it and `create`/`update` could fail.

**Root cause:** The generator read the previous index but only used it to rescue failed command executions — it never merged in entries not derived from config.

**Fix:** After building index entries from config, merge in previous index entries whose names don't appear in the config-managed set. Index is now sorted for determinism.

**Example:** Some GSM secrets (including `cherrypick-signing-key__id_ed25519`) were missing from the index while present in GSM. After this fix they persist across generator runs.

**Trade-off:** Secrets removed from config will linger in the index (indistinguishable from manual secrets). No GSM-side deletion occurs — cleanup is manual.

---

Fix `ci-secret-generator` dropping manually created GSM secrets during its periodic (12-hour) index rebuilds.

In the `cmd/ci-secret-generator` component, newly generated index entries from `_config.yaml` are now merged with the previous index contents for any entries that are not config-managed (including entries for disabled clusters). The resulting index is then sorted deterministically so `secret-manager` can reliably discover the full set of GSM secrets, not just those coming from config. If a secret is removed from `_config.yaml`, it is kept in the index and requires manual cleanup; the generator does not delete anything in GSM.

Unit tests were updated to verify preservation of manually created entries across rebuilds and to assert the expected deterministic ordering of index fields/items.

### DIFF
--- a/cmd/ci-secret-generator/main.go
+++ b/cmd/ci-secret-generator/main.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"reflect"
+	"sort"
 	"strings"
 
 	secretmanager "cloud.google.com/go/secretmanager/apiv1"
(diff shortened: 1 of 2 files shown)


Prucek (approved):

psalajova:
the e2e test ran for 3h without even starting....
/test e2e

PR #5316, opened 2026-07-16
Derive assemble release version prefix from stream name

deepsm007:
When assembling a release payload without a `release.openshift.io/config` annotation (typical for some presubmits), derive the version prefix from the stream name in `tag_specification` / `integration` (e.g. `5.0` → `5.0.0-0`, `4.19` → `4.19.0-0`) instead of a hardcoded fallback.

Release-controller annotated streams are unchanged: annotation `name` still wins (e.g. `5.0.0-0.ci`).

This keeps ClusterVersion / consumers that parse major in sync across OpenShift 4 and 5 without baking a single major into ci-operator.

/cc @openshift/test-platform

Updates the ci-operator release assembly step to pick the release image version prefix from the configured stream version (from `tag_specification` / `integration`) when the `release.openshift.io/config` annotation is not set. Unannotated streams now convert an input `major.minor` like `5.0` → `5.0.0-0` and `4.19` → `4.19.0-0`; if the derived stream name can’t be parsed as `major.minor` (or the config is nil), it falls back to `0.0.1-0`. When the `release.openshift.io/config` annotation is present, its resolved config `name` still takes precedence and leaves the prefix derivation unchanged. Adds unit coverage for nil/unparseable and multiple edge-case version strings in the new `assembleReleaseVersionPrefix` helper.

### DIFF
--- a/pkg/steps/release/create_release.go
+++ b/pkg/steps/release/create_release.go
@@ -3,6 +3,7 @@ package release
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


JoelSpeed:
/lgtm

This will get us out of a bind, but I'd like to see if there's a better long term solution so that we can keep the major version in sync here

JoelSpeed:
Nice!

/lgtm

deepsm007:
> Nice!
> 
> /lgtm

fixed coderabbit test case suggestions

deepsm007:
We will look into a proper solution, this is a temp fix and will be reverted later

PR #5330, opened 2026-07-24
KFLUXVNGD-1162: graceful teardown on EphemeralCluster deletion

amisstea:
When an EphemeralCluster is deleted, instead of hard-aborting the ProwJob, signal the test step to exit via the test-done-signal secret. This allows the ci-operator workflow to reach its post steps and deprovision the cluster, preventing leaked cloud resources.

When the graceful signal cannot be delivered — because the build client is unavailable or the ci-operator namespace does not exist yet — fall back to aborting the ProwJob and removing the finalizer. This prevents the EphemeralCluster from getting stuck in deletion indefinitely.

The EC reconciler's findCIOperatorTestNS now distinguishes API errors from "namespace not yet created", surfacing real failures instead of masking them as requeue-worthy.

Extract findCIOperatorTestNS and createTestDoneSignalSecret into shared package-level functions to eliminate duplication between the EC reconciler and the ProwJobReconciler.

The ProwJobReconciler also gains a terminal state guard so it no longer attempts to abort or signal ProwJobs that have already completed.

## Summary

EphemeralCluster deletion now performs a graceful ci-operator teardown instead of immediately aborting the ProwJob. The controller signals completion through the `test-done-signal` Secret, allowing post steps to run and the cluster to be deprovisioned before cleanup continues.

If the build client or ci-operator test namespace is unavailable, the controller falls back to aborting the ProwJob and removing the finalizer. Namespace lookup now distinguishes a missing namespace from API errors, with shared helpers used across reconcilers.

The ProwJob reconciler also ignores ProwJobs already in a terminal state, preventing unnecessary provisioning or termination actions. Tests cover graceful signaling, fallback abort behavior, missing namespaces, and completed ProwJobs.

### DIFF
--- a/pkg/controller/ephemeralcluster/prowjobreconciler.go
+++ b/pkg/controller/ephemeralcluster/prowjobreconciler.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/sirupsen/logrus"
 
-	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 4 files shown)


amisstea:
Closing in favor of https://github.com/openshift/ci-tools/pull/5332.

PR #5332, opened 2026-07-24
Konflux: deprovision ephemeral cluster on EphemeralCluster deletion

danilo-gemoli:
This is an alternative path for deprovisioning an ephemeral cluster that triggers when the `EphemeralCluster` object is being deleted.
Konflux users that run the `deprovision-ephemeral-cluster` task might experience a delay that is higher than the `provision-ephemeral-cluster` task. This is counter-intuitive and worsen the user experience.

This PR allows to implement a "fire-and-forget" approach that works that relies solely on the deletion event rather than the `deprovision-ephemeral-cluster` task:
1. The `PipelineRun` completes and a deletion is requested.  
    1.  From the user point of view the test is done, there is no need to wait for anything anymore.
1. The `EphemeralCluster` object deletion is requested too: it doesn't happend right away due to the `ephemeralcluster.ci.openshift.io/dependent-prowjob` finalizer.  
1. The `EphemeralCluster` reconciler starts the deprovision procedure as if `spec.tearDownCluster` was set to `true`.  
1. The procedure completes and the reconciler removes the finalizer.
1. The `EphemeralCluster` is deleted.
1. The `PipelineRun` is deleted.

The point (1i) makes a lot of difference for the users since they won't experience a delay anymore.

## Summary

Updates the `ephemeralcluster` controller to support a fire-and-forget deprovisioning workflow when an `EphemeralCluster` is deleted, so the user can delete the `PipelineRun` without waiting for the `deprovision-ephemeral-cluster` task.

- **New delete reconciliation path:** All delete-timestamp handling is routed through `reconcileDeleteEphemeralCluster`, which keeps the `ephemeralcluster.ci.openshift.io/dependent-prowjob` finalizer while the controller performs deprovision signaling/cleanup, and removes the finalizer once it can persist terminal state.
- **Deprovision via test-complete secret (instead of waiting on deprovision task):** The controller uses `notifyTestComplete` to mark `EphemeralCluster` as deprovisioning and relies on the CI-operator-side `test-done-signal` secret mechanism (the controller-created secret is what unblocks the workflow), allowing the pipeline to complete independently of the deprovision task.
- **Graceful termination skipped for terminal ProwJob states:** `pjInAFinalState` is introduced and used to short-circuit termination/graceful deprovision actions when the ProwJob is already in a terminal/final state.
- **Abort semantics when required dependencies aren’t ready:** If `notifyTestComplete` fails due to typed “not found” cases (missing build-client or ci-operator test namespace), the controller **aborts the ProwJob** and then removes the dependent finalizer when possible, updating `EphemeralCluster` status accordingly.
- **Typed “not found” error handling:** Introduces typed errors (`errBuildClientNotFound`, `errCIOperatorNSNotFound`) and switches logic to use `errors.Is` for consistent branching (including earlier terminal returns in `gracefullyTerminateClusterProvisioning`).
- **Tests/fixtures updated:** Refactors and expands reconciler tests to cover the new delete flow (`TestReconcileDeleteEphemeralCluster`), finalizer removal behavior, terminal-state early returns, and the new abort behavior; updates/refreshes the associated YAML fixtures.


pkg/controller/ephemeralcluster/reconciler.go
```diff
@@ -225,15 +248,18 @@ func (r *reconciler) Reconcile(ctx context.Context, req reconcile.Request) (reco
...
 
 	upsertCondition(&observedStatus, ephemeralclusterv1.ProwJobCreating, ephemeralclusterv1.ConditionFalse, r.now(), ProwJobCreatingDoneReason, "")
 	observedStatus.Phase = ephemeralclusterv1.EphemeralClusterProvisioning
 	observedStatus.ProwJobURL = pj.Status.URL
 
-	err := r.fetchSecrets(ctx, log, ec, &oldStatus, &observedStatus, &pj)
-	if err != nil {
+	if err := r.fetchSecrets(ctx, log, ec, &oldStatus, &observedStatus, &pj); err != nil {
+		if errors.Is(err, &errBuildClientNotFound{}) {
+			return r.handleBuildClientNotFoundError(ctx, log, ec, &observedStatus, &pj, err)
```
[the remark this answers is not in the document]
danilo-gemoli: If a PJ is in a terminal state, the reconciler wouldn't have run at all. You are describing a very unlikely edge case scenario:
1. The build client disappeared for whatever reason (cluster has been disabled perhaps?)
2. The reconciler is running and the PJ has completed in the middle of the current reconciliation loop

This is a special scenario, I'm ok if the PJ is marked as `Failed`

pkg/controller/ephemeralcluster/prowjobreconciler.go
```diff
@@ -104,23 +104,27 @@ func (r *prowJobReconciler) Reconcile(ctx context.Context, req reconcile.Request
...
 func (r *prowJobReconciler) gracefullyTerminateClusterProvisioning(ctx context.Context, log *logrus.Entry, pj *prowv1.ProwJob) (reconcile.Result, error) {
-	buildClient, err := r.buildClients.forCluster(pj.Spec.Cluster)
-	if err != nil {
+	if pjInAFinalState(pj) {
+		log.WithField("pj_state", pj.Status.State).Info("ProwJob in a final state, skipping graceful termination")
+		return reconcile.Result{}, nil
+	}
+
+	buildClient, err := buildClientFor(r.buildClients, pj.Spec.Cluster)
+	if err != nil && errors.Is(err, &errBuildClientNotFound{}) {
```
amisstea, line 113: If `err` is non-nil and not a `errBuildClientNotFound` error, this will likely fall through and pass a nil `buildClient` to `findCIOperatorTestNS` and cause a panic.

danilo-gemoli: By looking at the code, it seems it's not a possible scenario:
```golang
func buildClientFor(bc buildClients, cluster string) (ctrlruntimeclient.Client, error) {
	buildClient, ok := bc[cluster]
	if !ok {
		return nil, &errBuildClientNotFound{cluster: cluster}
	}
	return buildClient, nil
}
```
`err != nil` => `error.Is(err, &errBuildClientNotFound)`

amisstea: Ok, LGTM then.

amisstea (approved):

PR #5338, opened 2026-07-28
fix clonerefs in test pods to use the correct oauth secret from the temp namespace

droslean:
Updates ci-operator test pod generation so clonerefs uses the correct OAuth secret from the temporary namespace. Clone authentication now flows from test configuration into pod generation and sets the pod decoration’s OAuth token secret when cloning is enabled.


Prucek (approved):

deepsm007 (approved):

deepsm007:
/override ci/prow/integration 
unrelated failures
/test e2e

PR #5340, opened 2026-07-29
clusterimageset-updater: replace integration test with unit test

Prucek:
## Summary
- The `clusterimageset-updater` integration test hit the real release controller API, so it broke every time a new z-stream was published.
- Extract core logic into a testable `run(options, HTTPClient)` function.
- Add a `--release-controller-url` flag and `EndpointWithBase` helpers to support URL override.
- Add a unit test using `httptest.NewServer` with `testhelper.CompareWithFixture` golden files — deterministic, no external dependencies.
- Remove the shell-based integration test and its fixtures.


test/integration/clusterimageset-updater.sh
```diff
@@ -14,9 +17,39 @@ cp -a "${suite_dir}"/* "${workdir}"
 inputs="${workdir}/input"
 expected="${workdir}/output"
 
+# Start a mock release controller that returns a deterministic version.
+MOCK_PORT=$(python3 -c 'import socket; s=socket.socket(); s.bind(("",0)); print(s.getsockname()[1]); s.close()')
```
droslean: I don't think that you should do that.... I guess your AI is trying to hack the hell out of it.

PR #5343, opened 2026-07-29
feat(check-gh-automation): add caching, --warn-only and --only-org

Prucek:
## Summary
- Add org-level `IsMember` caching to eliminate redundant GitHub API calls across repos in the same org, reducing runtime from ~13h to ~5h
- Add `--warn-only` flag to report failures as warnings (exit 0) instead of fatal errors
- Add `--only-org` flag to restrict checking to specific orgs, enabling job splitting for parallel execution

## Summary

Improves the `check-gh-automation` CI validation tool:

- Caches GitHub organization membership checks to reduce API calls and runtime.
- Adds `--warn-only` to report missing automation without failing the job.
- Adds `--only-org` to support parallel checks for selected organizations.
- Applies organization filtering to explicit repositories, candidate paths, and default repository lists.
- Adds tests for organization filtering and membership-cache behavior.

### DIFF
--- a/cmd/check-gh-automation/main.go
+++ b/cmd/check-gh-automation/main.go
@@ -40,11 +40,38 @@ type options struct {
 	appName               string
 	appCheckMode          string
 	checkBranchProtection bool
+	warnOnly              bool
 	ignore                flagutil.Strings
+	onlyOrgs              flagutil.Strings
 	repos                 flagutil.Strings
 	releaseRepoPath       string
 	flagutil.GitHubOptions
 	pluginConfig prowpluginconfig.PluginOptions
+}
+
+type membershipCache struct {
+	client automationClient
+	cache  map[string]bool
+}
+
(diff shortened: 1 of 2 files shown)


psalajova:
I love this change 💪 is the `--only-org` flag to enable parallel executions something we plan to do, run the `check-gh-automation` that way?

cmd/check-gh-automation/main.go
psalajova, line 156: A nit, but if `--only-org` flag is set and doesn't match any repos, checkRepos processes zero repos and main() logs "All repos have github automation configured." -- technically true but misleading.
```suggestion
	repos := determineRepos(o, prowAgent, logger)
	if len(repos) == 0 {
    	  logger.Warnf("No repos to check.")
    	  return
        }
```

psalajova (commented):
Left one comment, up to you whether you think it's worth adding

PR #5346, opened 2026-07-30
Fix dispatcher capacity and add ipCapacity field to count for maxNodes

deepsm007:
Dispatcher used to balance jobs only with capacity (how much load we want on a cluster). That ignored how many nodes a cluster can actually grow to before it runs out of subnet IPs so tight /26 farms (e.g. build09) got the same pressure as roomy /24 ones (e.g. build05) and hit scale/IP limits first.

This adds optional ipCapacity (usable IPs / max nodes from the CIDR) and weights load as capacity × (ipCapacity / farm max). Same capacity → bigger CIDR takes more jobs; cutting capacity 100→50 still halves load. Missing ipCapacity keeps the old capacity-only behavior.

/cc @jmguzik @openshift/test-platform 

[Reference thread](https://redhat-internal.slack.com/archives/GB7NB0CUC/p1785437916744409?thread_ts=1785420991.228669&cid=GB7NB0CUC)

## Summary

Enhances the Prow job dispatcher’s load balancing to optionally account for farm subnet IP capacity.

- Cluster load uses `capacity × (ipCapacity / maxIPCapacity)`.
- Farms with larger CIDRs receive more jobs at equal capacity.
- Reduced `ipCapacity` proportionally reduces a farm’s load.
- Farms without `ipCapacity` retain capacity-only behavior.
- Fallback selection skips blocked, ineligible, and invalid-capacity clusters.
- Cluster configuration loading and change detection now support `ipCapacity`.
- Prometheus volume distribution uses the same weighting.
- Tests cover configuration validation, cluster selection, load scoring, and volume distribution.

### DIFF
--- a/cmd/prow-job-dispatcher/main.go
+++ b/cmd/prow-job-dispatcher/main.go
@@ -208,9 +208,11 @@ type clusterVolume struct {
 	clusterMap         dispatcher.ClusterMap
 }
 
+func clusterStressScore(volume float64, info dispatcher.ClusterInfo, maxIP int) float64 {
+	return volume / dispatcher.LoadWeight(info, maxIP)
+}
+
 // findClusterForJobConfig finds a cluster running on a preferred cloud provider for the jobs in a Prow job config.
-// The chosen cluster will be the one with minimal workload with the given cloud provider.
-// If the cluster provider is empty string, it will choose the one with minimal workload across all cloud providers.
 func (cv *clusterVolume) findClusterForJobConfig(cloudProvider string, jc *prowconfig.JobConfig, path string, config *dispatcher.Config, jobVolumes map[string]float64) (string, error) {
 	if _, ok := cv.clusterVolumeMap[cloudProvider]; !ok {
 		cloudProvider = ""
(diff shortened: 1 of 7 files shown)


deepsm007:
/override ci/prow/integration
Unrelated to PR changes, failing for cluster imageset updater. Peter is working on a fix.

jmguzik (commented):
The overall approach of combining dispatcher capacity with IP capacity is reasonable. The focused dispatcher tests pass on the current PR head. However, the PR should not be merged in its current form IMO.
  
Mixed configurations with omitted and configured ipCapacity values produce unstable weights. A constrained cluster can receive full weight until another configured cluster is added. This means a partial rollout may fail to protect clusters with limited IP capacity. The configuration should either require ipCapacity for all active clusters or use a stable default baseline. The dispatcher should also return an error when no eligible fallback or signalize somehow this problem. Negative ipCapacity values should be rejected during configuration loading.

jmguzik:
The best of course would be to have the same cluster ip ranges but that requires work.

PR #5348, opened 2026-08-01
release: wait for CLI tag spec visibility

not-stbenjam:
## What changed

- extend the shared ImageStream import evaluator with opt-in waiting for requested tags to appear in `spec.tags`
- use that mode in both release CLI resolution paths
- keep the existing ten-minute bound and preserve immediate missing-tag errors for callers that do not opt in
- add direct regression coverage for the evaluator, both call sites, and the timeout path

## Why

An ImageStreamTag update can complete before the parent ImageStream observation contains `spec.tags[cli]`. The shared import evaluator treated that temporary absence as terminal, so release jobs could fail before cluster installation.

The retained OpenShift 5.0 CI corpus contains 37 exact failures with this `[cli]` signature. Seven occurred from July 26 through July 30 across Azure and GCP jobs, stable and stable-initial streams, nightly and CI payloads, and three ci-operator versions. In the confirmed target run, the CLI extractor succeeded and the final ImageStream contained a populated CLI spec and status entry, but the initial release graph had already failed.

## Impact

CLI resolution now keeps watching through temporary spec visibility lag. Persistent missing tags and incomplete imports still fail within the existing ten-minute timeout. Other import-wait callers retain their current immediate missing-definition behavior.

## Validation

- `go test -p 1 ./pkg/steps/utils -run '^(TestGetEvaluator|TestWaitForImportingISTagSpecTimeout)$' -count=1 -v`
- `go test -p 1 ./pkg/steps/release -run '^(TestResolveCLIImageFromStreamWaitsForSpecVisibility|TestExtractAndTagCLIImageWaitsForSpecVisibility)$' -count=1 -v`
- `go test -p 1 ./pkg/steps/utils ./pkg/steps/release -count=1`
- `gofmt`
- `git diff --check`

The call-site tests model spec visibility and import status as separate events on one watch. Removing the `WaitForSpecTags` option fails on the original `missing definition in the spec` path.

## Summary

The release CLI now optionally waits for requested tags to appear in an ImageStream’s `spec.tags` before it evaluates import status.

Both release image-resolution paths use this wait mode. Existing callers still fail immediately when a requested tag is missing. The existing ten-minute timeout remains in effect for missing tags and incomplete imports.

This prevents failures caused by temporary ImageStream observation lag in CI release workflows.

## Testing

- Added regression tests for both release CLI resolution paths.
- Added timeout coverage for persistently missing spec tags.
- Ran targeted and package tests.
- Ran `gofmt` and `git diff --check`.

### DIFF
--- a/pkg/steps/release/import_release.go
+++ b/pkg/steps/release/import_release.go
@@ -512,7 +512,7 @@ func waitForCLIImageTagInSpec(ctx context.Context, client ctrlruntimeclient.With
 	imageStream := &imagev1.ImageStream{}
 	evaluate := func(obj runtime.Object) (bool, error) {
 		stream, ok := obj.(*imagev1.ImageStream)
-		if !ok {
+		if !ok || stream == nil {
 			return false, fmt.Errorf("got an event that did not contain an imagestream: %v", obj)
 		}
 		for _, tag := range stream.Spec.Tags {
--- a/pkg/steps/release/import_release.go
+++ b/pkg/steps/release/import_release.go
@@ -14,6 +14,7 @@ import (
 	kerrors "k8s.io/apimachinery/pkg/api/errors"
 	meta "k8s.io/apimachinery/pkg/apis/meta/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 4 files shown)


pkg/steps/release/import_release.go
```diff
@@ -487,13 +494,38 @@ func (s *importReleaseStep) extractAndTagCLIImage(ctx context.Context, targetCLI
...
 		return nil, fmt.Errorf("unable to wait for the 'cli' image in the stable stream to populate: %w", err)
 	}
 
 	return &cliImageRef, nil
 }
 
+// waitForCLIImageTagInSpec waits for the CLI tag definition to be visible before
+// inspecting its import status. An ImageStreamTag update can succeed before the
+// corresponding tag is returned in the parent ImageStream spec.
+func waitForCLIImageTagInSpec(ctx context.Context, client ctrlruntimeclient.WithWatch, namespace, streamName string, timeout time.Duration) error {
```
Prucek: I don't think this is the right fix. `utils.WaitForImportingISTag` does the same thing, and the fix should land there - in the `getEvaluator` where the real bug lies

Prucek: maybe a separate evaluator and timeout for the cli

stbenjam:
/ok-to-test

@coderabbitai review

Prucek (approved):

PR #5349, opened 2026-08-01
NO-JIRA: report unfinished aggregated jobs correctly

not-stbenjam:
A tolerated unfinished Prow job is currently appended to `AggregationConfiguration.FinishedJobs`, even though the result type has a dedicated `UnfinishedJobs` field. This makes `aggregation-config.yaml` report an unknown/incomplete child as finished precisely when downstream analysis needs an accurate child manifest.

This changes the assignment to `UnfinishedJobs` and adds regression coverage for the supported 7-finished/3-unfinished case, while preserving the existing terminal aggregate verdict for missing data.

Study context: an exact reconciliation of 570 OpenShift 5.0 aggregate parents found 95 parents with shared missing-data cascades. This PR intentionally does not change gating or claim green-job rescue; it corrects the machine-readable attribution boundary needed for follow-up missing-data collapse work.

Validation:
- `GOMAXPROCS=2 go test -p=2 ./pkg/jobrunaggregator/jobrunaggregatoranalyzer`
- `git diff --check`

## Summary

- Record tolerated unfinished Prow jobs in `AggregationConfiguration.UnfinishedJobs` instead of `FinishedJobs`.
- Preserve the terminal aggregate verdict when job data is missing.
- Add regression coverage for seven finished and three unfinished jobs.
- Do not change gating or green-job rescue behavior.

## Validation

- Targeted Go tests
- `git diff --check`

### DIFF
--- a/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
+++ b/pkg/jobrunaggregator/jobrunaggregatoranalyzer/analyzer.go
@@ -167,8 +167,8 @@ func (o *JobRunAggregatorAnalyzerOptions) Run(ctx context.Context) error {
 		if len(o.explicitGCSPrefix) > 0 {
 			jobRunGCSBucketRoot = filepath.Join(o.explicitGCSPrefix, jobRunName)
 		}
-		aggregationConfiguration.FinishedJobs = append(
-			aggregationConfiguration.FinishedJobs,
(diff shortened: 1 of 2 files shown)


stbenjam:
It is not clear to me this actually does anything useful,  finished = "jobs complete", it's a tracking mechanism

PR #5363, opened 2026-08-12
pkg/steps: retry network-unreachable builds, capture failed build logs

kaovilai:
## Summary
- `hintsAtInfraReason()` (`pkg/steps/source.go`) now matches `"network is unreachable"` and `"no route to host"`, so a transient ENETUNREACH-style dial failure during a build (e.g. a Go module/sumdb fetch mid-Dockerfile-build) is classified as an infra-side retryable failure instead of falling through to "classified as legitimate failure, will not be retried".
- Renamed the outcome-agnostic `gatherSuccessfulBuildLog` to `gatherBuildLog` (`pkg/steps/artifacts.go`) and call it from both the success path and the failure path (`waitForBuild`'s `BuildPhaseFailed`/`Cancelled`/`Error` case, alongside the existing `printBuildLogs` stdout dump), so failed builds get the same discrete `build-logs/<name>.log.gz` artifact successful builds already do, instead of only being interleaved into ci-operator's own stdout log.

No config schema changes. No behavior change for legitimate (non-infra) `DockerBuildFailed` failures — those still won't retry, they just also get a log artifact now.

## Context
Root-caused from a `DockerBuildFailed` flake on an `oadp-operator` rehearsal build (`oadp-operator-1.5-amd64`): a sibling job on the identical commit/Dockerfile succeeded in parallel on a different build pod, confirming a build-farm-node-level transient network flake rather than a real Dockerfile bug. The failing line (`dial tcp [...]:443: connect: network is unreachable`) was hard to find because it was buried inside the monolithic `build-log.txt` rather than a small dedicated per-build log artifact.

## Test plan
- [x] `go build ./pkg/steps/...`
- [x] `go vet ./pkg/steps/...`
- [x] `gofmt -l` clean on both changed files
- [x] `go test ./pkg/steps/...` — all packages pass

## Summary

- Updates `pkg/steps` to retry builds that fail with `"network is unreachable"` or `"no route to host"`.
- Captures `build-logs/<name>.log.gz` artifacts for successful, failed, cancelled, and errored builds.
- Preserves non-retryable behavior for legitimate `DockerBuildFailed` failures.
- Renames the internal helper to `gatherBuildLog`.
- No configuration schema changes.
- Tests and formatting checks pass.

### DIFF
--- a/pkg/steps/artifacts.go
+++ b/pkg/steps/artifacts.go
@@ -664,12 +664,12 @@ func gatherContainerLogsOutput(podClient kubernetes.PodClient, artifactDir, name
 	return utilerrors.NewAggregate(validationErrors)
 }
 
-// for gathering successful build logs to the artifacts, there is no way to augment the pod spec
+// for gathering build logs (successful or failed) to the artifacts, there is no way to augment the pod spec
 // created by the build controller to add the artifacts container; this method cherry picks elements
 // from downloadArtifacts and gatherContainerLogsOutput and munges them in conjunction with the build
 // api logging capabilities; also, without needing to inject an artifacts container, some of the complexities
 // around download/copy from the artifacts container's volume mount and multiple pods are avoided.
-func gatherSuccessfulBuildLog(buildClient BuildClient, namespace, buildName string) error {
+func gatherBuildLog(buildClient BuildClient, namespace, buildName string) error {
 	artifactDir, set := api.Artifacts()
 	if !set {
 		return nil
--- a/pkg/steps/source.go
+++ b/pkg/steps/source.go
@@ -590,7 +590,7 @@ func handleBuild(ctx context.Context, client BuildClient, podClient kubernetes.P
 			errs = append(errs, err)
 			return false, handleFailedBuild(ctx, client, ns, name, err)
 		}
-		if err := gatherSuccessfulBuildLog(client, ns, name); err != nil {
+		if err := gatherBuildLog(client, ns, name); err != nil {
 			// log error but do not fail successful build
 			logrus.WithError(err).Warnf("Failed gathering successful build %s logs into artifacts.", name)
 		}
@@ -659,7 +659,9 @@ func hintsAtInfraReason(logSnippet string) bool {
 		strings.Contains(logSnippet, "Could not resolve host: ") ||
 		strings.Contains(logSnippet, "net/http: TLS handshake timeout") ||
 		strings.Contains(logSnippet, "All mirrors were tried") ||
-		strings.Contains(logSnippet, "connection reset by peer")
+		strings.Contains(logSnippet, "connection reset by peer") ||
+		strings.Contains(logSnippet, "network is unreachable") ||
+		strings.Contains(logSnippet, "no route to host")
 }
 
 func waitForBuildOrTimeout(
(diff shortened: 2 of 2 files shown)


kaovilai:
/test core-valid

> [!Note]
> Responses generated with Claude

kaovilai:
/test owners

> [!Note]
> Responses generated with Claude

kaovilai:
/test release-controller-config

> [!Note]
> Responses generated with Claude

kaovilai:
/test openshift-image-mirror-mappings

> [!Note]
> Responses generated with Claude

kaovilai:
/test images

> [!Note]
> Responses generated with Claude

deepsm007 (approved):

kaovilai:
/test core-valid

> [!Note]
> Responses generated with Claude

kaovilai:
/test owners

> [!Note]
> Responses generated with Claude

kaovilai:
/test release-controller-config

> [!Note]
> Responses generated with Claude

kaovilai:
/test openshift-image-mirror-mappings

> [!Note]
> Responses generated with Claude

kaovilai:
Apologies for the duplicate `/test core-valid`, `/test owners`, and `/test release-controller-config` comments above — a monitoring script restart re-triggered nudges for checks it had already flagged, and it also didn't account for this repo's pipeline-controller gating (those second-stage tests aren't stuck, they trigger on `lgtm`/automatic mode by design). Fixed the script to dedupe against actual PR comment history and to recognize pipeline-controller repos so this won't recur.

> [!Note]
> Responses generated with Claude

kaovilai:
Heads up: `ci/prow/images` has now failed **5/5 times** on this exact commit (`29943ddd7ac`), always at the same spot — `bin-arm64` build, `go install` in `hack/install.sh:45` exiting 1 after ~3-7min — while `bin-amd64` succeeds every time. That's a reproducible failure, not a flake, so keeping `/test images` isn't the right hammer here.

Can't get the actual compiler error though: `artifacts/build-logs/` only captures logs for successful builds today (only `bin-amd64.log` is present), so the `bin-arm64` stderr that would show *why* `go install` fails is lost — ironically the log-capture-on-failure feature this very PR adds isn't live yet to catch its own build's failure. The diff itself (renaming a helper, adding failed-build log capture, 2 new infra-string matches) doesn't look like it should affect compilation on any arch, so this may be arm64-builder-side rather than code — but I don't have enough visibility to confirm either way from here.

> [!Note]
> Responses generated with Claude

kaovilai:
Update: root-caused (with help from a local repro on real arm64 hardware). This is **not** a regression from this PR's diff.

- Checked out this PR at head `29943ddd7ac` locally on arm64 and ran `GOOS=linux GOARCH=arm64 CGO_ENABLED=0 go build ./...`, plus the full `make production-install` (`hack/install.sh` + npm/webpack dist assets) — both succeed cleanly, no compile errors.
- `ci-operator.log` for these runs is running the *released* ci-operator (`v20260810-3e8efbe0b`, built from base SHA), so this PR's `gatherBuildLog` change isn't even active yet — and indeed the log shows `Unable to retrieve logs from failed build: pod bin-arm64-build not found`, literally the gap this PR closes.
- `bin-arm64` gets the same `cpu:1/memory:1Gi` request as `bin-amd64`. `bin-amd64` needed 23m18s to succeed; `bin-arm64` failed at ~7m16s — consistent with the arm64 build hitting a resource ceiling (OOM/starved) on that node rather than a real compile error.
- Cross-check: PR #5346, same base SHA, `images` job passed clean.

So: looks like an infra-side thin-margin issue on the arm64 builder, not code. Retrying once more; if it keeps failing deterministically, worth flagging to DPTP that the arm64 `images` build (go install of ~50 binaries + webpack bundle) may need more than 1Gi.

/test images

> [!Note]
> Responses generated with Claude

kaovilai:
mem bump pr: https://github.com/openshift/release/pull/83341 for arm64 build issue, similar did wonder for my teams' oadp-operator arm64 build flakes.

PR #5375, opened 2026-08-25
prowgen: add managed-repos config and atomic --from-file mode

Prucek:
## Summary

Part of the plan to move generated Prow jobs off of `ci-operator/jobs/` in `openshift/release` and onto EFS, without disrupting the existing git-based flow.

- Adds `pkg/prowgen.ManagedReposConfig`, a YAML config mapping `org/repo` to which of its branches are "managed elsewhere" (e.g. already onboarded onto EFS):
  - `allBranches: true` — the whole repo.
  - `fromRelease: "4.22"` — only `release-X.Y` branches at/after that version (non-release branches like `main`, and older release branches, are unaffected — supports mixed mode during a repo's migration).
  - `branches: [...]` — explicit extra managed branches.
  - `excludeBranches: [...]` — escape hatch that always wins.
- `ci-operator-prowgen` gains `--managed-repos-config <path>`. Managed org/repo@branch entries are skipped entirely: no jobs generated for them, and their existing files under `--to-dir` are left untouched (not pruned) since another process now owns them.
- `auto-config-brancher` passes `--managed-repos-config` through to the prowgen step when configured.
- Adds a `--from-file <path>` mode to `ci-operator-prowgen` that generates jobs for a single ci-operator config file (org/repo/branch read from its own `zz_generated_metadata`) and writes them via a new `jc.WriteBranchToDir`, which only touches that file's own branch and writes atomically (temp file + rename) via `jc.WriteToFileAtomic`. This is scoped to `--from-file` only — the existing bulk `--from-dir`/`WriteToDir` path is unchanged and still writes non-atomically, since it targets a git checkout that gets committed rather than a volume read concurrently by live Prow components.

Made with [Cursor](https://cursor.com)

## Summary

- Adds managed-repository rules to `ci-operator-prowgen`. Operators can skip all branches, selected branches, or release branches from a minimum version. Exclusions take precedence.
- Adds `--managed-repos-config` and passes it through `auto-config-brancher`. Managed branches are neither generated nor pruned, which preserves externally managed jobs.
- Adds `--from-file` for postsubmit-only workflows. The command generates jobs from one ci-operator configuration and atomically updates only its branch.
- Preserves other branch files and validates organization and repository path components during single-file output.
- Adds atomic file-writing APIs and tests for branch isolation, stale-file handling, cleanup, metadata validation, path safety, and managed-repository rules.


Prucek:
unrelated `✖  test/e2e/gsm-secret-sync` failure
/override ci/prow/e2e

cmd/ci-operator-prowgen/main.go
```diff
@@ -21,13 +23,29 @@ import (
...
 	"github.com/openshift/ci-tools/pkg/util"
+	"github.com/openshift/ci-tools/pkg/util/gzip"
 )
 
 type options struct {
 	config.Options
 
 	fromDir         string
 	fromReleaseRepo bool
+	// fromFiles, if non-empty, generates jobs for one or more single
```
droslean: I don't think that you need to add that big comment in code just to document it.

cmd/ci-operator-prowgen/from_file_test.go
```diff
@@ -0,0 +1,236 @@
+package main
```
droslean: Can you rewrite the whole file with correct Go tests? Also the name of the file is wrong as well.

cmd/ci-operator-prowgen/managed_test.go
```diff
@@ -0,0 +1,163 @@
+package main
```
droslean: Same here.  Also the name of the file is wrong. It should be main_test.go

pkg/prowgen/managed.go
```diff
@@ -0,0 +1,140 @@
...
+
+// releaseVersionFromBranch extracts the OCP release version encoded in
+// branch names like "release-4.22". Branches that don't follow this
+// convention (main, master, feature branches, ...) return ok=false.
+func releaseVersionFromBranch(branch string) (api.ParsedVersion, bool) {
+	m := releaseBranchRE.FindStringSubmatch(branch)
+	if m == nil {
+		return api.ParsedVersion{}, false
+	}
+	major, _ := strconv.Atoi(m[1])
```
droslean: This will panic for wrong m length. Also, you shouldn't eat any errors.

pkg/jobconfig/files_test.go
```diff
@@ -1110,3 +1112,231 @@ func TestStaleSelectorFor(t *testing.T) {
 		})
 	}
 }
+
+func TestWriteToFileAtomic(t *testing.T) {
```
droslean: Can you rewrite all those tests too?

pkg/jobconfig/files.go
```diff
@@ -609,6 +738,41 @@ func WriteToFile(path string, jobConfig *prowconfig.JobConfig) error {
 	return nil
 }
 
+// WriteToFileAtomic writes Prow job config to a YAML file the same way
+// WriteToFile does, but readers that concurrently poll/read the target path
+// (e.g. Prow components reading job configs off a shared EFS volume) never
+// observe a partially written file: the new content is written to a temp
+// file in the same directory and then atomically renamed into place. Used by
+// WriteBranchToDir (--from-file mode).
+func WriteToFileAtomic(path string, jobConfig *prowconfig.JobConfig) error {
```
droslean, line 630: The whole generation process seems like a hack. Creating a temp file and then rename etc... Why you can't just generate it? Are we using the same logic in the other parts of generation?

pkg/jobconfig/files.go
```diff
@@ -371,6 +372,134 @@ func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, gener
...
+// stale jobs (unlike WriteToDir, it never reads existing files to merge or
+// prune against) -- except for jobs of the same ci-operator variant being
+// written, which fully replace what was there before (see
+// mergeOtherVariants). This is used by --from-file mode, where prowgen
+// generates jobs for a single org/repo@branch@variant at a time and must not
+// affect any other branch's files, nor any sibling variant's jobs that share
+// the same branch's output file. Writes are atomic (see WriteToFileAtomic)
+// since this mode is intended for targets like EFS that are read
+// concurrently by live Prow components.
+func WriteBranchToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, generator Generator) error {
```
droslean: This looks like a logic that belongs to the Generator.

pkg/jobconfig/files.go
Prucek: Yes — this mirrors the same temp-file + rename pattern already used in `pkg/dispatcher/gob.go` (`writeGob`), for the same underlying reason: these job files are written directly onto the EFS volume while live Prow components (deck, sinker, etc.) are reading that same volume concurrently. A plain `os.WriteFile` isn't atomic, so a reader could see a truncated/partial YAML file mid-write.

This is the same problem `git-sync` solves for a whole checkout, by atomically swapping a symlink to point at a newly-synced worktree once it's complete. We need the same guarantee here, just at the level of a single file instead of a whole directory, since `--from-file` writes one job file per invocation rather than syncing an entire tree.

cmd/auto-config-brancher/main_test.go
```diff
@@ -3,9 +3,61 @@ package main
...
 	"os"
 	"os/exec"
+	"reflect"
 	"testing"
+
+	"github.com/openshift/ci-tools/pkg/config"
+	"github.com/openshift/ci-tools/pkg/promotion"
 )
 
+func optionsWithConfigDir(configDir string) options {
```
droslean: You shouldn't need a helper to use it once. Also this has to be inside the testcase to make them more readable.

cmd/ci-operator-prowgen/main.go
```diff
@@ -121,26 +214,38 @@ func (o *options) generateJobsToDir(subDir string) error {
 	return writeToDir(o.toDir, generated)
 }
 
-func generateJobs(resolver registry.Resolver, output map[string]*prowconfig.JobConfig) func(configSpec *cioperatorapi.ReleaseBuildConfiguration, info *config.Info) error {
+// clusterProfileResolverFor builds a prowgen.ClusterProfileResolver backed
+// by resolver, or a resolver that always errors if resolver is nil (no step
+// registry configured).
+func clusterProfileResolverFor(resolver registry.Resolver) func(name string) (*api.ClusterProfile, error) {
```
droslean: This looks like a method of the registry.Resolver

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -310,3 +311,399 @@ tests:
...
+			o := &options{managedRepos: tc.managedRepos}
+			if got := o.skipRepoBranch(tc.orgRepo, tc.branch); got != tc.expected {
+				t.Errorf("skipRepoBranch(%q, %q) = %v, want %v", tc.orgRepo, tc.branch, got, tc.expected)
+			}
+		})
+	}
+}
+
+// writeConfigYAML writes a minimal, valid ci-operator config for org/component@branch.
+func writeConfigYAML(t *testing.T, configDir, component, branch string) {
```
droslean: The fact that you need to do this in the tests indicates that the `generateJobsToDir` has to be refactored to make it more testable. Also you shouldn't need to add the logic to a helper fn

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -310,3 +311,399 @@ tests:
...
+  commands: make test-unit
+  container:
+    from: src
+`)
+	if err := os.WriteFile(path, content, 0664); err != nil {
+		t.Fatalf("failed to write config file: %v", err)
+	}
+}
+
+func TestGenerateJobsToDirManagedRepos(t *testing.T) {
```
droslean: There is no GenerateJobsToDirManagedRepos() anywhere. I guess this is custom logic for testing generated by the ai.

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -310,3 +311,399 @@ tests:
...
+}
+
+type fromFileConfig struct {
+	name    string
+	content []byte
+}
+
+// unitTestConfigYAML returns a minimal ci-operator config with an explicit
+// zz_generated_metadata block, as --from-file mode requires.
+func unitTestConfigYAML(repo, variant string) []byte {
```
droslean: Same here. The fact that you need to do this, something is smelly. The tested function should be able to operate on a ciop config as a struct.

pkg/jobconfig/files.go
```diff
@@ -371,6 +372,117 @@ func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, gener
...
+	if value == "" || value != filepath.Base(value) || value == "." || value == ".." {
+		return fmt.Errorf("invalid %s %q", name, value)
+	}
+	return nil
+}
+
+// WriteBranchToDir atomically writes only the branch's own files, used by
+// --from-file mode to update a single org/repo@branch without touching
+// anything else in jobDir.
+func WriteBranchToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, generator Generator) error {
```
droslean: But this logic should already exist. Can you just expand the previous logic instead of introducing a new one?

pkg/jobconfig/files_test.go
```diff
@@ -1110,3 +1113,235 @@ func TestStaleSelectorFor(t *testing.T) {
...
+			if tc.verify != nil {
+				if err := tc.verify(t, componentDir); err != nil {
+					t.Error(err)
+				}
+			}
+		})
+	}
+}
+
+func assertNoTempFiles(t *testing.T, dir string) {
```
droslean: This shouldn't happen at all. You are trying to run integration tests via a uni test hacked logic.

cmd/ci-operator-prowgen/main_test.go
```diff
@@ -310,3 +312,237 @@ tests:
...
+		t.Run(tc.name, func(t *testing.T) {
+			tempDir := t.TempDir()
+			configDir := filepath.Join(tempDir, "config")
+			jobsDir := filepath.Join(tempDir, "jobs")
+			for _, c := range tc.configs {
+				dir := filepath.Join(configDir, "org", c.component)
+				if err := os.MkdirAll(dir, os.ModePerm); err != nil {
+					t.Fatalf("failed to create config dir: %v", err)
+				}
+				if err := os.WriteFile(filepath.Join(dir, "org-"+c.component+"-"+c.branch+".yaml"), []byte(`build_root:
```
droslean: This shouldn't happen on unitesting. My guess is that you want to write integration tests. Why not adding your extra cases there?

pkg/jobconfig/files.go
```diff
@@ -319,58 +318,85 @@ func WriteToDir(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, gener
...
 		}
 	}
 	return nil
 }
 
+// WriteToDirAtomic writes job configuration atomically, merging with any
+// existing content but without scanning the full directory or pruning stale
+// jobs. Used by --from-file mode to update a single config's output files
+// without touching anything else in jobDir.
+func WriteToDirAtomic(jobDir, org, repo string, jobConfig *prowconfig.JobConfig, generator Generator) error {
```
droslean: WriteToDirAtomic seems like WriteToDir with some extra options.

pkg/jobconfig/files_test.go
```diff
@@ -1110,3 +1113,326 @@ func TestStaleSelectorFor(t *testing.T) {
...
+		{
+			name: "does not touch other branches' files",
+			org:  "org",
+			repo: "repo",
+			preseed: func(t *testing.T, componentDir string) {
+				if err := os.MkdirAll(componentDir, 0755); err != nil {
+					t.Fatalf("failed to create component dir: %v", err)
+				}
+				otherBranchContent := []byte(`presubmits:
+  org/repo:
```
droslean: Same here. This is not unitesting. The logic in WriteToDirAtomic seems wrong then.

pkg/prowgen/managed_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}}
+	if !cfg.IsManaged("org/repo", "main") {
+		t.Errorf("expected org/repo@main to be managed")
+	}
+	if cfg.IsManaged("org/other", "main") {
+		t.Errorf("expected org/other to not be managed (not in config)")
+	}
+}
+
+func TestLoadManagedReposConfig(t *testing.T) {
```
droslean: The test format are not idiomatic. Also, I don't think you need to test this here. In theory, you are not testing anything.

jmguzik (approved):

Prucek:
/hold
I want to watch the rolling

PR #5391, opened 2026-09-04
gsm-secret-sync: fix IAM binding limits and support unclaimed collections

psalajova:
## What

Two related fixes to `gsm-secret-sync`, in separate commits:

### 1. Chunk group IAM bindings and scope service accounts per collection
The reconciler emitted one combined viewer/updater binding per collection whose members were the owning groups **plus** the collection's updater service account. A group owning many collections thus appeared in that many bindings sharing the same `(role, member)`, hitting GCP's limit of **20 bindings per role+member** (`multiarch-devel`, 23 collections, failed to apply).

Now the desired bindings are split into:
- **Per-collection, service-account-only** bindings scoped to exactly one collection, so a service account never gains access to another collection's secrets.
- **Per-group** bindings whose condition covers the group's collections, chunked to at most `MaxCollectionsPerGroupBinding` (5). The viewer condition uses `2N+1` logical operators for N collections, so 5 stays within GCP's cap of **12 operators per condition**; chunking also keeps a group under the 20-bindings cap for up to 100 collections.

Conditions still use exact `==` (viewer) and `startsWith("<collection>__")` (updater), which are safe against collection-name prefix collisions because collection names cannot contain `__`.

### 2. Support unclaimed secret collections
Adds an `unclaimed` flag to `group.Target` for a holding-area rover group (`test-platform-gsm-unclaimed-secrets`) that owns Vault collections no team has claimed yet. For an `unclaimed: true` group, `GetDesiredState` keeps its collections in the active-collection set (so migrated data secrets are **not deleted**) but emits no service account, no SA/index secrets, and no IAM bindings. Claiming a collection later is just moving it under a normal group; the collection stays active throughout, so its data secrets are never deleted in the process. A collection listed under both a normal and the unclaimed group is treated as claimed.

## Testing
- Unit: new `chunked-config` and `unclaimed-config` fixtures for `TestGetDesiredState`; added `TestChunkCollections` and group-binding cases to `TestIsManagedBinding`.
- e2e: `config-create.yaml` now includes a 6-collection group (forces a 5-collection / 11-operator condition through real `SetIamPolicy`); new `config-unclaimed.yaml` + `TestUnclaimedCollectionPreserved` seeds a data secret and asserts it survives while no SA/binding/managed-secret is created for the unclaimed collection.

## Summary

- Updated `gsm-secret-sync` to generate separate service-account and group IAM bindings.
- Limited each group binding to five collections to avoid GCP binding limits.
- Added `unclaimed` collections that retain data secrets but receive no service accounts, managed secrets, or IAM bindings.
- Added unit fixtures, tests, and end-to-end coverage for IAM chunking, large groups, and unclaimed collection preservation.

### DIFF
--- a/pkg/group/config.go
+++ b/pkg/group/config.go
@@ -34,6 +34,10 @@ type Target struct {
 	ClusterGroups []string `json:"cluster_groups,omitempty" yaml:"cluster_groups,omitempty"`
 	// SecretCollections are the secret collections the group has access to.
 	SecretCollections []string `json:"secret_collections,omitempty" yaml:"secret_collections,omitempty"`
+	// Unclaimed marks a group as a holding area for secret collections that no team owns yet.
+	// The gsm-secret-sync reconciler keeps these collections' secrets alive but creates no
+	// updater service account and no IAM bindings for them until they are moved under a normal group.
+	Unclaimed bool `json:"unclaimed,omitempty" yaml:"unclaimed,omitempty"`
 }
 
 func (t Target) ResolveClusters(cg map[string][]string) sets.Set[string] {
--- a/pkg/gsm-secrets/config.go
+++ b/pkg/gsm-secrets/config.go
@@ -14,11 +14,15 @@ import (
 
 // GetDesiredState parses the configuration file and builds the desired state specifications.
 //
-// Each collection referenced by a group gets an updater service account, its SA secret, an index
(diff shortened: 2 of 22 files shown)


psalajova:
/override ci/prow/e2e

the gsm test within it passed, the failure is due to something else

PR #5394, opened 2026-09-04
Pruned registry-replacer

hector-vido:
This is a task from [DPTP-3528](https://redhat.atlassian.net/browse/DPTP-3528).
Functionality of `registry-replacer` was moved into `ci-operator` as we can see on [DPTP-4619](https://redhat.atlassian.net/browse/DPTP-4619).

## Summary

Removed the `registry-replacer` tool from `ci-tools`.

- Deleted its implementation, tests, fixtures, documentation, and container image.
- Removed the binary from the `auto-config-brancher` image.
- Updated configuration validation to run without registry-replacer preprocessing.

CI operators must use the replacement functionality now provided by `ci-operator`.


hector-vido:
/override ci/prow/images - registry-replace will fail since it was removed

hector-vido:
/override ci/prow/images
`registry-replace` will fail since it was removed

PR #5401, opened 2026-09-07
gsm-secrets: make the updater service account per collection opt-in

psalajova:
Every claimed secret collection gets its own updater service account, and with it two more bindings in the `openshift-ci-secrets` project IAM policy. At the ~115 collections the Vault migration will onboard that is 230 bindings and roughly 63KB — on its own enough to push the policy past GCP's ~67KB cap. It is the last thing standing between the policy and the migration.

Almost none of them are wanted. Group members can already write to every collection their group owns, via the group's own updater binding. The service account is for automation that cannot authenticate as a group member. Four collections have one today.

So a group now names the collections it wants one for:

```yaml
test-platform-gsm-secrets-owners:
  secret_collections:
  - psalajova-first-secret
  - test-platform-infra
  updater_service_accounts:
  - psalajova-first-secret
  - test-platform-infra
```

Collections not listed still get their index secret and are still covered by their group's bindings. They just have no service account, no SA secret, and no service-account-scoped bindings.

## Effect

At 115 collections / 50 groups, projected policy size:

| | size | bindings |
|---|---|---|
| service account for every collection | 95.6 KB | 330 |
| service accounts only where asked for | **32.6 KB** | 108 |

## The SA code is kept, not removed

It is unchanged and stays exercised by the collections that do opt in, so granting one to a team later is a one-line config change rather than reviving deleted or commented-out code.

## Withdrawing an account also removes its secret

`DiffSecrets` kept every secret belonging to a live collection, which is right for the collection's own data but left `<collection>__updater-service-account` holding a key that had just been revoked along with the account. Only that secret is affected; data secrets and index secrets are untouched. This also fixes the same latent problem on the existing `unclaimed` path, where the account was already being deleted while its secret stayed behind.

## e2e

Left alone, no e2e config would have asked for a service account, `GetDesiredState` would have returned none, and the comparison would have passed on zero either way — the whole lifecycle would have stopped being exercised. So the configs opt in: `alpha-secrets` asks for one while the six collections beside it do not, and the update step withdraws that account while granting one to `new-secrets`, so creation, key generation and deletion all still run against real GCP. The unclaimed test now has a claimed collection with an account next to the unclaimed one without, so that assertion means something again.

The reconciler binary path is also overridable via `GSM_SECRET_SYNC_BINARY` (defaulting to the container path), which is what makes the suite runnable outside CI.

## Ordering: this must merge before openshift/release#84712

`group.LoadConfig` parses `sync-rover-groups/_config.yaml` strictly, so until this ships and its images are rebuilt the release PR's own presubmits fail on the unknown field. That PR adds `updater_service_accounts` for the three collections whose accounts are in use, and merging it is what fires the reconciler.

One hazard in that order: between this image going live and the release PR merging, the reconciler would run new code against a config where nobody lists `updater_service_accounts`, and would delete all four service accounts. It only fires if something else changes that file in the window, so it wants watching rather than blocking.

## Reviewer notes

- The field is validated as a subset of the group's own `secret_collections`, rejects duplicates, and is rejected on `unclaimed` groups.
- Removing a collection from the list deletes its service account and revokes its keys. Re-adding it regenerates both, but with a **new** key, so anything holding the old one needs updating.

## Summary

The GSM secret collection updater now creates dedicated updater service accounts only for collections listed in `updater_service_accounts`.

- Group configuration validates ownership, duplicate entries, and unclaimed-group usage.
- Claimed collections receive index secrets and group-based IAM bindings.
- Opted-in collections receive updater service accounts, service-account secrets, and scoped IAM bindings.
- Unclaimed collections remain active without service accounts, index secrets, or bindings.
- The updater removes obsolete service-account secrets and bindings when opt-in is removed.
- E2E and fixture coverage verifies the opt-in behavior.


Prucek (approved):

psalajova:
/override e2e

test/e2e/gsm-secret-sync (2m2.499s) passed

Prucek (approved):

PR #5406, opened 2026-09-09
fix: omit visibility from private-org-peribolos-sync output

deepsm007:
- Stop writing `visibility`/`private` in `private-org-peribolos-sync` generated openshift-priv repo config
- `periodic-org-sync` strips `openshift-priv` before `peribolos --fix-repos`; visibility there is unused and caused bulk churn in auto-sync PRs to openshift/config

`private-org-peribolos-sync` no longer generates `visibility` or `private` for `openshift-priv` repositories. This prevents unused fields from causing bulk churn in automated `openshift/config` sync pull requests.

Tests: `go test ./cmd/private-org-peribolos-sync/...`


deepsm007:
Closing: wrong approach. Main already migrated private-org-peribolos-sync to prow Visibility (9a28914d7). Fix is ci-tools-standalone#31 (align determinize-peribolos prow), not dropping visibility from sync output.

PR #5432, opened 2026-09-22
fix(dptp-controller-manager): reduce cache memory usage

jmguzik:
Reduces `dptp-controller-manager` cache memory usage by removing managed fields from cached objects. Secret and ServiceAccount caches are namespace-scoped when only the service-account refresher runs. Registry-cluster caches retain only the newest ImageStream tag event.

The promotion reconciler reads Image objects through an uncached client while keeping ImageStreams cached. It uses the newest ImageStream status tag to determine recency and handles invalid image metadata as a terminal error.

Adds tests for cache trimming, uncached image reads, context propagation, and invalid image metadata.


pkg/controller/promotionreconciler/reconciler.go
```diff
@@ -75,14 +70,14 @@ func AddToManager(mgr controllerruntime.Manager, opts Options) error {
 	log := logrus.WithField("controller", ControllerName)
 	go func() {
 		for delta := range configChangeChannel {
-			if err := handleCIOpConfigChange(opts.RegistryManager.GetClient(), releaseBuildConfigs, prowJobEnqueuer, opts.GitHubClient, delta, log); err != nil {
+			if err := handleCIOpConfigChange(opts.RegistryManager.GetAPIReader(), releaseBuildConfigs, prowJobEnqueuer, opts.GitHubClient, delta, log); err != nil {
```
danilo-gemoli, line 73: Beware that this will put more pressure on the `apiserver`.

cmd/dptp-controller-manager/main.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/sirupsen/logrus"
```
danilo-gemoli, line 16: It's fine but I guess that at some point we might want to refactor this file into a builder class.

